PR #142, opened 2015-07-06
Add type checking utilities to the reference docs

beam2d:
I added type checking utilities to the reference section of the documentation. I also fixes some expressions in the docstrings in this PR.


unnonouno:
I'll check it.

unnonouno:
It looks good, but I found some errors. I'll fix them after merge.

unnonouno:
LGTM

PR #144, opened 2015-07-06
Add nodes to computational graph

delta2323:
Currently, `ComputationalGraph` is essentially a list of edges. So if we run `build_computational_graph` with variables part of which is unchained, the unchained variables are missing from the resulting computational graph. Further, if all of variables are unchained, the graph is empty set and no graph is plotted. This specification is not intuitive. 

This PR adds nodes to computational graph so that such isolated variables are restored.

### DIFF
--- a/chainer/computational_graph.py
+++ b/chainer/computational_graph.py
@@ -63,7 +63,7 @@ class ComputationalGraph(object):
 
         Args:
             nodes (list): List of nodes. Each node is either
-            :class:`Variable` object or :class:`Function` object.
+                 :class:`Variable` object or :class:`Function` object.
             edges (list): List of edges. Each edge consists of pair of nodes.
         """
         self.nodes = nodes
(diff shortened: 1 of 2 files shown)


unnonouno:
Use `assertIn` instead of `assertTrue(... in ...)` for readable error messages.

delta2323:
Since I check equivalence of two sets, I use `assertSetEqual` instead of membership test with `assertTrue`.

chainer/computational_graph.py
```diff
@@ -58,14 +58,15 @@ class ComputationalGraph(object):
       We assume that the computational graph is directed and acyclic.
     """
 
-    def __init__(self, edges):
+    def __init__(self, nodes, edges):
         """Initializes computational graph.
 
         Args:
+            nodes (list): List of nodes. Each node is either
+            :class:`Variable` object or :class:`Function` object.
```
unnonouno: indent?

unnonouno:
LGTM except a comment.

delta2323:
Fixed

unnonouno:
LGTM

PR #145, opened 2015-07-06
Add test decorators that make test cases run multiple times

### LINKED ISSUE #85: Run numerical test multiple times
Related #80 

To reduce undesired failure of numerical tests like `optimizers` or `backward_{cpu,gnu}`, we should add some mechanism in which we run targeted test multiple times and assert fails only if targeted one fails at all attempts.

One of possible implementation is to use decorator like:
```
@test_n_times(10, 8):
def test_something(self):
    ....
```

delta2323:
This PR fixes #85 .
This PR implements decorators for test cases that make them run multiple times.

Decorators are as follows
* repeat_with_success_at_least(n, m) : Pass the test case if it succeeded m times among n trials.
* retry(n) : Pass the test case if it succeeded at least once among n trials.
* repeat(n) : Pass the test case if it succeeded all n trials.

### DIFF
--- a/chainer/testing/condition.py
+++ b/chainer/testing/condition.py
@@ -13,15 +13,15 @@ class QuietTestRunner(object):
 
 
 def repeat_with_success_at_least(times, min_success):
-    """Decorator for multiple trial of the test case
+    """Decorator for multiple trial of the test case.
 
-    Decorated test case is launched multiple times.
+    The decorated test case is launched multiple times.
     The case is judged as passed at least specified number of trials.
     If the number of successful trials exceeds `min_success`,
     the remaining trials are skipped.
 
     Args:
-        times(int): The number of trials
+        times(int): The number of trials.
         min_success(int): Threshold that the decorated test
             case is regarded as passed.
 
--- a/chainer/testing/condition.py
+++ b/chainer/testing/condition.py
@@ -11,11 +11,10 @@ class QuietTestRunner(object):
         return result
 
 
-def success_at_least(times, min_success):
+def repeat_with_success_at_least(times, min_success):
     assert times >= min_success
 
-    def _multiple(f):
-
+    def _repeat_with_success_at_least(f):
         @functools.wraps(f)
         def wrapper(*args, **kwargs):
             assert len(args) > 0
@@ -39,5 +38,6 @@ def success_at_least(times, min_success):
     return _repeat_with_success_at_least
 
 
-def all_success(times):
-    return success_at_least(times, times)
+def repeat(times):
+    return repeat_with_success_at_least(times, times)
+
--- a/chainer/testing/condition.py
+++ b/chainer/testing/condition.py
@@ -33,8 +33,8 @@ def repeat_with_success_at_least(times, min_success):
         @functools.wraps(f)
         def wrapper(*args, **kwargs):
             assert len(args) > 0
-            cls = args[0]
-            assert isinstance(cls, unittest.TestCase)
+            instance = args[0]
+            assert isinstance(instance, unittest.TestCase)
             success_counter = 0
             failure_counter = 0
             for _ in six.moves.range(times):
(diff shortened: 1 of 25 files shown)


chainer/testing/condition.py
```diff
@@ -0,0 +1,46 @@
+import functools
+import unittest
+
+import six
+
+
+class QuietTestRunner(object):
+    def run(self, suite):
```
unnonouno: Insert a blank line after `class`

chainer/testing/condition.py
```diff
@@ -0,0 +1,46 @@
...
+                    unittest.FunctionTestCase(
+                        lambda: f(*args, **kwargs),
+                        setUp=cls.setUp,
+                        tearDown=cls.tearDown))
+                if QuietTestRunner().run(suite).wasSuccessful():
+                    success_counter += 1
+                if success_counter >= min_success:
+                    cls.assertTrue(True)
+                    return
+            cls.fail()
```
unnonouno: Is it possible to show an original error message?

chainer/testing/condition.py
```diff
@@ -0,0 +1,46 @@
...
+    def run(self, suite):
+        result = unittest.TestResult()
+        suite(result)
+        return result
+
+
+def repeat_with_success_at_least(times, min_success):
+    assert times >= min_success
+
+    def _repeat_with_success_at_least(f):
```
unnonouno: Do you have a test for this function?

unnonouno:
The code looks good. Please write a document about its specification, and make a unit test.

delta2323:
I added documents and unit tests for the decorators.

chainer/testing/condition.py
```diff
@@ -0,0 +1,86 @@
...
+    def run(self, suite):
+        result = unittest.TestResult()
+        suite(result)
+        return result
+
+
+def repeat_with_success_at_least(times, min_success):
+    """Decorator for multiple trial of the test case
+
+    Decorated test case is launched multiple times.
```
unnonouno: "A decorated test case" ?

chainer/testing/condition.py
```diff
@@ -0,0 +1,86 @@
...
+class QuietTestRunner(object):
+
+    def run(self, suite):
+        result = unittest.TestResult()
+        suite(result)
+        return result
+
+
+def repeat_with_success_at_least(times, min_success):
+    """Decorator for multiple trial of the test case
```
unnonouno: Is "." required?

tests/testing_tests/test_condition.py
```diff
@@ -0,0 +1,111 @@
...
+        _should_fail(self, f)
+
+    def test_all_trials_fail2(self):
+        f = self._decorate(MockUnitTest.failure_case, 10, 0)
+        _should_pass(self, f)
+
+    def test_all_trials_succeed(self):
+        f = self._decorate(MockUnitTest.success_case, 10, 10)
+        _should_pass(self, f)
+
```
unnonouno: Can you check the count of the trials. I mean `self.assertEqual(self.unit_test.counter, 10)`.

chainer/testing/condition.py
```diff
@@ -0,0 +1,86 @@
...
+
+    Decorated test case is launched multiple times.
+    The case is judged as passed at least specified number of trials.
+    If the number of successful trials exceeds `min_success`,
+    the remaining trials are skipped.
+
+    Args:
+        times(int): The number of trials
```
unnonouno: "." ?

chainer/testing/condition.py
```diff
@@ -0,0 +1,86 @@
...
+        suite(result)
+        return result
+
+
+def repeat_with_success_at_least(times, min_success):
+    """Decorator for multiple trial of the test case
+
+    Decorated test case is launched multiple times.
+    The case is judged as passed at least specified number of trials.
+    If the number of successful trials exceeds `min_success`,
```
unnonouno: "``" ?

unnonouno:
LGTM except the comments. And, note that tests show no adequate error messages, as they only show "fail".

unnonouno:
LGTM

PR #146, opened 2015-07-06
It is fix batchsize in training and evaluation.

airyou:
When batchsize is not chosen appropriately, accuracy may exceed 1.


delta2323:
LGTM :+1:

delta2323:
Thank you for the patch!

PR #148, opened 2015-07-06
Support a recent specification of caffe

rezoo:
This PR aims to support a recent specification of caffemodel.
A current ``chainer.functions.caffe.CaffeFunction`` does not support the ``.caffemodel`` created by the latest caffe because the ``CaffeFunction`` assumes that there exists positive values
in ``height`` and ``width`` fields in ``Blob`` (the current caffe explicitly uses the ``shape`` field in order to support a two-dimensional array).
Although the original code works well when we load an old version of ``.caffemodel`` (i.e., bvlc_reference_caffenet.caffemodel), I wrote an additional code for supporting the latest model.

Please note that I do not add any test files for this PR just like the original ``caffe_function.py`` (I tested it on my laptop and confirmed that it works correctly, though).

### DIFF
--- a/chainer/functions/caffe/caffe_function.py
+++ b/chainer/functions/caffe/caffe_function.py
@@ -221,9 +221,11 @@ class CaffeFunction(function.Function):
         ksize = _get_ksize(param)
         stride = _get_stride(param)
         pad = _get_pad(param)
+        num = _get_num(blobs[0])
+        channels = _get_channels(blobs[0])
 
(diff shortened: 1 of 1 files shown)


beam2d:
Thank you for the fix. I'll reivew it.

beam2d:
LGTM! I've also checked the fix by Caffe MNIST example, and confirmed that it works correctly.

PR #149, opened 2015-07-07
Fix installation and READMEs on it

beam2d:
I fixed following things:
- chainer.testing is now installed
- Add dependency on Nose (chainer.testing requires it)
- Move examples-specific dependencies to appropriate per-example READMEs
- Refactor the requirements section in README


unnonouno:
LGTM

PR #157, opened 2015-07-07
Fix buf: if arguments is a sections in SplitAxis

okuta:
I fixed `foward_gpu` function.

### DIFF
--- a/chainer/functions/split_axis.py
+++ b/chainer/functions/split_axis.py
@@ -27,17 +27,6 @@ class SplitAxis(function.Function):
         self.axis = axis
 
     def forward_cpu(self, x):
-        if isinstance(self.indices_or_sections, collections.Iterable):
-            cdimx = x[0].shape[self.axis]
-            ind = list(self.indices_or_sections)
-            ind.append(cdimx)
-            prev_i = 0
-            for i in ind:
-                cdimy = max(0, min(i, cdimx) - prev_i)
-                if cdimy == 0:
-                    raise ValueError('Not support if shape contains 0')
-                prev_i = i
-
         return tuple(numpy.split(x[0], self.indices_or_sections, self.axis))
 
     def forward_gpu(self, x):
@@ -63,9 +52,8 @@ class SplitAxis(function.Function):
             s = list(xshape)
             s[self.axis] = cdimy
             y = cuda.empty(s, dtype=x[0].dtype)
-            if cdimy == 0:
-                raise ValueError('Not support if shape contains 0')
-            kernel(y, x[0], cdimy, self.cdimx, self.rdim, prev_i)
+            if cdimy != 0:
+                kernel(y, x[0], cdimy, self.cdimx, self.rdim, prev_i)
             prev_i = i
             ys.append(y)
         return tuple(ys)
--- a/chainer/functions/split_axis.py
+++ b/chainer/functions/split_axis.py
@@ -80,7 +80,7 @@ def split_axis(x, indices_or_sections, axis):
 
     Args:
         x (tuple of Variables): Variables to be split.
-        indices_or_sections (int or 1-Darray): If this argument is an integer,
+        indices_or_sections (int or 1-D array): If this argument is an integer,
             N, the array will be divided into N equal arrays along axis.
             If it is a 1-D array of sorted integers, the entries
             indicate where along which axis the array is split.
--- a/chainer/functions/split_axis.py
+++ b/chainer/functions/split_axis.py
@@ -36,26 +36,26 @@ class SplitAxis(function.Function):
 
         if isinstance(self.indices_or_sections, collections.Iterable):
             ind = list(self.indices_or_sections)
-            ind.append(xshape[self.axis])
+            ind.append(self.cdimx)
         else:
             sec = self.indices_or_sections
-            if xshape[self.axis] % sec:
+            if self.cdimx % sec:
                 raise ValueError(
                     'array split does not result in an equal division')
-            ind = numpy.arange(1, sec + 1) * (xshape[self.axis] // sec)
+            ind = numpy.arange(1, sec + 1) * (self.cdimx // sec)
         ys = []
         kernel = cuda.elementwise(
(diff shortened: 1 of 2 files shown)


delta2323:
I'll check it

chainer/functions/split_axis.py
```diff
@@ -33,21 +32,20 @@ def forward_gpu(self, x):
...
-        indices = self.indices_or_sections
-        if isinstance(indices, collections.Iterable):
-            indices = list(indices)
-            indices.append(xshape[self.axis])
+        if isinstance(self.indices_or_sections, collections.Iterable):
+            ind = list(self.indices_or_sections)
+            ind.append(xshape[self.axis])
         else:
-            if xshape[self.axis] % indices:
+            sec = self.indices_or_sections
```
delta2323: It is better to check if `self.dices_or_sections` is integer and raise an error if this assumption is not satisfied.

chainer/functions/split_axis.py
```diff
@@ -33,21 +32,20 @@ def forward_gpu(self, x):
         self.cdimx = xshape[self.axis]
         self.rdim = numpy.prod(xshape[self.axis + 1:])
 
-        indices = self.indices_or_sections
-        if isinstance(indices, collections.Iterable):
-            indices = list(indices)
-            indices.append(xshape[self.axis])
+        if isinstance(self.indices_or_sections, collections.Iterable):
+            ind = list(self.indices_or_sections)
+            ind.append(xshape[self.axis])
```
delta2323: xshape[self.axis] is assigned to `self.cdimx`.

chainer/functions/split_axis.py
```diff
@@ -33,21 +32,20 @@ def forward_gpu(self, x):
...
         else:
-            if xshape[self.axis] % indices:
+            sec = self.indices_or_sections
+            if xshape[self.axis] % sec:
```
delta2323: `xshape[self.axis]` is assigned to `self.cdimx`.

chainer/functions/split_axis.py
```diff
@@ -33,21 +32,20 @@ def forward_gpu(self, x):
...
                 raise ValueError(
                     'array split does not result in an equal division')
-            indices = six.moves.range(
-                indices, xshape[self.axis] + indices, indices)
+            ind = numpy.arange(1, sec + 1) * (xshape[self.axis] // sec)
```
delta2323: `xshape[self.axis]` is assigned to `self.cdimx`.

chainer/functions/split_axis.py
```diff
@@ -33,21 +32,20 @@ def forward_gpu(self, x):
...
         ys = []
         kernel = cuda.elementwise(
             _args, 'COPY(y[i] = x[idx])', 'split_fwd', preamble=_preamble)
         bi = 0
-        for i in indices:
+        for i in ind:
             i = min(i, xshape[self.axis])
```
delta2323: `xshape[self.axis]` is assigned to `self.cdimx`.

chainer/functions/split_axis.py
```diff
@@ -33,21 +32,20 @@ def forward_gpu(self, x):
...
+            if xshape[self.axis] % sec:
                 raise ValueError(
                     'array split does not result in an equal division')
-            indices = six.moves.range(
-                indices, xshape[self.axis] + indices, indices)
+            ind = numpy.arange(1, sec + 1) * (xshape[self.axis] // sec)
         ys = []
         kernel = cuda.elementwise(
             _args, 'COPY(y[i] = x[idx])', 'split_fwd', preamble=_preamble)
         bi = 0
```
delta2323: It is better that the meaning of this variable should be clear from its name (like `i_prev`, `prev_i`)

delta2323:
Though it is not related to this PR directly, I found that we should add test case where `indices_or_section` is a list that contains one or more 0's. In that case, the output of forward is # of non-zero elements in `indices_or_sections` (i.e. tensors that corresponding to 0 are removed). I do not say it is not intuitive. But I think it is not obvious from function's name.

okuta:
The numpy supports zero size array.
But, pycuda partial supports zero size array.
`chainer.cuda.empty((1,0,1))` -> `array([], shape=(1, 0, 1), dtype=float32)`
`chainer.cuda.empty((1,0,1)) + 1` -> error

Therefore, I did not add test cases where indices_or_sections is a list that contains one or more 0's.

chainer/functions/split_axis.py
```diff
@@ -22,10 +21,23 @@ class SplitAxis(function.Function):
...
     def forward_cpu(self, x):
+        if isinstance(self.indices_or_sections, collections.Iterable):
+            cdimx = x[0].shape[self.axis]
+            ind = list(self.indices_or_sections)
+            ind.append(cdimx)
+            prev_i = 0
+            for i in ind:
+                cdimy = max(0, min(i, cdimx) - prev_i)
+                if cdimy == 0:
+                    raise ValueError('Not support if shape contains 0')
```
delta2323: More explanatory error is appropriate.
Output tensors whose shapes contain 0 are not supported.

chainer/functions/split_axis.py
```diff
@@ -33,29 +45,28 @@ def forward_gpu(self, x):
...
+        for i in ind:
+            cdimy = max(0, min(i, self.cdimx) - prev_i)
             s = list(xshape)
             s[self.axis] = cdimy
             y = cuda.empty(s, dtype=x[0].dtype)
-            if cdimy != 0:
-                kernel(y, x[0], cdimy, self.cdimx, self.rdim, bi)
-            bi = i
+            if cdimy == 0:
+                raise ValueError('Not support if shape contains 0')
```
delta2323: Same as the above.

delta2323:
Please add the test case where at least one of the output tensors has shape that contains 0. In this case , `forward` should raise `ValueError`.

okuta:
I separated `ValueError`.

delta2323:
LGTM :+1:

delta2323:
Good job! As we discussed above, please open new PR about zero-length output.

PR #158, opened 2015-07-08
Stop using OpenCV in examples

### LINKED ISSUE #88: Stop using OpenCV to just reading a JPEG file in the ImageNet example
Current ImageNet example uses the cv2 package to just load JPEG image files. This is overkill.

okuta:
I fixed #88

### DIFF
--- a/examples/imagenet/README.md
+++ b/examples/imagenet/README.md
@@ -1,9 +1,5 @@
 # Large Scale ConvNets
 
-## Requirements
-
-- OpenCV 2.4
-
 ## Description
 
 This is an experimental example of learning from the ILSVRC2012 classification dataset.
--- a/examples/imagenet/README.md
+++ b/examples/imagenet/README.md
@@ -1,5 +1,9 @@
 # Large Scale ConvNets
 
+## Requirements
+
+- Pillow
+
 ## Description
 
 This is an experimental example of learning from the ILSVRC2012 classification dataset.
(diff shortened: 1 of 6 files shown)


examples/imagenet/README.md
```diff
@@ -1,9 +1,5 @@
 # Large Scale ConvNets
 
-## Requirements
-
-- OpenCV 2.4
```
unnonouno: We need to add `pillow` instead of `opencv`

setup.py
```diff
@@ -17,6 +17,7 @@
               'chainer.utils'],
     install_requires=['nose',
                       'numpy',
+                      'pillow',
```
unnonouno: `pillow` is only required for examples. You don't need to add it here.

unnonouno:
We will need to fix the test for examples after this PR. https://github.com/pfnet/chainer-test/blob/master/test_example.sh

examples/modelzoo/evaluate_caffe_net.py
```diff
@@ -92,8 +92,7 @@ def forward(x, t):
 accum_loss = 0
 accum_accuracy = 0
 for path, label in dataset:
-    # TODO(beam2d): Remove dependency on OpenCV
-    image = cv2.imread(path).transpose(2, 0, 1)
+    image = np.asarray(Image.open(path)).transpose(2, 0, 1)[::-1]
```
unnonouno: Is `[2,1,0]` required?

okuta: It uses `[::-1]` instead of `[2,1,0]`

unnonouno: oh, i see.

unnonouno:
LGTM

PR #162, opened 2015-07-08
Add condition.retry decorator to undecorated numerical tests

delta2323:
This PR adds `condition.retry` decorator to following functions.

* `TestBinaryHierarchicalSoftmax.test_sum`
* All of numerical test fixtures of `matmul`, non-parameterized `convolution_2d`, non-parameterized `linear` and `split_axis`.

### DIFF
--- a/tests/functions_tests/test_concat.py
+++ b/tests/functions_tests/test_concat.py
@@ -7,7 +7,6 @@ from chainer import cuda
 from chainer import functions
 from chainer import gradient_check
 from chainer.testing import attr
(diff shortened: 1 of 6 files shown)


okuta:
`split_axis` uses only copy operation.
I think it is better to remove `condition.retry`.

delta2323:
I removed `retry` decorator from tests of `split_axis` . The situation is similar in `concat`. So I remove the decorator from `concat`, too.

okuta:
LGTM!

PR #165, opened 2015-07-08
Add missing explicit dtype to accum los variable.

umitanuki:
It might have been working without problems for this example data, but
technically speaking, by omitting dtype in the zeros() call the grad
could become too small beyond float32 precision.  As a result, it would
stop updating function parameters if you run it in GPU, as the grad
values just vanish.

### DIFF
--- a/examples/ptb/train_ptb.py
+++ b/examples/ptb/train_ptb.py
@@ -114,7 +114,7 @@ epoch = 0
 start_at = time.time()
 cur_at = start_at
 state = make_initial_state()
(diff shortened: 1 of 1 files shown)


okuta:
Thank you for finding and fixing the inconsistency!

Current `chainer.cuda.zeros` implementation is
`def zeros(shape, dtype=numpy.float32, stream=None)`.

Therefore,
`numy.zeros(1).dtype` is `dtype('float64')`.
`chainer.cuda.zeros(1).dtype` is  `dtype('float32')`.

PR #175, opened 2015-07-10
Fix document of split_axis

delta2323:
* Fix return value of `split_axis`
* Add note about zero-size outputs of `split_axis`
* Fix minor grammatical errors.


okuta:
LGTM!

PR #178, opened 2015-07-11
API: Consistent to_gpu/to_cpu API

sinhrks:
I felt inconsistencies in following differences.

- array-like must be moved to CPU/GPU via ``cuda.to_cpu`` / ``cuda.to_gpu``.
- ``Function`` and ``FunctionSet`` has its own ``to_cpu`` and ``to_gpu``. But ``Variable`` doesn't.

```
import chainer
f = chainer.FunctionSet(f=chainer.functions.Linear(2, 3))
chainer.cuda.to_gpu(f)
# AttributeError: 'FunctionSet' object has no attribute 'shape'

import numpy as np
chainer.Variable(np.array([1])).to_gpu()
# AttributeError: 'Variable' object has no attribute 'to_gpu'
```

This PR makes:

- All ``chainer`` instance movable between CPU/GPU  should have ``to_cpu`` and ``to_gpu``. In case of ``Variable``, internal data is moved to CPU/GPU resetting progressing calculation status.
- ``cuda.to_cpu`` and ``cuda.to_gpu`` should accept ``chainer`` instances.

### DIFF
--- a/chainer/cuda.py
+++ b/chainer/cuda.py
@@ -457,6 +457,9 @@ def to_gpu(array, device=None):
         copy GPUArray into specified device.
 
     """
+    if hasattr(array, 'to_gpu'):
+        return array.to_gpu(device=device)
+
     if isinstance(array, GPUArray):
         return array
     with using_device(device):
@@ -511,6 +514,9 @@ def to_cpu(array):
         ``array`` without performing any copy.
 
     """
+    if hasattr(array, 'to_cpu'):
+        return array.to_cpu()
+
     if isinstance(array, GPUArray):
         return array.get()
     return array
--- a/chainer/variable.py
+++ b/chainer/variable.py
@@ -45,8 +45,8 @@ class Variable(object):
         """Initializes a variable.
 
         Args:
-            data (:class:`~numpy.ndarray` or \
-                :class:`~pycuda.gpuarray.GPUArray`):
+            data (:class:`~numpy.ndarray`,  \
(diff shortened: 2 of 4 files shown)


beam2d:
Thank you for thinking about a better API around the CPU/GPU migration.

> All `chainer` instance movable between CPU/GPU should have `to_cpu` and `to_gpu`. In case of `Variable`, internal data is moved to CPU/GPU resetting progressing calculation status.

We should be careful about the interface of `Variable`. I think all computations on `Variable` except graph manipulations should be recorded into the computational graph. If you want to add `to_cpu` and `to_gpu` to `Variable`, I think it is better to make its own `Function` and make these methods a short cut for this `Function` call.

> `cuda.to_cpu` and `cuda.to_gpu` should accept chainer instances.

I don't think it is confusing that `cuda.to_cpu` and `cuda.to_gpu` do not accept chainer's objects, since chainer.cuda is purely an API for the CUDA array backend. There is also a difference between `cuda.to_cpu`/`cuda.to_gpu` and `FunctionSet.to_cpu`/`FunctionSet.to_gpu`: the former are copying functions, while the latter is in-place ones. We cannot use them interchangeably, so we cannot mix them into the same functions.

beam2d:
I added Variable.to_cpu and Variable.to_gpu at #573 (sorry, I forgot this PR at this point). Since we have CuPy now, we should rethink about the purpose of `chainer.cuda.to_cpu` and `chainer.cuda.to_gpu`. Anyway, since this PR is too old, I close it here. If you want to add it to the current implementation, please reopen or make a new PR. Thank you!

PR #188, opened 2015-07-14
Batch Normalization typecheck

delta2323:
* Add typecheck to BatchNormalization

Related to #123

### DIFF
--- a/chainer/functions/batch_normalization.py
+++ b/chainer/functions/batch_normalization.py
@@ -151,7 +151,7 @@ class BatchNormalization(function.Function):
 
     def check_type_backward(self, in_types, out_types):
         type_check.expect(out_types.size() == 1)
-        x_type, = out_types
+        x_type, = in_types
         y_type, = out_types
 
         type_check.expect(
--- a/chainer/functions/batch_normalization.py
+++ b/chainer/functions/batch_normalization.py
@@ -2,6 +2,7 @@ import numpy
 
 from chainer import cuda
 from chainer import function
+from chainer.utils import type_check
 
 
 def _kernel_with_I(args, expr, name):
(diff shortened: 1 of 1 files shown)


chainer/functions/batch_normalization.py
```diff
@@ -91,6 +92,7 @@ class BatchNormalization(function.Function):
     gradient_names = ('ggamma', 'gbeta')
 
     def __init__(self, size, decay=0.9, eps=1e-5):
+        self.size = size
```
beam2d: How about forcing `self.size` to be a tuple? E.g.: `self.size = size if isinstance(size, tuple) else (size,)`?

chainer/functions/batch_normalization.py
```diff
@@ -130,6 +132,35 @@ def __call__(self, x, test=False, finetune=False):
         self.is_finetune = finetune
         return function.Function.__call__(self, x)
 
+    def check_type_forward(self, in_types):
+        type_check.expect(in_types.size() == 1)
+        x_type, = in_types
+
+        if isinstance(self.size, tuple):
+            feature_shape = type_check.Variable(self.size, 'feature_shape')
```
beam2d: How about making `self` as a variable (e.g. named `self_`) instead of feature_shape?

chainer/functions/batch_normalization.py
```diff
@@ -130,6 +132,35 @@ def __call__(self, x, test=False, finetune=False):
...
+            feature_ndim = len(self.size)
+        else:
+            feature_shape = type_check.Variable((self.size,), 'feature_shape')
+            feature_ndim = 1
+
+        type_check.expect(
+            x_type.dtype == numpy.float32,
+            x_type.ndim >= type_check.Variable(
+                feature_ndim, 'feature_ndim') + 1,
```
beam2d: Combined with the above comment, how about using `self_.size.__len__()` instead of the `feature_ndim` variable?

beam2d:
I wrote some comments.

beam2d:
LGTM :+1:

beam2d:
Thank you!

PR #189, opened 2015-07-14
Add unittest(cpu-only) and typecheck for Copy

delta2323:
* Add type check for `Copy` Function 
* Add unittest(cpu-only) for `Copy` Function to test type check.

Related to #123

### DIFF
--- a/chainer/functions/copy.py
+++ b/chainer/functions/copy.py
@@ -1,5 +1,3 @@
-import numpy
-
 from chainer import cuda
 from chainer import function
 from chainer.utils import type_check
--- a/chainer/functions/copy.py
+++ b/chainer/functions/copy.py
@@ -14,8 +14,7 @@ class Copy(function.Function):
 
     def check_type_forward(self, in_types):
         type_check.expect(
-            in_types.size() == 1,
-            in_types[0].dtype == numpy.float32
+            in_types.size() == 1
         )
 
     def check_type_backward(self, in_types, out_types):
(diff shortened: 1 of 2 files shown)


chainer/functions/copy.py
```diff
@@ -9,6 +12,20 @@ class Copy(function.Function):
     def __init__(self, out_device):
         self.out_device = out_device
 
+    def check_type_forward(self, in_types):
+        type_check.expect(
+            in_types.size() == 1,
+            in_types[0].dtype == numpy.float32
```
unnonouno: Does it need to restrict input type?

tests/functions_tests/test_copy.py
```diff
@@ -0,0 +1,27 @@
...
+
+    def setUp(self):
+        self.x_data = numpy.random.uniform(
+            -1, 1, (10, 5)).astype(numpy.float32)
+        self.gy = numpy.random.uniform(-1, 1, (10, 5)).astype(numpy.float32)
+
+    def test_check_forward_cpu(self):
+        x = chainer.Variable(self.x_data)
+        y = functions.copy(x, -1)
+        gradient_check.assert_allclose(self.x_data, y.data, atol=0, rtol=0)
```
unnonouno: I think values are exactly the same?

delta2323: atol and troll is 0, so I think we do not admit any torelance.

unnonouno:
LGTM

PR #202, opened 2015-07-16
Add typecheck to split axis

delta2323:
Related #123

### DIFF
--- a/chainer/functions/split_axis.py
+++ b/chainer/functions/split_axis.py
@@ -4,6 +4,8 @@ import numpy
 
 from chainer import cuda
 from chainer import function
+from chainer.utils import type_check
+
 
 _args = 'float* y, float* x, int cdimy, int cdimx, int rdim, int coffset'
 _preamble = '''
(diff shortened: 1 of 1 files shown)


chainer/functions/split_axis.py
```diff
@@ -26,6 +29,50 @@ def __init__(self, indices_or_sections, axis):
         self.indices_or_sections = indices_or_sections
         self.axis = axis
 
+    def check_type_forward(self, in_types):
+        type_check.expect(
+            in_types.size() == 1,
+            in_types[0].ndim >= self.axis
```
unnonouno: Don't use `in_types[0]` before you check `in_types.size() == 1`.

unnonouno: Why not use `Variable` for `self.axis`?

chainer/functions/split_axis.py
```diff
@@ -26,6 +29,50 @@ def __init__(self, indices_or_sections, axis):
...
+        )
+
+        if isinstance(self.indices_or_sections, collections.Iterable):
+            max_index = type_check.Variable(
+                self.indices_or_sections[-1], 'max_index')
+            type_check.expect(in_types[0].shape[self.axis] >= max_index)
```
beam2d: Is this allowed to have the dimension equal to the `max_index`?

chainer/functions/split_axis.py
```diff
@@ -26,6 +29,50 @@ def __init__(self, indices_or_sections, axis):
...
+    def check_type_backward(self, in_types, out_types):
+        if isinstance(self.indices_or_sections, collections.Iterable):
+            num_indices = type_check.Variable(
+                len(self.indices_or_sections), 'len(indices)')
+            output_size = num_indices + 1
+        else:
+            output_size = type_check.Variable(
+                self.indices_or_sections, 'sections')
+
+        type_check.expect(out_types.size() == output_size.eval())
```
unnonouno: You don't need to call `eval()`.

chainer/functions/split_axis.py
```diff
@@ -26,6 +29,50 @@ def __init__(self, indices_or_sections, axis):
...
+
+        for out_type in out_types:
+            type_check.expect(
+                out_type.ndim == in_types[0].ndim,
+                out_type.dtype == out_types[0].dtype
```
unnonouno: `in_types[0].dtype`?

chainer/functions/split_axis.py
```diff
@@ -26,6 +29,50 @@ def __init__(self, indices_or_sections, axis):
...
+        for out_type in out_types:
+            for i in six.moves.range(in_types[0].ndim.eval()):
+                if i == self.axis:
+                    continue
+                type_check.expect(out_type.shape[i] == in_types[0].shape[i])
+
+        total_output_dim = sum(
+            out_type.shape[self.axis] for out_type in out_types)
+        type_check.expect(
+            total_output_dim == in_types[0].shape[self.axis]
```
unnonouno: Do we need to check each value of `out_type.shape[self.axis]`?

chainer/functions/split_axis.py
delta2323: I made a mistake. If the equality holds, last tensor of output is empty, which is forbidden.

chainer/functions/split_axis.py
delta2323: As discussed in #209, `check_type_backward` is no longer necessary. So I remove this function.

chainer/functions/split_axis.py
delta2323: I just intended to check that all out_types have same dtypes.

delta2323: As discussed in #209, `check_type_backward` is no longer necessary. So I remove this function.

chainer/functions/split_axis.py
delta2323: I understand that the check is somewhat loose.

delta2323: As discussed in #209, `check_type_backward` is no longer necessary. So I remove this function.

unnonouno:
LGTM

PR #203, opened 2015-07-16
typecheck for Reshape Function

delta2323:
Related to #123

### DIFF
--- a/chainer/functions/reshape.py
+++ b/chainer/functions/reshape.py
@@ -11,26 +11,19 @@ class Reshape(function.Function):
     def __init__(self, shape):
         self.shape = shape
 
-    def check_type_forward(self, in_type):
-        type_check.expect(in_type.size() == 1)
-        x_type, = in_type
-
-        in_shape_size = type_check.Variable(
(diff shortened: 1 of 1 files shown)


delta2323:
This PR currently cannot be merged since it needs changes of #204 .

beam2d:
#204 has a new interface that should be deferred to the next minor release. We want to merge this PR before #204, so would you define a local one for prod?

delta2323:
@beam2d I agree with you. For now, I add `prod` to `Reshape` class.

chainer/functions/reshape.py
```diff
@@ -1,13 +1,31 @@
+import numpy
+
 from chainer import function
+from chainer.utils import type_check
 
 
 class Reshape(function.Function):
+    type_check_prod = type_check.Variable(numpy.prod, 'prod')
```
beam2d: Do not write anything above the docstring. How about moving it to module level and name it like `_type_check_prod`?

beam2d:
LGTM except one comment.

beam2d:
[LGTM](http://lgtm.in/i/yPb27nh7q)

PR #204, opened 2015-07-16
Add type_check.prod

delta2323:
Typical usage (type check of `Reshape.forward` where size of input shape must equal to that of self.shape)

```
    type_check.expect(
        in_types.size() == 1,
        type_check.prod(in_types[0].shape) ==
        type_check.prod(self.shape)
)
```

Related to #203


beam2d:
Thank you!

Issue #209, opened 2015-07-16
check_type_backward may be removed


beam2d:
Current type-checking system allows one to define type checking on backward by `Function.check_type_backward`. Type checking at forward makes users easy to debug the type error in their network definition. On the other hand, almost all type errors on backward are not due to the users' failure but due to the bugs of Functions. I want to keep the type-checking system a debugging tool for users, not for developers. Writers of Function definition must write unit tests for its failures instead of type checking on backward.

There is only one exception for the above notion: a user can cause a type error on backward by setting wrong `grad` of the root variable (i.e., the variable whose `backward` the user calls). In such case, what we have to do is checking whether its dtype and shape match to those of `data`. This can be done at the setter of `grad`, which is far more helpful for the debugging.

As a consequence, I suggest to modify the interface as following:

* Remove `Function.check_type_backward`.
* Make the `Variable.grad` a property whose setter checks the dtype and shape of the argument.

Clarifying the types of outputs is future work. It may be done by document, though it cannot be verified automatically.

beam2d:
This issue is based on an offline conversation with @unnonouno (thank you!).

delta2323:
Now I am implementing type checking of various functions in #123. As discussed in this issue, we do not have to iplement `check_type_backward` from later on. Is my understanding correct?

PR #210, opened 2015-07-16
Check type at setting Variable.grad

beam2d:
It fixes the second point of #209. `Variable.grad` is now a property whose setter checks the type, dtype, and shape of the argument. This PR also contains several fixes on Function implementations and some tests.

### DIFF
--- a/chainer/variable.py
+++ b/chainer/variable.py
@@ -67,7 +67,7 @@ class Variable(object):
         self.volatile = volatile
 
         self.splitter = weakref.ref(lambda: 0)  # dead ref
-        self.grad = None
+        self._grad = None
         self.creator = None
 
     def __pos__(self):
@@ -89,6 +89,24 @@ class Variable(object):
             return str(self.data.dtype)
         return '%s, %s' % (str(self.data.shape),
                            str(self.data.dtype))
+
+    @property
+    def grad(self):
+        return self._grad
+
+    @grad.setter
+    def grad(self, g):
+        if g is not None:
+            if type(g) != type(self.data):
+                raise TypeError('Type of data and grad mismatch: %s != %s'
+                                % (type(self.data), type(g)))
+            if g.dtype != self.data.dtype:
+                raise TypeError('Dtype of data and grad mismatch: %s != %s'
+                                % (self.data.dtype, g.dtype))
+            if g.shape != self.data.shape:
+                raise ValueError('Shape of data and grad mismatch: %s != %s'
+                                 % (self.data.shape, g.shape))
+        self._grad = g
 
     def set_creator(self, gen_func):
         """Notifies the variable that the given function is its creator.
--- a/chainer/variable.py
+++ b/chainer/variable.py
@@ -96,16 +96,24 @@ class Variable(object):
 
     @grad.setter
     def grad(self, g):
+        error_msg = '''
+This error is occured in two cases. The first case is when the user manually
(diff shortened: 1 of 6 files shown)


chainer/variable.py
```diff
@@ -90,6 +90,24 @@ def label(self):
...
 
+    @property
+    def grad(self):
+        return self._grad
+
+    @grad.setter
+    def grad(self, g):
+        if g is not None:
+            if not isinstance(g, type(self.data)):
+                raise TypeError('Type of data and grad mismatch: %s != %s'
```
unnonouno: How do users fix this problem?
It's a bug?

tests/test_variable.py
```diff
@@ -137,5 +140,30 @@ def test_invalid_value_type(self):
         with self.assertRaises(AssertionError):
             chainer.Variable(1)
 
+    def test_grad_type_check_pass(self):
+        a = chainer.Variable(np.empty((3,), dtype=np.float32))
+        a.grad = np.ndarray((3,), dtype=np.float32)
+
+    def test_grad_type_check_type(self):
```
unnonouno: How about making a test for `dtype` and a test for `shape`?

unnonouno:
I wrote some comments.

chainer/variable.py
beam2d: It depends on the case. If the user manually sets the grad and its type is wrong, then the user should fix this problem. If there is a function returns a grad with the wrong type, it's a bug. The user may distinguish these situations by watching the stack trace, though is a bit complicated task. I have no idea to distinguish them at this line.

tests/test_variable.py
beam2d: `test_grad_type_check_dtype` and `test_grad_type_check_shape` are the ones. Aren't they enough?

beam2d:
I updated the error messages of the grad setter.

tests/test_variable.py
unnonouno: Ok, no problem.

unnonouno:
It looks good to me. How do you think about broad casting problem?

tests/test_variable.py
```diff
@@ -137,5 +140,30 @@ def test_invalid_value_type(self):
         with self.assertRaises(AssertionError):
             chainer.Variable(1)
 
+    def test_grad_type_check_pass(self):
+        a = chainer.Variable(np.empty((3,), dtype=np.float32))
+        a.grad = np.ndarray((3,), dtype=np.float32)
```
delta2323: Is there any tool that is equivalent to `assert_nothing_raised` in Ruby.
I am wondering if this test case can handle the exception when this assignment raises it.

beam2d: AFAIK, there is no such function that checks if the code raises nothing. When the code raises an exception without catch, the test is failed with the stack trace as same as usual assertion failures.

beam2d:
@unnonouno You mean broadcasting between `data` and `grad` of the same variable? That sounds an interesting idea. The `grad` is purely a matter of backward computation where the `data` array is fixed beforehand, so there is only one case: broadcasting a size-one axis of `grad`. It should only occur when the user manually sets the gradient. We can support such broadcasting, though I think the merit is limited compared to the effort to support it (note that PyCUDA does not support broadcasting). Therefore, I think we do not need to support broadcasting now.

beam2d:
I merged the current master.

unnonouno:
LGTM

PR #211, opened 2015-07-17
add spatial pyramid pooling 2d function

mitmul:
This PR adds a non-parameterized function which performs Spatial Pyramid Pooling (SPP). This function enables a network to accept input images of an arbitrary shape/size and generate a fixed-length feature vector regardless of the shape/size of feature maps in the next lower layer. The paper linked below shows that a network with SPP (SPP-net) boosts the accuracy of a variety of CNN for object recognition.

I tested `spatial_pyramid_pooling_2d` function as a pooling layer by replacing the last `max_pooling_2d` layer in LeNet architecture with SPP. I trained the network on MNIST dataset, and got the 99.9% accuracy. So I think it works fine.

Please see details about SPP in the paper:
"Spatial Pyramid Pooling in Deep Convolutional Networks for Visual Recognition"
http://arxiv.org/abs/1406.4729

Note that this function supports only `MaxPooling2D` as elemental pooling operator because `AveragePooling2D` does not support `cover_all` mode for now.

I also fixed some typos in documentations of `max_pooling_2d` and `average_pooling_2d` functions in this PR.

### DIFF
--- a/chainer/functions/pooling_2d.py
+++ b/chainer/functions/pooling_2d.py
@@ -383,8 +383,7 @@ class SpatialPyramidPooling2D(function.Function):
             ksize = (ksize_h, ksize_w)
             pad = (pad_h, pad_w)
 
-            if ((pooling_class == MaxPooling2D) or
-                    (pooling_class == AveragePooling2D)):
+            if pooling_class == MaxPooling2D:
                 pooler = pooling_class(ksize=ksize, stride=None, pad=pad,
                                        cover_all=True, use_cudnn=use_cudnn)
                 self.poolers.append(pooler)
--- a/chainer/functions/pooling_2d.py
+++ b/chainer/functions/pooling_2d.py
@@ -422,23 +422,47 @@ def spatial_pyramid_pooling_2d(x, pyramid_height, pooling_class,
 
     It outputs a fixed-length vector regardless of input feature map size.
 
+    It performs pooling operation to the input 4D-array ``x`` with different
+    kernel sizes and padding sizes, and then flattens all dimensions except
+    first dimension of all pooling results, and finally concatenates them along
+    2nd dimension.
+
+    At :math:`i`-th pyramid level, the kernel size
+    :math:`(k_h^{(i)}, k_w^{(i)})` and padding size
+    :math:`(p_h^{(i)}, p_w^{(i)})` of pooling operation are calculated as
+    below:
+
+    .. math::
+        k_h^{(i)} &= \\lceil b_h / 2^i \\rceil, \\\\
+        k_w^{(i)} &= \\lceil b_w / 2^i \\rceil, \\\\
+        p_h^{(i)} &= (2^i k_h^{(i)} - b_h) / 2, \\\\
+        p_w^{(i)} &= (2^i k_w^{(i)} - b_w) / 2,
+
+    where :math:`\\lceil \\cdot \\rceil` denotes the ceiling function, and
+    :math:`b_h, b_w` are height and width of input variable ``x``,
+    respectively. Note that index of pyramid level :math:`i` is zero-based.
+
     See detail in paper: `Spatial Pyramid Pooling in Deep Convolutional \
     Networks for Visual Recognition \
     <http://arxiv.org/abs/1406.4729>`_.
 
     Args:
-        x (~chainer.Variable): Input variable. The shape of `x` should be
+        x (~chainer.Variable): Input variable. The shape of ``x`` should be
             (batchsize, # of channels, height, width).
         pyramid_height (int): the number of pyramid levels
-        pooling_class (~functions.MaxPooling2D or ~functions.AveragePooling2D):
+        pooling_class (MaxPooling2D or AveragePooling2D):
             Only MaxPooling2D class can be available for now.
         use_cudnn (bool): If True and CuDNN is enabled, then this function
             uses CuDNN as the core implementation.
 
     Returns:
-        ~chainer.Variable: Ouptut variable.
+        ~chainer.Variable: Ouptut variable. The shape of the output variable
+            will be (batchsize, :math:`c \\sum_{h=0}^{H-1} 2^{2h}`, 1, 1),
+            where :math:`c` is the number of channels of input variable ``x``
+            and :math:`H` is the number of pyramid levels.
 
-    .. note:
+    .. note::
+
         This function uses some pooling classes as components to perform
         spatial pyramid pooling. Now it supports only
         :class:`~functions.MaxPooling2D` as elemental pooling operator so far.
--- a/chainer/functions/pooling_2d.py
+++ b/chainer/functions/pooling_2d.py
@@ -363,8 +363,6 @@ class SpatialPyramidPooling2D(function.Function):
 
     def __init__(self, x_shape, pyramid_height, pooling_class, use_cudnn=True):
         bottom_c, bottom_h, bottom_w = x_shape
-        assert bottom_h > 3 and bottom_w > 3, \
-            'input feature map size should be larger than 3'
         self.pyramid_height = pyramid_height
 
         # create pooling functions for different pyramid levels
--- a/chainer/functions/pooling_2d.py
+++ b/chainer/functions/pooling_2d.py
@@ -369,7 +369,7 @@ class SpatialPyramidPooling2D(function.Function):
 
         # create pooling functions for different pyramid levels
         split_inds = []
-        self.out_dim = 0
+        out_dim = 0
         self.poolers = []
         for pyramid_level in six.moves.range(pyramid_height):
             num_bins = int(2 ** pyramid_level)
@@ -393,9 +393,9 @@ class SpatialPyramidPooling2D(function.Function):
             else:
                 raise NotImplementedError()
 
-            self.out_dim += bottom_c * (num_bins ** 2)
+            out_dim += bottom_c * (num_bins ** 2)
             if pyramid_level < pyramid_height - 1:
-                split_inds.append(self.out_dim)
+                split_inds.append(out_dim)
 
         self.split = split_axis.SplitAxis(split_inds, axis=1)
 
--- a/chainer/functions/pooling_2d.py
+++ b/chainer/functions/pooling_2d.py
@@ -385,13 +385,12 @@ class SpatialPyramidPooling2D(function.Function):
             ksize = (int(ksize_h), int(ksize_w))
             pad = (int(pad_h), int(pad_w))
 
-            if pooling_class == MaxPooling2D:
+            if ((pooling_class == MaxPooling2D) or
+                    (pooling_class == AveragePooling2D)):
                 pooler = pooling_class(ksize=ksize, stride=None, pad=pad,
                                        cover_all=True, use_cudnn=use_cudnn)
                 self.poolers.append(pooler)
-
-            if pooling_class == AveragePooling2D:
-                # cover_all mode is needed also for average pooling
+            else:
                 raise NotImplementedError()
 
             self.out_dim += bottom_c * (num_bins ** 2)
--- a/chainer/functions/pooling_2d.py
+++ b/chainer/functions/pooling_2d.py
@@ -1,10 +1,13 @@
 import collections
 
 import numpy
+import six
 
 from chainer import cuda
 from chainer import cudnn
 from chainer import function
+from chainer.functions import concat
+from chainer.functions import split_axis
 from chainer.utils import conv
 
 if cudnn.available:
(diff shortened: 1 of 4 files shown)


delta2323:
Great work! Would you add this function to `docs/source/reference/functions.rst` to add documents of this function the manual?

chainer/functions/pooling_2d.py
```diff
@@ -352,3 +355,105 @@ def average_pooling_2d(x, ksize, stride=None, pad=0, use_cudnn=True):
 
     """
     return AveragePooling2D(ksize, stride, pad, False, use_cudnn)(x)
+
+
+class SpatialPyramidPooling2D(function.Function):
+
+    """Spatial pyramid pooling over a set of 2d planes."""
+
+    def __init__(self, x, pyramid_height, pooling_class, use_cudnn=True):
```
delta2323: `__init__`  uses only the shape of `x` and do not use its data. So second argument should be `shape` instead of `x` .

chainer/functions/pooling_2d.py
```diff
@@ -352,3 +355,105 @@ def average_pooling_2d(x, ksize, stride=None, pad=0, use_cudnn=True):
...
+        use_cudnn (bool): If True and CuDNN is enabled, then this function
+            uses CuDNN as the core implementation.
+
+    Returns:
+        ~chainer.Variable: Ouptut variable.
+
+    """
+
+    return SpatialPyramidPooling2D(x, pyramid_height, pooling_class,
+                                   use_cudnn=True)(x)
```
delta2323: Same as the above comment in `__init__` . Use `x.shape[1:]` instead of `x` as the first argument.

chainer/functions/pooling_2d.py
```diff
@@ -352,3 +355,105 @@ def average_pooling_2d(x, ksize, stride=None, pad=0, use_cudnn=True):
...
+    def forward(self, x):
+        self.ys = []
+        for pooler in self.poolers:
+            y = pooler.forward(x)[0]
+            n, c, h, w = pooler.out_shape = y.shape
+            self.ys.append(y.reshape((n, c * h * w, 1, 1)))
+
+        return self.concat.forward(self.ys)
+
+    def backward_cpu(self, x, gy):
```
delta2323: `backward_cpu` and `backward_gpu` is almost the same. So it would be better to unify the two. There are (at least) two choice of implementation. One is to override `Function.backward` directly and the other is to make common routine a sub-function (like `_backward`). If we choose the first alternative, `backward_cpu` and `backward_gpu` will raise an `NotImplementedError`. But I think it is not problematic because some of existing functions do override `backward`.

chainer/functions/pooling_2d.py
```diff
@@ -352,3 +355,105 @@ def average_pooling_2d(x, ksize, stride=None, pad=0, use_cudnn=True):
...
+        return gx,
+
+
+def spatial_pyramid_pooling_2d(x, pyramid_height, pooling_class,
+                               use_cudnn=True):
+    """Spatial pyramid pooling function.
+
+    This function uses some pooling classes as components to perform spatial
+    pyramid pooling. Now it supports only :class:`~functions.MaxPooling2D` as
+    elemental pooling operator so far. It outputs a fixed-length vector
```
delta2323: Second sentence is a note of current implementation. So please use `.. note `.

chainer/functions/pooling_2d.py
```diff
@@ -352,3 +355,105 @@ def average_pooling_2d(x, ksize, stride=None, pad=0, use_cudnn=True):
...
+    regardless of input feature map size.
+
+    See detail in paper: `Spatial Pyramid Pooling in Deep Convolutional \
+    Networks for Visual Recognition \
+    <http://arxiv.org/abs/1406.4729>`_.
+
+    Args:
+        x (~chainer.Variable): Input variable.
```
delta2323: Please add more detailed specification of input variable `x` . As long as I read the code, `x` must be a tensor of fourth-order whose shape is `(batchsize, # of channels, height, width)`

chainer/functions/pooling_2d.py
```diff
@@ -352,3 +355,105 @@ def average_pooling_2d(x, ksize, stride=None, pad=0, use_cudnn=True):
...
+
+            if pooling_class == AveragePooling2D:
+                # cover_all mode is needed also for average pooling
+                raise NotImplementedError()
+
+            self.out_dim += bottom_c * (num_bins ** 2)
+            if pyramid_level < pyramid_height - 1:
+                split_inds.append(self.out_dim)
+
+        self.concat = concat.Concat(axis=1)
```
delta2323: `self.concat` is used only once in this class. So I think we do not need to hold this function as an attribute.

chainer/functions/pooling_2d.py
```diff
@@ -352,3 +355,105 @@ def average_pooling_2d(x, ksize, stride=None, pad=0, use_cudnn=True):
...
+        self.split = split_axis.SplitAxis(split_inds, axis=1)
+
+    def forward(self, x):
+        self.ys = []
+        for pooler in self.poolers:
+            y = pooler.forward(x)[0]
+            n, c, h, w = pooler.out_shape = y.shape
+            self.ys.append(y.reshape((n, c * h * w, 1, 1)))
+
+        return self.concat.forward(self.ys)
```
delta2323: Same as the above comment on `self.concat`.

chainer/functions/pooling_2d.py
```diff
@@ -352,3 +355,105 @@ def average_pooling_2d(x, ksize, stride=None, pad=0, use_cudnn=True):
...
+        # create pooling functions for different pyramid levels
+        split_inds = []
+        self.out_dim = 0
+        self.poolers = []
+        for pyramid_level in six.moves.range(pyramid_height):
+            num_bins = int(2 ** pyramid_level)
+
+            ksize_h = numpy.ceil(bottom_h / (float(num_bins)))
+            remainder_h = ksize_h * num_bins - bottom_h
+            pad_h = remainder_h / 2
```
delta2323: Use `a//b` instead of `int(a/b)`.

chainer/functions/pooling_2d.py
```diff
@@ -352,3 +355,105 @@ def average_pooling_2d(x, ksize, stride=None, pad=0, use_cudnn=True):
...
+
+            ksize_w = numpy.ceil(bottom_w / (float(num_bins)))
+            remainder_w = ksize_w * num_bins - bottom_w
+            pad_w = remainder_w / 2
```
delta2323: Same as the above (use `a//b`)

tests/functions_tests/test_pooling_2d.py
```diff
@@ -179,4 +179,82 @@ def test_backward_gpu_no_cudnn(self):
         self.check_backward(cuda.to_gpu(self.x), cuda.to_gpu(self.gy), False)
 
 
+class TestSpatialPyramidPooling2D(unittest.TestCase):
+    pyramid_height = 3
+    output_dim = 63  # pyramid_height(=3) * (1 + 4 + 16) = 63
```
delta2323: I think multiplier 3 is derived from the number of channel.

chainer/functions/pooling_2d.py
```diff
@@ -352,3 +355,105 @@ def average_pooling_2d(x, ksize, stride=None, pad=0, use_cudnn=True):
...
+            pad = (int(pad_h), int(pad_w))
+
+            if pooling_class == MaxPooling2D:
+                pooler = pooling_class(ksize=ksize, stride=None, pad=pad,
+                                       cover_all=True, use_cudnn=use_cudnn)
+                self.poolers.append(pooler)
+
+            if pooling_class == AveragePooling2D:
+                # cover_all mode is needed also for average pooling
+                raise NotImplementedError()
```
delta2323: I think it would be better to raise `NotImplementedError` if pooling_class is neither `MaxPooling2D` nor `AveragePooling2D`.

mitmul:
Thank you for a lot of helpful comments. I fixed all the points you commented on, so please check them again.

chainer/functions/pooling_2d.py
```diff
@@ -352,3 +355,100 @@ def average_pooling_2d(x, ksize, stride=None, pad=0, use_cudnn=True):
...
+
+
+class SpatialPyramidPooling2D(function.Function):
+
+    """Spatial pyramid pooling over a set of 2d planes."""
+
+    def __init__(self, x_shape, pyramid_height, pooling_class, use_cudnn=True):
+        bottom_c, bottom_h, bottom_w = x_shape
+        assert bottom_h > 3 and bottom_w > 3, \
+            'input feature map size should be larger than 3'
```
delta2323: Where these constraints come from?

chainer/functions/pooling_2d.py
```diff
@@ -352,3 +355,100 @@ def average_pooling_2d(x, ksize, stride=None, pad=0, use_cudnn=True):
...
+            ksize_h = numpy.ceil(bottom_h / (float(num_bins)))
+            remainder_h = ksize_h * num_bins - bottom_h
+            pad_h = remainder_h // 2
+
+            ksize_w = numpy.ceil(bottom_w / (float(num_bins)))
+            remainder_w = ksize_w * num_bins - bottom_w
+            pad_w = remainder_w // 2
+
+            ksize = (int(ksize_h), int(ksize_w))
+            pad = (int(pad_h), int(pad_w))
```
delta2323: `pad_h` and `pad_w` is already int, so we do not need this coercion.

chainer/functions/pooling_2d.py
```diff
@@ -352,3 +355,100 @@ def average_pooling_2d(x, ksize, stride=None, pad=0, use_cudnn=True):
...
+
```
delta2323: I think the following is better because we can reduce float calculation.
```
ksize_h = int(numpy.ceil(bottom_h / (float(num_bins)))) # add int()
remainder_h = ksize_h * num_bins - bottom_h
pad_h = remainder_h // 2

# (same for ***_w)

ksize = (ksize_h, ksize_w) # remove int()
pad = (pad_h, pad_w) # remove int()
```

chainer/functions/pooling_2d.py
```diff
@@ -352,3 +355,100 @@ def average_pooling_2d(x, ksize, stride=None, pad=0, use_cudnn=True):
...
+            if ((pooling_class == MaxPooling2D) or
+                    (pooling_class == AveragePooling2D)):
```
delta2323: We also raise `NotImplementedError` when `pooling_class` is `AveragePooling2D` (Sorry, my last comment about this was ambiguous).

chainer/functions/pooling_2d.py
mitmul: I forgot to delete them. It was just for debugging.

chainer/functions/pooling_2d.py
```diff
@@ -352,3 +355,100 @@ def average_pooling_2d(x, ksize, stride=None, pad=0, use_cudnn=True):
...
+                                       cover_all=True, use_cudnn=use_cudnn)
+                self.poolers.append(pooler)
+            else:
+                raise NotImplementedError()
+
+            out_dim += bottom_c * (num_bins ** 2)
+            if pyramid_level < pyramid_height - 1:
+                split_inds.append(out_dim)
+
+        self.split = split_axis.SplitAxis(split_inds, axis=1)
```
delta2323: I think we need not hold `split_axis` function as an attribute for the same reason that `self.concat` is removed.

chainer/functions/pooling_2d.py
```diff
@@ -352,3 +355,100 @@ def average_pooling_2d(x, ksize, stride=None, pad=0, use_cudnn=True):
...
+            self.ys.append(y.reshape((n, c * h * w, 1, 1)))
+
+        return concat.Concat(axis=1).forward(self.ys)
+
+    def backward(self, x, gy):
+        if isinstance(x[0], cuda.GPUArray):
+            gx = cuda.zeros_like(x[0])
+        else:
+            gx = numpy.zeros_like(x[0])
+        gys = self.split.forward(gy)
```
delta2323: Same as the above (we do not have to hold `self.split`)

delta2323:
@mitmul I added comments which reflects your update. Please see to check them.

chainer/functions/pooling_2d.py
mitmul: OK, so can I make `split_inds` an attribute instead of `self.split`?

mitmul:
@delta2323 Thanks. I refined the codes again.

chainer/functions/pooling_2d.py
delta2323: I think it will be OK.

delta2323:
Great! LGTM. We can merge this function in the next minor release.

chainer/functions/pooling_2d.py
```diff
@@ -352,3 +355,95 @@ def average_pooling_2d(x, ksize, stride=None, pad=0, use_cudnn=True):
...
+    """Spatial pyramid pooling function.
+
+    It outputs a fixed-length vector regardless of input feature map size.
+
+    See detail in paper: `Spatial Pyramid Pooling in Deep Convolutional \
+    Networks for Visual Recognition \
+    <http://arxiv.org/abs/1406.4729>`_.
+
+    Args:
+        x (~chainer.Variable): Input variable. The shape of `x` should be
```
delta2323: Use ``(two back quotes) instead of `(single quote), since single quote make character italic

chainer/functions/pooling_2d.py
```diff
@@ -352,3 +355,95 @@ def average_pooling_2d(x, ksize, stride=None, pad=0, use_cudnn=True):
...
+            (batchsize, # of channels, height, width).
+        pyramid_height (int): the number of pyramid levels
+        pooling_class (~functions.MaxPooling2D or ~functions.AveragePooling2D):
```
delta2323: Remove "~functions." from argument type list. When compiling this line, MaxPooling2D is not shown in the argument type list of `pooling_class`.  "(MaxPooling2D or AveragePooling2D)" worked.

chainer/functions/pooling_2d.py
```diff
@@ -352,3 +355,95 @@ def average_pooling_2d(x, ksize, stride=None, pad=0, use_cudnn=True):
...
+            Only MaxPooling2D class can be available for now.
+        use_cudnn (bool): If True and CuDNN is enabled, then this function
+            uses CuDNN as the core implementation.
+
+    Returns:
+        ~chainer.Variable: Ouptut variable.
+
+    .. note:
```
delta2323: Add one more colon (".. note::")to enable note.

delta2323: Add one more empty line after "..note::" to enable note.

chainer/functions/pooling_2d.py
```diff
@@ -352,3 +355,95 @@ def average_pooling_2d(x, ksize, stride=None, pad=0, use_cudnn=True):
...
+        x (~chainer.Variable): Input variable. The shape of `x` should be
+            (batchsize, # of channels, height, width).
+        pyramid_height (int): the number of pyramid levels
+        pooling_class (~functions.MaxPooling2D or ~functions.AveragePooling2D):
+            Only MaxPooling2D class can be available for now.
+        use_cudnn (bool): If True and CuDNN is enabled, then this function
+            uses CuDNN as the core implementation.
+
+    Returns:
+        ~chainer.Variable: Ouptut variable.
```
delta2323: Could you write the shape of output variable? Especially the information about `output_variable.shape[1]` would be helpful.

delta2323:
I added some comments on documents. I would appreciate you if you fix them. Sorry for additional order.

delta2323:
Please let me know if you do not have time to fix the code. We do it instead and make a PR to your branch.

mitmul:
@delta2323 I added some additional information to the documentation part. Please check it again. Thank you.

delta2323:
Thank you very much! I will check in detail later. But basically it seems OK.

delta2323:
LGTM :+1: Thank you for answering my reviews.

PR #219, opened 2015-07-21
Add typecheck for unary operator

delta2323:
typecheck for following functions:
- Neg
- Absolute
- AddConstant
- SubFromConstatnt
- MulConstant
- DivFromConstant
- PowVarConst
- PowConstVar
- Exp
- Log
- Tanh
- Sigmoid

Related #123

### DIFF
--- a/chainer/function.py
+++ b/chainer/function.py
@@ -490,7 +490,7 @@ class Split(Function):
         return gx,
 
 
-class UnaryOperator(function.Function):
+class UnaryOperator(Function):
 
     def check_type_forward(self, in_types):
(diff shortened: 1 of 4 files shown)


beam2d:
Thank you for the work. `UnaryOperator` may be an interface addition, which should be deferred to v1.2.0. Though, it seems the type check for `Sigmoid` and `Tanh` should include checking of dtypes, so you can put `UnaryOperator` into `basic_math.py` as a module-local class. I also think it is also fine to expand these two lines into each Functions.

beam2d:
[LGTM](http://lgtm.in/i/Acdhfwhgq)

PR #226, opened 2015-07-21
Add typecheck to matmul

delta2323:
Related to #123

### DIFF
--- a/chainer/functions/matmul.py
+++ b/chainer/functions/matmul.py
@@ -100,12 +100,10 @@ def _convert_type(in_type, vector_ndim=1):
 
 
 def _get_check_index(trans, right, row_idx=0, col_idx=1):
-    use_col_idx = True
-    if trans:
-        use_col_idx = not use_col_idx
-    if right:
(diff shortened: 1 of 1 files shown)


chainer/functions/matmul.py
```diff
@@ -80,17 +81,62 @@ def _batch_matmul_gpu(a, b, out, transa=False, transb=False, transout=False):
...
+    return in_type
+
+
+def _get_check_index(trans, right, row_idx=0, col_idx=1):
+    use_col_idx = True
+    if trans:
+        use_col_idx = not use_col_idx
+    if right:
+        use_col_idx = not use_col_idx
+    return col_idx if use_col_idx else row_idx
```
beam2d: How about:
```
def _get_check_index(trans, right, row_idx=0, col_idx=1):
    if trans ^ right:
        return row_idx
    else:
        return col_idx
```

beam2d:
LGTM except one comment.

beam2d:
[LGTM](http://lgtm.in/i/yDB1)

Issue #228, opened 2015-07-21
Operators in basic_math supports broadcast


delta2323:

delta2323:
To make broader ranges of functions applicable to broadcast. We plan to implement broadcast as `function`.

entron:
Agree that when basic_math supports broadcast it could sometimes cause unexpected broadcast when a user makes a mistake. However, isn't making the API compatible to numpy thus making chainer easy to learn/use/port more meaningful?

PR #229, opened 2015-07-21
Add check_type_forward to binary operators in basic_math

delta2323:
Related to #123 

It is worth discussion whether we admit broadcast in binary operators in `basic_math` (i.e. shapes of first and second argument do not match, but their data can be computed by means of broadcast of numpy
). For now, we decided not to admit broadcast because in order to admit broadcast, we must fix backward of each operators to maintain consistency of forward and backward.

Binary operators' support of broadcast is issued as #228 .

### DIFF
--- a/chainer/functions/basic_math.py
+++ b/chainer/functions/basic_math.py
@@ -88,6 +88,13 @@ class Add(function.Function):
     def label(self):
         return '_ + _'
 
+    def check_type_forward(self, in_types):
+        type_check.expect(in_types.size() == 2)
+        type_check.expect(
+            in_types[0].dtype == in_types[1].dtype,
+            in_types[0].shape == in_types[1].shape
+        )
+
     def forward(self, x):
         y = utils.force_array(x[0] + x[1])
         return y,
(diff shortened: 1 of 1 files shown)


chainer/functions/basic_math.py
```diff
@@ -301,6 +329,13 @@ class PowVarVar(function.Function):
     def label(self):
         return '_ ** _'
 
+    def check_type_forward(self, in_types):
+        type_check.expect(in_types.size() == 2)
+        type_check.expect(
+            in_types[0].dtype == in_types[1].dtype,
```
unnonouno: GPU code assumes `numpy.float32`

chainer/functions/basic_math.py
```diff
@@ -171,6 +185,13 @@ class Mul(function.Function):
     def label(self):
         return '_ * _'
 
+    def check_type_forward(self, in_types):
+        type_check.expect(in_types.size() == 2)
+        type_check.expect(
+            in_types[0].dtype == in_types[1].dtype,
+            in_types[0].shape == in_types[1].shape
```
unnonouno: `numpy.float32` is expected in `backward_gpu`

chainer/functions/basic_math.py
```diff
@@ -222,6 +243,13 @@ class Div(function.Function):
     def label(self):
         return '_ / _'
 
+    def check_type_forward(self, in_types):
+        type_check.expect(in_types.size() == 2)
+        type_check.expect(
+            in_types[0].dtype == in_types[1].dtype,
+            in_types[0].shape == in_types[1].shape
```
unnonouno: `numpy.float32` is expected in `backward_gpu`

unnonouno:
LGTM

PR #230, opened 2015-07-21
Restrict dtype of arguments in some of unary operators in basic_math

delta2323:
The objective of this fix is basically same as d7836e .

Some unary operators in `basic_math` is implemented by explicitly writing kernel code. In that case, dtypes of arguments are restricted to float32. This PR imposes this restriction to `check_type_forward` of the functions.

Related to #219

### DIFF
--- a/tests/functions_tests/test_basic_math.py
+++ b/tests/functions_tests/test_basic_math.py
@@ -249,35 +249,98 @@ class TestBinaryOpConstant(unittest.TestCase):
         y4.backward()
         self.assertEqual(x4.grad.dtype, numpy.float32)
 
-    def test_add_constnt(self):
+    def _test_constant_array_one(self, func, lhs, rhs):
+        x = chainer.Variable(lhs)
+        y = func(x, rhs)
+        self.assertEqual(y.data.dtype, numpy.float32)
+        y.grad = numpy.ones_like(y.data, numpy.float32)
+        y.backward()
+        self.assertEqual(x.grad.dtype, numpy.float32)
+
+    def _test_constant_array(self, func):
+        x_data = numpy.array([1.0, 2.0], numpy.float32)
+
+        self._test_constant_array_one(
+            func, x_data, numpy.array([3.0, 4.0], numpy.int32))
+        self._test_constant_array_one(
+            func, x_data, numpy.array([3.0, 4.0], numpy.int64))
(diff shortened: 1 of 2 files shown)


beam2d:
Looks good. How about the dtypes of following functions?
- AddConstant
- SubFromConstant
- MulConstant
- Exp
- Log

delta2323:
The followings are my understandings.
As for `Exp` and `Log` it is enough that their arguments are `float`. We do not restrict its size.

For `***Constat`(take `Addconstant` which maps `x` to `x+a` for example)
- `a` is scalar :  It is no problem if the constant `a` is a scalar because of `_force_type`.
- `a` is array
    - `x` is int : We do not admit this case (as we do in other functions)
    - `x` is float64 : Promotion causes no problems
    - `x` is float32
        - `a` is int : Type check should be passed. We should coarse `a` to float32.
        - `a` is float32 : No problem
        - `a` is float64 : Type check should forbids the case.

beam2d:
Thank you for the investigation. How about forbidding an int array `a`? I feel inconsistency on prohibiting a float64 array while allowing an int array (by forcing the dtype).

delta2323:
The other alternative to handle the case float64 is to coarse it to float32.

delta2323:
In the end,  `Exp` and `Log` restrict the kind of input arguments to `float` and `***Constant` forbids to use `list` or `tuple` as a constant (use `numpy.ndarray` instead)

unnonouno:
Do we need to test GPUarray?

unnonouno:
How about `sin` and `cos`?

tests/functions_tests/test_basic_math.py
```diff
@@ -223,62 +223,232 @@ def make_data(self):
...
+
+        self._test_constant_one(func, x_data, 1)
+        self._test_constant_one(func, x_data, 1.0)
+        self._test_constant_one(func, x_data, numpy.int64(1))
+        self._test_constant_one(func, x_data, numpy.float64(1.0))
+
+    def _test_constant_gpu(self, func):
+        x_data = numpy.array(1, numpy.float32)
+
+        self._test_constant_one(func, x_data, 1)
```
unnonouno: `gpu=True`?

unnonouno:
LGTM

PR #272, opened 2015-07-28
Add Inception{,BN}.check_type_forward in which we explicitly pass

### LINKED ISSUE #232: Add type check to Inception and InceptionBN functions
Related #123

delta2323:
Fix #232

### DIFF
--- a/chainer/functions/inception.py
+++ b/chainer/functions/inception.py
@@ -63,6 +63,8 @@ class Inception(function.Function):
         y = relu.relu(concat.concat((out1, out3, out5, pool), axis=1))
         return y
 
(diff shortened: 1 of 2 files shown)


chainer/functions/inception.py
```diff
@@ -63,6 +63,9 @@ def __call__(self, x):
         y = relu.relu(concat.concat((out1, out3, out5, pool), axis=1))
         return y
 
+    def check_type_forward(self, in_types):
+        pass
```
unnonouno: How about to write a comment?

delta2323:
I added comments.

unnonouno:
LGTM

PR #288, opened 2015-08-04
Add a framework comparison

### LINKED ISSUE #32: Compare Chainer with existing frameworks
Candidates I suppose are Caffe, Torch7, and Theano-based frameworks (like Pylean2, keras etc.).

beam2d:
It adds a new page to the documentation. It currently consists of a comparison table.

fix #32


w0nk0:
Quick question, since it's mentioned in the comparison - is there 
documentation for pickling of models in chainer? I think I tried pickling 
the state of a model once and it wasn't working 'out of the box'.

beam2d:
There is no documentation about pickling right now. Classes in Chainer only consists of objects that can be pickled (including GPUArray), so the model (e.g. FunctionSet) can be pickled, too. Would you post a failure to [the forum](https://groups.google.com/forum/?hl=ja#!forum/chainer)? We can inspect causes and, if it is due to a bug of Chainer, we will fix it.

docs/source/comparison.rst
```diff
@@ -0,0 +1,69 @@
+Comparison with Other Frameworks
+================================
+
+A table for quick comparison
+----------------------------
+
+This table compares Chainer and other popular deep learning frameworks.
```
delta2323: It seems compare do not have such an usage. "... compare Chainer with other popular deep learning frameworks"
http://www.oxforddictionaries.com/definition/english/compare

docs/source/comparison.rst
```diff
@@ -0,0 +1,69 @@
...
+| Perf  | CUDA support                | Y                 | Y                      | Y                 | Y                                                  |
+|       +-----------------------------+-------------------+------------------------+-------------------+----------------------------------------------------+
+|       | cuDNN support               | Y                 | Y                      | Y (cudnn.torch)   | Y                                                  |
+|       +-----------------------------+-------------------+------------------------+-------------------+----------------------------------------------------+
+|       | FFT-based convolution       |                   | Y                      | Y (fbcunn)        | `#544 <https://github.com/BVLC/caffe/pull/544>`_   |
+|       +-----------------------------+-------------------+------------------------+-------------------+----------------------------------------------------+
+|       | CPU/GPU generic coding      | [1]_              | [4]_                   | Y                 |                                                    |
```
delta2323: I believe what we want to emphasize is whether we can write code that can usable both CPU and GPU mode. It would be better to add some explanations for it.

delta2323:
LGTM :+1:

PR #309, opened 2015-08-13
Inception unittest

### LINKED ISSUE #294: Make a unit test for inception

delta2323:
This PR fixes #294


delta2323:
I am considering how we check `Inception.backward` calls `backward` of each function in `Inception.f` properly.

delta2323:
I come to think that once `forward`s of some set of functions are called properly, we should check if their `backward`s are called as a test of `function` and `variable`. So we need not check the order of  `backward` calls.

unnonouno:
LGTM

PR #311, opened 2015-08-13
Remove check_type_backward and its relevants

### LINKED ISSUE #209: check_type_backward may be removed

delta2323:
As discussed in #209, `check_type_backward` is no longer needed. This PR removes this method and codes that relevant to it.

This PR fixes #209

### DIFF
--- a/chainer/function.py
+++ b/chainer/function.py
@@ -190,11 +190,6 @@ class Function(object):
         in_type = type_check.get_types(in_data, 'in_types', False)
         self.check_type_forward(in_type)
 
-    def _check_data_type_backward(self, in_data, grad_data):
-        in_type = type_check.get_types(in_data, 'in_types', False)
-        grad_type = type_check.get_types(grad_data, 'grad_types', True)
-        self.check_type_backward(in_type, grad_type)
-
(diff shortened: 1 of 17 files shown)


unnonouno:
LGTM

PR #312, opened 2015-08-13
Fix NumericalGradientTest that fails in python3

delta2323:
In python3 map returns iterator, which causes failure in unit tests of numerical_grad. This PR fixes this failure.


unnonouno:
LGTM

PR #340, opened 2015-08-22
Fix wrong device destination in optimizer.accumulate_grads().

umitanuki:
It looks to me simply wrong to copy the src variable to src device.  In fact, I bumped into "invalid memory access" error when I tried to accumulate grads between multiple GPUs.


okuta:
Thanks!
LGTM.

umitanuki:
Thanks, and I just found a doc bug in the data-parallel multi-gpu tutorial section.

https://chainer.readthedocs.org/en/stable/tutorial/gpu.html#data-parallel-computation-on-multiple-gpus
`acumulate_grads` should be `accumulate_grads`.

PR #341, opened 2015-08-23
Fix Convolution2D.backward_gpu so that it does not use cuda.linalg

delta2323:
When CuDNN is not used, `Convolution2D.backward_gpu` uses `cuda.linalg`, which is no longer available after CuPy is introduced. This PR removes dependency on `cuda.linalg.dot` and `cuda.linalg.add_dot` and uses `cuda.cupy.dot` instead.

### DIFF
--- a/chainer/functions/convolution_2d.py
+++ b/chainer/functions/convolution_2d.py
@@ -272,7 +272,6 @@ class Convolution2D(function.Function):
                 gy_desc.value, gy_arr.data.ptr, self.conv_desc.value,
                 zero, x_desc.value, gx.data.ptr)
         else:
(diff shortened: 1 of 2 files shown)


beam2d:
We should fix the test code to pass the im2col/col2im path even in cuDNN-enabled environments.

beam2d:
I fixed tests to check im2col version even if cuDNN is enabled.

delta2323:
As for @beam2d's fix, it looks good for me.

beam2d:
OK, then I'll merge.

beam2d:
Thank you!

PR #337, opened 2015-08-21
add axis option and apply cupy for functions.Sum

kikusu:
Add axis option like numpy.sum for functions.Sum

```
s = list(x[0].shape)
s[self.axis] = 1
gy = gy[0].reshape(s)
```
If `cupy.expand_dims` is implemented, this cord is replaced with `gy = xp.expand_dims(gy[0], axis=self.axis)`

PR #343, opened 2015-08-24
Implement expand_dims

okuta:
It is related to #337.

### DIFF
--- a/cupy/manipulation/dims.py
+++ b/cupy/manipulation/dims.py
@@ -183,8 +183,24 @@ def broadcast_arrays(*args):
 
 
 def expand_dims(a, axis):
-    # TODO(beam2d): Implement it
-    raise NotImplementedError
+    """Expands given arrays.
+
+    Args:
+        a (cupy.ndarray): Array to be expanded.
+        axis (int): Position where new axis is to be inserted.
+
+    Returns:
+        cupy.ndarray: The number of dimensions is one greater than that of
+            the input array.
+
+    .. seealso:: :func:`numpy.expand_dims`
+
+    """
+    a = cupy.asarray(a)
+    shape = a.shape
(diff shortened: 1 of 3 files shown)


cupy/manipulation/dims.py
```diff
@@ -183,8 +183,24 @@ def broadcast_arrays(*args):
...
+        cupy.ndarray: The number of dimensions is one greater than that of
+            the input array.
+
+    .. seealso:: :func:`numpy.expand_dims`
+
+    """
+    a = cupy.asarray(a)
+    shape = a.shape
+    if axis < 0:
+        axis = axis + len(shape) + 1
```
delta2323: What happens when `axis + len(shape) + 1` is still negative?

okuta:
I added negative test case.

cupy/manipulation/dims.py
beam2d: It should delegate the handling of negative axes to tuple indexing instead of handling by hand.

okuta:
> What happens when axis + len(shape) + 1 is still negative?

Slice can specify axis of the out of range.

And, this implementation has numpy compatible.
https://github.com/numpy/numpy/blob/v1.9.1/numpy/lib/shape_base.py#L273

okuta:
> What happens when axis + len(shape) + 1 is still negative?

Even if the axis is out of range , Slice will work.

And, this implementation has numpy compatible.
https://github.com/numpy/numpy/blob/v1.9.1/numpy/lib/shape_base.py#L273

delta2323:
@okuta Thank you for adding negative indices cases.

beam2d:
OK, I'll merge.

beam2d:
Thank you!

PR #346, opened 2015-08-25
Make cupy.cudnn importable only if cuDNN is available

### LINKED ISSUE #339: cupy.cudnn.reset_handles() fails for atexit run time
- chainer (fc79c3caf0a5ebea1bcbb1570b48c057577cf1ad) with GPU
- cuDNN not installed
- python (using pyenv)

```
Python 2.7.10 (default, Jun 19 2015, 20:41:53)
[GCC 4.8.4] on linux2
>>> import chainer
>>> exit()
Error in atexit._run_exitfuncs:
  File "****/2.7.10/lib/python2.7/atexit.py", line 24, in _run_exitfuncs
    func(*targs, **kargs)
  File "cupy/cudnn.py", line 35, in reset_handles
    for handle in six.itervalues(handles):
AttributeError: 'NoneType' object has no attribute 'itervalues'
Error in sys.exitfunc:
  File "***/2.7.10/lib/python2.7/atexit.py", line 24, in _run_exitfuncs
    func(*targs, **kargs)
  File "cupy/cudnn.py", line 35, in reset_handles
    for handle in six.itervalues(handles):
AttributeError: 'NoneType' object has no attribute 'itervalues'
```

I can see the problem as follows
1. cupy/cudnn.py:10 `from cupy.cuda import cudnn` fails because cuDNN is not installed.
2. However, `import cupy.cudnn` fails because `cudnn` is used in cupy/cudnn.py (name 'cudnn' is not defined)
3. The handling of the global variable(`cupy.cudnn._handles`) of this time seems to be different from python3.3~ in python2.

test_code.py
```
import atexit

global_variable = []

@atexit.register
def exit_func():
    global global_variable
    print("global_variable is ", global_variable)


fail_code # this is fail code
```

python 2.7.10
```
Python 2.7.10 (default, Jun 19 2015, 20:41:53)
[GCC 4.8.4] on linux2
>>> import test_code
  File "test_code.py", line 11, in <module>
    fail_code # this is fail code
NameError: name 'fail_code' is not defined
>>> exit()
('global_variable is ', None)
```

python 3.4.3
```
Python 3.4.3 (default, Aug 13 2015, 16:03:08)
[GCC 4.8.4] on linux
>>> import test_code
  File "test_code.py", line 11, in <module>
    fail_code # this is fail code
NameError: name 'fail_code' is not defined
>>> exit()
global_variable is  []
```
The result is `[]` in the python3.3.x~, but becomes `None` in python2.
But python3.0.1, 3.1.5, 3.2.6 result is 'None', python 2.7.0 is [].

beam2d:
Fix #339. It actually just makes `cupy.cudnn` not importable if cuDNN is not available.

### DIFF
--- a/cupy/cudnn.py
+++ b/cupy/cudnn.py
@@ -5,12 +5,7 @@ import numpy
 import six
 
 from cupy import cuda
-
-try:
-    from cupy.cuda import cudnn
-    available = True
-except Exception:
-    available = False
+from cupy.cuda import cudnn
 
 
 _handles = {}


delta2323:
This PR forbids to import `cupy.cudnn` when CuDNN is not available. The motivation of this change is two folds.

1. Since all functions in `cupy.cudnn` module assume that the import of `cupy.cuda.cudnn` is successful, there seems no reason to make the import of `cupy.cudnn` successful when CuDNN is not available.

2. In the current implementation, default arguments of `create_tensor_descriptor`, `create_filter_descriptor` and `create_convolution_descriptor` access to `cudnn` module and cause an error when the import of `cupy.cuda.cudnn` fails (So there are no cases where `cupy.cudnn.available` is `False`).
This error seems to cause the side effect on the post process of the program. Even if this error is raised, registration of `reset_handles` by `atexit` is successful, it is called at the end of program. But as the import of `cupy.cudnn` fails due to the error, `six` module is not imported and raises error at `six.itervalues(handles)`. I think this is the cause of #339. The simplest solution is to forbid the import of this module when `cudnn` is not available.

okuta:
LGTM!

PR #350, opened 2015-08-26
Update cuda dependency paragraph in README

nushio3:
The file mentioned at README
```
python cuda_deps/setup.py install
```
seems to have been vanished.
Would you please update the paragraph and let us, the users, know how to  install the latest cuda dependencies?


beam2d:
Thank you for the catch! I forgot to update the readme. Indeed, other points in the readme are also obsolete, so I will make further updates on it in another PR.

PR #351, opened 2015-08-26
Reorganize chainer.functions

### LINKED ISSUE #331: Reorganize the modules for Function implementations
Thanks to many contributors, we now have many Function implementations under the `chainer.functions` module. It is convenient that user can access to them via `chainer.functions`, though the source code is less readable and searchable since they have no grouping based on the type of their functionalities. I want to reorganize the `chainer.functions` module. Note that I still want to keep the current interface that all functions can be accessed via `chainer.functions`.

beam2d:
fix #331. It contains reorganization of the chainer.functions sub package. The diff looks very large, though it contains few essential changes other than the relocation of classes and functions under the module hierarchy. Almost all functions and classes of chainer.functions can be accessed basically as before. The changes in `chainer.functions` namespace are following:

- Modules are grouped into ten sub packages: `activation`, `array`, `caffe`, `connection`, `evaluation`, `loss`, `math`, `noise`, `normalization`, `pooling`.
- create_huffman_tree is moved to a static method of BinaryHierarchicalSoftmax.
- VAE loss components (bernoulli_nll, gaussian_kl_divergence, and gaussian_nll) are now accessible via `chainer.functions`.


beam2d:
I also moved NonparameterizedLinear and NonparameterizedConvolution2D to linear.py and convolution_2d.py, respectively.

unnonouno:
LGTM

PR #352, opened 2015-08-26
Update readme for v1.3

beam2d:
The main change is the CUDA support description. It also adds the supported NumPy version based on the internal testing.


unnonouno:
LGTM

PR #354, opened 2015-08-26
Update train_sentiment.py, bug fix

miljanm:
Fixing the following bug:

File "/Library/Python/2.7/site-packages/chainer/optimizer.py", line 42, in setup
    for p, g in zip(*params_grads):
TypeError: zip() argument after * must be a sequence, not FunctionSet

### DIFF
--- a/examples/sentiment/train_sentiment.py
+++ b/examples/sentiment/train_sentiment.py
@@ -152,7 +152,7 @@ if args.gpu >= 0:
 
 # Setup optimizer
 optimizer = optimizers.AdaGrad(lr=0.1)
-optimizer.setup(model)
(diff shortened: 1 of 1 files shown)


delta2323:
Which version do you use? If the version is prior to v1.2.0, I would recommend you to update Chainer because in v1.2.0 `optimizer.setup` can accept `FunctionSet` itself and `FunctionSet.collect_parameters()` gets obsolete, although it still works. (See #177)
You can check chainer's version by
```
>>> import chainer
>>> chainer.__version__
```

miljanm:
You're right, accidentally loaded the old version, my bad.

PR #355, opened 2015-08-28
Fix range bug of ptb example

okuta:
I fixed range bug of ptb example.


okuta:
It was a misunderstanding.

PR #357, opened 2015-08-30
fixed bug that --out argument in train_imagenet.py dimmed

mrfuzzybee:
has tested locally that model file had correct  name  was generated


okuta:
LGTM!

okuta:
Thank you for your PR.

PR #358, opened 2015-08-30
Add partial Windows support

okuta:
This PR add partial windows support.
- Use filelock package instead of fcntl.
- Do not use zero sized array.

I tested that the example of mnist can be run in the following environments.
- WIndows7
- CUDA 7.0
- Python 2.7.10 |Anaconda 2.3.0 (64-bit)|
However, some test cases is not completed successfully.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@ Minimum requirements:
 Requirements for some features:
 - CUDA support
   - CUDA 6.5+
+  - filelock
 - CuDNN support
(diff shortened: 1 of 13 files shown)


beam2d:
I did not check Windows yet (since I could not prepare an appropriate environment), though it runs correctly on my Ubuntu environment. I will merge it. Further trials by Windows users are appriciated!

PR #359, opened 2015-08-31
Speed up Python code of CuPy

### LINKED ISSUE #333: Make Python code of CuPy faster
While CuPy enables us to great flexibility and usability of using CUDA, it also adds great overheads due to the slow index manipulations and code generations written in pure Python. We must make it faster to improve the try-and-error cycle on small-scale networks.

There might be three options:

1. Make the pure Python code faster by finding bottlenecks and improve them.
2. Port some critical codes to Cython.
3. Reimplement the ndarray by C and add a Python C extension for it.

I have already tried the approach 1 a bit, and realized it is very hard and seems to be hopeless. Approach 3 is maybe ideal, since the resulting array implementation might be used by other languages as well, though this task is too heavy to complete shortly. Approach 2 is an intermediate solution and seems to be better as the first step of improving CuPy performance.

If we take the approach 2, we have to think about the installation process. Using Cython means we introduce C extensions for the CUDA-related code, which cannot be compiled (linked) in environments without CUDA. In order to support such environments, we have to do either of the following two:

A. Detect whether the CUDA is installed or not in setup.py, and if not, avoid the installation of CuPy.
B. Make CuPy an isolated package (like PyCUDA), and let users decide whether to install CuPy or not.

I recommend to choose the approach B, since it makes it clear which modules are installed from what the user did.

okuta:
I fix #333.

### DIFF
--- a/chainer/functions/convolution_2d.py
+++ b/chainer/functions/convolution_2d.py
@@ -167,7 +167,7 @@ class Convolution2D(function.Function):
     def forward_cpu(self, x):
         self.col = conv.im2col_cpu(
             x[0], self.kh, self.kw, self.sy, self.sx, self.ph, self.pw)
-        y = numpy.tensordot(self.col, self.W, ([1, 2, 3], [1, 2, 3]))
+        y = numpy.tensordot(self.col, self.W, ((1, 2, 3), (1, 2, 3)))
         if self.b is not None:
             y += self.b
         return numpy.rollaxis(y, 3, 1),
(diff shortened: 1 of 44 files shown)


beam2d:
[LGTM](http://lgtm.in/i/kcLOwU1lq)

bordingj:
This gave me a nice speed-up! Thanks!
Is it still in the plans to rewrite some of cupy to cython ?

PR #360, opened 2015-08-31
Fix _convert_value_to_string to support constant other than float

### LINKED ISSUE #353: XConstant functions whose value is integer cannot access to label attributes
Functions in basic_math which involve operations on constant values (namely, `AddConstant`, `SubFromConstant`, `MulConstant`, `DivFromConstant`, `PowConstVar`, and `PowVarConst`) admit that the constant value is integer. But when we do access their label properties, `ValueError` is raised because `_convert_value_to_string`, which is called when creating labels, forbids the value being integer. We should allow broader constant cases.

In the first place, it could be inappropriate to check the type of the constant value in `constant_value_to_string`. I think we should check the type in instantiation.

delta2323:
This PR fixes #353


chainer/functions/basic_math.py
```diff
@@ -15,13 +15,16 @@ def _convert_value_to_string(value):
     if isinstance(value, variable.Variable):
         value = value.data
 
-    if isinstance(value, float):
-        return str(value)
+    if numpy.isscalar(value):
+        if value < 0:
+            return '({})'.format(str(value))
```
beam2d: Remove str.

beam2d:
I added a comment. Would you test on one of numpy scalars? I think testing float32 is enough.

beam2d:
[LGTM](http://lgtm.in/i/MSPDzNKhq)

PR #362, opened 2015-08-31
Fix error message of _check_constant_type

delta2323:
I found a minor mistake in error message of `_check_constant_type` function in `basic_math`


beam2d:
Thanks!

PR #364, opened 2015-08-31
Comment out functions that is not implemented

okuta:
I was commenting out functions that is not used.


beam2d:
Would you delete NOQAs and related comments?

beam2d:
Sorry I now realize that NOQA is required to allow empty module (only with comments).

beam2d:
[LGTM](http://lgtm.in/i/QeE1)

beam2d:
Thank you!!!

PR #369, opened 2015-09-01
Fix nonparameterized linear without bias

### LINKED ISSUE #367: Non-parameterized linear without bias
Hello all. 

I'm sorry if this has been already reported.

According to the document, chainer looks to allow non-parameterized linear function call without bias (`b=None`), but when I try it, python raises an attribute error, which says "NoneType object has no attribute volatile."

This is because `chainer.functions.linear()` has `b=None` and `chainer.function.__call__()` tries to check `b.volatile` at line 132. 

A minor change in `chainer.functions.linear()` like:

    if b is None:
        return NonparameterizedLinear()(x, W)
    else:
        return NonparameterizedLinear()(x, W, b)

instead of just doing:

    return NonparameterizedLinear()(x, W, b)

may fix the problem.

beam2d:
fix #367. It also fixes nonparameterized convolution_2d without bias.

### DIFF
--- a/chainer/functions/connection/linear.py
+++ b/chainer/functions/connection/linear.py
@@ -52,7 +52,6 @@ class Linear(function.Function):
        and the other dimensions are reduced to one dimension.
 
     """
-
     def __init__(self, in_size, out_size, wscale=1, bias=0, nobias=False,
                  initialW=None, initial_bias=None):
         self.W = None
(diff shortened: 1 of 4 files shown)


chainer/functions/connection/linear.py
```diff
@@ -180,13 +180,13 @@ def backward(self, x, gy):
...
 
-def linear(x, W, b=None, stride=1, pad=0, use_cudnn=True):
+def linear(x, W, b=None):
     """Nonparameterized linear function.
 
     Args:
         x (~chainer.Variable): Input variable.
         W (~chainer.Variable): Weight variable.
-        b (~chainer.Variable): Bias variable.
+        b (~chainer.Variable): Bias variable (optional).
```
unnonouno: Do we need to fix docstring of convolution2d like this?: https://github.com/pfnet/chainer/pull/369/files#diff-34a15eecaef34d920b685d1f971f6a17L371

unnonouno:
LGTM except one comment.

chainer/functions/connection/linear.py
beam2d: Oh, I've forgotten to fix that.

unnonouno:
LGTM

PR #373, opened 2015-09-01
Remove redundant pooling modules

delta2323:
`pooling_2d.py` is duplicated after `functions` directory is reorganized. This PR removes redundant `pooling_2d.py`.


unnonouno:
LGTM

Issue #381, opened 2015-09-03
An error when I customized __init__ of subclass of FunctionSet


choni:
Hello.

I customized \__init\__ of subclass of FunctionSet like below
```
class NIN(chainer.FunctionSet):

    insize = 227
    numClasses = 0

    def __init__(self, numClasses):
        self.numClasses = numClasses
        w = math.sqrt(2)  # MSRA scaling
        super(NIN, self).__init__(
            conv1=F.Convolution2D(3,   96, 11, wscale=w, stride=4),
            conv1a=F.Convolution2D(96,   96,  1, wscale=w),
            conv1b=F.Convolution2D(96,   96,  1, wscale=w),
            conv2=F.Convolution2D(96,  256,  5, wscale=w, pad=2),
            conv2a=F.Convolution2D(256,  256,  1, wscale=w),
            conv2b=F.Convolution2D(256,  256,  1, wscale=w),
            conv3=F.Convolution2D(256,  384,  3, wscale=w, pad=1),
            conv3a=F.Convolution2D(384,  384,  1, wscale=w),
            conv3b=F.Convolution2D(384,  384,  1, wscale=w),
            conv4=F.Convolution2D(384, 1024,  3, wscale=w, pad=1),
            conv4a=F.Convolution2D(1024, 1024,  1, wscale=w),
            conv4b=F.Convolution2D(1024, self.numClasses,  1, wscale=w),
        )
```

And, I met an error like below
```
  File "/usr/local/lib/python2.7/dist-packages/chainer/function_set.py", line 59, in to_gpu
    func.to_gpu(device=device)
AttributeError: 'int' object has no attribute 'to_gpu'
```

Then, I modified chainer code for survey (to_gpu of chainer/function_set.py)
```
        for func in six.itervalues(self.__dict__):
            print(func)
            # func.to_gpu(device=device)
        return self
```

I got
```
<chainer.functions.convolution_2d.Convolution2D object at 0x7fec36d8ed50>
<chainer.functions.convolution_2d.Convolution2D object at 0x7fec36d8e910>
<chainer.functions.convolution_2d.Convolution2D object at 0x7fec2ed930d0>
<chainer.functions.convolution_2d.Convolution2D object at 0x7fec3cbcd6d0>
<chainer.functions.convolution_2d.Convolution2D object at 0x7fec36d8e410>
0
<chainer.functions.convolution_2d.Convolution2D object at 0x7fec2e0abf50>
[...]

PR #382, opened 2015-09-03
Added typecheck where self.__dict__ is used.

choni:
It's about #381 .
To avoid error when self.\__dict\__ contains something not function.

### DIFF
--- a/chainer/function_set.py
+++ b/chainer/function_set.py
@@ -149,5 +149,4 @@ class FunctionSet(object):
     def _get_sorted_funcs(self):
         return sorted(
             [func_tuple for func_tuple in six.iteritems(self.__dict__)
-             if isinstance(func_tuple, tuple) and
-             isinstance(func_tuple[1], (function.Function, FunctionSet))])
+             if isinstance(func_tuple[1], (function.Function, FunctionSet))])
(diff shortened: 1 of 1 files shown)


okuta:
Thank you for your PR.

`isinstance` returns true in the case of subclass.
And, `FunctionSet` may contain `FunctionSet`.
Therefore, `isinstance(func, (function.Function, function.FunctionSet))` is better.

Chainer project uses PEP8 and a part of OpenStack Style Guidelines related to general coding style.
http://docs.chainer.org/en/stable/contribution.html#coding-guidelines
Please fix coding style.

choni:
OK. Fixed.

choni:
Fixed typecheck in _get_sorted_funcs.
Because, When _get_sorted_funcs is called __dict__ not contains functions but tuples like below.

('conv1b', \<chainer.functions.convolution_2d.Convolution2D object at 0x7fc5a47295d0\>)

okuta:
I think that `__dict__.values()`

okuta:
I think to be good to use `__dict__.values()`.

choni:
I overlooked `iteritems` is used in \_get_sorted_funcs, instead of `itervalues` is used in to_gpu and to_cpu.
About \_get_sorted_funcs, iteritems is used for `sorted` function.
So, I think \__dict\__.values() is not suitable.
And, I simplified if clause of _get_sorted_funcs.

okuta:
OK. I misunderstood.

okuta:
LGTM!
Thank you.

PR #386, opened 2015-09-07
Fix device switching by "with" statement

beam2d:
Device switching does not work. This change fixes it.
(This bug is found by the investigation for the forum report by Amit. Thank you!)


delta2323:
I checked this code. It works well with the MNIST data parallel example.

PR #387, opened 2015-09-07
Fix Optimizer.accumulate_grads

delta2323:
PyCUDA-related codes remained in `Optimizer.accumulate_grads`. This PR removes them.


beam2d:
[LGTM](http://lgtm.in/i/2zJ7TeD5q)

PR #396, opened 2015-09-10
Add unittest for cupy.random.sample

delta2323:
This PR implements unit tests for `cupy.random.sample`

### DIFF
--- a/tests/cupy_tests/random_tests/test_sample.py
+++ b/tests/cupy_tests/random_tests/test_sample.py
@@ -12,12 +12,17 @@ class TestSample(unittest.TestCase):
 
     _multiprocess_can_split_ = True
 
-    def setUp(self):
-        random.random_sample = mock.Mock()
-
     def test_rand(self):
+        random.sample_.random_sample = mock.Mock()
         random.rand(1, 2, 3, dtype=numpy.float32)
-        random.random_sample.assert_call_once_with((1, 2, 3), numpy.float32)
+        random.sample_.random_sample.assert_called_once_with(
+            size=(1, 2, 3), dtype=numpy.float32)
+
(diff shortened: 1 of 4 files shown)


tests/cupy_tests/random_tests/test_sample.py
```diff
@@ -7,3 +11,22 @@
...
+        random.rand(1, 2, 3, dtype=numpy.float32)
+        random.random_sample.assert_call_once_with((1, 2, 3), numpy.float32)
+
+    def test_rand_invalid_argument(self):
+        with self.assertRaises(TypeError):
+            random.rand(1, 2, 3, unnecessary='unnecessary_argument')
+
+    def test_randn(self):
+        random.randn(1, 2, 3, dtype=numpy.float32)
+        random.random_sample.assert_call_once_with((1, 2, 3), numpy.float32)
```
unnonouno: `assert_called_once_with`

tests/cupy_tests/random_tests/test_sample.py
```diff
@@ -7,3 +11,22 @@
 class TestSample(unittest.TestCase):
 
     _multiprocess_can_split_ = True
+
+    def setUp(self):
+        random.random_sample = mock.Mock()
+
+    def test_rand(self):
+        random.rand(1, 2, 3, dtype=numpy.float32)
+        random.random_sample.assert_call_once_with((1, 2, 3), numpy.float32)
```
unnonouno: `assert_called_once_with`

unnonouno:
Do we need to test default `dtype`? I mean we need to call `random.rand(1, 2, 3)` and check if its default dtype is `float`.

unnonouno:
LGTM

PR #413, opened 2015-09-16
Fix memory init

okuta:
This PR modified to `self.ptr` exists when the exception raised(e.g. CUDARuntimeError).


delta2323:
LGTM :+1:

Issue #398, opened 2015-09-10
Strange behavior of 'Linear' transform with shape (x, 1) in GPU mode


lightcaster:
I tried to make a linear transformation from x to 1 dimension using this code:

```python
from chainer import cuda, Variable
from chainer import functions as F

a = Variable(cuda.ones((100,5)))
l = F.Linear(5,1).to_gpu()

print  l(a).data
```
The 'strange' thing is that I'm getting different types of errors. For this particular snippet I got:

```
 ** On entry to SGEMV  parameter number 6 had an illegal value
Traceback (most recent call last):
  File "test_linear_to_1.py", line 7, in <module>
    print  l(a).data
  File "/home/light/prj/chainer/chainer/function.py", line 174, in __call__
    outputs = self.forward(in_data)
  File "/home/light/prj/chainer/chainer/functions/connection/linear.py", line 111, in forward
    Wx = x.dot(self.W.T)
  File "/home/light/prj/chainer/cupy/__init__.py", line 1016, in dot
    return dot(self, b, out)
  File "/home/light/prj/chainer/cupy/linalg/product.py", line 72, in dot
    return _tensordot_core(a, b, out, n, m, k, ret_shape)
  File "/home/light/prj/chainer/cupy/linalg/product.py", line 349, in _tensordot_core
    c.data.ptr, 1)
  File "/home/light/prj/chainer/cupy/cuda/cublas.py", line 238, in sgemv
    check_status(status)
  File "/home/light/prj/chainer/cupy/cuda/cublas.py", line 60, in check_status
    raise CUBLASError(status)
cupy.cuda.cublas.CUBLASError: CUBLAS_STATUS_INVALID_VALUE
```
And with different values of batch size and the size of the first dimension I get NANs or extremely large numbers. 

But if I change this
```python
l = F.Linear(5,1).to_gpu()
```
to this
```python
l = F.Linear(5,2).to_gpu()
```
the error disappears. 

ps I tried to dig deeper but with no luck. Still new to your lib. 
pps Great job btw, I really like the Chainer approach!

jnory:
I met same error too. My chainer version is v1.3.0. 
In my environment, the shape of `a` (in your example) is also affect to the error.
When I changed the shape of `a`, the error disappears.

delta2323:
Hi, thank you for reporting.
We have reproduced the error and now investigating it.

[...]

PR #414, opened 2015-09-16
Add unittest for Bilinear where out_size equals to 1

delta2323:
This PR implements unit tests for the problem on Bilinear reported in #398 (Thanks @cemoody! ).


unnonouno:
LGTM

PR #420, opened 2015-09-21
Fix cupy.expand_dims() is missing in the docs

ronekko:
cupy.expand_dims() is available in version 1.3.1 (#343), but missing in the docs.


okuta:
Thank you for your PR.
LGTM!

PR #436, opened 2015-09-25
Fix type casting

### LINKED ISSUE #421: Typecasting bug
The result of multiplying a float32 cupy ndarray by a constant integer is a float64 array. This is not consistent with numpy.

    import numpy as np
    import cupy as cp
    a = np.zeros((2,)).astype('float32')
    b = cp.zeros((2,)).astype('float32')

    (2 * a).dtype  # float32
    (2 * b).dtype  # float64

okuta:
fix #421.

I fix some cases that are different behavior of the type conversion in numpy and cupy.

### DIFF
--- a/cupy/elementwise.py
+++ b/cupy/elementwise.py
@@ -506,27 +506,31 @@ def _guess_routine_from_dtype(ops, dtype):
     return None
 
 
+def _check_in_args_kind(in_args):
+    all_scalars = True
+    max_array_kind = -1
+    max_scalar_kind = -1
+    for i in in_args:
+        if isinstance(i, cupy.ndarray):
+            kind = _kind_score[i.dtype.kind]
+            all_scalars = False
+            if kind > max_array_kind:
+                max_array_kind = kind
+        else:
+            if isinstance(i, _python_scalar_type):
+                dtype = numpy.dtype(type(i))
+            else:
+                dtype = i.dtype
+            kind = _kind_score[dtype.kind]
+            if kind > max_scalar_kind:
+                max_scalar_kind = kind
+    return not all_scalars and max_array_kind >= max_scalar_kind
+
+
 def _guess_routine(name, cache, ops, in_args, dtype):
     if dtype is None:
-        all_scalars = True
-        max_array_kind = -1
-        max_scalar_kind = -1
-        for i in in_args:
-            if isinstance(i, cupy.ndarray):
-                kind = _kind_score[i.dtype.kind]
-                all_scalars = False
-                if kind > max_array_kind:
-                    max_array_kind = kind
-            else:
-                if isinstance(i, _python_scalar_type):
-                    dtype = numpy.dtype(type(i))
-                else:
-                    dtype = i.dtype
-                kind = _kind_score[dtype.kind]
-                if kind > max_scalar_kind:
-                    max_scalar_kind = kind
-        use_min_scalar = not all_scalars and max_array_kind >= max_scalar_kind
-        if use_min_scalar:
+        use_raw_value = _check_in_args_kind(in_args)
+        if use_raw_value:
             in_types = tuple(in_args)
             op = ()
         else:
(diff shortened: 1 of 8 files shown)


tests/cupy_tests/test_ndarray_elementwise_op.py
```diff
@@ -264,3 +265,42 @@ def test_array_reversed_sub(self):
 
     def test_array_reversed_mul(self):
         self.check_array_reversed_op(operator.mul)
+
+    @testing.for_all_dtypes(no_bool=True)
+    def check_typecast(self, val, dtype):
+        a = val + testing.shaped_arange((5,), numpy, dtype)
+        b = val + testing.shaped_arange((5,), cupy, dtype)
```
delta2323: I think `(shaped_arange((5,)) - 2)` will make stronger test case than `shaped_arange((5,))` because the former checks more boundary cases.

cupy/elementwise.py
```diff
@@ -501,19 +508,43 @@ def _guess_routine_from_dtype(ops, dtype):
...
-        if dtype is None:
-            op = _guess_routine_from_in_types(ops, key)
+        all_scalars = True
+        max_array_kind = -1
+        max_scalar_kind = -1
+        for i in in_args:
+            if isinstance(i, cupy.ndarray):
+                kind = _kind_score[i.dtype.kind]
+                all_scalars = False
+                if kind > max_array_kind:
```
delta2323: We can unify L518-519 and L526-527 into one.

delta2323:
Could you add test cases that typecast of other operations (including multiplication which is reported in #421) is consistent with NumPy and CuPy?

tests/cupy_tests/math_tests/test_misc.py
```diff
@@ -14,7 +14,7 @@ class TestMisc(unittest.TestCase):
     @testing.numpy_cupy_allclose(atol=1e-5)
     def check_unary(self, name, xp, dtype, no_bool=False):
         if no_bool and numpy.dtype(dtype).char == '?':
-            return 0
+            return numpy.int_(0)
```
delta2323: Let me know the reason for this change.

okuta: `testing.numpy_cupy_allclose` needs numpy array/scalar instance for cheking dtype.

tests/cupy_tests/test_ndarray_elementwise_op.py
okuta: Thank you for your suggestion!

okuta:
I added test.

tests/cupy_tests/math_tests/test_misc.py
delta2323: OK, I understand it.

cupy/elementwise.py
```diff
@@ -501,19 +508,43 @@ def _guess_routine_from_dtype(ops, dtype):
...
-                     if isinstance(i, _python_scalar_type) else i.dtype.type
-                     for i in in_args])
-    else:
-        key = dtype
-
-    op = cache.get(key, ())
-    if op is ():
-        if dtype is None:
-            op = _guess_routine_from_in_types(ops, key)
+        all_scalars = True
```
delta2323: Please extract the routine for determining the value of `use_min_scalar` as a separate function.

cupy/elementwise.py
```diff
@@ -501,19 +508,43 @@ def _guess_routine_from_dtype(ops, dtype):
...
+                    max_array_kind = kind
+            else:
+                if isinstance(i, _python_scalar_type):
+                    dtype = numpy.dtype(type(i))
+                else:
+                    dtype = i.dtype
+                kind = _kind_score[dtype.kind]
+                if kind > max_scalar_kind:
+                    max_scalar_kind = kind
+        use_min_scalar = not all_scalars and max_array_kind >= max_scalar_kind
```
delta2323: Please extract the routine for determining the value of `use_min_scalar` as a separate function.

delta2323:
LGTM except the comments

delta2323:
LGTM :+1:

Issue #438, opened 2015-09-25
MemoryPointer.copy_from_peer is missing


delta2323:
`cupy.creation.from_data.copy` uses `MemoryPointer.copy_from_peer` to copy an array from the device other than current one (See the [corresponding code](https://github.com/pfnet/chainer/blob/9f6fa3a43cd8f6718a25bf6b3cbec8fae34854da/cupy/creation/from_data.py#L144)). But, `copy_from_peer` is not implemented and cross-device copy is impossible if we solely use cupy module.

Note: We can do cross-device copy if we are allowed to use chainer module because `chainer.cuda.copy` supports it.

delta2323:
After the investigation, I found that current cross-device copy is realized by undocumented behavior of `cupy`. 

To be concrete, we can copy arrays across devices by `chainer.cuda.cupy`, `cupy.copy` or `cupy.copyto`. All of three methods realize cross-device copy by `MemoryPointer.copy_from_device`. But its document says this API is for the copy of GPU array within single device.

The followings items are notes for myself.
* `chainer.cuda.copy` invokes `cupy.copyto`, which is an alias of `cupy.manipulation.basic.copyto` ([Link](https://github.com/pfnet/chainer/blob/master/chainer/cuda.py#L346)).
* `cupy.copyto` invokes `cupy.elementwise.copy`, `cupy.elementwise.copy_where`, or `MemoryPointer.copy_from` depending on the arguments ([Link](https://github.com/pfnet/chainer/blob/master/cupy/manipulation/basic.py#L6)). The former two element-wise methods cannot operate cross-device copy. So the only way to do cross-device copy is to invoke `MemoryPointer.copy_from`.
* Likewise, `cupy.copy`, which is an alias of `cupy.creation.from_data.copy` invokes one of `cupy.elementwise.copy`, `MemoryPointer.copy_from_peer`, or  `MemoryPointer.copy_from` ([Link](https://github.com/pfnet/chainer/blob/master/cupy/creation/from_data.py#L117)). First one cannot do cross-device copy and second one is not implemented. So the situation is same as `cupy.copy` and we do cross-device copy by means of `MemoryPointer.copy_from`.
* If source array is allocated in the device, `MemoryPointer.copy_from` is same as `MemoryPoint.copy_from_device` ([Link](https://github.com/pfnet/chainer/blob/master/cupy/cuda/memory.py#L93)).

Possible solutions are:
* to change the role of `copy_from_device` so that it admits cross-device copy.
[...]

PR #440, opened 2015-09-28
Fixes relating to cross-device array copy

delta2323:
This PR implements
* multi_gpu decorator for cupy
* unittests for `cupy.copy`, `cupy.copyto` and `cupy.elementwise.copy` including multi-GPU ones.
* fixes for #438 
    * Contrary to what I commented there, I tried to support cross-device copy in non-UVA devices. But since I do not have such environment, I could not check the corresponding code.

### DIFF
--- a/cupy/creation/from_data.py
+++ b/cupy/creation/from_data.py
@@ -3,7 +3,6 @@ import ctypes
 import numpy
 
 import cupy
-from cupy import cuda
 from cupy import elementwise
 
 
--- a/cupy/creation/from_data.py
+++ b/cupy/creation/from_data.py
@@ -137,19 +137,10 @@ def copy(a):
     if a.size == 0:
         return cupy.empty_like(a)
 
-    if a.data.device != cuda.Device():
-        # peer copy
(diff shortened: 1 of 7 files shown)


delta2323:
I was pointed out that current implementation already assumes UVA functions, so it is of no use that only `copy_from_device` supports non-UVA devices.

cupy/creation/from_data.py
```diff
@@ -137,19 +136,10 @@ def copy(a):
...
-
-    # in-device copy
     newarray = cupy.empty_like(a)
-    if a.flags.c_contiguous:
-        newarray.data.copy_from(a.data, a.nbytes)
-    else:
-        elementwise.copy(a, newarray)
+    if not a.flags.c_contiguous:
+        a = ascontiguousarray(a)
+    newarray.data.copy_from(a.data, a.nbytes)
```
okuta: Copy will be performed twice If same device and c_contiguous.

okuta:
LGTM except comment.

okuta:
LGTM!

PR #442, opened 2015-09-29
Unittest for cupy.get

### LINKED ISSUE #430: Add tests on cupy.asnumpy
All of decorators in `cupy.testing.array` check the equivalence of two `cupy.ndarray`s by converting them into `numpy.ndarray` with `cupy.asnumpy`. Since these decorators are used in many unit tests, we should test `cupy.asnumpy` to validate the correctness of this function.

`cupy.asnumpy` is essentially same as `cupy.ndarray.get`. So we will also have to test the latter one.

delta2323:
This PR implements unit tests for `cupy.get`.

This PR essentially fixes #430 because `cupy.asnumpy` is same as `cupy.get` get the source array is on the device.

### DIFF
--- a/cupy/cuda/memory.py
+++ b/cupy/cuda/memory.py
@@ -198,8 +198,8 @@ class MemoryPointer(object):
 
         """
         if size > 0:
-            runtime.memcpyAsync(mem, self.ptr, size, stream,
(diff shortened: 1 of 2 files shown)


okuta:
LGTM!

PR #443, opened 2015-09-29
Restore mocked objects in teardown of RandomState's unit tests

delta2323:
Some test fixtures in `RandomState`'s unit tests substitute objects with mocking ones. But they leave the mocks as they are and it does harm to other unit tests.
This PR fixes so that mocked objects are properly restored in teardown.


okuta:
LGTM!

PR #462, opened 2015-10-01
Use function.__name__ instead of func_name to support Python3

delta2323:
`functon.func_name` is not available in Python 3. This PR change to use `__name__` instead of `func_name`


okuta:
LGTM!

PR #466, opened 2015-10-01
Minimize the effect of `numpy.seterr`

okuta:
I changed tests to minimize the effect of `numpy.seterr`.
Please merge #464 and #465 before this PR.


unnonouno:
LGTM

PR #473, opened 2015-10-02
Do not generate graph.wo_split.dot

delta2323:
Since `Splitter` is not available in `model` architecture, `graph.wo_split.dot` and `graph.dot` is exactly same DAG. So, we do not need to generate the former one.


beam2d:
Thanks!

PR #480, opened 2015-10-03
Add chainer Functions Transpose and Swapaxes.

tscohen:
Simple wrappers around the corresponding cupy functions.

### DIFF
--- a/chainer/functions/array/transpose.py
+++ b/chainer/functions/array/transpose.py
@@ -8,7 +8,7 @@ class Transpose(function.Function):
     """Permute the dimensions of an array."""
 
     def __init__(self, axes):
-        self.axes = axes
+        self.axes = tuple(ax % len(axes) for ax in axes)
 
     def check_type_forward(self, in_types):
         type_check.expect(in_types.size() == 1,)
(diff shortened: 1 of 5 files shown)


chainer/functions/array/transpose.py
```diff
@@ -0,0 +1,44 @@
...
+        return 'Transpose'
+
+    def forward(self, inputs):
+        x = inputs[0]
+        y = x.transpose(self.axes)
+        return y,
+
+    def backward(self, inputs, grad_outputs):
+        gy = grad_outputs[0]
+        invAxes = numpy.argsort(self.axes)
```
beam2d: Use snake case for local variables.

chainer/functions/array/transpose.py
```diff
@@ -0,0 +1,44 @@
...
+        gx = gy.transpose(invAxes)
```
beam2d: Currently, it seems CuPy's transpose does not support numpy.ndarray argument. Use a tuple instead.

beam2d:
I added two comments.

beam2d:
LGTM!

beam2d:
Thank you!

PR #489, opened 2015-10-06
Fix contiguity after reshape

beam2d:
Contiguity is wrong after cupy.reshape:
```python
a = cupy.arange(6).reshape(2, 3)
a.flags.f_contiguous  #=> 1
```
It causes incorrect results on successive computations on the reshaped array. This PR fixes this issue.


unnonouno:
LGTM

PR #496, opened 2015-10-07
Raise type error in to_cpu if array is neither numpy.ndarray nor cupy.ndarray

### LINKED ISSUE #491: cuda.to_cpu does not move a Variable to CPU
It even does not raises an error. It silently returns the original variable without any modification. This behavior is confusing.

delta2323:
This PR fixes #491


beam2d:
Would you add a test?

beam2d:
LGTM

beam2d:
Thank you!

PR #529, opened 2015-10-20
Fix setter of FunctionSet.parameters and FunctionSet.gradients

### LINKED ISSUE #502: The behavior of Function#parameters= is inconsistent with FunctionSet#parameters=
At aaf2aae14ad683734964590e6ceee9d775d9d7c5, an assertion was introduced into `Function#parameters=`, which requires the argument `values` to be a list-like object (implementing `__len__` method).
However, this assertion seems to be inconsistent with the behavior of `FunctionSet#parameters=`.
`FunctionSet#parameters=` passes the iterator of its argument `params` to its child functions through `Function#parameters=`, and it causes a TypeError exception.

```python
from chainer import FunctionSet
import chainer.functions as F
func_set = FunctionSet(f=F.Linear(10,10))
params = func_set.parameters
func_set.parameters = params # TypeError: object of type 'tuple_iterator' has no len()
```

delta2323:
Fix #502. `FunctionSet.gradients` have same problems as `FunctionSet.parameters`. This PR also fixes `gradients` .

### DIFF
--- a/chainer/function_set.py
+++ b/chainer/function_set.py
@@ -151,7 +151,7 @@ class FunctionSet(object):
     def gradients(self, grads):
         funcs = self._get_sorted_funcs()
 
-        expected_grads_size = sum([len(func.gradients) for _, func in funcs])
+        expected_grads_size = sum(len(func.gradients) for _, func in funcs)
         assert expected_grads_size == len(grads)
 
         idx = 0
--- a/chainer/function_set.py
+++ b/chainer/function_set.py
@@ -128,7 +128,7 @@ class FunctionSet(object):
     def parameters(self, params):
         funcs = self._get_sorted_funcs()
 
-        expected_params_size = sum([len(func.parameters) for _, func in funcs])
+        expected_params_size = sum(len(func.parameters) for _, func in funcs)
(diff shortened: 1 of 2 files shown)


chainer/function_set.py
```diff
@@ -126,9 +126,12 @@ def parameters(self):
...
     @parameters.setter
     def parameters(self, params):
-        param_iter = iter(params)
+        idx = 0
         for _, func in self._get_sorted_funcs():
-            func.parameters = param_iter
+            param_size = len(func.parameters)
+            func.parameters = params[idx: idx + param_size]
+            idx += param_size
+        assert idx == len(params)
```
unnonouno: Why don't you check sum of length of params before assigning? Such code is exception-safe.

unnonouno:
LGTM except one comment.

chainer/function_set.py
delta2323: I agree with you. I will change to check the assertion before assignment.

delta2323:
I fixed the point @unnonouno pointed out.

unnonouno:
LGTM

PR #532, opened 2015-10-20
Unify bias argument and initial_bias argument. Make bias deprecated.

### LINKED ISSUE #525: Deprecate bias parameter of Convolution2D?
Currently, Convolution2D has a bias and initial_bias argument. The documentation of bias is "Initial bias value". So it seems like they are redundant. The code currently expects an array for initial_bias and a scalar for bias, but these could both be supported by initial_bias. Or alternatively, we could remove the nobias parameter as well and just have a bias parameter that can be a number, an array, or None (for no bias).

delta2323:
Fix #525 
As @tscohen and @beam2d are now discussing the design of `Linear` and `Convolution2D`, I should defer this PR until the discussion.

### DIFF
--- a/chainer/functions/connection/convolution_2d.py
+++ b/chainer/functions/connection/convolution_2d.py
@@ -1,5 +1,6 @@
 import ctypes
 import math
+import warnings
 
 import numpy
 from six import moves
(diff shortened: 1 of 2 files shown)


delta2323:
Memo
* Add unittest that checks default behavior is unchanged
* Consider if we should apply same change to `Bilinear`.

delta2323:
Parameter initialization is redesigned in #820. It is possible and better to work on what I want to do here in the PR. So I close this PR.

PR #535, opened 2015-10-22
Add AlexNet to Imagenet example

delta2323:
AlexNet used in `imagenet` example is different from original one in that Local Response Normalization(LRN) layers are replaced with Batch Normalizations(BN) (so, the architecture name is `alexbn` rather than `alex`). The reason for this change is that LRN was not implemented at the time the example is made. This PR implements original AlexNet and adds it to `imagenet` example.

### DIFF
--- a/examples/imagenet/alex.py
+++ b/examples/imagenet/alex.py
@@ -4,7 +4,7 @@ import chainer.functions as F
 
 class Alex(chainer.FunctionSet):
 
-    """Single-GPU AlexNet.
+    """Single-GPU AlexNet without partition toward the channel axis.
 
     """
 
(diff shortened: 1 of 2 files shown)


okuta:
The original AlexNet has layers which are divided into two at channel axis.
Its structure has been reproduced in caffe.
What do you think is good for us.

delta2323:
Ah, sorry. I missed the structure. I think we can defer the implementation of the feature until we need to reproduce the result exactly. In that case, we should add comment the difference of my implementation and original one.

okuta:
OK. Please add comment.

okuta:
LGTM!

PR #539, opened 2015-10-22
Restore chainer_seed in the teardown of TestGetRandomState2

delta2323:
`CHAINER_SEED` is set but is not restored in the teardown of `TestGetRandomState2`. This PR fixes the problem.

### DIFF
--- a/tests/cupy_tests/random_tests/test_generator.py
+++ b/tests/cupy_tests/random_tests/test_generator.py
@@ -234,8 +234,7 @@ class TestGetRandomState2(unittest.TestCase):
         generator.RandomState = self.rs_tmp
         generator._random_states = self.rs_dict
         if self.chainer_seed is None:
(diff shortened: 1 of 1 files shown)


tests/cupy_tests/random_tests/test_generator.py
```diff
@@ -228,10 +228,16 @@ def setUp(self):
         generator.RandomState = mock.Mock()
         self.rs_dict = generator._random_states
         generator._random_states = {}
+        self.chainer_seed = os.getenv('CHAINER_SEED')
 
     def tearDown(self, *args):
         generator.RandomState = self.rs_tmp
         generator._random_states = self.rs_dict
+        if self.chainer_seed is None:
+            if 'CHAINER_SEED' in os.environ:
```
okuta: I think that `os.environ.pop('CHAINER_SEED', None)` is better.

okuta:
LGTM!

PR #545, opened 2015-10-23
Add min and max alias for amin and amax, to mimick numpy.

tscohen:
Numpy has np.min and np.max functions. It's true that np.min == np.amin and np.max == np.amax, so I added the same aliases to cupy.


unnonouno:
LGTM

PR #548, opened 2015-10-25
Improve reduction performance

okuta:
This PR improve reduction performance for a large shape array.

PR #549, opened 2015-10-25
Improve batch_normalization performance

okuta:
Speeding up from 1.6 to 1.8 times in the following test case .
It's tested in branch which PR #548 is merged.

before
0:00:00.312393 0:00:00.204991
0:00:00.321782 0:00:00.205670
after 
0:00:00.167545 0:00:00.124970
0:00:00.175646 0:00:00.125602

```python
import datetime

import chainer
import chainer.functions as F

def testbn(shape):
    bn = F.BatchNormalization(shape[0])
    bn.to_gpu()
    x_org = chainer.Variable(chainer.cuda.zeros(shape), volatile=False)
    x = x_org
    t1 = datetime.datetime.now()
    for i in range(200):
        x = bn(x)
    y = F.sum(x)
    y.data.get()
    t2 = datetime.datetime.now()
    y.backward()
    t3 = datetime.datetime.now()
    print(t2 - t1, t3 - t2)

#for memory allocation
testbn((10, 10, 10))
testbn((100, 100, 100))

#for performance test
testbn((10, 10, 10))
testbn((100, 100, 100))
```

### DIFF
--- a/chainer/functions/normalization/batch_normalization.py
+++ b/chainer/functions/normalization/batch_normalization.py
@@ -99,10 +99,18 @@ class BatchNormalization(function.Function):
             var = self.avg_var
 
         self.std = xp.sqrt(var, dtype=var.dtype)
-        x_mu = x - mean
-        self.x_hat = x_mu / self.std
-        y = self.gamma * self.x_hat
-        y += self.beta
+
+        if xp == numpy:
+            self.x_hat = (x - mean) / self.std
+            y = self.gamma * self.x_hat
+            y += self.beta
+        else:
+            self.x_hat, y = cuda.elementwise(
+                'T x, T mean, T std, T gamma, T beta', 'T x_hat, T y', '''
+                T x_hat_ = (x - mean) / std;
+                x_hat = x_hat_;
+                y = gamma * x_hat_ + beta;''', 'bn_fwd1')(
+                    x, mean, self.std, self.gamma, self.beta)
 
         # Compute exponential moving average
         if self.use_batch_mean:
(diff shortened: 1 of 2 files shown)


chainer/functions/normalization/batch_normalization.py
```diff
@@ -135,11 +152,19 @@ def backward(self, x_orig, gy):
...
         self.ggamma += ggamma
 
-        coeff = self.gamma / self.std
-        gbeta /= m
-        ggamma /= m
+        if cuda.get_array_module(*x_orig) == numpy:
+            coeff = self.gamma / self.std
+            gbeta *= inv_m
+            ggamma *= inv_m
+            gx = coeff * (gy - self.x_hat * ggamma - gbeta)
```
unnonouno: How about `gx = coeff * (gy - (self.x_hat * ggamma - gbeta) / m)`?
I think `inv_m` is not requried.

chainer/functions/normalization/batch_normalization.py
```diff
@@ -135,11 +152,19 @@ def backward(self, x_orig, gy):
...
+        else:
+            gx = cuda.elementwise(
+                'T gy, T gbeta, T ggamma, T x_hat, T gamma, T std, T inv_m',
+                'T gx',
+                'gx = gamma / std * (gy - (x_hat * ggamma + gbeta) * inv_m)',
```
unnonouno: ```gx = gamma / std * (gy - (x_hat * ggamma + gbeta) / (ldim * rdim))```?

unnonouno:
LGTM except the comments.

chainer/functions/normalization/batch_normalization.py
unnonouno: OK, I noticed that `ldim` and `rdim` are scalar.

okuta:
I fixed.

unnonouno:
LGTM

PR #550, opened 2015-10-25
Implement chainer.testing.product and , cupy.testing.product

delta2323:
This PR implements `chainer.testing.product` and `cupy.testing.product`. Basic usage (and implementation itself) is same as that of `maflib.util.product` ([Link](https://pfi.github.io/maf/maflib.html#maflib.util.product)).

### DIFF
--- a/chainer/testing/__init__.py
+++ b/chainer/testing/__init__.py
@@ -3,6 +3,7 @@ import nose
 from chainer.testing import parameterized
 
 parameterize = parameterized.parameterize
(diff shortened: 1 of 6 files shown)


unnonouno:
Do we need to write test code for this method?

delta2323:
Actually, yes. I will write it.

delta2323:
I added unit tests.

unnonouno:
LGTM

PR #553, opened 2015-10-26
Add contrastive loss

mitmul:
This PR adds contrastive loss function. It enables us to implement Siamese Network. Here is an example: [chainer-siamese](https://github.com/mitmul/chainer-siamese)


chainer/functions/loss/contrastive.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def check_type_forward(self, in_types):
+        type_check.expect(in_types.size() == 3)
+
+        x0_type, x1_type, y_type = in_types
+        type_check.expect(
+            x0_type.dtype == numpy.float32,
+            x1_type.dtype == numpy.float32,
+            x0_type.shape == x1_type.shape,
+            x0_type.shape[0] == x1_type.shape[0],
```
delta2323: This equality is redundant because we have already checked it in the previous line.

chainer/functions/loss/contrastive.py
```diff
@@ -0,0 +1,90 @@
...
+        gx0 += alpha * (1 - y) * mdist_p * mdist * -(self.diff / dist)
+        gx0 = gx0.astype(xp.float32)
+
+        return gx0, -gx0, None
+
+
+def contrastive(x0, x1, y, margin=1):
+    """Computes contrastive loss.
+
+    It takes a variable pair and a label as inputs. The label is 1 when those
```
delta2323: "a pair of variables" is more appropriate.

chainer/functions/loss/contrastive.py
```diff
@@ -0,0 +1,90 @@
...
+    two input variables are similar, or 0 when they are dissimilar. Let
+    :math:`N` and :math:`K` denote mini-batchsize and the dimension of input
+    variables, respectively, the shape of both input variables should be
```
delta2323: This sentence ends at "respectively". The latter half should be separated as another sentence.

delta2323:
Could you add the reference to this loss function?

delta2323:
I recommend you add the remark that this loss function is used in siamese network to the docstring.

chainer/functions/loss/contrastive.py
```diff
@@ -0,0 +1,90 @@
...
+    def __init__(self, margin):
+        self.margin = float(margin)
+
+    def check_type_forward(self, in_types):
+        type_check.expect(in_types.size() == 3)
+
+        x0_type, x1_type, y_type = in_types
+        type_check.expect(
+            x0_type.dtype == numpy.float32,
+            x1_type.dtype == numpy.float32,
```
delta2323: Can we restrict the dtype of `y_type` ?

chainer/functions/loss/contrastive.py
```diff
@@ -0,0 +1,90 @@
...
+
+    It takes a variable pair and a label as inputs. The label is 1 when those
+    two input variables are similar, or 0 when they are dissimilar. Let
+    :math:`N` and :math:`K` denote mini-batchsize and the dimension of input
+    variables, respectively, the shape of both input variables should be
+    (N, K).
+
+    .. math::
+        L = \\frac{1}{2N} \\left( \\sum_{n=1}^N y_n d_n
+            + (1 - y_n) \\max ({\\rm margin} - \\sqrt{d_n}, 0)^2 \\right)
```
delta2323: It seems this formula is different from the one implemented in `forward`. I guess `forward` operation is correct judging from the Caffe implementation. Is my suggestion correct? If we fix the document, `d_n` and `\sqrt{d_n}` should be replaced with `d_n^2` and `d_n`, respectively.

chainer/functions/loss/contrastive.py
```diff
@@ -0,0 +1,90 @@
...
+
+    where :math:`N` denotes the mini-batch size, and
+    :math:`d_n = \\| {\\bf x_0}_n - {\\bf x_1}_n \\|_2`, and
+    :math:`{\\bf x_0}_n` means the n-th K-dimensional vector in a mini-batch.
+
```
delta2323: I think we can treat `x_0_n` as well as `N` and `d_n` as the objectives of the verb "denotes", although it is just the matter of tastes.

delta2323: We should add the description of `{x_1}_n` , too.

chainer/functions/loss/contrastive.py
```diff
@@ -0,0 +1,90 @@
...
+    Args:
+        x0 (~chainer.Variable): The first input variable.
+        x1 (~chainer.Variable): The second input variable.
+        y (~chainer.Variable): Labels. All values should be 0 or 1.
+        margin (int): A parameter for contrastive loss.
```
delta2323: I think we should add the requirement on the shapes of `x0`, `x1` and y.

chainer/functions/loss/contrastive.py
```diff
@@ -0,0 +1,90 @@
...
+            x0_type.ndim == 2,
+            x1_type.ndim == 2,
+            y_type.ndim == 1
+        )
+
+    def forward(self, inputs):
+        xp = cuda.get_array_module(*inputs)
+        x0, x1, y = inputs
+
+        self.diff = x0 - x1  # N x 2
```
delta2323: What does this comment mean?

chainer/functions/loss/contrastive.py
```diff
@@ -0,0 +1,90 @@
...
+
+        return xp.array(loss, dtype=xp.float32),
+
+    def backward(self, inputs, gy):
+        xp = cuda.get_array_module(*inputs)
+        x0, x1, y = inputs
+
+        y = xp.vstack((y, y)).T
+        alpha = gy[0] / y.shape[0]
+        dist = xp.vstack((self.dist, self.dist)).T
```
delta2323: `xp.vstack` works well only when `x0.shape[1] = x1.shape[1] = 2`, which is the case tested.
It would work well if `xp.vstack` `xp.repeat` . But currently CuPy does not implemented it.

delta2323:
This branch has conflicts with master branch. So, please rebase or merge it.

mitmul:
Thank you for reviewing. I fixed some points except the problem around `xp.vstack`. Once `xp.repeat` is merged, I will fix it.

delta2323:
@mitmul Thank you very much! I will review it.

delta2323:
@mitmul Now `repeat` is added by #604. Could you fix the code?

delta2323:
We are planning to merge this PR as a new feature of v1.5.0. As we will adopt new model architecture (#573) in v1.5.0, we would like to make all functions (including Contrastive Loss) be based on this design.
I will send a PR applying this design change to Contrastive Loss function to your branch. So could you merge it when my PR is submitted?

mitmul:
@delta2323 Of course. Please send a PR to my repo :)

delta2323:
What happens when `margin` is negative? Although it works even if `margin` is negative, I think we do not set so. Should we repel the negative margin case?

chainer/functions/loss/contrastive.py
```diff
@@ -0,0 +1,103 @@
...
+
+    Args:
+        x0 (~chainer.Variable): The first input variable. The shape should be
+            (N, K), where N denotes the minibatch size, and K denotes the
+            dimension of x0.
+        x1 (~chainer.Variable): The second input variable. The shape should be
+            the same as x0.
+        y (~chainer.Variable): Labels. All values should be 0 or 1. The shape
+            should be (N,), where N denotes the minibatch size.
+        margin (int): A parameter for contrastive loss.
```
delta2323: I think margin should be `float` rather than int.

mitmul:
I added a check of margin's sign in the __init__ function of Contrastive class.

delta2323:
I agree with you about negative margin. Then could you add test cases for negative margin case?

mitmul:
@delta2323 I added an additional test case for checking negative margin value. I'd appreciate it if you could review it again.

chainer/functions/loss/contrastive.py
```diff
@@ -0,0 +1,107 @@
...
+from chainer.utils import type_check
+
+
+class Contrastive(function.Function):
+
+    """Contrastive loss function."""
+
+    def __init__(self, margin):
+        if margin <= 0:
+            raise Exception("margin should be positive value.")
```
delta2323: Use `ValueError` for invalid values.

delta2323:
LGTM :+1:
Thank you very much. We appreciate your great work!

PR #554, opened 2015-10-26
Directly manipulate os.environ to remove environment variable

delta2323:
As the [Python document](https://docs.python.org/2.7/library/os.html#os.unsetenv) indicates, it is preferable to manipulate `os.environ` directly to remove environment variable instead of using `os.unsetenv`. Actually my environment does not have bash command `unsetenv` and `os.unsetenv` does not take effect.

### DIFF
--- a/tests/cupy_tests/random_tests/test_generator.py
+++ b/tests/cupy_tests/random_tests/test_generator.py
@@ -239,7 +239,7 @@ class TestGetRandomState2(unittest.TestCase):
             os.environ['CHAINER_SEED'] = self.chainer_seed
 
     def test_get_random_state_no_chainer_seed(self):
-        os.unsetenv('CHAINER_SEED')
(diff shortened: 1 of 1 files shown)


tests/cupy_tests/random_tests/test_generator.py
```diff
@@ -239,7 +239,7 @@ def tearDown(self, *args):
             os.environ['CHAINER_SEED'] = self.chainer_seed
 
     def test_get_random_state_no_chainer_seed(self):
-        os.unsetenv('CHAINER_SEED')
+        os.environ.pop('CHAINER_SEED', None)
```
unnonouno: Why not use `del os.environ['CHAINER_SEED']`?

unnonouno: OK, it's a little long. `pop` is cool.

unnonouno:
LGTM

PR #558, opened 2015-10-27
Add min and max functions to chainer.

tscohen:
Add differentiable min and max functions to chainer. Code and tests are based on the code for Sum(). Other selections like the median could be easily added by deriving from the base class (but right now median is not implemented in cupy).


okuta:
I think that `@condition.retry(3)` is not needed for tests of min and max functions.
Because, min and max functions have no numerical error.

chainer/functions/math/max.py
```diff
@@ -0,0 +1,111 @@
...
+        y = self.y.reshape(shape)
+
+        # Compute the gradient
+        return gy * (x == y),
+
+
+class Max(SelectorBase):
+
+    def __init__(self, axis=None, keepdims=False):
+        super(Max, self).__init__(axis, keepdims)
```
delta2323: You can remove this `__init__` since it only calls `__init__` of super class. Base class's `__init__` is automatically called if derived class does not implement its own `__init__` method.

chainer/functions/math/max.py
```diff
@@ -0,0 +1,111 @@
...
+
+    def _fwd(self, x, xp):
+        return xp.amax(x, axis=self.axis, keepdims=self.keepdims)
+
+
+class Min(SelectorBase):
+
+    def __init__(self, axis=None, keepdims=False):
+        super(Min, self).__init__(axis, keepdims)
```
delta2323: ditto

okuta:
How about change module name to `minmax` or other?

tscohen:
I considered another name but chose the name 'max' because exponential.py also implements the log function. But I agree this is clearer so I changed it to 'minmax'.

okuta:
Thank you for changing file name.
Please change test file name too.

okuta:
I was mistaken about backward tests.
A gradient of min and max function is discontinuous.
Thus, results of `numerical_grad` in backward test might rarely unstable.
Please add `@condition.retry(3)` for backward tests, Sorry.

You can see that tests become unstable by using `numpy.ones` instead of `numpy.random.uniform` in `setUp`.

okuta:
Thank you for fixing.
Please add reference to `docs/source/reference/functions.rst`.

okuta:
LGTM except the comment.

okuta:
Thank you for fixing.

tscohen:
Will this get merged for 1.5?

PR #566, opened 2015-10-28
Fix 0-size malloc

okuta:
This PR fix error at the time of gpu memory release.

### DIFF
--- a/cupy/cuda/memory.py
+++ b/cupy/cuda/memory.py
@@ -307,7 +307,7 @@ class SingleDeviceMemoryPool(object):
 
     def malloc(self, size):
         if size == 0:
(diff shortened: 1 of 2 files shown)


unnonouno:
Can you write a test for this patch?

delta2323:
I run the test code you added without applying this PR and found that it does not fail. So I am afraid the test code cannot check it does solve the error.

unnonouno:
OK, please add `assertNotIsInstance(mem, PooledMemory)`.

okuta:
Thank you for the information.
I fixed the test case.

unnonouno:
LGTM

PR #586, opened 2015-11-01
Fix count_nonzero and argmin/argmax

### LINKED ISSUE #585: cupy.count_nonzero does not work correctly
count_nonzero(x) should return the count of non-zero elements in x. The current implementation is not correct.

```python
>>> a = cupy.eye(4)
>>> cupy.count_nonzero(a)
1
```

Note that this bug is overlooked because the test also has a bug. `testing.shaped_random` with dtype=bool almost always returns an array filled by False. It seems a bug of shaped_random(dtype=bool), since it should return a random boolean array. We should fix it, too.

beam2d:
Fix #585.

### DIFF
--- a/tests/cupy_tests/sorting_tests/test_count.py
+++ b/tests/cupy_tests/sorting_tests/test_count.py
@@ -1,5 +1,8 @@
 import unittest
 
+import numpy
+
+import cupy
 from cupy import testing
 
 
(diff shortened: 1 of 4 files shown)


beam2d:
After fixing shaped_random, I found that cupy.argmin and cupy.argmax also behave different from those in numpy, when there are multiple maximum values in the target axis. I also fixed it in this PR.

tests/cupy_tests/sorting_tests/test_count.py
```diff
@@ -9,13 +12,12 @@ class TestCount(unittest.TestCase):
...
-        m = testing.shaped_random((2, 3), xp, xp.bool_)
-        a = testing.shaped_random((2, 3), xp, dtype) * m
-        c = xp.count_nonzero(a)
-        self.assertIsInstance(c, int)
-        return c
+    def test_count_nonzero(self, dtype):
+        m = testing.shaped_random((2, 3), numpy, bool)
+        a = testing.shaped_random((2, 3), numpy, dtype) * m
+        c = numpy.count_nonzero(a)
+        d = cupy.count_nonzero(cupy.array(a))
```
unnonouno: Do we need to check types of `c` and `d`?

beam2d: I added a type test.

beam2d:
I reverted the test code.

unnonouno:
LGTM

PR #588, opened 2015-11-01
Fix RmsProp

kashif:
need to add the eps so that sqrt does not overflow


okuta:
Thank you for PR.
Please add comments to class docstring if there is reference for fixed RMSProp.

beam2d:
Would you experience the overflow in RMSprop? If so, you can raise the eps value via initializer to avoid the overflow instead of altering the update rule.

kashif:
true @beam2d I am thinking of closing this commit...

PR #609, opened 2015-11-09
Add index of connectionist temporal classification to documents.

jnishi:
Add index of connectionist temporal classification to documents.


okuta:
LGTM!

PR #617, opened 2015-11-11
Fix sentiment example for Chain

delta2323:
This PR fixes sentiment example to catch up the latest `chain` branch.

Changes themselves are almost same as #494 (except using `Chain` instead of `DictLink`). But I could not merge `chain` branch into #494 because of tremendous conflicts.

### DIFF
--- a/examples/sentiment/train_sentiment.py
+++ b/examples/sentiment/train_sentiment.py
@@ -19,6 +19,7 @@ import numpy as np
 import chainer
 from chainer import cuda
 import chainer.functions as F
+import chainer.links as L
 from chainer import optimizers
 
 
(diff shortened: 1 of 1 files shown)


beam2d:
LGTM except one point: use `model.zerograds()` instead of `optimizer.zero_grads()`.

examples/sentiment/train_sentiment.py
```diff
@@ -166,14 +184,13 @@ def evaluate(test_trees):
...
     random.shuffle(train_trees)
     for tree in train_trees:
-        loss, v = traverse(tree, train=True)
+        loss, v = traverse(model, tree, train=True)
         accum_loss += loss
         count += 1
 
         if count >= batchsize:
-            optimizer.zero_grads()
+            model.zero_grads()
```
beam2d: no underscore here (model.zerograds())

beam2d:
Thank you!!

PR #621, opened 2015-11-14
Fix transpose() when called with no "axes"

kiyukuta:
transpose() causes error when argument “axes” is None because Transpose.\_\_init\_\_()  assumes that the “axes” is iterable.

I added a line to fix it (the default behavior of “axes” is to reverse the dimension of input x).

### DIFF
--- a/chainer/functions/array/transpose.py
+++ b/chainer/functions/array/transpose.py
@@ -7,8 +7,10 @@ from chainer.utils import type_check
 class Transpose(function.Function):
     """Permute the dimensions of an array."""
 
-    def __init__(self, axes):
-        self.axes = tuple(ax % len(axes) for ax in axes)
+    def __init__(self, axes=None):
+        self.axes = axes
+        if axes:
+            self.axes = tuple(ax % len(axes) for ax in axes)
 
     def check_type_forward(self, in_types):
         type_check.expect(in_types.size() == 1,)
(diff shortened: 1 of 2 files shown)


okuta:
I think that it may be better to fix the `Transpose.__init__`.
What do you think?

kiyukuta:
I agree with you.
Since Transpose do not know the input shape until forward() is called, I changed both __init__() and forward().

okuta:
How about change `backward`?
Because `x.transpose(None)` is possible, it becomes simpler.

And, please add test for `None` case.

kiyukuta:
Yes, you're right. I moved ```axes = tuple(ax % len(self.axes) for ax in self.axes)``` to ```Transpose.backward``` and finally it looks simpler. How do you like it?

By the way, when I test it, I encountered another problem with cupy.transpose in GPU mode.
Specifically, ```x.transpose(None)``` causes error (```x.transpose()``` and ```cupy.transpose(x, None)``` work fine).

The reason is argument ```*axes``` in ```cupy.ndarray.transpose``` (```None``` becomes ```(None,)``` ).
https://github.com/pfnet/chainer/blob/master/cupy/__init__.py#L424

I think we should add another if-statement to catch ```axes==(None,)``` in ```cupy.ndarray.transpose```.
Anyway, should I make another issue/PR about this or continue in here?

okuta:
Thank you for information of bug.
Please make another issue/PR.

Please fix style.
https://travis-ci.org/pfnet/chainer/jobs/91315975#L831

kiyukuta:
Thanks for your comments.

ok, I'll send another PR.

okuta:
Thank you!
LGTM.

okuta:
The axes issue fixed #587.

PR #629, opened 2015-11-16
Remove h5py from install_requires on Read the Docs.

okuta:
The h5py is reguired hdf5.h.
But, there is no way to install the hdf5.h on Read the Docs.


unnonouno:
LGTM

PR #632, opened 2015-11-16
Fix cupy.ndarray.transpose and add test

kiyukuta:
```x.transpose(None)``` causes error because  ```cupy.ndarray.transpose``` receives variable-length argument list ```*axes``` and thus the ```None``` become ```(None,)```.


okuta:
LGTM!

PR #637, opened 2015-11-17
Remove duplicated lines

mitmul:
This PR removes duplicated lines in docs. I think `Noise injecting functions` part should be deleted.


delta2323:
LGTM :+1: Thank you for the catch.

PR #591, opened 2015-11-02
Add Deconvolution2D

t-abe:
#315

I updated @tnarihi's deconvolution implementation to make it working on current chainer, and added tests.

PR #645, opened 2015-11-19
Another PR for #591: Add Deconvolution2D

wkentaro:
I added some commits from #591 according to the comments from @delta2323.


wkentaro:
The hard thing is to adapt this function to the reorganized function style..

wkentaro:
@delta2323 do you have any idea?

delta2323:
I sent you the PR that make `Deconvolution2D` function support the latest master branch. Could you check it?

wkentaro:
@delta2323 Ok, I will.

wkentaro:
A problem I'm concerning about is the return value `get_deconv_outsize` [here](https://github.com/wkentaro/chainer/blob/add_deconv2d/chainer/utils/conv.py#L14) can differ from the input of `get_conv_outsize`.
Because, `get_conv_outsize` ignores decimal values at [here](https://github.com/wkentaro/chainer/blob/add_deconv2d/chainer/utils/conv.py#L11) by `// s`.
So the usecase may be limited because of this.

okuta:
Would you like to be able to use the output size option?

wkentaro:
I agree with your idea, that kind of option is necessary in most cases if `s` is not `1`.
I'm adding that option.

wkentaro:
The original test was not actually tested deconvolution_2d but convolution_2d, so fixed.
And I added `outsize` as optional argument.

delta2323:
Just to be sure. I guess `outsize` is expected to be the value that satisfies `get_conv_outsize(outsize[0]) == in_h` and `get_conv_outsize(outsize[1]) == in_w` where `in_h` and `in_w` is the input height and width of deconvolution layer, respectively. The reason I think so is that the output size of deconvolution layer is expected to be same as input size of convolution layer when they are used as the building blocks of "convolutional autoencoder".
But the implementation allows the case when the equalities do not hold, too. Is this behavior agree with your intention?

wkentaro:
Good catch, thanks. I think it should be asserted.

wkentaro:
I think I have done, please review.

okuta:
Thank you very much!
LGTM.

PR #664, opened 2015-11-23
Warning on no cuDNN environment

okuta:


unnonouno:
How about showing a method to enable cudnn? I mean, 'Please reinstall chainer after you install cudnn.'

okuta:
I fixed the message.

unnonouno:
LGTM

PR #689, opened 2015-11-26
Add ctc padding

jnishi:
Support padding for training CTC with mini batch. Using this, you can train CTC by training data that are different length.

If valid length of each inputs of batch are given, values of input array after the given length are ignored.


jnishi:
I support padding and its computation is all on GPU. Please check.

jnishi:
Add test case for the case which all of inputs are padded.

okuta:
LGTM!

PR #754, opened 2015-12-12
Implement ELU

### LINKED ISSUE #694: ELU activation function
Djork-Arné Clevert, Thomas Unterthiner and Sepp Hochreiter propose a new activation function which seems to work well:

Fast and Accurate Deep Network Learning by Exponential Linear Units (ELUs)
http://arxiv.org/abs/1511.07289

The ELU (eq. 16) and its gradient (eq. 17) should be very easy to implement.

muupan:
This PR fixes #694.

### DIFF
--- a/chainer/functions/activation/elu.py
+++ b/chainer/functions/activation/elu.py
@@ -1,0 +1,72 @@
+import numpy
+
+from chainer import cuda
+from chainer import function
+from chainer.utils import type_check
+
+
+class ELU(function.Function):
+
+    """Exponential Linear Unit."""
+
+    def __init__(self, alpha=1.0):
+        self.alpha = numpy.float32(alpha)
+
+    def check_type_forward(self, in_types):
+        type_check.expect(in_types.size() == 1)
+        x_type, = in_types
+
+        type_check.expect(
+            x_type.dtype == numpy.float32,
+        )
+
+    def forward_cpu(self, x):
+        y = x[0].copy()
+        neg_indices = x[0] < 0
+        y[neg_indices] = self.alpha * (numpy.exp(y[neg_indices]) - 1)
+        return y,
+
+    def forward_gpu(self, x):
+        y = cuda.elementwise(
+            'T x, T alpha', 'T y',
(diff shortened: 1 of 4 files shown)


chainer/functions/activation/elu.py
```diff
@@ -0,0 +1,72 @@
+import numpy
+
+from chainer import cuda
+from chainer import function
+from chainer.utils import type_check
+
+
+class ELU(function.Function):
+
+    """Exponential Linear Unit."""
```
delta2323: Could you add the reference to ELU?

chainer/functions/activation/elu.py
```diff
@@ -0,0 +1,72 @@
...
+        gx = cuda.elementwise(
+            'T x, T gy, T alpha', 'T gx',
+            'gx = x >= 0 ? gy : gy * alpha * exp(x)', 'elu_bwd')(
+                x[0], gy[0], self.alpha)
+        return gx,
+
+
+def elu(x, alpha=1.0):
+    """Exponential Linear Unit function.
+
```
delta2323: Could you add the reference to ELU?

tests/chainer_tests/functions_tests/activation_tests/test_elu.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+class TestELU(unittest.TestCase):
+
+    def setUp(self):
+        # Avoid unstability of numeraical grad
+        self.x = numpy.random.uniform(-1, 1, (5, 4)).astype(numpy.float32)
+        for i in range(self.x.size):
+            if -0.01 < self.x.flat[i] < 0.01:
+                self.x.flat[i] = 0.5
```
delta2323: Unlike ReLU, ELU is differentiable at x = 0. So I do not think we need this avoid sampling around x = 0.

delta2323:
LGTM except the comment.

muupan:
ELU is not differentiable at x = 0 if alpha != 1. Should I add a test case where alpha is exactly 1?

delta2323:
I am sorry. I missed `alpha`. You do not have to treat 1 as a special case.

delta2323:
LGTM

PR #757, opened 2015-12-13
Add binary accuracy

t-abe:
This is the function to compute accuracy of binary classification. (typically used with sigmoid_cross_entropy)

### DIFF
--- a/chainer/functions/evaluation/binary_accuracy.py
+++ b/chainer/functions/evaluation/binary_accuracy.py
@@ -1,0 +1,44 @@
+import numpy
+
+from chainer import cuda
+from chainer import function
+from chainer.utils import type_check
+
+
+class BinaryAccuracy(function.Function):
+    def check_type_forward(self, in_types):
+        type_check.expect(in_types.size() == 2)
+        x_type, t_type = in_types
+
+        type_check.expect(
+            x_type.dtype == numpy.float32,
+            t_type.dtype == numpy.int32,
+            t_type.shape[0] == x_type.shape[0],
+        )
+
+    def forward(self, inputs):
+        xp = cuda.get_array_module(*inputs)
+        y, t = inputs
+        # flatten
+        y = y.reshape(len(y), -1)
+        t = t.reshape(len(t), -1)
+        c = (y >= 0)
+        return xp.asarray((c == t).mean(dtype='f')),
+
+
+def binary_accuracy(y, t):
+    """Computes binary classification accuracy of the minibatch.
+
+    Args:
+        y (Variable): Variable holding a matrix whose i-th element
+            indicates the score of positive at the i-th example.
+        t (Variable): Variable holding an int32 vector of groundtruth
+            labels (0 or 1).
+
+    Returns:
+        Variable: A variable holding a scalar array of the accuracy.
+
+    .. note:: This function is non-differentiable.
+
+    """
+    return BinaryAccuracy()(y, t)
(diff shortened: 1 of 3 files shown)


chainer/functions/evaluation/binary_accuracy.py
```diff
@@ -0,0 +1,44 @@
...
+        c = (y >= 0)
+        return xp.asarray((c == t).mean(dtype='f')),
+
+
+def binary_accuracy(y, t):
+    """Computes binary classification accuracy of the minibatch.
+
+    Args:
+        y (Variable): Variable holding a matrix whose i-th element
+            indicates the score of positive at the i-th example.
```
delta2323: `j-th` ?

chainer/functions/evaluation/binary_accuracy.py
```diff
@@ -0,0 +1,44 @@
...
+
+class BinaryAccuracy(function.Function):
+    def check_type_forward(self, in_types):
+        type_check.expect(in_types.size() == 2)
+        x_type, t_type = in_types
+
+        type_check.expect(
+            x_type.dtype == numpy.float32,
+            t_type.dtype == numpy.int32,
+            t_type.shape[0] == x_type.shape[0],
```
delta2323: As `x` and `t` are flatten below `x` and `t` can be more than one dimensional. So I think the requirement for the shapes is `prod(t_type.shape) == prod(x_type)` rather than the consistency of first dimensions.

delta2323: Further, I think we should impose more tighter restriction on the shapes of `x` and `t`. `t_type.shape == x_type.shape`, that is, to impose that `x` and `t` must have same shape. How do you think of it?

tests/chainer_tests/functions_tests/evaluation_tests/test_binary_accuracy.py
```diff
@@ -0,0 +1,46 @@
...
+from chainer import testing
+from chainer.testing import attr
+from chainer.testing import condition
+
+
+class TestAccuracy(unittest.TestCase):
+
+    def setUp(self):
+        self.x = numpy.random.uniform(-1, 1, (30,)).astype(numpy.float32)
+        self.t = numpy.random.randint(3, size=(30,)).astype(numpy.int32)
```
delta2323: Please add test case where `x` and `t` has more than one dimensions.

delta2323: Please add test case where the shapes of `x` and `t` are contradicted.

t-abe:
I updated the PR based on the review comments.

delta2323:
Could add test case where function correctly raise `InvalidType` error when the shapes of x and t are not same.

chainer/functions/evaluation/binary_accuracy.py
```diff
@@ -0,0 +1,44 @@
...
+            x_type.dtype == numpy.float32,
+            t_type.dtype == numpy.int32,
+            t_type.shape == x_type.shape,
+        )
+
+    def forward(self, inputs):
+        xp = cuda.get_array_module(*inputs)
+        y, t = inputs
+        # flatten
+        y = y.reshape(-1)
```
okuta: How about use `ravel` method?

chainer/functions/evaluation/binary_accuracy.py
```diff
@@ -0,0 +1,44 @@
...
+        t = t.reshape(-1)
```
okuta: How about use `ravel` method?

t-abe:
updated

delta2323:
LGTM :+1:

PR #758, opened 2015-12-13
Fix handling of temporary files

okuta:


unnonouno:
`NamedTempfile` is removed when it is close, isn't it? 
https://docs.python.org/2/library/tempfile.html#tempfile.NamedTemporaryFile

unnonouno:
OK, `delete=True` is not supported on Windows. Please write a comment about it.

unnonouno:
LGTM

PR #785, opened 2015-12-19
Add destroy pool.

jnory:
Hi,

I recognize that current `MemoryPool` implementation does not have features to free GPU memory, which is possibly a problem for server side applications.

This small PR gives a feature to release GPU memory forcibly.
We can now `chainer.cuda.pool.destroy()` after each jobs.


okuta:
Please write test.

okuta:
Hou about use `free_all_free`, and change method name `destroy` to `free_all_free`?

jnory:
Thank you for your comments. I will try them ASAP.

jnory:
@okuta I've added tests but some tests other than my changes (e.g. test_reversed_inner (test_product.TestProduct)) are now failing. I'm under investigation.

By the way, I have a question: Why `TestMemoryPool.pool` is instance of `SingleDeviceMemoryPool`? https://github.com/pfnet/chainer/blob/ad7a68fe5c549b477ef13003c836ae782e253472/tests/cupy_tests/cuda_tests/test_memory.py#L143

jnory:
Oops, I downgrade numpy to 1.9.3, all tests has passed.

okuta:
https://github.com/pfnet/chainer/blob/ad7a68fe5c549b477ef13003c836ae782e253472/tests/cupy_tests/cuda_tests/test_memory.py#L143

jnory:
I fixed it together with original PR.

cupy/cuda/memory.pyx
```diff
@@ -341,6 +341,9 @@ cdef class SingleDeviceMemoryPool:
     cpdef free_all_free(self):
         self._free = collections.defaultdict(list)
 
+    cpdef n_free(self):
+        return sum([len(free) for _, free in self._free.iteritems()])
```
okuta: How about above code?
```python
Py_ssize_t sum
for v in self._free.itervalues():
    sum += len(v)
return sum
```

cupy/cuda/memory.pyx
```diff
@@ -391,3 +394,21 @@ cdef class MemoryPool(object):
...
         return self._pools[dev].malloc(size)
+
+    cpdef free_all_free(self):
+        """release unused memory pool.
+        """
+        dev = device.get_device_id()
+        if dev in self._pools:
+            self._pools[dev].free_all_free()
+
+    cpdef n_free(self):
```
okuta: I think more meaningful name is good. For example, `n_free_blocks`.

jnory:
Sorry for the late reply. I will fix them today.

cupy/cuda/memory.pyx
jnory: Thank you! I've fixed here.

cupy/cuda/memory.pyx
jnory: Ok, I've changed the name from `n_free` to `n_free_blocks`.

jnory:
The code is now updated. Thank @okuta for the review.

cupy/cuda/memory.pyx
```diff
@@ -341,6 +341,12 @@ cdef class SingleDeviceMemoryPool:
     cpdef free_all_free(self):
         self._free = collections.defaultdict(list)
 
+    cpdef n_free_blocks(self):
+        cdef Py_ssize_t n = 0
+        for v in self._free.itervalues():
```
okuta: Sorry, I made a mistake.
Please use `six.itervalues`, because `itervalues` was removed in Python3.

okuta:
LGTM except comment!

cupy/cuda/memory.pyx
jnory: Oops, thank you. I've just fixed here now.

cupy/cuda/memory.pyx
```diff
@@ -391,3 +398,21 @@ cdef class MemoryPool(object):
         """
         dev = device.get_device_id()
         return self._pools[dev].malloc(size)
+
+    cpdef free_all_free(self):
+        """release unused memory pool.
```
okuta: How about `"""Release free blocks in memory pool.""""`?

cupy/cuda/memory.pyx
```diff
@@ -391,3 +398,21 @@ cdef class MemoryPool(object):
...
+        """
+        dev = device.get_device_id()
+        if dev in self._pools:
+            self._pools[dev].free_all_free()
+
+    cpdef n_free_blocks(self):
+        """
+        :return:
```
okuta: Please fix style of document.

cupy/cuda/memory.pyx
```diff
@@ -391,3 +398,21 @@ cdef class MemoryPool(object):
...
+        the total number of unused memory pool.
```
okuta: How about `int: The number of free blocks.`?

cupy/cuda/memory.pyx
```diff
@@ -391,3 +398,21 @@ cdef class MemoryPool(object):
...
+
+    cpdef free_all_free(self):
+        """release unused memory pool.
+        """
+        dev = device.get_device_id()
+        if dev in self._pools:
+            self._pools[dev].free_all_free()
+
+    cpdef n_free_blocks(self):
+        """
```
okuta: Please write one-line document.

tests/cupy_tests/cuda_tests/test_memory.py
```diff
@@ -140,14 +140,30 @@ def test_free_all_free(self):
...
 
     def setUp(self):
-        self.pool = memory.SingleDeviceMemoryPool()
+        self.pool = memory.MemoryPool()
 
     def test_zero_size_alloc(self):
-        mem = self.pool.malloc(0).mem
-        self.assertIsInstance(mem, memory.Memory)
-        self.assertNotIsInstance(mem, memory.PooledMemory)
+        from cupy.cuda import Device
```
okuta: Please import at the head of source.
And, don't import objects, only modules.

We use PEP8 and a part of OpenStack Style Guidelines related to general coding style as our basic style guidelines.
see http://docs.openstack.org/developer/hacking/#imports

cupy/cuda/memory.pyx
```diff
@@ -6,6 +6,7 @@ from cupy.cuda import runtime
 
 from cupy.cuda cimport device
 from cupy.cuda cimport runtime
+import six
```
okuta: Please move import to L4(and create new import group), because six is a third party library.
see https://www.python.org/dev/peps/pep-0008/#imports

jnory:
Thank you for the comments. I will fix them today.

jnory:
The codes are now updated again.

chainer/cuda.py
```diff
@@ -123,7 +123,8 @@ def __ne__(self, other):
 # Global states
 # ------------------------------------------------------------------------------
 if available:
-    cuda.set_allocator(cuda.MemoryPool().malloc)
+    pool = cuda.MemoryPool()
```
okuta: How about using more specific name (e.g. memory_pool)?

cupy/cuda/memory.pyx
```diff
@@ -391,3 +398,21 @@ cdef class MemoryPool(object):
...
+        """Release free blocks."""
+        dev = device.get_device_id()
+        if dev in self._pools:
+            self._pools[dev].free_all_free()
+
+    cpdef n_free_blocks(self):
+        """Count the total number of free blocks.
+
+        Returns:
+        int: The total number of free blocks.
```
okuta: Need indent.

cupy/cuda/memory.pyx
```diff
@@ -391,3 +398,21 @@ cdef class MemoryPool(object):
         """
         dev = device.get_device_id()
         return self._pools[dev].malloc(size)
+
+    cpdef free_all_free(self):
+        """Release free blocks."""
+        dev = device.get_device_id()
+        if dev in self._pools:
```
okuta: The type of `self._pools` is defaultdict.
Is this check is necessary?

cupy/cuda/memory.pyx
```diff
@@ -391,3 +398,21 @@ cdef class MemoryPool(object):
...
+            self._pools[dev].free_all_free()
+
+    cpdef n_free_blocks(self):
+        """Count the total number of free blocks.
+
+        Returns:
+        int: The total number of free blocks.
+        """
+        dev = device.get_device_id()
+        if dev in self._pools:
```
okuta: The type of `self._pools` is defaultdict.
Is this check is necessary?

cupy/cuda/memory.pyx
```diff
@@ -1,6 +1,7 @@
 import collections
 import ctypes
 import weakref
+import six
```
okuta: Please add a blank line before this line.
`coolections`, `ctypes` and `weakref` is standard library.
`six` is related third party library.

jnory:
Thank you for the comments. I've just updated codes.
In order to confirm that  we can call `free_all_free()` or `n_free_blocks()` without calling `malloc()`, I've added more tests.

okuta:
Thank you for many fixing!
Added test is good.

okuta:
LGTM!

jnory:
Thank you for many reviews!

PR #790, opened 2015-12-22
Add ignore_label into sigmoid_cross_entropy

t-abe:
I added ``ignore_label`` to ``sigmoid_cross_entropy`` as ``softmax_cross_entropy`` has.
When groundtruth label is -``1``, corresponding prediction is ignored while computing loss.

Added ``normalize`` option works same as ``softmax_cross_entropy``.

### DIFF
--- a/chainer/functions/loss/sigmoid_cross_entropy.py
+++ b/chainer/functions/loss/sigmoid_cross_entropy.py
@@ -10,8 +10,11 @@ class SigmoidCrossEntropy(function.Function):
 
     """Sigmoid activation followed by a sigmoid cross entropy loss."""
 
-    def __init__(self, use_cudnn=True):
+    ignore_label = -1
+
+    def __init__(self, use_cudnn=True, normalize=True):
         self.use_cudnn = use_cudnn
+        self.normalize = normalize
 
     def check_type_forward(self, in_types):
         type_check.expect(in_types.size() == 2)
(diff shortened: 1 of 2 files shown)


chainer/functions/loss/sigmoid_cross_entropy.py
```diff
@@ -23,48 +26,44 @@ def check_type_forward(self, in_types):
...
 
     Args:
         x (Variable): A variable object holding a matrix whose (i, j)-th
             element indicates the unnormalized log probability of the j-th unit
             at the i-th example.
-        t (Variable): A variable object holding an int32 vector of groundtruth
-            binary labels.
+        t (Variable): Variable holding an int32 vector of groundtruth labels.
+            If ``t[i] == -1``, correspondig ``x[i]`` is ignored.
+        normalize (Variable): Variable holding a boolean value which
```
delta2323: `normalize` is not an instance of `Variable` but a boolean.

delta2323: I think the name `normalize` is not appropriate because it is natural to expect this option to enable/disable normalization. But what this option actually does it to determine the normalization constant, meaning that output is normalized regardless of the value of this option. How about changing the name that emphasizes the function of this option?

chainer/functions/loss/sigmoid_cross_entropy.py
```diff
@@ -23,48 +26,44 @@ def check_type_forward(self, in_types):
...
-        return gx, None
-
-    def backward_gpu(self, inputs, grad_outputs):
-        t, gloss = inputs[1], grad_outputs[0]
-        gx = cuda.elementwise(
-            'T y, S t, raw T gloss, T inv_cnt', 'T gx',
-            'gx = gloss[0] * inv_cnt * (y - t)',
-            'sigmoid_crossent_bwd')(self.y, t, gloss, 1.0 / t.shape[0])
+        gloss = grad_outputs[0]
+        y, = sigmoid.Sigmoid().forward((x,))
```
delta2323: Pass `self.use_cudnn` to `Sigmoid`

chainer/functions/loss/sigmoid_cross_entropy.py
```diff
@@ -23,48 +26,44 @@ def check_type_forward(self, in_types):
...
+            count = x.shape[0]
+        self.count = count
         # stable computation of the cross entropy.
-        loss = -numpy.sum(
-            x * (t - (x >= 0)) - numpy.log1p(numpy.exp(-numpy.abs(x))))
-        return numpy.array(loss / t.shape[0], dtype=x.dtype),
+        loss = -xp.sum(
+            self.ignore_mask * (x * (t - (x >= 0)) -
+                                xp.log1p(xp.exp(-xp.abs(x)))))
+        return xp.array(loss / self.count, dtype=x.dtype),
```
delta2323: `self.count` is 0 when all training data is ignored, which causes 0 division error. We must handle such a case.

delta2323: Please add test cases where all training data are ignored.

t-abe:
I updated it based on your comments except `normalize`.

> I think the name `normalize` is not appropriate because it is not strange to expect this option to enable/disable normalization. But what this option actually does it to determine the normalization constant, meaning that output is normalized regardless of the value of this option. How about changing the name to the one that emphasizes the functionality?
I agree with your comment, and `normalize_by_size` is better as the name of this. (there may be better one)
But `normalize` is also used in `softmax_cross_entropy`, so changing it causes a small consistency problem.
How should I do?

delta2323:
This change adds the specification that loss is 0 when all data are ignored. So could you add this remark to the document (e.g. in the description of `t` in `sigmoid_cross_entropy`)?

delta2323:
I agree with you as for `normalize` option. I come to think change of the option name is not so important enough to break the inconsistency between `sigmoid_cross_entropy` and `softmax_cross_entropy`.

chainer/functions/loss/sigmoid_cross_entropy.py
```diff
@@ -23,48 +26,45 @@ def check_type_forward(self, in_types):
...
 
     Args:
         x (Variable): A variable object holding a matrix whose (i, j)-th
             element indicates the unnormalized log probability of the j-th unit
             at the i-th example.
-        t (Variable): A variable object holding an int32 vector of groundtruth
-            binary labels.
+        t (Variable): Variable holding an int32 vector of groundtruth labels.
+            If ``t[i] == -1``, correspondig ``x[i]`` is ignored.
+            Loss becomes to zero if all groundtruth labels are ``-1``.
```
delta2323: I guess "Loss is zero" is more grammatically appropriate.

delta2323:
LGTM :+1:

PR #791, opened 2015-12-22
Improve _sqnorm

okuta:


chainer/optimizer.py
```diff
@@ -7,11 +7,21 @@
...
-    with cuda.get_device(x):
-        x = x.ravel()
-        return float(x.dot(x))
+def _sum_sqnorm(arr):
+    sum = {}
+    for x in arr:
+        with cuda.get_device(x) as dev:
+            x = x.ravel()
+            s = x.dot(x)
+        id = int(dev)
```
beam2d: It hides the builtin function `id`. Use a different name (e.g. `dev_id`).

chainer/optimizer.py
```diff
@@ -7,11 +7,21 @@
 import chainer.link as link_module
 
 
-def _sqnorm(x):
-    # TODO(delta2323): Make it public function and move it to common directory.
-    with cuda.get_device(x):
-        x = x.ravel()
-        return float(x.dot(x))
+def _sum_sqnorm(arr):
+    sum = {}
```
beam2d: It hides the builtin function `sum`. Use a different name (e.g. `sq_sum`).

chainer/optimizer.py
```diff
@@ -7,11 +7,21 @@
...
+    for x in arr:
+        with cuda.get_device(x) as dev:
+            x = x.ravel()
+            s = x.dot(x)
+        id = int(dev)
+        if id in sum:
```
beam2d: How about using defaultdict?

beam2d:
Would you add multi-gpu tests (both with and without cpu)?

okuta:
@beam2d I fixed the code.

beam2d:
LGTM!

PR #810, opened 2015-12-28
Implement numpy serializer

okuta:


chainer/serializers/npz.py
```diff
@@ -0,0 +1,96 @@
...
+    This is the standard serializer in Chainer. The chain hierarchy is simply
+    mapped to dictionary key.
+
+    Args:
+        dict (dict): The dict that this serializer represents.
+        path (str): The base path.
+
+    """
+    def __init__(self, dict, path=''):
+        self.dict = dict
```
unnonouno: Don't use `dict` as it is reserved.

chainer/serializers/npz.py
```diff
@@ -0,0 +1,96 @@
...
+        self.path = path
+
+    def __getitem__(self, key):
+        return DictionarySerializer(self.dict, self.path + key + '/')
```
unnonouno: Do we need to escape the key?

chainer/serializers/npz.py
```diff
@@ -0,0 +1,96 @@
...
+
+    def __call__(self, key, value):
+        ret = value
+        if isinstance(value, cuda.ndarray):
```
unnonouno: I think we don't need this check.

beam2d: This check is required to accept Python scalar values (to_cpu only accepts numpy/cupy arrays).

beam2d:
Fixed. I once think about escaping slashes in the keys, but it made the serialization of optimizers full of escapes, so I do not include the escaping in the current version.

unnonouno:
OK. It looks good. Please fix the document.

beam2d:
Added a document. I also added a compression support.

beam2d:
Added tests without compression.

unnonouno:
LGTM

PR #813, opened 2015-12-30
Fix broken math expressions in reference

ronekko:
In the document of [`F.linear`](http://docs.chainer.org/en/stable/reference/functions.html#chainer.functions.linear) and [`F.embed_id`](http://docs.chainer.org/en/stable/reference/functions.html#chainer.functions.embed_id), math expression `\times` and `\top` are not properly rendered. To fix them, I replaced `\t` with `\\t` at the first commit.

In principle, as [Math support in Sphinx](http://sphinx-doc.org/ext/math.html) says:
> Keep in mind that when you put math markup in Python docstrings read by autodoc, you either have to double all backslashes, or use Python raw strings (r"raw").

So I further replaced single-backslash in doncstrings with double-backslashes.

### DIFF
--- a/chainer/functions/activation/clipped_relu.py
+++ b/chainer/functions/activation/clipped_relu.py
@@ -9,7 +9,8 @@ class ClippedReLU(function.Function):
 
     """Clipped Rectifier Unit function.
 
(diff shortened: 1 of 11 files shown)


okuta:
Thank you very much!
LGTM.

PR #817, opened 2016-01-02
Make GRU stateful

### LINKED ISSUE #789: Same api for GRU and LSTM
Currently, GRU is stateless while LSTM is stateful. It would be great to make GRU stateful as well, for the stake of consistency.

delta2323:
Fix #789 

I propose to add several APIs to `link.GRU`
* `reset_state`: This same API as `link.LSTM`
* `set_state`: The pair instance of `reset_state` (it is not existed in current `link.LSTM`)
I think we should add `set_state` to `link.LSTM`, too.

Also, we need to override `to_cpu` and `to_gpu` because when the link is moved to CPU (resp. GPU), we must move its state, too. Maybe we must override these methods of `link.LSTM`.

### DIFF
--- a/chainer/links/connection/gru.py
+++ b/chainer/links/connection/gru.py
@@ -120,11 +120,13 @@ class StatefulGRU(GRUBase):
             self.h.to_gpu(device)
 
     def set_state(self, h):
+        assert isinstance(h, chainer.Variable)
+        h_ = h
         if self.xp == numpy:
-            h = chainer.cuda.to_cpu(h)
+            h_.to_cpu()
         else:
-            h = chainer.cuda.to_gpu(h)
-        self.h = chainer.Variable(h, volatile='auto')
+            h_.to_gpu()
+        self.h = h_
 
     def reset_state(self):
         self.h = None
--- a/chainer/links/connection/gru.py
+++ b/chainer/links/connection/gru.py
@@ -98,9 +98,8 @@ class StatefulGRU(GRUBase):
     Use :class:`~chainer.links.GRU` as a stateless version of GRU.
 
     Args:
-        n_units(int): Dimension of hidden vector :math:`h`.
-        n_inputs(int): Dimension of input vector :math:`x`. If ``None``,
-        it is set to the same value as ``n_units``.
+        in_size(int): Dimension of input vector :math:`x`.
+        out_size(int): Dimension of hidden vector :math:`h`.
 
     Attributes:
         h(~chainer.Variable): Hidden vector that indicates the state of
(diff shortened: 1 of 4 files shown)


beam2d:
To keep the compatibility, I recommend you to make it another class rather than modifying the existing GRU class.

delta2323:
@beam2d  Thank you for suggestion, I agree with you.

chainer/links/connection/gru.py
```diff
@@ -56,3 +68,74 @@ def __call__(self, h, x):
...
+            h = chainer.cuda.to_cpu(h)
+        else:
+            h = chainer.cuda.to_gpu(h)
+        self.h = chainer.Variable(h)
+
+    def reset_state(self):
+        self.h = None
+
+    def __call__(self, x):
+        if self.h is None:
```
beam2d: If self.h is None, you can omit the lateral computations (`U_r(h)`, `U_z(h)`, `U(r*h)`, and `(1-z)*h`) instead of initializing it by zeros. It would slightly improve the first iteration on a sequence.

beam2d:
LGTM except one comment.

chainer/links/connection/gru.py
```diff
@@ -56,3 +68,83 @@ def __call__(self, h, x):
...
+        z = self.W_z(x)
+        if self.h is not None:
+            r += self.U_r(self.h)
+            z += self.U_z(self.h)
+        r = sigmoid.sigmoid(r)
+        z = sigmoid.sigmoid(z)
+
+        h_bar = self.W(x)
+        if self.h is not None:
+            h_bar += self.U(r * self.h)
```
beam2d: I further found that the computation of `r` can also be omitted when self.h is None.

chainer/links/connection/gru.py
```diff
@@ -56,3 +68,81 @@ def __call__(self, h, x):
...
+        super(StatefulGRU, self).to_gpu(device)
+        if self.h is not None:
+            self.h.to_gpu(device)
+
+    def set_state(self, h):
+        if self.xp == numpy:
+            h = chainer.cuda.to_cpu(h)
+        else:
+            h = chainer.cuda.to_gpu(h)
+        self.h = chainer.Variable(h)
```
beam2d: Add volatile='auto'.

chainer/links/connection/gru.py
beam2d: Can you combine these two if statements?

beam2d:
I found that set_state should accept a variable instead of an array to make it able to use the output of another network as the initial state.

delta2323:
I add `n_inputs` to `GRU` and `StatefulGRU` to accept input data whose size can be different from `n_units` other than as suggested in #837.

beam2d:
It would be confusing that the order of the arguments are reversed between (Stateful)GRU and LSTM, though we do not want to break the interface of the original GRU... How about making the interface of StatefulGRU matched to that of LSTM, while leaving that of GRU as is (at least the position of n_units)?

beam2d:
LGTM! (sorry for being late)

Issue #118, opened 2015-07-02
Flexible weight initialization of Parametrized Functions


delta2323:
I would like to change the way `chainer.functions.Linear` is initialized, especially weight initialization. In current implementation, initialization by Gaussian is hard-coded in `Linear.__init__`.

This problem is generalized how to initialize parametrized functions.

EdwardRaff:
Is there anyone working on this / plans for this? This is particularly important for LSTMs, where the forget bias terms are usually initialized to a value >= 1, and all other bias terms initialized to 0. Its also common to initialize the internal activation in a manner different from the initialization of the input weights. 

There are some compatibility licensed initialization schemes in the keras library that could be used to significantly reduce the effort in this.

delta2323:
Thank you for suggestion on the initialization on LSTM. For now, we do not have any concrete plan to implement this feature. I think we need to design the interface first. Thank you.

EdwardRaff:
If the Keras style of doing it works for you (constructor arguments with default values) I'd be willing to work on merging in the Keras code and adding appropriate initializers.

delta2323:
Although it is vague for now, we also have requirements on initialization scheme. So how about making a WIP (work in progress) PR and discussing the design? Since changes will be huge, I would recommend implement step by step (e.g. first apply change to `Linear Link`).

EdwardRaff:
Sure, I'll make a pull soon.

PR #820, opened 2016-01-03
[WIP] Flexible weight initialization of Parametrized Functions

EdwardRaff:
Work space to tackle Issue #118 

As requested, initial commit with only changes to Linear link. Documentation isn't correct any more, I made the initial pull to just show how I would plan on adding everything and make it backwards compatible with existing behavior/code. 

The initialization used for Linear is very close to `he_normal`. If minor change in default behavior is acceptable, that could be used instead of the hardcoded stuff (ie: if `is None`, set to `he_normal`).


delta2323:
Thank you very much! Your work is very helpful.

If you mind the failure of travis CI, you should apply flake8, although we can do it afterward.
http://docs.chainer.org/en/stable/contribution.html#coding-guidelines

EdwardRaff:
At this point I'm not so concerned with travis right now on style errors, I'll take care of that in the end :) 

I'm more concerned with the setup and the way I've introduced the functionality. If both look good I can start adding it to other links, if not - any recomendations on what to change / make it better would be appreciated. Or if there is some potential issue I'm not aware of.

beam2d:
Thank you for the work! How about passing callable objects that only accept the shape argument, and pushing other arguments (e.g. scale) to the construction of these callable objects? It may make the interface simpler and more flexible. Adding scaling parameter to `he_normal` etc. is also good (it makes it easy to implement the current default behavior as an initializer, like `HeNormal(scale=0.5)`).

I'd also like to add `__repr__()` to the callable object, though it can be deferred to another PR if you do not have enough time to do it.

EdwardRaff:
>How about passing callable objects that only accept the shape argument, 

I can do that, but is it necessary / makes sense? I don't see any reason for the initializer to be an actual object. Unless you want to be. 

>and pushing other arguments (e.g. scale) to the construction of these callable objects? It may make the interface simpler and more flexible.

I could do that, but that would get a little weird for initializing the bias a well. Unless you can do more than one **kwargs in python (I don't write a huge amount of python, if thats doable than no problem). 

Would you want me to remove the current arguments ie: its ok to break some current code? Or leave them int? 

>Adding scaling parameter to he_normal etc. is also good (it makes it easy to implement the current default behavior as an initializer, like HeNormal(scale=0.5))

A scale parameter make sense, but scaling he_normal by 0.5 will not produce the current initialization used by Chainer. That would get you `1/2 sqrt(2/fan_in)` 

>I'd also like to add __repr__() to the callable object, though it can be deferred to another PR if you do not have enough time to do it.

Currently I'm just using a bunch of functions, I'm not aware of how to define `__repr__` for a function in python.

EdwardRaff:
Ok, so you can't do multiple **kwargs for a function, which is problematic for having options for both the weight vector and the bias term initialized with different options. Its also problematic for any model where you would want to initialize multiple parts differently (this is true for LSTMs). 

One option would be to define the initialization functions like this:

```python

def __init__(self, in_size, out_size, wscale=1, bias=0, nobias=False,
                   initialW=I.glorot_uniform, initialW_args={}, initial_bias=I.zeros, initial_bias_args={}):
    #...
    elif isinstance(initialW, types.FunctionType):
        initialW = initialW(shape, **initialW_args)
    #...
    elif isinstance(initial_bias, types.FunctionType):
        initial_bias = initial_bias(shape, **initialW_args)
```

This will allow for specifying function specific parameters for each initialization function. 

Does that look good to you as a way to code it up? 


>>and pushing other arguments (e.g. scale) to the construction of these callable objects? It may make the interface simpler and more flexible.
>I could do that, but that would get a little weird for initializing the bias a well. Unless you can do more than one **kwargs in python (I don't write a huge amount of python, if thats doable than no problem).

I didn't understand when I first read that. Now I understand why you thought of callables. I think it works for the caller, but makes the code in `initializations.py` boiler-plate heavy to make each one an object. I don't think it would be a big need, so my personal preference would be for the idea I proposed above. 

Obviously I'll defer to yalls judgment - let me know.

delta2323:
> I think it works for the caller, but makes the code in initializations.py boiler-plate heavy to make each one an object.
Then, how about put `(Abstract)Initializer` class to `initialization.py` and each concrete initializer class to `initializers/` directory? This is same idea as how `Function`, `Link`, `Optimizer` and `Serializer` and their subclasses are organized.

beam2d:
> One option would be to define the initialization functions like this:
> ...
> This will allow for specifying function specific parameters for each initialization function.

I do not think it is good to push the arguments of each initializer into arguments of chains. I want to keep the constructor of chains simpler, since new chain implementations are written much more often than initializers.

You can use closures to keep the code simple yet hiding the options from chains. E.g.:
```
def normal(scale=0.05):
    return lambda shape: np.random.normal(loc=0.0, scale=scale, size=shape)
```
Letting it have a customized repr requires writing a class, though.

EdwardRaff:
@beam2d using closures also makes sense to me. The default __repr__ for a function will include the function name, and people can then look up the settings in the docs. 

If you're OK with just the functions defined as they are, I can switch the Linear to `callable` like you wanted and start adding it to others. 

The only other question then is if you want me to remove things like `wscale` from the Chain options and if you want the default initializations to behave like they did before, or if new initializations are ok.

beam2d:
`wscale` option should be left as is, and we should keep the default initialization behavior for backward compatibility. I think it is better to write a new initializer that does the default initialization. It can be used both for Linear and Convolution2D.

EdwardRaff:
@beam2d updated the code for Linear. Now uses callable as the designator, and made the initial_bias have the same type of behavior. Also uses a scaling of `1/sqrt(2)` to make it compatible with the original initialization scheme after my brain started working again.

delta2323:
I think it is better to make `bias` argument in `Linear` and `Convolution2D` deprecated because its role has some overlap with `initial_bias`. See #525 and #532 for detail discussion.

EdwardRaff:
@delta2323 So do you want me to make those changes in this PR, or just do it after? 

Once I get an "ok" on the style I'll add the initialization style I've implemented in Linear to all the links.

delta2323:
@EdwardRaff I come to think we do not have to apply change about `bias` in this PR because `bias` problem is a different one from initializers. So you do not have to mind it. I am sorry that my comment bothered you.

To avoid conflict, we should work on `bias` affairs after finishing implementation of initializers (this was principal reason I mentioned `bias` here).

EdwardRaff:
I wasn't bothered by your comment. Though I would appreciate an answer if the approach is "acceptable". I've gone ahead and implemented the initialization for more method, and refactored part of the code into a helper method. This cleans up a good amount of code. 

I've gone ahead and done the LSTM with initializations, and made it break previous behavior. Specifically, initializing the forget gates is important to performance (see: Jozefowicz, R., Zaremba, W., & Sutskever, I. (2015). An Empirical Exploration of Recurrent Network Architectures. Proceedings of The 32nd International Conference on Machine Learning, 37, 2342–2350. Retrieved from [here](http://machinelearning.wustl.edu/mlpapers/paper_files/icml2015_jozefowicz15.pdf) ). Since I was making changes, I also made the lateral connections use the orthogonal initialization (see: Saxe, A. M., McClelland, J. L., & Ganguli, S. (2013). Exact solutions to the nonlinear dynamics of learning in deep linear neural networks. arXiv Preprint arXiv:1312.6120. Retrieved from [here](http://arxiv.org/pdf/1312.6120v3.pdf) )

For links like inception, that have other links inside of them - is there a preference on initialization? My initial thought would be to add a single `initial_x` argument to `__init__`. If `initial_x` is a list, it needs to be the same size as the number of layers with type x, and they will be applied individually. If it is not, the same value will be used for all of them. 

ie: [func1, func2, func3] would be used to initialize conv1, conv2, and conv3 in a layer. If only func1 was given, we turn it into [func1, func1, func1] and again apply to conv1, conv2, and conv3.

EdwardRaff:
I went with the simpler approach for inception layer, one initialization scheme used for all components. I also ran pyformat on all the code I added. @beam2d  @delta2323 Please tell me what you think.

delta2323:
I think dictionary is more preferable than list for passing initializers to one Chain because there is no a priori ordering to Links that the Chain has (ChainList has such an ordering). So one of the possible way is to pass dictionary that maps the names of Links to the Initializers.

But as for inception, I agree with your idea that we use single initializer for initializing `Convolution2D` Links in it. Further, it will be difficult to adopt single strategy on how to pass initializers to all kinds of Chains. We must think differently how to pass initializers according to the type of Chains. So we should treat my proposal about the dictionary of Initializer as one of the common strategies of initialization.

Generally speaking

EdwardRaff:
I'm not sure where that dictionary initialization scheme would be needed. My original thought was for Inception, but just using one initializer seemed better. Could you clarify where you would want to see that added? 

Also, not sure why the checks/travis have stopped running on this pull.

delta2323:
For existing Chains, I do not have any one that I want to use dictionary for initialization. I was thinking Chains that will be implemented in the future.

Any Chain can have the "initialization-by-dictionary" scheme if they have multiple Links or Chains in it (, especially when they have different roles). One of the such examples will be VAE. I suppose VAEs are implemented as a Chain which includes two Chains, encoder and decoder as children. In that case, some user may want to initialize encoder and decoder in different ways.

delta2323:
Sorry, I do not come up with why travis does not work. We will investigate it.

EdwardRaff:
It looks like the `add_link` method in links.py requires unique names withing children. One could write code to fully specify the path to a link, something like `def initLink(self, path_representation, field_name, initializer):`. Do you want me to work on this? I personally don't see any particular need for it at this time.

delta2323:
Sorry for late response. I agree with you. I do not think we need this functionality for now (or the priority is not so high at least).

EdwardRaff:
Ok then. What else do I need to do / clean up for this to be acceptable for merge? 

Also, installing flake8 and running it doesn't give the same errors as your CI builds, which is very confusing.

EdwardRaff:
I have to re-merge master, but I wanted to check in on what more I need to do to make this mergeable into master.

delta2323:
Thank you for updating your implementation and sorry for my late response. As I agree with overall design of initialization scheme, I start to review this PR as an ordinal one.

You can merge the master any time you want to, but as merge can be costly, I recommend to merge after you apply my comments, which will be available soon.

chainer/__init__.py
```diff
@@ -24,6 +25,7 @@
 Optimizer = optimizer.Optimizer
 Serializer = serializer.Serializer
 Variable = variable.Variable
+Initializations = initializations
```
delta2323: What is the reason for this rename?

delta2323:
If we apply same directory structures as `functions`, `links`, `optimizers`, and `serializers`, you should create `initializers` directory, make files for each initializers and put alias for them in `__init__`. 

I think directory structures for optimizers are helpful (e.g. [Here](https://github.com/pfnet/chainer/blob/master/chainer/optimizers/__init__.py)).

chainer/initializations.py
```diff
@@ -0,0 +1,120 @@
+from __future__ import absolute_import
+
+import numpy as np
+
+# Original code forked from MIT licensed keras project
+# https://github.com/fchollet/keras/blob/master/keras/initializations.py
+
+
+def init_weight(weights, initWith, scale=1,
```
delta2323: We ordinary use snake case for argument names. So replace `initWith` with `init_with`

chainer/initializations.py
```diff
@@ -0,0 +1,120 @@
+from __future__ import absolute_import
```
delta2323: Why this import is needed?

chainer/initializations.py
```diff
@@ -0,0 +1,120 @@
...
+
+import numpy as np
+
+# Original code forked from MIT licensed keras project
+# https://github.com/fchollet/keras/blob/master/keras/initializations.py
+
+
+def init_weight(weights, initWith, scale=1,
+                none_default=lambda shape: he_normal(shape, 1 / np.sqrt(2))):
```
delta2323: I agree that `none_default` makes the call statement of `init_weight` in some functions simpler. But `none_default` seems redundant if we only look at the interface of `init_weight`. It can be the matter of taste, but I suggest that we remove `none_default` and make the user code be in charge of setting default value of `initWith`.

chainer/initializations.py
```diff
@@ -0,0 +1,120 @@
...
+                `None`.
+
+    """
+    if initWith is None:
+        initWith = none_default
+    if callable(initWith):
+        initWith = scale * initWith(weights.shape)
+    if hasattr(initWith, 'shape'):
+        # check needed for bilinear tests to pass
+        assert weights.shape == initWith.shape
```
delta2323: Do you think it is better to impose `initWith` array to have exactly same shape as `weights` (i.e. broadcast is allowed only when `initWith` is scalar). In my opinion, it is one of the choice that we loosen the condition and admit `initWith` to have the shape broadcast-able to `weights`.

chainer/initializations.py
```diff
@@ -0,0 +1,120 @@
...
+    weights[...] = initWith
+
+
+def get_fans(shape):
```
delta2323: What does `fan` stands for in this context?

EdwardRaff:
I have less free time to work on this now, but I'll get around to it soon hopefully. Thanks for comments.

chainer/initializations.py
EdwardRaff: fan in and fan out, the terminology is fairly common in a number of papers.

chainer/initializations.py
```diff
@@ -0,0 +1,120 @@
...
+    u, _, v = np.linalg.svd(a, full_matrices=False)
+    # pick the one with the correct shape
+    q = u if u.shape == flat_shape else v
+    q = q.reshape(shape)
+    return scale * q[:shape[0], :shape[1]]
+
+
+def identity(shape, scale=1):
+    if len(shape) != 2 or shape[0] != shape[1]:
+        raise Exception('Identity matrix initialization can only be used '
```
delta2323: Use `ValueError` to include more information to Exception object itself.

chainer/initializations.py
```diff
@@ -0,0 +1,120 @@
...
+    Reference: Saxe et al., http://arxiv.org/abs/1312.6120
+
+    """
+    flat_shape = (shape[0], np.prod(shape[1:]))
+    a = np.random.normal(0.0, 1.0, flat_shape)
+    u, _, v = np.linalg.svd(a, full_matrices=False)
+    # pick the one with the correct shape
+    q = u if u.shape == flat_shape else v
+    q = q.reshape(shape)
+    return scale * q[:shape[0], :shape[1]]
```
delta2323: Could you add test cases for this function as this initializer is more complicated than others?

chainer/initializations.py
EdwardRaff: Personally, I'm not positive on where it would be useful to have an initialization that is broadcastable with the correct weight. If a matrix is being passed in, the only logical assumption that makes sense to me (IMHO) is that the caller wants this specific weight matrix to be the starting value of the weights. 

So if we are loading saved weights from another run, it should never be an issue. If for some reason they are manually constructing a matrix to use, I see no reason that the user shouldn't make their code produce a matrix of the correct shape. My personal intuition is that if you allow it to be broadcastable too, it may occasionally result in unintuitive behavior where something initializes that one would more logically expect to error out.

chainer/initializations.py
EdwardRaff: My only concern is that a number of the interfaces in chainer use `None` to indicate a default weight initalizaton scheme. I would rather keep the `none_default` to better handle/matchup with the inputs being passed from the constructors. If you really want it the other way though I will change it.

chainer/initializations.py
EdwardRaff: It may not be needed. It was in the original code from Keras.

chainer/__init__.py
EdwardRaff: I'm not good with python projects and that's how I got it to work :)

chainer/initializations.py
EdwardRaff: Would porting the tests from keras https://github.com/fchollet/keras/blob/master/tests/keras/test_initializations.py be sufficent for this one in particular, or would you like me to come up with a more rigiorus test?

chainer/links/connection/bilinear.py
```diff
@@ -23,14 +22,20 @@ class Bilinear(link.Link):
...
             Each element of this tuple must have the shapes of
             ``(left_size, output_size)``, ``(right_size, output_size)``,
             and ``(output_size,)``, respectively. If ``None``, :math:`V^1`
             and :math:`V^2` is initialized by scaled centered Gaussian
-            distributions and :math:`b` is set to :math:`0`.
+            distributions and :math:`b` is set to :math:`0`. May also be
+            a callable that takes in a tuple of the matrix shape and
+            returns an initialization, in which case :math:`b` is set to
+            :math:`0` but :math:`V^1` and :math:`V^2` will be initialized
+            from the callable.
```
delta2323: How about passing a 3-tuple of initializers to `initial_bias` and calling `init_weight` with each initializer. It works even if the tuple consists of `ndarry`s because `init_weight` can accept `ndarry`s as an argument `initWith`. Further it makes initialization of `b` more flexible.

chainer/links/connection/convolution_2d.py
```diff
@@ -25,9 +24,13 @@ class Convolution2D(link.Link):
         nobias (bool): If True, then this link does not use the bias term.
         use_cudnn (bool): If True, then this link uses CuDNN if available.
         initialW (4-D array): Initial weight value. If ``None``, then this
-            function uses to initialize ``wscale``.
+            function uses to initialize ``wscale``. May also be a callable
+             that takes a tuple of the matrix shape and returns a matrix
```
delta2323: a tuple of matrices

chainer/links/connection/convolution_2d.py
```diff
@@ -25,9 +24,13 @@ class Convolution2D(link.Link):
...
+             of the same dimensions to use for initialization.
         initial_bias (1-D array): Initial bias value. If ``None``, then this
-            function uses to initialize ``bias``.
+            function uses to initialize ``bias``. May also be a callable
+            that takes a tuple of the matrix shape and returns a matrix of
```
delta2323: ditto

chainer/links/connection/deconvolution_2d.py
```diff
@@ -31,9 +28,13 @@ class Deconvolution2D(link.Link):
             input size, stride and pad.
         use_cudnn (bool): If True, then this function uses CuDNN if available.
         initialW (4-D array): Initial weight value. If ``None``, then this
-            function uses to initialize ``wscale``.
+            function uses to initialize ``wscale``. May also be a callable
+            that takes a tuple of the matrix shape and returns a matrix of the
```
delta2323: ditto

chainer/links/connection/embed_id.py
```diff
@@ -15,6 +14,11 @@ class EmbedID(link.Link):
         in_size (int): Number of different identifiers (a.k.a. vocabulary
             size).
         out_size (int): Size of embedding vector.
+        initialW (2-D array): Initial weight value. If ``None``, then the
+                matrix is initialized from the standard normal distribution.
+            May also be a callable that takes a tuple of the matrix shape
+            and returns a matrix of the same dimensions to use for
```
delta2323: ditto

chainer/links/connection/embed_id.py
```diff
@@ -22,9 +26,12 @@ class EmbedID(link.Link):
...
 
     """
-    def __init__(self, in_size, out_size):
+
+    def __init__(self, in_size, out_size, initialW=None):
         super(EmbedID, self).__init__(W=(in_size, out_size))
-        self.W.data[...] = numpy.random.randn(in_size, out_size)
+        initializations.init_weight(
+            self.W.data, initialW,
+            none_default=lambda s: initializations.normal(s, scale=1.0))
```
delta2323: I remember that we had discussed whether we should make the initializer a class that takes parameters (`scale` in this case) in `__init__` and invoke initialization with `__call__`. If we use this design, we can use `initializers.Normal(scale)` and avoid using lambda expressions. So could you consider the possibility of this design?

chainer/links/connection/gru.py
```diff
@@ -100,16 +113,30 @@ class StatefulGRU(GRUBase):
     Args:
         in_size(int): Dimension of input vector :math:`x`.
         out_size(int): Dimension of hidden vector :math:`h`.
+        init: A callable that takes a tuple of the matrix shape and returns a
```
delta2323: ditto

chainer/links/connection/gru.py
```diff
@@ -100,16 +113,30 @@ class StatefulGRU(GRUBase):
...
+            matrix of the same dimensions to use for initialization of the
+            GRU's input units (:math:`W`). Maybe be `None` to use default
+            initialization.
+        inner_init: A callable that takes a tuple of the matrix shape and
+            returns a matrix of the same dimensions to use for initialization
+            of the GRU's inner recurrent units (:math:`U`). Maybe be `None` to
+            use default initialization
+        bias_init: A callable or scalar used to initiate the bias values for
```
delta2323: initialize?

chainer/links/connection/gru.py
delta2323: Please write the type of `init` after the argument name.

chainer/links/connection/gru.py
```diff
@@ -100,16 +113,30 @@ class StatefulGRU(GRUBase):
...
         in_size(int): Dimension of input vector :math:`x`.
         out_size(int): Dimension of hidden vector :math:`h`.
+        init: A callable that takes a tuple of the matrix shape and returns a
+            matrix of the same dimensions to use for initialization of the
+            GRU's input units (:math:`W`). Maybe be `None` to use default
+            initialization.
+        inner_init: A callable that takes a tuple of the matrix shape and
+            returns a matrix of the same dimensions to use for initialization
+            of the GRU's inner recurrent units (:math:`U`). Maybe be `None` to
+            use default initialization
```
delta2323: put a period at the end of the sentence.

chainer/links/connection/gru.py
```diff
@@ -100,16 +113,30 @@ class StatefulGRU(GRUBase):
...
+        bias_init: A callable or scalar used to initiate the bias values for
+            both the GRU's inner and input units. Maybe be `None` to use
+            default initialization
```
delta2323: ditt (add a period)

chainer/links/connection/inception.py
```diff
@@ -29,16 +29,38 @@ class Inception(link.Chain):
         proj5 (int): Projection size of 5x5 convolution path.
         out5 (int): Output size of 5x5 convolution path.
         proj_pool (int): Projection size of max pooling path.
+        convInit: A callable or scalar that takes a tuple of the matrix shape
```
delta2323: ditto

chainer/links/connection/inception.py
```diff
@@ -29,16 +29,38 @@ class Inception(link.Chain):
...
+        biasInit: A callable or scalar that takes a tuple of the matrix shape
+                and returns a matrix of the same dimensions to use for
+                initialization of the convolution bias weights. Maybe be `None`
+                to use default initialization.
 
     """
-    def __init__(self, in_channels, out1, proj3, out3, proj5, out5, proj_pool):
+
+    def __init__(self, in_channels, out1, proj3, out3, proj5, out5, proj_pool,
+                 convInit=None, biasInit=None):
```
delta2323: Use snake cases (`conv_init` and `bias_init`)

chainer/links/connection/inceptionbn.py
```diff
@@ -29,6 +29,10 @@ class InceptionBN(link.Chain):
         pooltype (str): Pooling type. It must be either ``'max'`` or ``'avg'``.
         proj_pool (bool): If True, do projection in the pooling path.
         stride (int): Stride parameter of the last convolution of each path.
+        convInit: A callable or scalar that takes a tuple of the matrix shape
```
delta2323: ditto(use snake case)

delta2323: ditto (matrix shape)

chainer/links/connection/inceptionbn.py
```diff
@@ -37,19 +41,24 @@ class InceptionBN(link.Chain):
             training mode. If False, they are used in testing mode.
 
     """
+
     def __init__(self, in_channels, out1, proj3, out3, proj33, out33,
-                 pooltype, proj_pool=None, stride=1):
+                 pooltype, proj_pool=None, stride=1, convInit=None):
```
delta2323: ditto (use snake case)

chainer/links/connection/linear.py
```diff
@@ -25,9 +23,12 @@ class Linear(link.Link):
         bias (float): Initial bias value.
         nobias (bool): If True, then this function does not use the bias.
         initialW (2-D array): Initial weight value. If ``None``, then this
-            function uses to initialize ``wscale``.
+            function uses to initialize ``wscale``. May also be a function
+            that takes a tuple of (outpu_size, input_size) and returns a
```
delta2323: outpu_size -> output_size

delta2323: enclose with double back quotations as ```(output_size, input_size)```

chainer/links/connection/linear.py
```diff
@@ -25,9 +23,12 @@ class Linear(link.Link):
...
+            matrix of the same dimensions to use for initialization
         initial_bias (1-D array): Initial bias value. If ``None``, then this
-            function uses to initialize ``bias``.
+            function uses to initialize ``bias``. May also be a function
+            that behaves in the same manner as initalW
```
delta2323: Put a period at the end of the sentence.

chainer/links/connection/lstm.py
```diff
@@ -27,14 +28,31 @@ class LSTM(link.Chain):
         h (chainer.Variable): Output at the previous timestep.
 
     """
-    def __init__(self, in_size, out_size):
+
+    def __init__(self, in_size, out_size,
+                 lateral_init=initializations.orthogonal, upward_init=None,
+                 bias_init=0, forget_bias_init=1):
```
delta2323: Add documentation about `lateral_init`, `upward_init`, `bias_init`, and `forget_bias_init`.

chainer/initializations.py
delta2323: I think `q[:shape[0], :shape[1]]` is same as `q` itself. So why do you use this indexing?

chainer/links/connection/gru.py
EdwardRaff: What should I put as the type since it can be one of many things? (callable or matrix or scalar)?

chainer/links/connection/embed_id.py
EdwardRaff: I don't personally think it's worth the code bloat, but I'll work on it.

chainer/initializations.py
EdwardRaff: The code is from keras, I don't know why it did that.

delta2323: OK, then could you replace `q[:shape[0], :shape[1]]` with `q` ?

delta2323: As for tests, I imagined the tests that checks orthogonality of the result. But it is worth adding such unit tests you mentioned. Further, I come to think we should also add such unit tests to other initializers as some of them have not been called.

chainer/initializations.py
delta2323: OK. Then, could you remove it?

chainer/initializations.py
delta2323: I understand it. Then I agree with keeping your API as it is. I will propose you the alternative if I come up with it.

chainer/initializations.py
delta2323: Then, it is OK as it is.

chainer/initializations.py
```diff
@@ -0,0 +1,120 @@
...
+    if callable(initWith):
+        initWith = scale * initWith(weights.shape)
+    if hasattr(initWith, 'shape'):
+        # check needed for bilinear tests to pass
+        assert weights.shape == initWith.shape
+    weights[...] = initWith
+
+
+def get_fans(shape):
+    fan_in = shape[1] if len(shape) == 2 else np.prod(shape[1:])
```
delta2323: I think we do not have to divide the case based on the length of `shape`. Just `fan_in = np.prod(shape[1:])` is enough.

chainer/links/connection/gru.py
delta2323: > What should I put as the type since it can be one of many things? (callable or matrix or scalar)?

I imagined the list of possible types of arguments as you suggest.

chainer/initializations.py
delta2323: I agree with your idea to avoid implicit broadcast here. Generally speaking we do not need to broaden the possible types of inputs of a function. We can broaden the possible type afterwards if necessary.

chainer/initializations.py
delta2323: I realized `init_weight` can handle not only matrices but also tensors with arbitrary dimension as weights. Actually, existing `initializer`s like `uniform` or `normal` do be able to initialize tensors. So could you change the description of `init_weight` to support tensors?

delta2323:
Hi, as we would like to add it to 1.8.0, I would appreciate you if you tell me whether you have time to work on fixing this PR.

EdwardRaff:
I'm afraid I don't have time right now. I should have more time in the summer, but right now I just don't have the bandwidth anymore. I have no problem if someone wants to pick up where I left off, otherwise I'll work my way back to it once I'm no longer swamped.

delta2323:
@EdwardRaff OK, I understand it. Then I will take over this PR. Let me ask you if I have some questions about your implementation. Thank you for your considerable contribution so far!

PR #828, opened 2016-01-05
add l2 norm squared function

hrantzsch:
Adds a function to compute square of l2 norm and a test for it.


hrantzsch:
Oh, flake8 does not agree with the imports, I will adjust this.
Btw. this style was introduced by isort, which I used as described in the coding guidelines.

delta2323:
Thank you for PR!

> Btw. this style was introduced by isort, which I used as described in the coding guidelines.
Yes, the result of isort does not necessarily pass flake8 (e.g. isort introduce multiple import, whereas flake8 forbids it).
As we check coding styleby flake8 in CI, we manually change the imports.

chainer/functions/__init__.py
```diff
@@ -43,6 +43,7 @@
 from chainer.functions.noise import dropout
 from chainer.functions.noise import gaussian
 from chainer.functions.normalization import batch_normalization
+from chainer.functions.normalization import l2_norm_squared
```
delta2323: I think this function should not be in `normalization` because it does not normalize vectors. A alternative directory is `math`.

chainer/functions/normalization/l2_norm_squared.py
```diff
@@ -0,0 +1,44 @@
...
+
+        type_check.expect(
+            x_type.dtype == np.float32,
+            x_type.ndim >= 2,
+        )
+
+    def forward_cpu(self, inputs):
+        x, = inputs
+        N = x.shape[0]
+        self.norm = np.zeros(N, dtype=np.float32)
```
delta2323: `self.norm` need not to initialize with zero because it will be overwritten. `np.empty` is more appropriate.

chainer/functions/normalization/l2_norm_squared.py
```diff
@@ -0,0 +1,44 @@
...
+            'T x', 'T y', 'x * x', 'a + b', 'y = a', '0', 'l2normsquared'
+        )
+        return l2normsquared_kernel(x, axis=1),
+
+    def backward(self, inputs, gy):
+        return 2 * inputs * gy[0],
+
+
+def l2_norm_squared(x):
+    """L2 normalization (a.k.a. Euclidean norm) squared.
```
delta2323: I think "L2 norm" is more common than "L2 normalization".

chainer/functions/normalization/l2_norm_squared.py
```diff
@@ -0,0 +1,44 @@
...
+
+    This function implements the square of L2 normalization on a vector.
```
delta2323: ditto(L2 normalization)

tests/chainer_tests/functions_tests/normalization_tests/test_l2_norm_squared.py
```diff
@@ -0,0 +1,64 @@
...
+        self.x = numpy.random.uniform(-1, 1, (2, 7)).astype(numpy.float32)
+        self.gy = numpy.random.uniform(-1, 1, (2, 7)).astype(numpy.float32)
+
+    def check_forward(self, x_data):
+        x = chainer.Variable(x_data)
+        y = functions.l2_norm_squared(x)
+        self.assertEqual(y.data.dtype, numpy.float32)
+        y_data = cuda.to_cpu(y.data)
+
+        y_expect = numpy.zeros(self.x.shape[0])
```
delta2323: `len(self.x)` is simpler. It is equivalent to `self.x.shape[0]`.

delta2323:
Please add test cases where ndim of input `x` is more than two dimensional.

chainer/functions/normalization/l2_norm_squared.py
```diff
@@ -0,0 +1,44 @@
...
+        )
+
+    def forward_cpu(self, inputs):
+        x, = inputs
+        N = x.shape[0]
+        self.norm = np.zeros(N, dtype=np.float32)
+        for i in range(N):
+            self.norm[i] = x[i].dot(x[i])
+        return self.norm,
+
```
delta2323: How about `self.norm = numpy.array([b[i].dot(b[i]) for i in range(len(b))], dtype=numpy.float32)` . It need not to initialize `self.norm`.

tests/chainer_tests/functions_tests/normalization_tests/test_l2_norm_squared.py
delta2323: ditto(use `numpy.empty` instead of `numpy.zeros`)

tests/chainer_tests/functions_tests/normalization_tests/test_l2_norm_squared.py
```diff
@@ -0,0 +1,64 @@
...
+        for n in six.moves.range(self.x.shape[0]):
+            y_expect[n] = sum(map(lambda x: x * x, self.x[n]))
```
delta2323: I think it does not work when `x` is more than 2-dimentional.

chainer/functions/normalization/l2_norm_squared.py
delta2323: Please add description that this function does not do reduction along batch axis.

delta2323: Please add description when `x` is more than two dimentional.

delta2323:
@hrantzsch I added comments. Please check them.

hrantzsch:
Thank you for your comments! I adjusted accordingly.
Please note that I introduced significant changes to the backprop function, since I discovered mistakes there.
Furthermore I don't store self.norm anymore, as it seemed unnecessary.

delta2323:
I am coming to think that we should rename this function to something that emphasizes data-wise operation (like `BatchL2NormSquared`), meaning that we take the squared norm for each data in the input minibatch.

It is natural someone to misunderstand and expect this function to return `x.flatten().dot(x.flatten())` , i.e. regard input `x` as single vector and take square of its L2 norm.

How do you think of it?

hrantzsch:
I see what you mean. I looked at batch_matmul vs matmul and given that it makes a lot of sense.
I think I'll change the name as you proposed.

delta2323:
OK, please go ahead.

tests/chainer_tests/functions_tests/math_tests/test_batch_l2_norm_squared.py
```diff
@@ -0,0 +1,72 @@
...
+    def test_forward_cpu(self):
+        self.check_forward(self.x)
+
+    @attr.gpu
+    @condition.retry(3)
+    def test_forward_gpu(self):
+        self.check_forward(cuda.to_gpu(self.x))
+
+    def check_backward(self, x_data, y_grad):
+        x = chainer.Variable(_as_two_dim(x_data))
```
delta2323: I think `x` should not wrap reshaped `x_data` because it makes two dimensional and three dimensional case identical.

delta2323:
Please add `batch_l2_norm_squared` to `docs/source/reference/functions.rst`.

delta2323:
Please add test case in which `batch_l2_norm_squared` raises `type_check.InvalidType` when input array is one-dimensional. I think [this test](https://github.com/pfnet/chainer/blob/23f7d25ef9e2db84cc6735f5bdeb0a6b9ba280ba/tests/chainer_tests/functions_tests/evaluation_tests/test_binary_accuracy.py#L55) is helpful for implementation.

delta2323:
Thank you for applying comments. LGTM except the commented above.

hrantzsch:
Great! I'll fix the remaining issues you stated and wait for #848 to be merged, so I can rebase and make the tests pass.
One more thing, I noticed GPU implementation of the backward function might be desirable. If you agree I would also implement that before we go on.

delta2323:
> One more thing, I noticed GPU implementation of the backward function might be desirable. If you agree I would also implement that before we go on.

We would like to include this PR in v1.6. So if you can apply this change today, I will be happy to do so.
If you cannot afford to do so, it is still OK because we can separate this change as another PR and merge this change after we release v1.6 (as an improvement of `BatchL2NormSquared`).

hrantzsch:
Alright, in that case I will just implement the changes you suggested above today.

delta2323:
Now #848 is merged. I will merge this PR after you rebase or merge the master branch and fix errors on flake8. Do you have time to work on it?

delta2323:
OK, then I will merge this PR as it is and fix the errors regarding #848 error after the merge. As for GPU implementation of backward function, could you send it as another PR? Thank you for your great work!

hrantzsch:
Thanks! Sorry I couldn't rebase it myself. The time offset...
Anyways, I created #871 for the GPU backward function.

PR #830, opened 2016-01-06
Add Variable.debug_print method

### LINKED ISSUE #786: Add debug_print function to Variable
Currently, it is little hard to confirm ``Variable`` properties one by one. How about adding ``Variable.debug_print`` method to show required info at once? I'm willing to send a PR if it looks useful.

I've created utility func to show info as below:
- device which variable is on
- volatile flag
- background class of ``data`` 
- ``data.shape``
- ``data.dtype``
- ``data`` statistics (mean and std)
- ``grad`` ( none, all 0 or mean/std )

```
arr = np.random.randn(5, 3, 5, 5)

vcpu = Variable(arr)
vcpu.debug_print()
# <variable at 0x7fdc22141810>
# - device: CPU
# - volatile: OFF
# - background: <type 'numpy.ndarray'>
# - shape: (5, 3, 5, 5)
# - dtype: float64
# - statistics: mean=-0.0130620493847, std=1.00905897902
# - grad: None

vgpu = Variable(arr)
vgpu.to_gpu()
vgpu.zerograd()
vgpu.debug_print()
# <variable at 0x7fdc22141a50>
# - device: <CUDA Device 0>
# - volatile: OFF
# - background: <type 'cupy.core.core.ndarray'>
# - shape: (5, 3, 5, 5)
# - dtype: float64
# - statistics: mean=-0.0130620493847, std=1.00905897902
# - grad: 0

vcpu.grad = np.random.randn(5, 3, 5, 5)
vcpu.debug_print()
# <variable at 0x7fdc22141810>
# - device: CPU
# - volatile: OFF
# - background: <type 'numpy.ndarray'>
# - shape: (5, 3, 5, 5)
# - dtype: float64
# - statistics: mean=-0.0130620493847, std=1.00905897902
# - grad: mean=0.0081744735799, std=1.01002850276
```

sinhrks:
Closes #786. 

@delta2323  Could you take a look?

### DIFF
--- a/chainer/variable.py
+++ b/chainer/variable.py
@@ -80,7 +80,7 @@ class Variable(object):
 - statistics: {stats}
 - grad: {grad}"""
 
-        stats_msg = 'mean={0}, std={1}'
+        stats_msg = 'mean={0:.8f}, std={1:.8f}'
 
         try:
             device = self.data.device
--- a/chainer/variable.py
+++ b/chainer/variable.py
@@ -68,6 +68,45 @@ class Variable(object):
     def __str__(self):
         return self.name or ('<var@%x>' % id(self))
 
+    def debug_print(self):
+        """Display a summary of the stored data and location of the Variable"""
+
+        msg = """{summary}
+- device: {device}
+- volatile: {volatile}
+- backend: {background}
+- shape: {shape}
+- dtype: {dtype}
+- statistics: {stats}
+- grad: {grad}"""
+
+        stats_msg = 'mean={0}, std={1}'
+
+        try:
+            device = self.data.device
+        except AttributeError:
+            device = 'CPU'
+
+        with cuda.get_device(self.data) as dev:
+            xp = numpy if int(dev) == -1 else cuda.cupy
+
+            if self.grad is None:
+                grad = None
+            elif xp.all(self.grad == 0):
+                grad = 0
+            else:
+                grad = stats_msg.format(float(xp.mean(self.grad)),
+                                        float(xp.std(self.grad)))
+
+            stats = stats_msg.format(float(xp.mean(self.data)),
+                                     float(xp.std(self.data)))
+
+        return msg.format(summary=repr(self), volatile=self.volatile,
+                          grad=grad, shape=self.data.shape,
+                          background=type(self.data),
+                          dtype=self.data.dtype, device=device,
+                          stats=stats)
+
     def __pos__(self):
         return self
 
--- a/tests/chainer_tests/test_variable.py
+++ b/tests/chainer_tests/test_variable.py
@@ -7,6 +7,7 @@ from chainer import cuda
 from chainer import testing
 from chainer.testing import attr
 
+import re
 import six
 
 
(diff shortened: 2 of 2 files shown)


chainer/variable.py
```diff
@@ -68,6 +68,46 @@ def __repr__(self):
...
+        except AttributeError:
+            device = 'CPU'
+
+        with cuda.get_device(self.data) as dev:
+            xp = numpy if int(dev) == -1 else cuda.cupy
+
+            if self.grad is None:
+                grad = None
+            else:
+                if xp.all(self.grad == 0):
```
delta2323: How about using `elif`? It reduces indent blocks.

tests/chainer_tests/test_variable.py
```diff
@@ -366,6 +366,67 @@ def test_addgrad_gpu_to_another_gpu(self):
...
+
+        msg = "grad: mean={mean}, std={std}".format(mean=grad.mean(),
+                                                    std=grad.std())
+        self.assertTrue(msg in result)
+
+    @attr.gpu
+    def test_debug_print_gpu(self):
+        arr = np.random.randn(5, 3, 5, 5).astype(np.float32)
+        v = chainer.Variable(arr)
+        v.to_gpu()
```
delta2323: I think it is better to explicitly send `v` to GPU 0 because default device can be other than 0 in some environment (although I am not certain such an environment does exist).

tests/chainer_tests/test_variable.py
```diff
@@ -366,6 +366,67 @@ def test_addgrad_gpu_to_another_gpu(self):
             c = cp.full(3, 30, dtype=np.float32)
         self.check_addgrad(a, b, c)
 
+    def test_debug_print_cpu(self):
```
delta2323: `test_debug_print_cpu` and `test_debug_print_gpu` do not use attributes of `self`. So these need not be in `TestVariable` class. Could you consider the possibility of separate them as another class?

delta2323: Can we apply DRY principle to `test_debug_cpu` and `test_debug_gpu` and unify part of them?

chainer/variable.py
```diff
@@ -68,6 +68,46 @@ def __repr__(self):
     def __str__(self):
         return self.name or ('<var@%x>' % id(self))
 
+    def debug_print(self):
+        """Display stored data and location of the Variable"""
```
delta2323: I think "Display **summary** of stored data" is more appropriate because it does not show whole data.

delta2323:
@sinhrks I added comments. Could you check them?

sinhrks:
Thanks, updated except one point.

tests/chainer_tests/test_variable.py
sinhrks: Other test seems to have separate / duplicated methods for ``debug_cpu`` and ``debug_gpu``. Is there any example which you intend for?

delta2323: Although it is not a rule but just a convention, we made `check_xxx` method and call it from `test_xxx_cpu` and `test_xxx_gpu` to test some single features with CPU and GPU.
You can see the example e.g. [here](https://github.com/pfnet/chainer/blob/master/tests/chainer_tests/links_tests/connection_tests/test_linear.py#L38).

delta2323:
@sinhrks  I updated the comment. Could you check it?
https://github.com/pfnet/chainer/pull/830#discussion_r50502613

sinhrks:
@delta2323 Updated based on your comments. Pls review.

tests/chainer_tests/test_variable.py
```diff
@@ -367,6 +367,54 @@ def test_addgrad_gpu_to_another_gpu(self):
...
 
 
+class TestDebugPrint(unittest.TestCase):
+
+    def setUp(self):
+        self.arr = np.random.randn(5, 3, 5, 5).astype(np.float32)
+
+    def check_debug_print(self, v):
+        result = v.debug_print()
+        self.assertTrue(repr(v) in result)
```
delta2323: Use [`self.assertIn`](http://docs.python.jp/2/library/unittest.html#unittest.TestCase.assertIn)

tests/chainer_tests/test_variable.py
```diff
@@ -367,6 +367,54 @@ def test_addgrad_gpu_to_another_gpu(self):
...
+        self.assertTrue('volatile: OFF' in result)
```
delta2323: ditto (`assertIn`)

tests/chainer_tests/test_variable.py
```diff
@@ -367,6 +367,54 @@ def test_addgrad_gpu_to_another_gpu(self):
...
+        self.assertTrue("dtype: float32" in result)
```
delta2323: ditto (`assertIn`)

tests/chainer_tests/test_variable.py
```diff
@@ -367,6 +367,54 @@ def test_addgrad_gpu_to_another_gpu(self):
...
+
+        # no grad
+        msg = "statistics: mean="
+        self.assertTrue(msg in result)
```
delta2323: ditto (`assertIn`)

tests/chainer_tests/test_variable.py
```diff
@@ -367,6 +367,54 @@ def test_addgrad_gpu_to_another_gpu(self):
...
+        self.assertTrue("grad: None" in result)
```
delta2323: ditto (`assertIn`)

tests/chainer_tests/test_variable.py
```diff
@@ -367,6 +367,54 @@ def test_addgrad_gpu_to_another_gpu(self):
...
+
+        # zero grad
+        v.zerograd()
+        result = v.debug_print()
+        self.assertTrue("grad: 0" in result)
```
delta2323: ditto (`assertIn`)

tests/chainer_tests/test_variable.py
```diff
@@ -367,6 +367,54 @@ def test_addgrad_gpu_to_another_gpu(self):
...
+
+        # add grad
+        v.grad = v.data
+        result = v.debug_print()
+
+        msg = "grad: mean="
+        self.assertTrue(msg in result)
```
delta2323: ditto (`assertIn`)

tests/chainer_tests/test_variable.py
```diff
@@ -367,6 +367,54 @@ def test_addgrad_gpu_to_another_gpu(self):
...
+
+    def test_debug_print_cpu(self):
+        v = chainer.Variable(self.arr)
+        result = v.debug_print()
+        self.assertTrue('device: CPU' in result)
```
delta2323: ditto (`assertIn`)

tests/chainer_tests/test_variable.py
```diff
@@ -367,6 +367,54 @@ def test_addgrad_gpu_to_another_gpu(self):
...
+        self.assertTrue("numpy.ndarray" in result)
```
delta2323: ditto (`assertIn`)

tests/chainer_tests/test_variable.py
```diff
@@ -367,6 +367,54 @@ def test_addgrad_gpu_to_another_gpu(self):
...
+
+        self.check_debug_print(v)
+
+    @attr.gpu
+    def test_debug_print_gpu(self):
+        v = chainer.Variable(self.arr)
+        v.to_gpu(0)
+
+        result = v.debug_print()
+        self.assertTrue('device: <CUDA Device 0>' in result)
```
delta2323: ditto (`assertIn`)

tests/chainer_tests/test_variable.py
```diff
@@ -367,6 +367,54 @@ def test_addgrad_gpu_to_another_gpu(self):
...
+        self.assertTrue("cupy.core.core.ndarray" in result)
```
delta2323: ditto (`assertIn`)

tests/chainer_tests/test_variable.py
delta2323: Please add assertion that checks `shape`.

tests/chainer_tests/test_variable.py
delta2323: Could you add the assertion that checks statistics should include stddev?

tests/chainer_tests/test_variable.py
delta2323: ditto (check that statistics should include stddev)

delta2323:
I updated the comments. Could you check them?

sinhrks:
OK, fixed.

delta2323:
Please use quotes consistently, that is, do not mix single quotes or double quotes.
https://www.python.org/dev/peps/pep-0008/#string-quotes

delta2323:
I have added one comment. LGTM if you fix it.

delta2323:
LGTM :+1:

delta2323:
Thank you very much!

PR #834, opened 2016-01-08
Fix assertion name in cupy.testing

delta2323:
Related to #779, #833

### DIFF
--- a/tests/cupy_tests/testing_tests/test_helper.py
+++ b/tests/cupy_tests/testing_tests/test_helper.py
@@ -82,40 +82,22 @@ class TestCheckCupyNumpyError(unittest.TestCase):
                 numpy_error, numpy_tb, accept_error=False)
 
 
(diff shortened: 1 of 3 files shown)


unnonouno:
It looks correct, but i think the test is complicated.

tests/cupy_tests/testing_tests/test_helper.py
```diff
@@ -78,3 +80,140 @@ def test_forbidden_error(self):
             helper._check_cupy_numpy_error(
                 self, cupy_error, cupy_tb,
                 numpy_error, numpy_tb, accept_error=False)
+
+
+def make_func(np_result, cp_result):
```
unnonouno: It is too generalized to understand. How about making simplified functions like `numpy_error` and `cupy_error`?

unnonouno:
LGTM

PR #853, opened 2016-01-17
Fix the order of dimentsions of Maxout's weight tensor and bias vector.

### LINKED ISSUE #849: Change the order of indices of Maxout's weight tensor
In the current implementation, Maxout's weight tensor is of the form `(input_dim, channel_dim, output_dim)`, but I would like to change it to `(input_dim, output_dim, channel_dim)` in two reasons.

* The order of original paper is the latter one.
* As Maxout take maximum along channel axis, we can expect faster computation when the last axis is channel one.

As Maxout is not included in any release, I would like to apply this change until the next release v1.6.

delta2323:
Fix #849 

I am wondering we should change the order of argument of `chainer.links.Maxout.__init__` from `(self, in_size, num_channels, out_size, ...)` to `(self, in_size, out_size, num_channels, ...)`.

### DIFF
--- a/chainer/functions/activation/maxout.py
+++ b/chainer/functions/activation/maxout.py
@@ -43,8 +43,8 @@ class MaxoutFunction(function.Function):
         ys = xp.tensordot(x, W, axes=1)
         if len(inputs) == 3:
             ys += inputs[2]
-        self.argmax = xp.argmax(ys, axis=1)
-        return xp.max(ys, axis=1),
+        self.argmax = xp.argmax(ys, axis=2)
+        return xp.max(ys, axis=2),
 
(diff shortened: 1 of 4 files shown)


beam2d:
The order of arguments of the initializer and the shape of the weight tensor are independent and we can choose any combination, though I think `(in_size, out_size, num_channels)` is better than `(in_size, num_channels, out_size)` as the order of arguments because it is consistent to other links that let the first two arguments represent the input size and the output size (e.g. Linear, Convolution2D, etc.).

delta2323:
As we reimplemented Maxout in #856, this PR is obsolete. So I close it.

PR #871, opened 2016-01-19
add backward_gpu for batch_l2_norm_squared

hrantzsch:


chainer/functions/math/batch_l2_norm_squared.py
```diff
@@ -35,12 +35,20 @@ def forward_gpu(self, inputs):
...
-    def backward(self, inputs, gy):
+    def backward_cpu(self, inputs, gy):
         x = _as_two_dim(inputs[0])
         xp = cuda.get_array_module(x, gy)
         return xp.array([2 * x[i] * gy[0][i] for i in six.moves.range(len(x))],
                         dtype=np.float32).reshape(inputs[0].shape),
 
+    def backward_gpu(self, inputs, gy):
+        x = _as_two_dim(inputs[0])
+        gy_repeat = gy[0].reshape(len(gy[0]), 1).repeat(x.shape[1], axis=1)
```
hrantzsch: I'm not really happy with repeating the grad array like this, but it was the only way I found to feed it into the elementwise kernel. Can you suggest a better way?

delta2323: How about using `raw` keyword?
```
        kernel = cuda.elementwise(
            'T x, raw T gy, int32 N', 'T gx', 'gx = 2 * x * gy[i / N]', 'l2normsquared_bwd')
        gx = kernel(x, gy[0], x.shape[1]).reshape(inputs[0].shape)
```

hrantzsch:
Very nice. I'll do exactly that. Thanks!

delta2323:
LGTM :+1:

delta2323:
Thank you very much!

PR #874, opened 2016-01-20
Add gradient_check.check_backward to check.rst

ronekko:
`gradient_check.check_backward` is missing in the reference manual.


delta2323:
Thank you very much! I found that generated document violates the syntax of docstring. I will fix it later.

PR #877, opened 2016-01-21
Deduplicate Chain.to_cpu/to_gpu

beam2d:
to_cpu/to_gpu methods of Chain/ChainList call these methods of their children recursively, while it also calls Link.to_cpu/to_gpu which call these methods of all parameter variables in the hierarchy. It results in duplicated call of to_cpu/to_gpu of parameters in the hierarchy. I fixed this duplication in this PR by only calling to_cpu/to_gpu of parameters directly belonging to the link.

### DIFF
--- a/chainer/link.py
+++ b/chainer/link.py
@@ -192,9 +192,9 @@ class Link(object):
         """
         if self._cpu:
             return self
-        for param in self.params():
-            param.to_cpu()
         d = self.__dict__
+        for name in self._params:
+            d[name].to_cpu()
         for name in self._persistent:
             value = d[name]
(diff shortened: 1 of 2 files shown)


delta2323:
Can you add the test case where we exactly call `to_cpu` or `to_gpu` methods of each `Variable` in a hierarchical `Chain` or `ChainList` (not zero nor more than once) ?

beam2d:
fixed.

delta2323:
Even the code before change passes the the test code added. We must check not `Link.to_cpu`s but  `Variable.to_cpu`s are called exactly once.

delta2323:
LGTM :+1:

PR #883, opened 2016-01-25
Tutorial patch

zori:
Minor fixes of typos. For clarity, added explicitly `model.to_gpu()` to the code example, as the text immediately before it suggests.

### DIFF
--- a/docs/source/tutorial/gpu.rst
+++ b/docs/source/tutorial/gpu.rst
@@ -179,7 +179,7 @@ The :meth:`~Link.to_gpu` method also accepts a device ID like ``model.to_gpu(0)`
 In this case, the link object is transferred to the appropriate GPU device.
 The current device is used by default.
 
-Then, all we have to do is transfer each minibatch to the GPU:
+Then we have to transfer each minibatch to the GPU:
 
 .. testcode::
    :hide:
--- a/docs/source/tutorial/gpu.rst
+++ b/docs/source/tutorial/gpu.rst
@@ -175,11 +175,11 @@ A :class:`Link` object can be transferred to the specified GPU using the :meth:`
    optimizer.setup(model)
 
 This time, we make the number of input, hidden, and output units configurable.
-The :meth:`~Link.to_gpu` method also accpets a device ID like ``model.to_gpu(0)``.
+The :meth:`~Link.to_gpu` method also accepts a device ID like ``model.to_gpu(0)``.
 In this case, the link object is transferred to the appropriate GPU device.
 The current device is used by default.
 
-Then, all we have to do is transferring each minibatch to the GPU:
+Then, all we have to do is transfer each minibatch to the GPU:
 
 .. testcode::
    :hide:
@@ -189,6 +189,7 @@ Then, all we have to do is transferring each minibatch to the GPU:
 
 .. testcode::
 
+   model.to_gpu()
    batchsize = 100
    datasize = len(x_train)
    for epoch in range(20):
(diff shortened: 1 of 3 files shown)


docs/source/tutorial/gpu.rst
```diff
@@ -175,11 +175,11 @@ A :class:`Link` object can be transferred to the specified GPU using the :meth:`
    optimizer.setup(model)
 
 This time, we make the number of input, hidden, and output units configurable.
-The :meth:`~Link.to_gpu` method also accpets a device ID like ``model.to_gpu(0)``.
+The :meth:`~Link.to_gpu` method also accepts a device ID like ``model.to_gpu(0)``.
 In this case, the link object is transferred to the appropriate GPU device.
 The current device is used by default.
 
-Then, all we have to do is transferring each minibatch to the GPU:
+Then, all we have to do is transfer each minibatch to the GPU:
```
delta2323: Is it grammatically incorrect?

docs/source/tutorial/type_check.rst
```diff
@@ -232,7 +231,7 @@ More complicated cases
 
 How to write a more complicated condition that can't be written with these operators?
 You can evaluate :class:`utils.type_check.Expr` and get its result value with :meth:`eval` method.
-And, check the condition and show warning message by your hand:
+Then check the condition and show warning message by hand:
```
delta2323: Don't we need comma after `Then`?

docs/source/tutorial/type_check.rst
```diff
@@ -281,9 +280,9 @@ And then, get each type:
 
    x_type, y_type = in_types
 
-Don't get each value before check ``in_types.size()``.
-When the number of argument is illegal, this process may fail.
-For example, this code doesn't work when the size of ``in_types`` is zero:
+Don't get each value before checking ``in_types.size()``.
+When the number of argument is illegal, this will fail.
```
delta2323: This change alter the meaning of the document. I am now investigating this change is acceptable.

delta2323:
@zori I added the comment. Could you check it, please?

docs/source/tutorial/type_check.rst
delta2323: I think the wording "may fail" is appropriate because there are some situations `type_check.expect` does not "fail", or, `type_check.expect` can correctly throw an appropriate error even if the number of argument of `__call__` is incorrect.

For example, suppose the expected number of argument is 2 and `ndim` of first argument is 3 (this is same example as the tutorial). If the number of argument is 0, `type_check.expect` will "fail", or it emits strange error message. But if the number of argument is 3, it does not "fail", or, `type_check.expect` can correctly throw `InvalidType` error. 

I think the meaning of `fail` here is difficult to understand. How about changing the description to "When the number of argument is illegal, `type_check.expect` can output strange error messages"?

docs/source/tutorial/gpu.rst
zori: I think after "have to / must / should" we use infinitive (i.e., "transfer", not the gerund, which would be "transferring").

Simpler version is "we have to transfer each minibatch..."

docs/source/tutorial/type_check.rst
zori: Then doesn't need a comma, unless there is a secondary clause after it. This would be like a second sentence giving more information; without it it is still a complete sentence. Like this:

"""
Then, since we know how to evaluate, check the condition and show warning message by hand.
"""

delta2323: OK, I understand it. You do not need to add a comma here.

docs/source/tutorial/type_check.rst
zori: SGTM.

zori:
If it looks good now, I think that's all from me.

delta2323:
LGTM :+1:

delta2323:
Thank you for your PR!

PR #884, opened 2016-01-26
Direct DtoD copy

### LINKED ISSUE #882: Device-to-device transfer unnecessarily visits the host memory
CUDA support direct copy between multiple GPUs, while it seems CuPy's device-to-device copy does not utilize this feature correctly. Here is the minimum example.

```python
import cupy
x = cupy.array(1)  # x is on the device 0
with cupy.cuda.Device(1):
    x.copy()  # copy x to the device 1
```

The following command shows this code invokes device-to-host and host-to-device copies.

```
$ nvprof --devices 0,1 --print-gpu-summary python dtod.py
==16767== NVPROF is profiling process 16767, command: python dtod.py
==16767== Profiling application: python dtod.py
==16767== Profiling result:
 66.03%  5.5360us         2  2.7680us  1.1200us  4.4160us  [CUDA memcpy HtoD]
 33.97%  2.8480us         1  2.8480us  2.8480us  2.8480us  [CUDA memcpy DtoH]
```

beam2d:
fix #882.
This PR adds cudaDeviceEnablePeerAccess call right before the first copy between given two devices.

### DIFF
--- a/cupy/cuda/memory.pyx
+++ b/cupy/cuda/memory.pyx
@@ -124,7 +124,7 @@ cdef class MemoryPointer:
         if size > 0:
             _set_peer_access(src.device.id, self.device.id)
             runtime.memcpy(self.ptr, src.ptr, size,
-                           runtime.memcpyDeviceToDevice)
+                           runtime.memcpyDefault)
 
     cpdef copy_from_device_async(self, MemoryPointer src, size_t size, stream):
         """Copies a memory sequence from a (possibly different) device asynchronously.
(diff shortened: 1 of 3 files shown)


cupy/cuda/memory.pyx
```diff
@@ -39,6 +39,27 @@ cdef class Memory:
         return self.ptr
 
 
+cdef set _peer_access = set()
+
+
+cpdef _set_peer_access(int device, int peer):
+    global _peer_access
```
okuta: Please remove.

beam2d:
Fixed. I also changed the _peer_access flag behavior: it now stops checking the peer access availability after the first call even if the peer access is unavailable.

okuta:
LGTM!

PR #888, opened 2016-01-26
Add ignore_label to binary_accuracy

t-abe:
Added `ignore_label` to `binary_accuracy`.
`sigmoid_cross_entropy` already have it, and it should be also considered when computing accuracy.

### DIFF
--- a/chainer/functions/evaluation/binary_accuracy.py
+++ b/chainer/functions/evaluation/binary_accuracy.py
@@ -6,6 +6,9 @@ from chainer.utils import type_check
 
 
 class BinaryAccuracy(function.Function):
+
+    ignore_label = -1
+
     def check_type_forward(self, in_types):
         type_check.expect(in_types.size() == 2)
         x_type, t_type = in_types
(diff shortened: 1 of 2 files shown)


chainer/functions/evaluation/binary_accuracy.py
```diff
@@ -23,7 +26,10 @@ def forward(self, inputs):
         y = y.ravel()
         t = t.ravel()
         c = (y >= 0)
-        return xp.asarray((c == t).mean(dtype='f')),
+        count = (t != self.ignore_label).sum()
+        if int(count) == 0:
```
okuta: Please use `xp.max`

t-abe: I do not understand your comment.
Do you mean I should use `xp.maximum(count)` instead of `int(count)`?

okuta: How about `count = xp.maximum(1, (t != self.ignore_label).sum())`?
`int(count)` cause synchronization of GPU and CPU.

t-abe: OK! I got it.

okuta:
LGTM except comment!

t-abe:
Updated.

okuta:
LGTM!

PR #891, opened 2016-01-27
Improve softmax_cross_entropy

### LINKED ISSUE #887: Remove sync point of softmax_cross_entropy
softmax_cross_entropy synchronizes the host and the device when `normalize` is True (https://github.com/pfnet/chainer/blob/master/chainer/functions/loss/softmax_cross_entropy.py#L99). This synchronization should be removed to improve the performance. It can be done by including the conditional branch into the post-map part of the crossent_fwd kernel.

okuta:
fix #887.

### DIFF
--- a/chainer/functions/loss/softmax_cross_entropy.py
+++ b/chainer/functions/loss/softmax_cross_entropy.py
@@ -81,11 +81,11 @@ class SoftmaxCrossEntropy(function.Function):
             count = (t != self.ignore_label).sum()
         else:
             count = len(x)
-        count = numpy.maximum(count, 1)
-        self._inv_count = 1.0 / count
+        count = max(count, 1)
+        self._coeff = 1.0 / count
 
         y = (log_p * (t.ravel() != self.ignore_label)).sum(keepdims=True) \
-            * (-self._inv_count)
+            * (-self._coeff)
         return y.reshape(()),
 
     def forward_gpu(self, inputs):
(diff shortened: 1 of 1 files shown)


chainer/functions/loss/softmax_cross_entropy.py
```diff
@@ -74,20 +74,18 @@ def forward_cpu(self, inputs):
...
         if getattr(self, 'normalize', True):
             count = (t != self.ignore_label).sum()
         else:
-            count = x.shape[0]
-        self.count = count
+            count = len(x)
+        count = numpy.maximum(count, 1)
+        self._inv_count = 1.0 / count
 
-        if count == 0:
```
delta2323: Why `count == 0` case is removed?

chainer/functions/loss/softmax_cross_entropy.py
```diff
@@ -96,29 +94,24 @@ def forward_gpu(self, inputs):
...
-            return cupy.zeros((), dtype=x.dtype),
+            count = len(t)
+        count = cupy.maximum(count, 1)
+        self._inv_count = cupy.divide(1.0, count, dtype=x.dtype)
 
         log_y = cupy.rollaxis(log_y, 1, log_y.ndim)
         ret = cuda.reduce(
-            'S t, raw T log_y, int32 n_channel, T inv_count', 'T out',
+            'S t, raw T log_y, int32 n_channel, raw T inv_count', 'T out',
```
delta2323: Why `raw` argument specifier is used here?

chainer/functions/loss/softmax_cross_entropy.py
```diff
@@ -96,29 +94,24 @@ def forward_gpu(self, inputs):
...
         self.y = cupy.exp(log_y)
         if getattr(self, 'normalize', True):
-            count = float((t != self.ignore_label).sum())
+            count = (t != self.ignore_label).sum()
         else:
-            count = t.shape[0]
-        self.count = count
-
-        if count == 0:
-            return cupy.zeros((), dtype=x.dtype),
```
delta2323: I think we should treat `count == 0` as a special case for readability, if we can ignore the difference of computing complexity between the creation of zero-filled array and reduction kernel.

The reason for it is the interpretation of `count`. In usual case, `count` is the number of labels that are not ignored. But if all labels are ignored and `self.normalize` is `True`, `count` is 1, although I think we can expect is `0`. So by treating the latter case as a special one, we make the interpretation of `count` more complex.

okuta: I will change variable name.

delta2323:
LGTM :+1:

PR #892, opened 2016-01-28
Fix documents of check_backward

delta2323:
Docstring in `check_backward` was broken. This PR fixes to it.

Related to #874


unnonouno:
LGTM

PR #901, opened 2016-02-02
Improve matmul

okuta:


chainer/functions/math/matmul.py
```diff
@@ -181,11 +175,9 @@ def __init__(self, transa=False, transb=False):
         self.transb = transb
 
     def _output_shape(self, a, b):
-        batch_size = a.shape[0]
-        a_mat_shape = _as_mat(a[0]).shape
-        b_mat_shape = _as_mat(b[0]).shape
-        m = a_mat_shape[1 if self.transa else 0]
-        n = b_mat_shape[0 if self.transb else 1]
+        batch_size = len(a)
+        m = _as_batch_mat(a).shape[2 if self.transa else 1]
```
delta2323: We do not have to do reshape `a` itself. We just need the shape of reshaped `a`.

chainer/functions/math/matmul.py
```diff
@@ -181,11 +175,9 @@ def __init__(self, transa=False, transb=False):
...
+        n = _as_batch_mat(b).shape[1 if self.transb else 2]
```
delta2323: ditto

delta2323:
LGTM except comments.

delta2323:
LGTM :+1:

PR #903, opened 2016-02-02
Fix Stream and Event nullptr value

### LINKED ISSUE #793: Stream object should have .ptr set to 0, not None.
The event object expects the stream.ptr to be an integer (size_t) here:
https://github.com/pfnet/chainer/blob/master/cupy/cuda/stream.py#L56
https://github.com/pfnet/chainer/blob/master/cupy/cuda/runtime.pyx#L309

In trunk at the moment, recording events with default stream fails via:
  File "train_imagenet.py", line 85, in <module>
    train_loop()
  File "train_imagenet.py", line 67, in train_loop
    start.record()
  File "/home/awesomebox/anaconda/lib/python2.7/site-packages/chainer-1.5.1-py2.7-linux-x86_64.egg/cupy/cuda/stream.py", line 56, in record
    runtime.eventRecord(self.ptr, stream.ptr)
  File "cupy/cuda/runtime.pyx", line 309, in cupy.cuda.runtime.eventRecord (cupy/cuda/runtime.cpp:6139)
TypeError: an integer is required

The fix seems simple:

https://github.com/pfnet/chainer/blob/master/cupy/cuda/stream.py#L103
self.ptr = 0

beam2d:
Fix #793. It still lacks tests (as is).

### DIFF
--- a/cupy/cuda/stream.py
+++ b/cupy/cuda/stream.py
@@ -22,7 +22,7 @@ class Event(object):
 
     """
     def __init__(self, block=False, disable_timing=False, interprocess=False):
-        self.ptr = None
+        self.ptr = 0
 
         if interprocess and not disable_timing:
             raise ValueError('Timing must be disabled for interprocess events')
@@ -100,7 +100,7 @@ class Stream(object):
     """
     def __init__(self, null=False, non_blocking=False):
         if null:
-            self.ptr = None
+            self.ptr = 0
         elif non_blocking:
             self.ptr = runtime.streamCreateWithFlags(runtime.streamNonBlocking)
         else:
@@ -109,7 +109,7 @@ class Stream(object):
     def __del__(self):
         if self.ptr:
             runtime.streamDestroy(self.ptr)
-            self.ptr = None
+            self.ptr = 0
 
     @property
     def done(self):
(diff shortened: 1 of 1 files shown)


cupy/cuda/stream.py
```diff
@@ -22,7 +22,7 @@ class Event(object):
 
     """
     def __init__(self, block=False, disable_timing=False, interprocess=False):
-        self.ptr = None
+        self.ptr = 0
```
delta2323: `self.ptr` is set at the last line of `__init__`.  So I guess it is not necessary to initialize `self.ptr` here.

cupy/cuda/stream.py
```diff
@@ -109,7 +109,7 @@ def __init__(self, null=False, non_blocking=False):
     def __del__(self):
         if self.ptr:
             runtime.streamDestroy(self.ptr)
-            self.ptr = None
+            self.ptr = 0
```
delta2323: Do we have to set `self.ptr` to 0 explicitly even if `Stream` object itself will be deleted at the same time?

cupy/cuda/stream.py
beam2d: It is indeed required for the case that __init__ raises an exception before setting the ptr. If we omit the first line, __del__ tries to access ptr, and raises an AttributeError. This is problematic.

delta2323: OK, I understand it.

delta2323:
LGTM :+1:

PR #904, opened 2016-02-02
Fix check_cuda_version and check_cudnn_version and add their tests

### LINKED ISSUE #880: check_cuda_version and check_cudnn_version failed although compile and link are successful.
`check_cuda_version` and `check_cudnn_version` check the version of CUDA and cuDNN respectively by making a temporary cpp file, compiling and linking it.

In some environment, the compiler used is gcc although the extension is `.cpp`. As checking code does not  return exit code explicitly, the return code can be other than 0 although all procedures are successful. If g++ was used, this problem would not occur because the code return 0 by default.

delta2323:
* Fix test codes in `check_cuda_version` and `check_cudnn_version` to add `return 0;`
* Move these functions to `install/build.py`.
* Move `print_warning` function to `install/utils.py`.
* Add small tests for these functions.

To enable these tests, we need to add commands that invokes these tests to chainer-test.

This PR fixes #880


delta2323:
As the change of the directory structure of build scripts take so much time to review, we decided not to include this change in v1.6.0. We will decide we will merge it in the future release later. I extracted minimum changes and send PR as #908 .

install/test_build.py
```diff
@@ -0,0 +1,20 @@
+import distutils
+import setuptools  # NOQA
+import unittest
+
+import build
```
beam2d: Add `from install`.

install/test_utils.py
```diff
@@ -0,0 +1,9 @@
+import unittest
+
+import utils
```
beam2d: Add `from install`.

beam2d:
What is the intended way to test these modules? `nosetests install` at the repository root does not work on Python3 (my second comment). With the current code, `nosetests` inside of `install` directory works, while once you make a change for the second comment (i.e. add `from install` to the import statement), it is then not allowed. It might be good to add an instruction somewhere for the test of install script (additional readme in the `install` directory, the contribution guide, or both).

delta2323:
@beam2d Thank you for your comments. I did not think much about Py3.
I move testing files to `tests` directory, resulting in the same directory structure as Chainer and CuPy. We have a restriction that we cannot execute tests for installation modules by `nosetests` other than from the root directory. But I think it is OK because root directory is the only place where the contribution guide advises to invoke `nosetests`.

I also update Contribution Guide.

beam2d:
OK. Running `nosetests` at the repository root for testing installation utilities is acceptable, though we should have to think about running `nosetests` under the `tests` directory for testing Chainer itself, which might be done for checking the success of installation.

install/build.py
```diff
@@ -0,0 +1,96 @@
+import distutils
+import os
+import setuptools  # NOQA
+import shutil
+import subprocess
+import tempfile
+
+import utils
```
beam2d: Fix this, too.

install/build.py
```diff
@@ -0,0 +1,96 @@
+import distutils
+import os
+import setuptools  # NOQA
```
beam2d: Move it to a new line between tempfile and utils, and remove NOQA.

tests/install_tests/test_build.py
```diff
@@ -0,0 +1,20 @@
+import distutils
+import setuptools  # NOQA
```
beam2d: ditto: move it to a new line between unittest and build, and remove NOQA.

beam2d:
Would you merge master? (do not miss this fix: https://github.com/pfnet/chainer/commit/47344b0652d6e62f6da7219e73e979efdd103bb1)

beam2d:
LGTM!

PR #908, opened 2016-02-03
Add return value to test code of check_cuda_version and check_cudnn_version

### LINKED ISSUE #880: check_cuda_version and check_cudnn_version failed although compile and link are successful.

delta2323:
Fix #880


okuta:
LGTM!

PR #917, opened 2016-02-05
Reduce memory usage

okuta:


chainer/variable.py
```diff
@@ -308,6 +308,7 @@ def add_cand(cand):
                             need_copy.remove(id_x)
                         else:  # 3rd or later visit
                             x._grad += gx
+            del gxs  # for reduce memory usage
```
delta2323: The comment is not grammatically correct. "to reduce memory usage" is more appropriate.

delta2323:
LGTM :+1:

PR #923, opened 2016-02-05
Implement asfortranarray

okuta:


tests/cupy_tests/manipulation_tests/test_kind.py
```diff
@@ -7,3 +7,15 @@
 class TestKind(unittest.TestCase):
 
     _multiprocess_can_split_ = True
+
+    @testing.for_all_dtypes()
+    @testing.numpy_cupy_array_equal(type_check=False)
+    def test_asfortranarray1(self, xp, dtype):
+        x = xp.zeros(2, 3)
```
delta2323: It should be `xp.zeros((2, 3))`. So this caused an error, but it is caught and grasped [here](https://github.com/pfnet/chainer/blob/master/cupy/testing/helper.py#L24).

cupy/manipulation/kind.py
```diff
@@ -1,11 +1,26 @@
...
+        dtype (str or dtype object, optional): By default, the data-type is
+            inferred from the input data.
+
+    Returns:
+        ~cupy.ndarray: The input `a` in Fortran, or column-major, order.
+
+    .. seealso:: :func:`numpy.asfortranarray`
+
+    """
+    ret = cupy.empty(a.shape.T, a.dtype if dtype is None else dtype)
```
delta2323: `tuple` does not have an attribute `T`.

delta2323: What happens if we feed `a` with a fortran array?

tests/cupy_tests/manipulation_tests/test_kind.py
okuta: Thank you! 
I will fix.

okuta:
@delta2323 I fixed.

tests/cupy_tests/manipulation_tests/test_kind.py
```diff
@@ -7,3 +7,21 @@
 class TestKind(unittest.TestCase):
 
     _multiprocess_can_split_ = True
+
+    @testing.for_all_dtypes()
+    @testing.numpy_cupy_array_equal(type_check=False)
```
delta2323: Why `type_check` is disabled?

delta2323:
Please add tests in which `ndarray.flags` are set correctly after the operation.

delta2323:
@okuta I updated comments.

tests/cupy_tests/manipulation_tests/test_kind.py
okuta: This test checks the stride. It is a tuple.

delta2323:
LGTM :+1:

PR #940, opened 2016-02-12
HuberLoss: Add loss function for robust regression (new function)

wkentaro:


delta2323:
I think `SmoothL1Loss` does not need to accept two inputs and calculate their difference. It is simpler that we calculate the difference outside the function and feed it to the function.

delta2323:
Smooth L1 Loss is the special case of [Huber Loss](https://en.wikipedia.org/wiki/Huber_loss) where delta equals to 1. So I think it is beneficial that we introduce the parameter `delta`.

beam2d:
As Oono-san wrote, Huber loss is a generalization of the smooth L1 loss, and the smooth_l1_loss can be written with this loss function with negligible overhead. Could you modify the current code to implement it? For example, replace SmoothL1Loss by HuberLoss, add huber_loss, and rewrite smooth_l1_loss to use huber_loss.

wkentaro:
I have updated the function to support `delta` argument, and implemented the HuberLoss function.

chainer/functions/loss/huber_loss.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def forward(self, inputs):
+        xp = cuda.get_array_module(*inputs)
+        x0, x1 = inputs
+        self.diff = x0 - x1
+        flag = xp.abs(self.diff) <= self.delta
+        y = (flag * 0.5 * xp.square(self.diff) +
+             (~flag) * self.delta * (xp.abs(self.diff) - 0.5 * self.delta))
+        y = y.astype(numpy.float32)
+        return y.sum(axis=1),
```
beam2d: It would be slightly faster.
```python
y = xp.square(self.diff)
mask = y > (self.delta ** 2)
y -= mask * xp.square(abs(self.diff) - self.delta)
y /= 2
return y.sum(axis=1),
```
It would be great if you have a time to write a custom kernel for GPU, which should be much faster. Otherwise, though, I will merge this device-agnostic implementation (since it would be enough for most use cases).

chainer/functions/loss/huber_loss.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def backward(self, inputs, gy):
+        xp = cuda.get_array_module(*inputs)
+        flag = xp.abs(self.diff) <= self.delta
+        gx = flag * self.diff + ~flag * self.delta * xp.sign(self.diff)
+        gx = gx.astype(numpy.float32)
```
beam2d: It would be made slightly faster.
```python
flag = abs(self.diff) <= sef.delta
gx = xp.where(flag, self.diff, self.delta * xp.sign(self.diff))
```

chainer/functions/loss/huber_loss.py
```diff
@@ -0,0 +1,40 @@
...
+        return gx, -gx
+
+
+def huber_loss(x, t, delta):
```
beam2d: Would you write a docstring and add a link to it from the reference? If you don't have enough time, I will make one.

beam2d:
Sorry for being late. I added some comments.

chainer/functions/loss/huber_loss.py
wkentaro: > It would be made slightly faster.
y = xp.square(self.diff)
mask = y > (self.delta ** 2)
y -= mask * xp.square(abs(self.diff) - self.delta)
y *= 0.5
return y.sum(axis=1),

Is this correct? It looks different from the definition. https://en.wikipedia.org/wiki/Huber_loss
It fails on the test previously created.

chainer/functions/loss/huber_loss.py
wkentaro: Ok, I will.

chainer/functions/loss/huber_loss.py
wkentaro: > It would be great if you have a time to write a custom kernel for GPU, which should be much faster. Otherwise, though, I will merge this device-agnostic implementation (since it would be enough for most use cases).

I'd like to work for that when it become necessary.

wkentaro:
@beam2d Thank you for your reviews and comments. I have updated the commits, and commented above. Please browse them.

chainer/functions/loss/huber_loss.py
beam2d: > Is this correct? It looks different from the definition. https://en.wikipedia.org/wiki/Huber_loss
> It fails on the test previously created.

Oh, really? It passed the tests in my environment.
The Huber Loss definition can be transformed into the following equation:
```
L_delta(a) = 0.5 * a^2  (if a^2 <= delta^2),
             0.5 * (a^2 - (|a| - delta)^2)  (otherwise)
```
and my implementation uses it. Note that `self.diff = x0 - x1` must be left as is (I omitted it in my previous comment).

beam2d:
@wkentaro Thank you for reflecting my comments! I added a comment.

chainer/functions/loss/huber_loss.py
wkentaro: Sorry, it seems my mistake. It works.

wkentaro:
@beam2d I have updated the commits. Could you please review again?

beam2d:
Thank you for quick response! I will review.

chainer/functions/loss/huber_loss.py
```diff
@@ -0,0 +1,69 @@
...
+    def backward(self, inputs, gy):
+        xp = cuda.get_array_module(*inputs)
+        mask = xp.abs(self.diff) <= self.delta
+        gx = xp.where(mask, self.diff, self.delta * xp.sign(self.diff))
+        return gx, -gx
+
+
+def huber_loss(x, t, delta):
+    """Loss function used in robust regression,
+    that is less sensitive to outliers in data than the squared error loss.
```
beam2d: Would you shorten the head line description to one line?
http://docs.openstack.org/developer/hacking/#docstrings See the example here.

wkentaro: I have updated the line.

tests/chainer_tests/functions_tests/loss_tests/test_huber_loss.py
```diff
@@ -0,0 +1,61 @@
...
+        self.assertEqual(loss.data.dtype, numpy.float32)
+        loss_value = cuda.to_cpu(loss.data)
+
+        diff_data = cuda.to_cpu(x_data) - cuda.to_cpu(t_data)
+        expected_result = numpy.zeros(self.shape)
+        mask = numpy.abs(diff_data) < 1
+        expected_result[mask] = 0.5 * diff_data[mask]**2
+        expected_result[~mask] = numpy.abs(diff_data[~mask]) - 0.5
+        loss_expect = numpy.sum(expected_result, axis=1)
+        numpy.testing.assert_allclose(loss_value, loss_expect)
```
beam2d: Would you relax the threshold of testing? It seems be failed sometimes. Just using gradient_check.assert_allclose is also good, since it uses more loose thresholds by default.

beam2d:
Added a few comments. LGTM except them!

wkentaro:
Thanks! I have added a commit.

beam2d:
Merged. Thank you!

wkentaro:
Thanks!

PR #953, opened 2016-02-17
Skip cythonizing with no changes

wkentaro:
To make it developer friendly.

### DIFF
--- a/chainer_setup_build.py
+++ b/chainer_setup_build.py
@@ -194,6 +194,10 @@ def parse_args():
     if check_readthedocs_environment():
         _arg_options['no_cuda'] = True
 
+    _arg_options['only_changed'] = '--cupy-only-changed' in sys.argv
+    if _arg_options['only_changed']:
+        sys.argv.remove('--cupy-only-changed')
+
 
 def get_cython_pkg():
     try:
@@ -243,8 +247,8 @@ def _is_changed(filename):
     return md5_cached != md5_new.encode('utf-8')
 
 
-def cythonize(
-        extensions, force=False, annotate=False, compiler_directives=None):
+def cythonize(extensions, force=False, annotate=False,
+              compiler_directives=None only_changed=False):
     cython_location = get_cython_pkg().location
     cython_path = path.join(cython_location, 'cython.py')
     print("cython path:%s" % cython_location)
@@ -258,7 +262,7 @@ def cythonize(
             cmd.append('%s=%s' % i)
 
     for ext in extensions:
-        if all(not _is_changed(f) for f in ext.sources):
+        if all(not _is_changed(f) for f in ext.sources) and only_changed:
             print("skipping because no changes: '{}'"
                   .format(','.join(ext.sources)))
             continue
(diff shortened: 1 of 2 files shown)


okuta:
I recommend using the `ccache`. You can reduce the compile time safety.

wkentaro:
Thanks for the comment, but skipping compilation is still much more faster, isn't it?

On OS X Yosemite, ccache setup
```
brew install ccache
sudo ln -s /usr/local/bin/ccache /usr/local/bin/clang
sudo ln -s /usr/local/bin/ccache /usr/local/bin/clang++
```


1 without ccache
```
Using /usr/local/lib/python2.7/site-packages
Finished processing dependencies for chainer==1.6.2
python setup.py develop  33.12s user 2.20s system 97% cpu 36.118 total
```

2 with ccache
```
Using /usr/local/lib/python2.7/site-packages
Finished processing dependencies for chainer==1.6.2
python setup.py develop  8.25s user 1.43s system 97% cpu 9.967 total
```

3 skipping with no change
```
Using /usr/local/lib/python2.7/site-packages
Finished processing dependencies for chainer==1.6.2
python setup.py develop  0.82s user 0.44s system 95% cpu 1.330 total
```

okuta:
Thank you for the presentation of evidence.
It is necessary to check environment variables and header files that in order to determine the skipping compilation.
We need a more safety method.
For example, to control this function by command line flag.

wkentaro:
> It is necessary to check environment variables and header files that in order to determine the skipping compilation.
We need a more safety method.

I see, that's reasonable. I added a command line option `--cupy-only-changed` to skip cythonizing if there're no changes.

chainer_setup_build.py
```diff
@@ -388,6 +421,10 @@ def cythonize(extensions, force=False, annotate=False, compiler_directives={}):
 
     cython_cmdbase.extend(['--fast-fail', '--verbose', '--cplus'])
     for ext in extensions:
+        if all(not _is_changed(f) for f in ext.sources) and only_changed:
```
okuta: How about `if only_changed and all(not _is_changed(f) for f in ext.sources):`?

wkentaro: I need to call `_is_changed` function to update the `*.pyx.md5` file.

wkentaro: I added a comment to describe that, looks good?

okuta: I think that the calculation of md5 is not necessary for most users.
So, it is better that extra file is not made.
If you want to check the md5 of all files, following code is good.
`if all([not _is_changed(f) for f in ext.sources]) and only_changed:`

wkentaro: What's the difference between 
`if all(not _is_changed(f) for f in ext.sources) and only_changed:`
and
`if all([not _is_changed(f) for f in ext.sources]) and only_changed:`
?

okuta: `not _is_changed(f) for f in ext.sources` return the generator.
And, `all` method performs short-circuit evaluation.
`[not _is_changed(f) for f in ext.sources]` return the list.
`_is_changed` is called for all elements of `ext.sources`.

wkentaro: Ah, I understood. Thanks.

wkentaro: I updated the commit.

okuta:
I think that the calculation of md5 is not necessary for most users.
So, it is better that extra file is not made.

wkentaro:
Ok. In that case, maybe there is no way to fix this issue.
And at the current master branch, the speed is fast enough, so this PR can be closed.

PR #972, opened 2016-02-22
Fix batch_matmul to avoid ZeroDivisionError in GPU mode

kiyukuta:
The `batch_matmul` (in GPU mode) uses  `strides[0]` of an input array as a divisor in `_mat_ptrs` (i.e. the third argument of `cuda.cupy.arange`).
https://github.com/pfnet/chainer/blob/master/chainer/functions/math/matmul.py#L23 

And strides of ndarray will be 0 when it is a broadcasted array because of rule 4 of the ufunc’s broadcasting rules.
http://docs.scipy.org/doc/numpy-1.10.0/reference/ufuncs.html

So, ZeroDivisionError will happen.
```python
A = chainer.Variable(xp.random.random((3, 2, 4), dtype=np.float32))
B = chainer.Variable(xp.random.random((4, 6), dtype=np.float32))
B = F.expand_dims(B, 0)
B = F.broadcast_to(B, (3, 4, 6))
print(B.data.shape, B.data.strides)     # —> (3, 4, 6) (0, 48, 8)

C = F.batch_metmul(A, B)  # ZeroDivisionError (only on GPU mode)
```

We can fix the strides by simply copying the ndarray.

```python
b = xp.random.random((4, 6))
b = xp.expand_dims(b, 0)
b = xp.broadcast_to(b, (3,4,6))
print(b.shape, b.strides)   # —> (3, 4, 6) (0, 48, 8) 
_b = b.copy()
print(_b.shape, _b.strides) # —> (3, 4, 6) (192, 48, 8) 
```

### DIFF
--- a/chainer/functions/math/matmul.py
+++ b/chainer/functions/math/matmul.py
@@ -55,8 +55,8 @@ def _get_ld(a):
 
 
 def _batch_matmul_gpu(a, b, out, transa=False, transb=False, transout=False):
-    a = _as_batch_mat(a.copy())
-    b = _as_batch_mat(b.copy())
+    a = _as_batch_mat(cupy.ascontiguousarray(a))
+    b = _as_batch_mat(cupy.ascontiguousarray(b))
     trans_axis = (0, 2, 1)
     if transout:
         out = out.transpose(trans_axis)
--- a/chainer/functions/math/matmul.py
+++ b/chainer/functions/math/matmul.py
@@ -55,8 +55,8 @@ def _get_ld(a):
 
 
 def _batch_matmul_gpu(a, b, out, transa=False, transb=False, transout=False):
-    a = _as_batch_mat(a)
-    b = _as_batch_mat(b)
+    a = _as_batch_mat(a.copy())
+    b = _as_batch_mat(b.copy())
     trans_axis = (0, 2, 1)
     if transout:
         out = out.transpose(trans_axis)
--- a/chainer/functions/math/matmul.py
+++ b/chainer/functions/math/matmul.py
@@ -55,8 +55,8 @@ def _get_ld(a):
 
 
 def _batch_matmul_gpu(a, b, out, transa=False, transb=False, transout=False):
-    a = _as_batch_mat(cupy.ascontiguousarray(a))
-    b = _as_batch_mat(cupy.ascontiguousarray(b))
+    a = _as_batch_mat(cuda.cupy.ascontiguousarray(a))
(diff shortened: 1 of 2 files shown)


kiyukuta:
By the way, I don’t know hot to write a test code for this case.
First I tried the following code but it causes error because the numpy’s `broadcast_to` returns read-only array.

```python
class TestBatchMatMulBroadcastedMatrix(_TestMatMul):

    def setUp(self):
        self.x1 = numpy.random.uniform(
            .5, 1, (batch_size, m, k)).astype(numpy.float32)
        x2 = numpy.random.uniform(.5, 1, (k, n)).astype(numpy.float32)
        self.x2 = numpy.broadcast_to(
            numpy.expand_dims(x2, 0), (batch_size, k, n))
        self.gy = numpy.random.uniform(
            -1, 1, (batch_size, m, n)).astype(numpy.float32)
        self.op = lambda x, y: F.batch_matmul(x, y)
        self.forward_answer = numpy.array([
            numpy.dot(self.x1[i], self.x2[i])
            for i in six.moves.range(batch_size)])
```

fabiencro:
I actually ran into this issue as well. 

I think systematically copying the data as you propose would create some performance issues. It should be possible to solve this problem without the copying. Actually, for such cases, I have a workaround by using F.matmul (the non-batch version) and F.reshape

In the case of your first example, the intended behaviour can be obtained this way, without copying (did not try to run this code, but I think it should work...) 
```
A = chainer.Variable(xp.random.random((3, 2, 4), dtype=np.float32))
B = chainer.Variable(xp.random.random((4, 6), dtype=np.float32))

A_reshaped = F.reshape(A, (6,4))
C_reshaped = F.matmul(A_reshaped, B)
C = F.reshape(C_reshaped, (3,2,6))
```

Still, it would be good if batch_matmul could do the right thing when there is some broadcast happening.

chainer/functions/math/matmul.py
```diff
@@ -55,8 +55,8 @@ def _get_ld(a):
 
 
 def _batch_matmul_gpu(a, b, out, transa=False, transb=False, transout=False):
-    a = _as_batch_mat(a)
```
okuta: How about using `cupy.ascontiguousarray`?

okuta:
Please add tests.

kiyukuta:
Sorry for my late response.

@fabiencro Oh, I didn’t notice that !!
It’s a nice solution, thanks for your suggestion :)

@okuta I pushed some changes, please check it!

okuta:
Thank you!
LGTM!

PR #1002, opened 2016-03-02
Add PrintHook

delta2323:


chainer/function_hooks/debug_print.py
```diff
@@ -0,0 +1,54 @@
...
+
+import chainer
+from chainer import cuda
+from chainer import function
+
+
+class PrintHook(function.FunctionHook):
+    """Function hook that prints debug information.
+
+    Attributes are same as the keyword argument of print function.
```
beam2d: Explain what this hook actually does, and how/when to use. It is also better to compare with just using print function in the forward computation (capability on printing at backward, and printing every function within the `with` statement, right?).

chainer/function_hooks/debug_print.py
```diff
@@ -0,0 +1,54 @@
...
+        self._print('input data')
+        for d in in_data:
+            self._print(chainer.Variable(d).debug_print())
+        if out_grad is not None:
+            self._print('output gradient')
+            for d in out_grad:
+                xp = cuda.get_array_module(d)
+                v = chainer.Variable(xp.empty_like(d, dtype=d.dtype))
+                v.grad = d
+                self._print(v.debug_print())
```
beam2d: Maybe better to flush the stream here.

beam2d:
LGTM except two comments.

beam2d:
Oh, I found that it is not added to the reference.

chainer/function_hooks/debug_print.py
```diff
@@ -0,0 +1,77 @@
...
+            at the end of preprocessing.
+    """
+
+    name = 'PrintHook'
+
+    def __init__(self, sep='', end='\n', file=sys.stdout, flush=True):
+        self.sep = sep
+        self.end = end
+        self.file = file
+        self.flush = True
```
beam2d: rhs should be flush?

PR #1004, opened 2016-03-02
Add profiling example

delta2323:


chainer/function_hooks/timer.py
```diff
@@ -6,56 +6,105 @@
...
+
+    def reset(self):
+        self.total_time = 0.0
+        self.last_increment = None
+        self.count = 0
+
+    def start(self):
+        if self.running:
+            return
+        if self.xp == numpy:
```
beam2d: Use `is` to compare modules. It is cheaper.

chainer/function_hooks/timer.py
```diff
@@ -6,56 +6,105 @@
...
+        else:
+            self._start = cuda.Event()
+            self._start.record()
+        self.running = True
+
+    def stop(self):
+        if not self.running:
+            return 0.0
+
+        if self.xp == numpy:
```
beam2d: ditto

chainer/function_hooks/timer.py
```diff
@@ -6,56 +6,105 @@
...
+        self.running = False
+        self.total_time += elapsed_time
+        self.last_increment = elapsed_time
+        self.count += 1
+
+        return elapsed_time
+
+    def mean(self):
+        if self.count == 0:
+            raise ValueError('count is 0')
```
beam2d: Should be more informative message.

examples/profile/profile.py
```diff
@@ -0,0 +1,157 @@
...
+                    help='random seed')
+parser.add_argument('--iteration', '-i', type=int, default=10,
+                    help='the number of iterations to be averaged over.')
+parser.add_argument('--gpu', '-g', type=int, default=-1,
+                    help='GPU to use. Negative value to use CPU')
+parser.add_argument('--cudnn', '-c', action='store_true',
+                    help='True if using cudnn')
+parser.add_argument('--cache-level', '-C', type=str, default=None,
+                    choices=(None, 'memory', 'disk'),
+                    help='If `None`, memory cache and disk cache are removed '
```
beam2d: Can we pass None value explicitly? If not, you should write the help message without using this word, like "by default, ..." (since the default values are not printed in the help message).

chainer/function_hooks/timer.py
```diff
@@ -6,56 +6,105 @@
...
+    def stop(self):
+        if not self.running:
+            return 0.0
+
+        if self.xp == numpy:
+            self._stop = time.time()
+            elapsed_time = self._stop - self._start
+        else:
+            self._stop = cuda.Event()
+            self._stop.record()
```
beam2d: Is this record run immediately here? I mean, I want to know whether there is a possibility that the elapsed time is measured from the first kernel launched after this record query, or just from the timing of this record query.

examples/profile/profile.py
```diff
@@ -0,0 +1,157 @@
...
+    data = chainer.Variable(xp.asarray(data))
+    label = numpy.empty((args.batchsize,), dtype=numpy.int32)
+    label.fill(0)
+    label = chainer.Variable(xp.asarray(label))
+
+    # forward
+    with function_hooks.TimerHook() as h:
+        loss = model(data, label)
+    forward_time_one = h.total_time()
+
```
beam2d: Don't we have to wait for all kernels finish here?

delta2323: Just for confirmation. Are you thinking of the case where `Timer.xp` eqauls to `numpy`?

chainer/function_hooks/timer.py
delta2323: I think I cannot fully understand your question. Is your question that if the kernel is launched after this record query, we cannot include this kernel to the elapse time?

beam2d: I wanted to know whether the execution time of Python code below is included or not.

examples/profile/profile.py
beam2d: OK, I found that the timer hook actually synchronizes with the CUDA streams, so it actually waits for all kernels to finish.

delta2323:
@beam2d  I came to realize that I need to think more the design of `Timer` module. I had sent PR tackling this issue as #1004. So please review it first.

delta2323:
We are working on clearing branches of this repository. As this PR is too old, I close this PR.

PR #1009, opened 2016-03-07
Needed to link with MSVC

samehkhamis:


okuta:
Please tell me why we need this PR?

samehkhamis:
Without this change for MSVC 12.0+, "import cupy" gives "Cannot import name core", similar to https://github.com/pfnet/chainer/issues/702 . cupy doesn't build because the setup fails to get the versions for cuda and cudnn (linking fails for the executables, mt error 31).
Linking with a version of MSVC different from the one that python/numpy was built with is messy. The easy solution for me was to include a manifest to add proper metadata to different versions of CRT functions. Cheers!

okuta:
Thank you!
LGTM!

PR #1049, opened 2016-03-22
Fix elementwise kernel's document

### LINKED ISSUE #1031: ElmentwiseKernel: "n" is undefined
http://docs.chainer.org/en/stable/cupy-reference/kernel.html
```
We can use the special variables n and i for the manual indexing. n indicates total number of elements to apply the elementwise operation. i indicates the index within the loop. For example, a kernel that adds two vectors with reversing one of them can be written as follows: ...
```

I tested follow sample code. I was not able to execute a code.
```python
import cupy
import numpy as np
x = cupy.arange(5, dtype=np.float32)
y = cupy.arange(5, dtype=np.float32)

add_reverse = cupy.ElementwiseKernel(
    'T x, raw T y', 'T z',
    'z = x + y[n - i - 1]',
    'add_reverse')

add_reverse(x, y)
```

```bash
stdout/stderr:
/tmp/tmpsaehXm/kern.cu(279): error: identifier "n" is undefined
```

delta2323:
Fix #1031


docs/source/cupy-reference/kernel.rst
```diff
@@ -108,24 +108,32 @@ The ElementwiseKernel class does the indexing with broadcasting automatically, w
...
 We can tell the ElementwiseKernel class to use manual indexing by adding the ``raw`` keyword preceding the type specifier.
 
-We can use the special variables ``n`` and ``i`` for the manual indexing.
-``n`` indicates total number of elements to apply the elementwise operation.
+We can use the special variable ``i`` and ``_ind`` for the manual indexing.
 ``i`` indicates the index within the loop.
+``_ind`` holds the infromation about the shape of the underlying array.
+
+* ``_ind.size()`` returns the number of elements to apply the elementwise operation.
+* ``_ind.shape`` is the array that represents the shape of the array.
```
okuta: Is this member visible?
`shape_` is private member.
https://github.com/pfnet/chainer/blob/master/cupy/core/carray.cuh#L197

okuta:
LGTM!

PR #1051, opened 2016-03-23
Add function_hooks to setup.py

### LINKED ISSUE #1050: function_hook is missing in the package
As function_hook is missing in `setup.py`, it is not installed when we install chainer through `pip`.

delta2323:
Fix #1050


okuta:
LGTM!

PR #1054, opened 2016-03-23
Add functions and links to TOC

### LINKED ISSUE #946: Suggestion: reorganize navigation menu in the document (show Links/Functions etc.)
Hi, I am new to Chainer, I think the concept is great.  

Just a little bit confused when reading the online document of chainer. 

As a regular user, what I want to know most is what kind of layers (Links) and functions is available in Chainer. But currently there is no such list available in the document, instead there are classified menus such as "learnable connections" and "loss functions" which is not intuitive for me. 

It will be surely helpful if you put all the links and functions right inside the navigation, and try to remove those redundant upper layers. Since the references are heavily used during the entire usage of chainer, it's really important to make a good navigation without too deep wrapping. (Tutorials are only for beginners, and CuPy is more for developers, thus they are not so heavily used.)

A good example for me is http://keras.io/layers/core/

What do you think?

beam2d:
Fix #946.

In this PR, I added a heading to each function/link in the document. It makes sphinx add these items to the side menu and users can easily look through all functions/links.

NOTE: I couldn't find a reasonable way to add arbitrary items to the side menu. This solution makes the document somewhat redundant (i.e. the name of each function/class must be written twice), though the readability is (maybe) not lowered.

The updated document can be found [here](http://docs.chainer.org/en/reference-toc-catalog/).

### DIFF
--- a/docs/source/reference/functions.rst
+++ b/docs/source/reference/functions.rst
@@ -23,95 +23,291 @@ functions, which users should use.
 
    KEEP EACH LIST IN LEXICOGRAPHICAL ORDER.
 
+
 Activation functions
 --------------------
+
+clipped_relu
+~~~~~~~~~~~~
 .. autofunction:: clipped_relu
+
+elu
+~~~
 .. autofunction:: elu
+
+leaky_relu
+~~~~~~~~~~
 .. autofunction:: leaky_relu
+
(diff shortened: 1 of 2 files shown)


okuta:
LGTM!

PR #1061, opened 2016-03-23
Use immutable object for default argument

okuta:


chainer_setup_build.py
```diff
@@ -375,11 +376,13 @@ def cythonize(extensions, force=False, annotate=False, compiler_directives={}):
     run_command(cython_cmdbase + ['--version'])
 
     cython_cmdbase.extend(['--fast-fail', '--verbose', '--cplus'])
-    for ext in extensions:
-        cmd = list(cython_cmdbase)
+    cmd = list(cython_cmdbase)
```
delta2323: `cython_cmdbase` seems a list from the outset. Why do we apply list function to it?

okuta:
I fixed.

delta2323:
LGTM :+1:

PR #1063, opened 2016-03-23
Fix code in array_split

okuta:


delta2323:
I think this is not a bug as this change does not change the behavior of this function.

okuta:
Thank you for your opinion.
I change title and label.

delta2323:
LGTM :+1:

PR #1068, opened 2016-03-24
Add support for SMORMS3 optimizer

benob:
This PR adds support for the SMORMS3 optimizer described here:
http://sifter.org/~simon/journal/20150420.html
This optimizer yields faster convergence on certain workloads. The cost is higher memory requirements (stores 3 parameters per weight). A CUDA kernel is provided as well.

### DIFF
--- a/chainer/optimizers/smorms3.py
+++ b/chainer/optimizers/smorms3.py
@@ -1,0 +1,52 @@
+import numpy
+
+from chainer import cuda
+from chainer import optimizer
+
+
+class SMORMS3(optimizer.GradientMethod):
+
+    """Simon Funk's SMORMS3.
+
+    See http://sifter.org/~simon/journal/20150420.html.
+
+    """
+
+    def __init__(self, lr=0.001, eps=1e-16):
+        self.lr = lr
+        self.eps = eps
+
+    def init_state(self, param, state):
+        xp = cuda.get_array_module(param.data)
+        state['mem'] = xp.ones_like(param.data)
+        state['g'] = xp.zeros_like(param.data)
+        state['g2'] = xp.zeros_like(param.data)
+
+    def update_one_cpu(self, param, state):
+        mem, g, g2 = state['mem'], state['g'], state['g2']
+        grad = param.grad
+
+        r = 1 / (mem + 1)
+        g = (1 - r) * g + r * grad
+        g2 = (1 - r) * g2 + r * grad * grad
+        x = g * g / (g2 + self.eps)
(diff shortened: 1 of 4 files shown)


chainer/optimizers/smorms3.py
```diff
@@ -0,0 +1,48 @@
...
+
+    """ Simon Funk's SMORMS3, see http://sifter.org/~simon/journal/20150420.html """
+
+    def __init__(self, lr=0.001, eps=1e-16):
+        self.lr = lr
+        self.eps = eps
+
+    def init_state(self, param, state):
+        xp = cuda.get_array_module(param.data)
+        state['mem'] = xp.zeros_like(param.data) + 1
```
okuta: How about using `ones_like`?

benob: Working on a version that matches chainer/master.

benob:
Squashed commits for cleaner history (maybe that was not a good idea).
Tagging with WIP because I could not test the GPU code.

benob:
I don't understand why the checks failed.

delta2323:
I do not know why CI does not work. Adding new commit could trigger the CI. As this branch has conflicts, how about adding the commit that merges or rebases master branch?

delta2323:
Thank you for your PR. Could you tell me about this optimizer as I do not know about it well? Specifically, is there a published paper for it and how much does it used?

benob:
SMORMS3 is described here: http://sifter.org/~simon/journal/20150420.html
I haven't seen it described in a proper paper. However, some people use it around the web and it is mentioned in a few papers. There is a theano implementation:
https://github.com/diogo149/treeano/blob/master/treeano/sandbox/nodes/smorms3.py

Feel free to include it or not in chainer. I figured it could be interesting to its users.

delta2323:
>I haven't seen it described in a proper paper. However, some people use it around the web and it is mentioned in a few papers. There is a theano implementation:

OK, I understand it. In my opinion, it is acceptable to add the optimizer.

delta2323:
Please register this optimizer to `docs/source/reference/optimizers.rst`.

chainer/optimizers/smorms3.py
delta2323: Could you apply this change?

benob:
Registered the optimizer and fixed the ones_like(...) line.

delta2323:
LGTM 👍

delta2323:
Thank you very much!

PR #1098, opened 2016-04-07
Fix decay rate in fine tuning of batch normalization

### LINKED ISSUE #1091: Bug in batch normalization when finetune = True?
Correct me if I'm wrong, but from my understanding, if the option `finetune` is `True`, we want to compute the population statistics of the training data so we can use this mean/var during testing.
In the current implementation, once the finetune mode is set, `decay` gets smaller the more training batches we see. However, the code weights the accumulated mean/var with `decay` and the current mean/var with `1-decay`.
Hence, if `N` gets very big, `self.avg_mean`/`self.avg_var` are almost equal to the mean/var of the current batch. It should be the other way around, e.g. the more data we saw, the less influence should the current mean/var have on the average statistics.

TD;LR `decay = 1. / self.N` should be replaced by `decay = 1. - 1. / self.N`

delta2323:
Fix #1091

### DIFF
--- a/chainer/links/normalization/batch_normalization.py
+++ b/chainer/links/normalization/batch_normalization.py
@@ -94,7 +94,7 @@ class BatchNormalization(link.Link):
 
             if finetune:
                 self.N += 1
-                decay = 1. / self.N
+                decay = 1. - 1. / self.N
             else:
                 decay = self.decay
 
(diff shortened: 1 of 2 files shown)


okuta:
Can you write a test?

delta2323:
@okuta I added unit tests.

okuta:
LGTM!

PR #1112, opened 2016-04-12
Async to_gpu

### LINKED ISSUE #886: Asynchronous Host-to-Device copy
Current CuPy does not have a way to allocate page-locked host memory to enable asynchronous HtoD copy. There are e.g. two options to support asynchronous HtoD copy:

1. Allocate page-locked host memory and copy the source buffer to it implicitly in the MemoryPointer.copy_from_host_async() method. It adds an overhead on copying, which might be problematic.
2. Provide a way to create a numpy array on page-locked host memory. I do not investigate the feasibility yet (maybe it can be done with using np.frombuffer and adding a destructor?).

okuta:
fix #886

### DIFF
--- a/chainer/cuda.py
+++ b/chainer/cuda.py
@@ -96,6 +96,9 @@ class DummyDeviceType(object):
     This class is used to represent CPU device.
 
     """
+
+    id = -1
+
     def __int__(self):
         return -1
 
@@ -116,10 +119,6 @@ class DummyDeviceType(object):
 
     def __ne__(self, other):
         return not (self == other)
-
-    @property
-    def id(self):
-        return -1
 
 
 DummyDevice = DummyDeviceType()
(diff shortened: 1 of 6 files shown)


chainer/cuda.py
```diff
@@ -115,6 +115,10 @@ def __eq__(self, other):
     def __ne__(self, other):
         return not (self == other)
 
+    @property
+    def id(self):
+        return -1
```
beam2d: Is it better to make it a class attribute?

chainer/cuda.py
```diff
@@ -190,15 +195,33 @@ def to_gpu(array, device=None, stream=None):
...
+                # cpu to gpu
+                src = array.copy(order='C')
+                ret.set(src, stream)
+            else:
+                # gpu to gpu
+                with array_dev:
+                    src = array.copy()
+                ret.data.copy_from_device_async(src.data, src.nbytes, stream)
+
+            # to hold a reference to the end of the asynchronously memcpy
```
beam2d: to the end -> until the end
asynchronously -> asynchronous

beam2d:
Sorry for being late. LGTM except above comments!

okuta:
@beam2d I fixed.

beam2d:
LGTM!

PR #1149, opened 2016-04-26
Optimized version of the GRU link

fabiencro:
As mentioned in issue #1143 , the current implementation of links.GRU is not very well optimized (while links.LSTM seem to be much more optimized).

Here is an initial pull request. I still need to add the tests.

The performance improvement will depend on the settings of the GRU. Performances gains are expected to be lower with large layer size / batch sizes. As an example, this implementation is more than twice faster than links.GRU in my benchmarks using size 512 for state and input, and a minibatch size of 64, on a Titan X GPU. 

The optimizations were inspired by this article: https://devblogs.nvidia.com/parallelforall/optimizing-recurrent-neural-networks-cudnn-5/

Mostly, I implemented optimizations 1 and 3.

I would like to implement optimization 2 (ie parallelizing the matrix multiplications), but could not find how to apply a Linear link on two different cuda streams. I tried something like this, but it did not seem to work:

```
device = chainer.cuda.get_device(h)
with device:
            stream1 = chainer.cuda.Stream(1)
            stream2 = chainer.cuda.Stream(2)

            chainer.cuda.cublas.setStream(device.cublas_handle, stream1)
            r_z_h_x = self.W_r_z_h(x)

            chainer.cuda.cublas.setStream(device.cublas_handle, stream2)
            r_z_h = self.U_r_z(h)

            stream1.synchronize()
            stream2.synchronize()
```

I would be interested in some advice on this.


fabiencro:
I added fast_gru to the list of tested variants in test_gru.py. I added some functions for converting the fast_gru parameter layout ( 3 linear links) to the gru.py layout (6 linear links).

It would probably be good to add more tests.

I would also like some comments about the fact that I created a separate fast_gru.GRU class instead of modifying the gru.GRU. I somehow like it that way because it allows to have both an easy-to-understand implementation and an optimized one. And the simpler implementation can be used to test the optimized one. But it might be confusing to have several GRU classes doing the same thing ...

delta2323:
Thank you for your PR and sorry for our late reply. For now, I cannot say whether we should separately implement `FastGRU`. But there are several points we must consider.

* Current `FastGRU` cannot initialize `W_r`, `W_z` and `W_h` independently as we merge these matrices into one (same is true for `U_*`s). We can separately initialize them as we do in `LSTM` link (See [this](https://github.com/pfnet/chainer/blob/master/chainer/links/connection/lstm.py#L28)).
* To preserve [backward compatibility](http://docs.chainer.org/en/stable/compatibility.html), we cannot replace completely existing `GRU` with `FastGRU` as the resulting serialized objects will be different, at least if we use current `FastGRU` implementation.
* As you mentioned, it is confusing to have two `GRU`s, that do same things. If we can do what existing `GRU` can do with `FastGRU`, one of the possible solution is to make existing `GRU` deprecated.

PR #1162, opened 2016-05-06
Loosen type condition in linear, convolution_2d and deconvolution_2d

### LINKED ISSUE #519: Convolution2D does not support 64 bit
When I change dtype=numpy.float32 to dtype=numpy.float64 in the constructor of Convolution2D, and then run nosetests I get tons of errors. I think 64 bit is simply not supported, so we should raise a NotImplementedError.

okuta:
See also #555.
Fix #519.

### DIFF
--- a/chainer/functions/connection/convolution_2d.py
+++ b/chainer/functions/connection/convolution_2d.py
@@ -18,6 +18,11 @@ if cuda.cudnn_enabled:
             libcudnn.CUDNN_CONVOLUTION_BWD_DATA_SPECIFY_WORKSPACE_LIMIT
 
 
+def _check_cudnn_acceptable_type(x_dtype, W_dtype):
+    return x_dtype == W_dtype and (
+        _cudnn_version >= 3000 or x_dtype != numpy.float16)
+
+
 def _pair(x):
     if hasattr(x, '__getitem__'):
         return x
@@ -39,8 +44,8 @@ class Convolution2DFunction(function.Function):
         x_type = in_types[0]
         w_type = in_types[1]
         type_check.expect(
-            x_type.dtype == numpy.float32,
-            w_type.dtype == numpy.float32,
+            x_type.dtype.kind == 'f',
+            w_type.dtype.kind == 'f',
             x_type.ndim == 4,
             w_type.ndim == 4,
             x_type.shape[1] == w_type.shape[1],
@@ -49,7 +54,7 @@ class Convolution2DFunction(function.Function):
         if n_in.eval() == 3:
             b_type = in_types[2]
             type_check.expect(
-                b_type.dtype == numpy.float32,
+                b_type.dtype == x_type.dtype,
                 b_type.ndim == 1,
                 b_type.shape[0] == w_type.shape[0],
             )
@@ -61,7 +66,8 @@ class Convolution2DFunction(function.Function):
         self.col = conv.im2col_cpu(
             x, kh, kw, self.sy, self.sx, self.ph, self.pw,
             cover_all=self.cover_all)
-        y = numpy.tensordot(self.col, W, ((1, 2, 3), (1, 2, 3)))
+        y = numpy.tensordot(
+            self.col, W, ((1, 2, 3), (1, 2, 3))).astype(x.dtype)
         if b is not None:
             y += b
         return numpy.rollaxis(y, 3, 1),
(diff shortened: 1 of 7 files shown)


chainer/functions/connection/convolution_2d.py
```diff
@@ -218,8 +227,12 @@ def backward_gpu(self, inputs, grad_outputs):
             W_mat = W.reshape(out_c, -1)
             gcol = cuda.cupy.empty_like(self.col)
             gcol_mats = gcol.reshape(n, c * kh * kw, out_h * out_w)
+
             for i in moves.range(n):
-                cuda.cupy.dot(W_mat.T, gy_mats[i], gcol_mats[i])
+                gcol_mats[i] = cuda.cupy.dot(W_mat.T, gy_mats[i])
+
+            gcol = cuda.cupy.tensordot(W, gy, (0, 1)).astype(x.dtype)
+            gcol = cuda.cupy.rollaxis(gcol, 3)
```
delta2323: Could you explain why we need additional mathematical operations, although changes so far are related to dtypes?

chainer/functions/connection/convolution_2d.py
```diff
@@ -74,8 +80,9 @@ def forward_gpu(self, inputs):
         out_h = conv.get_conv_outsize(h, kh, self.sy, self.ph)
         out_w = conv.get_conv_outsize(w, kw, self.sx, self.pw)
 
-        y = cuda.cupy.empty((n, out_c, out_h, out_w), dtype=x.dtype)
-        if cuda.cudnn_enabled and self.use_cudnn:
+        y = cuda.cupy.zeros((n, out_c, out_h, out_w), dtype=x.dtype)
```
delta2323: Why do you initialize `y` with zero?

delta2323:
Please merge or rebase the master branch after finishing the fix.

chainer/functions/connection/convolution_2d.py
okuta: This is miss. I will remove it.

chainer/functions/connection/convolution_2d.py
```diff
@@ -61,7 +66,8 @@ def forward_cpu(self, inputs):
         self.col = conv.im2col_cpu(
             x, kh, kw, self.sy, self.sx, self.ph, self.pw,
             cover_all=self.cover_all)
-        y = numpy.tensordot(self.col, W, ((1, 2, 3), (1, 2, 3)))
+        y = numpy.tensordot(
+            self.col, W, ((1, 2, 3), (1, 2, 3))).astype(x.dtype)
         if b is not None:
             y += b
```
delta2323: Is it intentional that we do not convert `dtype` after we add `b` to `y`?

delta2323:
ditto (don't we have to convert `dtype`?)
https://github.com/pfnet/chainer/pull/1162/files#diff-16a7c64c13ba043b2f7f21283bf70f47R162

chainer/utils/conv.py
```diff
@@ -105,7 +105,7 @@ def col2im_gpu(col, sy, sx, ph, pw, h, w):
              for (int out_x = out_x_0; out_x < out_x_1; ++out_x) {
                int kx = x - out_x * sx;
                int k = out_y + out_h * (kx + kw * (ky + kh * c0));
-               val += col[out_x + out_w * k];
+               val = val + col[out_x + out_w * k];
```
delta2323: Could you tell me why the previous code does not work.

okuta: `+=` is not supported, if type of `val` is `float16`.

delta2323:
I updated comments.

chainer/functions/connection/convolution_2d.py
okuta: `b.dtype` and `y.dtype` are same.

okuta:
Please merge #1212 before this PR.

chainer/functions/connection/convolution_2d.py
delta2323: OK, I understand it.

chainer/utils/conv.py
delta2323: OK.

delta2323:
LGTM 👍

delta2323:
Now I had merged #1212. As changes of `cupy/core/carray.cuh` in this PR is different from #1212, please remove the change in this file from this PR.

okuta:
I fixed.

delta2323:
LGTM 👍

PR #1165, opened 2016-05-08
Loosen type condition in evaluation functions

okuta:
See also #555

### DIFF
--- a/chainer/functions/evaluation/binary_accuracy.py
+++ b/chainer/functions/evaluation/binary_accuracy.py
@@ -1,3 +1,5 @@
+from __future__ import division
+
 import numpy
 
 from chainer import cuda
@@ -14,7 +16,7 @@ class BinaryAccuracy(function.Function):
         x_type, t_type = in_types
 
         type_check.expect(
-            x_type.dtype == numpy.float32,
+            x_type.dtype.kind == 'f',
             t_type.dtype == numpy.int32,
             t_type.shape == x_type.shape,
         )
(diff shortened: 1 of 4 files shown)


delta2323:
I have also fixed `Accuracy` in #1162.

chainer/functions/evaluation/binary_accuracy.py
```diff
@@ -27,7 +27,7 @@ def forward(self, inputs):
         t = t.ravel()
         c = (y >= 0)
         count = xp.maximum(1, (t != self.ignore_label).sum())
-        return xp.asarray((c == t).sum(dtype='f') / count, dtype='f'),
+        return xp.asarray((c == t).sum(dtype=y.dtype) / count, dtype=y.dtype),
```
delta2323: Do we have to change `dtype` twice?

okuta: Yes. But I have better idea. I will fix it.

chainer/functions/evaluation/binary_accuracy.py
```diff
@@ -27,7 +29,7 @@ def forward(self, inputs):
         t = t.ravel()
         c = (y >= 0)
         count = xp.maximum(1, (t != self.ignore_label).sum())
-        return xp.asarray((c == t).sum(dtype='f') / count, dtype='f'),
+        return xp.asarray((c == t).sum() / count, dtype=y.dtype),
```
delta2323: I think it is a division of an integer by an integer. So it causes a rounding to the integer, doesn't it?

okuta: It uses Python3 mode.
https://github.com/pfnet/chainer/pull/1165/files/91b37a010ae0ff64115e1fbc54b427f292b9920c#diff-36d8cd578e4268189b044f7fad32c5e3R1

delta2323: Sorry, I missed it.

delta2323:
LGTM 👍

PR #1168, opened 2016-05-09
added gradient noise optimizer hook

kashif:
This hook function simply adds time-dependent annealed Gaussian noise
to the gradient at every training step. See: [Adding Gradient Noise
Improves Learning for Very Deep Networks](http://arxiv.org/pdf/1511.06807)


chainer/optimizer.py
```diff
@@ -518,3 +518,52 @@ def __call__(self, opt):
...
+
+    def __call__(self, opt):
+        if cuda.available:
+            kernel = cuda.elementwise(
+                'T noise', 'T g', 'g += noise', 'gradient_noise')
+
+        for param in opt.target.params():
+            p, g = param.data, param.grad
+            xp = cuda.get_array_module(p)
+            noise = xp.random.normal(0,
```
jheymann85: Thanks for the code. One bug I encountered: This must be moved into the device context (e.g. to line 566), otherwise the hook does not work when using multiple GPUs.

chainer/optimizer.py
```diff
@@ -518,3 +518,51 @@ def __call__(self, opt):
...
+            kernel = cuda.elementwise(
+                'T noise', 'T g', 'g += noise', 'gradient_noise')
+
+        for param in opt.target.params():
+            g = param.grad
+            xp = cuda.get_array_module(g)
+            with cuda.get_device(g) as dev:
+                noise = xp.random.normal(0, numpy.sqrt(
+                    self.eta / numpy.power(1 + opt.t, 0.55)), g.shape).astype(
+                    numpy.float32)
```
delta2323: Use `g.dtype` instead of `numpy.float32`.

delta2323:
Do you think it is possible to separate the mechanism that determines the quantity of the noise added from the one that does inject the noise to the gradient? Currently these two mechanisms are strongly tied to each other.

kashif:
@delta2323 yes I could do that... but can you kindly let me know where to add the two functions? At the moment the noise is very dependent on the time step of the optimisation, so it made sense to me to confound the two together into this single gradient noise hook function. but if you can tell me where to add the noise helper, then I can try to separate it. Thanks!

delta2323:
One option is to extract the calculation of noise as a function as follows:

```
def exponential_decay_noise(xp, shape, dtype, hook, opt):
    std = numpy.sqrt(hook.eta / numpy.power(1 + opt.t, 0.55))
    return xp.random.normal(0, std, shape).astype(dtype)


class GradientNoise(object):
    def __init__(self, eta, noise_func=exponential_decay_noise):
        self.eta = eta
        self.noise_func = noise_func

    def __call__(self, opt):
        ...
        for param in opt.target.params():
            g = param.grad
            xp = cuda.get_array_module(g)
            with cuda.get_device(g) as dev:
                noise = self.noise_func(xp, g.shape, g.dtype, self, opt)
                ...
```

The function name (`exponential_decay_noise`) and its arguments are not yet considered well. The arguments are the most "informative" ones in the sense that the function takes the hook and optimizer themselves and extract the information that is necessary. I guess `shape` and `dtype` are the most essential arguments.

kashif:
thanks @delta2323 my question was about the noise helper, in which file should it be? would it make sense to put it in `optimizer.py` or somewhere else? thanks!

delta2323:
@kashif Sorry, I forgot to answer the question. I think we put the helper function in `optimizer.py`, or same location as `GradientNoise` hook. We need not to separate it from the hook because we can expect we do not use it other than for this hook.

kashif:
@delta2323 thanks changed as per your suggestion.

chainer/optimizer.py
```diff
@@ -518,3 +523,50 @@ def __call__(self, opt):
...
+
+    . math::
+
+        \\sigma_t^2 = \\frac{\\eta}{(1+t)^\\gamma}
+
+    with :math:`\\eta` selected from {0.01, 0.3, 1.0} and
+    :math:`\\gamma = 0.55`.
+
+    Args:
+        eta (float): parameter for the standard deviation.
```
delta2323: Please add `noise_func` argument to the description of `Args`.

chainer/optimizer.py
```diff
@@ -518,3 +523,50 @@ def __call__(self, opt):
...
+
+class GradientNoise(object):
+    """Optimizer hook function for adding gradient noise.
+
+    This hook function simply adds time-dependent annealed Gaussian noise
+    to the gradient at every training step:
+
+    . math::
+
+        g_t \\leftarrow g_t + N(0, \\sigma_t^2)
```
delta2323: As we design this hook so that we can flexibly change the noise function (by `noise_func`), we need to change the document so that this behavior is the default one. We also need to write how to change the noise function as well.

chainer/optimizer.py
```diff
@@ -518,3 +523,50 @@ def __call__(self, opt):
...
+
+    where
+
+    . math::
+
+        \\sigma_t^2 = \\frac{\\eta}{(1+t)^\\gamma}
+
+    with :math:`\\eta` selected from {0.01, 0.3, 1.0} and
```
delta2323: We do not restrict the range of `eta` to these three values in this hook. So we need to clear that this choice is done in the original paper.

delta2323:
Sorry for my late response. LGTM except comments.

chainer/optimizer.py
```diff
@@ -518,3 +530,52 @@ def __call__(self, opt):
...
+    :math:`\\gamma = 0.55`.
+
+    Args:
+        eta (float): parameter for the noise function, which for the default
```
delta2323: Could you explain the role of the attribute in more detail? For example, "parameter that defines the scale of the noise".

delta2323:
@kashif Sorry, I have one more comments. Could you check that?

delta2323:
LGTM 👍

delta2323:

kashif:
thank you 🙇

PR #1178, opened 2016-05-14
Fix HuberLoss's backward computation

### LINKED ISSUE #1175: HuberLoss's backward() should not ignore gy
`HuberLoss`'s `backward()` ignores `gy`, so any computation after `HuberLoss` has no effect for gradients backpropagated by `HuberLoss`. I think such a behavior is not correct.

```python
>>> x = chainer.Variable(np.zeros((1,1), dtype=np.float32))
>>> t = chainer.Variable(np.ones((1,1), dtype=np.float32))
>>> F.huber_loss(x, t, 1.0).backward()
>>> x.grad
array([[-1.]], dtype=float32)
```
```python
>>> x = chainer.Variable(np.zeros((1,1), dtype=np.float32))
>>> t = chainer.Variable(np.ones((1,1), dtype=np.float32))
>>> (F.huber_loss(x, t, 1.0) * 0).backward()  # Multiply the loss by zero
>>> x.grad
array([[-1.]], dtype=float32)
```

muupan:
This PR fixes #1175 .

- Modified `HuberLoss`'s `backward()` so that it will compute correct gradients even when `gy` is not ones.
- Modified its test case so that `gy` is set to random values.

Since `backward()` will start to raise an error when `gy` is `None` after this PR, some existing code might stop to work.

### DIFF
--- a/chainer/functions/loss/huber_loss.py
+++ b/chainer/functions/loss/huber_loss.py
@@ -31,7 +31,8 @@ class HuberLoss(function.Function):
     def backward(self, inputs, gy):
         xp = cuda.get_array_module(*inputs)
         mask = xp.abs(self.diff) <= self.delta
-        gx = xp.where(mask, self.diff, self.delta * xp.sign(self.diff))
+        gx = gy[0].reshape(gy[0].shape + (1,) * (self.diff.ndim - 1)) * \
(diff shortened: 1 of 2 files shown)


delta2323:
As the previous behavior is an implementation bug, my opinion is to fix as you did.

delta2323:
LGTM 👍

PR #1189, opened 2016-05-18
Add `deterministic` option into convolution_2d and deconvolution_2d

t-abe:
Some of convolution algorithm cuDNN provides are non-deterministic.
So outputs of current chainer convolution and deconvolution can be non-deterministic when cuDNN is available.
I think this might make trouble when you want to do reproducible experiments.

This PR add `deterministic` option into these two functions and links.
If `True`, the functions force cuDNN to use deterministic algorithms.

### DIFF
--- a/chainer/functions/connection/convolution_2d.py
+++ b/chainer/functions/connection/convolution_2d.py
@@ -26,11 +26,12 @@ def _pair(x):
 
 class Convolution2DFunction(function.Function):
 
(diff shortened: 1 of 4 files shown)


delta2323:
Could you add test cases for this option where we call the function twice and compare results.

delta2323:
We had reverted and resent this PR as #1321. So this feature is not merged yet.

PR #1191, opened 2016-05-18
Loosen inputs' shape condition of Accuracy Function

### LINKED ISSUE #1190: Inconsistency of type checking between SoftmaxCrossEntropy and Accuracy
We often evaluate training procedure with both `softmax_cross_entropy` and `accuracy`, as default `chainer.links.Classifier` does. But there are cases `softmax_cross_entorpy(x, t)` is accepted but `accuracy(x, t)` raises error because of its type checking (and vise versa). Replicated softmax is one of a such examples (e.g. `x.shape = (B, C, N)` and `t.shape = (B, N)`). We need to reshape them manually in such a case.

It is convenient for users that shape checking of the two be more consistent.

delta2323:
Fix #1190

Let the shape of `t` be of the form `(S0, S1, ..., Sn)`, then accepted shapes of `x` is of the form `(S0, C, S1, ..., Sn[, 1, 1, ..., 1])` where `n >= 0` and the length of 1's is more than or equals to 0. This condition includes the previous condition. And it also supports the condition of `SoftMaxCrossEntropy`.

See the test cases for the combination of shapes accepted.

### DIFF
--- a/chainer/functions/evaluation/accuracy.py
+++ b/chainer/functions/evaluation/accuracy.py
@@ -32,8 +32,6 @@ class Accuracy(function.Function):
     def forward(self, inputs):
         xp = cuda.get_array_module(*inputs)
         y, t = inputs
-        y = xp.rollaxis(y, 1)
-        y = xp.reshape(y, (-1,) + t.shape)
 
         if self.ignore_label is not None:
             mask = (t == self.ignore_label)
@@ -43,7 +41,8 @@ class Accuracy(function.Function):
             # TODO(henry0312)
             #   If cupy.where returns indexes, we could make the code better.
             #   Also, we would need Advanced Indexing.
-            pred = xp.where(mask, self.ignore_label, y.argmax(axis=0))
+            pred = xp.where(mask, self.ignore_label,
+                            y.argmax(axis=1).reshape(t.shape))
             count = (pred == t).sum() - ignore_cnt
             total = t.size - ignore_cnt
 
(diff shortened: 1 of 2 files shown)


delta2323:
`dtype` conditions are changed in #1165, so we will fix them in this PR, too.

okuta:
LGTM!

okuta:
Please merge or rebase the master branch.

delta2323:
I merged master branch. And I also made changes regarding `ignore_label, which was supported` by #1113.

chainer/functions/evaluation/accuracy.py
```diff
@@ -37,16 +43,16 @@ def forward(self, inputs):
...
-            total = len(t) - ignore_cnt
+            total = t.size - ignore_cnt
 
             if total == 0:
                 return xp.asarray(0.0, dtype='f'),
             else:
                 return xp.asarray(float(count) / total, dtype='f'),
         else:
-            pred = y.argmax(axis=1)
+            pred = y.argmax(axis=0)
```
okuta: How about `pred = y.argmax(axies=1).reshape(t.shape)`?

chainer/functions/evaluation/accuracy.py
```diff
@@ -37,16 +43,16 @@ def forward(self, inputs):
             # TODO(henry0312)
             #   If cupy.where returns indexes, we could make the code better.
             #   Also, we would need Advanced Indexing.
-            pred = xp.where(mask, self.ignore_label, y.argmax(axis=1))
+            pred = xp.where(mask, self.ignore_label, y.argmax(axis=0))
```
okuta: How about `pred = xp.where(mask, self.ignore_label, y.argmax(axis=1).reshape(t.shape))`?

okuta:
LGTM except comments.

okuta:
OK.
Please merge or rebase the master branch.

delta2323:
I have merged master branch.

PR #1212, opened 2016-05-25
Add compound assignment operators of float16

okuta:
When `T` is `float16`, I want to use the following code.
```python
cuda.elementwise(
    'T grad, T lr', 'T param',
    'param -= lr * grad',
    'sgd')
```

### DIFF
--- a/cupy/core/carray.cuh
+++ b/cupy/core/carray.cuh
@@ -61,28 +61,28 @@ public:
   inline __device__ float16& operator+=(const T& rhs)
   {
     *this = *this + rhs;
-	return *this;
+    return *this;
   }
 
   template<typename T>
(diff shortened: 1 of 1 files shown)


delta2323:
LGTM. It is not directly relevant to this PR but where do we define `float16::operator+` etc.?

okuta:
`float16::operator+` is not defined.
`float16` is castted to `float32` automatically when we use `float16` arithmetic operation.
Because the old CUDA dose not support `float16` arithmetic operation.

The CDUA7.5 provides half arithmetic operation.
http://docs.nvidia.com/cuda/cuda-math-api/group__CUDA__MATH____HALF__ARITHMETIC.html#group__CUDA__MATH____HALF__ARITHMETIC

I created Issue #1224.

PR #1173, opened 2016-05-11
Add log1p and expm1 function

yuji-mizobuchi:
Add log1p and expm1 function.  #1160

PR #1222, opened 2016-05-27
Add log1p and expm1 functions.

yuji-mizobuchi:
I am really sorry for opening another pull request that is the same as #1173 .
If there is something wrong, please give me comments.
Thanks.


delta2323:
Thank you for your work!

PR #1223, opened 2016-05-29
[WIP] Add aggregate_option option to loss functions

### LINKED ISSUE #1088: Elementwise negative loglikelihood of VAE
In some situation, we need negative loglikelihood values for each output of the decoder of VAE. We already have `F.gaussian_nll` or `F.bernoulli_nll`, but they are loss functions, that is, they sum up these values and return the sum as a loss. We want a `Function`s that return each summand.

delta2323:
This PR is one of the solutions that fix #1088.

Each loss function first calculates _raw_ loss values and aggregates them with `Aggregator` in the forward operation. The behavior of `Aggregator` is determined by `aggregate_option`. Currently, it supports five options.
- `elementwise` : Actually, `Aggregator` does nothing in this case.
- `samplewise` : `Aggregator` takes sum along all axes except the first one.
- `sum` : `Aggregator` takes sum along all axes.
- `mean` : `Aggregator` takes sum along all axes and divides the resultant by the number of samples.
- `divide` : `Aggregator` takes sum along all axes and divides the resultant by specified value. This option is used to implement the default behavior of `sigmoid_cross_entropy` and `softmax_cross_entropy`.
- [x] Fix `NegativeSampling` Class to support `aggregate_option`.
- [ ] Test GPU mode.
- [ ] Add tests to `AxesAggregator`, `Aggregator`, and `AggregateFunction`.
- [ ] Move AggregateFunction to `functions` directory.
- [ ] Naming is not thought out well.
- [ ] Fix documents of each loss functions.
- [ ] Add documents to `AxesAggregator`, `Aggregator`, and `AggregateFunction`
- [ ] Make an abstract class for loss functions.

### DIFF
--- a/chainer/functions/loss/contrastive.py
+++ b/chainer/functions/loss/contrastive.py
@@ -2,6 +2,7 @@ import numpy
 
 from chainer import cuda
 from chainer import function
+from chainer.utils import aggregator
 from chainer.utils import type_check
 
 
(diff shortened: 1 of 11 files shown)


delta2323:
We are now redesigning loss functions and planing to resend PR. So we close this PR.

PR #1232, opened 2016-05-30
Fix checking shape in BatchNormalizationFunction

okuta:
Type checking is loose.
```
In [1]: import numpy

In [2]: import chainer

In [3]: bn0 = chainer.links.BatchNormalization(64)

In [4]: bn0(chainer.Variable(numpy.zeros((1,100,1,1),dtype='f')))
ValueError                                Traceback (most recent call last)
<ipython-input-4-ddda6c9bb1a9> in <module>()
----> 1 bn0(chainer.Variable(numpy.zeros((1,100,1,1),dtype='f')))

/home/okuta/env/chainer/chainer/chainer/links/normalization/batch_normalization.pyc in __call__(self, x, test, finetune)
     92         if use_batch_mean:
     93             func = batch_normalization.BatchNormalizationFunction(self.eps)
---> 94             ret = func(x, self.gamma, self.beta)
     95
     96             if finetune:

/home/okuta/env/chainer/chainer/chainer/function.pyc in __call__(self, *inputs)
    121         # Forward prop
    122         with cuda.get_device(*in_data):
--> 123             outputs = self.forward(in_data)
    124             assert type(outputs) == tuple
    125         for hook in six.itervalues(hooks):

/home/okuta/env/chainer/chainer/chainer/functions/normalization/batch_normalization.pyc in forward(self, inputs)
     62             x_mu = x - mean[expander]
     63             self.x_hat = x_mu / self.std[expander]
---> 64             y = gamma * self.x_hat
     65             y += beta
     66         else:

ValueError: operands could not be broadcast together with shapes (1,64,1,1) (1,100,1,1)
```

### DIFF
--- a/chainer/functions/normalization/batch_normalization.py
+++ b/chainer/functions/normalization/batch_normalization.py
@@ -18,9 +18,11 @@ class BatchNormalizationFunction(function.Function):
                 '%s == %s' % (in_types.size(), n_in))
 
         x_type, gamma_type, beta_type = in_types[:3]
+        M = gamma_type.ndim.eval()
         type_check.expect(
             x_type.dtype.kind == 'f',
             x_type.ndim >= gamma_type.ndim + 1,
+            x_type.shape[1:1 + M] == gamma_type.shape,
             # TODO(beam2d): Check shape
             gamma_type.dtype == x_type.dtype,
             beta_type.dtype == x_type.dtype,
--- a/tests/chainer_tests/links_tests/normalization_tests/test_batch_normalization.py
+++ b/tests/chainer_tests/links_tests/normalization_tests/test_batch_normalization.py
@@ -245,11 +245,13 @@ class BatchNormalizationTestWithoutGammaAndBeta(unittest.TestCase):
         self.check_backward(x, gy)
 
 
+@testing.parameterize(*testing.product({
+    'size': [3, (2, 3)],
+}))
 class TestInitialize(unittest.TestCase):
 
     def setUp(self):
         self.decay = 0.9
-        self.size = 3
         self.initial_gamma = numpy.random.uniform(-1, 1, self.size)
         self.initial_gamma = self.initial_gamma.astype(numpy.float32)
         self.initial_beta = numpy.random.uniform(-1, 1, self.size)
--- a/tests/chainer_tests/links_tests/normalization_tests/test_batch_normalization.py
+++ b/tests/chainer_tests/links_tests/normalization_tests/test_batch_normalization.py
@@ -289,22 +289,30 @@ class TestDefaultInitializer(unittest.TestCase):
         testing.assert_allclose(numpy.zeros(self.size), self.link.beta.data)
 
 
+@testing.parameterize(*testing.product({
+    'shape': [(2, 4), (2, 5, 3, 4)],
+}))
 class TestInvalidInput(unittest.TestCase):
 
     def setUp(self):
         self.link = links.BatchNormalization(3)
 
-    def test_invalid_initialize(self):
-        with self.assertRaises(TypeError):
-            self.link = links.BatchNormalization({})
-
-    def test_invalid_shape(self):
-        with self.assertRaises(type_check.InvalidType):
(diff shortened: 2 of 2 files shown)


chainer/functions/normalization/batch_normalization.py
```diff
@@ -21,6 +21,7 @@ def check_type_forward(self, in_types):
         type_check.expect(
             x_type.dtype.kind == 'f',
             x_type.ndim >= gamma_type.ndim + 1,
+            x_type.shape[1] == gamma_type.shape[0],
```
delta2323: I think this condition supports only the case `ndim` of `gamma` equals to 1. If `x.shape` is `(B, s1, s2, ..., sN)`, then `gamma.shape` should be `(s1, ..., sM)` where `M <= N`. So the condition is `x_type.shape[1:1+M] == gamma_type.shape` where `M` is the `ndim` of `gamma`.

tests/chainer_tests/links_tests/normalization_tests/test_batch_normalization.py
```diff
@@ -175,4 +176,17 @@ def test_statistics2_gpu(self):
             cuda.to_gpu(self.y))
 
 
+class TestInvalidInput(unittest.TestCase):
+
+    def setUp(self):
+        self.link = links.BatchNormalization(3)
```
delta2323: Please add test case where `size` argument is other than `int` type.

delta2323:
I added comments.

okuta:
@delta2323: I fixed.

tests/chainer_tests/links_tests/normalization_tests/test_batch_normalization.py
```diff
@@ -243,4 +244,23 @@ def test_backward_gpu(self):
         gy = cuda.to_gpu(self.gy)
         self.check_backward(x, gy)
 
+
+class TestInvalidInput(unittest.TestCase):
+
+    def setUp(self):
+        self.link = links.BatchNormalization(3)
+
+    def test_invalid_initialize(self):
```
delta2323: This test fixture should be separated as another class because it does not depend on `setUp` of this class.

tests/chainer_tests/links_tests/normalization_tests/test_batch_normalization.py
```diff
@@ -243,4 +244,23 @@ def test_backward_gpu(self):
...
+        with self.assertRaises(TypeError):
+            self.link = links.BatchNormalization({})
+
+    def test_invalid_shape(self):
```
delta2323: Please add GPU version of this test.

tests/chainer_tests/links_tests/normalization_tests/test_batch_normalization.py
```diff
@@ -243,4 +244,23 @@ def test_backward_gpu(self):
...
+        with self.assertRaises(type_check.InvalidType):
+            self.link(chainer.Variable(numpy.zeros((2, 5, 3, 4), dtype='f')))
+
+    def test_invalid_shape_cpu(self):
```
delta2323: Please add GPU version of this test.

delta2323:
Thank you for your work. Could you consider applying this [comment](https://github.com/pfnet/chainer/pull/1232#discussion_r66796772), too?

okuta:
@delta2323 I fixed.

delta2323:
Could you add test case where the argument of Batchnormalization.__init__ is other than int?

delta2323:
@okuta  LGTM except the comment.

okuta:
https://github.com/pfnet/chainer/pull/1232/files#diff-9ba9992ebffe5029a4bb23a18efb9691R315
I tested. Is it OK?

delta2323:
I see. What I meant was `tuple of int` case. Actually `size` argument can also accept it (see the description of `L.BatchNormalization` [here](https://github.com/pfnet/chainer/blob/master/chainer/links/normalization/batch_normalization.py#L34)).

okuta:
I fixed.

delta2323:
LGTM 👍

PR #1240, opened 2016-05-31
Fix coding style

okuta:


chainer/links/caffe/caffe_function.py
```diff
@@ -348,7 +348,7 @@ def _setup_eltwise(self, layer):
         # stable_prod_grad parameter is not supported now.
         operation = layer.eltwise_param.operation
         coeffs = layer.eltwise_param.coeff
```
gwtnb: How about `coeffs = layer.eltwise_param.coeff or None`?

gwtnb:
LGTM

PR #1247, opened 2016-06-02
gradient_check.check_backward supports dtype option

okuta:
`gradient_check.check_backward` is poor accuracy for `numpy.float16` array.
I add `dtype` option.
`check_backward` casts input arrays.

### DIFF
--- a/chainer/gradient_check.py
+++ b/chainer/gradient_check.py
@@ -257,7 +257,7 @@ def check_backward(func, x_data, y_grad, params=(),
         if dtype is None:
             assert gx.dtype == x.grad.dtype
         else:
-            assert gx.dtype == dtype
+            assert gx.dtype.kind == 'f' and gx.dtype == dtype
 
     for p in params:
         gp, = numerical_grad(f, (p.data,), y_grad, eps=eps)
--- a/chainer/gradient_check.py
+++ b/chainer/gradient_check.py
@@ -187,7 +187,7 @@ def check_backward(func, x_data, y_grad, params=(),
         no_grads (list of bool): Flag to skip variable for gradient assertion.
             It should be same length as ``x_data``.
         dtype (~numpy.dtype): `x_data` and `y_grad` are casted to this dtype
-            when calculating numerical gradients.
+            when calculating numerical gradients. Only float type is allowed.
 
     See:
        :func:`numerical_grad`
@@ -227,15 +227,18 @@ def check_backward(func, x_data, y_grad, params=(),
     y[0].backward()
 
     if dtype is None:
-        cxs = [variable.Variable(x) for x in x_data]
+        casted_xs = [variable.Variable(x) for x in x_data]
     else:
+        if numpy.dtype(dtype).kind != 'f':
+            raise ValueError('`dtype` is allowed only float type')
         if len(params) > 0:
             raise ValueError('`dtype` is available only if `params` is empty')
-        cxs = [variable.Variable(x.astype(dtype) if x.dtype.kind == 'f' else x)
-               for x in x_data]
+        casted_xs = [variable.Variable(x.astype(dtype)
+                     if x.dtype.kind == 'f' else x)
+                     for x in x_data]
 
     def f():
-        ys = func(*cxs)
+        ys = func(*casted_xs)
         ys = _as_tuple(ys)
         return tuple(y.data for y in ys)
 
@@ -245,7 +248,7 @@ def check_backward(func, x_data, y_grad, params=(),
         if len(no_grads) != len(xs):
             raise ValueError(
                 'Length of no_grads param and xs should be same.')
-    for skip, x, cx in six.moves.zip(no_grads, xs, cxs):
+    for skip, x, cx in six.moves.zip(no_grads, xs, casted_xs):
         if skip:
             assert x.grad is None
             continue
--- a/chainer/gradient_check.py
+++ b/chainer/gradient_check.py
@@ -93,7 +93,7 @@ def _as_tuple(x):
 
 
 def check_backward(func, x_data, y_grad, params=(),
-                   eps=1e-3, atol=1e-5, rtol=1e-4, no_grads=None):
+                   eps=1e-3, atol=1e-5, rtol=1e-4, no_grads=None, dtype=None):
     """Test backward procedure of a given function.
 
     This function automatically check backward-process of given function.
@@ -186,6 +186,8 @@ def check_backward(func, x_data, y_grad, params=(),
             :func:`chainer.testing.assert_allclose`.
         no_grads (list of bool): Flag to skip variable for gradient assertion.
             It should be same length as ``x_data``.
+        dtype (~numpy.dtype): `x_data` and `y_grad` are casted to this dtype
+            when calculating numerical gradients.
 
     See:
        :func:`numerical_grad`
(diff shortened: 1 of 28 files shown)


chainer/gradient_check.py
```diff
@@ -222,16 +226,24 @@ def check_backward(func, x_data, y_grad, params=(),
     # `Variable.backward` method calls `Function.backward` of its creator.
     y[0].backward()
 
+    if dtype is None:
+        cxs = [variable.Variable(x) for x in x_data]
+    else:
+        if len(params) > 0:
+            raise ValueError('`dtype` is available only if `params` is empty')
```
delta2323: Could you tell me why we must forbid the case where dtype is not `None` and `params` is empty?

chainer/gradient_check.py
```diff
@@ -184,6 +186,8 @@ def check_backward(func, x_data, y_grad, params=(),
             :func:`assert_allclose`.
         rtol (float): Relative tolerance to be passed to
             :func:`assert_allclose`.
+        dtype (~numpy.dtype): `x_data` and `y_grad` are casted to this dtype
```
delta2323: Document says `y_grad` is also casted to specified `dtype`. But it seems to me there is no corresponding code.

chainer/gradient_check.py
```diff
@@ -222,16 +226,24 @@ def check_backward(func, x_data, y_grad, params=(),
     # `Variable.backward` method calls `Function.backward` of its creator.
     y[0].backward()
 
+    if dtype is None:
+        cxs = [variable.Variable(x) for x in x_data]
```
delta2323: Variable name is not understandable. What does c in `cxs` stand for?

tests/chainer_tests/test_gradient_check.py
```diff
@@ -324,6 +324,9 @@ def backward(self, inputs, grads):
         return grads
 
 
+@testing.parameterize(*testing.product({
+    'dtype': [None, numpy.float32, numpy.float64],
```
delta2323: Why don't you test numpy.float16 case?

delta2323:
I added comments.

delta2323:
What happens when we feed `dtype` argument with dtypes other than float like `int32`?

chainer/gradient_check.py
okuta: We can not change `dtype` in `params`.

okuta:
Maybe `int32` is acceptable.
But I do not know what it makes sense.

delta2323:
I read `check_backward` again and found that it should fail at [this line](https://github.com/pfnet/chainer/blob/a608dfe3cb1d45872b69b9a990d764a767901bb4/chainer/gradient_check.py#L261) when `dtype` equals to `int32`. So I think we should restrict `dtype` to float.

tests/chainer_tests/test_gradient_check.py
okuta: `numpy.float16` is not tested because it type is low of the low precision.

okuta:
@delta2323 I fixed. Please review this.

chainer/gradient_check.py
```diff
@@ -186,6 +186,8 @@ def check_backward(func, x_data, y_grad, params=(),
             :func:`chainer.testing.assert_allclose`.
         no_grads (list of bool): Flag to skip variable for gradient assertion.
             It should be same length as ``x_data``.
+        dtype (~numpy.dtype): `x_data` and `y_grad` are casted to this dtype
+            when calculating numerical gradients. Only float type is allowed.
```
delta2323: If we restrict `dtype` to `float`, I think we should raise some error in other case. Because it can detect unexpected mistakes.

chainer/gradient_check.py
```diff
@@ -224,8 +226,19 @@ def check_backward(func, x_data, y_grad, params=(),
     # `Variable.backward` method calls `Function.backward` of its creator.
     y[0].backward()
 
+    if dtype is None:
+        casted_xs = [variable.Variable(x) for x in x_data]
+    else:
+        if numpy.dtype(dtype).kind != 'f':
+            raise ValueError('`dtype` is allowed only float type')
```
delta2323: We do not need backquotes here because it is not compiled by sphinx.

chainer/gradient_check.py
```diff
@@ -224,8 +226,19 @@ def check_backward(func, x_data, y_grad, params=(),
...
+        if len(params) > 0:
+            raise ValueError('`dtype` is available only if `params` is empty')
```
delta2323: ditto (backquotes are not needed.)

chainer/gradient_check.py
```diff
@@ -211,7 +213,7 @@ def check_backward(func, x_data, y_grad, params=(),
         if len(y) != len(y_grad):
             raise ValueError(
                 '`y_grad` must have the same length of output values')
```
delta2323: Although it is out of the scope of this PR but we can also remove these backquotes, too.

delta2323:
Just a comment, I believe it does not change your implementation.

> I read check_backward again and found that it should fail at this line when dtype equals to int32. 

I found that when the function to be tested returns `int32`, `numerical_grad` returns `gx` with `int32`. So the assertion is satisfied and this comment was wrong in that sense. But I still think we should restrict `dtype` to float because it is natural that the inputs and outputs of the function tested by `gradient_check` are floats because they are perturbed by small float values.

chainer/gradient_check.py
```diff
@@ -235,13 +248,16 @@ def f():
...
             assert x.grad is None
             continue
-        gx, = numerical_grad(f, (x.data,), y_grad, eps=eps)
+        gx, = numerical_grad(f, (cx.data,), y_grad, eps=eps)
         testing.assert_allclose(gx, x.grad, atol=atol, rtol=rtol)
-        assert gx.dtype is x.grad.dtype
+        if dtype is None:
+            assert gx.dtype == x.grad.dtype
+        else:
+            assert gx.dtype == dtype
```
delta2323: I found that when that when the `dtype` of `x` is `int32`, `gx` will be `int32`. As `x` is not cast to `dtype`, whose kind is `float` in the current implementation, the assertion should fail. The possible solutions are

* to convert `x` even if `x` is not float.
* to impose `x` to be float.

Could you consider the problem?

okuta:
I fixed.

chainer/gradient_check.py
delta2323: Only float type*s* and `None` are allow.

tests/chainer_tests/functions_tests/activation_tests/test_elu.py
```diff
@@ -27,11 +27,11 @@ def setUp(self):
         self.gy = numpy.random.uniform(-1, 1, self.shape).astype(self.dtype)
         self.alpha = random.random()
         self.check_forward_options = {}
-        self.check_backward_options = {}
+        self.check_backward_options = {'dtype': numpy.float64}
         if self.dtype == numpy.float16:
             self.check_forward_options = {'atol': 5e-4, 'rtol': 5e-3}
             self.check_backward_options = {
-                'eps': 2.0 ** -5, 'atol': 1e-3, 'rtol': 1e-2}
+                'dtype': numpy.float64, 'atol': 5e-4, 'rtol': 5e-3}
```
delta2323: This assignment is redundant.

tests/chainer_tests/test_gradient_check.py
delta2323: Could you explain it in more detail?

delta2323:
@okuta LGTM except the comment

okuta:
I added a comment

delta2323:
LGTM 👍

PR #1249, opened 2016-06-03
Extract timer functionality from TimerHook

delta2323:
This PR extracts `Timer` class from `TimerHook` to make this functionality reusable. Specifically, I made an abstract `Timer` class, concrete realizations of this class (`CPUTimer` and `GPUTimer`), and factory method of `Timer` objects.
- [x] Unittests
- [ ] Document

Related to #1004

### DIFF
--- a/chainer/function_hooks/timer.py
+++ b/chainer/function_hooks/timer.py
@@ -40,6 +40,10 @@ class TimerHook(function_.FunctionHook):
         self._preprocess()
 
     def _postprocess(self, function):
+        # For backward compatibility
+        if self.xp is numpy:
+            self.stop = self.timer.stop_times[-1]
+
         self.timer.stop()
         self.call_history.append((function, self.timer.total_time()))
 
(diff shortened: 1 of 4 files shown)


delta2323:
I think we need the explanation on the synchronization of `GPUTimer`.

On the start of measurement (i.e. when we call `GPUTimer.start`), we record an event and optionally may synchronize the event depending on the blocking pattern, which is specified by `GPUTimer.blocking_method`. If blocking_method is `block_every_time`, every measurement is synchronized at its launch. If blocking_method is `block_first_time`, only first measurement is synchronized. If blocking_method is `non_block`, we do not synchronize any measurement at all.

When we call `GPUTimer.synchronize`, the event issued at the end of last measurement is synchronized and the timer is ready for calculating the elapsed times (maybe we need to change to synchronize all events if we use multiple streams). Once synchronization occurred, we cannot add measurements any more by calling `GPUTimer.start`.

I have prepared three types of synchronization pattern, namely, `non_block` and `block`

delta2323:
We are working on clearing branches of this repository. I close this PR as it is too old.

PR #1076, opened 2016-03-26
[WIP] Add 3d convolution

ywatanabex:
I have implemented 3d convolution.
This enhancement is proposed in #945 .

Though the code passes tests on my environment. (with cudnn v4 and gpu),
I need to improve it to support various environment.

Beforehand, I would like to obtain some feedback.

Issue #945, opened 2016-02-14
Support 3D Convolutional Layers


oeway:
It's going to be great to support 3d convolution for volumetric and spatio-temporal data processing.

In theano, they use 2d convolution to do 3d convolution:
https://github.com/Theano/Theano/blob/master/theano/tensor/nnet/conv3d2d.py#L170

I am new to chainer, before I am working on Keras. Is it possible to use the same trick as theano to do 3d convolution in chainer?

PR #1258, opened 2016-06-07
[WIP] Support N-dimentional Convolutional Layers.

takagi:
Rel #945 and #1076.

**Goal**
- Implement nd-Convolution function and link
- Implement nd-Deconvolution function and link
- Implement nd-Pooling function and link
- cuDNN support as well as CUDA, from which version?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,7 +28,7 @@ chainer_setup_build.parse_args()
 
 setup(
     name='chainer',
-    version='1.9.0',
+    version='1.9.1',
     description='A flexible framework of neural networks',
     author='Seiya Tokui',
     author_email='tokui@preferred.jp',


takagi:
I first implement naive generalization from 2D convolution to ND one, which is working on, then think about how we should maintain them later.

takagi:
How should we generalize `im2col_gpu` and `col2im_gpu` functions in conv.py which use elementwise CUDA kernels? There may be some possible ways as illustrated in the following discision tree. As first implementation, I adopt code generation and will push it shortly for reference.

- Actually generalize as code?
  - Yes
    - Use `cupy.elementwise`?
      - Yes
        - When to take care of the generality?
          - CUDA runtime
            - Implement CUDA C operations in`cupy.elementwise`.
          - CUDA compile time
            - Code generation.
       - No
         - Implement in cupy with adding necessary functions.
  - No
    - Implement codes for required dimensions individually: 1D, 2D, 3D, 4D...

delta2323:
> Implement CUDA C operations in cupy.elementwise.
> Implement in cupy with adding necessary functions.

I do not fully understand how we realize the generalization of `im2col` and `col2im` with these two strategies. But if you think performance of code-generation and the complexity of its implementation are reasonable, I think code-generation is good as a first trial.

PR #1272, opened 2016-06-14
Update release cycle in the documents

beam2d:
As I noted in the release note of v1.9.1, we have decided to change the release cycle.
From the next release, we can include any feature updates that do not break the backward compatibility to any biweekly releases (i.e., we can made minor releases more frequently than once per six weeks). The versioning policy is kept as is.

I reflected this change to the documentation in this PR.


okuta:
LGTM!

PR #1273, opened 2016-06-14
fix bug in roi pooling forwarding with cpu

apple2373:
I implemented fast r-cnn https://github.com/apple2373/chainer-simple-fast-rnn, and realized roi pooling with cpu does not work. It always outputs only one roi when forwarding. I just copied two lines from GPU implementation. Also I am not sure backward function is also wrong or not because I only use forwarding now.

### DIFF
--- a/chainer/functions/pooling/roi_pooling_2d.py
+++ b/chainer/functions/pooling/roi_pooling_2d.py
@@ -185,7 +185,8 @@ class ROIPooling2D(function.Function):
 
     def backward_cpu(self, inputs, gy):
         bottom_data, bottom_rois = inputs
-        n_rois, channels, height, width = bottom_data.shape
+        channels, height, width = bottom_data.shape[1:]
+        n_rois = bottom_rois.shape[0]
         bottom_delta = numpy.zeros_like(bottom_data, dtype=numpy.float32)
 
         for i_roi in six.moves.range(n_rois):
--- a/chainer/functions/pooling/roi_pooling_2d.py
+++ b/chainer/functions/pooling/roi_pooling_2d.py
@@ -68,7 +68,8 @@ class ROIPooling2D(function.Function):
 
     def forward_cpu(self, inputs):
         bottom_data, bottom_rois = inputs
-        n_rois, channels, height, width = bottom_data.shape
+        channels, height, width = bottom_data.shape[1:]
+        n_rois = bottom_rois.shape[0]
         top_data = numpy.empty((n_rois, channels, self.outh, self.outw),
                                dtype=numpy.float32)
         self.argmax_data = numpy.empty_like(top_data).astype(numpy.int32)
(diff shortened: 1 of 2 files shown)


okuta:
Thank you for your PR.
Would you write the test?

apple2373:
Unfortunately I am not familiar with software testing or software engineering. It seems that roi pooling already has the test: https://github.com/apple2373/chainer/blob/c4b5082e3b0d2973152022aecefb07486b806f5e/tests/chainer_tests/functions_tests/pooling_tests/test_roi_pooling_2d.py

 I want to ask inputs from @wkentaro who is the developer. I guess this is the easy mistake to understand for him.

wkentaro:
The code looks good to me.
But I'm not sure how to test this.. the `test_roi_pooling_2d.py` already has `test_forward_cpu_gpu_equal` and it passed unexpectedly..

mitmul:
@apple2373 As for the test, how about modifying this line from N = 4 to N = 3 here?: [pooling_tests/test_roi_pooling_2d.py#L17](https://github.com/pfnet/chainer/blob/master/tests/chainer_tests/functions_tests/pooling_tests/test_roi_pooling_2d.py#L17)
The number of images in a minibatch and number of rois are incidentally same. That's why the test was passed I think, so please consider to try this modification.

t-abe:
mitmul's idea looks good for testing this PR.
@apple2373 Could you modify the test as he mentioned?

apple2373:
I really do not know about software testing. Is it ok just to modify the test_roi_pooling_2d.py as mitmul mentioned, and add it to this PR?

t-abe:
Yes. You can add commits into fix_roi_pooling_cpu branch, then this PR is automatically updated.

apple2373:
I did it.

t-abe:
Added test failed.
Probably we should fix same bug in `backward`.
https://github.com/apple2373/chainer/blob/cbfd52bf0d9b26aa8451c4dbf65d991b2b0ed824/chainer/functions/pooling/roi_pooling_2d.py#L188

apple2373:
I modified cpu backward in the same way.

t-abe:
Thank you.
The test still failed. I am going to check the code.
Could you do it too?
You can run a test by `$ nosetests -v tests/chainer_tests/function_tests/pooling_tests/....py`.

apple2373:
I am not sure what is going on but the test will end up "RuntimeError: CUDA environment is not correctly set up". Maybe test cannot be run only with CPU.  I'll try to run on GPU server later.

t-abe:
You can run only tests only on cpu by giving `-a '!gpu'` option to `nosetests`.

apple2373:
Thank you. I was able to reproduce test failure on my laptop, but I do not think I can fix it because I do not fully understand the implementation.

t-abe:
I fixed the bug and open PR #1352.

PR #1286, opened 2016-06-17
Add evaluator argument to Classifier

delta2323:
Current `L.Classifier` cannot change evaluator that calculates accuracy from default `Accuracy` Function with default argument. But we have other kind of accuracy such as `binary_accuracy`. Also, `Accuracy` Function itself has an argument `ignore_label`. This PR add the argument `evaluator` to `Classifier` and make the evaluator pluggable.

### DIFF
--- a/tests/chainer_tests/links_tests/model_tests/test_classifier.py
+++ b/tests/chainer_tests/links_tests/model_tests/test_classifier.py
@@ -1,0 +1,59 @@
+import unittest
+
+import numpy
+
+import chainer
+from chainer import links
+from chainer.functions.evaluation import accuracy
+from chainer import testing
+from chainer.testing import attr
+
+
+class accuracy_with_ignore_label(object):
+
+    def __call__(self, y, t):
+        return accuracy.accuracy(y, t, ignore_label=1)
+
+
+@testing.parameterize(*testing.product({
+    'evaluator': [accuracy_with_ignore_label(), None],
+    'compute_accuracy': [True, False]
+}))
+class TestClassifier(unittest.TestCase):
+
+    def setUp(self):
+        predictor = links.Linear(10, 3)
+        if self.evaluator is None:
+            self.link = links.Classifier(predictor)
+        else:
+            self.link = links.Classifier(predictor, evaluator=self.evaluator)
+        self.link.compute_accuracy = self.compute_accuracy
+
+        self.x = numpy.random.uniform(-1, 1, (5, 10)).astype(numpy.float32)
+        self.t = numpy.random.randint(3, size=(5)).astype(numpy.int32)
+
+    def check_call(self):
+        xp = self.link.xp
+        x = chainer.Variable(xp.asarray(self.x))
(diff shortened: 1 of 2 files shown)


tests/chainer_tests/links_tests/model_tests/test_classifier.py
```diff
@@ -1,22 +1,35 @@
...
 import numpy
 
 import chainer
+from chainer import functions
 from chainer import links
 from chainer import testing
 from chainer.testing import attr
 
 
+class accuracy_with_ignore_label(object):
```
muupan: I guess you needed to define a callable object instead of a function here because if a function is passed to `testing.parameterize` it would unexpectedly receive the `self` argument, and I think it is OK. But as it looks strange, I suppose you should write a comment about the reason.

muupan:

delta2323:
@muupan Thank you for your review I updated the code.

beam2d:
@delta2323 Is it OK to fix the attribute name to `accuracy`? I mean, do you think there is a user that wants to compute other measurements in the `evaluator`? The current one is OK if you think, in that case, they should write their own chain possibly by inheriting the Classifier.

delta2323:
@beam2d My original motivation of this PR is to provide user with a way to flexibly change the method that calculates accuracy. So I did not expect evaluator calculate metrics other than accuracy. So in my option, it is OK to assign the return value of `evaluator` to `accuracy`.

If we generalize the behavior of `Classifier`, it should calculate the loss value and optionally the summary of classification, (accuracy by default). I think some users do want to summarize the classification results by some metrics other than accuracy (e.g. precision, recall, F1 score etc.). But this generalization leads to the specification change of `Classifier`. So we cannot do it because of the backward compatibility. One of the possible solutions (other than your proposal) is to add another attribute like `summary` and remain the accuracy as it is. But it does not solve my original motivation.

beam2d:
Thanks for the detailed explanation. Then in that case, I recommend you to use a more precise name for the argument, other than `evaluator`, like `accuracy_func`.

delta2323:
OK

delta2323:
@beam2d @muupan I change the name from `evaluator` to `accfun`.

tests/chainer_tests/links_tests/model_tests/test_classifier.py
```diff
@@ -1,22 +1,39 @@
...
 }))
 class TestClassifier(unittest.TestCase):
 
     def setUp(self):
-        self.link = links.Classifier(chainer.Link())
+        if self.evaluator is None:
+            self.link = links.Classifier(chainer.Link())
+        else:
+            self.link = links.Classifier(chainer.Link(),
+                                         evaluator=self.evaluator)
```
muupan: `evaluator` should be replaced with `accfun` in the test code too. CI tests failed due to this line.

delta2323: Oops, sorry. I fixed it.

muupan:
LGTM.

@beam2d Could you check?

PR #1295, opened 2016-06-20
Initializer supports dtype option and initialization from shape

okuta:


chainer/initializer.py
```diff
@@ -14,7 +17,12 @@ def __call__(self, array):
 
         Args:
             array (numpy.ndarray or cupy.ndarray):
-                An array to be initialized by this initializer.
+                An array to be initialized by this initializer. If ``None``,
+                this method returns new array.
+            shape (tuple): Shape of a return array. If ``None``, this method
+                uses ``array.shape``.
+            xp (module): :mod:`cupy` or :mod:`numpy`. If ``None``, this method
+                uses ``cuda.get_array_module(array)``.
```
ir5: I feel that this documentation is insufficient. This method can either initialize the given array or return a new array. In the latter case, the sentence """Initializes given array.""" and """This method destructively changes the value of array.""" are not true.
As for `Args:` section, It is probably good to specify that `shape` and `xp` are valid only if `array` is `None`.

chainer/initializers/constant.py
```diff
@@ -17,10 +17,24 @@ class Identity(initializer.Initializer):
...
 
-    def __init__(self, scale=1.0):
+    def __init__(self, scale=1.0, **kwargs):
+        super(Identity, self).__init__(**kwargs)
         self.scale = scale
 
-    def __call__(self, array):
+    def __call__(self, array=None, shape=None, xp=None):
+        if array is None:
+            assert isinstance(shape, tuple)
```
ir5: How about add type check for `xp` as well? The same things may hold for other initializers below.

ir5:
It seems that you changed many parts of `__call__` methods in all inherited classes of `Initializer`, but is this really necessary? This implementation looks somewhat complicated due to the branching of `array is None` case and `array is not None` case. I think this can be simplified.

Suppose that we add a private method to `Initializer`, say, `_set_value`, which corresponds to the previous `__call__` method. So `_set_value` just destructively changes the value of given array.
`_set_value` a method to implement in each inherited class.
And then, suppose we change the implementation of `Initializer.__call__` so that it just calls `_set_value` if `array is not None`, and otherwise it first initializes an array appropriately according to `shape` and `xp` then calls `_set_value`.
What we should do is just to rename current `__call__` method in each inherited class to `_set_value`.

I may overlook something, but please consider this way to simplify the implementation.

okuta:
Thank you for your review.
I changed the implementation more simple.
Please review it.

chainer/initializer.py
```diff
@@ -3,6 +3,30 @@
 
 class Initializer(object):
 
+    dtype = numpy.float32
+
+    def __init__(self, dtype=None):
+        if dtype is not None:
+            self.dtype = dtype
+
+    def get_array(self, shape, xp):
```
ir5: (Minor comment) `create_array` or `make_array` look more appropriate than `get_array`.

chainer/initializer.py
```diff
@@ -16,6 +40,9 @@ def __call__(self, array):
             array (numpy.ndarray or cupy.ndarray):
                 An array to be initialized by this initializer.
 
+        Returns:
+            numpy.ndarray or cupy.ndarray: An initialized array.
```
beam2d: Does the call operator returns the array? It seems the overridden implementations do not return anything.

chainer/initializers/constant.py
```diff
@@ -17,8 +17,9 @@ class Identity(initializer.Initializer):
 
     """
 
-    def __init__(self, scale=1.0):
+    def __init__(self, scale=1.0, **kwargs):
```
beam2d: I think it is better to explicitly write down the dtype argument and add it to the docstring.

chainer/initializers/constant.py
```diff
@@ -46,23 +47,24 @@ class Constant(initializer.Initializer):
             Broadcast is allowed on this assignment.
     """
 
-    def __init__(self, fill_value):
+    def __init__(self, fill_value, **kwargs):
```
beam2d: ditto

chainer/initializers/constant.py
```diff
@@ -46,23 +47,24 @@ class Constant(initializer.Initializer):
...
 
     def __call__(self, array):
         xp = cuda.get_array_module(array)
         array[...] = xp.asarray(self.fill_value)
 
 
-def Zero():
+def Zero(dtype=None):
 
     """Returns initializer that initializes array with the all-zero array."""
```
beam2d: Add the description of dtype option.

chainer/initializers/constant.py
```diff
@@ -46,23 +47,24 @@ class Constant(initializer.Initializer):
...
 
-    return Constant(0.0)
+    return Constant(0.0, dtype=dtype)
 
 
-def One():
+def One(dtype=None):
 
     """Returns initializer that initializes array with the all-one array."""
```
beam2d: ditto

chainer/initializers/normal.py
```diff
@@ -19,8 +19,9 @@ class Normal(initializer.Initializer):
         scale(float): Standard deviation of Gaussian distribution.
     """
 
-    def __init__(self, scale=0.05):
+    def __init__(self, scale=0.05, **kwargs):
```
beam2d: ditto (explicit dtype option)

chainer/initializers/normal.py
```diff
@@ -47,13 +48,14 @@ class GlorotNormal(initializer.Initializer):
 
     """
 
-    def __init__(self, scale=1.0):
+    def __init__(self, scale=1.0, **kwargs):
```
beam2d: ditto

chainer/initializers/normal.py
```diff
@@ -74,10 +76,11 @@ class HeNormal(initializer.Initializer):
...
```
beam2d: ditto

chainer/initializers/orthogonal.py
```diff
@@ -33,8 +33,9 @@ class Orthogonal(initializer.Initializer):
 
     """
 
-    def __init__(self, scale=1.1):
+    def __init__(self, scale=1.1, **kwargs):
```
beam2d: ditto

chainer/initializers/uniform.py
```diff
@@ -20,8 +20,9 @@ class Uniform(initializer.Initializer):
 
     """
 
-    def __init__(self, scale=0.05):
+    def __init__(self, scale=0.05, **kwargs):
```
beam2d: ditto

chainer/initializers/uniform.py
```diff
@@ -47,13 +48,14 @@ class LeCunUniform(initializer.Initializer):
 
     """
 
-    def __init__(self, scale=1.0):
+    def __init__(self, scale=1.0, **kwargs):
```
beam2d: ditto

chainer/initializers/uniform.py
```diff
@@ -72,13 +74,14 @@ class GlorotUniform(initializer.Initializer):
...
```
beam2d: ditto

chainer/initializers/uniform.py
```diff
@@ -96,10 +99,11 @@ class HeUniform(initializer.Initializer):
...
```
beam2d: ditto

okuta:
Thank you for your review.
I fixed.

chainer/initializer.py
```diff
@@ -3,6 +3,30 @@
 
 class Initializer(object):
 
+    dtype = numpy.float32
+
+    def __init__(self, dtype=None):
```
ir5: It seems that `self.dtype` is used only in `generate_array` method. So, I wonder if we really need `dtype` in the constructor.
At least to me, current interface looks confusing. Suppose that we specify `dtype=np.float32` in the constructor while we call `__call__` method with an array of `dtype=np.float64`. In this situation, it is not clear if the initialized array's `dtype` is `np.float32` or `np.float64`. Actually, it will depend on the implementation of each inherited class.
I suggest we remove `dtype` argument here, and make `generate_array` have the `dtype` argument instead. (Sorry, I should have said this earlier...)

chainer/initializer.py
```diff
@@ -3,6 +3,9 @@
 
 class Initializer(object):
 
+    def __init__(self, dtype=None):
```
ir5: Please specify what is expected to `dtype` argument here. Unless I overlook something, it is for type check of `__call__` functions and `generate_array` function.

chainer/initializers/__init__.py
```diff
@@ -20,6 +20,30 @@
 HeUniform = uniform.HeUniform
 
 
+def generate_array(initializer, shape, xp):
+    """Return initialized array.
+
+    The algorithms used to make the new values depend on the
+    concrete derived classes.
```
ir5: Please note that the `dtype` of a generated array depends on `initializer.dtype`.

okuta:
@ir5 Thank you for your review. I fixed.

chainer/initializers/__init__.py
```diff
@@ -20,6 +20,31 @@
...
 
+def generate_array(initializer, shape, xp):
+    """Return initialized array.
+
+    The algorithms used to make the new values depend on the
+    concrete derived classes. The dtype of a generated array depends on
+    ``initializer.dtype``.
+
+    Args:
+        initializer: The callable object that takes :class:`numpy.ndarray`
```
beam2d: The -> A

beam2d:
LGTM except one small comment

beam2d:
LGTM!

PR #1297, opened 2016-06-21
Avoid division by zero

hitottiez:
In Contrastive.backward(), division by zero occurs when ```dist``` is equal to zero.

```
gx0 += alpha * (1 - y) * mdist_p * mdist * -(self.diff / dist)
```

It causes that gradient value is nan.

Please see my proposed modification.

### DIFF
--- a/tests/chainer_tests/functions_tests/loss_tests/test_contrastive.py
+++ b/tests/chainer_tests/functions_tests/loss_tests/test_contrastive.py
@@ -82,18 +82,12 @@ class TestContrastive(unittest.TestCase):
 
     @condition.retry(3)
     def test_backward_zero_dist_cpu(self):
-        x0 = self.x0
-        x1 = self.x0
-        self.dist = xp.sqrt(numpy.sum((x0 - x1) ** 2, axis=1))
-        self.check_backward(x0, x1, self.t)
+        self.check_backward(self.x0, self.x0, self.t)
 
     @attr.gpu
     @condition.retry(3)
     def test_backward_zero_dist_gpu_no_cudnn(self):
-        x0 = self.x0
-        x1 = self.x0
-        self.dist = cuda.to_gpu(numpy.sqrt(numpy.sum((x0 - x1) ** 2, axis=1)))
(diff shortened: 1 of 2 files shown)


okuta:
Please add test case what ``dist`` is zero.

okuta:
Commits are strange.
Would you organize commits?

hitottiez:
I added test cases as follows.
e0e95e5cb2aac451f62510d3e0f0c8e3b66cf383

tests/chainer_tests/functions_tests/loss_tests/test_contrastive.py
```diff
@@ -80,5 +80,20 @@ def test_backward_gpu_no_cudnn(self):
         self.check_backward(cuda.to_gpu(self.x0), cuda.to_gpu(self.x1),
                             cuda.to_gpu(self.t))
 
+    @condition.retry(3)
+    def test_backward_zero_dist_cpu(self):
+        x0 = self.x0
+        x1 = self.x0
+        self.dist = xp.sqrt(numpy.sum((x0 - x1) ** 2, axis=1))
+        self.check_backward(x0, x1, self.t)
```
okuta: How about `self.check_backward(self.x0, self.x0, self.t)`?

tests/chainer_tests/functions_tests/loss_tests/test_contrastive.py
```diff
@@ -80,5 +80,20 @@ def test_backward_gpu_no_cudnn(self):
...
+
+    @attr.gpu
+    @condition.retry(3)
+    def test_backward_zero_dist_gpu_no_cudnn(self):
+        x0 = self.x0
+        x1 = self.x0
+        self.dist = cuda.to_gpu(numpy.sqrt(numpy.sum((x0 - x1) ** 2, axis=1)))
+        self.check_backward(cuda.to_gpu(x0), cuda.to_gpu(x1),
```
okuta: ditto

hitottiez:

I modified test cases as follows.
ebbfa7b33b314715e43028c630cbfa7816301b54

okuta:
LGTM

PR #1298, opened 2016-06-21
Restore gradient value when Variable is unPickled

hitottiez:
We want to use Chainer in multi-process environment. We have tried to use multiprocessing.Queue to share a model object (```chainer.Chain```) between processes. However, this strategy didn't work because gradient value was not restored when ```Variable``` instance was unPickled.

This pull request wants to fix the problem by keeping gradient values after unPickling ```Variable```.

Please see below for the details.
https://groups.google.com/forum/#!topic/chainer-jp/KneGQWbaPtw


t-abe:
Thank you for the PR.
Could you make a test to check pickled & unpickled Variable has .grad correctly?

hitottiez:
We added tests to check pickled & unpickled Variable has .grad correctly.
Please see our modified code.

t-abe:
Thank you.  LGTM :smile:

chainer/variable.py
```diff
@@ -86,7 +86,7 @@ class Variable(object):
             :class:`~chainer.Flag` for the detail of ternary flags.
 
     """
-    def __init__(self, data, volatile=flag.OFF, name=None):
+    def __init__(self, data, volatile=flag.OFF, name=None, grad=None):
```
beam2d: Would you add a document for the new argument?

hitottiez: We added a docstring for the new argument.
Please check it.

beam2d:
LGTM

beam2d:
Thank you!

PR #1307, opened 2016-06-22
Fix document on multi-GPU test

delta2323:
As nose's document [says](http://nose.readthedocs.io/en/latest/plugins/attrib.html#expression-evaluation), we must use `--eval-attr` option to handle expression involving inequality.


bkvogel:
This looks good to met.

PR #1309, opened 2016-06-22
Make _check_available() function public.

BlGene:
Hi chainer team,

its a good idea to check if the serializer works _before_ starting training, this information should be exposed to the user. 

I didn't change the `__init__.py` file though.

BR, Max


beam2d:
Good, thank you. Would you write a docstring and add it to the reference document?

PR #1310, opened 2016-06-23
Add cuDNN support to batch normalization function

bkvogel:
I have added cuDNN support to the batch normalization function and updated the corresponding link. This involved writing Python wrappers for the new cuDNN functions and using the cuDNN implementation if cuDNN is enabled. Since cuDNN updates the running average statistics (mean and variance), I also moved these computations from the link into the function.


okuta:
Why don’t you support float16 and float64?

bkvogel:
I disabled float16 because the tests failed using cuDNN. I see other functions such as relu.py allow it to run anyway if the cudnn version is at least 3000. I didn't realize I did not support float64. Could you give me an example of another function that handles all float types correctly so that I could use it as a template? Thanks.

bkvogel:
I have modified the tests to be similar to the newer batch normalization tests in master that check all float types. float16 and float64 types are tested now. I still do not allow the cuDNN version to use float16 since it fails the test on my machine using cuDNN v5. So if float16 is detected, cupy version is used instead. How does this look?

chainer/functions/normalization/batch_normalization.py
```diff
@@ -4,20 +4,53 @@
...
+    elif arr.ndim == 4:
+        return arr
+    else:
+        return arr.reshape(arr.shape[0], -1, 1, 1)
+
 
 class BatchNormalizationFunction(function.Function):
 
-    def __init__(self, eps=1e-5):
+    def __init__(self, mean, var, train, decay=0.9, eps=2e-5, use_cudnn=True):
```
okuta: Please do not change argument order.

chainer/functions/normalization/batch_normalization.py
```diff
@@ -27,100 +60,266 @@ def check_type_forward(self, in_types):
...
-                mean_type.dtype == x_type.dtype,
-                mean_type.shape == gamma_type.shape,
-                var_type.dtype == x_type.dtype,
-                var_type.shape == gamma_type.shape,
-            )
-
     def forward(self, inputs):
         xp = cuda.get_array_module(*inputs)
-        x, gamma, beta = inputs[:3]
+        x, gamma, beta = inputs
```
okuta: Do not delete forward function argument for backward compatibility.

chainer/functions/normalization/batch_normalization.py
```diff
@@ -27,100 +60,266 @@ def check_type_forward(self, in_types):
...
 
-        self.std = xp.sqrt(var, dtype=var.dtype)
+        if (x.ndim == 2) or (x.ndim == 4):
+            # cuDNN only supports these tensor dimensions because they are
+            # the most commonly used. If there is a need to support other
+            # dimensions with cuDNN, we could consider reshaping the input
+            # into a 2-dim array with channels as second dim and m=<product
+            # of all dimensions except the 2nd dimension> as the first
+            # dimension.
+            self.cudnn_dim_ok = True
```
okuta: How about `self.cudnn_dim_ok = x.ndim == 2 or x.ndim == 4`?

chainer/functions/normalization/batch_normalization.py
```diff
@@ -27,100 +60,266 @@ def check_type_forward(self, in_types):
...
+                var += self.eps
+            else:
+                mean = self.fixed_mean
+                var = self.fixed_var
+            self.std = xp.sqrt(var, dtype=var.dtype)
             x_mu = x - mean[expander]
             self.x_hat = x_mu / self.std[expander]
             y = gamma * self.x_hat
             y += beta
+        elif cuda.cudnn_enabled and self.use_cudnn and self.cudnn_dim_ok:
```
okuta: Please check cuDNN version. Chainer supports cuDNN from v2 to v5.

chainer/functions/normalization/batch_normalization.py
```diff
@@ -27,100 +60,266 @@ def check_type_forward(self, in_types):
...
+                                              x_desc.value, self.mode)
+            one = numpy.array(1, dtype=dtype).ctypes
+            zero = numpy.array(0, dtype=dtype).ctypes
+            y = cuda.cupy.empty_like(x)
+            # Factor used in the moving average
+            factor = 1 - self.decay
+            # computation runningMean
+            if self.mean_cache is None:
+                # Output cache to speed up bacward pass (recommended to enable)
+                self.mean_cache = xp.zeros_like(gamma)
```
okuta: How about using `empty_like`?

chainer/functions/normalization/batch_normalization.py
```diff
@@ -27,100 +60,266 @@ def check_type_forward(self, in_types):
...
+                # Output cache to speed up bacward pass (recommended to enable)
+                self.var_cache = xp.zeros_like(gamma)
```
okuta: How about using `empty_like`?

chainer/functions/normalization/batch_normalization.py
```diff
@@ -27,100 +60,266 @@ def check_type_forward(self, in_types):
...
+            del temp_ar
         return y,
 
     def backward(self, inputs, grad_outputs):
         x, gamma = inputs[:2]
         gy = grad_outputs[0]
 
+        if x[0].dtype == numpy.float16:
+            # cuDNN v5 batch normalization does not seem to support float16.
+            self.use_cudnn = False
```
okuta: `self.use_cudnn` had assigned at `forward`.

chainer/functions/normalization/batch_normalization.py
```diff
@@ -27,100 +60,266 @@ def check_type_forward(self, in_types):
...
+                    'T gy, T x_hat, T gamma, T std',
+                    'T gx',
+                    'gx = (gamma / std) * gy',
+                    'bn_bwd')(gy, self.x_hat, gamma[expander],
+                              self.std[expander])
         return gx, ggamma, gbeta
 
 
-def batch_normalization(x, gamma, beta, eps=1e-5):
+def batch_normalization(x, gamma, beta, running_mean=None,
```
okuta: Do not change argument order for backward compatibility.

chainer/functions/normalization/batch_normalization.py
```diff
@@ -129,28 +328,30 @@ def batch_normalization(x, gamma, beta, eps=1e-5):
     .. seealso:: :class:`links.BatchNormalization`
 
     """
-    return BatchNormalizationFunction(eps)(x, gamma, beta)
+    return BatchNormalizationFunction(running_mean, running_var, True,
+                                      decay, eps)(x, gamma, beta)
 
 
-def fixed_batch_normalization(x, gamma, beta, mean, var, eps=1e-5):
+def fixed_batch_normalization(x, gamma, beta, fixed_mean, fixed_var, eps=2e-5):
```
okuta: Do not change arguments name for backward compatibility.

chainer/links/normalization/batch_normalization.py
```diff
@@ -101,7 +100,8 @@ def __call__(self, x, test=False, finetune=False):
 
         """
         use_batch_mean = not test or finetune
-
+        print('use_batch_mean = ', use_batch_mean)
```
okuta: Please remove this line.

chainer/links/normalization/batch_normalization.py
```diff
@@ -114,31 +114,23 @@ def __call__(self, x, test=False, finetune=False):
...
-                adjust = m / max(m - 1., 1.)  # unbiased estimation
-                self.avg_mean *= decay
-                func.mean *= 1 - decay  # reuse buffer as a temporary
-                self.avg_mean += func.mean
-                del func.mean
-                self.avg_var *= decay
-                func.var *= (1 - decay) * adjust  # reuse buffer as a temporary
-                self.avg_var += func.var
-                del func.var
+            print('decay = ', decay)
```
okuta: Please remove this line.

tests/chainer_tests/functions_tests/normalization_tests/test_batch_normalization.py
```diff
@@ -132,19 +140,22 @@ def test_forward_cpu(self):
     @attr.gpu
     @condition.retry(3)
     def test_forward_gpu(self):
+        print('In: test_forward_gpu()...')
```
okuta: Please remove this line.

cupy/cuda/cudnn.pyx
```diff
@@ -585,6 +618,71 @@ cpdef poolingBackward(
...
+        size_t resultRunningMean, size_t resultRunningVariance,
+        float epsilon, size_t resultSaveMean, size_t resultSaveInvVariance):
+        status = cudnnBatchNormalizationForwardTraining(
+        <Handle>handle, <BatchNormMode> mode,
+        <void*>alpha, <void*>beta, <TensorDescriptor>xDesc,
+        <void*>x, <TensorDescriptor>yDesc, <void*>y,
+        <TensorDescriptor>bnScaleBiasMeanVarDesc, <void*>bnScale,
+        <void*>bnBias, <float>exponentialAverageFactor,
+        <void*>resultRunningMean, <void*>resultRunningVariance,
+        <float>epsilon, <void*>resultSaveMean, <void*>resultSaveInvVariance)
```
okuta: `<double>epsilon` is correct.

cupy/cuda/cudnn.pyx
```diff
@@ -585,6 +618,71 @@ cpdef poolingBackward(
...
+        size_t bnScaleBiasMeanVarDesc, size_t bnScale,
+        size_t bnBias, float exponentialAverageFactor,
+        size_t resultRunningMean, size_t resultRunningVariance,
+        float epsilon, size_t resultSaveMean, size_t resultSaveInvVariance):
+        status = cudnnBatchNormalizationForwardTraining(
+        <Handle>handle, <BatchNormMode> mode,
+        <void*>alpha, <void*>beta, <TensorDescriptor>xDesc,
+        <void*>x, <TensorDescriptor>yDesc, <void*>y,
+        <TensorDescriptor>bnScaleBiasMeanVarDesc, <void*>bnScale,
+        <void*>bnBias, <float>exponentialAverageFactor,
```
okuta: `<double>exponentialAverageFactor` is correct.

cupy/cuda/cudnn.pxd
```diff
@@ -235,6 +236,39 @@ cpdef poolingBackward(
...
+
+cpdef deriveBNTensorDescriptor(
+        size_t derivedBnDesc, size_t xDesc, int mode)
+
+cpdef batchNormalizationForwardTraining(
+        size_t handle, int mode,
+        size_t alpha, size_t beta, size_t xDesc,
+        size_t x, size_t yDesc, size_t y,
+        size_t bnScaleBiasMeanVarDesc, size_t bnScale,
+        size_t bnBias, float exponentialAverageFactor,
```
okuta: `double exponentialAverageFactor` is correct.

cupy/cuda/cudnn.pxd
```diff
@@ -235,6 +236,39 @@ cpdef poolingBackward(
...
+        size_t resultRunningMean, size_t resultRunningVariance,
+        float epsilon, size_t resultSaveMean, size_t resultSaveInvVariance)
```
okuta: `double epsilon` is correct.

cupy/cuda/cudnn.pxd
```diff
@@ -235,6 +236,39 @@ cpdef poolingBackward(
...
+
+cpdef batchNormalizationForwardInference(
+        size_t handle, int mode,
+        size_t alpha, size_t beta, size_t xDesc,
+        size_t x, size_t yDesc, size_t y,
+        size_t bnScaleBiasMeanVarDesc, size_t bnScale,
+        size_t bnBias, size_t estimatedMean, size_t estimatedVariance,
+        float epsilon)
```
okuta: `double epsilon` is correct.

cupy/cuda/cudnn.pxd
```diff
@@ -235,6 +236,39 @@ cpdef poolingBackward(
...
+
+cpdef batchNormalizationBackward(
+        size_t handle, int mode,
+        size_t alphaDataDiff, size_t betaDataDiff,
+        size_t alphaParamDiff, size_t betaParamDiff,
+        size_t xDesc, size_t x, size_t dyDesc,
+        size_t dy, size_t dxDesc, size_t dx,
+        size_t dBnScaleBiasDesc, size_t bnScale,
+        size_t dBnScaleResult, size_t dBnBiasResult,
+        float epsilon, size_t savedMean, size_t savedInvVariance)
```
okuta: `double epsilon` is correct.

cupy/cuda/cudnn.pyx
```diff
@@ -151,6 +151,39 @@ cdef extern from "cupy_cudnn.h":
...
+    int cudnnDeriveBNTensorDescriptor(
+    	    TensorDescriptor derivedBnDesc, TensorDescriptor xDesc,
+	    BatchNormMode mode)
+
+    int cudnnBatchNormalizationForwardTraining(
+            Handle handle, BatchNormMode mode,
+	    void* alpha, void* beta, TensorDescriptor xDesc,
+	    void* x, TensorDescriptor yDesc, void* y,
+	    TensorDescriptor bnScaleBiasMeanVarDesc, void* bnScale,
+	    void* bnBias, float exponentialAverageFactor,
```
okuta: Type miss. `double` is correct.

cupy/cuda/cudnn.pyx
```diff
@@ -151,6 +151,39 @@ cdef extern from "cupy_cudnn.h":
...
+	    void* resultRunningMean, void* resultRunningVariance,
+	    float epsilon, void* resultSaveMean, void* resultSaveInvVariance)
```
okuta: Type miss. `double` is correct.

cupy/cuda/cudnn.pyx
```diff
@@ -151,6 +151,39 @@ cdef extern from "cupy_cudnn.h":
...
+
+    int cudnnBatchNormalizationForwardInference(
+            Handle handle, BatchNormMode mode,
+	    void* alpha, void* beta, TensorDescriptor xDesc,
+	    void* x, TensorDescriptor yDesc, void* y,
+	    TensorDescriptor bnScaleBiasMeanVarDesc, void* bnScale,
+	    void* bnBias, void* estimatedMean, void* estimatedVariance,
+	    float epsilon)
```
okuta: Type miss. `double` is correct.

cupy/cuda/cudnn.pyx
```diff
@@ -151,6 +151,39 @@ cdef extern from "cupy_cudnn.h":
...
+    int cudnnBatchNormalizationBackward(
+            Handle handle, BatchNormMode mode,
+	    void* alphaDataDiff, void* betaDataDiff,
+	    void* alphaParamDiff, void* betaParamDiff,
+	    TensorDescriptor xDesc, void* x,
+	    TensorDescriptor dyDesc, void* dy,
+	    TensorDescriptor dxDesc, void* dx,
+	    TensorDescriptor dBnScaleBiasDesc, void* bnScale,
+	    void* dBnScaleResult, void* dBnBiasResult,
+	    float epsilon, void* savedMean, void* savedInvVariance)
```
okuta: Type miss. `double` is correct.

cupy/cuda/cudnn.pyx
```diff
@@ -585,6 +618,71 @@ cpdef poolingBackward(
...
+        <TensorDescriptor>derivedBnDesc, <TensorDescriptor>xDesc,
+        <BatchNormMode> mode)
+    check_status(status)
+
+cpdef batchNormalizationForwardTraining(
+        size_t handle, int mode,
+        size_t alpha, size_t beta, size_t xDesc,
+        size_t x, size_t yDesc, size_t y,
+        size_t bnScaleBiasMeanVarDesc, size_t bnScale,
+        size_t bnBias, float exponentialAverageFactor,
```
okuta: Type miss. `double` is correct.

cupy/cuda/cudnn.pyx
```diff
@@ -585,6 +618,71 @@ cpdef poolingBackward(
...
+        size_t resultRunningMean, size_t resultRunningVariance,
+        float epsilon, size_t resultSaveMean, size_t resultSaveInvVariance):
```
okuta: Type miss. `double` is correct.

cupy/cuda/cudnn.pyx
```diff
@@ -585,6 +618,71 @@ cpdef poolingBackward(
...
+        size_t bnScaleBiasMeanVarDesc, size_t bnScale,
+        size_t bnBias, size_t estimatedMean, size_t estimatedVariance,
+        float epsilon):
+        status = cudnnBatchNormalizationForwardInference(
+        <Handle>handle, <BatchNormMode> mode,
+        <void*>alpha, <void*>beta, <TensorDescriptor>xDesc,
+        <void*>x, <TensorDescriptor>yDesc, <void*>y,
+        <TensorDescriptor>bnScaleBiasMeanVarDesc, <void*>bnScale,
+        <void*>bnBias, <void*>estimatedMean, <void*>estimatedVariance,
+	    <float>epsilon)
```
okuta: `<float>` is not needed.

cupy/cuda/cudnn.pyx
```diff
@@ -585,6 +618,71 @@ cpdef poolingBackward(
...
+        status = cudnnBatchNormalizationBackward(
+        <Handle>handle, <BatchNormMode>mode,
+	    <void*>alphaDataDiff, <void*>betaDataDiff,
+	    <void*>alphaParamDiff, <void*>betaParamDiff,
+	    <TensorDescriptor>xDesc, <void*>x,
+	    <TensorDescriptor>dyDesc, <void*>dy,
+	    <TensorDescriptor>dxDesc, <void*>dx,
+	    <TensorDescriptor>dBnScaleBiasDesc, <void*>bnScale,
+	    <void*>dBnScaleResult, <void*>dBnBiasResult,
+	    <float>epsilon, <void*>savedMean, <void*>savedInvVariance)
```
okuta: `<float>` is not needed.

cupy/cuda/cudnn.pyx
```diff
@@ -585,6 +618,71 @@ cpdef poolingBackward(
...
+        <float>epsilon, <void*>resultSaveMean, <void*>resultSaveInvVariance)
+        check_status(status)
+
+cpdef batchNormalizationForwardInference(
+        size_t handle, int mode,
+        size_t alpha, size_t beta, size_t xDesc,
+        size_t x, size_t yDesc, size_t y,
+        size_t bnScaleBiasMeanVarDesc, size_t bnScale,
+        size_t bnBias, size_t estimatedMean, size_t estimatedVariance,
+        float epsilon):
```
okuta: Type miss. `double` is correct.

cupy/cuda/cudnn.pyx
```diff
@@ -585,6 +618,71 @@ cpdef poolingBackward(
...
+
+cpdef batchNormalizationBackward(
+        size_t handle, int mode,
+        size_t alphaDataDiff, size_t betaDataDiff,
+        size_t alphaParamDiff, size_t betaParamDiff,
+        size_t xDesc, size_t x, size_t dyDesc,
+        size_t dy, size_t dxDesc, size_t dx,
+        size_t dBnScaleBiasDesc, size_t bnScale,
+        size_t dBnScaleResult, size_t dBnBiasResult,
+        float epsilon, size_t savedMean, size_t savedInvVariance):
```
okuta: Type miss. `double` is correct.

okuta:
Please add with/without cuDNN tests.
See also
https://github.com/pfnet/chainer/blob/master/tests/chainer_tests/functions_tests/activation_tests/test_relu.py#L49
https://github.com/pfnet/chainer/blob/master/tests/chainer_tests/functions_tests/activation_tests/test_relu.py#L57


And, please check calling cuDNN.
See also
https://github.com/pfnet/chainer/blob/master/tests/chainer_tests/functions_tests/activation_tests/test_relu.py#L90

okuta:
> I have modified the tests to be similar to the newer batch normalization tests in master that check all float types. float16 and float64 types are tested now. I still do not allow the cuDNN version to use float16 since it fails the test on my machine using cuDNN v5. So if float16 is detected, cupy version is used instead. How does this look?

I tested float16. It fails too...
This PR works to use float32 and float64.

chainer/functions/normalization/batch_normalization.py
bkvogel: Thank you for the comment. I was just made aware that we should try to maintain backward compatibility on any documented functions, even if they will probably only ever be used by the corresponding link. I will try to rewrite to maintain backward compatibility.

chainer/functions/normalization/batch_normalization.py
bkvogel: OK, I will rewrite to keep "eps" as first default arg.

bkvogel:
Thank you for the detailed comments. I will make the suggested changes.

chainer/functions/normalization/batch_normalization.py
bkvogel: This function was never documented, though. Is it okay to remove the optional arguments as long as I do not brake any of the documented APIs?

bkvogel:
I have implemented the requested changes and added additional tests (calling cuDNN with/without cuDNN). I fixed the batch normalization function so that backwards compatibility with existing documented APIs is not broken. Could you check if it looks okay now? Thank you.

tests/chainer_tests/functions_tests/normalization_tests/test_batch_normalization.py
```diff
@@ -134,20 +149,69 @@ def test_forward_cpu(self):
...
+        shape = (7, 3) + (2,) * ndim
+        self.x = cuda.cupy.random.uniform(-1, 1, shape).astype(self.dtype)
+        self.gy = cuda.cupy.random.uniform(-1, 1, shape).astype(self.dtype)
+        self.args = [self.x, self.gamma, self.beta]
+        self.aggr_axes = (0,) + tuple(six.moves.range(2, ndim + 2))
+        self.mean = self.x.mean(axis=self.aggr_axes)
+        self.var = self.x.var(axis=self.aggr_axes) + self.eps
+        self.expect = self.use_cudnn and (
+            cuda.cudnn.cudnn.getVersion() >= 5000 or
+            self.dtype != numpy.float16)
```
okuta: `self.dtype` is not `numpy.float16` always.

chainer/functions/normalization/batch_normalization.py
```diff
@@ -27,130 +66,294 @@ def check_type_forward(self, in_types):
...
         eps (float): Epsilon value for numerical stability.
 
     .. seealso::
        :func:`functions.batch_normalization`,
        :class:`links.BatchNormalization`
 
     """
-    return BatchNormalizationFunction(eps)(x, gamma, beta, mean, var)
+    return BatchNormalizationFunction(eps, mean, var, False,
+                                      0.0, use_cudnn)(x, gamma, beta)
```
okuta: The gradients of `mean` and `var` is not calculated.
It is different from previous version.

bkvogel: OK, I will change to back so that `mean` and `var` are passed as in the existing version. I don't see the need to compute the gradients of these (and cuDNN does not compute them and so the revised code will be a bit slower) but I agree it is good not to break compatibility.

okuta: Actually, `links.BatchNormalization` dose not need gradient of `mean` and `var` .
I think reducing extra calculation is OK. For example, adding a control flag.

bkvogel: Yes, since the BN link does not need the gradients of `mean` and `var`, I think it should be possible to maintain backward compatibility of the BN function while also reducing computation by not computing these gradients. I will try this.

chainer/functions/normalization/batch_normalization.py
okuta: Sorry, This comment was very confusing.

We should not keep backward compatibility of undocumented functionality.
But, `fixed_batch_normalization` wants gradients of `mean` and `var`.
And the computational graph includes `mean` and `var` Variable in that case.

bkvogel:
I have modified the BN function so that it now computes the gradients of all Variables, maintaining backward compatibility with the existing version. The (slow) code for computing the gradients of `mean' and `var` has been moved to the backward() function (which is not used by the BN link) and so the forward() call will use cuDNN if available and should be fast.
Could you check again if it looks ok now?

chainer/functions/normalization/batch_normalization.py
```diff
@@ -42,105 +73,289 @@ def forward(self, inputs):
...
+            derivedBnDesc = cudnn.create_uninitialized_tensor_descriptor()
+            libcudnn.deriveBNTensorDescriptor(derivedBnDesc.value,
+                                              x_desc.value, self.mode)
+            one = numpy.array(1, dtype=dtype).ctypes
+            zero = numpy.array(0, dtype=dtype).ctypes
+            y = cuda.cupy.empty_like(x)
+            # Factor used in the moving average
+            factor = 1 - self.decay
+            # computation runningMean
+            if self.mean_cache is None:
```
okuta: How about moving this block into `if self.train` block?
https://github.com/pfnet/chainer/pull/1310/files#diff-148edea535bb806260c487ca31869a91R152

chainer/functions/normalization/batch_normalization.py
```diff
@@ -42,105 +73,289 @@ def forward(self, inputs):
...
+                              self.std[expander],
+                              ggamma[expander], gbeta[expander], inv_m)
+            else:
+                # Note: Under normal conditions, this will never be executed
+                # because fixed-mean-variance mode is normally
+                # only used at test/evaluation time.
+                # If cuDNN was called on the forward call, self.x_hat will
+                # be None.
+                if self.x_hat is None:
+                    self.std = xp.sqrt(self.fixed_var,
```
okuta: How about writing function to calculate `std` and `x_hat`.
There are similar codes.
https://github.com/pfnet/chainer/pull/1310/files#diff-148edea535bb806260c487ca31869a91R116
https://github.com/pfnet/chainer/pull/1310/files#diff-148edea535bb806260c487ca31869a91R179
https://github.com/pfnet/chainer/pull/1310/files#diff-148edea535bb806260c487ca31869a91R220

chainer/functions/normalization/batch_normalization.py
```diff
@@ -42,105 +73,289 @@ def forward(self, inputs):
...
+            else:
+                mean = self.fixed_mean
+                var = self.fixed_var
+            self.std = xp.sqrt(var, dtype=var.dtype)
             x_mu = x - mean[expander]
             self.x_hat = x_mu / self.std[expander]
             y = gamma * self.x_hat
             y += beta
+            # cuDNN Batch normalization requires v5 or greater.
+        elif cuda.cudnn_enabled and self.use_cudnn and self.cudnn_dim_ok and \
```
okuta: How about changing to following style?
``` python
if (xp is not numpy and cuda.cudnn_enabled and self.use_cudnn and self.cudnn_dim_ok and _cudnn_version >= 5000):
     ~~~~~
else:
    if self.train:
        axis = (0,) + tuple(range(head_ndim, x.ndim))
        mean = x.mean(axis=axis)
        var = x.var(axis=axis)
        var += self.eps
    else:
        mean = self.fixed_mean
        var = self.fixed_var
        self.std = xp.sqrt(var, dtype=var.dtype)
        if xp is numpy:
        ~~~~~
````

File: chainer/functions/normalization/batch_normalization.py [hunk 34]
```diff
@@ -42,105 +73,289 @@ def forward(self, inputs):
...
+    access the running_mean and/or running_var attributes. See the
+    corresponding Link class for an example of how to do this.
 
     Args:
         x (Variable): The input variable.
         gamma (Variable): The scaling parameter of normalized data.
         beta (Variable): The shifting parameter of scaled normalized data.
         eps (float): Epsilon value for numerical stability.
+        running_mean (array): The running average of the mean. This is a
+        running average of the mean over several
```
okuta: Please fix an indent.

chainer/functions/normalization/batch_normalization.py
```diff
@@ -42,105 +73,289 @@ def forward(self, inputs):
...
+            mini-batches using the decay parameter. If None, the running
```
okuta: Please use ```` ``None`` ````.

chainer/functions/normalization/batch_normalization.py
```diff
@@ -42,105 +73,289 @@ def forward(self, inputs):
...
+            average is not computed. If this is None,
```
okuta: ditto

chainer/functions/normalization/batch_normalization.py
```diff
@@ -42,105 +73,289 @@ def forward(self, inputs):
...
+            then runnng_var must also be None.
+        running_var (array): The running average of the variance. This is a
+        running average of the variance
```
okuta: Please fix an indent.

chainer/functions/normalization/batch_normalization.py
```diff
@@ -42,105 +73,289 @@ def forward(self, inputs):
...
+            over several mini-batches using the decay parameter. If None, the
```
okuta: Please use ```` ``None`` ````.

chainer/functions/normalization/batch_normalization.py
```diff
@@ -42,105 +73,289 @@ def forward(self, inputs):
...
+            running average is not computed. If
+            this is None, then running_mean must also be None.
```
okuta: ditto

chainer/functions/normalization/batch_normalization.py
```diff
@@ -42,105 +73,289 @@ def forward(self, inputs):
...
-def fixed_batch_normalization(x, gamma, beta, mean, var, eps=1e-5):
+def fixed_batch_normalization(x, gamma, beta, mean, var, eps=2e-5,
+                              use_cudnn=True):
     """Batch normalization function with fixed statistics.
 
     This is a variant of batch normalization, where the mean and variance
-    statistics are given by the caller as variables. This is used on testing
-    mode of the batch normalization layer, where batch statistics cannot be
-    used for prediction consistency.
+    statistics are given by the caller as fixed parameters. This is
```
okuta: Is this fix needed?

okuta:
How about changing below statement?
https://github.com/pfnet/chainer/pull/1310/files#diff-0fb203b8a923ff41fabe22b3b10508fbR132
`BatchNormalization` do not need gradients of `mean` and `bar`.

chainer/functions/normalization/batch_normalization.py
```diff
@@ -42,105 +73,289 @@ def forward(self, inputs):
...
-            self.var = var
+        if self.train:
+            if self.running_mean is None:
+                self.running_mean = xp.zeros_like(gamma)
+                self.running_var = xp.zeros_like(gamma)
+            else:
+                self.running_mean = xp.array(self.running_mean)
+                self.running_var = xp.array(self.running_var)
+        elif len(inputs) == 5:
+                self.fixed_mean = xp.array(inputs[3])
```
okuta: How about `self.fixed_mean = inputs[3]`?

chainer/functions/normalization/batch_normalization.py
bkvogel: Yes, this will reduce code duplication. I will make the change.

chainer/functions/normalization/batch_normalization.py
bkvogel: Not needed. I will change wording back to "variables" instead of "parameters."

chainer/functions/normalization/batch_normalization.py
bkvogel: OK.

chainer/functions/normalization/batch_normalization.py
```diff
@@ -4,19 +4,56 @@
...
+    elif arr.ndim == 4:
+        return arr
+    else:
+        return arr.reshape(arr.shape[0], -1, 1, 1)
+
+
+def _xhat(x, mean, std, expander):
+    x_mu = x - mean[expander]
+    # Return x_hat
+    return x_mu / std[expander]
```
okuta: How about following code?
```python
x_mu /= std[expander] # reduce memory usage
return x_mu
```

chainer/links/normalization/batch_normalization.py
```diff
@@ -101,7 +100,7 @@ def __call__(self, x, test=False, finetune=False):
 
         """
         use_batch_mean = not test or finetune
-
+        # use_batch_mean = (not test) or finetune
```
okuta: Please remove this line.

okuta:
LGTM except comments!

bkvogel:
Thank you for the comments. I have made the requested changes.

cupy/cuda/cupy_cudnn.h
```diff
@@ -118,6 +118,76 @@ cudnnStatus_t cudnnSetConvolutionNdDescriptor_v3(
 #define cudnnSetPooling2dDescriptor_v3 cudnnSetPooling2dDescriptor
 #define cudnnSetPoolingNdDescriptor_v3 cudnnSetPoolingNdDescriptor
 
+cudnnStatus_t cudnnDeriveBNTensorDescriptor(
+         cudnnTensorDescriptor_t derivedBnDesc,
+         const cudnnTensorDescriptor_t xDesc,
+         cudnnBatchNormMode_t mode) {
```
okuta: Please define `cudnnBatchNormMode_t`.
Build is failed when CuPy uses lower than cuDNN v5.

okuta:
LGTM!

jekbradbury:
I'm seeing cuDNN errors when using batchnorm at inference time for 2D tensors, on both cuDNN 5.0 and 5.1.

```
  File "/home/james.bradbury/chainer-dev/chainer/chainer/links/normalization/batch_normalization.py", line 132, in __call__
    x, gamma, beta, mean, var, self.eps)
  File "/home/james.bradbury/chainer-dev/chainer/chainer/functions/normalization/batch_normalization.py", line 347, in fixed_batch_normalization
    use_cudnn)(x, gamma, beta, mean, var)
  File "/home/james.bradbury/chainer-dev/chainer/chainer/function.py", line 130, in __call__
  File "/home/james.bradbury/chainer-dev/chainer/chainer/functions/normalization/batch_normalization.py", line 157, in forward
    self.eps)
  File "cupy/cuda/cudnn.pyx", line 655, in cupy.cuda.cudnn.batchNormalizationForwardInference (cupy/cuda/cudnn.cpp:8636)
  File "cupy/cuda/cudnn.pyx", line 669, in cupy.cuda.cudnn.batchNormalizationForwardInference (cupy/cuda/cudnn.cpp:8439)
  File "cupy/cuda/cudnn.pyx", line 243, in cupy.cuda.cudnn.check_status (cupy/cuda/cudnn.cpp:1507)
cupy.cuda.cudnn.CuDNNError: CUDNN_STATUS_NOT_SUPPORTED: b'CUDNN_STATUS_NOT_SUPPORTED'
```
This is the error I get for both `_cudnn_version=v5005` and `_cudnn_version=v5105`. The cuDNN User Guide for v5.1 says (for all three batchnorm API functions):
> Only 4D and 5D tensors are supported.

And its description of `CUDNN_BATCHNORM_PER_ACTIVATION` (which is for non-convolutional layers, which Chainer assumes are 2D) is
> Normalization is performed per-activation.
> This mode is intended to be used after non-
> convolutional network layers. In this mode bnBias
> and bnScale tensor dimensions are 1xCxHxW.

But this PR explicitly says cuDNN supports 2D tensors:
`self.cudnn_dim_ok = x.ndim == 2 or x.ndim == 4`
Is that true for an older cuDNN version (?), or only at train time (??) or am I doing something wrong...

PR #1317, opened 2016-06-26
Implementing highway net function

fukatani:
This PR will add `highway` network (https://arxiv.org/pdf/1505.00387v2.pdf).
This network is intended to shorten learning epoch of very deep network (>20 layer), and it is used for character language model. (https://arxiv.org/pdf/1508.06615v4.pdf)

Now, Forward / Backward calculation is not validated.


fukatani:
Hi!
I have implemented highway network.
Keras supported this network, so I think this feature is valuable to merge.

I'd appreciate it if I could get any opinion.

mattya:
Hi, thank you for PR!
Could you also update https://github.com/pfnet/chainer/blob/master/docs/source/reference/links.rst ?

chainer/links/connection/highway.py
```diff
@@ -0,0 +1,79 @@
...
+            x (~chainer.Variable): Input variable.
+
+        Returns:
+            Variable: Output variable. Its array has the same spatial size and
+            the same minibatch size as the input array.
+
+        """
+        out_plain = self.activate(self.plain(x))
+        out_transform = sigmoid.sigmoid(self.transform(x))
+        out_transform_copy = copy.copy(out_transform)
```
mattya: Could you explain why you copy out_transform?

chainer/links/connection/highway.py
```diff
@@ -0,0 +1,79 @@
...
+        a = basic_math.mul(out_plain, out_transform)
+        ones = cuda.get_array_module(x).ones_like(x.data)
```
mattya: This code doesn't work on GPU.
  ones = cuda.get_array_module(x.data).ones_like(x.data) 
is correct.

chainer/links/connection/highway.py
```diff
@@ -0,0 +1,79 @@
...
+        b = basic_math.rsub(out_transform_copy, ones)
+        c = basic_math.mul(x, b)
+        y = basic_math.add(a, c)
```
mattya: I think 
`y = out_plain * out_transform + x * (1 - out_transform)`
is easy to understand.

fukatani:
Thanks for your attentive comments!

The reason why I copied 'out_transform' is the note of https://github.com/pfnet/chainer/blob/master/chainer/function.py as follows.
```
       As of v1.5, a function instance cannot be used twice in any
       computational graphs. In order to reuse a function object multiple
       times, use :func:`copy.copy` before the function applications to make a
       copy of the instance.
```
But I may misunderstand the meaning of it.

I added test for checking back-prop value and previous PR doesn't pass test.
So I fixed it to as you suggested (don't use copy).
And I reflected the other your pointing out (edit links.rst).

mattya:
Thank you for the update!  LGTM.

chainer/links/connection/highway.py
```diff
@@ -0,0 +1,71 @@
...
+from chainer.functions.activation import sigmoid
+from chainer import link
+from chainer.links.connection import linear
+
+
+class Highway(link.Chain):
+
+    """Highway module.
+
+    Network allowing unimpeded information flow across layer on information
```
beam2d: Make it into a sentence, e.g. "This is a network fragment allowing ..."

chainer/links/connection/highway.py
```diff
@@ -0,0 +1,71 @@
...
+    flow.
+    It applies two different gate. One gate is transform gate, and the other
```
beam2d: two different gate*s*

chainer/links/connection/highway.py
```diff
@@ -0,0 +1,71 @@
...
+    is carry gate. Carry gate enable to propagete large gradient value to
```
beam2d: Carry gate enable**s** **us** to ...

chainer/links/connection/highway.py
```diff
@@ -0,0 +1,71 @@
...
+    previous layer.
+
+    The output is sum of linear multiplied transform gate, and input array
+    multiplied carry gate.
```
beam2d: The output is a sum of the output of a fully-connected layer with the specified activation function multiplied by the transform gate, and the input array multiplied by the carry gate.

(Maybe this corrected sentence is too long, so it would be better to split it into multiple sentences.)

beam2d:
I added some comments on the document.
I also think that it is better to explicitly write down the equations that defines the forward computation of this chain.
If you need, I can take over the documentation task.

fukatani:
Thanke for your review!

I appreciate if you take over the documentation task.
My English is not natural,
repeating exchange about doc might be a time loss for you.

And I relaxed test `atol` and `rtol` for backword test.
Highway including product of two linear, 
I decided them to be square root of it in linear.

fukatani:
Because CI failed, I relaxed test `atol` and `rtol` for backword test.
Highway including product of two linear, 
I decided them to be square root of it in linear. (Or is there any reasonable method?)

chainer/links/connection/highway.py
```diff
@@ -0,0 +1,79 @@
...
+class Highway(link.Chain):
+
+    """Highway module.
+
+    This is a network allowing unimpeded information flow across layer on
+    information highways.
+    It applies two different gates. One gate is transform gate
+    (:math:`tanh(W_h x + b_h)`), and the other is carry gate
+    (:math:`sigma(W_t x + b_t)`). Carry gate enable us to propagete large
+    gradient value to previous layer.
```
mattya: One gate is ordinal non-linear transformation (:math:`activate(W_h x + b_h)`),
and the other is carry gate (:math:`1 - sigma(W_t x + b_t)`)

chainer/links/connection/highway.py
```diff
@@ -0,0 +1,79 @@
...
+
+    Highway module returns :math:`y` defined as
+
+    .. math::
+
+    y &=& Relu(W_h x + b_h)\\odot\\sigma(W_t x + b_t)) +
```
mattya: Relu -> activate

fukatani:
Thank you for your review!
And I'm sorry for late response.
I fixed doc.

fukatani:
Thanks for mattya-san, document is fixed. Please check this PR.

chainer/links/connection/highway.py
```diff
@@ -0,0 +1,75 @@
...
+    For simplicity, the author defined :math:`C = 1 - T`.
+    Highway module returns :math:`y` defined as
+
+    .. math::
+
+        y = activate(W_h x + b_h) \\odot \\sigma(W_t x + b_t) +
+        x \\odot(1 - \\sigma(W_t x + b_t))
+
+    The output array has the same spatial size as the input. In order to
+    satisfy this, :math:`W_h` and :math:`W_t` must be square matrix.
```
beam2d: matrix -> matrices

chainer/links/connection/highway.py
```diff
@@ -0,0 +1,75 @@
...
+
+    Args:
+        in_out_size (int): Dimension of input and output vectors.
+        nobias (bool): If ``True``, then this function does not use the bias.
+        activate: Activation function of plain array. :math:`tanh` is also
+        enable.
```
beam2d: enable -> available (if I understand your intention correctly?)

chainer/links/connection/highway.py
```diff
@@ -0,0 +1,75 @@
...
+        init_Wh (2-D array): Initial weight value of plain array. If ``None``,
+            then this function uses to initialize ``wscale``.
```
beam2d: uses to -> uses it to

chainer/links/connection/highway.py
```diff
@@ -0,0 +1,75 @@
...
+            May also be a callable that takes ``numpy.ndarray`` or
+            ``cupy.ndarray`` and edits its value.
+        init_bh (1-D array): Initial bias value of plain array. If ``None``,
+            then this function uses to initialize zero vector.
```
beam2d: ditto

chainer/links/connection/highway.py
```diff
@@ -0,0 +1,75 @@
...
+            May also be a callable that takes ``numpy.ndarray`` or
+            ``cupy.ndarray`` and edits its value.
+        init_Wt (2-D array): Initial weight value of transform array.
+            If ``None``, then this function uses to initialize ``wscale``.
```
beam2d: ditto

chainer/links/connection/highway.py
```diff
@@ -0,0 +1,75 @@
...
+            May also be a callable that takes ``numpy.ndarray`` or
+            ``cupy.ndarray`` and edits its value.
+        init_bt (1-D array): Initial bias value of transform array.
+            Default value is -1 vector.
+            May also be a callable that takes ``numpy.ndarray`` or
+            ``cupy.ndarray`` and edits its value.
+            Negative value is sufficient for learning (e.g. -1, -3, ...).
```
beam2d: sufficient for learning -> recommended by the author of the paper? (if I understand your intention correctly)

beam2d:
I added some comments on the documentation. LGTM except those!

fukatani:
Thanks for your attentive review! :smiley:
I applied your comments.

mattya:
Thank you! 👍

beam2d:
LGTM!

PR #1325, opened 2016-06-27
refactoring: range -> six.moves.range

fukatani:
Use `six.moves.range` instead of `range`.
If any reason for using `range`, this PR should be rejected.


delta2323:
LGTM 👍

delta2323:
Thank you!

PR #1329, opened 2016-06-28
Added data parallel example

amitibo:
This PR adds ParallelUpdater that implements data parallelism on multiple GPU. An example of data parallel model using 2 GPUs is given for the mnist dataset.

### DIFF
--- a/examples/mnist/train_mnist_data_parallel.py
+++ b/examples/mnist/train_mnist_data_parallel.py
@@ -1,0 +1,99 @@
+#!/usr/bin/env python
+from __future__ import print_function
+import argparse
+
+import chainer
+import chainer.links as L
+from chainer import training
+from chainer.training import extensions
+
+import train_mnist
+
+
+def main():
+    parser = argparse.ArgumentParser(description='Chainer example: MNIST')
+    parser.add_argument('--batchsize', '-b', type=int, default=400,
+                        help='Number of images in each mini batch')
+    parser.add_argument('--epoch', '-e', type=int, default=20,
+                        help='Number of sweeps over the dataset to train')
+    parser.add_argument('--gpu0', '-g', type=int, default=0,
+                        help='First GPU ID')
+    parser.add_argument('--gpu1', '-G', type=int, default=1,
+                        help='Second GPU ID')
+    parser.add_argument('--out', '-o', default='result_parallel',
+                        help='Directory to output the result')
+    parser.add_argument('--resume', '-r', default='',
+                        help='Resume the training from snapshot')
+    parser.add_argument('--unit', '-u', type=int, default=1000,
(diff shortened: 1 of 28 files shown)


examples/mnist/train_mnist_data_parallel.py
```diff
@@ -0,0 +1,99 @@
...
+    updater = training.ParallelUpdater(
+        train_iter,
+        optimizer,
+        models={'main':model_0, 'second':model_1},
+        devices={'main':args.gpu0, 'second':args.gpu1},
+    )
+    trainer = training.Trainer(updater, (args.epoch, 'epoch'), out=args.out)
+
+    # Add observer for model on second gpu.
+    trainer.reporter.add_observer('second', model_1)
```
beam2d: Can it be made into a part of ParallelUpdater?

beam2d:
LGTM except above comment.

amitibo:
@beam2d I am not sure how to implement your comment above. Do you have anything specific in mind?
On a similar note, the need to copy the original model (chain) to several devices seems redundant. Maybe it is possible to define a ``ParallelModel`` or ``ParallelChain`` which will abstract away the need to copy the model to the different devices and the separate forward and backward passes of the model copies.It can also add the observers of all the model copies.

beam2d:
> I am not sure how to implement your comment above. Do you have anything specific in mind?

I though it can be moved to the first call of the update method, though it is ugly a little (we still have to prepare a way to send the trainer to the updater in this case).
Now I'm thinking it is better to add an API for it (see below).

> Maybe it is possible to define a ParallelModel or ParallelChain which will abstract away the need to copy the model to the different devices and the separate forward and backward passes of the model copies.It can also add the observers of all the model copies.

It seems a good solution. One issue is that parameters on devices other than the main one should not be involved in the optimization, so we have to modify the behaviors of the default iterator methods (e.g. params).
I don't think it is a good idea to modify the definition of such a common interface.

At least for hiding the model copying, you can copy the model at `ParallelUpdater.__init__` as well, instead of making a new chain. I mean making this initializer method only accept the main model and clone it into a specified set of devices automatically.

With regard to how to send the trainer to the updater, now I think it is better to add an API to `Updater` (and make `Trainer.__init__` call it automatically). Once I check the validity of this approach, I will change the trainer2 branch implementation, then you can use it to move the observers registration to inside of ParallelUpdater.

beam2d:
@amitibo I added `Updater.connect_trainer` to send the trainer to the updater. It is called at the end of `Trainer.__init__`, so you can use it to register the models on other devices to `trainer.reporter`.

beam2d:
LGTM. I will fix some minor point after the merge (e.g. py3 support and reference). Thank you for your quick and great contribution!

Issue #1283, opened 2016-06-16
Support Dilated Convolution


yasunorikudo:
See https://arxiv.org/abs/1511.07122 . Dilated convolutions support exponentially expanding receptive fields without losing resolution or coverage. I am going to implement this layer.

delta2323:
Now the dilated convolution is supported, we can close this issue.

PR #1335, opened 2016-06-30
Dilated Convolution Layer

yasunorikudo:
I created the issue in #1283. This PR will add dilated convolution layer (https://arxiv.org/abs/1511.07122).

### DIFF
--- a/chainer/functions/__init__.py
+++ b/chainer/functions/__init__.py
@@ -96,7 +96,8 @@ from chainer.functions.util import forget
 from chainer.links.activation import prelu as links_prelu
 from chainer.links.connection import bilinear as links_bilinear
 from chainer.links.connection import convolution_2d as links_convolution_2d
-from chainer.links.connection import dilated_convolution_2d as links_dilated_convolution_2d
+from chainer.links.connection import dilated_convolution_2d \
+    as links_dilated_convolution_2d
 from chainer.links.connection import embed_id as links_embed_id
 from chainer.links.connection import inception
 from chainer.links.connection import inceptionbn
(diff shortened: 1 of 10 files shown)


mattya:
Hello, thank you for the cool PR.
I've tested your code and it looks working without critical problems.
I have  some comments at this time.

1. It could be helpful if there's more detail description about the relationships between `ksize`, `dilate`, `patch size` and `output size` in the document for `chainer.functions.dilated_convolution_2d`.  
2. It's needed to add DilatedConvolution2D to `chainer/docs/source/reference/links.rst`.
3. The author of the paper says "random initialization schemes were not effective".  Do you have any plans to implement the "identity initialization" proposed in the paper?
4. I briefly benchmarked your function by forward and backward (32, 32, 64, 64) tensor to DilatedConvolution2D(32, 32, ksize=3, stride=1, pad=4, dilate=4, use_cudnn=use_cudnn) and got the result that `use_cudnn=False` was about 2 times faster than `use_cudnn=True`.  Is it likely behaviour?

yasunorikudo:
Thank you for your reviewing. I fixed documents and added DilatedConvolution2D to ```chainer/docs/source/reference/links.rst```.
- About identity initialization
  Chainer doesn't support 4D identity Initialization. I will raise an issue about 4D identity initialization and make PR.
- About difference between ```use_cudnn=True``` and ```use_cudnn=False```
  cuDNN doesn't support dilated convolution yet, but I implemented this layer using 1×1 convolution operation in cuDNN library. Therefore in some cases  ```use_cudnn=True``` option is much slower. In my experiments under the same condition above, forward prop is about 50% faster and backward prop is 20% slower than ```use_cudnn=False```.

mattya:
Thank you!  LGTM

PR #1340, opened 2016-07-01
F.classification_summary (precision, recall, f1_score)

delta2323:
This PR implements `F.classification_summary` which calculates precision, recall, f1_score, and support  of the minibatch of interest for each class label. We also provide utility functions that returns each metric only. Note that these functions implements forward propagation only, as `F.accuracy` does.


rezoo:
Thank you for your PR! I'll check this review.

tests/chainer_tests/functions_tests/evaluation_tests/test_classification_summary.py
```diff
@@ -0,0 +1,98 @@
...
+    ret = numpy.zeros((label_num,), dtype=numpy.int32)
+    for t in ts.ravel():
+        ret[t] += 1
+    return ret
+
+
+@testing.parameterize(
+    *testing.product_dict(
+        [{'y_shape': (30, 3), 't_shape': (30,)},
+         {'y_shape': (30, 3, 5), 't_shape': (30, 5)}],
```
rezoo: I found that the above settings sometimes raise RuntimeWarning because a denominator of F1 score may contains zero due to the small batchsize.
How about setting the batch size to a larger one (e.g., 100) to disable such annoying messages?

chainer/functions/evaluation/classification_summary.py
```diff
@@ -0,0 +1,124 @@
...
+
+    This function calculates the following quantities for each label value.
+
+    - Precision: :math:`\frac{tp}{tp + fp}`
+    - Recall: :math:`\frac{tp}{tp + tn}`
+    - F1 Score: The harmonic average of Precision and Recall.
+    - Support: The number of data points of each ground truth label.
+
+    Here, ``tp``, ``fp``, and ``tn`` stand for true positive, false positive,
+    and true negative, respectively.
```
rezoo: Here, ``tp``, ``fp``, and ``tn`` stand for true positives, false positives, and true negatives, respectively.

chainer/functions/evaluation/classification_summary.py
```diff
@@ -0,0 +1,124 @@
...
+
+        f1 = _f1_score(precision, recall, self.beta)
+
+        return precision, recall, f1, support
+
+
+def classification_summary(y, t, label_num=None, beta=1.0):
+    """Calculates Precision, Recall, F1 Score, and support.
+
+    This function calculates the following quantities for each label value.
```
rezoo: This function calculates the following quantities for each class.

chainer/functions/evaluation/classification_summary.py
```diff
@@ -0,0 +1,124 @@
...
+
+    - Precision: :math:`\frac{tp}{tp + fp}`
+    - Recall: :math:`\frac{tp}{tp + tn}`
+    - F1 Score: The harmonic average of Precision and Recall.
+    - Support: The number of data points of each ground truth label.
+
+    Here, ``tp``, ``fp``, and ``tn`` stand for true positive, false positive,
+    and true negative, respectively.
+
+    If ``ignore_label`` is not ``None``, data points whose labels are ignored.
```
rezoo: If ``ignore_label`` is not ``None``, instances with the given label are ignored.

chainer/functions/evaluation/classification_summary.py
```diff
@@ -0,0 +1,124 @@
...
+    ``label_num`` should specifies the number of label types.
+    Each value in ``t`` should be the integer in ``[0, label_num)``.
+    If ``label_num`` is ``None``, we regard ``label_num`` to be the maximum
+    value of in ``t`` plus 1.
```
rezoo: ``label_num`` specifies the number of classes, that is, each value in ``t`` must be an integer in the range of ``[0, label_num)``.
If ``label_num`` is ``None``, this function regards ``label_num`` as a maximum value of ``t`` plus one.

chainer/functions/evaluation/classification_summary.py
```diff
@@ -0,0 +1,124 @@
...
+
+    Args:
+        y (~chainer.Variable): Variable holding a vector of scores.
+        t (~chainer.Variable): Variable holding ground truth label.
+        label_num (int): the number of label types.
+        beta (float): Strength of recall against precision in F1 score.
```
rezoo: label_num (int): the number of classes.
beta (float): The parameter determines the weight of precision in the F-beta score.

chainer/functions/evaluation/classification_summary.py
```diff
@@ -0,0 +1,124 @@
...
+import numpy
+import six
+
+import chainer
+from chainer import cuda
+from chainer import function
+from chainer.utils import type_check
+
+
+def _f1_score(precision, recall, beta):
```
rezoo: I think it is better that ``_f1_score`` is renamed ``_fbeta_score``.

chainer/functions/evaluation/classification_summary.py
```diff
@@ -0,0 +1,124 @@
...
+    ret = ClassificationSummary(1.0, label_num)(y, t)
+    return ret[0], ret[-1]
+
+
+def recall(y, t, label_num=None):
+    ret = ClassificationSummary(1.0, label_num)(y, t)
+    return ret[1], ret[-1]
+
+
+def f1_score(y, t, label_num=None, beta=1.0):
```
rezoo: ``f1_score`` should be renamed ``fbeta_score``, and we have to implement the new function called ``f1_score`` like:
```
def f1_score(y, t, label_num=None):
    return fbeta_score(y, t, label_num=label_num, beta=1.0)
```

chainer/functions/evaluation/classification_summary.py
```diff
@@ -0,0 +1,124 @@
...
+        t_ndim = t_type.ndim.eval()
+        type_check.expect(
+            x_type.ndim >= t_type.ndim,
+            x_type.shape[0] == t_type.shape[0],
+            x_type.shape[2: t_ndim + 1] == t_type.shape[1:]
+        )
+        for i in six.moves.range(t_ndim + 1, x_type.ndim.eval()):
+            type_check.expect(x_type.shape[i] == 1)
+
+    def forward(self, inputs):
```
rezoo: As you mentioned, the current implementation is insufficient when the number of labels is relatively large.
How about using ``xp.bincount()`` instead of ``for`` statement, and replace it with the following?
```python
    def forward(self, inputs):
        y, t = inputs

        if self.label_num is None:
            label_num = xp.maximum(t) + 1
        else:
            label_num = self.label_num
            if chainer.is_debug():
                assert (t < label_num).all()

        mask = (t == self.ignore_label).ravel()
        pred = xp.where(mask, label_num, y.argmax(axis=1).ravel())
        true = xp.where(mask, label_num, t.ravel())
        support = xp.bincount(true, minlength=label_num + 1)[:label_num]
        relevant = xp.bincount(pred, minlength=label_num + 1)[:label_num]
        tp_mask = xp.where(pred == true, true, label_num)
        tp = xp.bincount(tp_mask, minlength=label_num + 1)[:label_num]

        precision = tp / relevant
        recall = tp / support

        f1 = _f1_score(precision, recall, self.beta)

        return precision, recall, f1, support
```
I confirmed that the above code passes the test.

rezoo:
LGTM except for the above comments!

chainer/functions/evaluation/classification_summary.py
```diff
@@ -0,0 +1,124 @@
...
+    beta_square = beta * beta
+    return ((1 + beta_square) * precision * recall /
+            (beta_square * precision + recall)).astype(precision.dtype)
+
+
+class ClassificationSummary(function.Function):
+
+    ignore_label = -1
+
+    def __init__(self, beta, label_num):
```
rezoo: I found an additional problem; it seems that ``ignore_label`` option is missing.

chainer/functions/evaluation/classification_summary.py
```diff
@@ -0,0 +1,124 @@
...
+            support[i] = supp.sum()
+            precision[i] = tp / relevant.sum()
+            recall[i] = tp / support[i]
+
+        f1 = _f1_score(precision, recall, self.beta)
+
+        return precision, recall, f1, support
+
+
+def classification_summary(y, t, label_num=None, beta=1.0):
```
rezoo: Is it better to add ``ignore_label`` option?

chainer/functions/evaluation/classification_summary.py
delta2323: done

chainer/functions/evaluation/classification_summary.py
delta2323: done

chainer/functions/evaluation/classification_summary.py
delta2323: Thank you for the suggestion. It is just because I could not spare my time to consider the test cases (especially, the corner cases). Further, as we can expand the API afterwards, I sent this PR without `ignore_label` option. So I do not implement the option somewhat intentionally. But as you may think, it is better, to provide users with the interface that changes `ignore_label` easily. So I will consider adding it.

By the way, `F.accuracy` can change the value used as `ignore_label`, whereas it is fixed to `-1` in `F.binary_accuracy`, `F.sigmoid_cross_entropy` and `F.softmax_cross_entropy`. So do you think we should admit users to flexibly change the value of `ignore_label`?

But as you may think, it is better to add `ignore_label`

chainer/functions/evaluation/classification_summary.py
delta2323: I commented about `ignore_label` in another place.

chainer/functions/evaluation/classification_summary.py
delta2323: done

chainer/functions/evaluation/classification_summary.py
delta2323: done

chainer/functions/evaluation/classification_summary.py
delta2323: done

tests/chainer_tests/functions_tests/evaluation_tests/test_classification_summary.py
delta2323: I concerned that larger batch size can cause slowdown of unit tests. But I found that change of mini batch size from 30 to 100 does not increase test time so much. So I will change the mini batch size.

chainer/functions/evaluation/classification_summary.py
delta2323: done

chainer/functions/evaluation/classification_summary.py
delta2323: Thank you for the suggesion! `xp.bincount` perfectly suits this function.

delta2323:
I found test cases that consider `ignore_label` is missing. I will add it.

delta2323:
Suppose we provide `ignore_label` option, I am wondering what is the preferred behavior when use specified the integer in `[0, label_num)` as `ignore_label`?

rezoo:
> By the way, F.accuracy can change the value used as ignore_label, whereas it is fixed to -1 in F.binary_accuracy, F.sigmoid_cross_entropy and F.softmax_cross_entropy. So do you think we should admit users to flexibly change the value of ignore_label?

We can choose two options: i) we will *fix* the value of ``ignore_label`` to -1 in all the functions, or ii) we will improve all the functions to accept ``ignore_label`` except for -1.
The option i) is simple and easier to implement, but it would break backward compatibility.
Therefore, in my opinion it would be better for us to add ``ignore_label`` option to ``Accuracy`` and ``classification_summary``, and unite these specifications after the major release.

rezoo:
> Suppose we provide ignore_label option, I am wondering what is the preferred behavior when user specifies the integer in [0, label_num) as ignore_label? Especially, what should we return precision etc. that corresponds to the ignored value?

Although in the current implementation ``ignore_label`` must be outside of ``[0, label_num)``, its specification is the completely the same as that of the ``accuracy`` layer.
I think the preferred behavior would be that ``classification_summary`` computes summary for each class without the given label, and we need to improve the both layers in near future.

delta2323:
@rezoo Thank you for the comment. I update the implementation, could you check that?

rezoo:
Thank you for your commits! I'll check it.

tests/chainer_tests/functions_tests/evaluation_tests/test_classification_summary.py
```diff
@@ -0,0 +1,134 @@
...
+
+        pred = self.y.argmax(axis=1).reshape(self.t.shape)
+        p_expect = precision(pred, self.t, self.dtype,
+                             self.label_num, self.ignore_label)
+        r_expect = recall(pred, self.t, self.dtype,
+                          self.label_num, self.ignore_label)
+        fbeta_expect = fbeta_score(p_expect, r_expect, self.beta)
+        s_expect = support(self.t, self.dtype,
+                           self.label_num, self.ignore_label)
+        chainer.testing.assert_allclose(p_actual.data, p_expect,
```
rezoo: Could you add a comment that describe the reason why we suppress RuntimeWarning like that?
```python
        # Suppress warning that arises from zero division.
        warnings.filterwarnings('ignore', category=RuntimeWarning)

File: tests/chainer_tests/functions_tests/evaluation_tests/test_classification_summary.py [hunk 13]
```diff
@@ -0,0 +1,134 @@
...
+        t_upper = 3 if self.label_num is None else self.label_num
+        self.label_num = 3
+        self.y = numpy.random.uniform(-1, 1, self.y_shape).astype(self.dtype)
+        self.t = make_ground_truth(t_upper, self.t_shape,
+                                   self.ignore_label, self.has_ignore_label)
+        self.check_forward_options = {}
+        if self.dtype == numpy.float16:
+            self.check_forward_options = {'atol': 1e-4, 'rtol': 1e-3}
+
+        warnings.filterwarnings('ignore', category=RuntimeWarning)
```
rezoo: Could you add a comment that describe the reason why we suppress RuntimeWarning like that?
```python
        # Suppress warning that arises from zero division.
        warnings.filterwarnings('ignore', category=RuntimeWarning)
```

chainer/functions/evaluation/classification_summary.py
```diff
@@ -90,6 +90,9 @@ def classification_summary(y, t, label_num=None, beta=1.0, ignore_label=-1):
     into account for calculating the above quantities.
     By default, it is set to -1 so that all instances are taken
     into consideration, as labels are supposed to be non-negative integers.
+    If ``ignore_label`` is "valid", that is, it is a non-negative
+    integer less than ``label_num``, then the ``ignore_label``-th
+    entries in output arrays do not contain correct quantities.
```
rezoo: Note that if ``ignore_label`` is a non-negative integer less than ``label_num``, this function arises ``RuntimeWarning``, and ``ignore_label``-th entries in output arrays do not contain correct quantities.

rezoo:
LGTM except for the above small comments! 👍 👍 👍

chainer/functions/evaluation/classification_summary.py
delta2323: I have several issues which I do not have clear answers.
* By raising `RuntimeWarning`, we declare that setting a non-negative integer to `ignore_label` is an abuse of this function. But I guess some users sometimes would like to use `ignore_label` in this way to ignore some specified label. So to me we should not raise warning here.
* We will be in the position that we should invalidate to assign an non-negative integer to `ignore_label` in this paragraph. The message attached to this `RuntimeWarning` is division-by-zero warning. It is by product of the inappropriate `ignore_label` and user cannot understand the cause of the warning easily. So we should wrap this warning and attach the more meaningful message.

Thank you for your suggestion on adding description on ``

rezoo: I'm not sure if my understanding w.r.t. your comment is correct, but it seems that you have considered that setting ``ignore_label`` to a value less than ``n_classes`` is **illegal** in this function.

My opinion is the following; it is the best for us to completely support ``ignore_label`` within {0, ..., n_label}, but it requires to deal with the problem that setting a small non-negative value raises ``RuntimeWarning``. Therefore, currently we define that setting a non-negative integer less than ``n_label`` to``ignore_label`` is "incorrect", and yields "undefined behavior" (even in this setting this function works well to a certain degree, though).
In this problem we should solve it by some PRs, but it is a future-work.

delta2323:
Sorry for late response. I agree with your idea. I will change the document.

rezoo:
Good, then I would like to merge it.

PR #1353, opened 2016-07-04
Support N-dimensional max pooling funciton

takagi:
Rel. #945 and #1076. Split from #1279 .

This PR generalizes two-dimensional max pooling function to N-dimensional one.


takagi:
I show current status of the first implementation of this PR.

■…Done □…Doing

**ND Pooling**

|Pooling|Items|Documentation|Test|Implementation|
|---|---|---|---|---|
|Pooling|Function - CPU - forward|N/A|N/A|■|
|Pooling|Function - CPU - backward|N/A|N/A|■|
|Pooling|Function - GPU - forward|N/A|N/A|■|
|Pooling|Function - GPU - backward|N/A|N/A|■|
|MaxPooling|Function - CPU - forward|N/A|■|■|
|MaxPooling|Function - CPU - backward|N/A|■|■|
|MaxPooling|Function - GPU - forward|N/A|■|■|
|MaxPooling|Function - GPU - generate_forward|N/A|N/A|■|
|MaxPooling|Function - GPU - backward|N/A|■|■|
|MaxPooling|Function - GPU - generate_backward|N/A|N/A|■|
|MaxPooling|Function - max_pooling_nd|■|■|■|
|MaxPooling|Function - cuDNN - call forward|N/A|■|N/A|
|MaxPooling|Function - cuDNN - call backward|N/A|■|N/A|
|AveragePooling|Function - CPU - forward|N/A|□|□|
|AveragePooling|Function - CPU - backward|N/A|□|□|
|AveragePooling|Function - GPU - forward|N/A|□|□|
|AveragePooling|Function - GPU - backward|N/A|□|□|
|...|||||

takagi:
cuDNN now has newer version of pooling APIs that Chainer uses.

- cudnnSetPooling2dDescriptor_v4
- cudnnSetPoolingNdDescriptor_v4

The difference from v3 is that v4 has `cudnnNanPropagation_t *maxpoolingNanOpt` parameter that specifies routines should propagate `Nan` numbers or not.

I suppose that specifiying `Nan` propagation is not needed in Chainer so I leave them as they are v3.

takagi:
I will follow the change #1334 which is performance improvement for 2D max pooling.

takagi:
I split N-dimensional average pooling to #1391.

takagi:
Same as #1334, b86ccc9 gets the following efficiency improvement in my profiling:

    y = max_pooling_nd(x, ksize)

    x shape: (100, 3, 8, 8, 8)
    kernel size: (3, 3, 3)
    y shape: (100, 3, 3, 3, 3)

    backward_cpu: 11.729 [s] -> 2.2435[s]
    (forward_cpu): 4.019[s]

chainer/functions/pooling/pooling_nd_kernel.py
```diff
@@ -0,0 +1,300 @@
+import chainer
+from chainer.utils.conv_nd_kernel import _map
+from chainer.utils.conv_nd_kernel import mulexp
+from chainer.utils.conv_nd_kernel import succ_sublists
+from chainer.utils.conv_nd_kernel import vars
+from chainer.utils.conv_nd_kernel import Writer
```
delta2323: Direct import of objects is not allowed in pep8 (`_map`, `mulexp`, `succ_sublists`, `vars`, `Writer`).

chainer/functions/pooling/max_pooling_nd_kernel.py
```diff
@@ -0,0 +1,92 @@
+from chainer.functions.pooling import pooling_nd_kernel
+from chainer.utils.conv_nd_kernel import _map
+from chainer.utils.conv_nd_kernel import muladdexp
+from chainer.utils.conv_nd_kernel import vars
+from chainer.utils.conv_nd_kernel import Writer
```
delta2323: Direct import of objects is not allowed in pep8.

tests/chainer_tests/functions_tests/pooling_tests/test_max_pooling_nd.py
```diff
@@ -0,0 +1,199 @@
...
+
+    @attr.gpu
+    @condition.retry(3)
+    def test_forward_gpu_no_cudnn(self):
+        self.check_forward(cuda.to_gpu(self.x), False)
+
+    def test_forward_consistency_regression(self):
+        # Regression test to max_pooling_2d.
+
+        if len(self.dims) != 2:
```
takagi: I suppose that `max_pooling_2d` does not work with `self.dims` other than two, so the result of `max_pooling_nd` can not be tested with that of `max_pooling_2d`.

chainer/functions/pooling/pooling_nd.py
```diff
@@ -0,0 +1,94 @@
...
+if cuda.cudnn_enabled:
+    cudnn = cuda.cudnn
+    libcudnn = cudnn.cudnn
+    _cudnn_version = libcudnn.getVersion()
+
+
+_check_cudnn_acceptable_type = pooling_2d._check_cudnn_acceptable_type
+
+
+class PoolingND(function.Function):
```
takagi: Yes, I've fixed it!

chainer/functions/pooling/max_pooling_nd_kernel.py
takagi: I fixed them to be module import!

tests/chainer_tests/functions_tests/pooling_tests/test_max_pooling_nd.py
```diff
@@ -0,0 +1,199 @@
...
+from chainer.utils import conv
+
+
+def pooling_patches(dims, ksize, stride, pad, cover_all):
+    """Return tuples of slices that indicate pooling patches."""
+    # Left-top indeces of each pooling patch.
+    if cover_all:
+        xss = itertools.product(
+            *[six.moves.range(-p, d + p - k + s, s)
+              for (d, k, s, p) in zip(dims, ksize, stride, pad)])
```
delta2323: Strictly speaking, if it is better to use `six.moves.range`, it is also better to use `six.moves.zip` as well (ref. [here](https://pythonhosted.org/six/#module-six.moves)). But it is not critical because it is not expected to affect the performance so much.

tests/chainer_tests/functions_tests/pooling_tests/test_pooling_nd_kernel.py
```diff
@@ -0,0 +1,39 @@
...
+        chainer.cuda.clear_memo()
+
+    def test_pooling_nd_kernel_forward_memo(self):
+        ndim = self.ndim
+        with mock.patch('chainer.functions.pooling.pooling_nd_kernel.'
+                        'PoolingNDKernelForward._generate') as m:
+            pooling_nd_kernel.PoolingNDKernelForward.generate(ndim)
+            m.assert_called_once_with(ndim)
+            pooling_nd_kernel.PoolingNDKernelForward.generate(ndim)
+            m.assert_called_once_with(ndim)
```
delta2323: Could you add comment that you would like to check the kernel code is cached.

tests/chainer_tests/functions_tests/pooling_tests/test_pooling_nd_kernel.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def test_pooling_nd_kernel_backward_memo(self):
+        ndim = self.ndim
+        with mock.patch('chainer.functions.pooling.pooling_nd_kernel.'
+                        'PoolingNDKernelBackward._generate') as m:
+            pooling_nd_kernel.PoolingNDKernelBackward.generate(ndim)
+            m.assert_called_once_with(ndim)
+            pooling_nd_kernel.PoolingNDKernelBackward.generate(ndim)
+            m.assert_called_once_with(ndim)
```
delta2323: ditto (please add comment on the cache of the kernel code).

tests/chainer_tests/functions_tests/pooling_tests/test_max_pooling_nd.py
```diff
@@ -0,0 +1,199 @@
...
+from chainer import gradient_check
+from chainer import testing
+from chainer.testing import attr
+from chainer.testing import condition
+from chainer.utils import conv
+
+
+def pooling_patches(dims, ksize, stride, pad, cover_all):
+    """Return tuples of slices that indicate pooling patches."""
+    # Left-top indeces of each pooling patch.
```
takagi: Fixed!

takagi:
> Is there any reason why the underscore in _map remains, while those in others are removed?

It is just because I wanted to give different name with standard `map` built-in but was not sure proper name, not for making it private. Anyway, it is not a good name.

delta2323:
>It is just because I wanted to give a different name with standard map built-in

OK, I understand it.

chainer/utils/conv_nd_kernel.py
```diff
@@ -28,9 +29,9 @@ def aux(exp, arg):
         return functools.reduce(aux, zip(xs, ys))
 
 
-def _map(fn, *lst):
+def map(fn, *lst):
```
delta2323: I think map is not appropriate because it hides builtin `map` function. So how about `map_`?

takagi: Good! I will fix it again.

takagi: Fixed.

chainer/functions/pooling/max_pooling_nd_kernel.py
```diff
@@ -0,0 +1,92 @@
...
+        #     if (maxval < v) {
+        #       maxval   = v;
+        #       argmax_0 = x_0;
+        #       argmax_1 = x_1;
+        #     }
+        w = Writer()
+        w.write('T v = in[{}];'.format(offset))
+        w.write('if (maxval < v) {', 'inc')
+        w.write('maxval = v;')
+        for (argmax, x) in zip(self.argmaxs, xs):
```
takagi: Removed!

tests/chainer_tests/functions_tests/pooling_tests/test_pooling_nd_kernel.py
takagi: Added!

tests/chainer_tests/functions_tests/pooling_tests/test_pooling_nd_kernel.py
takagi: Added!

tests/chainer_tests/functions_tests/pooling_tests/test_max_pooling_nd.py
takagi: Ah~, yes. I've replaced builtin `zip` with `six.moves.zip` in files changed in this PR, they were not so much.

takagi:
I should have responded all your comments (The new review system of github seems not matured enough, comments that I have not responded are hidden as outdated.)

delta2323:
It is just a note but we found that current implementation does not work well with cuDNN v2.

takagi:
> We found that current implementation does not work well with cuDNN v2.

I've found that `cudnnSetPoolingNdDescriptor` in cuDNN v2 works with two-dimensional inputs but does not work with others including one-dimensional and three-dimensional inputs.

I checked the following cases for three-dimensional inputs with cuDNN v2, all of which raise `CUDNN_STATUS_NOT_SUPPORTED`:

- `pad` of `0`
- `stride` of `1`
- `ksize` of `1`

where `cover_all` is `True` and `False`, and `dtype` is `numpy.float32`.

The reference of cuDNN v2 mentions that `cudnnPoolingForward` returns `CUDNN_STATUS_NOT_SUPPORTED` when `The wStride of input tensor or output tensor is not 1.`. I checked the strides of an input array with the shape of `(2, 3, 4, 4, 4)`, and got strides of `(192, 64, 16, 4, 1)`, which is along the restriction.

takagi:
After discussion with @delta2323, we've decided to take the following actions:

- On cuDNN v2 environment, use CuPy implementation for N-dimensional arrays other than two-dimensional ones.
- Leave the problem as another issue.

takagi:
I've fixed to use CuPy implementation for N-dimensional arrays other than two on cuDNN v2 environment.

takagi:
I've added experimental feature.

chainer/functions/pooling/max_pooling_nd.py
```diff
@@ -21,6 +21,13 @@ class MaxPoolingND(pooling_nd._PoolingND):
 
     """Max pooling over a set of N-dimensional planes."""
 
+    def __init__(self, ndim, ksize, stride=None, pad=0, cover_all=True,
+                 use_cudnn=True):
+        experimental('chainer.functions.pooling.MaxPoolingND')
```
delta2323: Could you use `util.experimental` as direct import of modules are prohibited by pep8. (Should we change the document of `chainer.util.experimental as well?`)

takagi: I've fixed it! It would be better to change the document as well.

mitmul:
LGTM 👍

delta2323:
Thank you for you considerable work!

PR #1355, opened 2016-07-04
Fix the description of cover_all option of unpooling_2d

### LINKED ISSUE #1341: Possibly wrong doc or code for deconvolution cover_all
The documentation says:

```
cover_all (bool): If True, all spatial locations are convoluted into
            some output pixels. It may make the output size larger.
```

However, when I prepare a small toy example, the output is larger when `cover_all=True`. I feel like either the code or the documentation needs to be inverted.

See an [gist notebook](https://gist.github.com/LukasDrude/8a9ebbaa3a6ba4ae0e2bef611afefd5a) for the toy example or the attached screenshot. I had set the weight matrices to ones and disabled normalization for clarity.

screen shot 2016-07-01 at 14 54 08

delta2323:
This PR fixes #1341

### DIFF
--- a/chainer/functions/pooling/unpooling_2d.py
+++ b/chainer/functions/pooling/unpooling_2d.py
@@ -89,7 +89,7 @@ def unpooling_2d(x, ksize, stride=None, pad=0, outsize=None, cover_all=True):
             :func:`~chainer.utils.conv.get_conv_outsize` must be equal to
             the shape of the 2d array in the input batch ``x``.
         cover_all (bool): If ``True``, all spatial locations are pooled
-            into some output pixels, and the output size is larger than that
+            into some output pixels, and the output size is smaller than that
             when cover_all is ``False``.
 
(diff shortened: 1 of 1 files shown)


chainer/functions/pooling/unpooling_2d.py
```diff
@@ -89,7 +89,7 @@ def unpooling_2d(x, ksize, stride=None, pad=0, outsize=None, cover_all=True):
             :func:`~chainer.utils.conv.get_conv_outsize` must be equal to
             the shape of the 2d array in the input batch ``x``.
         cover_all (bool): If ``True``, all spatial locations are pooled
```
zori: cover_all (bool): If ``True``, the output size may be smaller than if cover_all is ``False``. This flag serves to align behavior to the pooling functions which can cover all input locations, see :func:`max_pooling_2d`, :func:`convolution_2d`

PR #1358, opened 2016-07-05
Trainer2 multiprocess

amitibo:
The multiprocess iterator hangs while waiting for the worker processes to join.
I think there is a race condition that causes the workers to push more data into the out queue even after the iterator process finalizes. In this push I added exiting of the worker process without waiting for the queue thread to join:
[https://docs.python.org/2/library/multiprocessing.html#multiprocessing.Queue.cancel_join_thread](https://docs.python.org/2/library/multiprocessing.html#multiprocessing.Queue.cancel_join_thread)


beam2d:
Thank you for the PR! Can you omit the ParallelUpdater part from this PR?

beam2d:
LGTM! Some of my colleagues had been also trapped by this issue, and have confirmed that your PR works. Thank you for the helpful support!!

PR #1362, opened 2016-07-06
Add shuffle argument to the multiprocessiterator.

amitibo:
Allow for non shuffled multi-process iterator, similar to Sequential/ShuffledIterator.


beam2d:
Sorry, I overlooked this PR and implemented shuffle argument independently.

amitibo:
@beam2d no problem, don't miss #1372 :-), thanks

PR #1372, opened 2016-07-10
A feature extractor extension.

amitibo:
This pull request adds an ``Extractor`` extension. This extension can be used to extract features calculated using a trained model. Similarly to the ``Evaluator`` extension, it can be run standalone without a ``Trainer``.

An example on the ``mnist`` dataset is given.


beam2d:
Trainer2 branch has been merged to master. Would you resend this PR to master?

PR #1389, opened 2016-07-14
Fix broken documentations in maximum and minimum.

takagi:
This PR fixes broken documentations in maximum and minimum functions.


delta2323:
LGTM 👍

PR #1400, opened 2016-07-18
min/max value trigger

dsanno:
* Implement `MaxValueTrigger` that fires when specific value (e.g. the value associated with 'validation/main/accuracy') is larger than that of past epochs.
* Implement `MinValueTrigger` that fires when specific value is minimum.
* To take snapsot when specific value is minimum or maximum,  add `trigger` parameter to `snapshot_object` and `snapshot`.

This is [WIP] pull request because there is no test code.
I don't have an idea of policy to write test code of trigger and objects related to training.


delta2323:
Thank you for your PR. Is it ready for reviewing?

delta2323:
Is it possible to separate this PR into two, namely, one for changes of `snapshot` and the other one for `MaxValueTrigger` and `MinValueTrigger`, as these are logically separated?

delta2323:
@beam2d Do you have any policy for testing triggers?

dsanno:
> Is it possible to separate this PR into two, namely, one for changes of snapshot and the other one for MaxValueTrigger and MinValueTrigger, as these are logically separated?

I've separated changes of `snapshot` into #1450.
This PR is for `MaxValueTrigger` and `MinValueTrigger`.

uchida:
👍 and I have tried this and a following change is required to use the triggers in this PR.

```diff
--- a/setup.py
+++ b/setup.py
@@ -61,6 +61,7 @@ setup(
               'chainer.serializers',
               'chainer.testing',
               'chainer.training',
+              'chainer.training.triggers',
               'chainer.training.extensions',
               'chainer.utils',
               'cupy',
```

dsanno:
@uchida 
Thank you for your comment.
I've fixed. 01b80e915a4185550703b86747ddcc22e50078cf

chainer/training/triggers/minmax_value_trigger.py
```diff
@@ -0,0 +1,89 @@
...
+
+    Args:
+        key (str): Key of value.
+        compare (function): Compare function which takes current best value and
+        new value and returns whether new value is better than current best.
+
+    """
+    def __init__(self, key, compare):
+        self._key = key
+        self._min_value = None
```
delta2323: It is appropriate to replace it with `_best_value`?

delta2323:
Do you think we should expose `BestValueTrigger` to users?

delta2323:
Currently, `BestValueTrigger` accumulates the target value throughout the epoch and compares the value with the current best value only at the end of each epoch. We can make this trigger more flexible by adding an tuple that determines the comparison interval (e.g. `1, 'epoch'`) and create `IntervalTrigger` with the tuple.

chainer/training/triggers/minmax_value_trigger.py
```diff
@@ -0,0 +1,89 @@
...
+        self._interval_trigger = IntervalTrigger(1, 'epoch')
```
delta2323: Currently, `BestValueTrigger` accumulates the target value throughout the epoch and compares the value with the current best value only at the end of each epoch. We can make this trigger more flexible by adding an tuple that determines the comparison interval (e.g. `1, 'epoch'`) and create `IntervalTrigger` with the tuple.

delta2323:
I talked with @beam2d. According to him, there are no policy for testing trigger. He said he had discussed how to test extensions and thought triggers would test in the similar way. One of the cheapest ways to test them is to use mock. But it should be the last option.

chainer/training/triggers/minmax_value_trigger.py
dsanno: Fixed 329ab35f55fccec19638cdea4db8e2368c96bc42

chainer/training/triggers/minmax_value_trigger.py
dsanno: It's a nice idea.
I've fixed `BestValueTrigger` to accept `trigger` option.
7be533b0fd879edd093f765a6e4ed75b659a9079

dsanno:
Thank you for your review.
I've fixed according to your comments and
* Added unit test b441cda5885c3da1f4c42105e7cb2cfabc058a5e
* I also fixed `BestValueTrigger`  to avoid keeping best value on GPU memory f81ea5dcd13d0883904f774f928c17804cb476ea

dsanno:
> @dsanno Do you think we should expose BestValueTrigger to users as well?

I don't think of the case to use compare function other than simple comparison of volume,
so I don't thik it is necessary to expose `BestValueTrigger `.

delta2323:
Now that we add unit tests for the classes, can we remove [WIP] from this PR?

tests/chainer_tests/training_tests/triggers_tests/test_minmax_trigger.py
```diff
@@ -0,0 +1,64 @@
+import unittest
+
+import numpy
+
+from chainer import testing
+from chainer import training
+from chainer.training import triggers
+
+
+class dummyUpdater(training.Updater):
```
delta2323: Capitalize the first character of class names.

chainer/training/triggers/minmax_value_trigger.py
```diff
@@ -0,0 +1,102 @@
...
+        trigger: Trigger that decides the comparison interval between current
+            best value and new value. This must be a tuple in the form of
+            ``<int>, 'epoch'`` or ``<int>, 'iteration'`` which is passed to
+            :class`IntervalTrigger`.
+
+    """
+    def __init__(self, key, compare, trigger=(1, 'epoch')):
+        self._key = key
+        self._best_value = None
+        self._interval_trigger = IntervalTrigger(*trigger)
```
delta2323: You can use [get_trigger](https://github.com/pfnet/chainer/blob/master/chainer/training/trigger.py#L44) to create `Trigger` object. If you use this function, `trigger` argument can be either of a callable, None, or tuple.

chainer/training/triggers/minmax_value_trigger.py
```diff
@@ -0,0 +1,102 @@
...
+    def __call__(self, trainer):
+        """Decides whether the extension should be called on this iteration.
+
+        Args:
+            trainer (Trainer): Trainer object that this trigger is associated
+                with. The updater associated with this trainer is used to
+                determine if the trigger should fire.
+
+        Returns:
+            bool: True if the corresponding extension should be invoked in this
```
delta2323: Enclose True with double backquotes.

chainer/training/triggers/minmax_value_trigger.py
```diff
@@ -0,0 +1,102 @@
...
+        self._best_value = None
+        self._interval_trigger = IntervalTrigger(*trigger)
+        self._init_summary()
+        self._compare = compare
+
+    def __call__(self, trainer):
+        """Decides whether the extension should be called on this iteration.
+
+        Args:
+            trainer (Trainer): Trainer object that this trigger is associated
```
delta2323: trainer (`~chainer.training.Trainer`): ...

tests/chainer_tests/training_tests/triggers_tests/test_minmax_trigger.py
dsanno: Fixed c3c8c1ad98634fcb9b7d37ac5b5feb64fc13e47f

chainer/training/triggers/minmax_value_trigger.py
dsanno: Fixed 5ec9913f2aa414d317b122cd74f641c51338ebd0

chainer/training/triggers/minmax_value_trigger.py
dsanno: Fixed 120dea9c9355e97d13658dc994dd76a57d9f7cbc

chainer/training/triggers/minmax_value_trigger.py
dsanno: Fixed 120dea9c9355e97d13658dc994dd76a57d9f7cbc

dsanno: Also I fixed the comment "The updater associated with this trainer is" to "The ``observation`` of this trainer is" to match actual implementation.

dsanno:
Thank you for your review. I've updated PR.

delta2323:
@beam2d Is it OK to create new directory `training/triggers` and put new Triggers to it?

delta2323:
LGTM 👍 . I will ask other developer team mates the appropriate place to put these triggers.

beam2d:
@delta2323 OK, but it would be better to put IntervalTrigger under training/triggers, too (it can be deferred to another PR).

delta2323:
@beam2d Thank you for comments, I agree with your idea that we put `IntervalTrigger` to the same directory.

PR #1402, opened 2016-07-19
[WIP] Implements cupy.linalg

rezoo:
This PR aims to implement `cupy.linalg` functions using CUSolver, which provides LAPACK-like features, and make `cupy` module applicable to various types of large-scale optimization problems.


delta2323:
Is it possible to separate this PR into smaller ones? As this PR implements Cholesky, QR, and SVD. The first PR would be to add either of them.

rezoo:
Good idea, but currently I have to add a build function to check whether it contains cusolver and the version is up to date. After finishing that job I will do split this PR into the smaller ones.

rezoo:
Finished the implementation! Since the current commits are like an intricate jigsaw puzzle, I will make additional PRs.

Issue #1294, opened 2016-06-20
CuPy fill_diagonal


okuta:
We need `fill_diagonal`.

https://github.com/pfnet/chainer/blob/ed8ef4786be325899119c3827febc9286da7e354/cupy/indexing/insert.py#L13
http://docs.scipy.org/doc/numpy-1.11.0/reference/generated/numpy.fill_diagonal.html

PR #1404, opened 2016-07-20
Implement cupy.fill_diagonal()

rezoo:
Related to #1294 

http://docs.scipy.org/doc/numpy-1.11.0/reference/generated/numpy.fill_diagonal.html

### DIFF
--- a/cupy/indexing/insert.py
+++ b/cupy/indexing/insert.py
@@ -1,5 +1,4 @@
-# flake8: NOQA
-# "flake8: NOQA" to suppress warning "H104  File contains nothing but comments"
+import numpy
 
 # TODO(okuta): Implement place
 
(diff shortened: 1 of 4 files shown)


cupy/indexing/insert.py
```diff
@@ -10,4 +9,85 @@
...
+    if a.ndim == 2:
+        step = a.shape[1] + 1
+        if not wrap:
+            end = a.shape[1] * a.shape[1]
+    else:
+        if not numpy.alltrue(numpy.diff(a.shape) == 0):
+            raise ValueError("All dimensions of input must be of equal length")
+        step = 1 + numpy.cumprod(a.shape[:-1]).sum()
+
+    # we use a.ravel() instead of a.flat
```
muupan: Please provide the reason why we do so, because it is different from the original numpy code. I guess it is because current `cupy.ndarray` doesn't support `.flat`.

tests/cupy_tests/indexing_tests/test_insert.py
```diff
@@ -7,3 +7,31 @@
 class TestInsert(unittest.TestCase):
 
     _multiprocess_can_split_ = True
+
+    @testing.for_all_dtypes()
+    @testing.numpy_cupy_array_equal()
+    def test_fill_diagonal_shape1(self, xp, dtype):
+        a = testing.shaped_arange((3, 3), xp, dtype)
+        xp.fill_diag(a, val=0)
```
muupan: `fill_diag` may be typo of `fill_diagonal`?

Please specify `accept_error=False` for `testing.numpy_cupy_array_equal` unless you actually intend to accept errors in test code. Otherwise, errors such as typo are just ignored and tests would still pass (see https://github.com/pfnet/chainer/issues/1419.)

tests/cupy_tests/indexing_tests/test_insert.py
```diff
@@ -7,3 +7,31 @@
...
+        a = testing.shaped_arange((3, 5), xp, dtype)
+        xp.fill_diag(a, val=0, wrap=True)
+        return a
+
+    @testing.for_all_dtypes()
+    @testing.numpy_cupy_array_equal()
+    def test_fill_diagonal_wrap2(self, xp, dtype):
+        a = testing.shaped_arange((5, 3), xp, dtype)
+        xp.fill_diag(a, val=0, wrap=True)
+        return a
```
muupan: Please add tests that specifies `wrap=False` to tall and wide matrices.

I suggest to use `testing.parameterize`, for example:
```
    'shape': [(3, 3), (2, 2, 2), (3, 5), (5, 3)],
    'val': [1, 0],
    'wrap': [True, False],
}))
@testing.gpu
class TestInsert(unittest.TestCase):

    @testing.for_all_dtypes()
    @testing.numpy_cupy_array_equal(accept_error=False)
    def test_fill_diagonal(self, xp, dtype):
        a = testing.shaped_arange(self.shape, xp, dtype)
        xp.fill_diagonal(a, val=self.val, wrap=self.wrap)
        return a
```

muupan:
LGTM except the comments.

cupy/indexing/insert.py
rezoo: Done.

tests/cupy_tests/indexing_tests/test_insert.py
rezoo: Done.

tests/cupy_tests/indexing_tests/test_insert.py
rezoo: Done.

rezoo:
I fixed the above problems. Please check it.

muupan:
Thanks. LGTM

cupy/indexing/insert.py
```diff
@@ -10,4 +9,86 @@
...
+    modifies the input array in-place, it does not return a value.
+
+    Args:
+        a (cupy.ndarray): The array, at least 2-D.
+        val (scalar): The value to be written on the diagonal.
+            Its type must be compatible with that of the array a.
+        wrap (bool): If specified, the diagonal is "wrapped" after N columns.
+            This affects only tall matrices.
+
+    Examples
```
beam2d: I think it would be enough to just let users see NumPy documentation, because the examples are identical to those of NumPy's one.

rezoo: This means that I should simplify the documentation like the other ones in cupy, right?

tests/cupy_tests/indexing_tests/test_insert.py
```diff
@@ -3,7 +3,19 @@
...
 class TestInsert(unittest.TestCase):
 
     _multiprocess_can_split_ = True
+
+    @testing.for_all_dtypes()
+    @testing.numpy_cupy_array_equal(accept_error=False)
+    def test_fill_diagonal(self, xp, dtype):
+        a = testing.shaped_arange(self.shape, xp, dtype)
+        xp.fill_diagonal(a, val=self.val, wrap=self.wrap)
+        return a
```
beam2d: Could you add a fail test? (e.g. for arrays with less than 2 dimensions)

cupy/indexing/insert.py
rezoo: Done.

tests/cupy_tests/indexing_tests/test_insert.py
rezoo: Done.

rezoo:
@beam2d I fixed the problems you have mentioned. Please check it.

beam2d:
LGTM! (sorry for being late)

PR #1406, opened 2016-07-20
Add stack function missed in Chainer reference.

takagi:
This PR adds `stack` function to Chainer reference, it suppose to be missed in #1262 .


okuta:
LGTM!

PR #1409, opened 2016-07-21
[Experimental] patch toward out-of-core training on GPU

anaruse:
I've implemented a method "to_swap()" that allow you to swap-out any cupy data on GPU memory to page-locked memory on CPU (so called pinned memory).  You can use "to_gpu()" to swap-in the data on pinned memory back to GPU memory.
You may wonder what is difference of "to_swap()" to "to_cpu()".  The key difference is whether asynchronous data transfer is naturally available or not.  As you know, "to_cpu()" moves (and transforms as well?)  cupy data on GPU to numpy data on CPU.  Considering numy data is typically on standard CPU memory, in other words, page-able memory, you have almost no chance to see real asynchronous data transfer with "to_cpu()" unless you change numpy's memory allocator.  However, if you use this "to_swap()", you can move data asynchronously and accordingly you may run data transfer and GPU kernels in parallel (that is my goal).
This is kind of experimental patch, so any feedback would be appreciated.
BR.

### DIFF
--- a/chainer/cuda.py
+++ b/chainer/cuda.py
@@ -193,6 +193,10 @@ def to_gpu(array, device=None, stream=None):
     """
     check_cuda_available()
     with get_device(device):
+        if isinstance(array, ndarray) and array.data.mem._swap is True:
+            # anaruse: move data from SWAP to GPU
+            return cupy.array(array, copy=True, stream=stream)
+
         array_dev = get_device(array)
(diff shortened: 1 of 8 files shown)


okuta:
Thank you for your PR.
This function is interesting. It is very important to reduce the memory usage of GPU.

But, this PR is difficult to merge because it has no backward compatibility. 
Could you create an issue about this function. In order to be able to discussion.
And, please close this PR.

Issue #1394, opened 2016-07-15
FileExistsError at first epoch finished


danpansa:
I try the Chainer v1.11.0 on windows 10(64bit). My python is python3.5 on anaconda. I tested the MNIST sample code was obtained from  http://qiita.com/_329_/items/bcc306194d52f7b81b5a. The code was good working at start. However, it stopped when the first epoch had been finished.  The error message is below. Is this error global or depend on windows 10?
---------------------------------------------------------------------------
FileExistsError                           Traceback (most recent call last)
<ipython-input-49-a920a1206cd1> in <module>()
     33 trainer.extend(extensions.ProgressBar())
     34 
---> 35 trainer.run()

C:\Anaconda3\lib\site-packages\chainer\training\trainer.py in run(self)
    266                     for name, entry in extensions:
    267                         if entry.trigger(self):
--> 268                             entry.extension(self)
    269         finally:
    270             for _, entry in extensions:

C:\Anaconda3\lib\site-packages\chainer\training\extensions\log_report.py in __call__(self, trainer)
     92                 with os.fdopen(fd, 'w') as f:
     93                     json.dump(self._log, f, indent=4)
---> 94                 os.rename(path, os.path.join(trainer.out, log_name))
     95             # reset the summary for the next output
     96             self._init_summary()

FileExistsError: [WinError 183] Cannot create a file when that file already exists: 'C:\\Users\\name\\chainer1.11\\result\\logq7tunkxu' -> 'result\\log'

td2sk:
On windows (only windows?), os.rename cannot override file that already exists.
I now use work-around below.

94  log_path = os.path.join(os.path.join(trainer.out, log_name))
95  if os.path.exists(log_path):
96     os.remove(log_path)
97  os.rename(path, log_path)

danpansa:
Thank you very much for quick response, and I'm sorry for my late reply.

I fixed the "log_report.py" according to  your reply, I found that the problem had fixed.

Thank you for your help! I'm closing this issue.

delta2323:
Thank you for reporting the problem. I sent the PR that fixes this problem, although this issue is closed.

delta2323:
I fixed the implementation in #1412 . As I do not have Windows environment, so does anyone check it with Windows?

okuta:
I checked #1412 on Windows 10.

PR #1412, opened 2016-07-22
Remove log file if it exists in LogReport

delta2323:
This PR fixes the problem reported in #1394.

### DIFF
--- a/chainer/training/extensions/log_report.py
+++ b/chainer/training/extensions/log_report.py
@@ -91,7 +91,11 @@ class LogReport(extension.Extension):
                 fd, path = tempfile.mkstemp(prefix=log_name, dir=trainer.out)
                 with os.fdopen(fd, 'w') as f:
                     json.dump(self._log, f, indent=4)
-                os.rename(path, os.path.join(trainer.out, log_name))
+
+                log_path = os.path.join(trainer.out, log_name)
+                if os.path.exists(log_path):
+                    os.remove(log_path)
+                os.rename(path, log_path)
 
             # reset the summary for the next output
             self._init_summary()
--- a/chainer/training/extensions/log_report.py
+++ b/chainer/training/extensions/log_report.py
@@ -1,5 +1,6 @@
 import json
 import os
+import shutil
 import tempfile
 
 import six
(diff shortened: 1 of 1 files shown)


chainer/training/extensions/log_report.py
```diff
@@ -91,7 +91,11 @@ def __call__(self, trainer):
                 fd, path = tempfile.mkstemp(prefix=log_name, dir=trainer.out)
                 with os.fdopen(fd, 'w') as f:
                     json.dump(self._log, f, indent=4)
-                os.rename(path, os.path.join(trainer.out, log_name))
+
+                log_path = os.path.join(trainer.out, log_name)
+                if os.path.exists(log_path):
+                    os.remove(log_path)
+                os.rename(path, log_path)
```
okuta: How about using `shutil.move`?
Please see also https://docs.python.org/3/library/shutil.html#shutil.move

delta2323: Thank you for the suggestion. I guess `shutil.move` cannot solve the problem occurred in Windows. According to the document, if the destination exists and is not a directory, `os.move` is used to overwrite the destination. But on Windows, `os.move` raises `OSError` when the destination exists (regardless of the type of the destination).

beam2d: Can we use `os.replace`?

beam2d: Oh, I found that `os.replace` is not available in Py2. Sorry.

okuta: ``shutil.move`` caches ``OSError``.
https://github.com/python/cpython/blob/master/Lib/shutil.py#L537

delta2323:
I fixed the code.

chainer/training/extensions/log_report.py
miyazakishogun: I have tried on Windows. `shutil.move` worked well. Also, `os.path.exists` version worked well. In addition, `os.path.isfile` instead of `os.path.exists` also worked well.

okuta:
@delta2323 I checked on Windows 10.
LGTM!

delta2323:
Thank you!

Issue #1384, opened 2016-07-13
chainer v1.11.0 ptb example training with --test option


odanado:
I run ptb example with --test option:

```
./train_ptb.py -g 0 --test
```

Screen during training:

```
#vocab = 10000
     total [##############################################################] 125.64%
this epoch [..................................................]  0.00%
         7 iter, 49 epoch / 39 epochs
   0.15797 iters/sec. Estimated time to finish: -1 day, 23:59:50.956683.
```

result is:

```
#vocab = 10000
epoch       iteration   perplexity  val_perplexity
70          10          3918.07     43725.3
     total [###############################################################################################################################################] 287.18     total [#######################################################################################################################################################     total [#######################################################################################################################################################     total [#######################################################################################################################################################140         20          1003.55     45304.3
     total [#######################################################################################################################################################     total [#######################################################################################################################################################     total [#######################################################################################################################################################     total [#######################################################################################################################################################     total [#######################################################################################################################################################     total [#######################################################################################################################################################     total [#######################################################################################################################################################     total [#######################################################################################################################################################     total [#######################################################################################################################################################     total [#######################################################################################################################################################210         30          205.992     13718.8
     total [#######################################################################################################################################################     total [#######################################################################################################################################################     total [#######################################################################################################################################################     total [#######################################################################################################################################################     total [#######################################################################################################################################################     total [##########################################################################################################################################################################################################################################################################################################################     total [##########################################################################################################################################################################################################################################################################################################################     total [##########################################################################################################################################################################################################################################################################################################################     total [##########################################################################################################################################################################################################################################################################################################################     total [##########################################################################################################################################################################################################################################################################################################################test
test perplexity: 28380.350504
```

But when the number of epochs is multiple of 7 (e.g. 35), training is a success:

```
./train_ptb.py -g 0 --test -e 35
```

result: 

```
#vocab = 10000
test
test perplexity: 4670.25469318
```

Smerity:
This is also a problem when it comes to the proper training as well, not just `--test`, so it's likely an issue with the epoch iterator or similar?

kissge:
Until this bug would be fixed, how about changing the default value to 35, not 39?
Otherwise new users might get confused.

delta2323:
Now that #1423 is merged, the script stops even if the number of epochs is not divisible by 7.

[...]

PR #1423, opened 2016-07-26
Make IntervalTrigger stateful to fix PTB example bug #1384

jekbradbury:
Bug #1384 is caused by epochs in PTB not being evenly divisible into iterations. An IntervalTrigger set on a particular number of epochs should trigger after exactly that many epochs, if that corresponds to an integer number of iterations, or on the first iteration after exactly that many epochs, if it doesn't.


delta2323:
Thank you for your PR. Is it possible to write unitests?

delta2323:
@jekbradbury How are you? If you cannot spare time, I will write unittests and send PR to your repository. I would appreciate you if you let me know the way that is convenient for you?

jekbradbury:
Thank you for reminding me -- I will write unittests this weekend.

delta2323:
Thank you! Let me know if you encounter any problem.

delta2323:
Is it OK to start reviewing this PR?

jekbradbury:
Yes, I am done pushing

chainer/training/trigger.py
```diff
@@ -36,7 +40,8 @@ def __call__(self, trainer):
         """
         updater = trainer.updater
         if self.unit == 'epoch':
-            return updater.is_new_epoch and updater.epoch % self.period == 0
+            prev, self.count = self.count, updater.epoch_detail // self.period
```
delta2323: It is better to split this assignment into two.

tests/chainer_tests/training_tests/triggers_tests/test_interval_trigger.py
```diff
@@ -0,0 +1,80 @@
...
+    def epoch(self):
+        return self.iteration // self.iters_per_epoch
+
+    @property
+    def epoch_detail(self):
+        return self.iteration / self.iters_per_epoch
+
+    @property
+    def is_new_epoch(self):
+        return 0 <= self.iteration % self.iters_per_epoch < 1
```
delta2323: I guess `self.iteration % self.iters_per_epoch == 0` is simpler, if I understand your intension correctly.

jekbradbury: No, because `self.iters_per_epoch` may not be an integer (in this dummy iterator, which replicates the behavior of the real PTB iterator where iterations don’t evenly divide epochs)

delta2323: Ah, I did not expect `iters_per_epoch` to be float. I will check if `IntervalTrigger` works even in this situation.

chainer/training/extensions/log_report.py
```diff
@@ -80,6 +80,7 @@ def __call__(self, trainer):
 
             updater = trainer.updater
             stats_cpu['epoch'] = updater.epoch
+            stats_cpu['epoch_detail'] = updater.epoch_detail
```
delta2323: Is this change related to the bug in `IntervalTrigger`?

jekbradbury: No, I must have included it accidentally.

delta2323: Then, could you remove it? Also, could you send it as an another PR if you think it is worth doing so.

chainer/training/trigger.py
```diff
@@ -20,10 +20,14 @@ def __init__(self, period, unit):
         self.period = period
         assert unit == 'epoch' or unit == 'iteration'
         self.unit = unit
+        self.count = 0
 
     def __call__(self, trainer):
         """Decides whether the extension should be called on this iteration.
 
+        In order to support non-integer epoch intervals, it does not rely on
+        ``updater.iterator.is_new_epoch``.
```
delta2323: So this description is for developers of Chainer, not its users because it explains *how* it is implemented not *what* is implemented. Further, I do not think it is natural for users to use `updater.is_new_epoch` to realize this method.

So I think we need not to write it (or we write the motivation of the design a  note with slight detailed background).  For example, how about the followings?
"For developers: it would be natural at first sight to use updater.is_new_epoch to decide if the trigger fires or not when `unit` is 'epoch'. But we cannot judge solely from it if the period is floating values. That is why IntervalTrigger directly sees updater.epoch_detail and decides if it fires."

delta2323:
In the original document, we restrict `period` argument to be integer as you can see [here](https://github.com/pfnet/chainer/blob/master/chainer/training/triggers/interval.py#L13). So could you change it as well?

delta2323:
The conflict is derived from the change by#1506, in which `IntervalTrigger` is moved under `training.triggers` namespace. So could you merge or rebase the latest master branch?

delta2323:
In the current implementation, we allow `period` to be float even if `unit` is `iteration`. Do you think it is acceptable? It could do harm to the operation [here](https://github.com/pfnet/chainer/blob/master/chainer/training/triggers/interval.py#L42). So I think we should change it if we want to do this extension.

jekbradbury:
Well, the period should have to be an `int` if the unit is `iteration` -- there's no such thing as a partial iteration. Is it ok to specify that the period must be an `int` if the unit is `iteration` but it can be a `float` if the unit is `epoch`?

delta2323:
> Is it ok to specify that the period must be an int if the unit is iteration but it can be a float if the unit is epoch?

Yes, it is reasonable.

delta2323:
A comment on the conflict occurred. We moved `IntervalTrigger` to `training.triggers` namespace, so could you rebase the master branch to resolve the conflict?

delta2323:
Also, I would like you to consider the comment on the document of `__call__` method.

delta2323:
I ran the PTB example in test mode. The example did stop, but the number of epoch processed is longer than specified. Specifically, we set the number epoch to 39 in test mode. But as 1 iteration consumes 7 epochs, the test finishes with 42 epochs, the least number which is multiple of 7 and is larger than or equals to 39.

I would like to ask if you think it is acceptable behavior or we should fix it, although I do not think we need not include this change if any.

delta2323:
LGTM, I am now running on the GPU server. I will merge it when I confirm the test passes.

delta2323:

PR #1430, opened 2016-07-26
Show progress bar at first iteration

okuta:
```
epoch       main/loss   validation/main/loss  main/accuracy  validation/main/accuracy
     total [..................................................]  0.83%
this epoch [########..........................................] 16.67%
       100 iter, 0 epoch / 20 epochs
       inf iters/sec. Estimated time to finish: 0:00:00.
```

### DIFF
--- a/chainer/training/extensions/progress_bar.py
+++ b/chainer/training/extensions/progress_bar.py
@@ -27,8 +27,6 @@ class ProgressBar(extension.Extension):
 
     """
 
-    invoke_before_training = False
-
     def __init__(self, training_length=None, update_interval=100,
                  bar_length=50, out=sys.stdout):
         self._training_length = training_length
--- a/chainer/training/extensions/progress_bar.py
+++ b/chainer/training/extensions/progress_bar.py
@@ -27,6 +27,8 @@ class ProgressBar(extension.Extension):
 
     """
 
+    invoke_before_training = False
+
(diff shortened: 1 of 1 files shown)


chainer/training/extensions/progress_bar.py
```diff
@@ -26,6 +26,9 @@ class ProgressBar(extension.Extension):
         out: Stream to print the bar. Standard output is used by default.
 
     """
+
+    invoke_before_training = True
```
ir5: Well, why do we need to invoke this extension before training? When I ran mnist example, this progress bar was shown for just a moment at first, followed by `PrintReport` output. This was a little strange.
When I ran the example with `invoke_before_training = False`, the output was good. So I suspect that this is actually unnecessary.

okuta:
Sorry, I changed this PR title.

chainer/training/extensions/progress_bar.py
ir5: I'm not sure if you understand what I'm saying, so I put a screen capture here. You can see progress bar disappears after the `PrintReport` output.
aaa

okuta: Oh, I understood.
I will fix and change title.

ir5:
Now the code looks good to me.

beam2d:
LGTM

PR #1434, opened 2016-07-27
Deduplicate Chain.zerograd

okuta:
See also #877

### DIFF
--- a/chainer/link.py
+++ b/chainer/link.py
@@ -315,8 +315,9 @@ class Link(object):
         iteration of the optimization.
 
         """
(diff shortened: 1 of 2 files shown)


chainer/link.py
```diff
@@ -315,8 +315,9 @@ def zerograds(self):
         iteration of the optimization.
 
         """
-        for param in self.params():
-            param.zerograd()
```
gwtnb: It seems better to remove `Chain.zerograds` and `ChainList.zerograds` than removing the above statement.

PR #1435, opened 2016-07-27
Use memset in cupy.ndarray.fill

okuta:
Calling ``elementwise_copy`` is high cost.

### DIFF
--- a/cupy/core/core.pyx
+++ b/cupy/core/core.pyx
@@ -6,6 +6,7 @@ import numpy
 import six
 
 from cupy.core import flags
+from cupy.cuda import stream
 from cupy import util
 
 cimport cpython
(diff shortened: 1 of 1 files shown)


cupy/core/core.pyx
```diff
@@ -351,7 +352,10 @@ cdef class ndarray:
         .. seealso:: :meth:`numpy.ndarray.fill`
 
         """
-        elementwise_copy(value, self, dtype=self.dtype)
+        if value == 0 and self._c_contiguous:
+            self.data.memset_async(0, self.nbytes, stream.Stream(True))
```
gwtnb: This memset ignores a sign of the zero. Is it okay? (I think it doesn't matter.)

okuta: I think okay.

PR #1444, opened 2016-07-30
Reduce extra copy

okuta:
`astype` returns new array in default mode.
This PR uses `copy=False` option.
And, PR extends `utils.force_array` that receives `dtype` argument.


gwtnb:
LGTM

PR #1447, opened 2016-08-01
Fix incorrect argument name in docstring

iwiwi:
There is no `eval_loop` argument. I suppose it should be `eval_func`.


okuta:
LGTM!

Issue #1422, opened 2016-07-25
chainer.datasets.get_cifar*() not working due to incorrect path


yanii:
At least in Linux (ubuntu 15.10), the chainer.datasets.get_cifar*() function is currently broken:

> Traceback (most recent call last):
  File "./train_cifar10.py", line 114, in <module>
    main()
  File "./train_cifar10.py", line 74, in main
    train, val = datasets.get_cifar10()
  File "/usr/local/lib/python2.7/dist-packages/chainer-1.11.0-py2.7-linux-x86_64.egg/chainer/datasets/cifar.py", line 41, in get_cifar10
    raw = _retrieve_cifar('cifar-10')
  File "/usr/local/lib/python2.7/dist-packages/chainer-1.11.0-py2.7-linux-x86_64.egg/chainer/datasets/cifar.py", line 139, in _retrieve_cifar
    return download.cache_or_load_file(path, creator, numpy.load)
  File "/usr/local/lib/python2.7/dist-packages/chainer-1.11.0-py2.7-linux-x86_64.egg/chainer/dataset/download.py", line 145, in cache_or_load_file
    content = creator(temp_path)
  File "/usr/local/lib/python2.7/dist-packages/chainer-1.11.0-py2.7-linux-x86_64.egg/chainer/datasets/cifar.py", line 122, in creator
    d = pickle.load(archive.extractfile(file_name))
  File "/usr/lib/python2.7/tarfile.py", line 2136, in extractfile
    tarinfo = self.getmember(member)
  File "/usr/lib/python2.7/tarfile.py", line 1821, in getmember
    raise KeyError("filename %r not found" % name)
KeyError: "filename 'cifar-10-batches-py//data_batch_1' not found"

This is easily fixed by the following change in path to avoid the double slash (which for some reason isn't treated as a valid path?):

```
diff --git a/chainer/datasets/cifar.py b/chainer/datasets/cifar.py
index 694258d..b8b0f90 100644
--- a/chainer/datasets/cifar.py
+++ b/chainer/datasets/cifar.py
@@ -113,7 +113,7 @@ def _retrieve_cifar(name):
         train_y = numpy.empty((5, 10000), dtype=numpy.uint8)
         test_y = numpy.empty(10000, dtype=numpy.uint8)
 
-        dir_name = '{}-batches-py/'.format(name)
+        dir_name = '{}-batches-py'.format(name)
 
         with tarfile.open(archive_path, 'r:gz') as archive:
             # training set
```

[...]

PR #1449, opened 2016-08-01
modify datasets/cifar.py

owruby:
#1422 
I tried get_cifar10 method, But output error.
This is typo?


zori:
LGTM, thank you for the fix.

delta2323:
I would like to confirm my understanding. This PR fixes two problems: one is #1422 and the other is that `_retrieve_cifar`, which is called internally by `get_cifar10` and `get_cifar100` returns different objects when the source data is cached and downloaded.

zori:
Yes, without the second fix it, after once downloading the DS to disk, on the second run it raises `KeyError: 'train_x is not a file in the archive'`

PR #1457, opened 2016-08-02
Implement zoneout #1339

tohmae:
I implemented zoneout function and LSTM with zoneout #1339 .
Test Codes has not been implemented yet.

### DIFF
--- a/chainer/functions/noise/zoneout.py
+++ b/chainer/functions/noise/zoneout.py
@@ -23,7 +23,7 @@ class Zoneout(function.Function):
         else:
             flag_x = (xp.random.rand(*x.shape) >=
                       self.zoneout_ratio)
-        self.flag_h = xp.ones_like(flag_x) - flag_x
+        self.flag_h = xp.ones_like(flag_x) ^ flag_x
         self.flag_x = flag_x
         return h * self.flag_h + x * self.flag_x,
 
--- a/chainer/functions/noise/zoneout.py
+++ b/chainer/functions/noise/zoneout.py
@@ -7,7 +7,12 @@ from chainer.utils import type_check
 
 class Zoneout(function.Function):
 
-    """Zonepout regularization."""
+    """Zoneout regularization.
+
+    See the paper: `Zoneout: Regularizing RNNs by Randomly Preserving Hidden \
+    Activations <http://arxiv.org/abs/1606.013050>`_.
+
+    """
 
     def __init__(self, zoneout_ratio):
         self.zoneout_ratio = zoneout_ratio
@@ -37,20 +42,17 @@ class Zoneout(function.Function):
 def zoneout(h, x, ratio=.5, train=True):
     """Instead of dropping out, units zone out and are set to their previous value.
 
-    This function adds their previous value randomly with probability ``ratio``
-    to hidden activation ``x``. In testing mode, returns ``x``.
+    This function stochastically forces some hidden units to maintain their
+    previous values.
 
     Args:
         h (~chainer.Variable): Previous variable.
         x (~chainer.Variable): Input variable.
         ratio (float): Zoneout ratio.
-        train (bool): If ``True``, executes dropout. Otherwise, return x.
+        train (bool): If ``True``, executes zoneout. Otherwise, return x.
 
     Returns:
         ~chainer.Variable: Output variable.
-
-    See the paper: `Zoneout: Regularizing RNNs by Randomly Preserving Hidden \
-    Activations <http://arxiv.org/abs/1606.013050>`_.
 
     """
     if train:
--- a/chainer/functions/noise/zoneout.py
+++ b/chainer/functions/noise/zoneout.py
@@ -7,12 +7,7 @@ from chainer.utils import type_check
 
 class Zoneout(function.Function):
 
-    """Zoneout regularization.
-
-    See the paper: `Zoneout: Regularizing RNNs by Randomly Preserving Hidden \
-    Activations <http://arxiv.org/abs/1606.013050>`_.
-
-    """
+    """Zoneout regularization."""
 
     def __init__(self, zoneout_ratio):
         self.zoneout_ratio = zoneout_ratio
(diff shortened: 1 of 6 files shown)


chainer/functions/noise/zoneout.py
```diff
@@ -0,0 +1,58 @@
+import numpy
+
+from chainer import cuda
+from chainer import function
+from chainer.utils import type_check
+
+
+class Zoneout(function.Function):
+
+    """Zonepout regularization."""
```
delta2323: Could you add the reference to this regularization if any?

chainer/functions/noise/zoneout.py
```diff
@@ -0,0 +1,58 @@
...
+        h (~chainer.Variable): Previous variable.
+        x (~chainer.Variable): Input variable.
+        ratio (float): Zoneout ratio.
+        train (bool): If ``True``, executes dropout. Otherwise, return x.
+
+    Returns:
+        ~chainer.Variable: Output variable.
+
+    See the paper: `Zoneout: Regularizing RNNs by Randomly Preserving Hidden \
+    Activations <http://arxiv.org/abs/1606.013050>`_.
```
delta2323: Should it be `http://arxiv.org/abs/1606.01305` ?

chainer/functions/noise/zoneout.py
```diff
@@ -0,0 +1,58 @@
...
+    def __init__(self, zoneout_ratio):
+        self.zoneout_ratio = zoneout_ratio
+
+    def check_type_forward(self, in_types):
+        type_check.expect(in_types.size() == 2)
+
+    def forward(self, inputs):
+        h, x = inputs
+        xp = cuda.get_array_module(*x)
+        if xp == numpy:
```
delta2323: `xp is numpy` is more efficient.

chainer/functions/noise/zoneout.py
```diff
@@ -0,0 +1,58 @@
...
+            flag_x = xp.random.rand(*h[0].shape) >= self.zoneout_ratio
+        else:
+            flag_x = (xp.random.rand(*h[0].shape, dtype=numpy.float32) >=
```
delta2323: What is the reason why you use float32 instead of float64(default)? I am wondering if we need it as we convert it into boolean array anyway.

chainer/functions/noise/zoneout.py
```diff
@@ -0,0 +1,58 @@
...
+                      self.zoneout_ratio)
+        flag_h = xp.ones_like(flag_x) - flag_x
```
delta2323: We need not intermediate variable `flag_h` and should directly assign it to `self.flag_h`.

chainer/functions/noise/zoneout.py
```diff
@@ -0,0 +1,58 @@
...
+        return h * self.flag_h + x * self.flag_x,
+
+    def backward(self, inputs, gy):
+        h, x = inputs
+
+        return gy[0] * self.flag_h, gy[0] * self.flag_x,
+
+
+def zoneout(h, x, ratio=.5, train=True):
+    """Instead of dropping out, units zone out and are set to their previous value.
```
delta2323: Although the wording "zone out" is used in the original paper, I think it is better to avoid to use it as it is figurative. How about "units are stochastically retain their previous value"?

chainer/functions/noise/zoneout.py
```diff
@@ -0,0 +1,58 @@
...
+
+    This function adds their previous value randomly with probability ``ratio``
+    to hidden activation ``x``. In testing mode, returns ``x``.
+
+    Args:
+        h (~chainer.Variable): Previous variable.
+        x (~chainer.Variable): Input variable.
+        ratio (float): Zoneout ratio.
+        train (bool): If ``True``, executes dropout. Otherwise, return x.
```
delta2323: zoneout, not dropout.

chainer/links/connection/zoneoutlstm.py
```diff
@@ -0,0 +1,113 @@
...
+        Args:
+            c (~chainer.Variable): A new cell states of LSTM units.
+            h (~chainer.Variable): A new output at the previous time step.
+
+        """
+        assert isinstance(c, chainer.Variable)
+        assert isinstance(h, chainer.Variable)
+        c_ = c
+        h_ = h
+        if self.xp == numpy:
```
delta2323: `self.xp is numpy` is better

chainer/links/connection/zoneoutlstm.py
```diff
@@ -0,0 +1,113 @@
...
+from chainer.functions.array import reshape
+from chainer.functions.array import split_axis
+from chainer.functions.noise import zoneout
+
+from chainer import link
+from chainer.links.connection import linear
+from chainer import variable
+
+
+class ZoneoutLSTM(link.Chain):
```
delta2323: Could you change the name of the chain to "StatefulZoneoutLSTM" to follow the naming convention of other RNN units?

delta2323:
I added comments. Could you apply them? And I also appreciate you if you add unit tests. Thank you in advance!

tohmae:
Thank you for your comment. I want to apply your comments and add unit tests by myself.
Please wait for about 2 weeks.

delta2323:
@tohmae OK. Thank you for your contribution!

tests/chainer_tests/functions_tests/noise_tests/test_zoneout.py
```diff
@@ -0,0 +1,56 @@
...
+        self.h = numpy.random.uniform(-1, 1, (2, 3)).astype(numpy.float32)
+        self.x = numpy.random.uniform(-1, 1, (2, 3)).astype(numpy.float32)
+        self.gy = numpy.random.uniform(-1, 1, (2, 3)).astype(numpy.float32)
+
+    def check_forward(self, h_data, x_data):
+        h = chainer.Variable(h_data)
+        x = chainer.Variable(x_data)
+        h_next = functions.zoneout(h, x, ratio=0)
+        testing.assert_allclose(h_next.data, x.data)
+        h_next = functions.zoneout(h, x, ratio=1.0)
```
delta2323: Could you check the case in which ratio is not 0 nor 1.0?

tohmae:
I apply your comments and add unit tests. But I don't know how I check the case 
in which ratio is not 0 nor 1.0.　And I don't test in GPU enviroment.

tohmae:
I add test case in which ratio is not 0 nor 1 to zoneout function.
But not to StatefulZoneoutLSTM Link.

tohmae:
I apply your comments and add unit tests. I add test case in which ratio is not 0 nor 1.
But I don't test in GPU enviroment.

delta2323:
OK, I will check the GPU test in my environment. If there will be the problem, I will send the PR to your branch.

delta2323:
Thank you for applying my comment.

The previous code checks the output in the special case in which the ratio is either 0 or 1. I think we do not have to remove it.

```
h_next = functions.zoneout(h, x, ratio=self.ratio)
if self.ratio == 0:
    h_next_expect = x
elif self.ratio == 1:
    h_next_expect = h
```

delta2323:
I confirmed GPU tests are passed. BTW, is it possible to remove [WIP] from the title?

tohmae:
Thank you for testing in GPU environment.
I fix implementaion where warning is out in test.

delta2323:
LGMT

delta2323:

PR #1461, opened 2016-08-03
Make not to assign a lambda expression to conform to PEP8

gwtnb:
PEP8 recommends to use `def` instead of binding a lambda expression to an identifier.


delta2323:
LGTM 👍

PR #1463, opened 2016-08-03
Fix the document in sigmoid_cross_entropy

henry0312:
follow the document in softmax_cross_entropy
cf. https://github.com/pfnet/chainer/blob/master/chainer/functions/loss/softmax_cross_entropy.py#L177


henry0312:
_however, I'm not sure this PR is right._

rezoo:
Thank you for pointing it out! I conclude that you are right. I'd like to merge it.

henry0312:
thanks

PR #1466, opened 2016-08-04
Fix UnicodeDecodeError in get_cifar on python3

Fohte:
This PR fixes `UnicodeDecodeError` in `chainer.datasets.get_cifar10` and `chainer.datasets.get_cifar100` when using Python 3.

The following code raises error:
```python
import chainer
chainer.datasets.get_cifar10()
```

And give a trackback:
```
Downloading from https://www.cs.toronto.edu/~kriz/cifar-10-python.tar.gz...
  File "/home/fohte/.pyenv/versions/3.5.2/lib/python3.5/site-packages/chainer/datasets/cifar.py", line 43, in get_cifar10
    raw = _retrieve_cifar('cifar-10')
  File "/home/fohte/.pyenv/versions/3.5.2/lib/python3.5/site-packages/chainer/datasets/cifar.py", line 144, in _retrieve_cifar
    return download.cache_or_load_file(path, creator, numpy.load)
  File "/home/fohte/.pyenv/versions/3.5.2/lib/python3.5/site-packages/chainer/dataset/download.py", line 145, in cache_or_load_file
    content = creator(temp_path)
  File "/home/fohte/.pyenv/versions/3.5.2/lib/python3.5/site-packages/chainer/datasets/cifar.py", line 126, in creator
    d = pickle.load(archive.extractfile(file_name))
UnicodeDecodeError: 'ascii' codec can't decode byte 0x8b in position 6: ordinal not in range(128)
```

### DIFF
--- a/chainer/datasets/cifar.py
+++ b/chainer/datasets/cifar.py
@@ -123,13 +123,13 @@ def _retrieve_cifar(name):
             # training set
             for i in range(5):
                 file_name = '{}/data_batch_{}'.format(dir_name, i + 1)
-                d = pickle.load(archive.extractfile(file_name))
+                d = pickle.load(archive.extractfile(file_name), encoding='latin-1')
                 train_x[i] = d['data']
                 train_y[i] = d['labels']
 
             # test set
             file_name = '{}/test_batch'.format(dir_name)
-            d = pickle.load(archive.extractfile(file_name))
+            d = pickle.load(archive.extractfile(file_name), encoding='latin-1')
             test_x = d['data']
             test_y[...] = d['labels']  # copy to array
 
--- a/chainer/datasets/cifar.py
+++ b/chainer/datasets/cifar.py
@@ -1,4 +1,5 @@
 import os
+import sys
 import tarfile
 
 import numpy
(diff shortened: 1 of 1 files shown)


chainer/datasets/cifar.py
```diff
@@ -123,13 +123,13 @@ def creator(path):
             # training set
             for i in range(5):
                 file_name = '{}/data_batch_{}'.format(dir_name, i + 1)
-                d = pickle.load(archive.extractfile(file_name))
+                d = pickle.load(archive.extractfile(file_name), encoding='latin-1')
```
zori: pickle.load only takes `encoding` keyword in python3. How about defining a function that chooses the right function to call based on python's version:
`def pickle_load(f):
    if sys.version_info > (3, ):
        # python3
        return pickle.load(f, encoding='latin-1')
    else:
        # python2
        return pickle.load(f)`

chainer/datasets/cifar.py
```diff
@@ -123,13 +123,13 @@ def creator(path):
...
                 train_x[i] = d['data']
                 train_y[i] = d['labels']
 
             # test set
             file_name = '{}/test_batch'.format(dir_name)
-            d = pickle.load(archive.extractfile(file_name))
+            d = pickle.load(archive.extractfile(file_name), encoding='latin-1')
```
zori: ditto: call the correct function based on python's version

zori:
Confirmed. The problem comes from the difference in pickling format between python2 and python3 for the CIFAR dataset. The data was serialized in Python2. As a workaround for python3, the 'latin-1' encoding can be given in order to unpickle the data.

Fohte:
Thank you for your review. I fixed it.

zori:
LGTM, thank you for the fix!

PR #1467, opened 2016-08-06
deconvolution size calculation, odd case

cryptexis:
When let's say the width of the input size  is even and kernel is odd during the convolution step we lose from 1-3 pixels, due to integer operation in the division part of convolution size equation:
 (size + p * 2 - k) // s

However during the inverse calculation in deconvolution operation s * (size - 1) + k - 2 * p this loss is not included. 
This issue can be solved by adding k%s term.

### DIFF
--- a/chainer/utils/conv.py
+++ b/chainer/utils/conv.py
@@ -15,11 +15,10 @@ def get_deconv_outsize(size, k, s, p, cover_all=False):
     if cover_all:
         return s * (size - 1) + k - s + 1 - 2 * p
     else:
-        if size%2 == 0 and k > 1 and k%2 == 1:
-            return s * (size - 1) + k - 2 * p + k%s
+        if size % 2 == 0 and k > 1 and k % 2 == 1:
+            return s * (size - 1) + k - 2 * p + k % s
         else:
             return s * (size - 1) + k - 2 * p
-
 
 
 def im2col_cpu(img, kh, kw, sy, sx, ph, pw, pval=0, cover_all=False):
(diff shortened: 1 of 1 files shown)


t-abe:
Thank you for your PR.

I probably understand your idea.
However, I think there is ambiguity in deconvolution output size.

For example, when ``k=5, s=2, p=2`` and the width of input size is 4, it is impossible to determine which 7 or 8 correct output width is.
Because we get size-4 output when we apply ``k=5, s=2, p=2``-convolution on both 7-size and 8-size inputs.

We can specify the output size by ``out_size`` option of Deconvoluition2d.
I think we should use this option in this case.

cryptexis:
thanks for the prompt response.

you're right :) but what if I don't want to specify the output size? What if my network can get any input size and then it should figure out itself what size should be there ?
For example: https://github.com/yusuketomoto/chainer-fast-neuralstyle/blob/master/net.py
Here the paper clearly says that StyleNet should be built on kernel size 3 both for convolution and deconvolution. What should one do there ? StyleNet has the ability to take any size input.

Thanks

t-abe:
> what if I don't want to specify the output size? What if my network can get any input size and then it should figure out itself what size should be there ?

This is not excellent solution.
You can change links.Deconvolution2D's member ``.out_size`` after initialization.
So you can calculate outsize by your get_deconv_outsize from the actual size of the inputl.
Then you can set ``out_size`` before deconvolution is called as follows:

```
def __call__(self, x, ...):
  h = self.conv(x)
  out_h = your_get_deconv_outsize(h.data.shape[2], k=5, s=2, p=2)
  out_w = your_get_deconv_outsize(h.data.shape[3], k=5, s=2, p=2)
  h = self.deconv(h, outsize=(out_w, out_h)
  ...
```

cryptexis:
hehe, thanks will try it and let you know :)

t-abe:
Thank you.
I am going to close this PR.

Issue #1280, opened 2016-06-16
Make Chainer.Function objects correspond to stacking functions.


delta2323:
- [x] `F.stack`
- [ ] `F.vstack`
- [x] `F.hstack`
- [ ] `F.dstack`

We need to create corresponding CuPy functions if necessary.

delta2323:
`F.stack` is already implemented as #1262 and #1276.

niboshi:
I think all of the functions above have been implemented.

PR #1475, opened 2016-08-08
Support hstack function

fukatani:
This PR is for supporting `hstack` mentioned in #1280 .
It is equivalent to hstack method in numpy.

### DIFF
--- a/chainer/functions/array/hstack.py
+++ b/chainer/functions/array/hstack.py
@@ -1,4 +1,5 @@
 import numpy
+import six
 
 from chainer import cuda
 from chainer import function
@@ -13,14 +14,14 @@ class Hstack(function.Function):
         type_check.expect(in_types.size() > 0)
 
         ndim = in_types[0].ndim.eval()
-        for i in range(1, in_types.size().eval()):
+        for i in six.moves.range(1, in_types.size().eval()):
             type_check.expect(
                 in_types[0].dtype == in_types[i].dtype,
                 in_types[0].ndim == in_types[i].ndim,
             )
             if ndim <= 1:
                 continue
-            for d in range(0, ndim):
+            for d in six.moves.range(0, ndim):
                 if d == 1:
                     continue
                 type_check.expect(in_types[0].shape[d] == in_types[i].shape[d])
@@ -30,9 +31,9 @@ class Hstack(function.Function):
         return xp.hstack(xs),
 
     def backward(self, xs, gy):
-        if not xs[:-1]:
+        if len(xs) == 1:
             if xs[0].ndim == 0:
-                return (gy[0].reshape(()), )
+                return (gy[0].reshape(()),)
             return gy
 
         xp = cuda.get_array_module(*xs)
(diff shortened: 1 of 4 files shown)


delta2323:
Thank you for your PR. As you did, creating new Chainer function from existing ones is standard and easy way. But CuPy has `hstack` and `hsplit`, which is the inverse function of `hstack`. So, why don't you use these functions to implement forward and backward methods, respectively? I guess it would be efficient because it directly handles numpy and cupy ndarrys and creates fewer number of Variables.

fukatani:
Thanks for your review! :smiley:
I reimplemented this PR using `hstack` and `hsplit`.

By the way, the same optimization is also effective for `function.stack`?
If so , I would get on it after this PR are merged.

delta2323:
> By the way, the same optimization is also effective for function.stack?

Certainly, it will improve the performance of the function. But I think it is more important for Chainer to implement another stacking function because Chainer has low coverage of primitive functions.

chainer/functions/array/hstack.py
```diff
@@ -1,5 +1,51 @@
...
+                if d == 1:
+                    continue
+                type_check.expect(in_types[0].shape[d] == in_types[i].shape[d])
+
+    def forward(self, xs):
+        xp = cuda.get_array_module(*xs)
+        return xp.hstack(xs),
+
+    def backward(self, xs, gy):
+        if not xs[:-1]:
```
delta2323: Does it check if the length of `xs` is 1 or not? If so, `len(xs) == 1` is simpler.

tests/chainer_tests/functions_tests/array_tests/test_hstack.py
```diff
@@ -61,4 +66,34 @@ def test_backward_gpu(self):
             [cuda.to_gpu(x) for x in self.xs], cuda.to_gpu(self.g))
 
 
+@testing.parameterize(
+    {'a_shape': (3, 4), 'b_shape': (4, 4), 'valid': False},
+    {'a_shape': (3, 4), 'b_shape': (3, 3), 'valid': True},
+    {'a_shape': (3,), 'b_shape': (4,), 'valid': True},
+    {'a_shape': (3), 'b_shape': (3, 3), 'valid': False},
+)
```
delta2323: Could you add the test case where at least one of `a_shape` and `b_shape` are longer than 2?

delta2323:
LGTM except the comments

chainer/functions/array/hstack.py
fukatani: > Certainly, it will improve the performance of the function. But I think it is more important for Chainer to implement another stacking function because Chainer has low coverage of primitive functions.

OK!
I'd give priority to implementation of `vstack` and `dstack`.

> Does it check if the length of xs is 1 or not? If so, len(xs) == 1 is easier to understand.
I did it.

In honest, I quoted it from `function.concat`.
My interpretation of `if not xs[:-1]:` is the same as yours.
So I changed it as you suggested.

tests/chainer_tests/functions_tests/array_tests/test_hstack.py
fukatani: OK!
I added the test for `n_dim` == 3.

delta2323:
LGTM 👍

delta2323:
Thank you very much for your contributon!

PR #1485, opened 2016-08-10
Restrict the cuDNN warning not to print more than once

### LINKED ISSUE #1453: Stop printing cuDNN warning more than once
Users do not need multiple cuDNN warnings. https://github.com/pfnet/chainer/blob/71eef3e90baea61d3140b75741520515f8882f9d/chainer/cuda.py#L86

gwtnb:
Fix #1453

### DIFF
--- a/chainer/cuda.py
+++ b/chainer/cuda.py
@@ -82,11 +82,13 @@ def check_cuda_available():
                '(see https://github.com/pfnet/chainer#installation).')
         msg += str(_resolution_error)
         raise RuntimeError(msg)
-    if not cudnn_enabled:
+    if (not cudnn_enabled and
+            not getattr(check_cuda_available, '_already_warned', False)):
         warnings.warn(
             'cuDNN is not enabled.\n'
             'Please reinstall chainer after you install cudnn\n'
             '(see https://github.com/pfnet/chainer#installation).')
+        check_cuda_available._already_warned = True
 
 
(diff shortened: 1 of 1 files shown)


chainer/cuda.py
```diff
@@ -82,11 +82,13 @@ def check_cuda_available():
                '(see https://github.com/pfnet/chainer#installation).')
         msg += str(_resolution_error)
         raise RuntimeError(msg)
-    if not cudnn_enabled:
+    if (not cudnn_enabled and
+            not getattr(check_cuda_available, '_already_warned', False)):
```
okuta: Is it possible not to use `getattr`?

gwtnb: It is possible. Could you tell me why `getattr` is undesirable?

okuta: `check_cuda_available._already_warned` is simple, isn't it?

gwtnb: To eliminate `getattr`, `check_cuda_available._already_warned = False` should be executed before any call of `check_cuda_available`. Is it simple?

Alternatively, I can use a global variable instead, but it seems unclear for a reader. `_already_warned` should be belong to `check_cuda_available` for readability.

okuta: OK. I understood.

okuta:
LGTM

PR #1496, opened 2016-08-12
Implement hyperbolic cosine and sine functions.

takagi:
This PR implements hyperbolic cosine and sine functions.

Hyperbolic tangent function `tanh` is already supported as an activation function, so I added a link to show that in reference to let users jump.


rezoo:
I think this PR looks good and it is OK to merge this without modification.
If nobody points out some critical problems, I would like to merge.

takagi:
I've fixed the conflict.

rezoo:
Thank you! Then I would like to merge this.

PR #1497, opened 2016-08-12
Fix division of float16

### LINKED ISSUE #1495: Bug in chainer/cupy/core/carray.cuh float16 /=
in the implemention of float16 the operator /= calculates +
 ``` template<typename T>
  inline __device__ float16& operator/=(const T& rhs)
  {
    *this = *this + rhs;
    return *this;
  }```

okuta:
Fix #1495


delta2323:
LGTM 👍

PR #1498, opened 2016-08-12
Fix overflow in reduction kernel

### LINKED ISSUE #1494: Bug in cupy.sum?
I found a case in which the results of cupy.sum largely differ from those of numpy.sum.
Here is an example:

```python
import numpy
from chainer import cuda
import cupy

numpy.random.seed(0)

x_cpu=numpy.random.rand(69,64,2,947).astype(numpy.float32)
sum_cpu=numpy.sum(x_cpu, axis=2)

x_gpu=cupy.asarray(x_cpu)
sum_gpu=cupy.sum(x_gpu, axis=2)

sum_gpu_cpu=cuda.to_cpu(sum_gpu)
diff=numpy.abs(sum_gpu_cpu-sum_cpu)

print(numpy.max(diff))



x_cpu=numpy.random.rand(68,64,2,947).astype(numpy.float32)
sum_cpu=numpy.sum(x_cpu, axis=2)

x_gpu=cupy.asarray(x_cpu)
sum_gpu=cupy.sum(x_gpu, axis=2)

sum_gpu_cpu=cuda.to_cpu(sum_gpu)
diff=numpy.abs(sum_gpu_cpu-sum_cpu)

print(numpy.max(diff))
```

output:
```
3.90482
0.0
```

I found the shape of the input array (x_cpu) matters, but could not find what kind of shapes will cause this strange behavior. Large array size seems to cause the problem?

Chainer version: 1.13.0
Cuda version: 7.5
GPU: GeForce GTX TIT
Linux: Ubuntu 14.04.4 LTS

okuta:
Fix #1494

### DIFF
--- a/tests/cupy_tests/core_tests/test_reduction.py
+++ b/tests/cupy_tests/core_tests/test_reduction.py
@@ -51,7 +51,8 @@ class SimpleReductionFunction(unittest.TestCase):
         self.check_int8_sum((512 + 1, 256 * 256 + 1), axis=1)
 
     def test_shape5(self):
-        size = 2 << 32 // cupy.core.core.simple_reduction_function._block_size
+        size = ((2 << 32)
+                // cupy.core.core.simple_reduction_function._block_size)
         self.check_int8_sum((size, 1), axis=1)
         self.check_int8_sum((size, 1), axis=0)
 
--- a/tests/cupy_tests/core_tests/test_reduction.py
+++ b/tests/cupy_tests/core_tests/test_reduction.py
@@ -50,6 +50,10 @@ class SimpleReductionFunction(unittest.TestCase):
         self.check_int8_sum((512 + 1, 256 * 256 + 1), axis=0)
         self.check_int8_sum((512 + 1, 256 * 256 + 1), axis=1)
 
+    def test_shape5(self):
+        self.check_int8_sum((4 * 1024 * 1024, 1), axis=1)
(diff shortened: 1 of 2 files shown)


rezoo:
I manually confirmed that the current cupy does not pass ``test_shape5``, and this patch correctly fix the bug. LGTM.

rezoo:
Ah I have one comment. In the current implementation ``block_size`` in the reduction kernel is always fixed, i.e., 512.
If you currently consider that this value should not be changed depending on the situation, this PR looks good. Otherwise, I think it is better to explicitly declare ``block_size`` in ``simple_reduction_function`` as a class variable, and reuse it in ``test_shape5``.

tests/cupy_tests/core_tests/test_reduction.py
```diff
@@ -51,8 +51,9 @@ def test_shape4(self):
         self.check_int8_sum((512 + 1, 256 * 256 + 1), axis=1)
 
     def test_shape5(self):
-        self.check_int8_sum((4 * 1024 * 1024, 1), axis=1)
-        self.check_int8_sum((4 * 1024 * 1024, 1), axis=0)
+        size = 2 << 32 // cupy.core.core.simple_reduction_function._block_size
```
zori: Shouldn't it be `(2 << 30) // cupy.core.core.simple_reduction_function._block_size` in order to get the same size?

rezoo:
Great! Then I would like to merge it. Thank you!

PR #1501, opened 2016-08-15
Fix ProgressBar iters per sec

### LINKED ISSUE #1499: ProgressBar iters/sec is too low in CPU mode
The 'iters/sec' speed displayed in the `ProgressBar` extension seems too low when running a model in CPU mode. I suspect this is due to the use of `time.clock()` instead of `time.time()`. `time.clock()` apparently measures the total time spent in all CPU cores, which can cause the measured time to (sometimes greatly) exceed the wall clock time.

bkvogel:
Fix #1499

### DIFF
--- a/chainer/training/extensions/progress_bar.py
+++ b/chainer/training/extensions/progress_bar.py
@@ -61,7 +61,7 @@ class ProgressBar(extension.Extension):
         if iteration % self._update_interval == 0:
             epoch = trainer.updater.epoch_detail
             recent_timing = self._recent_timing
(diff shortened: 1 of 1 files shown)


rezoo:
I agree that using ``time.time()`` is better than ``time.clock()`` because ``iters/sec`` usually represents the number of iterations per second in *real time*.
If nobody takes objection to this opinion, I'd like to merge it.

rezoo:
Good. Then I would like to merge.

PR #1508, opened 2016-08-15
Implement logarithm function to the base two.

takagi:
This PR implements logarithm to the base two, `log2`, function. It might be used to compute negative log-likelihood values in bits.


takagi:
I will add an issue to test the functions for FP16 and FP64.

takagi:
There are following options how to give base of logarithm function:
- accept base as an optional parameter as `math.log`
- provide functions for specific bases as `numpy.log`, `numpy.log2` and `numpy.log10`

Here I choose the latter respect to numpy. If we need arbitrary bases, we can convert a base easily with formula. We are also able to add the former spec later when we would actually need it.

bkvogel:
Your #1519 also seems to include the log2 of this PR. Should that PR replace this one?

bkvogel:
LGTM!

PR #1519, opened 2016-08-18
Implement logarithm function to the base ten, common logarithm.

takagi:
This PR implements logarithm to the base ten, `log10`, function. Rel #1508 .

Now it has only tests for FP32, tests for FP16 and FP64 would be added as another issue.

It also has a bit boring test code with repeated lines, I think of providing abstraction for lifting basic numpy/cupy functions to Chainer functions later. I would be another issue.


takagi:
I've added tests for FP16 and FP64.

bkvogel:
This looks good, but I see that your #1508 also has log2. Do you intend this PR to replace that one?

takagi:
No, it looks so is because log10 branch branches off log2 branch and this PR is compared to master, so the commit list includes log2 branch's commits as well as log10 branch's. Please leave them as they are, merging log2 branch makes all clear.

bkvogel:
I see. Thanks for the clarification.

bkvogel:
LGTM!

takagi:
Thanks!

PR #1528, opened 2016-08-19
Replace zerograd(s) with cleargrad(s)

gwtnb:
`Variable.zerograd` requires allocated memory and fills with zero, but [these operations are not necessary](https://github.com/pfnet/chainer/blob/v1.13.0/chainer/variable.py#L375).

TODOs:
- ~~write tests~~
- write docs (I will send another PR for docs)

### DIFF
--- a/chainer/link.py
+++ b/chainer/link.py
@@ -357,6 +357,16 @@ class Link(object):
         for name in self._params:
             dst[name].copydata(src[name])
 
+    def cleargrads(self):
+        """Clears all gradient arrays.
+
+        This method should be called before the backward computation at every
+        iteration of the optimization.
+
+        """
+        for param in self.params():
+            param.cleargrad()
+
     def zerograds(self):
         """Initializes all gradient arrays by zero.
 
--- a/chainer/link.py
+++ b/chainer/link.py
@@ -370,8 +370,8 @@ class Link(object):
     def zerograds(self):
         """Initializes all gradient arrays by zero.
 
-        This method should be called before the backward computation at every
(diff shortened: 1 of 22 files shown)


chainer/optimizer.py
```diff
@@ -384,8 +384,12 @@ def update(self, lossfun=None, *args, **kwds):
 
         """
         if lossfun is not None:
+            use_cleargrads = kwds.pop('use_cleargrads', False)
```
beam2d: I think it has no problem, but strictly speaking it breaks the compatibility. How do you think about adding a flag to the optimizer object itself?

chainer/link.py
```diff
@@ -357,6 +357,16 @@ def copyparams(self, link):
         for name in self._params:
             dst[name].copydata(src[name])
 
+    def cleargrads(self):
+        """Clears all gradient arrays.
+
+        This method should be called before the backward computation at every
+        iteration of the optimization.
```
beam2d: The note is almost same as that of zerograds, so users may be confused. Would you update the docstring of zerograds? I think new users should use cleargrads in most cases.

beam2d:
LGTM except above comments

chainer/optimizer.py
gwtnb: If the name `use_cleargrads` is already used, this PR breaks backward compatibility. I think it is too pessimistic to consider such case.

beam2d: OK, I agree.

beam2d:
LGTM

PR #1490, opened 2016-08-11
Implement function for reciprocal of square root.

takagi:
This PR implements a function that computes reciprocal of square root.

Currently it is implemented with combining Chainer functions, though it would be more efficient to use CUDA math API `rsqrt` for GPU implementation.

In contrast, numpy seems not to provide such optimized `rsqrt` counterpart.

PR #1534, opened 2016-08-20
Helper for testing Chainer functions lifting unary numpy/cupy functions.

takagi:
Merge #1490 first.

This PR provides a mechanism to easily test Chainer functions lifted from unary numpy/cupy functions such as `exp`, `log` and `tan`. There are now duplicated test codes for that task.

This mechanism is implemented as a decorator that is applied to a test class for Chainer functions. For example, the following code is enough for testing `tan` Chainer function.

```Python
import unittest
from chainer import testing

@testing.unary_function_test(F.tan)
class TestTan(unittest.TestCase):
    pass
```

With decorated, `TestTan` class tests `tan` Chainer function for forward and backward computation on CPU and GPU, parameterized with `dtype` of FP16, FP32 and FP64, and `shape` of rank zero and more. It also tests the label of corresponding Chainer function class `F.Tan`. For detailed specification, see its docstring and the example part would help.

Implementation possibilities
--
I've implemented the same mechanism in the following ways:

- decorator meta programming
- inheritance

This PR is bound to the former, achieving clearer interface but a bit more complicated implementation than the latter. [unary-funciton-test.1](https://github.com/takagi/chainer/tree/unary-function-test.1) branch implements the same thing with inheritance, which might be more Pythonic and better way, where the example above is written as:

```Python
from chainer import testing

class TestTan(testing.UnaryFunctionTestBase):
    def func(self):
        return F.tan
```

Which would be better?


chainer/testing/unary_function_test.py
```diff
@@ -0,0 +1,179 @@
...
+
+       We define ``make_data`` function to return input and gradient ndarrays
+       generated in proper value domains with given ``dtype`` and ``shape``
+       parameters, then passing it to the decorator's ``make_data`` argument.
+
+    """
+
+    # Import here to avoid mutual import.
+    from chainer import gradient_check
+    from chainer import testing
```
delta2323: Is flake8 passed?

chainer/testing/unary_function_test.py
```diff
@@ -0,0 +1,179 @@
...
+
+    if func_expected is None:
+        name = func.__name__
+        func_expected = getattr(numpy, name)
```
delta2323: How about raising error when we cannot find the corresponding function from numpy?

chainer/testing/unary_function_test.py
```diff
@@ -0,0 +1,179 @@
...
+
+        @attr.gpu
+        @condition.retry(3)
+        def test_backward_gpu(self):
+            self.check_backward(cuda.to_gpu(self.x), cuda.to_gpu(self.gy))
+        setattr(klass, "test_backward_gpu", test_backward_gpu)
+
+        def test_label(self):
+            klass = func_class(func)
+            self.assertEqual(klass().label, func.__name__)
```
delta2323: We can overwrite label name in subclass of chainer Functions. So, we should some way for developers to feed expected label name.
\

chainer/testing/unary_function_test.py
```diff
@@ -0,0 +1,179 @@
...
+    from chainer import testing
+
+    if func_expected is None:
+        name = func.__name__
+        func_expected = getattr(numpy, name)
+
+    if make_data is None:
+        make_data = make_data_default
+
+    def f(klass):
```
delta2323: Please decorate this function with [functools.wraps](https://docs.python.org/2.7/library/functools.html#functools.wraps). Otherwise, `func_name` of this class would be always `f`, regardless of the value of `klass`.

delta2323:
It is a bit strange to me that the decorated class is always empty and filled with pass. How about make `unary_function_test` a function that generates the subclass of `unittest.TestCase` ? 

```
TestTan = testing.unary_function_test(F.tan)
```

Actually, I did similar thing in #1395 in which I generate Stateful RNN classes from Stateless ones. [Here](https://github.com/pfnet/chainer/pull/1395/files#diff-df1f7648760de0741daf30fd78bf9644R14) is the code.

chainer/testing/unary_function_test.py
takagi: Now I resolved conflict, will check if flake8 would pass.

chainer/testing/unary_function_test.py
takagi: `getattr` raises `AttributeError` if it cannot find a function with given `name` in numpy, you mean wrapping it with another exception under this context?

chainer/testing/unary_function_test.py
takagi: You mean like following case?

```python
class MySqrt(Sqrt):

    @property
    def label(self):
        return 'name_other_than_mysqrt'
    ...

def mysqrt(x):
    return MySqrt()(x)
```

chainer/testing/unary_function_test.py
takagi: Is this needed for class decorators as well as ones for functions? https://docs.python.org/2.7/library/functools.html#functools.update_wrapper

takagi:
Pros of the decorated style in declarative manner is that we can imply readers the main responsibility of the code part is defining some test class, without making them refer its documentation or dive into its implementation. In contrast, generating subclasses with functions would leave users wider possibilities to guess what the function actually does. So I prefer the former style here.

Of course generating subclasses via function has logically the same semantics and is one of reasonable options, I'm willing to fix as you suggest if you would like it. How about that?

chainer/testing/unary_function_test.py
takagi: It passes flake8. https://travis-ci.org/pfnet/chainer/builds/154988181

chainer/testing/unary_function_test.py
delta2323: Yes, it is better to add explanatory messages to the error.

chainer/testing/unary_function_test.py
delta2323: Yes, actually, some existing Chainer functions overwrites `label` property (e.g. [neg](https://github.com/pfnet/chainer/blob/3f546e8bb63738b760d88b4f9666b1aaccfb73f0/chainer/functions/math/basic_math.py#L52)).

chainer/testing/unary_function_test.py
takagi: Sure.

chainer/testing/unary_function_test.py
takagi: Ah~ okay!

chainer/testing/unary_function_test.py
delta2323: Sorry, it was my mistake.

delta2323:
After the offline discussion, we decided to change the decorator into a function that generates unit tests.

takagi:
As you said, there is a problem to make unittest reconize test cases.

I find that a generated test case, which is to be parameterized, needs to belong to caller test module, so we have to pass it explicitly as an argument of `unary_function_test` or to get it implicitly somehow, maybe inspecting caller function but it would not be recommended approach.

https://github.com/pfnet/chainer/compare/master...takagi:unary-function-test.2#diff-c446668fa92d10c8a25e1c50b42e2c6aR182

Any idea?

delta2323:
I see. I am afraid I cannot come up with better ideas. I think inspect is better idea because it makes API exposed to users simpler.

takagi:
Okay. A remaining thing I concern on inspection is that it forces `unary_function_test` function to be called directly in test module where test classes are defined, it might not be easy to expect for its users, so at least I will leave a note about it on its docstring.

chainer/testing/unary_function_test.py
takagi: I fixed to provide ValueError with explanatory message.

chainer/testing/unary_function_test.py
takagi: I added `label_expected` parameter keyword.

chainer/testing/unary_function_test.py
takagi: No problem.

takagi:
I keep decorator approach as we discussed today's meeting, would you review it again? I added `label_expected` keyword parameter with some small fixes, and renamed the decorator's name from `unary_function_test` to `math_function_test` because it is valid only for mathematical functions, not for tensor manipulation functions even if they are 'unary'.

chainer/testing/math_function_test.py
```diff
@@ -61,48 +68,59 @@ def unary_function_test(func, func_expected=None, make_data=None):
...
+       >>> from chainer import testing
+       >>> from chainer import functions as F
        >>>
        >>> def rsqrt(x, dtype=numpy.float32):
        >>>     return numpy.reciprocal(numpy.sqrt(x, dtype=dtype))
        >>>
-       >>> @unary_function_test(F.rsqrt, rsqrt)
+       >>> @testing.math_function_test(F.rsqrt, func_expected=rsqrt)
        >>> class TestRsqrt(unittest.TestCase):
        >>>     pass
```
delta2323: Can we use doctest to check the code here?

tests/chainer_tests/testing_tests/test_math_function_test.py
```diff
@@ -0,0 +1,22 @@
...
+    def test_no_numpy_function(self):
+        with self.assertRaises(ValueError):
+            testing.math_function_test(F.rsqrt)  # no numpy.rsqrt
+
+
+class TestInvalidExpectedLabel(unittest.TestCase):
+
+    def test_invalid_expected_label(self):
+        with self.assertRaises(ValueError):
+            # no numpy.rsqrt
```
delta2323: Should we use existing numpy function to test the behavior of `label_expected`?

delta2323:
Can we use apply this decorator to mathematical operation which have more than one operands?

delta2323:
Float16 test of rsqrt is unstable (see this [failure](https://travis-ci.org/pfnet/chainer/jobs/158062194)). May be inputs more bigger values will help to make the test stable.

chainer/testing/math_function_test.py
```diff
@@ -0,0 +1,207 @@
...
+import chainer
+from chainer import cuda
+import chainer.functions as F
+from chainer.testing import attr
+from chainer.testing import condition
+
+
+def func_class(func):
+    name = func.__name__
+    name = name[0].upper() + name[1:]
```
delta2323: Use [string.capitalize](http://docs.python.jp/2/library/string.html#string.capitalize)

chainer/testing/math_function_test.py
```diff
@@ -0,0 +1,207 @@
...
+                       make_data=None):
+    """Decorator to test Chainer functions lifting mathematical numpy functions.
+
+    This decorator is for testing Chainer functions lifted from corresponding
+    mathematical numpy and cupy functions, and optionally ones composed with
+    such other Chainer functions. Forward and backward computations on CPU and
+    GPU across parameterized ``dtype`` and ``shape`` are tested.
+
+    Args:
+        func(~chainer.Function): Chainer function to be tested by
```
delta2323: According to the code that uses this decorator [here](https://github.com/pfnet/chainer/pull/1534/files?diff=unified#diff-0b1a99e803a1ab8f3e3e660fab939647R18), it must be a python function that has corresponding Chainer function, rather than the Chainer function itself.

delta2323: Alternatively, directory pass Chainer Functions to this decorator as `func` argument and do not use `func_class`.

chainer/testing/math_function_test.py
```diff
@@ -0,0 +1,207 @@
...
+            func_expected = getattr(numpy, name)
+        except AttributeError:
+            raise ValueError("numpy has no function corresponding "
+                             "to Chainer function '{}'.".format(name))
+
+    if label_expected is None:
+        label_expected = func.__name__
+    elif func_class(func) is None:
+        raise ValueError('Expected label is given even though Chainer '
+                         'function does not have its label.')
```
delta2323: If we feed Chainer function directly to `func`, this error checking is no longer needed.

chainer/testing/math_function_test.py
```diff
@@ -0,0 +1,207 @@
...
+       returns another with the same ``dtype``.
+
+       >>> import unittest
+       >>> from chainer import testing
+       >>> from chainer import functions as F
+       >>>
+       >>> @testing.math_function_test(F.sin)
+       >>> class TestSin(unittest.TestCase):
+       >>>     pass
+
```
delta2323: If possible, it is better to use doctest to check the code is appropriate.

takagi:
> Can we use apply this decorator to mathematical operation which have more than one operands?

No. I actually thought of it. Assuming the following classes of mathematical operations:

1. Math operations with one parameter.
2. Math operations with one parameter and one or more hyper parameter, such as `clip` that has `x_min` and `x_max` arguments that specify clip boundary.
3. Math operations with more than one parameters.

Since most mathematical operations in Chainer, at least currently, are those of a. and few are of b. or c., I leave the decorator as it is for now.

chainer/testing/math_function_test.py
takagi: What I intend is to make it possible to test functions such as `F.rsqrt` that has no `chainer.Function` subclass, composed with other Chainer math functions. How do you think about this idea?

    def rsqrt(x):
        return 1.0 / sqrt(x)

chainer/testing/math_function_test.py
takagi: As above, I intended to make it possible to test functions that have no `chainer.Function` subclass. Here I mean:

- If label_expected is not provided, label test may be skipped if `func` has no corresponding `chainer.Function` subclass.
- If label_expected is provided, it means that we want to test label so we need `func` to have corresponding `chainer.Function` subclass, so this error checking works.

chainer/testing/math_function_test.py
takagi: I fixed it.

chainer/testing/math_function_test.py
takagi: Or defining `rsqrt` function with `Rsqrt` function class would be better?

takagi:
> Float16 test of rsqrt is unstable (see this failure). May be inputs more bigger values will help to make the test stable.

I fixed as you suggested, it passed for tens of times on my local CPU environment. I will check it on Travis.

takagi:
> Float16 test of rsqrt is unstable (see this failure). May be inputs more bigger values will help to make the test stable.

I fixed as you suggested, it passed for tens of times on my local CPU environment. I will check it on Travis.

tests/chainer_tests/testing_tests/test_math_function_test.py
takagi: Yes for testing invalid expected label with actual label, but what I intended here was giving `label_expected` for testing a function that does not have corresponding `chainer.Function` subclass, so the comment was less appropriate. I fixed it.

Relating the discussion below.
> According to the code that uses this decorator here, it must be a python function that has corresponding Chainer function, rather than the Chainer function itself.
> Alternatively, directory pass Chainer Functions to this decorator as func argument and do not use func_class (I prefer it to the above change).

chainer/testing/math_function_test.py
takagi: I tried to make it tested in `doctest` directive. I found that the following doctest tested parameterized conterparts of `TestSin` test class, but including GPU tests:

```
.. doctest::

   >>> import unittest
   >>> from chainer import testing
   >>> from chainer import functions as F
   >>> import nose
   >>>
   >>> @testing.math_function_test(F.sin)
   ... class TestSin(unittest.TestCase):
   ...     pass
   ...
   >>> nose.main(defaultTest=__name__, argv=[''], exit=False).sucess
   True
```

And the following modification, adding `"--attr='!gpu'"` to `nose.main` 's `argv` parameter, did not work, testing no tests with just returning `True`.

```
.. doctest::

   >>> import unittest
   >>> from chainer import testing
   >>> from chainer import functions as F
   >>> import nose
   >>>
   >>> @testing.math_function_test(F.sin)
   ... class TestSin(unittest.TestCase):
   ...     pass
   ...
   >>> nose.main(defaultTest=__name__, argv=['', "--attr='!gpu'"], exit=False).sucess
   True
```

I'm not sure why it does not work with refering `nose` 's document, need more time to dive into its implementation. Let me talk to you about it.

(`nose` seems to be basically designed to be used in command line, not well in scripting maybe because of repeated update. It often went not as I expected, and required me hours of study to get here.)

chainer/testing/math_function_test.py
takagi: Please see another comment about doctest below.

takagi:
@delta2323 Please let me talk to you about this PR in the next 集中開発 because its comments become a bit complicated, you can leave them until then.

takagi:
Note:

There would be the following possible classes for Chainer functions that could be supported by `math_function_test`:

- Basic one input and one output mathematical functions.
- Composed with other Chainer functions, no `chainer.Function` subclasses: `rsqrt`
- Multiple input parameters.
- Multiple outputs.
- With hyper parameters: `clip`

And these are possible goals of providing it:

- Removing `UnaryFunctionsTest` boiler plates for at least existing Chainer functions, hoping possible classes above are already covered enough.
- Broader generalization for testing facility for mathematical Chainer functions.

chainer/testing/math_function_test.py
takagi: I solved the problem and now doctests in the docstring work!

takagi:
I've fixed the codes following today's discussion, please review it again.

chainer/testing/unary_math_function_test.py
```diff
@@ -0,0 +1,184 @@
...
+    - shape: rank of zero, and rank of more than zero
+    - dtype: ``numpy.float16``, ``numpy.float32`` and ``numpy.float64``
+
+    Additionally, it tests the label of the Chainer function.
+
+    Chainer functions tested by the test class decorated with the decorator
+    should have the following properties:
+
+    - Unary, taking one parameter and returning one value
+    - ``dtype`` of input and output are the same
```
delta2323: Is it OK that the dtype is one of the integer types?

chainer/testing/unary_math_function_test.py
```diff
@@ -0,0 +1,184 @@
...
+    # Import here to avoid mutual import.
+    from chainer import gradient_check
+    from chainer import testing
+
+    if func_expected is None:
+        name = func.__class__.__name__.lower()
+        try:
+            func_expected = getattr(numpy, name)
+        except AttributeError:
+            raise ValueError("numpy has no function corresponding "
```
delta2323: NumPy

delta2323: no function*s*

tests/chainer_tests/functions_tests/math_tests/test_sqrt.py
```diff
@@ -2,115 +2,36 @@
...
-from chainer import cuda
 import chainer.functions as F
-from chainer import gradient_check
 from chainer import testing
-from chainer.testing import attr
-from chainer.testing import condition
 
 
-class UnaryFunctionsTestBase(unittest.TestCase):
+#
```
delta2323: Do we need this empty comment?

tests/chainer_tests/functions_tests/math_tests/test_sqrt.py
```diff
@@ -2,115 +2,36 @@
...
+@testing.unary_math_function_test(F.Sqrt(), make_data=make_data)
+class TestSqrt(unittest.TestCase):
+    pass
 
-    def check_forward_cpu(self, op, op_xp):
-        self.check_forward(op, op_xp, self.x)
 
-    def check_forward_gpu(self, op, op_xp):
-        self.check_forward(op, op_xp, cuda.to_gpu(self.x))
+#
```
delta2323: Do we need this empty comment?

chainer/testing/unary_math_function_test.py
```diff
@@ -0,0 +1,184 @@
...
+            in the tests. It takes ``shape`` and ``dtype`` as its arguments,
+            and returns a tuple of input and gradient data. By default, uniform
+            destribution ranged ``[-1, 1]`` is used for both.
+
+    The decorated test class tests forward and backward computations on CPU and
+    GPU across the following :func:`~chainer.testing.parameterize` ed
+    parameters:
+
+    - shape: rank of zero, and rank of more than zero
+    - dtype: ``numpy.float16``, ``numpy.float32`` and ``numpy.float64``
```
delta2323: Do you think it is OK to check the function with `numpy.integer16`, 32 and 64? Some functions like those in `basic_math` accept integer types as well.

chainer/testing/unary_math_function_test.py
takagi: Fixed.

chainer/testing/unary_math_function_test.py
takagi: Yes, it would be. As discussed, it is supposed to be enough to support testing integer types when it would be actually needed.

chainer/testing/unary_math_function_test.py
takagi: The same as the reply above.

tests/chainer_tests/functions_tests/math_tests/test_sqrt.py
takagi: I removed it.

chainer/testing/unary_math_function_test.py
takagi: I noted about the future extension on Chainer functions that could be tested with the decorator in the comment.

delta2323:
LGTM!

Issue #1448, opened 2016-08-01
Want contribution for providing more primitive operations.


takagi:
## Background

It would be helpful that Chainer had more primitive operations, especially driving novel directions of deep learning research. Providing flexibility in such is one of Chainers most important objectives.
## Goal

Because we have quite restricted development resource, we would appreciate if we had Chainer community's precious contribution to provide more primitive operations.
## What primitive operations Chainer does not support now?

Here shows a list of primitive operations that Chainer does not provide currently compared with other major deep learning frameworks.

https://docs.google.com/spreadsheets/d/1tHQcuKhqYiDUo57uOWOtW_mUqrHmsP9fZzaQRA2dW5k/edit?usp=sharing
## How to implement a primitive operation?

**Sample works**
There are some PRs that implements primitive operations and you can follow with.
- #1007 clip function
- #1262 stack function
- #1401 rollaxis function

**Overall steps**
The following are overall steps to implement a Chainer function.
1. Keep a WIP pull request
   - Keep a work-in-progress pull request to show you are on work for a primitive operations.
   - You can make an empty commit with `git --allow-empty -m "Empty commit to keep PR."` to send an empty pull request to begin with.
2. Implement and test a Chainer function.
   - Implement a Chainer function and its test. You need to implement forward computation and backward computation for CPU and GPU respectively.
   - For detail on how to write a Chainer function, please see the following tutorial.
     http://docs.chainer.org/en/stable/tutorial/function.html
3. Document a Chainer function.
   - Write a documentation for your Chainer function and add it to sphinx source file.
4. Take a review.
   - Ask Chainer team to review your PR. After your PR passes the review, it is merged into Chainer's master branch.

entron:
I am thinking it would be good to make array related operations the same as numpy if possible. This will make chainer easier to learn and port.

takagi:
Yes, it would be good.

takagi:
I added Priority column to the list.

aksub99:
I'd like to implement ELiSH activation function as introduced in this paper.
https://arxiv.org/abs/1808.00783

aksub99:
I've opened issue #6376  to get some suggestions on the operation I can start implementing. Please let me know if any of the mentioned functions could be useful additions at present. Thanks in advance!

UmashankarTriforce:
[...]

PR #1537, opened 2016-08-21
Support square function.

fukatani:
This PR is for supporting `functions.square`. ( #1448 )
But I'm wondering how to avoid relative import `UnaryFunctionsTestBase`.

How about move it to `chainer.testing`? Or not using it?


takagi:
> How about move it to chainer.testing? Or not using it?

I sent PR #1534 which provides a helper for testing such functions, you may use it after it will be merged. Before that, I suppose that duplicated base class would be acceptable.

fukatani:
OK !
I'd wait for merger and reimplement test using helper class.
There are many duplicated codes in test, helper class seems useful!

delta2323:
So it is better to check #1534 with higher priority, isn't it?

takagi:
Yes, it would be better.

delta2323:
got it.

chainer/functions/math/square.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def square(x):
+    """Elementwise square function.
+
+    .. math::
+       y_i = x_i ^ 2.
+
+    Args:
+        x (~chainer.Variable): Input variable.
```
delta2323: Now, Chainer functions can accept numpy and cupy ndarrays as well (c.f. #1221). So could you change the acceptable types to that (c.f. #1687 )?

```
(chainer.Variable or :class:`numpy.ndarray` or cupy.ndarray)
```

fukatani: Sure! I fixed doc.

delta2323:
Now #1534 is merged, we can use unary function testing tools and proceed this PR.

takagi:
`unary_math_function_test` provides the default input data in range `[-1, 1]` and it is enough for `square` function that accepts real numbers including negative numbers. So you can remove `make_data` optional parameter to use the default data provider:

```Python
@testing.unary_math_function_test(F.Square())
class TestSquare(unittest.TestCase):
    pass
```

The custom `make_data` function is needed for `sqrt` function because it does not accept negative numbers.

fukatani:
Sure. I fixed it.

delta2323:
https://github.com/pfnet/chainer/issues/1740

fukatani:
I got it! No problem :smiley:

delta2323:
Now `unary_math_function_test` is renamed to `unary_math_function_unittest` by #1741. Could you fix decorator's name and merge the latest master branch?

fukatani:
I fixed it! :smiley:

chainer/functions/math/square.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def square(x):
+    """Elementwise square function.
+
+    .. math::
+       y_i = x_i ^ 2.
+
+    Args:
+        x (~chainer.Variable): Input variable., :class:`numpy.ndarray` or
```
delta2323: Description seems corrupted.

fukatani:
Thanks!
I'm sorry for silly mistake. :cry:

delta2323:
LGTM 👍

PR #1538, opened 2016-08-22
Fix document for softmax_cross_entropy

gwtnb:


chainer/functions/loss/softmax_cross_entropy.py
```diff
@@ -154,10 +154,9 @@ def softmax_cross_entropy(
             number of dimensions is greater than 2.
         t (Variable): Variable holding an int32 vector of ground truth labels.
             If ``t[i] == -1``, corresponding ``x[i]`` is ignored.
-        normalize (Variable): Variable holding a boolean value which
-            determines the normalization constant. If true, this function
-            normalizes the cross entropy loss across all instances. If else,
-            it only normalizes along a batch size.
+        normalize (bool): If true, this function normalizes the cross entropy
```
zori: Actually, isn't it that if ``True``, it normalizes w.r.t. non-ignored labels in the batch, and otherwise, w.r.t. all elements in the batch? https://github.com/pfnet/chainer/blob/3ca579ee17cb2ef328f3a8a33473d3c9b690db41/chainer/functions/loss/softmax_cross_entropy.py#L60-L63

gwtnb: Thank you for reviewing. I will fix it.

gwtnb:
@zori I discussed with @beam2d and we agreed that we should modify the behavior of `normalize` argument to accept a meaningful value (e.g. `normalize='excluding_batches_with_ignored_label'`). Although description of `normalize` in this PR is inaccurate, this PR should be merged without modification to reduce documentation work. This PR corrects the type of `normalize`. We will write accurate document when modifying the behavior of `normalize`.

zori:
LGTM, this fixes the type of the `normalize` argument in the description.

PR #1551, opened 2016-08-24
Use shape placeholder in example

okuta:
Update examples and documentation to use shape placeholder.


delta2323:
> This PR still includes bug.

Then, could you add [WIP] to the title?

okuta:
I fixed bugs. Please review this PR.

gwtnb:
`train_mnist_data_parallel.py` failed.

zori:
I created a PR that fixes `train_mnist_data_parallel.py`. I have tested the code with and without parameter shape placeholder. My proposed fix is https://github.com/pfnet/chainer/pull/1576

okuta:
Thank you @gwtnb and @zori .
I merged and fixed.

chainer/training/updater.py
```diff
@@ -307,6 +308,10 @@ def update_core(self):
                 in_vars = variable.Variable(in_arrays)
                 losses.append(loss_func(in_vars))
 
+        # For _uninitialized_params
```
zori: I think this call to cleargrads(), just before the backward pass, is enough for all parameters. There is no need to clear them here: https://github.com/pfnet/chainer/blob/0ceb2b20d1cd7b16b90ceb8e78d55774e77777cb/chainer/training/updater.py#L292-L293 before the forward pass.

zori:
I updated the tutorial https://github.com/pfnet/chainer/pull/1584

chainer/training/updater.py
okuta: `add_param` fill `grad` with `nan`.
https://github.com/pfnet/chainer/blob/master/chainer/link.py#L141

disktnk: Why `clear_grads` is set before call loss_func?
https://github.com/pfnet/chainer/pull/1551/files#diff-32a6fe336029b64713fe8980fe17f085R293

bkvogel:
LGTM!

PR #1556, opened 2016-08-26
Support N-dimensional convolution function and link

takagi:
This PR implements N-dimensional convolution function and link, following #1279 which is reverted by #1542, and now support cuDNN 5.x.

In this PR, it is to support FP16, FP32 and FP64 of storing precision, and FP32 and FP64 of computing precision. FP16 computing precision will be supported through another issue, since we would require policy making on how to treat FP16 computing precision on GPUs with compute capability that does not support it.

### DIFF
--- a/chainer/functions/connection/convolution_nd.py
+++ b/chainer/functions/connection/convolution_nd.py
@@ -144,9 +144,15 @@ class ConvolutionND(function.Function):
         # Add bias if given.
         # TODO(takagi) Support unshared bias
         if b is not None:
-            cudnn.add_tensor(
-                handle, one.data, self.bias_desc.value, b.data.ptr,
-                one.data, y_desc.value, y.data.ptr)
+            if _cudnn_version >= 3000 or ndim == 2:
+                cudnn.add_tensor(
+                    handle, one.data, self.bias_desc.value, b.data.ptr,
+                    one.data, y_desc.value, y.data.ptr)
+            else:
+                # cuDNN v2 seems not to support bias addition in spatial
+                # dimensions of three or more.
+                b_index = (None, colon) + (None,) * ndim
+                y += b[b_index]
 
         return y,
 
--- a/chainer/functions/connection/convolution_nd.py
+++ b/chainer/functions/connection/convolution_nd.py
@@ -150,7 +150,7 @@ class ConvolutionND(function.Function):
                     one.data, y_desc.value, y.data.ptr)
             else:
                 # cuDNN v2 seems not to support bias addition in spatial
-                # dimensions of three or more.
+                # dimensions other than two.
                 b_index = (None, colon) + (None,) * ndim
                 y += b[b_index]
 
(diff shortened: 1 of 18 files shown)


takagi:
With cuDNN v2, `cudnnAddTensor_v2` seems not to work with spatial dimensions of three or more, returns `CUDNN_STATUS_NOT_SUPPORTED`. See jenkins py2 #1573 . I will fix it to use cupy implementation for bias addition in the case.

takagi:
I fixed as following on using cuDNN:

- Use `cudnn.setConvolutionNdDescriptor_v3` in cuDNN v3 or higher since cuDNN v5 no longer has `cudnn.setConvolutionNdDescriptor_v2`.
- Use cupy implementation for bias forward and backward computation in cuDNN v2 since it seems not to support them for spatial dimensions of three or more.
- Use computing precision of FP32 for storing precision of FP16 for `cudnn.setConvolutionNdDescriptor_v3` for now.

Now Jenkins passes for this PR and we would be able to release it in the next release v1.15.0 after passing review again, except supporting computing precision of FP16. Computing precision of FP16 will be supported in #1558.

takagi:
Note: `cudnn.setConvolutionNdDescriptor_v3` works as strange as that it returns `CUDNN_STATUS_ARCH_MISMATCH` for computing precision FP16 for spatial dimensions of tree, but does not for spatial dimensions of two.

takagi:
@delta2323 Please review it again, I will share you its detail at today's meeting.

chainer/functions/connection/convolution_nd.py
```diff
@@ -0,0 +1,350 @@
...
+
+        # Add bias if given.
+        # TODO(takagi) Support unshared bias
+        if b is not None:
+            if _cudnn_version >= 3000 or ndim == 2:
+                cudnn.add_tensor(
+                    handle, one.data, self.bias_desc.value, b.data.ptr,
+                    one.data, y_desc.value, y.data.ptr)
+            else:
+                # cuDNN v2 seems not to support bias addition in spatial
```
delta2323: does not seem to

cupy/cudnn.py
```diff
@@ -114,8 +129,19 @@ def create_convolution_descriptor(pad, stride,
         c_pad = _to_ctypes_array(pad)
         c_stride = _to_ctypes_array(stride)
         c_upscale = _to_ctypes_array((1,) * ndim)
-        cudnn.setConvolutionNdDescriptor_v2(
-            desc.value, ndim, c_pad.data, c_stride.data, c_upscale.data, mode)
+        if _cudnn_version >= 3000:
+            data_type = get_data_type(dtype)
+            # TODO(takag) Temporarily use computing precision of FP32 for
```
delta2323: (takagi)

chainer/functions/connection/convolution_nd.py
```diff
@@ -0,0 +1,350 @@
...
+        if b is None:
+            return gx, gW
+        elif _cudnn_version >= 3000 or self.ndim == 2:
+            gb = cuda.cupy.empty_like(b)
+            libcudnn.convolutionBackwardBias(
+                handle, one.data, gy_desc.value, gy.data.ptr,
+                zero.data, self.bias_desc.value, gb.data.ptr)
+            return gx, gW, gb
+        else:
+            # cuDNN v2 seems not to support bias backward in spatial dimensions
```
delta2323: does not seem to

delta2323:
LGTM except the comments.

takagi:
I fixed the comments, please review again!

delta2323:
LGTM 👍

Issue #1549, opened 2016-08-23
Imagenet training example missing learning rate decay?


yanii:
As far as I can tell there doesn't seem to be any change in learning rate for the imagenet example, or for that matter in any of the examples. Am I missing something?

Before the Trainer abstraction (c67fa7869071c6470fd666be4be7f3c021d507d2), this line in the training loop decayed learning rate:
`optimizer.lr *= 0.97`

From what I can tell, the ExponentialShift extension should be used in the Trainer abstraction to perform a similar function, but there is none used. In fact there are no examples whatsoever on how to decrease learning rate during training that I can find. It looks like this might be correct however:

`trainer.extend(extensions.ExponentialShift('lr', 0.97), trigger=(1, epoch))`

How to print the current learning rate with a reporter is something I can't figure out within this framework. I suppose you could put a reporter in the ExponentialShift/Optimizer?

I'm personally not a fan of the abstraction to begin with, but if you are going to add this stuff, the examples should be fully fleshed out and easy to understand for newcomers. I've resorted to going back in the history and using the old examples more than a few times now, they were far clearer and more flexible (which frankly is what we need for research purposes - not an inordinate number of layers of abstraction). The harsh truth is that if I was coming across chainer for the first time now, with only these examples to look at, I'd probably have never ended up using it.

delta2323:
Thank you for your comments and sorry for our late response.

As you pointed out, we accidentally dropped the learning rate decay from the imagenet example. As this is a bug of the example, we should restore it.

As for the reporting of the learning rate, we can use `observe_lr` extension, which is implemented in PR #1472. You can check its typical usage in `examples/imagenet/train_imagenet.py`.

beam2d:
Thank you for giving us your opinion on the current direction towards training loop abstraction. Our motivation is to avoid boilerplate coding on writing training loop and support frequently-used but heavy-to-support features around managing the training process. We really do not want to hinder users' attempts to write their own training loop, but I believe that at least for beginners it would be better to use the prepared training loop instead of their own, because it makes them much easier to avoid a bug and can easily support frequently-used features (the most important one that I expected is suspend/resume of the training process with properly managing all stuffs involved in the training process and its monitoring).

[...]

PR #1563, opened 2016-08-29
Add weight decay into imagenet example

t-abe:
#1549


rezoo:
Is this modification an "annealing" rather than "weight decay"? (I do not know the strict name of the above modification...)
I think the implementation of weight decay would be the following:
```python
optimizer.add_hook(chainer.optimizer.WeightDecay(weight_decay_rate), 'hook_optimizer')
```

PR #1570, opened 2016-08-30
Modify GradientMethod to use a flag of an object for cleargrads.

gwtnb:
This is useful than passing an argument of the update method.

### DIFF
--- a/chainer/optimizer.py
+++ b/chainer/optimizer.py
@@ -446,7 +446,13 @@ class GradientMethod(Optimizer):
 
         Args:
             use (bool): If true, this function enables use of `cleargrads`.
-                If false, disables use of `cleargrads`.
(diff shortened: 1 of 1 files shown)


chainer/optimizer.py
```diff
@@ -437,6 +441,16 @@ def update_one_gpu(self, param, state):
         """
         raise NotImplementedError
 
+    def use_cleargrads(self, use=True):
+        """Enables or disables use of :func:`~chainer.Link.cleargrads` in `update`.
+
+        Args:
+            use (bool): If true, this function enables use of `cleargrads`.
+                If false, disables use of `cleargrads`.
```
beam2d: It would be better to note that zerograds is used by default and in the case of `use=False`.

beam2d:
LGTM except one comment

beam2d:
Thank you!

PR #1572, opened 2016-08-30
Fixed to exit by once ctrl+c

okuta:


gwtnb:
How does this PR work? AFAIK, C-c sends `SIGINT` to the foreground process group, so this PR is unnecessary if code written in C :)

gwtnb:
This PR discards children's stack traces, so cannot be merged. Let's press C-c multiple times!

okuta:
Thank you for the opinion of another point of view.
I will press C-c multiple times.

PR #1576, opened 2016-08-31
This fixes the 'train_mnist_data_parallel.py' example when using shape placeholder in MLP.

zori:
Use deep copy for slave models in ParallelUpdater. Zero parameter gradients just before backward computation (to allow for parameters initialized on first call).

In ParallelUpdater, each of the main model and the slave models needed to have its own version of `_uninitialized_params`. In the previous code that was a set shared between the models (because of the shallow copy). When one model has the delayed parameters initialized at call time, the set became empty and no other models could have their parameters initialized. So I used deep copy. 

The change also means that the parameters initialized at call time (using parameter shape place holder) will not be the same across the main and the slave models. 

Additionally, the gradients are zeroed only once all the parameters have been initialized.

### DIFF
--- a/chainer/training/updater.py
+++ b/chainer/training/updater.py
@@ -1,3 +1,4 @@
+import copy
 import six
 
 from chainer.dataset import convert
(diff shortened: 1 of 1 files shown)


okuta:
Thank you for your PR!

PR #1577, opened 2016-08-31
Take a snapshot only at the end of the learning in MNIST example

beam2d:
This change decreases the number of snapshots taken in the MNIST example. It makes the example quickly finish.


gwtnb:
LGTM

PR #1580, opened 2016-09-01
A bug-fix in BatchNormalization both test and finetune are True.

motootom:
The population statistics should not be updated while testing even if in
finetune mode.
Changed to use pre-computed population statistics when test and finetune
are both True.

### DIFF
--- a/chainer/links/normalization/batch_normalization.py
+++ b/chainer/links/normalization/batch_normalization.py
@@ -95,9 +95,9 @@ class BatchNormalization(link.Link):
                 for normalization, and normalizes the input using batch
                 statistics.
 
-        If ``test`` is ``False``, then BatchNormalization runs in training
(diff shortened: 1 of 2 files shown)


bkvogel:
Thank you for the bugfix PR! It looks like the Travis build failed, maybe due to flake8 errors. Could you fix the errors? See

http://docs.chainer.org/en/stable/contribution.html

Thanks.

motootom:
I fixed the flake8 error. But Travis CI failed in the linux 3.5.1 test. It
seems to be an installation failure on Travis CI.
Please check it.

2016年9月1日(木) 17:05 Brian Vogel <notifications@github.com>:

bkvogel:
Thank you for the fix. I re-ran the Travis CI and it passed this time. LGTM!

PR #1604, opened 2016-09-05
Implement square function

tek1031:
This PR implements square function with covering test for input of dtype float16, float32 and float64.
The way to test this function is almost same as sqrt function. (see tests/chainer_tests/functions_tests/math_tests/test_sqrt.py)


gwtnb:
Thank you for sending the PR but it duplicates. https://github.com/pfnet/chainer/pull/1537

tek1031:
Thank you for checking my pull request.
I understand this PR is duplicated and close this.

PR #1611, opened 2016-09-07
Fix cupy.clip

asi1024:
cupy.clip returns different values (compared to numpy.clip) when a_min is greater than a_max.
`cupy.clip(cupy.arange(10), 7, 3)` must be evaluated to [7, 7, 7, 7, 7, 7, 7, 3, 3, 3].

### DIFF
--- a/cupy/core/core.pyx
+++ b/cupy/core/core.pyx
@@ -2319,7 +2319,7 @@ cdef _clip = create_ufunc(
     'cupy_clip',
     ('???->?', 'bbb->b', 'BBB->B', 'hhh->h', 'HHH->H', 'iii->i', 'III->I',
      'lll->l', 'LLL->L', 'qqq->q', 'QQQ->Q', 'eee->e', 'fff->f', 'ddd->d'),
-    'out0 = (in0 < in1 ? in1 : (in0 > in2 ? in2 : in0))')
+    'out0 = in0 < in1 ? in1 : (in0 > in2 ? in2 : in0)')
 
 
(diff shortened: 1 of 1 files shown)


gwtnb:
Thank you for sending a PR. Is this behavior defined in `numpy.clip`'s specification?

asi1024:
According to the documentation, numpy.clip returns "An array with the elements of a, but where values < a_min are replaced with a_min, and those > a_max with a_max."
So, clip([0], 7, 3) should be evaluated to [7].

gwtnb:
> So, clip([0], 7, 3) should be evaluated to [7].

I see, but it seems unclear for `[4]`. 4 is less than the minimum and is greater than the maximum.

asi1024:
Yes, it is unclear (undefined probably), but it is clear for `[0]` and cupy.clip returns `[3]`

gwtnb:
I see. This PR fixes `cupy.clip`'s defined and undefined behavior, so should be merged. Before merging, could you remove the outermost parenthesis? `out0 = (...)` -> `out0 = ...`

asi1024:
Certainly. I fixed.

asi1024:
Oops! This PR might not make sense...
```
>>> numpy.clip([844], [536], [475])
array([536])
>>> numpy.clip(844, 536, 475)
475
```

gwtnb:
Although this PR fails to emulate numpy's buggy behavior, it should be merged to reduce branches.

PR #1618, opened 2016-09-08
Fix build for Cython 0.23

gwtnb:
Fixes https://github.com/pfnet/chainer/issues/1609.

Ref: https://github.com/pfnet/chainer/pull/1612


okuta:
LGMT!

PR #1626, opened 2016-09-09
Add use_cudnn flag to BatchNormalization link

beam2d:
`batch_normalization` and `fixed_batch_normalization` functions support the `use_cudnn` argument, while `BatchNormalization` link does not have a corresponding attribute, which is added by this PR.


gwtnb:
LGTM

PR #1628, opened 2016-09-09
add chainer.training.extensions.snapshot_json

Hiroshiba:
Add chainer.training.extensions.snapshot_json and it's test.
There are no way to save dict or list object with trainer carefree, like settings or texts.


beam2d:
I think it is better to provide a function that dumps an object to a json file, and just let user to combine `snapshot_object` with that function specified at the `savefun` argument. It is still ok to provide a wrapper of it as a dedicated function (named like `snapshot_json`), but I think we do not need to introduce a new extension for it, because the functionality is fully covered by `snapshot_object`.

Hiroshiba:
Thank you. I think so.

PR #1629, opened 2016-09-09
Treat future.types.newint.newint properly in get_device

gwtnb:
Note that we do not use `isinstance` in `get_device` for performance.

### DIFF
--- a/chainer/cuda.py
+++ b/chainer/cuda.py
@@ -136,6 +136,16 @@ if available:
     cuda.set_allocator(memory_pool.malloc)
 
 
+if six.PY2:
+    try:
+        from future.types.newint import newint as _newint
+        _integer_types = six.integer_types + (_newint,)
+    except ImportError:
+        _integer_types = six.integer_types
+else:
+    _integer_types = six.integer_types
+
+
 # ------------------------------------------------------------------------------
 # Global states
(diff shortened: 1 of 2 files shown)


gwtnb:
```python
import chainer
from chainer import cuda
from builtins import int

gpu = int(3)
model = L.Linear(1,1)
cuda.get_device(gpu).use()
model.to_gpu(gpu)
# before merging this PR, this print says device 0 (should be device 3)
print(cuda.get_device(model.W.data))
```

gwtnb:
future: https://pypi.python.org/pypi/future/

muupan:
Shouldn't we add tests for this feature? Though I guess it would not easy.

gwtnb:
Writing the test is not difficult, but running the test may be difficult.

muupan:
I ran `nosetests -s tests/chainer_tests/test_cuda.py` with (py2, py3) x (with-future, without-future) without error.

py2 without future
```
nosetests tests/chainer_tests/test_cuda.py
..S...........................................
Ran 46 tests in 0.394s

OK (SKIP=1)
```

py2 with future, py3 without future, py3 with future
```
nosetests tests/chainer_tests/test_cuda.py
..............................................
Ran 46 tests in 0.397s

OK
```

LGTM.

PR #1633, opened 2016-09-09
Accept dtype to add parameters via Link's __init__ method.

### LINKED ISSUE #1631: Adding parameter via Link's constructor does not accept its dtype.
Relating to #1295 .

While Link's `add_param` method has `dtype` keyword parameter, adding parameter via its constructor's keyword parameter dictionary does not support `dtype` option, ending up with making parameter with `dtype` of FP32.

http://docs.chainer.org/en/stable/reference/core/link.html#chainer.Link

This prevents us from adding parameters via Link's constructor when their expected `dtype` are not FP32, the following weight initialization does not work:

```Python
W_shape = (in_channels, out_channels) + ksize
super(DeconvolutionND, self).__init__(W=W_shape)
initializers.init_weight(self.W.data, initialW)    # initialW.dtype is other than numpy.float32, raising an exception
```

To solve this, we would be able to extend the protocol of Link constructor's keyword parameter as following:

```
<parameter> := <name>=<shape>
             | <name>=(<shape>, <dtype>)
```

With this, we can write the previous code part as:

```Python
W_shape = (in_channels, out_channels) + ksize
super(DeconvolutionND, self).__init__(W=(W_shape, W_dtype))    # W_dtype is other than numpy.float32
initializers.init_weight(self.W.data, initialW)                # actually initialW.dtype is not needed
```

This solution is based on that initializers are not responsible of converting parameter data types, so data types need to be determined when they are added by `add_param`.

takagi:
Fix #1631.

This PR makes Link's `__init__` method to accept dtypes as well as shapes to register parameters as the following:

    l = link.Link(W=((2, 3), 'd'))

where link `l` has parameter `W` with its shape of `(2, 3)` and dtype of `'d'`.

### DIFF
--- a/chainer/link.py
+++ b/chainer/link.py
@@ -1,3 +1,4 @@
+import collections
 import copy
 import warnings
 
@@ -8,14 +9,24 @@ from chainer import cuda
 from chainer import variable
 
 
-def _ensure_shape_dtype(value):
-    if isinstance(value, tuple) and \
-       len(value) == 2 and \
-       not isinstance(value[1], six.integer_types):
-        # Approximately identify a shape-dtype pair.
-        return value
+def _is_shape(value):
+    if value is None:
+        return True
+    elif isinstance(value, six.integer_types):
+        return True
+    elif isinstance(value, collections.Sequence):
+        return all(isinstance(x, six.integer_types) for x in value)
     else:
-        return (value, numpy.float32)
+        return False
+
+
+def _ensure_shape_dtype(value):
+    # Return value paired with dtype FP32 if it is a shape.
+    if _is_shape(value):
+        return (value, 'f')
+    # Otherwise, returns it with assuming a shape-dtype pair.
+    else:
+        return value
 
 
 class Link(object):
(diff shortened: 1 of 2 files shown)


takagi:
I find that I need to fix classes other than `Link`, at least `Chain` and `ChainList`.

chainer/link.py
```diff
@@ -83,7 +104,9 @@ def __call__(self, x):
 
     Args:
         params: Shapes of initial parameters. The keywords are used as their
-            names. The names are also set to the parameter variables.
+            names. The names are also set to the parameter variables. You may
+            pass tuples of a shape and a dtype ``(shape, dtype)`` to add
+            parameters with specifying their dtypes.
```
bkvogel: Consider rewording for clarity. Maybe something like:
params: Names, shapes, and optional dtypes of initial parameters. The keywords are used as the parameter names and the corresponding values consist either of the shape or a tuple of shape and a dtype ``(shape, dtype)``. If only the shape is supplied, the default dtype will be used.

takagi: I fixed as you suggested!

bkvogel:
LGTM!

PR #1643, opened 2016-09-12
Fix nvtx on Read the Docs

okuta:
`cupy_cuda.h` is broken.

### DIFF
--- a/cupy/cuda/cupy_cuda.h
+++ b/cupy/cuda/cupy_cuda.h
@@ -589,17 +589,59 @@ int cudaProfilerStop() {
 // nvToolsExt.h
 ///////////////////////////////////////////////////////////////////////////////
 
+#define NVTX_VERSION 1
+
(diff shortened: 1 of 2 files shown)


cupy/cuda/cupy_cuda.h
```diff
@@ -589,17 +589,59 @@ int cudaProfilerStop() {
 // nvToolsExt.h
 ///////////////////////////////////////////////////////////////////////////////
 
+#define NVTX_VERSION 1
```
gwtnb: `NVTX_VERSION` of CUDA 8.0RC is 2. Is this value okay?

okuta: Version 1 and 2 are both OK.
If use version 1, stub statement is small.

gwtnb: I see.

gwtnb:
LGTM

PR #1650, opened 2016-09-14
Fix AlexNet in examples to use the correct order of layers

rezoo:
Despite the original AlexNet applies the local response normalization layer after the relu, the current example applies the relu after the local response normalization layer.
(This bug was originally reported by another guy in the nvidia event)

### DIFF
--- a/examples/imagenet/alex.py
+++ b/examples/imagenet/alex.py
@@ -23,10 +23,10 @@ class Alex(chainer.Chain):
         self.train = True
 
     def __call__(self, x, t):
-        h = F.max_pooling_2d(F.relu(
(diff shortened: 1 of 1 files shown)


okuta:
Is bvlc_alexnet wrong?
https://github.com/BVLC/caffe/blob/master/models/bvlc_alexnet/train_val.prototxt

rezoo:
@okuta No. The original paper describes:

    We applied this normalization after applying the ReLU nonlinearity in certain layers (see Section 3.5).

Please see https://papers.nips.cc/paper/4824-imagenet-classification-with-deep-convolutional-neural-networks.pdf

okuta:
Thank you!
LGTM

PR #1652, opened 2016-09-14
Use shared memory in multiprocess iterator

okuta:
This PR reduce training time.

master
```
% ./train_imagenet.py /data/ILSVRC/train10k.txt /data/ILSVRC/val1k.txt --arch alex --gpu 2 --epoch 2 --batchsize 64
228.25s user 55.60s system 296% cpu 1:35.78 total
```

sharedmem
```
% ./train_imagenet.py /data/ILSVRC/train10k.txt /data/ILSVRC/val1k.txt --arch alex --gpu 2 --epoch 2 --batchsize 64
209.88s user 55.53s system 552% cpu 48.033 total
```

### DIFF
--- a/chainer/iterators/multiprocess_iterator.py
+++ b/chainer/iterators/multiprocess_iterator.py
@@ -35,9 +35,8 @@ class MultiprocessIterator(iterator.Iterator):
         n_processes (int): Number of worker processes. The number of CPUs is
             used by default.
         n_prefetch (int): Number of prefetch batches.
-        shared_mem (int):
-
-
+        shared_mem (int): The size of using shared memory per data.
+            If ``None``, size is adjusted automatically.
 
     """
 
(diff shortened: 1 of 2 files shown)


chainer/iterators/multiprocess_iterator.py
```diff
@@ -29,11 +34,15 @@ class MultiprocessIterator(iterator.Iterator):
             order of indexes.
         n_processes (int): Number of worker processes. The number of CPUs is
             used by default.
+        n_prefetch (int): Number of prefetch batches.
```
bkvogel: Is there any reason to use a value other than the default of 1? Of so, please consider to document the benefit or tradeoffs in using larger values.

chainer/iterators/multiprocess_iterator.py
```diff
@@ -29,11 +34,15 @@ class MultiprocessIterator(iterator.Iterator):
...
+        shared_mem (int):
```
bkvogel: Please document.

bkvogel:
LGTM! I also verified that sharedmem is faster than existing code on your imagenet example.

PR #1657, opened 2016-09-14
Use tanh instead of exp

okuta:
rel #1047

### DIFF
--- a/chainer/functions/activation/lstm.py
+++ b/chainer/functions/activation/lstm.py
@@ -26,7 +26,7 @@ def _grad_tanh(x):
 
 _preamble = '''
 template <typename T> __device__ T sigmoid(T x) {
-    T half = 0.5;
+    const T half = 0.5;
     return tanh(x * half) * half + half;
 }
 template <typename T> __device__ T grad_sigmoid(T y) { return y * (1 - y); }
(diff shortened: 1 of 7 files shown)


chainer/functions/activation/lstm.py
```diff
@@ -24,7 +25,9 @@ def _grad_tanh(x):
 
 
 _preamble = '''
-template <typename T> __device__ T sigmoid(T x) { return 1 / (1 + exp(-x)); }
+template <typename T> __device__ T sigmoid(T x) {
+    return tanh(x * 0.5) * 0.5 + 0.5;
```
gwtnb: `0.5` is `double`. Is it okay?

gwtnb:
flake8

chainer/functions/activation/lstm.py
```diff
@@ -24,7 +25,10 @@ def _grad_tanh(x):
 
 
 _preamble = '''
-template <typename T> __device__ T sigmoid(T x) { return 1 / (1 + exp(-x)); }
+template <typename T> __device__ T sigmoid(T x) {
+    T half = 0.5;
```
gwtnb: Is it okay to declare without `const`? In other words, does `nvcc` treat `half` as a constant?

gwtnb:
LGTM

PR #1662, opened 2016-09-16
Fix: InceptionBN.__call__

cympfh:
BatchNormalization (http://docs.chainer.org/en/stable/_modules/chainer/links/normalization/batch_normalization.html#BatchNormalization) can set test-mode on `__call__(self, x, test=False)`,
but InceptionBN (http://docs.chainer.org/en/stable/_modules/chainer/links/connection/inceptionbn.html#InceptionBN) has a difference interface `__call__(self, x)` and stores test-mode as a property.
The difference confuses us.

### DIFF
--- a/chainer/links/connection/inceptionbn.py
+++ b/chainer/links/connection/inceptionbn.py
@@ -40,10 +40,6 @@ class InceptionBN(link.Chain):
 
     .. seealso:: :class:`Inception`
 
-    Attributes:
-        train (bool): If ``True``, then batch normalization layers are used in
-            training mode. If ``False``, they are used in testing mode.
-
     """
 
     def __init__(self, in_channels, out1, proj3, out3, proj33, out33,
@@ -96,10 +92,7 @@ class InceptionBN(link.Chain):
         if pooltype != 'max' and pooltype != 'avg':
             raise NotImplementedError()
 
-        self.train = True
-
-    def __call__(self, x):
-        test = not self.train
+    def __call__(self, x, test=False):
         outs = []
 
         if self.out1 > 0:
--- a/chainer/links/connection/inceptionbn.py
+++ b/chainer/links/connection/inceptionbn.py
@@ -98,8 +98,16 @@ class InceptionBN(link.Chain):
 
         self.train = True
 
-    def __call__(self, x):
-        test = not self.train
+    def __call__(self, x, test=None):
+        """
+        Args:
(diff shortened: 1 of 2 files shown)


delta2323:
Thank you for suggestion. I agree with you in that identical APIs for `BatchNormalization` and `InceptionBN` reduce the confusion. On the other hand, It is difficult for us to remove `train` property because we have the [policy for the API compatibility](http://docs.chainer.org/en/stable/compatibility.html). One of the solution is to offer both ways to set the train/test mode of `InceptionBN` and define the priority between the two (e.g. if the `test` argument of `__call__` is explicitly set, we adopt it).

By the way, we make use of the fact that `train` mode is set through the attribute in the ImageNet example ([here](https://github.com/pfnet/chainer/blob/master/examples/imagenet/train_imagenet.py#L141) is the code). If we remove the `train` property or make it obsolete, it is better not to use it.

cympfh:
I tried to keep the compatibility.

Following 2 codes are same.

```
model.__call__(x, test=bl)
```

```
model.train = not bl
model.__call__(x)
```

delta2323:
Thanks! I will check your change.

delta2323:
We discussed within the developer team and concluded that it is confusing for users to be provided two ways to do the same thing (in this case, to set the test mode through the attribute and through the argument of `__call__` method). So, it is better to choose the one as the recommended way and make the other deprecated.

On the other hand, it is much confusing to change the "recommended way" within short terms, so we should discuss this point enough before making a change. I would like your opinion on this problem. Thanks in advance.

delta2323:
@beam2d Could you check this PR?

beam2d:
I'll check it shortly.

beam2d:
Note that I'm preparing a training mode configuration in #2186 and then this change will not be needed in v2, and so this change can be left unmerged for v2. It is still meaningful to merge it to v1. @delta2323 How do you think of it?

delta2323:
I agree that we do not need this feature any longer in v2.

I come think it is OK to merge in v1 if we add sufficient explanations. I concerned the multiple standard ways of setting the training mode could confuse users. But I come to think it is acceptable if we write explicitly that there are two ways to do the same thing and that it is for keeping backward compatibility.

delta2323:
I decided to merge this PR to v1. We will send another PR for v2 that renews its implementation.

PR #1663, opened 2016-09-16
Support FP16 and FP64 in N-dimensional convolution link

### LINKED ISSUE #1625: Test N-dimensional convolution link for dtypes of FP16 and FP64
Follows #1279 and #1556.

Since #1295 is now merged to master, we can add test for dtypes of FP16 and FP64 to N-dimensional convolution **LINK**.

takagi:
This PR make `ConvolutionND` link support `dtype` of FP16 and FP64, which are currently supported only in `convolution_nd` function. Fix #1625.

There are the following possible approaches how to supply its parameter `dtype` to the link:
- Make `ConvolutionND` link to accept parameter `dtype` so that it can `add_param` parameters with the supplied `dtype`.
- Make initializers to be responsible of converting parameters `dtype`

This PR implements in the former approach and, as a future issue, it would be expected that shape placeholder facility also supports `dtype` of uninitialized parameters as well as their `shape`.

### DIFF
--- a/tests/chainer_tests/links_tests/connection_tests/test_convolution_nd.py
+++ b/tests/chainer_tests/links_tests/connection_tests/test_convolution_nd.py
@@ -29,7 +29,7 @@ class TestConvolutionND(unittest.TestCase):
 
         self.link = convolution_nd.ConvolutionND(
             ndim, 3, 2, self.ksize, stride=self.stride, pad=self.pad,
-            initial_bias=initializers.Uniform(1), bias_dtype=self.dtype)
+            initial_bias=initializers.Uniform(scale=1., dtype=self.dtype))
         self.link.cleargrads()
 
         x_shape = (2, 3) + self.dims
--- a/tests/chainer_tests/links_tests/connection_tests/test_convolution_nd.py
+++ b/tests/chainer_tests/links_tests/connection_tests/test_convolution_nd.py
@@ -66,7 +66,7 @@ class TestConvolutionND(unittest.TestCase):
         self.assertEqual(y_cpu.data.dtype, self.x.dtype)
 
         self.link.to_gpu()
-        x_gpu = chainer.Variable(cuda.to_gpu(self.x.dtype))
+        x_gpu = chainer.Variable(cuda.to_gpu(self.x))
         y_gpu = self.link(x_gpu)
         self.assertEqual(y_gpu.data.dtype, self.x.dtype)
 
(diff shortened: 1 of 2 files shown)


takagi:
I will fix this change to use `Link.add_param` 's `initializer` parameter.

takagi:
> I will fix this change to use `Link.add_param` 's `initializer` parameter, which @okuta pointed out to me.

I've fixed.

takagi:
Fix #1718.

tests/chainer_tests/links_tests/connection_tests/test_convolution_nd.py
```diff
@@ -57,12 +63,12 @@ def test_col2im_consistency(self):
     def check_forward_consistency(self):
         x_cpu = chainer.Variable(self.x)
         y_cpu = self.link(x_cpu)
-        self.assertEqual(y_cpu.data.dtype, numpy.float32)
+        self.assertEqual(y_cpu.data.dtype, self.x.dtype)
```
okuta: How about using `self.dtype`?.

tests/chainer_tests/links_tests/connection_tests/test_convolution_nd.py
```diff
@@ -57,12 +63,12 @@ def test_col2im_consistency(self):
...
 
         self.link.to_gpu()
         x_gpu = chainer.Variable(cuda.to_gpu(self.x))
         y_gpu = self.link(x_gpu)
-        self.assertEqual(y_gpu.data.dtype, numpy.float32)
+        self.assertEqual(y_gpu.data.dtype, self.x.dtype)
```
okuta: How about using `self.dtype`?.

okuta:
LGTM except comments.

tests/chainer_tests/links_tests/connection_tests/test_convolution_nd.py
takagi: Fixed!

tests/chainer_tests/links_tests/connection_tests/test_convolution_nd.py
takagi: Fixed!

takagi:
I've fixed the points as you commented, as well as the the docstring as beam2d pointed out on the discussion.

okuta:
OK.

PR #1668, opened 2016-09-16
Remove unused variable in trigonometric functions.

takagi:
This PR removes an unused variable in trigonometric functions.


gwtnb:
LGTM

PR #1672, opened 2016-09-20
Remove nvtx on Windows for a while

gwtnb:
Exclude `cupy.cuda.nvtx` on Windows for a while to solve a build issue. This should be fixed in near future.

Ref: https://github.com/pfnet/chainer/issues/1660


okuta:
LGTM!

PR #1673, opened 2016-09-20
Added parameter shape placeholder support to deconvolution_2d.

bkvogel:
Add support for the "parameter shape placeholder" feature in the Deconvolution2D link. The `in_channels` parameter can now optionally be set to `None` to cause the initialization of the weights parameters to be deferred until the first forward pass.


tests/chainer_tests/links_tests/connection_tests/test_deconvolution_2d.py
```diff
@@ -92,4 +92,75 @@ def test_backward_gpu(self):
         self.check_backward(cuda.to_gpu(self.x), cuda.to_gpu(self.gy))
 
 
+@parameterize(
+    *testing.product({
+        'out_channels': [2],
```
delta2323: We need not to parameterize a variable if there is only one possible value.

bkvogel:
Thank you, I have responded to your comments. Could you check if the changes look okay?

delta2323:
Thank you!

PR #1680, opened 2016-09-22
Fix addgrad for cleargrad

gwtnb:
TODO: write test

### DIFF
--- a/chainer/variable.py
+++ b/chainer/variable.py
@@ -289,17 +289,19 @@ Actual: {0}'''.format(type(data))
         dst_dev = cuda.get_device(self.data)
 
         if src_dev.id == dst_dev.id:
-            if dst is None:
-                xp = cuda.get_array_module(src)
-                self._grad = xp.copy(src)
-            else:
-                self._grad += src
+            with dst_dev:
+                if dst is None:
+                    xp = cuda.get_array_module(src)
+                    self._grad = xp.copy(src)
+                else:
+                    self._grad += src
             return
 
         if dst_dev.id < 0:
             src_grad = cuda.to_cpu(src)
         else:
             src_grad = cuda.to_gpu(src, device=dst_dev)
+            print(dst_dev, cuda.get_device(src_grad))
 
         if dst is None:
             self._grad = src_grad
--- a/chainer/variable.py
+++ b/chainer/variable.py
@@ -304,7 +304,8 @@ Actual: {0}'''.format(type(data))
         if dst is None:
             self._grad = src_grad
         else:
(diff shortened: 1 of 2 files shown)


gwtnb:
This bug is reported on a blog post in Japanese http://hiho-developer.hatenablog.com/entry/20160922/1474541060.

chainer/variable.py
```diff
@@ -283,22 +283,29 @@ def addgrad(self, var):
...
+        dst_dev = cuda.get_device(self.data)
+
+        if src_dev.id == dst_dev.id:
+            if dst is None:
+                xp = cuda.get_array_module(src)
+                self._grad = xp.copy(src)
             else:
-                with dst_dev:
-                    dst += xp.copy(src)
+                self._grad += src
```
beam2d: Switch the device here again. It is better to wrap this `if` statement by a device-switching `with` block.

chainer/variable.py
```diff
@@ -283,22 +283,29 @@ def addgrad(self, var):
...
-            dst_dev = dst.device
-            if dst_dev == src.device:
-                dst += src
+        src_dev = cuda.get_device(src)
+        dst_dev = cuda.get_device(self.data)
+
+        if src_dev.id == dst_dev.id:
+            if dst is None:
+                xp = cuda.get_array_module(src)
+                self._grad = xp.copy(src)
```
beam2d: Switch the device here appropriately. Note that cupy.copy allocates an array on the current device (instead of the device that `src` reside). See also the next comment.

okuta:
I wrote test.
And, please merge #1778 before this PR.

chainer/variable.py
```diff
@@ -283,22 +283,31 @@ def addgrad(self, var):
...
+            src_grad = cuda.to_cpu(src)
         else:
-            dst_dev = dst.device
-            if dst_dev == src.device:
-                dst += src
-            else:
-                with dst_dev:
-                    dst += xp.copy(src)
+            src_grad = cuda.to_gpu(src, device=dst_dev)
+            print(dst_dev, cuda.get_device(src_grad))
```
beam2d: Remove this print

beam2d:
I removed a print. LGTM except that point. I'll ask other members to review it.

delta2323:
Could you add test cases in which both `clear_src_grad` and `clear_dst_grad` is True?

beam2d:
Added a test case.

delta2323:
Thanks. Now LGTM.

okuta:
Thank you!

PR #1685, opened 2016-09-22
Improve col2im and im2col

okuta:


rezoo:
I assume that it does simplify the code, but its computational cost is almost the same. Is it right?

okuta:
Yes, almost the same.
But the computational cost is not same in `col2im_gpu` with dilation mode.

rezoo:
Ok I understand. In the case of dilated convolution, the previous one is not efficient, whereas the proposed one efficiently handles such case.

rezoo:
Reading the PR, I conclude that there does not any crucial problems.

gwtnb:
LGTM

rezoo:
Thank you! then I would like to merge this.

PR #1697, opened 2016-09-23
Implementation of CuPy Fusion

asi1024:
I Implemented the fusion class for CuPy ufuncs.


okuta:
#2297 includes this PR. And #2297 was merged.
Thank you!

PR #1698, opened 2016-09-23
Fixed bug in link when passing numpy int types in shape.

bkvogel:
This fixes a bug when supplying a shape to the `__init__()` of link that contains numpy integer types. For example a shape of the form `(numpy.int64, int)` should be considered a valid shape.

### DIFF
--- a/chainer/link.py
+++ b/chainer/link.py
@@ -15,11 +15,11 @@ def _is_shape(value):
         return True
     elif isinstance(value, collections.Sequence):
         try:
-            return all(isinstance(int(x), six.integer_types) for x in value)
+            return all(int(x) for x in value)
         except TypeError:
             return False
     try:
-        return isinstance(int(value), six.integer_types)
+        return int(value)
     except TypeError:
         return False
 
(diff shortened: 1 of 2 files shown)


chainer/link.py
```diff
@@ -13,13 +13,18 @@
 def _is_shape(value):
     if value is None:
         return True
-    elif isinstance(value, six.integer_types):
-        return True
     elif isinstance(value, collections.Sequence):
-        return all(isinstance(x, six.integer_types) for x in value)
-    else:
+        try:
+            return all(isinstance(int(x), six.integer_types) for x in value)
```
okuta: Please remove `isinstance`.
`int` function returns integer type value.

chainer/link.py
```diff
@@ -13,13 +13,18 @@
...
+        except TypeError:
+            return False
+    try:
+        return isinstance(int(value), six.integer_types)
```
okuta: ditto

bkvogel: Thanks for the suggestion. I have made the change.

okuta:
LGTM!

PR #1705, opened 2016-09-24
without using threading.local()

Hiroshiba:
I used multi thread and multi GPU on training for faster training.
But with chainer1.16.0, I cannot use multi thread by causing error about `_thread_local.default_backprop` is not defined.
This PR resolve this error by using global value.

### DIFF
--- a/chainer/function.py
+++ b/chainer/function.py
@@ -1,7 +1,6 @@
 import collections
 import contextlib
 import os
-import threading
 import traceback
 import weakref
 
--- a/chainer/function.py
+++ b/chainer/function.py
@@ -1,6 +1,7 @@
 import collections
 import contextlib
 import os
+import threading
 import traceback
 import weakref
 
@@ -12,7 +13,8 @@ from chainer import flag
 from chainer.utils import type_check
 from chainer import variable
 
-_backprop_mode = True
+
+_thread_local = threading.local()
 
 
 @contextlib.contextmanager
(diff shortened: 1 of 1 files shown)


gwtnb:
Thank you for sending the PR, but the current patch cannot be merged because it introduces race condition. Could you fix this patch with `threading.local` and `getattr`?

Hiroshiba:
Thank you for answering. I want to fix this PR. Could you tell me the way to fix in more detail.
To replace `_thread_local.default_backprop` to `getattr(_thread_local, 'default_backprop')` ?

Hiroshiba:
I wrote tha [Qiita article (in japanese)](http://qiita.com/Hiroshiba/items/1ab9fb9cea7ab8402f08) that explain what I want to do.

Hiroshiba:
sample code. (need python >= 3.2)
```
import chainer

import numpy
import concurrent.futures


def calc_sigmoid(i):
    value = numpy.array(i, dtype=numpy.float32)
    return chainer.functions.sigmoid(value)

with concurrent.futures.ThreadPoolExecutor(max_workers=4) as executor:
    values = list(executor.map(calc_sigmoid, range(4)))

print(values)

# AttributeError: '_thread._local' object has no attribute 'default_backprop'
```

gwtnb:
OK, to tell the way to fix this PR, I will explain about Python object's attributes.

Unlike C++'s member variables or Java's fields, Python's attributes do not need declarations beforehand:

```python
class Foo(object):
    pass

x = Foo()
x.abcd = 123  # no error
print(x.abcd)  # 123
```

but reading an attribute before assignment is invalid:

```python
print(x.efgh)  # error!
```

This is the cause of this bug. `_thread_local.default_backprop` is not assigned before [reading it](https://github.com/pfnet/chainer/pull/1705/files#diff-648bb1423391876eb3df278b3e720babL41) on a thread.

To fix this problem, [`getattr`](https://docs.python.org/2/library/functions.html#getattr) is usable. This function accesses an attribute with a default value:

```python
print(getattr(x, 'efgh'))  # prints None
print(getattr(x, 'efgh', 321))  # prints 321
```

Hiroshiba:
I got it. Thanks.

gwtnb:
Thank you!

PR #1706, opened 2016-09-24
Revert "Fixed the link to paper pdf"

hido:
Reverts pfnet/chainer#1686, since the paper URL is back to the previous one.


gwtnb:
LGTM

PR #1709, opened 2016-09-25
Use nogil statements to enable cupy to use async callback

okuta:
This PR fixes dead locking.

### DIFF
--- a/cupy/cuda/cublas.pyx
+++ b/cupy/cuda/cublas.pyx
@@ -11,81 +11,84 @@ from cupy.cuda cimport runtime
 
 cdef extern from 'cupy_cuda.h':
     # Context
-    int cublasCreate(Handle* handle)
-    int cublasDestroy(Handle handle)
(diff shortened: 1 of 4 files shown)


gwtnb:
Could you describe how a deadlock occurs?

okuta:
This PR is needed by #1707, because it uses async callback.

rezoo:
I briefly read this PR, and it seems good.

rezoo:
Are there any reasons why ``setStream`` and ``getStream`` do not use ``with noggin:`` statement?

rezoo:
I also think that this PR involves ``enhancement`` or ``feature`` label rather than ``bug``.

okuta:
> Are there any reasons why getStream and setStream do not use with nogil: statement?
I don't know about need or not need nogil statement in this case.

> It seems that this PR involves enhancement or feature label rather than bug.
OK, Please change it.

rezoo:
Thanks. I would like to merge this one.

PR #1713, opened 2016-09-26
CuPy Fusion for Chainer

asi1024:
I rewrote chainer/optimizers by using CuPy-Fusion


okuta:
We are waiting for fusion to become formal function.

PR #1717, opened 2016-09-26
Use destructive astype in convolution-nd function for efficiency.

takagi:
This PR fixes `convolution-nd` function to use destructive `astype` method for efficiency, which is the same approach with #1444 . It improves nearly 10% of its efficiency in my handy profiling.


delta2323:
LGTM. I am checking with Jenkins to check if it passes the tests.

delta2323:
LGTM 👍

PR #1719, opened 2016-09-27
LSTM supports parameter shape place holder

delta2323:
This PR enables us to put `None` as the input size of `LSTM` and `StatelessLSTM` links.

```
a = chainer.Variable(numpy.random.uniform(0, 1, (5, 10)).astype(numpy.float32))
l = L.LSTM(None, 5)
l(a)
```

### DIFF
--- a/chainer/links/connection/lstm.py
+++ b/chainer/links/connection/lstm.py
@@ -22,19 +22,27 @@ class LSTMBase(link.Chain):
                                   initialW=0, nobias=True),
         )
         self.state_size = out_size
+        self.lateral_init = lateral_init
+        self.upward_init = upward_init
+        self.bias_init = bias_init
+        self.forget_bias_init = forget_bias_init
 
-        for i in six.moves.range(0, 4 * out_size, out_size):
+        if in_size is not None:
+            self._initialize_params(in_size)
+
+    def _initialize_params(self, in_size):
+        for i in six.moves.range(0, 4 * self.state_size, self.state_size):
             initializers.init_weight(
-                self.lateral.W.data[i:i + out_size, :], lateral_init)
+                self.lateral.W.data[i:i + self.state_size, :], self.lateral_init)
             initializers.init_weight(
-                self.upward.W.data[i:i + out_size, :], upward_init)
+                self.upward.W.data[i:i + self.state_size, :], self.upward_init)
 
         a, i, f, o = lstm._extract_gates(
-            self.upward.b.data.reshape(1, 4 * out_size, 1))
-        initializers.init_weight(a, bias_init)
-        initializers.init_weight(i, bias_init)
-        initializers.init_weight(f, forget_bias_init)
-        initializers.init_weight(o, bias_init)
+            self.upward.b.data.reshape(1, 4 * self.state_size, 1))
+        initializers.init_weight(a, self.bias_init)
+        initializers.init_weight(i, self.bias_init)
+        initializers.init_weight(f, self.forget_bias_init)
+        initializers.init_weight(o, self.bias_init)
 
 
 class StatelessLSTM(LSTMBase):
@@ -70,6 +78,11 @@ class StatelessLSTM(LSTMBase):
                 output of LSTM units.
 
         """
+        if self.upward.has_uninitialized_params:
+            in_size = x.size // len(x.data)
+            self.upward._initialize_params(in_size)
+            self._initialize_params(in_size)
+
         lstm_in = self.upward(x)
         if h is not None:
             lstm_in += self.lateral(h)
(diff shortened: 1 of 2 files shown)


chainer/links/connection/lstm.py
```diff
@@ -197,6 +211,11 @@ def __call__(self, x):
             ~chainer.Variable: Outputs of updated LSTM units.
 
         """
+        if self.upward.has_uninitialized_params:
+            in_size = x.size // len(x.data)
+            self.upward._initialize_params(in_size)
```
delta2323: Should we make this method public?

bkvogel: Is there any cases where the user might want to call it directly? Maybe if writing another subclass of LSTMBase? If so, it may be ok to make public.

chainer/links/connection/lstm.py
```diff
@@ -70,6 +79,11 @@ def __call__(self, c, h, x):
                 output of LSTM units.
 
         """
+        if self.upward.has_uninitialized_params:
+            in_size = x.size // len(x.data)
```
bkvogel: Is this line needed? I am not sure if the following two lines really need `in_size`. For example, `self._initialize_params(in_size)` does not seem to use `in_size` anywhere.

chainer/links/connection/lstm.py
```diff
@@ -197,6 +211,11 @@ def __call__(self, x):
             ~chainer.Variable: Outputs of updated LSTM units.
 
         """
+        if self.upward.has_uninitialized_params:
+            in_size = x.size // len(x.data)
```
bkvogel: Dito. Is `in_size` needed?

chainer/links/connection/lstm.py
delta2323: As `upward._initialize_params` does need `in_size`, we need `in_size` anyway. But as you pointed out, `self._initialize_params` does not use this argument. So, we should remove it from this method.

chainer/links/connection/lstm.py
delta2323: In the original implementation of the LSTM link, the LSTM link must use its initializers after Linear links are initialized. Specifically, we first initialize weight matrices with temporal initializers to create ones and initialize them with genuine initializers afterwards. We follow this seemingly redundant procedure because passing `lateral_init` and `upward_init` directly to `Linear` links inside leads to wrong initialization (see the discussion [here](https://github.com/pfnet/chainer/pull/1086#r62287522)).

On the other hand, initializations should be done before concrete matrix multiplications, of course. So, the LSTM link must use its initializers between initialization by Linear links and concrete calculations. But as these are coupled in `Linear.__call__`, we cannot insert it between the two. So, we have to call `upward`'s initialization directly here.

I would like to know if there is more clever way of initialization.

bkvogel: I think I understand now. It seems that `Linear._initialize_params` only needs to be called from LSTM link because LSTM wants to use the performance optimization of grouping the 4 linear layers for each gate into a single matrix multiply and also needs to initialize the forget gate differently from the other gates. Otherwise, it would not be necessary to call `Linear._initialize_params` from the LSTM link. I can understand why it is written this way (to improve performance), but a slight drawback is that the initialization code gets trickier. 

Since `Linear._initialize_params` is only needed as a performance optimization, I think maybe it is better to keep this method private for now.

Perhaps another option to avoid calling this private method of Linear link would be to change the LSTM link so that it no longer uses Linear link at all and instead just calls the linear function directly.

delta2323: I agree with keeping the method private as it is.

> Perhaps another option to avoid calling this private method of Linear link would be to change the LSTM link so that it no longer uses Linear link at all and instead just calls the linear function directly.

As `upward` and `lateral` links are [documented](https://github.com/pfnet/chainer/blob/master/chainer/links/connection/lstm.py#L133), we must keep them to support compatibility. Further, if `LSTM` link would hold weight matrices as parameters, we would have to change how to access them (e.g. from `LSTM.upward.W` to `LSTM.upward_W` etc). Although such a compatibility is not supported, it could break some users' codes. So, we need to elaborate the implementation anyway.

bkvogel: I did not notice that `upward` and `lateral` were documented. Since they are, I agree that we must keep them.

bkvogel:
Please document that `in_size` can be `None`:
https://github.com/pfnet/chainer/pull/1719/files#diff-ffcf2278ad274f346becb8b476d7e80bR59
and
https://github.com/pfnet/chainer/pull/1719/files#diff-ffcf2278ad274f346becb8b476d7e80bR121

bkvogel:
LGTM!

PR #1722, opened 2016-09-27
Add extern "C"

gwtnb:
Note that the header files are always included from a C++ file, checking `__cplusplus` is unnecessary.


okuta:
LGTM!

PR #1730, opened 2016-09-28
Progress_bar and print_report support Windows

tapdo:
“progress_bar” and “print_report” does not work well on windows, because
Windows command console does not support ANSI escape sequence. Adding
some functions for this. And these extensions can work well on Windows.

### DIFF
--- a/chainer/training/extensions/print_report.py
+++ b/chainer/training/extensions/print_report.py
@@ -1,7 +1,9 @@
+import os
 import sys
 
 from chainer.training import extension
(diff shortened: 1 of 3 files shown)


gwtnb:
Thank you for sending PR. I confirmed to run on `cmd.exe` properly. I will read code carefully. Please wait for a few days.

okuta:
I also tested on Windows. This PR is good.

gwtnb:
I am sorry to be late. Thank you for your contribution!

PR #1731, opened 2016-09-29
Add elapsed_time to Trainer and LogReport

beam2d:
This PR adds `elapsed_time` property to Trainer. This property returns the elapsed time in seconds since the training began. It properly accumulates the elapsed time even if user resumes the training from snapshot.

It also adds `elapsed_time` entry to the log taken by LogReport, and usages in MNIST examples.

Note: an extension to record the elapsed time was originally proposed by @iwiwi in #1472. At that time, I discussed with him and concluded that the elapsed-time recording should be the standard feature of Trainer, which is realized by the current PR.


gwtnb:
flake8

beam2d:
sorry, fixed

okuta:
Please add test.
But, maybe it is difficult..

chainer/training/trainer.py
```diff
@@ -135,8 +137,25 @@ def __init__(self, updater, stop_trigger=None, out='result'):
         self._done = False
         self._extensions = collections.OrderedDict()
 
+        self._start_at = None
+        self._snapshot_at = 0.0
```
okuta: How about `_snapshot_elapsed_time`?

beam2d:
@okuta I fixed the attribute name and added a small test. Would you check them?

okuta:
The test is OK.
Please fix coding style.

okuta:
LGTM except flake8.

beam2d:
Oops, sorry, fixed.

okuta:
Thank you

PR #1737, opened 2016-10-03
Extract evaluation procedure from training extension `Evaluator`

cocuh:
Add `chainer.training.evaluator.Evaluator` and `chainer.training.evaluator.StandardEvaluator` to evaluate testing error and accuracy.
Although `chainer.training.extensions.Evaluator` can evaluate them *by itself*, *standalone extension* is strange and it should be pure training extension which call external defined evaluation procedure. Thus, I define evaluator interface(`chainer.training.Evaluator`) and move evaluating codes from `chainer.training.extensions.Evaluator` to `chainer.training.evaluator.StandardEvaluator`.

I tested `examples/mnist/train_mnist.py` and it worked.

### DIFF
--- a/chainer/training/evaluator.py
+++ b/chainer/training/evaluator.py
@@ -1,0 +1,197 @@
+import copy
+import six
+
+from chainer import link
+from chainer import reporter as reporter_module
+from chainer import variable
+from chainer.dataset import convert
+from chainer.dataset import iterator as iterator_module
+
+
+class Evaluator(object):
+    """Base class of all evaluators.
+    """
+
+    def run(self):
+        raise NotImplementedError
+
+    def evaluate(self):
+        raise NotImplementedError
+
+    def get_iterator(self, name):
+        raise NotImplementedError
+
+    def get_all_iterators(self):
+        raise NotImplementedError
+
+    def get_target(self, name):
+        raise NotImplementedError
+
+    def get_all_targets(self):
+        raise NotImplementedError
+
+
+def eval_func_with_volatile(eval_func, in_arrays):
+    if isinstance(in_arrays, tuple):
+        in_vars = tuple(
+                variable.Variable(x, volatile='on') for x in in_arrays
+        )
+        eval_func(*in_vars)
+    elif isinstance(in_arrays, dict):
+        in_vars = {
+            key: variable.Variable(x, volatile='on')
+            for key, x in six.iteritems(in_arrays)
+        }
+        eval_func(**in_vars)
+    else:
+        in_var = variable.Variable(in_arrays, volatile='on')
+        eval_func(in_var)
+
+
+class StandardEvaluator(Evaluator):
+    """Standard implementation of Evaluator.
+
+    This is the standard implementation of :class:`Evaluator`.
+    It evaluate *target link* with given iterator and summarize the reported
+    values in the link by calling :func:`~chainer.reporter.report`.
+    It accepts one or more iterators and one or more links.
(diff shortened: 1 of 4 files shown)


chainer/training/extensions/evaluator.py
```diff
@@ -1,17 +1,10 @@
...
-
-import six
-
 from chainer.dataset import convert
-from chainer.dataset import iterator as iterator_module
-from chainer import link
 from chainer import reporter as reporter_module
 from chainer.training import extension
-from chainer import variable
+from chainer.training import StandardEvaluator
```
delta2323: We forbid the direct import of class objects (See the [coding guideline](http://docs.chainer.org/en/stable/contribution.html#coding-guidelines) for detail). So could you replace it with `from chainer.training import evaluator` and use `evaluator.StandardEvaluator` instead?

chainer/training/evaluator.py
```diff
@@ -0,0 +1,195 @@
...
+        >>>     (np.asarray([0, 1], dtype=np.float32), np.int32(1)),  # true
+        >>>     (np.asarray([0, 1], dtype=np.float32), np.int32(0)),  # false
+        >>> ]
+        >>>
+        >>> iterator = chainer.iterators.SerialIterator(data, 5, repeat=False)
+        >>>
+        >>> evaluator = StandardEvaluator(iterator, model)
+        >>> result = evaluator.run()
+        >>> print(result)
+        {'main/accuracy': 0.8000000119209289, 'main/loss': 0.5132616162300109}
```
delta2323: I think this example is useful for testing `StandardEvaluator`. So, could you write unittests from it?

chainer/training/extensions/evaluator.py
cocuh: Oops I missed it.

delta2323:
Thank you for updating. It looks good to me. I am now running tests and will merge it if tests will pass.

cocuh:
Thank you for reviewing! This PR is bit large. Please decide carefully whether merge or not.

delta2323:
LGTM 👍

delta2323:

jekbradbury:
This breaks backwards compatibility in a slightly confusing way if you were using a custom Evaluator (now you have to pass that custom Evaluator to the new Evaluator extension as the `evaluator=` kwarg...but you still have to pass `iterator` and `target` positional arguments, even though they're completely ignored now).

delta2323:
@jekbradbury Thank you for reporting the problem. I sent PR #1788 that hopefully fixes your problem. Could you check that to see if I understand your problem correctly?

jekbradbury:
That helps if you are using Evaluator on its own (outside of Trainer) but the use case I'm talking about is when you are using an evaluator as an extension to Trainer. If you want to use a custom evaluator class as a Trainer extension, you now have to create that evaluator and pass it as a keyword argument `evaluator=` to the new Evaluator extension...but you also have to redundantly pass `iterator` and `target` as positional arguments even though they're ignored.

old: `trainer.extend(CustomEvaluator(dev_iter, model), eval_trigger)`
new: `trainer.extend(Evaluator(dev_iter, model, evaluator=CustomEvaluator(dev_iter, model)), eval_trigger)`

This change is documented, but it's still unintuitive. Maybe it would be better to give `iterator` and `target` defaults of `None`?

delta2323:
@jekbradbury Thank you for the detail explanation. We found another problem on the backward compatibility. So, I think it is reasonable to revert this PR (#1790) and reopen this PR.

PR #1741, opened 2016-10-04
Rename unary_math_function_test

### LINKED ISSUE #1740: Rename unary_math_function_test.py
`nose` searches test files using [a regex](http://nose.readthedocs.io/en/latest/usage.html#extended-usage). It matches to `unary_math_function_test.py`, so we have to rename before releasing 1.17.0.

delta2323:
Fix #1740. Related #1534.

### DIFF
--- a/chainer/testing/__init__.py
+++ b/chainer/testing/__init__.py
@@ -12,7 +12,7 @@ parameterize = parameterized.parameterize
 product = parameterized.product
 product_dict = parameterized.product_dict
 
-unary_math_function_test = unary_math_function_test.unary_math_function_test
+unary_math_function_unittest = unary_math_function_test.unary_math_function_unittest
 
 with_requires = helper.with_requires
 
--- a/chainer/testing/__init__.py
+++ b/chainer/testing/__init__.py
@@ -12,7 +12,8 @@ parameterize = parameterized.parameterize
 product = parameterized.product
 product_dict = parameterized.product_dict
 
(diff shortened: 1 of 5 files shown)


gwtnb:
It seems better to add `_` prefix (e.g. `_unary_math_function_unittest`, `_unary_math_function_test.py` for a while to rename the entities in future.

gwtnb:
Documents should be fixed. https://github.com/pfnet/chainer/pull/1534/files#diff-6e9377c0cf634ee425fa26edc48c3e6a

delta2323:
> It seems better to add _ prefix (e.g. _unary_math_function_unittest, _unary_math_function_test.py for a while to rename the entities in future.

You mean we should change the names temporally? If so, I am afraid I am not sure why we should do so.

gwtnb:
> You mean we should change the names temporally?

Yes.

If an entity will be renamed, users should not use it. Adding `_` prefix is an easy way to express that the entity is private. Ref: https://github.com/pfnet/chainer/issues/1716

delta2323:
What I am not sure is that why should it be temporal? Do you think we should remove the prefix `_` again? 

The developers of Chainer functions can use this decorator to test them. So another question is that users in this context is those who do not create official Chainer functions, isn't it?

gwtnb:
> What I am not sure is that why should it be temporal?

If the name is good to use for long time, we do not have to add `_`. I will merge this to

PR #1744, opened 2016-10-05
Adds Subpixel Depth2Space and Space2Depth Functionality.

tjtorres:
This adds some new functionality to take stacked outputs from a convolution
and then intersperse them to create an upscaled array (Depth2Space).

It also adds the inverse operation (Space2Depth) for the purpose of building symmetrical
architectures for subpixel autoencoders etc.

Details of the algorithm's underpinnings can be found at https://arxiv.org/abs/1609.05158

### DIFF
--- a/chainer/functions/__init__.py
+++ b/chainer/functions/__init__.py
@@ -94,6 +94,8 @@ from chainer.functions.pooling import max_pooling_2d
 from chainer.functions.pooling import roi_pooling_2d
 from chainer.functions.pooling import spatial_pyramid_pooling_2d
 from chainer.functions.pooling import unpooling_2d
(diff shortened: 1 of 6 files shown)


delta2323:
Thank you for sending the PR! It seems Depth2Space and Space2Depth are independent, except that they are in the same module. So, is it possible to divide this PR into two smaller PRs so that we can narrow the scope of review? Thank you for your corporation.

tjtorres:
Sure, I will submit them as separate PRs. Thanks.

delta2323:
Thanks. Then, I will close this PR when you send the PRs.

tjtorres:
Ok added PRs #1748 and #1749 splitting this PR into its component halves.

PR #1748, opened 2016-10-07
Adds Subpixel Depth2Space Functionality

tjtorres:
This adds some new functionality to take stacked outputs from a convolution
and then intersperse them to create an upscaled array (Depth2Space).

Details of the algorithm's underpinnings can be found at https://arxiv.org/abs/1609.05158

This PR splits commit #1744 into two pieces. One for Depth2Space and the other for Space2Depth.


mattya:
Thank you for the PR.
I think we can reduce the number of operations by like following code (the result is not same as your code but I think it's essentially same),
```
bsize, c, a, b = X.shape
c /= r ** 2
X = np.transpose(X, (0, 2, 3, 1))
X = np.reshape(X, (bsize, a, b, r, r, c))
X = np.transpose(X, (0, 1, 3, 2, 4, 5))
X = np.reshape(X, (bsize, a*r, b*r, c))
X = np.transpose(X, (0, 3, 1, 2))
return X,
```
do you have any reason you use split&squeeze&concatenate ?

chainer/functions/subpixel/depth2space.py
```diff
@@ -0,0 +1,69 @@
...
+        gy = xp.transpose(gy, (0, 3, 1, 2))
+        return gy,
+
+
+def depth2space(X, r):
+    """Computes the depth2space transformation for subpixel calculations.
+
+    Args:
+        X (Variable): Variable holding a 4d array of
+        shape (batch, dim1, dim2, channel*r)
```
mattya: perhaps (batch, channel*r, dim1, dim2) ?

tjtorres: Yes, I forgot to rewrite the docstring here after switching to the chainer indexing scheme for convolution data. Will change this.

tjtorres:
@mattya In response to your comment about reducing operation number, I think you're correct. Your sequence seems to produce the same interleaving as before with fewer operations. The only difference appears to be how the ordering of the original channels is interwoven to make the full resolution data and that's inconsequential here given commutativity. I'll work in your changes and update the tests for the expected output. Thanks for that catch.

duschendestroyer:
I tried your code and was not able to successfully train a network.
I believe the backward pass is incorrect.
This is the code I used for comparison:

def depth2space_(X, r=2):
```
        bsize, c, a, b = X.data.shape
        c /= r ** 2
        X = F.transpose(X, (0, 2, 3, 1))
        X = F.reshape(X, (bsize, a, b, r, r, c))
        X = F.transpose(X, (0, 1, 3, 2, 4, 5))
        X = F.reshape(X, (bsize, a * r, b * r, c))
        X = F.transpose(X, (0, 3, 1, 2))
        return X
```

It produces the same forward result, but different gradients.

tjtorres:
Hmmmm, I was able to train a DCGAN with the previous version (larger number of operations) just fine before. I'll investigate the issues with this new version.

chainer/functions/subpixel/depth2space.py
```diff
@@ -0,0 +1,60 @@
...
+        X = xp.reshape(X, (bsize, a * self.r, b * self.r, c))
+        X = xp.transpose(X, (0, 3, 1, 2))
+        return X,
+
+    def backward(self, inputs, grad_outputs):
+        gy, = grad_outputs
+        xp = cuda.get_array_module(gy)
+        bsize, c, a, b = gy.shape
+        gy = xp.transpose(gy, (0, 2, 3, 1))
+        gy = xp.reshape(gy, (bsize, a / self.r, b / self.r, self.r, self.r, c))
```
mattya: I think the correct order of arguments are:
`gy = xp.reshape(gy, (bsize, a / self.r, self.r, b / self.r, self.r, c))`

tjtorres:
Ok I have updated the branch to correct for the transposition error in dimensions when reshaping previously. The backwards method now correctly propagates the gradients in the expected way, downsampling from the high resolution, low depth domain to the high depth, low resolution domain. 

Calling the backwards method on high resolution images now down-samples them in the expected manner. 

downsample

The left image was passed through the backwards method to give the expected 4 copies of down-sampled outputs (an example of which is seen on the right).

mattya:
Thank you for the update.
We tested the code and found that in python3.5 + GPU environment the argument of F.reshape function become float and crashed.
(numpy reshape function accepts float but cupy doesn't)

So could you use '//' instead of '/' and '//=' instead of '/=' ?

tjtorres:
OK, I made the proposed compatibility changes for the Python 3.5 + GPU env.

gwtnb:
A test with GPU failed:

```
FAIL: test_backward_gpu (test_depth_2_space.TestDepth2Space)
  File "/work/chainer/chainer/testing/condition.py", line 71, in wrapper
    fail()
  File "/work/chainer/chainer/testing/condition.py", line 51, in fail
    instance.fail(msg)
AssertionError:
Fail: 3, Success: 0

The first error message:
  File "/work/chainer/chainer/testing/condition.py", line 57, in <lambda>
    lambda: f(*args, **kwargs),
  File "/work/chainer/tests/chainer_tests/functions_tests/subpixel_tests/test_depth_2_space.py", line 90, in test_backward_gpu
    self.check_backward(cuda.to_gpu(self.rand_array))
  File "/work/chainer/tests/chainer_tests/functions_tests/subpixel_tests/test_depth_2_space.py", line 74, in check_backward
    y.grad = numpy.random.randn(*y.data.shape).astype(numpy.float32)
  File "/work/chainer/chainer/variable.py", line 192, in grad
    _check_grad_type(None, self, g)
  File "/work/chainer/chainer/variable.py", line 41, in _check_grad_type
    raise TypeError(make_message(msg))
TypeError: Type of data and grad mismatch
<class 'cupy.core.core.ndarray'> != <class 'numpy.ndarray'>
```

tjtorres:
Tested new commit on my GPU machine with this newest commit and all 4 tests are passing now.

chainer/functions/__init__.py
```diff
@@ -99,6 +99,7 @@
 from chainer.functions.pooling import roi_pooling_2d
 from chainer.functions.pooling import spatial_pyramid_pooling_2d
 from chainer.functions.pooling import unpooling_2d
+from chainer.functions.subpixel import depth2space as depth2space_
```
gwtnb: `_` suffix is unnecessary. https://github.com/tjtorres/chainer/blob/d665ea8c7047b63a858c26a690814194545c7ab6/chainer/functions/__init__.py#L121

gwtnb:
Could you tell us your intent to add a new directory `chainer/functions/subpixel`? Is `chainer/functions/array` inappropriate?

chainer/functions/__init__.py
tjtorres: It is for the current ordering, but I was trying to protect against possible future reordering by others where

```
depth2space = depth2space.depth2space
Depth2Space = depth2space.Depth2Space
```
would throw an error.

tjtorres:
I'm fine with this getting added into the `array` module, but it seemed a bit out of place when I was initially making it since all of the `array` functions appear to have functionalities with direct corollaries in NumPy whereas this does not.

mattya:
As you know, although the `array` module mainly focus on functions of numpy, it does not mean that they have to be implemented with numpy functions only.
Since this PR aims to shuffle the elements in an array, I recommend to add it on `array` module.

tjtorres:
Ok sounds good. I will move it over.

mattya:
LGTM!

PR #1749, opened 2016-10-07
Adds Subpixel Space2Depth Transformation Functionality

tjtorres:
This adds the inverse operation to Depth2Space called Space2Depth
and should be merged with the PR #1748 . 

This PR is a split of #1744 into its functional components.


gwtnb:
LGTM!

PR #1754, opened 2016-10-10
Support dropconnect

fukatani:
This PR is proposal for supporting dropconnect activation.
Dropconnect is generalization of dropout.
It drops weight element instead of input element.

Deteil is here.
http://cs.nyu.edu/~wanli/dropc/dropc.pdf


chainer/functions/connection/dropconnect.py
```diff
@@ -0,0 +1,100 @@
...
+        gW = gy.T.dot(x).astype(W.dtype, copy=False)
+        if len(inputs) == 3:
+            gb = gy.sum(0)
+            return gx, gW, gb
+        else:
+            return gx, gW
+
+
+def dropconnect(x, W, b=None, ratio=.5, train=True):
+    """Linear function using dropconnect.
```
delta2323: Could you add the reference to the paper to the document?

delta2323: To me, "linear function *with* dropconnect" seems more appropriate because "using dropconnect" sounds like implementation of Linear function made from dropconnect.

chainer/functions/connection/dropconnect.py
```diff
@@ -0,0 +1,100 @@
...
+
+    Dropconnect drops weight matrix elements randomly with probability ``ratio``
+    and scales the remaining elements by factor ``1 / (1 - ratio)``.
+    It accepts two or three arguments: an input minibatch ``x``, a weight
+    matrix ``W``, and optionally a bias vector ``b``. It computes
+    :math:`Y = xW^\\top + b`.
+
+    In testing mode, zero will be used as dropconnect ratio instead of
+    ``ratio``.
```
delta2323: As a confirmation, you are thinking you will implement testing mode, aren't you?

chainer/links/activation/dropconnect.py
```diff
@@ -0,0 +1,83 @@
...
+                 ratio=.5, initialW=None, initial_bias=None):
+        super(Dropconnect, self).__init__()
+
+        # For backward compatibility
+        self.initialW = initialW
+        self.wscale = wscale
+
+        self.out_size = out_size
+        self.ratio = ratio
+        # For backward compatibility, the scale of weights is proportional to
```
delta2323: `DropConnect` does not have backward compatibility to support. So this comment is not needed. As for implementation, I agree with this scale change because it is same behavior of `Convolution2D`

chainer/functions/connection/dropconnect.py
fukatani: Like `dropout`, `dropconnect` ratio have to become zero, when testing.
`test` is control flag for such behavior.
I implemented the testing mode in today's commit.

chainer/links/activation/dropconnect.py
fukatani: OK! I removed the comment.

chainer/functions/connection/dropconnect.py
fukatani: OK! I added reference to `links.dropconnect`.
In case of maxout, the document of `link.maxout` has reference, the document of `function.maxout` doesn't.

fukatani: Now I adopt "Linear unit regularized by dropconnect".
Or "Linear unit with dropconnect regularization" is more better?

fukatani:
Thanks for your comment :smiley: 
`link.dropconnect` test is failing, the document has not been completed and 
there is a need to improve the overall quality.

I'm going to get down when there is a time.

chainer/functions/connection/dropconnect.py
delta2323: OK, it is enough to add the reference to the link only.

delta2323: The former is OK.

delta2323:
Thank you for updating the implementation. I replied to your comments.

delta2323:
Could you add test cases in which `dropconnect` link work well with parameter shape placeholder (i.e. the first argument of the link is `None`)?

fukatani:
Thanks for your comment.
I added shape placeholder test and fixed bugs.

Especially, I struggled to reproduce mask for tests and to write docs.

chainer/links/activation/dropconnect.py
```diff
@@ -0,0 +1,94 @@
...
+        out_size (int): Dimension of output vectors.
+        wscale (float): Scaling factor of the weight matrix.
+        initialW (3-D array or None): Initial weight value.
+            If ``None``, then this function uses ``wscale`` to initialize.
+        initial_bias (2-D array, float or None): Initial bias value.
+            If it is float, initial bias is filled with this value.
+            If it is ``None``, bias is omitted.
+        reuse_mask (boolean):
+            If ``True``, once dropconnect mask is generated,
+            same mask will be reuse when link is called.
```
delta2323: So far as I know, users usually do not want `dropconnect` link to use same mask throughout the training. I would like to know if there is a usecase in which users want to fix the mask throughout whole training procedure. If there weren't, I think it is better not to add this option, considering the balance of the complexity of API and benefit for users.

On the other hand, I agree with `debug_mask` option in the `dropout` function because user may want to use this option for debugging.

So far as I check the implementation, `reuse_mask` is used to pass `check_backward` test, in which we call the link several times to calculate numerical gradient. So, if we could test the behavior of backward of the link, we may have to add this argument. Could you consider the feasibility of this proposal? One of my ideas is to set mask manually (or through the APIs of `dropconnect` function) in calculating the numerical grad.

chainer/links/activation/dropconnect.py
```diff
@@ -55,29 +55,40 @@ def __init__(self, in_size, out_size, wscale=1, bias=0, nobias=False,
...
             self._initialize_params(in_size)
 
-        if nobias:
-            self.b = None
-        else:
-            if initial_bias is None:
-                initial_bias = bias
-            bias_initializer = initializers._get_initializer(initial_bias)
-            self.add_param('b', out_size, initializer=bias_initializer)
+        bias_initializer = initializers._get_initializer(initial_bias)
```
fukatani: I did it!

chainer/links/activation/dropconnect.py
fukatani: As you said, `dropconnect` mask must not be fixed in normal use.
`reuse_mask` is only for paasing `check_backward`, but dirty, not preferable API.

Your idea's exactly what I wanted!

fukatani:
I responded to your comments.

chainer/functions/noise/dropconnect.py
```diff
@@ -0,0 +1,116 @@
...
+    if x.ndim == 2:
+        return x
+    return x.reshape(len(x), -1)
+
+
+class Dropconnect(function.Function):
+
+    """Linear unit regularized by dropconnect."""
+
+    def __init__(self, dropconnect_ratio, mask=None):
```
delta2323: The argument that specifies the ratio of dropping of connection if `dropconnect_ratio` whereas that of `dropconnect` function is `ratio`. Could you consider use the same option names?

chainer/functions/noise/dropconnect.py
```diff
@@ -0,0 +1,116 @@
...
+    ``ratio``.
+
+    Args:
+        x (chainer.Variable or :class:`numpy.ndarray` or cupy.ndarray):
+            Input variable. Its first dimension is assumed
+            to be the *minibatch dimension*. The other dimensions are treated
+            as concatenated one dimension whose size must be ``N``.
+        W (~chainer.Variable): Weight variable of shape ``(M, N)``.
+        b (~chainer.Variable): Bias variable (optional) of shape ``(M,)``.
+        ratio (float): Dropconnect ratio.
```
delta2323: If `mask` is not `None`, `ratio` is not used. I think we should add it to the document.

chainer/links/activation/dropconnect.py
```diff
@@ -0,0 +1,86 @@
...
+
+    .. seealso::
+        Li, W., Matthew Z., Sixin Z., Yann L., Rob F. (2013).
+        Regularization of Neural Network using DropConnect.
+        International Conference on Machine Learning.
+        `URL <http://cs.nyu.edu/~wanli/dropc/>`_
+    """
+
+    def __init__(self, in_size, out_size, wscale=1,
+                 ratio=.5, initialW=None, initial_bias=0, mask=None):
```
delta2323: I come to think there is another alternative of design that we add `mask` option to `__call__`, not `__init__`. The benefit of this design is that we can support the use case in which they use fixed but different masks during training (of course we can do it even if `mask` option is in `__init__`, but we need to change `self.mask` directory). On the other hand, it is more cumbersome because user have to specify `mask` at every iteration. According to @unnonouno , the problem looks similar to that in `where` function. As I do not have definitive reason for supporting either of them, I would like your opinion. Thank you.

fukatani: Initially, I implemented `mask` option to `__call__`. So the alternative is natural for me.

But I found it is difficult that setting `ndarray` type `mask` by `gradient_check.check_backward`.
Because `check_backward` convert `ndarray` to `Variable`.
And current version `functions.dropconnect` expects `mask` to be `ndarray`.

I know it is not problem if I change `dropconnect` to allow `Variable` type `mask`.
Add to `__call__` not `__init__` is better?

chainer/functions/noise/dropconnect.py
fukatani: It was in accordance with the `dropout`, but OK.
I did it.

chainer/functions/noise/dropconnect.py
fukatani: Sure :smiley:

fukatani:
Thanks. I replied to review!

chainer/links/activation/dropconnect.py
delta2323: I think it is OK for `dropconnect` to accept mask as a Variable because we can add `mask` to the computational graph. The possible drawbacks are (1) the overhead introduced by making computational graphs (2) that users need to wrap mask ndarray with Variable. But I think the overhead is not so problematic and wrapping of ndarray is not so cumbersome. If Variable mask does occur problem, we can change to ndarray as `mask` afterwards.

delta2323: One of the ways to realize this is to consider `mask` as an input Variable of `Dropconnect` Function as well as `x`, `W` and `b`. Could you consider the feasibility of this design?

fukatani: I agree with `Variable` type `mask` acceptance is preferable and less drawback.

fukatani: And I think both of `functions.dropconnect` and `links.dropconnect` mask option move from __init__ to__call__.
I appreciate if you give me any opinion.

delta2323: I agree with your design. Then, could you try to implement it?

fukatani:
I'm struggling to implement API.

To pass `check_backword`, all explicit input arguments have to be `Variable`.
But train is `bool`, so first, I implemented `__call__` as follows.

```python
    def __call__(self, x, mask=None, train=True):
```

This API enable us to test by `check_backward` as follows.
Because `train` is last argument and we don't have to input explicitly.
 
```python
    def check_backward(self, x_data, y_grad):
            self.link, (x_data, mask), y_grad, (self.link.W, self.link.b), eps=1e-2)
```

But user may think that prior argument is important, and posterior argument is optional.
Conversely, train is important, and mask is optional.

I appreciate if you have a good idea.
If you want, I can upload draft commit.

delta2323:
How about wrapping the link with a function?

```python
def f(data, mask):
    return self.link(data, True, mask)

def check_backward(self, x_data, y_grad):
        f, (x_data, mask), y_grad, (self.link.W, self.link.b), eps=1e-2)
```

fukatani:
I really enjoy discussion here :smiley: 

And I tried your idea.
But unfortunately, It seems not works well.

Here is a part of `gradient_check`.
```python
    for skip, x, cx in six.moves.zip(no_grads, xs, casted_xs):
        if skip:
            assert x.grad is None
            continue
        gx, = numerical_grad(f, (cx.data,), y_grad, eps=eps)
        testing.assert_allclose(gx, x.grad, atol=atol, rtol=rtol)
        if dtype is None:
            assert gx.dtype == x.grad.dtype
        else:
            assert gx.dtype.kind == 'f' and gx.dtype == dtype
```
Current `gradient_check` regards tuple of `Variable` as `data_x1` and `data_x2`.
And iterates test about `data_x1` and `data_x2` respectively.

But in our case, `func(*xs)` returns one array, and iterating about each data is impossible.
So it is difficult that we test `dropconnect` using `data_x` and `mask` by current API `gradient_check`.

One way is giving up and implement to set `mask` in `__init__` like my previous draft.

Another way is extending `gradient_check` API.
If similar tests is needed for another links, I'd like to implement it.
But if no other links, I think the extension of the API is somewhat exaggerated.

How about you?

delta2323:
I am not confident I understand your comment perfectly but you mean the problem is that `gradient_check` tries to calculate the numerical gradient with respect to `mask`, as well as `x_data`?

fukatani:
I overlooked `no_grads` option and I succeeded to implement test. Thanks!

tests/chainer_tests/functions_tests/noise_tests/test_dropconnect.py
```diff
@@ -0,0 +1,105 @@
...
+    @condition.retry(3)
+    def test_forward_cpu_nobias(self):
+        self.check_forward(self.x, self.W, None)
+
+    @attr.gpu
+    @condition.retry(3)
+    def test_forward_gpu(self):
+        self.check_forward(
+            cuda.to_gpu(self.x), cuda.to_gpu(self.W), cuda.to_gpu(self.b),
+            cuda.to_gpu(self.x.dot(self.W.T) + self.b))
```
delta2323: Too many argument. Maybe we should remove the last one.

tests/chainer_tests/functions_tests/noise_tests/test_dropconnect.py
```diff
@@ -0,0 +1,105 @@
...
+
+    @attr.gpu
+    @condition.retry(3)
+    def test_forward_gpu_nobias(self):
+        self.check_forward(
+            cuda.to_gpu(self.x), cuda.to_gpu(self.W), None,
+            cuda.to_gpu(self.x.dot(self.W.T)))
```
delta2323: ditto (remove the last argument)

tests/chainer_tests/functions_tests/noise_tests/test_dropconnect.py
```diff
@@ -0,0 +1,105 @@
...
+
+    def check_backward(self, x_data, W_data, b_data, y_grad):
+        args = (x_data, W_data)
```
delta2323: We can remove parentheses: `args = x_data, W_data`

fukatani:
Thanks for your attentive review.
Could you check GPU test results?

delta2323:
OK, I will push the code to the Jenkins.

delta2323:
Sorry for being late for letting you know the result of GPU tests. The following error occurred. Can you find out the cause of the problem?

```
14:46:24 ======================================================================
14:46:24 FAIL: test_backward_gpu (test_dropconnect.TestDropconnectParameterShapePlaceholder)
14:46:24 ----------------------------------------------------------------------nose.plugins.multiprocess: ERROR: Worker 2 STOPPED
14:46:24 NoneType
14:46:24 nose.plugins.multiprocess: ERROR: Worker 1 STOPPED
14:46:24 NoneType
14:46:24 nose.plugins.multiprocess: ERROR: Worker 3 STOPPED
14:46:24 NoneType
14:46:25 nose.plugins.multiprocess: ERROR: Worker 0 STOPPED
14:46:25 NoneType
14:47:41 
14:47:41 Traceback (most recent call last):
14:47:41   File "/work/chainer/chainer/testing/condition.py", line 71, in wrapper
14:47:41     fail()
14:47:41   File "/work/chainer/chainer/testing/condition.py", line 51, in fail
14:47:41     instance.fail(msg)
14:47:41 AssertionError: 
14:47:41 Fail: 3, Success: 0
14:47:41 
14:47:41 The first error message:
14:47:41 Traceback (most recent call last):
14:47:41   File "/work/chainer/chainer/testing/condition.py", line 57, in <lambda>
14:47:41     lambda: f(*args, **kwargs),
14:47:41   File "/work/chainer/tests/chainer_tests/links_tests/activation_tests/test_dropconnect.py", line 156, in test_backward_gpu
14:47:41     self.check_backward(cuda.to_gpu(self.x), cuda.to_gpu(self.gy))
14:47:41   File "/work/chainer/tests/chainer_tests/links_tests/activation_tests/test_dropconnect.py", line 146, in check_backward
14:47:41     (self.link.W, self.link.b), eps=1e-2, no_grads=(False, True))
14:47:41   File "/work/chainer/chainer/gradient_check.py", line 210, in check_backward
14:47:41     y = func(*xs)
14:47:41   File "/work/chainer/tests/chainer_tests/links_tests/activation_tests/test_dropconnect.py", line 141, in link_wrapper
14:47:41     return self.link(data[0], True, data[1])
14:47:41   File "/work/chainer/chainer/links/activation/dropconnect.py", line 86, in __call__
14:47:41     self.ratio, train, mask)
14:47:41   File "/work/chainer/chainer/functions/noise/dropconnect.py", line 121, in dropconnect
14:47:41     return Dropconnect(ratio, mask)(x, W, b)
14:47:41   File "/work/chainer/chainer/function.py", line 199, in __call__
14:47:41     outputs = self.forward(in_data)
14:47:41   File "/work/chainer/chainer/functions/noise/dropconnect.py", line 57, in forward
14:47:41     W = inputs[1] * self.mask
14:47:41   File "cupy/core/core.pyx", line 865, in cupy.core.core.ndarray.__mul__ (cupy/core/core.cpp:17251)
14:47:41     return multiply(x, y)
14:47:41   File "cupy/core/elementwise.pxi", line 731, in cupy.core.core.ufunc.__call__ (cupy/core/core.cpp:40840)
14:47:41     args = _preprocess_args(args)
14:47:41   File "cupy/core/elementwise.pxi", line 107, in cupy.core.core._preprocess_args (cupy/core/core.cpp:29538)
14:47:41     raise TypeError('Unsupported type %s' % type(arg))
14:47:41 TypeError: Unsupported type <class 'numpy.ndarray'>
```

delta2323:
As you see, CPU tests is failing. But the cause of the failure seems different from the change this PR made. We are now investigating the problem and will let you know how to fix the problem.

fukatani:
Thanks for GPU tests.
The reason why GPU tests fail is `self.mask` was not converted to `cupy`.

I use `add_persistent` to fix it.

delta2323:
Thank you for updating the code. I will run the GPU tests again.

delta2323:
Hi, thank you for updating the code. But unfortunately, it seems the same error remains. Is it possible to investigate the problem? Let me know if it is difficult to do so without GPUs.

fukatani:
I'm sorry for taking your time.

Now, I got a GPU environment.
I fixed tests and confirmed to pass GPU tests.

delta2323:
Ok, I will check it.

delta2323:
I confirmed that all GPU tests have passed successfully. I will check the code again and will merge if there are no problems.

chainer/functions/noise/dropconnect.py
```diff
@@ -0,0 +1,121 @@
...
+            )
+
+    def forward(self, inputs):
+        scale = inputs[1].dtype.type(1. / (1 - self.ratio))
+        xp = cuda.get_array_module(*inputs)
+        if xp == numpy:
+            flag = xp.random.rand(*inputs[1].shape) >= self.ratio
+        else:
+            flag = (xp.random.rand(*inputs[1].shape, dtype=numpy.float32) >=
+                    self.ratio)
```
delta2323: If self.mask is not `None`, we do not have to create `flag`  variable. So I think it is better to move it under `if self.mask is None`.

chainer/functions/noise/dropconnect.py
```diff
@@ -0,0 +1,121 @@
...
+                b_type.dtype == x_type.dtype,
+                b_type.ndim == 1,
+                b_type.shape[0] == w_type.shape[0],
+            )
+
+    def forward(self, inputs):
+        scale = inputs[1].dtype.type(1. / (1 - self.ratio))
+        xp = cuda.get_array_module(*inputs)
+        if xp == numpy:
+            flag = xp.random.rand(*inputs[1].shape) >= self.ratio
```
delta2323: Just for confirmation. The reason why we separate NumPy and CuPy case is to reduce the space size in CuPy case by using `float32` isn't it?

fukatani: In honest, I just referred to `Dropout`.

According to here, the purpose seems to be memory saving.
https://github.com/pfnet/chainer/pull/620

However, it seems that float16 can be good, and it may be also applicable to numpy.

fukatani: `numpy.random.rand` doesn't have `dtype` argument, so NumPy and CuPy cases are separated.

chainer/functions/noise/dropconnect.py
fukatani: Thanks. I did it.

delta2323:
@fukatani I realize that `mask` in `dropconnect` Function need not to be `Variable`, unless I miss something. As `isinstance` is costly, it is better if we can remove it in `forward` method [here](https://github.com/pfnet/chainer/pull/1754/files#diff-7f22745d81eead3103cbeb84c6962972R53). How do you think of it?

fukatani:
hmm...difficult
In honest, I'm not comfortable with it `isinstance`.

But it is a consistent interface in chainer that `__call__` can be input both `Variable` and `ndarray`.
And these arguments check by `isinstance` every `forward` call.

Another reason for allowing `Variable` is current `gradient_check.check_backword` permits only `Variable type`.

In addition, `mask` is `None` in normal case, so `isinstance` will not be executed.

delta2323:
I see. I could forget the discussion so far with you. So forgive me I could bring the same problem again. Do you remember why we did not choose to include `mask` as an argument of the function, as well as `x`, `W`, and `b`?

fukatani:
Don't mind. Actually, I think we should discuss about this topic too.

The reason why `mask` isn't argument of `forward` is as follows.
1)There are cases that we have to input only `mask` but not bias.
Inversely, input only bias but not `mask`.

We should identify these cases in `forward`.
For tht purpose, not only by using len(inputs), but also  by `ndarray.shape`
It's somewhat irregular.

2)`mask` is expected not to change by backward, backward returns zeros.

When doing the above, it will dirtier than the current pull request.
And current `mask` implementation resembles `function.MulConstant`, constant value of `function.MulConstant` is set in `__init__`.

delta2323:
I found the PR we are discussing is implemented differently from the [original one](http://cs.nyu.edu/%7Ewanli/dropc/dropc.pdf). Original dropconnect creates masks for each sample in a minibatch, whereas the PR uses single same mask for all samples. Also, the forward propagation in inference is different (see Algorithm 2 in the original paper). Unlike dropout-like inference, original paper uses approximation with the Gaussian distribution.

It seems to need sophisticated techniques to implement the original method (c.f. section 5) as naive implementation requires too much memory (proportional to minibatch size). So I guess it will take some additional time to reimplement to align the original one. 

As the PR is almost finished, I am thinking that we should complete the PR and create another PR that implements the original method if necessary. How do you think of it?

fukatani:
I really sorry for my mistake. :disappointed: 

I want to correct dropconnect in near future, current PR had better to be merged so far.
I wrote the differences with original paper in doc.

As original paper, can be saved memory by boolean `mask` ?

delta2323:
Please do not worry, I was at fault. 

> I wrote the differences with original paper in doc.

Thank you, could you write the difference between the implementation and the original paper and add notice that this implementation cannot be used for reproduction of the paper.

Also I think it is better to rename the names of the function and the link to reserve `Dropconnect` for future use (e.g. SimplifiedDropconnect). How do you think of it?

delta2323:
> As original paper, can be saved memory by boolean mask ?

Yes, it is better if possible. Could you try to change like that?

fukatani:
I decided to implement `dropconnect` as the paper.

GPU test failure remains.
It may be caused by `cupy.random` bug.
I'm investigating it now.

Changes:
1) Abolished scaling (as original implementation). 
Instead, `ratio` is multiplied at inference.

2) Sampling at inference is optional .
Default num_samples is `1000`  (as original implementation). 

3) Support only float32.
This is because current `matmul` is support only float32.
I sent PR (#2151).

Cons:
1) Memory saving is not perfect.
`mask` which is class attribute is boolean, 
But memory consumption by batch-sized array in the process of calculation still remains.
I don't consider it as a big problem.
e.g. There is a similar situation in `bilinear`.

fukatani:
I need #2169 for this PR.

delta2323:
Sorry for my late reply. So, you have decided not to make simplified version of DropConnect but to align the implementation with original one haven't you? I will check the implementation.

fukatani:
Don't worry.
Yes.
Actually, It is not difficult to implement batch-common-mask as option of Dropconnect`.

But I want to split PR, Since it is already big.

delta2323:
> But I want to split PR, Since it is already big.

 I agree. Could you do that?

fukatani:
OK! I'll do that.

fukatani:
I confirmed GPU tests at local. This PR can be reviewed.

delta2323:
Thank you for the update. I thought you would split this PR and send new smaller PRs because it ease our review. I would appreciate you if you would do so. If it is difficult, it is OK to remain this PR as it is.

delta2323:
Sorry for not working on this PR. Is it possible to splitting this PR into smaller ones.  I would appreciate you if you make a comment. Thank you in advance.

fukatani:
Sorry for late response.
I deleted commit about sampling inference. Is it OK?

delta2323:
Thank you for updating the code. I will take a look. Are you considering to send Sampling inference as another PR?

chainer/functions/noise/dropconnect.py
```diff
@@ -0,0 +1,140 @@
...
+    matrix ``W``, and optionally a bias vector ``b``. It computes
+    :math:`Y = xW^\\top + b`.
+
+    In testing mode, zero will be used as dropconnect ratio instead of
+    ``ratio``.
+
+    Notice:
+    This implementation cannot be used for reproduction of the paper.
+    There is a difference between the current implementation and the
+    original version dropconnect.
```
delta2323: Suggestion: ... the current implementation and the original one.

chainer/links/activation/dropconnect.py
```diff
@@ -0,0 +1,88 @@
...
+
+        Args:
+            x (chainer.Variable or :class:`numpy.ndarray` or cupy.ndarray):
+                Batch of input vectors.
+            train (bool):
+                If ``True``, executes dropconnect.
+                Otherwise, dropconnect link works as a linear unit.
+            mask (chainer.Variable or :class:`numpy.ndarray` or cupy.ndarray):
+                If ``None``, randomized dropconnect mask is generated.
+                If not ``None``, this value is used as a dropconnect mask.
```
delta2323: Suggestion: "If not `None`" -> "Otherwise"

chainer/links/activation/dropconnect.py
```diff
@@ -0,0 +1,88 @@
...
+                Scaling will not be executed.
```
delta2323: Please add "in this case" at the end of this sentence so that users do not wrongly understand that scaling is not done when if `mask` is `None`.

chainer/links/activation/dropconnect.py
```diff
@@ -0,0 +1,88 @@
...
+    def __call__(self, x, train=True, mask=None):
+        """Applies the dropconnect layer.
+
+        Args:
+            x (chainer.Variable or :class:`numpy.ndarray` or cupy.ndarray):
+                Batch of input vectors.
+            train (bool):
+                If ``True``, executes dropconnect.
+                Otherwise, dropconnect link works as a linear unit.
+            mask (chainer.Variable or :class:`numpy.ndarray` or cupy.ndarray):
```
delta2323: I think it is better to add `None` to the list of possible types of `mask` as `None` is default and expected value of this option.

chainer/links/activation/dropconnect.py
```diff
@@ -0,0 +1,88 @@
...
+                If ``None``, randomized dropconnect mask is generated.
+                If not ``None``, this value is used as a dropconnect mask.
+                Scaling will not be executed.
+                The mask shape must be ``(M, N)``.
+                Main purpose of the latter option is debugging.
```
delta2323: Although it is a problem of tastes, it is easier to read for me if we move this sentence after the sentence "If not ``None``, this value is used as a dropconnect mask." and change "latter option" to "this option".

chainer/links/activation/dropconnect.py
```diff
@@ -0,0 +1,88 @@
...
+            x (chainer.Variable or :class:`numpy.ndarray` or cupy.ndarray):
+                Batch of input vectors.
+            train (bool):
+                If ``True``, executes dropconnect.
+                Otherwise, dropconnect link works as a linear unit.
+            mask (chainer.Variable or :class:`numpy.ndarray` or cupy.ndarray):
+                If ``None``, randomized dropconnect mask is generated.
+                If not ``None``, this value is used as a dropconnect mask.
+                Scaling will not be executed.
+                The mask shape must be ``(M, N)``.
```
delta2323: As the shape of `mask` is `(M, N)`, we apply the same mask to every sample in the mini batch, if my understanding is correct. I would like to confirm it is your expected behavior.

fukatani: Your understanding is correct. I fixed here.

chainer/links/activation/dropconnect.py
fukatani: I changed implement and dropconnect always scale output.
Because I want to test scaling behavior.

So I deleted this sentence.

fukatani:
If the performance seems to rise with the original inference, I want to send PR, but as far as I implemented, the variance due to random sampling was large and the result was unstable.

I will rename this PR to `SimplifiedDropconnect`

Not-batchwise mask, as I said above, I will send another PR.

delta2323:
> I will rename this PR to SimplifiedDropconnect
OK, go ahead.

> Not-batchwise mask, as I said above, I will send another PR.
OK, I will check the PR.

delta2323:
As you can see, currently Travis tests of this PR fail. But this error seems independent of this PR as I encountered the same problem in another PR #2067. I am now investigating the problem.

fukatani:
Sure!
Basically, I confirmed failed contents, and I ignore them if it does not seem to be related.

And I renamed Dropconnect to SimplifiedDropconnect.

chainer/functions/noise/simplified_dropconnect.py
```diff
@@ -0,0 +1,144 @@
...
+            If ``True``, executes simplified dropconnect.
+            Otherwise, simplified dropconnect function works as a linear
+            function.
+        mask (None or chainer.Variable or :class:`numpy.ndarray` or
+            cupy.ndarray):
+            If ``None``, randomized dropconnect mask is generated.
+            Otherwise, The mask must be ``(n, M, N)`` shaped array.
+            Main purpose of this option is debugging.
+            `mask` array will be used as a dropconnect mask.
+
```
delta2323: Remove redundant empty line.

chainer/functions/noise/simplified_dropconnect.py
```diff
@@ -0,0 +1,144 @@
...
+
```
delta2323: Remove redundant empty line.

chainer/functions/noise/simplified_dropconnect.py
```diff
@@ -0,0 +1,144 @@
...
+def simplified_dropconnect(x, W, b=None, ratio=.5, train=True, mask=None):
+    """Linear unit regularized by simplified dropconnect.
+
+    Simplified dropconnect drops weight matrix elements randomly with
+    probability ``ratio`` and scales the remaining elements by factor
+    ``1 / (1 - ratio)``. It accepts two or three arguments: an input minibatch
+    ``x``, a weight matrix ``W``, and optionally a bias vector ``b``. It
+    computes :math:`Y = xW^\\top + b`.
+
```
delta2323: Please write that masks are generated for each sample in a minibatch (i.e. we use different mask for each sample).

chainer/functions/noise/simplified_dropconnect.py
```diff
@@ -0,0 +1,144 @@
...
+    In testing mode, zero will be used as simplified dropconnect ratio instead
+    of ``ratio``.
+
+    Notice:
+    This implementation cannot be used for reproduction of the paper.
```
delta2323: Please add the reference to this paper, as we did in `SimplifiedDropconnect`, too.

chainer/functions/noise/simplified_dropconnect.py
```diff
@@ -0,0 +1,144 @@
...
+    There is a difference between the current implementation and the
+    original one.
+    The original version uses sampling with gaussian distribution before
+    passing activation function, the current implementation averages
```
delta2323: add "whereas" after the comma.

chainer/links/activation/simplified_dropconnect.py
```diff
@@ -0,0 +1,90 @@
...
+            train (bool):
+                If ``True``, executes simplified dropconnect.
+                Otherwise, simplified dropconnect link works as a linear unit.
+            mask (None or chainer.Variable or :class:`numpy.ndarray` or
+                cupy.ndarray):
+                If ``None``, randomized simplified dropconnect mask is
+                generated. Otherwise, The mask must be ``(n, M, N)``
+                shaped array. Main purpose of this option is debugging.
+                `mask` array will be used as a dropconnect mask.
+
```
delta2323: Please add the notice as we did in `simplified_dropconnect`.

tests/chainer_tests/functions_tests/noise_tests/test_simplified_dropconnect.py
```diff
@@ -0,0 +1,105 @@
...
+        if b_data is None:
+            y = functions.simplified_dropconnect(x, W, None, self.ratio,
+                                                 self.train)
+        else:
+            b = chainer.Variable(b_data)
+            y = functions.simplified_dropconnect(x, W, b, self.ratio,
+                                                 self.train)
+        self.assertEqual(y.data.dtype, self.x_dtype)
+
+    @condition.retry(3)
```
delta2323: We do not have to retry because the main purpose of `retry` is to reduce unexpected failures caused by numerical instability.

tests/chainer_tests/functions_tests/noise_tests/test_simplified_dropconnect.py
```diff
@@ -0,0 +1,105 @@
...
+    def test_forward_cpu(self):
+        self.check_forward(self.x, self.W, self.b)
+
+    @condition.retry(3)
```
delta2323: ditto (remove `retry`)

tests/chainer_tests/functions_tests/noise_tests/test_simplified_dropconnect.py
```diff
@@ -0,0 +1,105 @@
...
+    def test_forward_cpu_nobias(self):
+        self.check_forward(self.x, self.W, None)
+
+    @attr.gpu
+    @condition.retry(3)
```
delta2323: ditto (remove `retry`)

tests/chainer_tests/functions_tests/noise_tests/test_simplified_dropconnect.py
```diff
@@ -0,0 +1,105 @@
...
+    def test_forward_gpu(self):
+        self.check_forward(
+            cuda.to_gpu(self.x), cuda.to_gpu(self.W), cuda.to_gpu(self.b))
+
+    @attr.gpu
+    @condition.retry(3)
```
delta2323: ditto (remove `retry`)

tests/chainer_tests/links_tests/activation_tests/test_simplified_dropconnect.py
```diff
@@ -0,0 +1,184 @@
...
+
+    def setUp(self):
+        in_size = numpy.prod(self.in_shape)
+
+        self.link = links.SimplifiedDropconnect(
+            in_size, self.out_size,
+            initialW=chainer.initializers.Normal(1, self.W_dtype),
+            initial_bias=chainer.initializers.Normal(1, self.x_dtype))
+        self.link.cleargrads()
+        self.mask = gen_mask(self.ratio, self.link.W.shape)
```
delta2323: According to the specification written in the document, the shape of a mask should be `(n, M, N)`, where as this test actually creates a mask of `(M, N)`. So, please fix the shape of the mask.

delta2323: This test unexpectedly works because broadcast occurred and the same mask is used.

tests/chainer_tests/links_tests/activation_tests/test_simplified_dropconnect.py
```diff
@@ -0,0 +1,184 @@
...
+        temp_x = numpy.random.uniform(-1, 1,
+                                      (self.out_size,
+                                       self.in_size)).astype(numpy.float32)
+        self.link(chainer.Variable(temp_x))
+        W = self.link.W.data
+        W[...] = numpy.random.uniform(-1, 1, W.shape)
+        b = self.link.b.data
+        b[...] = numpy.random.uniform(-1, 1, b.shape)
+        self.link.cleargrads()
+        self.mask = gen_mask(self.ratio, self.link.W.shape)
```
delta2323: ditto (create sample-wise mask).

tests/chainer_tests/links_tests/activation_tests/test_simplified_dropconnect.py
```diff
@@ -0,0 +1,184 @@
...
+
+        x_shape = (4,) + self.in_shape
+        self.x = numpy.random.uniform(-1, 1, x_shape).astype(self.x_dtype)
+        self.gy = numpy.random.uniform(
+            -1, 1, (4, self.out_size)).astype(self.x_dtype)
+        W = self.link.W.data
+        b = self.link.b.data
+        self.y_expect = self.x.reshape(4, -1).dot(W.T * self.mask.T) + b
```
delta2323: We need to change how to calculate `y_expect` as we do `mask`.

tests/chainer_tests/links_tests/activation_tests/test_simplified_dropconnect.py
```diff
@@ -0,0 +1,184 @@
...
+        b = self.link.b.data
+        b[...] = numpy.random.uniform(-1, 1, b.shape)
+        self.link.cleargrads()
+        self.mask = gen_mask(self.ratio, self.link.W.shape)
+
+        x_shape = (4,) + self.in_shape
+        self.x = numpy.random.uniform(-1, 1, x_shape).astype(numpy.float32)
+        self.gy = numpy.random.uniform(
+            -1, 1, (4, self.out_size)).astype(numpy.float32)
+        self.y_expect = self.x.reshape(4, -1).dot(W.T * self.mask.T) + b
```
delta2323: ditto (fix how to calculate `y_expect` appropriately)

fukatani:
Thanks! I fixed them.

I also fixed bug on gen_mask at test.
It is irrelevant to the result of the test, but the shape of `temp_x` was also inappropriate and I fixed it.

delta2323:
Thank you for your change. I will check it.

delta2323:
As you can see, there occurs the following errors in Travis. This is because old NumPy (< 1.10) does not have `matmul` (see [reference](https://docs.scipy.org/doc/numpy/reference/generated/numpy.matmul.html)). As Chainer support NumPy older than 1.10, we should make this function work with it. We have several choices:

* Implement this function without `numpy.matmul`
* Check the version of NumPy and switch the implementation depending on the version.
* Check the version of NumPy and raise an error if NumPy is older than 1.10.

Could you consider the implement this function without `matmul`? If it is difficult, I think the workaround is to take the third option and support older NumPy in another PR.

```
ERROR: test_forward_cpu (test_simplified_dropconnect.TestSimplifiedDropconnect_param_0)  parameter: {'train': True, 'ratio': 0.0, 'W_dtype': <type 'numpy.float16'>, 'x_dtype': <type 'numpy.float16'>}
  File "/home/travis/build/pfnet/chainer/tests/chainer_tests/functions_tests/noise_tests/test_simplified_dropconnect.py", line 56, in test_forward_cpu
    self.check_forward(self.x, self.W, self.b)
  File "/home/travis/build/pfnet/chainer/tests/chainer_tests/functions_tests/noise_tests/test_simplified_dropconnect.py", line 52, in check_forward
    self.train)
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/chainer/functions/noise/simplified_dropconnect.py", line 148, in simplified_dropconnect
    return SimplifiedDropconnect(ratio, mask)(x, W, b)
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/chainer/function.py", line 199, in __call__
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/chainer/functions/noise/simplified_dropconnect.py", line 61, in forward
    y = xp.matmul(W, x[:, :, None])
AttributeError: 'module' object has no attribute 'matmul'
```

fukatani:
As #2067, I use `numpy.einsum` instead of `numpy.matmul`.

delta2323:
@fukatani Thank you! As we use `matmul` in the test code, could you fix it in the same way?

fukatani:
Sorry, I fixed test.

delta2323:
LGTM, I will run the GPU test.

delta2323:

PR #1767, opened 2016-10-14
Fix BN problem, it works with copy().

t-abe:
In current BN implementation, it assign new values to avg_mean and avg_var.
It causes a problem when you use copy() to make a new model for validation (as ImageNet example does).


rezoo:
LGTM, and I think this PR is important and should be merged asap.

PR #1774, opened 2016-10-15
Fix fan_out that is different from other frameworks.

dsanno:
I fixed `fan_out` in initializer.py that is different from other frameworks.

I think `get_fans` is based on keras implementation and keras used `shape[0]` as `fan_out`, but it is fixed in https://github.com/fchollet/keras/pull/2252 
Other implementation of `fan_out`:
- Caffe  
  https://github.com/BVLC/caffe/blob/master/include/caffe/filler.hpp#L193
- Tensorflow  
  https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/layers/python/layers/initializers.py#L126

### DIFF
--- a/chainer/initializer.py
+++ b/chainer/initializer.py
@@ -41,6 +41,7 @@ def get_fans(shape):
     if len(shape) < 2:
         raise ValueError('shape must be of length >= 2: shape={}', shape)
 
-    fan_in = numpy.prod(shape[1:])
-    fan_out = shape[0]
+    receptive_field_size = numpy.prod(shape[2:], dtype=numpy.int32)
+    fan_in = shape[1] * receptive_field_size
+    fan_out = shape[0] * receptive_field_size
     return fan_in, fan_out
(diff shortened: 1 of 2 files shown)


delta2323:
Thank you for reporting the problem. As fchollet mentioned in the original PR, this change can affect the existing training process. So could you have time to think of it?

delta2323:
One option I come up with is to adopt this change in the next major release.

delta2323:
@beam2d Could you check this PR?

beam2d:
Thank you for the proposal. I'm wondering what should be the definition of _fan out_. Should the equation `fan_out = shape[0] * receptive_field_size` be the definition? A popular explanation would be the number of output edges from each input unit, though it does not match to the equation for Convolution layer with a stride larger than one. I think it's reasonable to align the behavior with other frameworks and I don't mean I'm against this PR, but I'm just curious about this point.

dsanno:
As mentioned in fchollet/keras#2252, it is described in the paper from He et al (http://arxiv.org/abs/1502.01852).
I think `\hat{n} = k^2 d` after the equation (14) is the definition of _fan out_. (_k_ is kernel size and _d_ is output channel size.)
Please see the paper for details.

delta2323:
I agree with merging this PR to v2.

PR #1779, opened 2016-10-17
fix scale of elapsed_time for GPU case

ywatanabex:
This is a bug fix.
Since the returned value is in millisecond scale, it must be divided by 1000 to change second scale.


delta2323:
Thank you for reporting the bug. I will merge after releasing 1.17.0 because code is frozen for today's release.

delta2323:
Thank you for your contribution. By the way, do you think we should add feature that show the elapsed time in millisecond (not second)?

ywatanabex:
I don't think we need that kind of option. Because we can just multiply 1000 to convert it in millisecond scale.

PR #1781, opened 2016-10-18
Add norm function in cupy.linalg.

ywatanabex:
The function can be used mostly the same as numpy version, but
complex value and nuclear norms are not supported.


ywatanabex:
This feature is requested in #677

ywatanabex:
The implemented norm function is not faster than cupy version. It mainly because the sum function of cupy is not first.
Here is an example:

```python
>>> import numpy, cupy, time
>>> x = numpy.random.random((100000000)).astype(numpy.float32)
>>> y = cupy.array(x)
>>> start=time.time();  x.sum(); end=time.time(); print(end-start)
0.0444009304047
>>> start=time.time();  y.sum(); end=time.time(); print(end-start)
0.0568418502808
```

We can improve the speed of sum on GPU by 

```
>>> y0 = y.reshape((100000, 1000))  
>>> start=time.time();  y0.sum(axis=1).sum(axis=0); end=time.time(); print(end-start)
0.0075089931488
```

but this is tricky and depends on the size of input.
Is there a good way to execute summation first?

okuta:
Please split commits.
* First, only rename commit
* Next, modify commit

okuta:
I checked `cupy.sum`. I confirmed a tricky result.
My test code.
https://gist.github.com/okuta/1387d563024c8a5364358212babd2cbf


It is recommended that you use normal sum operation.
We need improve `cupy.sum`.


And, please use `cupy.Event` to profile cupy code.
`cupy.sum` is async function. If you use `time.time()`, you get incorrect results.

```python
import chainer  # for using memory pool
import cupy


start = cupy.cuda.Event()
end = cupy.cuda.Event()
start.record()

# target code
for i in range(1000):
     y0.sum(axis=1).sum(axis=0)

end.record()
end.synchronize()
print(cupy.cuda.get_elapsed_time(start, end))
```

ywatanabex:
Hi, thank you for your advices.
I split the commit into two commits.

cupy/linalg/norms.py
```diff
@@ -0,0 +1,151 @@
...
+    Complex valued matrices and vectors are not supported.
+    See numpy.linalg.norm for more detail.
+
+    Args:
+        x (cupy.ndarray): Array to take norm. If ``axis`` is None,
+            ``x`` must be 1-D or 2-D.
+        ord (non-zero int, inf, -inf, 'fro'): Norm type.
+        axis (int, 2-tuple of ints, None): 1-D or 2-D norm is cumputed over
+            ``axis``.
+        keepdims (bool): If this is set True, the axes which are normed over
```
okuta: Please change to ```If this is set ``True``,``` .

cupy/linalg/norms.py
```diff
@@ -0,0 +1,151 @@
...
+        axis = (axis,)
+
+    if len(axis) == 1:
+        if ord == numpy.Inf:
+            return abs(x).max(axis=axis, keepdims=keepdims)
+        elif ord == -numpy.Inf:
+            return abs(x).min(axis=axis, keepdims=keepdims)
+        elif ord == 0:
+            # Zero norm
+            return (x != 0).astype(x.dtype).sum(axis=axis, keepdims=keepdims)
```
okuta: How about `(x != 0).sum(axis=axis, keepdims=keepdims, dtype=x.dtype)`?

cupy/linalg/norms.py
```diff
@@ -0,0 +1,151 @@
...
+        elif ord == 1:
+            # special case for speedup
+            return abs(x).sum(axis=axis, keepdims=keepdims)
+        elif ord is None or ord == 2:
+            # special case for speedup
+            s = x**2
+            return cupy.sqrt(s.sum(axis=axis, keepdims=keepdims))
+        else:
+            try:
```
okuta: How about using `float(ord)`?

tests/cupy_tests/linalg_tests/test_norms.py
```diff
@@ -0,0 +1,57 @@
...
+    'keepdims': [True, False],
+})
+)
+@testing.gpu
+class TestNorm(unittest.TestCase):
+
+    _multiprocess_can_split_ = True
+
+    @testing.for_all_dtypes()
+    @testing.numpy_cupy_allclose(rtol=1e-3, atol=1e-4, type_check=False)
```
okuta: Why using `type_check=False`?

ywatanabex: Given `float32` array as an in input, numpy norm function returns sometimes `flaot32` and sometimes `float64`. For example,

```python
>>> import numpy
>>> x = numpy.array([1.1, 1.2], dtype=numpy.float32)
>>> numpy.linalg.norm(x, ord=2).dtype
dtype('float32')
>>> numpy.linalg.norm(x, ord=3).dtype
```

ywatanabex:
Sorry, I happen to delete the branch. I made a new PR #1833

PR #1795, opened 2016-10-19
Add seed option to split functions

rezoo:
This PR aims to add ``seed`` option to ``split_dataset_random()`` and ``get_cross_validation_datasets_random()`` functions.
It would be useful when you split a dataset to train and test datasets, train a NN with the train dataset, and test the trained NN with the test dataset by another script.

### DIFF
--- a/chainer/datasets/sub_dataset.py
+++ b/chainer/datasets/sub_dataset.py
@@ -116,7 +116,8 @@ def split_dataset_random(dataset, first_size, seed=None):
         dataset: Dataset to split.
         first_size (int): Size of the first subset.
         seed (int): Seed the generator used for the permutation of indexes.
-            If an integer is specified, it is guaranteed that each sample
+            If an integer beging convertible to 32 bit unsigned integers is
+            specified, it is guaranteed that each sample
             in the given dataset always belongs to a specific subset.
             If ``None``, the permutation is changed randomly.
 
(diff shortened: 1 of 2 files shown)


chainer/datasets/sub_dataset.py
```diff
@@ -115,6 +115,10 @@ def split_dataset_random(dataset, first_size):
     Args:
         dataset: Dataset to split.
         first_size (int): Size of the first subset.
+        seed (int): Seed the generator used for the permutation of indexes.
+            If an integer is specified, it is guaranteed that each sample
+            in the given dataset always belongs to a specific subset.
+            If ``None``, the permutation is changed randomly.
```
muupan: It is better to note that `seed` must be 32 bit unsigned integers (or None), otherwise numpy will return `ValueError`.

See: https://docs.scipy.org/doc/numpy/reference/generated/numpy.random.seed.html

chainer/datasets/sub_dataset.py
```diff
@@ -176,10 +182,16 @@ def get_cross_validation_datasets_random(dataset, n_fold):
     Args:
         dataset: Dataset to split.
         n_fold (int): Number of splits for cross validation.
+        seed (int): Seed the generator used for the permutation of indexes.
+            If an integer is specified, it is guaranteed that each sample
+            in the given dataset always belongs to a specific subset.
+            If ``None``, the permutation is changed randomly.
```
muupan: Ditto.

chainer/datasets/sub_dataset.py
rezoo: Done.

chainer/datasets/sub_dataset.py
```diff
@@ -123,7 +127,9 @@ def split_dataset_random(dataset, first_size):
             dataset.
 
     """
+    numpy.random.seed(seed)
     order = numpy.random.permutation(len(dataset))
+    numpy.random.seed(None)
```
rezoo: That's a good idea. I replaced it.

rezoo:
I fixed it based on the comments. Can you check this one?

muupan:
LGTM. I'll merge it.

PR #1798, opened 2016-10-20
fix broken link of convolution_2d and convolution_nd document.

mmurooka:
links of "See also" part in following documents are broken
http://docs.chainer.org/en/stable/reference/functions.html#convolution-2d
http://docs.chainer.org/en/stable/reference/functions.html#convolution-nd

### DIFF
--- a/chainer/functions/connection/convolution_2d.py
+++ b/chainer/functions/connection/convolution_2d.py
@@ -308,7 +308,7 @@ def convolution_2d(x, W, b=None, stride=1, pad=0, use_cudnn=True,
     If the bias vector is given, then it is added to all spatial locations of
     the output of convolution.
 
-    .. seealso:: :class:`Convolution2D`
+    .. seealso:: :class:`~chainer.links.Convolution2D`
 
     """
     func = Convolution2DFunction(stride, pad, use_cudnn, cover_all)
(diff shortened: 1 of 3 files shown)


chainer/functions/connection/convolution_2d.py
```diff
@@ -308,7 +308,7 @@ def convolution_2d(x, W, b=None, stride=1, pad=0, use_cudnn=True,
     If the bias vector is given, then it is added to all spatial locations of
     the output of convolution.
 
-    .. seealso:: :class:`Convolution2D`
+    .. seealso:: :class:`chainer.links.Convolution2D`
```
delta2323: Could you change `~chainer.links.Convolution2D` (i.e. add tilde at the head of the class name). It hides the namespace `chainer.links` from the display name.

chainer/functions/connection/convolution_nd.py
```diff
@@ -342,7 +342,7 @@ def convolution_nd(x, W, b=None, stride=1, pad=0, use_cudnn=True,
       than v3)
 
 
-    .. seealso:: :class:`ConvolutionND`, :func:`convolution_2d`
+    .. seealso:: :class:`chainer.links.ConvolutionND`, :func:`convolution_2d`
```
delta2323: ditto (add tilde).

delta2323:

PR #1809, opened 2016-10-24
Set zero as grad in initialize of using placeholder parameter

### LINKED ISSUE #1799: I got NaN when only using shape placeholder.
I found that a gradient in link is not updated when using shape placeholder.
An example is as follows. You will get NaN in the last print function.:
```python
import chainer
from chainer import Variable
import numpy as np

l = L.Linear(None, 100)

x = Variable(np.zeros((10, 20)).astype("f"))

o = chainer.optimizers.SGD()

o.setup(l)
l.cleargrads()
y = l(x)
print(l.W.data.shape)

y.grad = np.zeros_like(y.data) + 1.0
y.backward(retain_grad=True)

print(l.W.data)
o.update()
print(l.W.data)
```
I tested on the recent chainer. I suspect that it may arise from the reference  of the optimizer but I don't have any detailed specifications. Do you have any ideas?

okuta:
Fix #1799
This bug is critical.
If we use shape placeholder, NN learning will fail. It is difficult for users to notice this bug.

But, this PC is like a work-around. Please tell me about good design.

### DIFF
--- a/tests/chainer_tests/optimizers_tests/test_optimizers_by_linear_model.py
+++ b/tests/chainer_tests/optimizers_tests/test_optimizers_by_linear_model.py
@@ -6,6 +6,7 @@ import six
 import chainer
 from chainer import cuda
 import chainer.functions as F
+from chainer import initializers
 import chainer.links as L
 from chainer import optimizers
 from chainer import testing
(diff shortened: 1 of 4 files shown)


tests/chainer_tests/optimizers_tests/test_optimizers_by_linear_model.py
```diff
@@ -189,7 +203,8 @@ def create(self):
 
 
 @testing.parameterize(*testing.product({
-    'dtype': [numpy.float16, numpy.float32, numpy.float64]
+    'dtype': [numpy.float16, numpy.float32, numpy.float64],
+    'use_placeholder': [True, False],
```
delta2323: Why we test placeholders for some specific optimizers only?

delta2323:
Could you add the failed case in #1799 so that the issue is surely resolved.

delta2323:
> Please tell me about good design.

One alternative is to call cleargrad (or diractly fill grad attribute with nan).

delta2323:
@okuta Should we close this PR and work on #1811?

okuta:
Yes, Please see #1811 .

PR #1810, opened 2016-10-24
Fix link placeholder problem caused by ndarray input.

dsanno:
I fixed the shape placeholder problem caused when ndarray is input.
It occurs in `Linear`, `StatelessLSTM` and `LSTM` links.
This problem caused by `len(x.data)` used for input size calculation.
When `x` is ndarray `len(x.data)` is buffer size of `x`.

Here is an example to cause the problem: (on Chainer v1.17.0)
```
>>> from chainer import links as L
>>> import numpy as np
>>> fc = L.Linear(None, 10)
>>> x = np.zeros((10, 10), dtype=np.float32)
>>> y = fc(x)
  File "chainer/links/connection/linear.py", line 91, in __call__
    self._initialize_params(x.size // len(x.data))
  File "chainer/links/connection/linear.py", line 77, in _initialize_params
    initializer=self._W_initializer)
  File "chainer/link.py", line 180, in add_param
    data = initializers.generate_array(initializer, shape, self.xp)
  File "chainer/initializers/__init__.py", line 45, in generate_array
    initializer(array)
  File "chainer/initializers/normal.py", line 93, in __call__
    s = self.scale * numpy.sqrt(2. / fan_in)
ZeroDivisionError: float division by zero
```

### DIFF
--- a/chainer/links/connection/linear.py
+++ b/chainer/links/connection/linear.py
@@ -88,5 +88,5 @@ class Linear(link.Link):
         """
         if self.has_uninitialized_params:
             with cuda.get_device(self._device_id):
-                self._initialize_params(x.size // len(x.data))
+                self._initialize_params(x.size // x.shape[0])
         return linear.linear(x, self.W, self.b)
--- a/chainer/links/connection/lstm.py
+++ b/chainer/links/connection/lstm.py
@@ -82,7 +82,7 @@ class StatelessLSTM(LSTMBase):
 
         """
         if self.upward.has_uninitialized_params:
-            in_size = x.size // len(x.data)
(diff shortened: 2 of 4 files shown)


gwtnb:
This problem appears on Python 2.

gwtnb:
Thank you!

gwtnb:
Python 2:
```python
>>> import numpy
>>> x = numpy.ndarray((2, 3), dtype=numpy.float32)
>>> len(x.data)
24
```

Python 3:
```python
>>> import numpy
>>> x = numpy.ndarray((2, 3), dtype=numpy.float32)
>>> len(x.data)
2
```

PR #1814, opened 2016-10-25
Add forward and backward test to dropout

tohmae:
This PR adds tests to dropout function.
There is only typecheck test in dropout function.
I add forward and backward tests to dropout test.
I think that these tests may not be enough but are better than nothing.

### DIFF
--- a/tests/chainer_tests/functions_tests/noise_tests/test_dropout.py
+++ b/tests/chainer_tests/functions_tests/noise_tests/test_dropout.py
@@ -49,12 +49,11 @@ class TestDropout(unittest.TestCase):
     def check_backward(self, x_data, y_grad):
         x = chainer.Variable(x_data)
         y = functions.dropout(x, self.ratio)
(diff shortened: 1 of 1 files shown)


tests/chainer_tests/functions_tests/noise_tests/test_dropout.py
```diff
@@ -5,31 +5,84 @@
...
+        if self.ratio == 0.0:
+            y_expect = x_data
+        else:
+            y_expect = _dropout(x_data, y.creator)
+        testing.assert_allclose(y_expect, y.data)
+
+    def check_backward(self, x_data, y_grad):
+        x = chainer.Variable(x_data)
+        y = functions.dropout(x, self.ratio)
+        d = {'creator': y.creator}
```
okuta: How about `creater = y.creater`?

tohmae:
Thank you for your comment.

okuta:
LGTM!

PR #1832, opened 2016-11-03
Add integer array indexing on one axis for __getitem__

yuyu2172:
The PR change `ndarray.__getitem__` to accept `ndarray` or `numpy.ndarray` in the argument. It checks for the `dtype` of arrays in the argument, and raise Error when necessary. Currently, only integer arrays are supported. 

The modification simply integrates `take` and `__getitem__`. With this PR, `a[:, idx]` returns same value as `a.take(idx, axis=1)`.

### DIFF
--- a/cupy/core/core.pyx
+++ b/cupy/core/core.pyx
@@ -1017,22 +1017,6 @@ cdef class ndarray:
         else:
             slices = list(slices)
 
-        # Check if advanced is true and if there are multiple integer indexing
-        advanced = False
-        axis = None
-        for i, s in enumerate(slices):
-            if isinstance(s, (numpy.ndarray, ndarray)):
-                if issubclass(s.dtype.type, numpy.integer):
-                    if advanced:
-                        advanced = True
-                        axis = None
-                    else:
-                        advanced = True
-                        axis = i
-                else:
-                    raise ValueError('Advanced indexing with ' +
-                                     'non-integer array is not supported')
-
         # Expand ellipsis into empty slices
         ellipsis = -1
         n_newaxes = n_ellipses = 0
@@ -1051,6 +1035,22 @@ cdef class ndarray:
             slices[ellipsis:ellipsis + 1] = noneslices * ellipsis_size
 
         slices += noneslices * (ndim - <Py_ssize_t>len(slices) + n_newaxes)
+
+        # Check if advanced is true and if there are multiple integer indexing
+        advanced = False
+        axis = None
+        for i, s in enumerate(slices):
+            if isinstance(s, (numpy.ndarray, ndarray)):
+                if issubclass(s.dtype.type, numpy.integer):
+                    if advanced:
+                        advanced = True
+                        axis = None
+                    else:
+                        advanced = True
+                        axis = i
+                else:
+                    raise ValueError('Advanced indexing with ' +
+                                     'non-integer array is not supported')
 
         if advanced:
             if axis is not None:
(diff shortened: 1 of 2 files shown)


cupy/core/core.pyx
```diff
@@ -1016,17 +1017,30 @@ cdef class ndarray:
...
 
-        for s in slices:
-            if isinstance(s, ndarray):
-                raise ValueError('Advanced indexing is not supported')
+        # Check if advanced is true and if there are multiple integer indexing
+        advanced = False
+        axis = None
+        for i, s in enumerate(slices):
+            if isinstance(s, (numpy.ndarray, ndarray)):
+                if s.dtype in [numpy.int64, numpy.int32,
```
okuta: How about using `issubclass(s.dtype.type, numpy.integer)`?

cupy/core/core.pyx
```diff
@@ -1039,52 +1053,64 @@ cdef class ndarray:
...
-                if ind < 0:
-                    ind += self._shape[j]
-                if not (0 <= ind < self._shape[j]):
-                    msg = ('Index %s is out of bounds for axis %s with size %s'
-                           % (s, j, self._shape[j]))
-                    raise IndexError(msg)
-                offset += ind * self._strides[j]
-                j += 1
+        if advanced:
+            if (axis is not None and
```
okuta: How about following code?
```
if axis is not None:
    noneSlice = slice(None)
    flag = True
    for i, s in enumerate(slices):
        if i != axis and s != slice:
            flag = False
            break
    if flag:
        return self.take(slices[axis], axis)
raise NotImplementedError(
    'Advanced indexing with 2 or more arrays is not supported')
```

cupy/core/core.pyx
```diff
@@ -1039,52 +1053,64 @@ cdef class ndarray:
...
             else:
-                raise TypeError('Invalid index type: %s' % type(slices[i]))
+                raise NotImplementedError('Adv indexing with 2 or ' +
+                                          'more arrays is not supported')
 
-        v = self.view()
-        v.data = self.data + offset
-        v._set_shape_and_strides(shape, strides)
-        return v
+        else:
```
okuta: Please remove this `else`.

yuyu2172:
Sorry. I pushed rebased code, and the commit log got messed up.
The first three commits (95590d8 cf8a700 81223b1) were reviewed.

Some additional commits (c5900aa, 94df0b3, fa8b1c5) are all about handling special cases.

cupy/core/core.pyx
yuyu2172: Thank you!

cupy/core/core.pyx
yuyu2172: That is much better.

cupy/core/core.pyx
yuyu2172: Thank you!

cupy/core/core.pyx
```diff
@@ -1039,6 +1050,42 @@ cdef class ndarray:
...
+        # Check if advanced is true and if there are multiple integer indexing
+        advanced = False
+        axis = None
+        for i, s in enumerate(slices):
+            if isinstance(s, list):
+                s = numpy.array(s)
+                slices[i] = s
+            if isinstance(s, (numpy.ndarray, ndarray)):
+                if issubclass(s.dtype.type, numpy.integer):
+                    if advanced:
```
yuyu2172: Thank you for commenting.

`advanced` is `True` when there are two or more arrays in `slices`.
In that case, `NotImplementedError` at line 1086 will be raised.

okuta: Oh, sorry. This code is in the loop. I understood.

okuta:
LGTM!

PR #1837, opened 2016-11-04
Fix multi gpu mnist exmaple

okuta:
This PR selects default device to fix memory allocation bug.

This problem reported in Japanese forum.
https://groups.google.com/forum/#!msg/chainer-jp/RvnEJ_mtM3U/KlplzOdFCAAJ


gwtnb:
LGTM

PR #1844, opened 2016-11-08
Add cupy put

yuyu2172:
Merge after #1855 

This PR adds `cupy.put` that functions similarly to `numpy.put`, but different in the following way.
+ `numpy.put` handles values sequentially, whereas `cupy.put` does not. 

Here is an example that illustrates the difference.

```python
>>> a = xp.zeros((2,))

>>> i = xp.arange(10000) % 2
>>> v = xp.arange(10000).astype(np.float)

>>> xp.put(a, i, v)
```

On numpy, whenever there is duplicating indices, the index that appeared last is used to store a value.
```python
>>> print a  # when xp is numpy
[ 9998. 9999.]
```

Whereas, in my implementation of `cupy.put`, the duplicating indices do not get handled as Numpy does. Instead, the index that gets handled by the last called kernel thread that stores values.
```python
>>> print a  # when xp is cupy
[ 9982. 9983.]
```

This can be changed to be compatible to Numpy. However, it requires an extra kernel call to look for duplicates, which I am not sure if worth it.
I would like to hear feedback. Thanks.

##### UPDATE:
I decided to leave CuPy different from NumPy with a index of duplicating values for performance reason.

### DIFF
--- a/cupy/core/core.pyx
+++ b/cupy/core/core.pyx
@@ -599,7 +599,7 @@ cdef class ndarray:
 
         .. seealso::
            :func:`cupy.put` for full documentation,
-           :meth: `numpy.ndarray.put`
+           :meth:`numpy.ndarray.put`
 
         """
         _put(self, ind, v, axis=None)
@@ -1987,10 +1987,8 @@ cpdef _put(ndarray a, ind, v, axis=None):
         (1,) * len(lshape) + ind.shape + (1,) * len(rshape))
     if axis == 0 or axis is None:
         _put_kernel_0axis(v, ind, rdim, index_range, a.reduced_view())
-        return
     else:
         _put_kernel(v, ind, cdim, rdim, adim, index_range, a.reduced_view())
-        return
 
 
 cpdef ndarray _diagonal(ndarray a, Py_ssize_t offset=0, Py_ssize_t axis1=0,
--- a/cupy/core/core.pyx
+++ b/cupy/core/core.pyx
@@ -594,7 +594,16 @@ cdef class ndarray:
         """
         return _take(self, indices, axis, out)
 
-    # TODO(okuta): Implement put
+    cpdef put(self, ind, v, axis=None):
+        """Replaces specified elements of an array with given values.
+
+        .. seealso::
+           :func:`cupy.put` for full documentation,
+           :meth: `numpy.ndarray.put`
+
+        """
+        _put(self, ind, v, axis)
+        return
 
     cpdef repeat(self, repeats, axis=None):
         """Returns an array with repeated arrays along an axis.
@@ -1863,6 +1872,32 @@ cdef _take_kernel_0axis = ElementwiseKernel(
     'cupy_take_0axis')
 
 
+cdef _put_kernel = ElementwiseKernel(
+    'T v, S indices, int32 cdim, int32 rdim, int32 adim, S index_range',
+    'raw T a',
+    '''
+      S wrap_indices = indices % index_range;
+      if (wrap_indices < 0) wrap_indices += index_range;
+
+      int li = i / (rdim * cdim);
+      int ri = i % rdim;
+      a[(li * adim + wrap_indices) * rdim + ri] = v;
+    ''',
+    'cupy_put')
+
+
+cdef _put_kernel_0axis = ElementwiseKernel(
+    'T v, S indices, int32 rdim, S index_range',
+    'raw T a',
+    '''
+      S wrap_indices = indices % index_range;
+      if (wrap_indices < 0) wrap_indices += index_range;
+
+      a[wrap_indices * rdim + i % rdim] = v;
+      ''',
+    'cupy_put_0axis')
+
+
 cpdef ndarray _take(ndarray a, indices, axis=None, ndarray out=None):
     if a.ndim == 0:
         a = a[None]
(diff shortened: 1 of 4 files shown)


tests/cupy_tests/core_tests/test_ndarray.py
```diff
@@ -181,3 +204,107 @@ def test_output_type_mismatch(self, xp):
...
+        v_shape = compute_v_shape(
+            self.shape, numpy.array(self.indices).shape, self.axis)
+        v = testing.shaped_arange(v_shape, xp, dtype)
+        wrap_put(a, self.indices, v, self.axis)
+        return a
+
+
+@testing.parameterize(
+    {'shape': (3, 4, 5), 'indices': (2,), 'axis': 0, 'v_shape': (5,)},
+)
```
delta2323: We need not to use parameterized test if there is one combination of parameters.

delta2323:
Could you add test cases in which `axis` is `None` and some elements in indices are out of range?

delta2323:
According to the document, if `v` is shorter than `ind`, it will be repeated as necessary. Are there test cases that checks this behavior?

cupy/core/core.pyx
```diff
@@ -1921,6 +1956,43 @@ cpdef ndarray _take(ndarray a, indices, axis=None, ndarray out=None):
...
+    v_shape = lshape + ind.shape + rshape
+    v = broadcast_to(v, v_shape)  # this was in manipulation/dims.pyx
+
+    cdim = ind.size
+    rdim = internal.prod(rshape)
+    ind = ind.reshape(
+        (1,) * len(lshape) + ind.shape + (1,) * len(rshape))
+    if axis == 0 or axis is None:
+        _put_kernel_0axis(v, ind, rdim, index_range, a.reduced_view())
+        return
```
yuyu2172: Thanks you for comment.

cupy/core/core.pyx
```diff
@@ -1921,6 +1956,43 @@ cpdef ndarray _take(ndarray a, indices, axis=None, ndarray out=None):
...
+    else:
+        _put_kernel(v, ind, cdim, rdim, adim, index_range, a.reduced_view())
+        return
```
yuyu2172: Thanks you for comment.

tests/cupy_tests/core_tests/test_ndarray.py
yuyu2172: Thanks you for comment.

cupy/core/core.pyx
```diff
@@ -594,7 +594,16 @@ cdef class ndarray:
         """
         return _take(self, indices, axis, out)
 
-    # TODO(okuta): Implement put
+    cpdef put(self, ind, v):
+        """Replaces specified elements of an array with given values.
+
+        .. seealso::
+           :func:`cupy.put` for full documentation,
+           :meth: `numpy.ndarray.put`
```
yuyu2172: Thanks you for comment.

tests/cupy_tests/core_tests/test_ndarray.py
```diff
@@ -36,6 +36,29 @@ def wrap_take(array, *args, **kwargs):
     return array.take(*args, **kwargs)
 
 
+def wrap_put(a, ind, v, axis=None):
+    if get_array_module(a) == numpy:
```
yuyu2172: Thanks you for comment.

tests/cupy_tests/core_tests/test_ndarray.py
```diff
@@ -36,6 +36,29 @@ def wrap_take(array, *args, **kwargs):
...
+        if axis is not None and a.ndim != 0:
+            axis = axis % a.ndim
```
yuyu2172: Thanks you for comment.

tests/cupy_tests/core_tests/test_ndarray.py
```diff
@@ -36,6 +36,29 @@ def wrap_take(array, *args, **kwargs):
...
+            a.put(ind, v, mode='wrap')
+    else:
+        core._put(a, ind, v, axis)
+
+
+def compute_v_shape(in_shape, indices_shape, axis):
+    if axis is None or len(in_shape) == 0:
+        return indices_shape
+    else:
+        axis = axis % len(in_shape)
```
yuyu2172: Thanks you for comment.

cupy/core/core.pyx
```diff
@@ -1921,6 +1956,43 @@ cpdef ndarray _take(ndarray a, indices, axis=None, ndarray out=None):
             a.reduced_view(), indices, cdim, rdim, adim, index_range, out)
 
 
+cpdef _put(ndarray a, ind, v, axis=None):
```
yuyu2172: I am planning to use `_put` with `axis` when implementing `__setitem__` with array indexing in the future.

yuyu2172:
@delta2323 
Thanks for the comments.

>Could you add test cases in which axis is None and some elements in indices are out of range and confirm it certainly raises error.

Actually, it will not raise error because current implementation fixes `mode` to `wrap` as done in `cupy.take`. 

> According to the document, if v is shorter than ind, it will be repeated as necessary. Are there test cases that checks this behavior?

I totally made mistake here. The documentation was a mistake. Such behavior is not supported.

yuyu2172:
I start to feel that the functionality I would like to add `x[:, idx] = val` should not be called `put`, but something else.
I will resend PR with different name. 
One drawback is that no such method exists in `numpy`. I need to come up with the name of the method.

Sorry for inconvenience.

Issue #1649, opened 2016-09-13
ParallelUpdater cannot fully exploit multiple GPUs


iwiwi:
`ParallelUpdater` improves performance when using simple models and/or large mini batches. On the other hand, for complex models, `ParallelUpdater` currently fails to exploit multiple GPUs, and it becomes almost sequential.

I examined the profiler results, and concluded that Python is just too slow to issue kernels for multiple GPUs. The possible way to cope with this issue is to use multiple processes in `ParallelUpdater`. To enable that, cupy needs the feature to handle inter-process communication. It can be done using the inter-process memory handle of CUDA (http://docs.nvidia.com/cuda/cuda-c-programming-guide/#interprocess-communication).

The following is a profiler result with 4 GPUs and ResNet-152. It clearly shows that multiple GPUs are used in sequential.

<img width="1464" alt="2016-08-10 18 39 22" src="https://cloud.githubusercontent.com/assets/469803/18466558/21880536-79d7-11e6-9f8f-ebd117f22d68.png">

jekbradbury:
I have a patch which uses CUDA's interprocess communication and `multiprocessing` to give near-linear speedup for data parallelism across multiple GPUs, but it was written for Chainer 1.5 and I'd have to do some work to bring it up to compatibility with the latest version. Is there interest in this? I'm not sure when I'll get time to work on it, but it'll probably be in the next few weeks.

beam2d:
We are very interested in your code. It is great if you can spend some time to make your code compatible to the current Chainer and send a patch, but if it is difficult just sharing your code as is is still much helpful to us.

anaruse:
Last week, I spend some of my time to see if we can improve chainer's multi-GPU performance with python multiprocessing module. The experimental coding using the multiprocessing module with CUDA IPC APIs has been nearly finished. So, I'm attaching the results with ResNet-152 on 4 GPUs. As you can see from it, multiple GPUs are running in parallel and decent speed-up is achieved even in case of very deep network like ResNet-152.

![screenshot](https://cloud.githubusercontent.com/assets/20274068/20053166/ac2f56d0-a51a-11e6-9f17-3d85ef8fb5d7.png)

If you have interest in it, I will spend some more time to refactor and clean-up the code and send PR.  For now, some features are implemented in kind of not-easy-to-maintain manner...

beam2d:
[...]

PR #1848, opened 2016-11-08
Multi-GPUs training

anaruse:
Python multiprocessing module and CUDA IPC APIs are used (#1649)

### DIFF
--- a/chainer/link.py
+++ b/chainer/link.py
@@ -444,6 +444,8 @@ class Link(object):
     def gather_grads(self):
         """Put together all gradient arrays and make a single array
 
+        Return:
+            cupy.ndarray
         """
         size, num = self.size_num_grads()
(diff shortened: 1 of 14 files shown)


jekbradbury:
I am working on a PR for multiprocess multi-GPU training that works with Trainer/Updater/Optimizer; it will be ready in a few days.

anaruse:
Thank you, @jekbradbury. Sounds great. To be honest, I was expecting someone to make it available with Trainer/Updater/Optimizer module (^^).
Now, I'm considering to use [NCCL](https://github.com/NVIDIA/nccl) for summing up gradients of multiple processes (a.k.a all-reduce operation) that is required after each back-propagation phase. I think your work and use of NCCL are basically orthogonal, so I believe there is no problem to merge them.

PR #1847, opened 2016-11-08
Implement Eve

muupan:
This PR will add a new optimizer Eve #1846 .

Since Eve requires loss values to compute step sizes, `GradientMethod.update` is overwritten to get the outputs of `lossfun`. It will raise an error if `lossfun` is not specified. Tests for optimizers are also modified to use `lossfun`.

I found it is better than Adam when applied to `examples/mnist/train_mnist.py` with default parameters:
Eve
```
epoch       main/loss   validation/main/loss  main/accuracy  validation/main/accuracy
1           0.193573    0.0952835             0.9419         0.9696
2           0.0725242   0.0800335             0.977466       0.9765
3           0.0345596   0.06424               0.988915       0.981
4           0.013741    0.0567913             0.995832       0.9843
5           0.00571857  0.0562762             0.998649       0.985
6           0.00299029  0.053972              0.999333       0.9859
7           0.00175658  0.0581067             0.99965        0.9857
8           0.00100978  0.0592461             0.999917       0.9856
9           0.000664501  0.0626751             0.999933       0.9861
10          0.000589156  0.0727927             0.999933       0.9853
11          0.0013305   0.0659082             0.999733       0.9859
12          0.00049436  0.0670607             0.999883       0.9865
13          0.000336688  0.0687204             0.99995        0.9858
14          0.00010929  0.0690714             1              0.986
15          6.87041e-05  0.0699605             1              0.9865
16          5.16802e-05  0.0734176             1              0.986
17          3.75787e-05  0.0766912             1              0.9856
18          0.0028124   0.0766279             0.999183       0.985
19          0.000150661  0.0755002             1              0.9855
20          7.40332e-05  0.076406              1              0.9857
```
Adam
```
epoch       main/loss   validation/main/loss  main/accuracy  validation/main/accuracy
1           0.192746    0.118016              0.9411         0.9631
2           0.0727157   0.093093              0.977249       0.97
3           0.0501551   0.0860783             0.983615       0.9738
4           0.0367548   0.0736718             0.988399       0.9783
5           0.0272907   0.0791892             0.990932       0.9786
6           0.0258209   0.0827545             0.991516       0.9796
7           0.0203185   0.0644044             0.993332       0.9828
8           0.0177427   0.0757816             0.994432       0.9822
9           0.0161435   0.0873263             0.994815       0.9803
10          0.0139228   0.0853089             0.995249       0.9811
11          0.01513     0.081602              0.995066       0.983
[...]

Issue #1846, opened 2016-11-08
Eve optimization


delta2323:
Newly proposed optimization algorithm in [this paper](https://arxiv.org/abs/1611.01505). According to Fig. 1 in the paper, Eve algorithm achieves better loss value in CIFAR10 and CIFAR100

PR #1851, opened 2016-11-08
Eve optimizer

crcrpar:
#1846  
Most of code are the same as #1847.

I think it's possible to use Eve algorithm without `lossfun` by using `**kwds`.
To use loss value, just pass loss value, `optimizer.update(loss=loss_value)`.

I don't have GPU environment. So I just checked Eve worked without `lossfun` and I didn't make test code.

### DIFF
--- a/chainer/optimizers/eve.py
+++ b/chainer/optimizers/eve.py
@@ -36,14 +36,14 @@ class Eve(optimizer.GradientMethod):
         d, f = state['d'], state['f']
         if self.t > 1:
             old_f = float(cuda.to_cpu(state['f']))
-            if self.loss < old_f:
+            if self.loss > old_f:
                 delta = self.lower_threshold + 1.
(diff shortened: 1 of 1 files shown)


bkvogel:
Thank you for the PR! I like the idea of also being able to pass the loss value to `optimizer.update()`. Since there are now 2 PRs for this feature, please let me know which of the following options would work best for you:
- You can make a PR to muupan's branch (not master), and let him merge it into his branch. We can then continue the review from his PR.
- I can ask muupan to manually merge the difference in your code into his branch and we can all then collaborate to review his PR. I think only the `update()` method is different, so it would not be much work to do manually.

Thanks!

crcrpar:
I prefer the former. So I sent PR to muupan's branch. https://github.com/muupan/chainer/pull/1

Thank you.

Issue #1852, opened 2016-11-09
Cython version requirement in the install guide


iwiwi:
I failed to install Chainer from the master branch with Cython 0.23.4. Following @gwtnb -san's advise, the installation was successful after updating Cython to 0.25.1. So, probably the version requirement specified in the install guide is outdated.

https://github.com/pfnet/chainer/blob/master/docs/source/install.rst#install-chainer-for-developers

PR #1853, opened 2016-11-09
Update Cython's version requirement in the install guide

iwiwi:
This is to solve #1852. I heard that the correct version requirement for Cython is 0.24, which is also consistent to `chainer_setup_build.py`.


gwtnb:
LGTM

PR #1854, opened 2016-11-09
Update the error message about Cython's failure

iwiwi:
Even if old Cython is already installed, `setup.py` just says that `Please install Cython`. Making the message more informative would be beneficial.

image


gwtnb:
flake8

iwiwi:
sorry, fixed!

gwtnb:
LGTM

PR #1855, opened 2016-11-09
Move broadcast_to to cupy.core.core

yuyu2172:
This PR moves `broadcast_to` from `cupy/manipulation/dims.py` to `cupy/core/core.pyx`. This is necessary for methods inside of `core` to call `broadcast_to`.

Some changes being made include ...
+ the method is now defined by `cpdef`.
+ on line 1682 of `core.pyx`, `j` is changed to `j % array.ndim`. This is necessary because `_strides` does not accept negative values as index


okuta:
LGTM!

PR #1865, opened 2016-11-10
Fix GIL bug when call cuLaunchKernel

okuta:
`./train_imagenet.py` had stopped in my environment.


gwtnb:
Fails with python [run_test.py](https://github.com/pfnet/chainer-test/blob/d9180f4a0a84f163a5a021b32df91e72f7b9a143/run_test.py) --test py3

okuta:
I fixed. Please review this.

PR #1872, opened 2016-11-13
Fix corrupted doc

fukatani:


chainer/links/connection/highway.py
```diff
@@ -26,8 +26,9 @@ class Highway(link.Chain):
     Args:
         in_out_size (int): Dimension of input and output vectors.
         nobias (bool): If ``True``, then this function does not use the bias.
-        activate: Activation function of plain array. :math:`tanh` is also
-        available.
+        activate:
```
okuta: Why do you use a line break?

fukatani:
I deleted unnecessary line break.

okuta:
LGTM!

PR #1876, opened 2016-11-14
Fix bugs about multi gpu

kamo-naoyuki:
Fix bugs when using multi gpus for some links. I think  the StatefulGRU.set_state method has a same bug but I have no idea to fix it. 

Your cupy's specification is likely to cause similar bugs. I think you'd better to consider the possibility of changing it, or you should call attention to this specification on your documentation.

### DIFF
--- a/chainer/links/connection/lstm.py
+++ b/chainer/links/connection/lstm.py
@@ -5,6 +5,7 @@ import chainer
 from chainer.functions.activation import lstm
 from chainer.functions.array import concat
 from chainer.functions.array import split_axis
+from chainer import cuda
 from chainer import initializers
 from chainer import link
 from chainer.links.connection import linear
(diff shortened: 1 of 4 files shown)


zori:
Could you describe what the bug is and how to reproduce it? Also, the fix that you propose is removing the deferred parameter initialization. It would be better to find another fix, which still allows to pass `None` for `in_size`.

kamo-naoyuki:
See https://github.com/pfnet/chainer/pull/1170 or https://github.com/pfnet/chainer/issues/1869.

Sorry, I can' t find problems in my code and it works on my environment, but this problem is already known and some one could fix it.

kamo-naoyuki:
I found my bugs and my lstm might be old version. I'll close this PR and reopen.

PR #1881, opened 2016-11-14
Add mean absolute error loss

iwiwi:
This PR proposes to add mean absolute error (MAE) loss to the standard loss function collection of Chainer. MAE is the most popular loss function next to MSE for regression problems. Indeed, for example, Keras has built-in support for MAE (see https://keras.io/objectives/).

This implementation is largely based on MSE. I'm using this for several weeks in my week-end project.


chainer/functions/loss/mean_absolute_error.py
```diff
@@ -0,0 +1,46 @@
...
+        coeff = gy[0] * gy[0].dtype.type(1. / self.diff.size)
+        gx0 = coeff * xp.sign(self.diff)
+        return gx0, -gx0
+
+
+def mean_absolute_error(x0, x1):
+    """Mean squared error function.
+
+    This function computes mean squared error between two variables. The mean
+    is taken over the minibatch. Note that the error is not scaled by 1/2.
```
rezoo: 1.  s/mean squared error/mean absolute error/
2. I think a sentence of "Note that the error is not scaled by 1/2." is no longer needed.

iwiwi:
Thank you for pointing out my stupid copy-and-paste error. Fixed!

rezoo:
Thank you! Then I would like to merge.

PR #1882, opened 2016-11-14
Fix bug about multi gpu in LSTM

kamo-naoyuki:
Fix of the issue described here https://github.com/pfnet/chainer/issues/1869.

Reopen from https://github.com/pfnet/chainer/pull/1876(Sorry, this previous PR has some problems and I closed it.)

### DIFF
--- a/chainer/links/connection/lstm.py
+++ b/chainer/links/connection/lstm.py
@@ -2,6 +2,7 @@ import numpy
 import six
 
 import chainer
+from chainer import cuda
 from chainer.functions.activation import lstm
 from chainer.functions.array import concat
 from chainer.functions.array import split_axis
@@ -91,9 +92,10 @@ class StatelessLSTM(LSTMBase):
             lstm_in += self.lateral(h)
         if c is None:
             xp = self.xp
-            c = variable.Variable(
-                xp.zeros((x.shape[0], self.state_size), dtype=x.dtype),
-                volatile='auto')
+            with cuda.get_device(x.data):
+                c = variable.Variable(
+                    xp.zeros((x.shape[0], self.state_size), dtype=x.dtype),
+                    volatile='auto')
         return lstm.lstm(c, lstm_in)
 
 
@@ -239,9 +241,10 @@ class LSTM(LSTMBase):
                 lstm_in += self.lateral(self.h)
         if self.c is None:
             xp = self.xp
-            self.c = variable.Variable(
-                xp.zeros((batch, self.state_size), dtype=x.dtype),
-                volatile='auto')
+            with cuda.get_device(x.data):
+                self.c = variable.Variable(
+                    xp.zeros((batch, self.state_size), dtype=x.dtype),
+                    volatile='auto')
         self.c, y = lstm.lstm(self.c, lstm_in)
 
         if h_rest is None:
--- a/chainer/links/connection/lstm.py
+++ b/chainer/links/connection/lstm.py
@@ -195,8 +195,8 @@ class LSTM(LSTMBase):
             c_.to_cpu()
             h_.to_cpu()
         else:
-            c_.to_gpu()
-            h_.to_gpu()
+            c_.to_gpu(self._device_id)
+            h_.to_gpu(self._device_id)
         self.c = c_
         self.h = h_
 
--- a/chainer/links/connection/lstm.py
+++ b/chainer/links/connection/lstm.py
@@ -84,15 +84,16 @@ class StatelessLSTM(LSTMBase):
         """
         if self.upward.has_uninitialized_params:
             in_size = x.size // x.shape[0]
-            self.upward._initialize_params(in_size)
-            self._initialize_params()
+            with cuda.get_device(self._device_id):
+                self.upward._initialize_params(in_size)
+                self._initialize_params()
 
         lstm_in = self.upward(x)
         if h is not None:
             lstm_in += self.lateral(h)
         if c is None:
             xp = self.xp
-            with cuda.get_device(x.data):
+            with cuda.get_device(self._device_id):
                 c = variable.Variable(
                     xp.zeros((x.shape[0], self.state_size), dtype=x.dtype),
                     volatile='auto')
@@ -218,9 +219,10 @@ class LSTM(LSTMBase):
 
         """
         if self.upward.has_uninitialized_params:
-            in_size = x.size // x.shape[0]
-            self.upward._initialize_params(in_size)
-            self._initialize_params()
+            with cuda.get_device(self._device_id):
+                in_size = x.size // x.shape[0]
+                self.upward._initialize_params(in_size)
+                self._initialize_params()
 
         batch = x.shape[0]
         lstm_in = self.upward(x)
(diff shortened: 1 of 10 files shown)


okuta:
Could you add tests?

zori:
I updated the PR description to include a link to the issue (https://github.com/pfnet/chainer/issues/1869.)

kamo-naoyuki:
What do you mean "tests"? Do you mean the tests for these links on multi-gpus?
Sorry, I'm not familiar to your test codes.

kamo-naoyuki:
I'll write the omitted explanation of this bug.
 
By default, cupy selects device number 0. so the following array is allocated on device 0.
```
a = cupy.zeros(4)
assert a.device == cuda.get_device(0)
```

If you want to allocated on specified device, you need to use the get_device function with "with" sentence.

```
with with cuda.get_device(1):
  a = cupy.zeros(4)
assert a.device == cuda.get_device(1)
```
Or,

```
cuda.Device(1).use()
a = cupy.zeros(4)
assert a.device == cuda.get_device(1)
```

For example,  the array in https://github.com/pfnet/chainer/blob/285e6558640425f61b6aa8c00564ccf37643babb/chainer/links/connection/lstm.py of Line 220, was allocated on the default device independently of the input array.

I think this is not good specification because many users are note aware that cupy has a default device number and it can be changed by device.use method. In general global variable is not a good idea.
 
There is a another problem. Your chainer.Link class has xp property, so many people might think that an array which was allocated just  by using this module will works well, but really the proper device needs to be specified as following,

```
with cuda.get_device(x.data):
  self.xp.zeros(4)
```

Furthermore, I have another problem. If you'll define new chainer function, you don't have to do as it because Function.__call__ set the device before calling the forward method at line198 of https://github.com/pfnet/chainer/blob/master/chainer/function.py . I think this is a good solution locally but confusing on the whole.

okuta:
https://github.com/kamo-naoyuki/chainer/pull/1

chainer/links/connection/lstm.py
```diff
@@ -91,9 +92,10 @@ def __call__(self, c, h, x):
             lstm_in += self.lateral(h)
         if c is None:
             xp = self.xp
-            c = variable.Variable(
-                xp.zeros((x.shape[0], self.state_size), dtype=x.dtype),
-                volatile='auto')
+            with cuda.get_device(x.data):
```
okuta: Please use `with cuda.get_device(self._device_id):`.

chainer/links/connection/peephole.py
```diff
@@ -98,9 +99,10 @@ def __call__(self, x):
             lstm_in += self.lateral(self.h)
         if self.c is None:
             xp = self.xp
-            self.c = variable.Variable(
-                xp.zeros((x.shape[0], self.state_size), dtype=x.dtype),
-                volatile='auto')
+            with cuda.get_device(x.data):
```
okuta: Please use `with cuda.get_device(self._device_id):`.

chainer/links/connection/lstm.py
```diff
@@ -239,9 +241,10 @@ def __call__(self, x):
                 lstm_in += self.lateral(self.h)
         if self.c is None:
             xp = self.xp
-            self.c = variable.Variable(
-                xp.zeros((batch, self.state_size), dtype=x.dtype),
-                volatile='auto')
+            with cuda.get_device(x.data):
```
okuta: Please use `with cuda.get_device(self._device_id):`.

chainer/links/connection/zoneoutlstm.py
```diff
@@ -86,14 +87,18 @@ def __call__(self, x):
             lstm_in += self.lateral(self.h)
         else:
             xp = self.xp
-            self.h = variable.Variable(
-                xp.zeros((len(x.data), self.state_size), dtype=x.data.dtype),
-                volatile='auto')
+            with cuda.get_device(x.data):
```
okuta: Please use `with cuda.get_device(self._device_id):`.

chainer/links/connection/zoneoutlstm.py
```diff
@@ -86,14 +87,18 @@ def __call__(self, x):
...
+                self.h = variable.Variable(
+                    xp.zeros((len(x.data), self.state_size),
+                             dtype=x.data.dtype),
+                    volatile='auto')
         if self.c is None:
             xp = self.xp
-            self.c = variable.Variable(
-                xp.zeros((len(x.data), self.state_size), dtype=x.data.dtype),
-                volatile='auto')
+            with cuda.get_device(x.data):
```
okuta: Please use `with cuda.get_device(self._device_id):`.

chainer/links/normalization/batch_normalization.py
```diff
@@ -106,13 +107,15 @@ def __call__(self, x, test=False, finetune=False):
         if hasattr(self, 'gamma'):
             gamma = self.gamma
         else:
-            gamma = variable.Variable(self.xp.ones(
-                self.avg_mean.shape, dtype=x.dtype), volatile='auto')
+            with cuda.get_device(x.data):
```
okuta: Please use `with cuda.get_device(self._device_id):`.

chainer/links/normalization/batch_normalization.py
```diff
@@ -106,13 +107,15 @@ def __call__(self, x, test=False, finetune=False):
...
+                gamma = variable.Variable(self.xp.ones(
+                    self.avg_mean.shape, dtype=x.dtype), volatile='auto')
         if hasattr(self, 'beta'):
             beta = self.beta
         else:
-            beta = variable.Variable(self.xp.zeros(
-                self.avg_mean.shape, dtype=x.dtype), volatile='auto')
+            with cuda.get_device(x.data):
```
okuta: Please use `with cuda.get_device(self._device_id):`.

okuta:
@kamo-naoyuki: Do you have a time for this PR?

kamo-naoyuki:
@okuta : Sorry for the late  response, I merged your PR.

okuta:
Thank you for merging.

LGTM expect comments.
Please use with `cuda.get_device(self._device_id):`.

okuta:
@kamo-naoyuki: This bug is a critical for multi gpu user.
If you have no time, I merge this PR. After that, I will small problem.
Which is better?

kamo-naoyuki:
Sorry, I'm not used to Github and I don't know what you mean. This PR  can not be merged?

okuta:
Please use with `cuda.get_device(self._device_id):` instead of `with cuda.get_device(x.data):`.

https://github.com/pfnet/chainer/pull/1882#pullrequestreview-8638217

kamo-naoyuki:
It seems already done by your PR. What should i do?

okuta:
My PR fixed problem partially.
The fixing is not finished in `peephole.py`, `zoneoutlstm.py` and `batch_normalization.py`.
https://github.com/pfnet/chainer/pull/1882/files#diff-03a82c30057b5828d72241167e757938R102

kamo-naoyuki:
Sorry, I was not aware of it. So, I think chainer.links.lstm.LSTM.set_state has a same problem. Do you have any idea for it?

okuta:
Thank you for your fixing.

How about `c_.to_gpu(self._device_id)` at [this line](https://github.com/pfnet/chainer/blob/4df382824b6ff95a012643b7aacbc9ad3f738887/chainer/links/connection/zoneoutlstm.py#L61)?

okuta:
Thank you very much.
LGTM!

PR #1886, opened 2016-11-15
Use synchronous to_gpu to avoid deadlock

okuta:
This PR is workaround to avoid deadlock problem.


okuta:
This problem was resolved #1865

PR #1892, opened 2016-11-16
DCGAN example

mattya:
This is an example implementation of DCGAN.
Thanks @rezoo for helping me using trainer.

Below is the test result on cifar-10 after 34 epoch.
cifar10

There may be some rooms for improvement to make this example more helpful (e.g. adding celebA dataset example).
Any comments and suggestion will be appreciated.


mattya:
Thank you for your comments.  I fixed them.

examples/dcgan/README.md
```diff
@@ -0,0 +1,12 @@
+# DCGAN
+
+This is an example implementation of DCGAN (https://arxiv.org/abs/1511.06434) using trainer.
+
+This code uses Cifar-10 dataset by default.
+You can use your own dataset by specifying `--dataset` argument to the directory consisting image files for training.
```
muupan: I think `consisting of` is correct. You should not remove `of`.

muupan:
LGTM except the tiny grammar fix.

examples/dcgan/README.md
mattya: Ah, that's right.  Thanks

mattya:
Fixed it

PR #1893, opened 2016-11-16
Experimental feature

delta2323:
This PR adds a function that marks an API as *experimental*.

* Developer can mark any newly added API as experimental by calling `chainer.utils.experimental`.
* Chainer can modify the interfaces of experimental APIs at any version up.
* Chainer can remove experimental mark at any minor or major version up and consider the API as *stable*. Once APIs get stable, they cannot be *experimental* again in principle.
* If users use an experimental API for the first time,`FutureWarning` is raised (one warning for each experimental API)
* Users can disable the warning by setting `chainer.disable_experimental_feature_warning`to `False`

### DIFF
--- a/chainer/utils/experimental.py
+++ b/chainer/utils/experimental.py
@@ -6,7 +6,7 @@ import warnings
 def experimental(api_name=None):
     """A function for marking APIs as experimental
 
-    Developer of the API can mark it as *experimental* by calling
+    The developer of the API can mark it as *experimental* by calling
     this function. When users call experimental APIs, `FutureWarning`
     is raised once for each experimental API.
     The presentation of `FutureWarning` is disabled by setting
--- a/chainer/utils/experimental.py
+++ b/chainer/utils/experimental.py
@@ -3,7 +3,7 @@ import warnings
 
 
 def experimental(api_name):
-    """A function for marking APIs as experimental.
+    """Declares that user is using an experimental feature.
 
     The developer of an API can mark it as *experimental* by calling
     this function. When users call experimental APIs, :class:`FutureWarning`
@@ -65,8 +65,8 @@ The interface can change in the future.
         ... FutureWarning: chainer.foo.C is an experimental API. \
 The interface can change in the future
 
-    If we want to mark ``__init__`` method only, rather than class itself.
-    It is recommended that we explicitly feed its API name.
+    If we want to mark ``__init__`` method only, rather than class itself,
+    it is recommended that we explicitly feed its API name.
 
     .. testcode::
         :options: +ELLIPSIS, +NORMALIZE_WHITESPACE
--- a/chainer/utils/experimental.py
+++ b/chainer/utils/experimental.py
@@ -13,11 +13,64 @@ def experimental(api_name=None):
     `chainer.disable_experimental_warning` to `True`,
     which is `False` by default.
 
+    The basic usage is to call it in the function or method we want to
+    mark as experimental.
+
+    >>> from chainer.utils import experiment
+    ...
+    ... def f(x):
+    ...   experiment()
+    ...   # concrete implementation of f follows
+    ...
+    ... f(1)
+    FutureWarning: f is an experimental API. The interface can change in the future
+
+    By default, if the caller is a function, its name is used as an API name,
+    while if the caller is a method or a class mehtod,
+    `<class name>.<method name>` is used.
+
+    >>> class C(object):
+    ...   def f(self):
+    ...     experiment()
+    ...
+    ... c = C()
+    ... c.f()
+    FutureWarning: C.f is an experimental API. The interface can change in the future
+
+    We can also define the name of experimental API name as an argument.
+    It will use in the warning that will be issued.
+
+    >>> def g(x):
+    ...   experiment('My function')
+    ...
+    ... g(1)
+    FutureWarning: g is an experimental API. The interface can change in the future
+
+    If we want mark whole class, you should call this function
+    in its `__init__` method.
+
+    >>> class C():
+    ...   def __init__(self):
+    ...     experiment()
+
+    >>> C()
+    FutureWarning: C is an experimental API. The interface can change in the future
+    
+    If we want to mark `__init__` method only, rather than class itself.
+    It is recommended that we explicitly feed its API name.
+
+    >>> class D():
+    ...   def __init__(self):
+    ...     experiment('D.__init__')
+    FutureWarning: D.__init__ is an experimental API. The interface can change in the future
+
     Args:
         api_name(str): The name of an API marked as experimental.
-            If it is `None`, it is inferred from the caller.
-            If the caller is a function, its name is used.
-            If the caller is a method or a class mehtod,
+            If it is `None`, it is inferred from the caller by the following rule.
+            * If the caller is a function, its name is used.
+            * If the caller is `__init__` method of a class,
+            The class name of the method is used.
+            * If the caller is a method (other than `__init__`) of a class or a class mehtod,
             `<class name>.<method name>` is used.
     """
 
@@ -26,10 +79,13 @@ def experimental(api_name=None):
 
         frame = inspect.stack()[1][0]
         f_locals = frame.f_locals
-        if 'self' in f_locals:
+        if 'self' in f_locals:  # The caller is a method.
             class_name = f_locals['self'].__class__.__name__
-            api_name = class_name + '.' + api_name
-        elif 'cls' in f_locals:
+            if api_name == '__init__':
+                api_name = class_name
+            else:
+                api_name = class_name + '.' + api_name
+        elif 'cls' in f_locals:  # The caller is class method.
             class_name = f_locals['cls'].__name__
             api_name = class_name + '.' + api_name
 
--- a/chainer/utils/experimental.py
+++ b/chainer/utils/experimental.py
@@ -4,6 +4,22 @@ import warnings
 
 
 def experimental(api_name=None):
+    """A function for marking APIs as experimental
+
+    Developer of the API can mark it as *experimental* by calling
+    this function. When users call experimental APIs, `FutureWarning`
+    is raised once for each experimental API.
+    The presentation of `FutureWarning` is disabled by setting
+    `chainer.disable_experimental_warning` to `True`,
+    which is `False` by default.
+
+    Args:
+        api_name(str): The name of an API marked as experimental.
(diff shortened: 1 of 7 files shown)


chainer/utils/experimental.py
```diff
@@ -0,0 +1,40 @@
+import chainer
+import inspect
+import warnings
+
+
+def experimental(api_name=None):
+    """A function for marking APIs as experimental
+
+    Developer of the API can mark it as *experimental* by calling
```
bkvogel: Wording:
I suggest "The developer of an API can ..." instead of "Developer of the API can ...".

chainer/utils/experimental.py
```diff
@@ -0,0 +1,40 @@
...
+    this function. When users call experimental APIs, `FutureWarning`
+    is raised once for each experimental API.
+    The presentation of `FutureWarning` is disabled by setting
+    `chainer.disable_experimental_warning` to `True`,
+    which is `False` by default.
+
```
bkvogel: Can this also be used to mark a class as experimental? What is the suggested way to do that? Should I call `utils.experimental()` from inside the `__init__()` initializer? If so, please consider adding documentation that explains and/or gives examples of how to use the experimental feature to mark an API as experimental at the different granularities (function/method, class, module).

docs/source/compatibility.rst
```diff
@@ -68,6 +68,37 @@ The actual change should be done in the following steps:
 - At the major update announced in the above update, change the API.
 
 
+Experimental APIs
+-----------------
+
+Thanks to many contributors, we have introduced many new features to Chainer.
+But sometimes it was only after we released the new features that we noticed that their APIs are not appropriate.
+The objective of experimental APIs is to alleviate the design failure of them.
```
bkvogel: I suggest to replace:

But sometimes it was only after we released the new features that we noticed that their APIs are not appropriate.
 The objective of experimental APIs is to alleviate the design failure of them.
Any newly added API can be marked as *experimental*.

with the following:

However, we have sometimes released new features only to later notice that their APIs are not appropriate.
The objective of experimental APIs is to avoid such issues by allowing the developer to mark any newly added API as *experimental*.

docs/source/compatibility.rst
```diff
@@ -68,6 +68,37 @@ The actual change should be done in the following steps:
...
+Any newly added API can be marked as *experimental*.
+Any API that is not experimental is called *stable* in this document.
+
+.. note::
+    Undocumented behaviors are not considered as APIs. So they are not experimental nor stable.
+    The treatment of undocumented behaviors are descibed in :ref:`undocumented_behavior` section.
+
+Chainer can change the interfaces and documents of experimental APIs at **any** version up.
+This change is not considered as a break of backward compatibility.
+Chainer can remove experimental mark of APIs and make them stable at any **minor** or **major** version up.
```
bkvogel: Wording: I suggest to replace

Chainer can remove experimental mark of APIs and make them stable at any **minor** or **major** version up.

with the following:

Chainer can promote an experimental API to become stable at any **minor** or **major** version up.

chainer/utils/experimental.py
delta2323: As I wrote as unit tests, I expect developers call `utils.experimental` in `__init__` if a class itself is experimental.  We can emit warnings whenever users instantiate the class. Of course, this usage cannot support the case where users call static methods and class methods. But I think developers should call `utils.experimental` in them in that case.


But it is simply because I could not find out an appropriate alternative. If there is a clever way to cover all of instance methods, class methods, and static methods, I would like to choose it.

docs/source/compatibility.rst
delta2323: Fixed

docs/source/compatibility.rst
delta2323: Fixed

docs/source/compatibility.rst
```diff
@@ -68,6 +68,37 @@ The actual change should be done in the following steps:
...
+Once experimental APIs get stable, they cannot be experimental again.
```
delta2323: Fixed

chainer/utils/experimental.py
delta2323: If there is no better way, I will write the examples to the document.

delta2323:
Before merging this PR, I would like to confirm the following rule within the developer team: the promotion of experimental feature to stable is done in minor or major version up.

chainer/utils/experimental.py
```diff
@@ -0,0 +1,107 @@
...
+    The presentation of :class:`FutureWarning` is disabled by setting
+    ``chainer.disable_experimental_warning`` to ``True``,
+    which is ``False`` by default.
+
+    The basic usage is to call it in the function or method we want to
+    mark as experimental.
+
+    By default, if the caller is a function, its name is used as an API name.
+
+    >>> from chainer.utils import experiment
```
bkvogel: Should this be `from chainer.utils import experimental`? Consider replacing usages of `experiment` with `experimental` below in docstring.

delta2323: Fixed

chainer/utils/experimental.py
```diff
@@ -0,0 +1,107 @@
...
+            If the caller is a function, its name is used.
+            If the caller is a ``__init__`` method of a class,
+            the name of the class is used.
+            If the caller is a class method or
+            a method other than ``__init__``,
+            ``<class name>.<method name>`` is used.
+    """
+
+    if api_name is None:
+        api_name = inspect.stack()[1][3]
```
beam2d: Did you measure the performance of the stack inspection? I'm wondering how it slows down the execution of an experimental feature. If it is prohibitively slow, I recommend you to drop the automatic naming feature and make the api_name argument mandatory.

chainer/utils/experimental.py
```diff
@@ -0,0 +1,107 @@
+import chainer
+import inspect
+import warnings
+
+
+def experimental(api_name=None):
```
beam2d: What is the appropriate way to declare a specific usage of some existing feature as experimental (e.g., a new argument and a new type of an existing argument)? Do you want to use `experimental` for these cases, or prefer to directly use `warnings.warn`?

chainer/utils/experimental.py
delta2323: OK, I will conduct an experiment.

chainer/utils/experimental.py
delta2323: To be honest, I did not imagine such a usage, but I think it is better to support it. The most straightforward way is to check the argument in the function and call `experimental`

```python
def f(x, y, experimental_argument=None):
    if experimental_argument is not None:
        experimental()
```

I come to think we may have to change the warning message to support such usage.

chainer/utils/experimental.py
delta2323: I made a experiment ([code](https://gist.github.com/delta2323/67a49511ca0c6a0c4401ee4ebc1bf45a)) and found that the overhead is negligible. So I agree with making `api_name` mandatory.

chainer/utils/experimental.py
```diff
@@ -0,0 +1,66 @@
...
+    >>> def g(x, experimental_arg=None):
+    ...   if experimental_arg is not None:
+    ...     experimental('experimental_arg of chainer.foo.g')
+
+    Args:
+        api_name(str): The name of an API marked as experimental.
+    """
+
+    if not chainer.disable_experimental_feature_warning:
+        warnings.warn('{} is an experimental API. '
```
beam2d: Isn't `'{} is experimental'` easier to use than `'{} is an experimental API'`? If the API itself is experimental, the current message is ok, but if a certain usage of the API is experimental (e.g. passing a specific value of an argument to a function), it would be difficult to express the situation with the current message.

docs/source/compatibility.rst
```diff
@@ -68,6 +68,40 @@ The actual change should be done in the following steps:
...
+
+When users use experimental APIs for the first time, warnings are raised once for each experimental API,
+unless users explicitly disable the emission of the warnings in advance.
+
+See the document of :meth:`chainer.utils.experimental` how developers mark APIs as experimental
+and how users enable or disable the warnings practically..
+
+.. note::
+   It is up to developers if APIs should be annotated as experimental or not.
+   We recommend to make the APIs experimental if they implemnt large modules or
```
beam2d: typo: implemnt -> implement

chainer/utils/experimental.py
beam2d: `api_name` should not be optional.

chainer/utils/experimental.py
delta2323: I change the warning message, thanks.

docs/source/compatibility.rst
delta2323: I fixed it.

beam2d:
Thank you, now LGTM.

chainer/utils/experimental.py
```diff
@@ -0,0 +1,103 @@
+import chainer
+import warnings
+
+
+def experimental(api_name):
+    """A function for marking APIs as experimental.
```
beam2d: Start a docstring of function from a verb (of the third-person singular form), like this: `Declares that user is using an experimental feature.`

chainer/utils/experimental.py
```diff
@@ -0,0 +1,103 @@
...
+              experimental('chainer.foo.C')
+
+        C()
+
+    .. testoutput::
+
+        ... FutureWarning: chainer.foo.C is an experimental API. \
+The interface can change in the future
+
+    If we want to mark ``__init__`` method only, rather than class itself.
```
beam2d: This sentence is incomplete.

docs/source/compatibility.rst
```diff
@@ -68,6 +68,40 @@ The actual change should be done in the following steps:
...
+Chainer can change the interfaces and documents of experimental APIs at **any** version up.
+This change is not considered as a break of backward compatibility.
+Chainer can promote an experimental API to become stable at any **minor** or **major** version up.
+Once experimental APIs become stable, they cannot revert to experimental again.
+
+When users use experimental APIs for the first time, warnings are raised once for each experimental API,
+unless users explicitly disable the emission of the warnings in advance.
+
+See the document of :meth:`chainer.utils.experimental` how developers mark APIs as experimental
+and how users enable or disable the warnings practically..
```
beam2d: .. -> .

bkvogel:
Thank you. LGTM!

PR #1894, opened 2016-11-17
Extend element size limit for cupy array

uchida:
In current CuPy implementation cannot treat arrays whose element size larger than `2**31 - 1`,
this is because CuPy uses `int` for `size`/`shape`/`strides`.

For example:

```
>>> from chainer.cuda import cupy as xp
>>> arr = xp.empty(2**31 - 1, dtype='b')
>>> arr += 1 # it works
>>> arr = xp.empty(2**31, dtype='b')
>>> arr += 1 # raise Exception
...
TypeError: Unsupported type <class 'cupy.core.core.Indexer'>
```

This pull request remove this restriction by replacing these `int` with `ssize_t` or `Py_ssize_t`, as same as Python and NumPy does.

See also [PEP 353 - Using ssize_t as the index type](https://www.python.org/dev/peps/pep-0353/)

### DIFF
--- a/tests/cupy_tests/core_tests/test_ndarray_reduction.py
+++ b/tests/cupy_tests/core_tests/test_ndarray_reduction.py
@@ -20,6 +20,12 @@ class TestArrayReduction(unittest.TestCase):
         a = testing.shaped_random((3, 1000), xp, dtype)
         return a.max(axis=0)
 
+    @testing.for_dtypes(['?', 'b', 'h', 'e'])
+    @testing.numpy_cupy_allclose()
+    def test_max_axis_huge(self, xp, dtype):
+        a = testing.shaped_random((2, 1024, 1024), xp, dtype)
+        return a.max(axis=0)
+
     @testing.for_all_dtypes()
     @testing.numpy_cupy_allclose()
     def test_max_axis0(self, xp, dtype):
@@ -54,6 +60,12 @@ class TestArrayReduction(unittest.TestCase):
     @testing.numpy_cupy_allclose()
     def test_min_axis_large(self, xp, dtype):
         a = testing.shaped_random((3, 1000), xp, dtype)
+        return a.min(axis=0)
+
+    @testing.for_dtypes(['?', 'b', 'h', 'e'])
+    @testing.numpy_cupy_allclose()
+    def test_min_axis_huge(self, xp, dtype):
+        a = testing.shaped_random((2, 1024, 1024), xp, dtype)
         return a.min(axis=0)
 
     @testing.for_all_dtypes()
--- a/tests/cupy_tests/core_tests/test_ndarray_reduction.py
+++ b/tests/cupy_tests/core_tests/test_ndarray_reduction.py
@@ -20,10 +20,11 @@ class TestArrayReduction(unittest.TestCase):
         a = testing.shaped_random((3, 1000), xp, dtype)
         return a.max(axis=0)
 
+    @testing.attr.slow
     @testing.for_dtypes(['?', 'b', 'h', 'e'])
     @testing.numpy_cupy_allclose()
     def test_max_axis_huge(self, xp, dtype):
-        a = testing.shaped_random((2, 1024, 1024), xp, dtype)
+        a = testing.shaped_random((2048, 1024, 1024), xp, dtype)
         return a.max(axis=0)
 
     @testing.for_all_dtypes()
@@ -62,10 +63,11 @@ class TestArrayReduction(unittest.TestCase):
         a = testing.shaped_random((3, 1000), xp, dtype)
         return a.min(axis=0)
 
+    @testing.attr.slow
     @testing.for_dtypes(['?', 'b', 'h', 'e'])
     @testing.numpy_cupy_allclose()
     def test_min_axis_huge(self, xp, dtype):
-        a = testing.shaped_random((2, 1024, 1024), xp, dtype)
+        a = testing.shaped_random((2048, 1024, 1024), xp, dtype)
         return a.min(axis=0)
 
     @testing.for_all_dtypes()
--- a/tests/cupy_tests/creation_tests/test_basic.py
+++ b/tests/cupy_tests/creation_tests/test_basic.py
@@ -18,6 +18,7 @@ class TestBasic(unittest.TestCase):
         a.fill(0)
         return a
 
+    @testing.slow
     @testing.for_dtypes(['?', 'b', 'h', 'e'])
     @testing.numpy_cupy_array_equal()
     def test_empty_huge(self, xp, dtype):
@@ -39,6 +40,7 @@ class TestBasic(unittest.TestCase):
         a.fill(0)
         return a
 
+    @testing.slow
     @testing.for_dtypes(['?', 'b', 'h', 'e'])
     @testing.numpy_cupy_array_equal()
     def test_empty_huge_int(self, xp, dtype):
--- a/tests/cupy_tests/creation_tests/test_basic.py
+++ b/tests/cupy_tests/creation_tests/test_basic.py
@@ -18,6 +18,13 @@ class TestBasic(unittest.TestCase):
         a.fill(0)
         return a
 
+    @testing.for_dtypes(['?', 'b', 'h', 'e'])
+    @testing.numpy_cupy_array_equal()
+    def test_empty_huge(self, xp, dtype):
+        a = xp.empty((1024, 1024, 2048), dtype=dtype)
+        a.fill(0)
+        return a
+
     @testing.for_all_dtypes()
     @testing.numpy_cupy_array_equal()
     def test_empty_scalar(self, xp, dtype):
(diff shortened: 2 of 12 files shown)


uchida:
I think similar fixes will be required for some cupy kernels, using `int`/`int32` for indices, to work with these huge arrays.
However, currently I'm leaving them out in this pull request.

ir5:
Just for double-check: Can you run MNIST example in `examples/mnist/train_mnist.py` with your code? I encountered some CUDA related errors. The message was like as follows.
```
...
RuntimeError: `nvcc` command returns non-zero exit status.
command: ['nvcc', '--cubin', '-arch', 'sm_52', '/tmp/tmpbk80ba62/kern.cu']
return-code: 2
stdout/stderr:
b'/tmp/tmpbk80ba62/kern.cu(369): error: more than one operator "[]" matches these operands:\n            function "CArray<T, 0>::operator[](const ssize_t *) const [with T=float]"\n            function "CArray<T, 0>::operator[](ssize_t) const [with T=float]"\n            operand types are: const CArray<float, 0> [ int ]\n\n1 error detected in the compilation of "/tmp/tmpxft_000063fe_00000000-7_kern.cpp1.ii".\n'
```
I have checked that the result of nosetests looks good, so I cannot figure out the reason for the error right now. If you have this kind of error too, we will need to fix somewhere.

uchida:
@ir5 thank you for checking!
I found operator mismatch in `crossent_fwd` kernel in this case.
I will add workaround for these cupy kernels.

tests/cupy_tests/creation_tests/test_basic.py
```diff
@@ -18,6 +18,14 @@ def test_empty(self, xp, dtype):
         a.fill(0)
         return a
 
+    @testing.slow
```
ir5: It seems that `@testing.slow` is not defined in `cupy.testing`. Probably we need to define the decorator in `cupy/testing/attr.py`.

tests/cupy_tests/creation_tests/test_basic.py
```diff
@@ -32,6 +40,14 @@ def test_empty_int(self, xp, dtype):
...
```
ir5: Ditto.

tests/cupy_tests/creation_tests/test_basic.py
uchida: defined `cupy.testing.attr.slow` in b2e5f26

uchida:
Similar errors occured in following tests in `tests/chainer_tests`:

- test_embed_id
- test_select_item
- test_black_out
- test_crf1d
- test_ctc
- test_hinge
- test_negative_sampling
- test_hierarchical_softmax
- test_softmax_cross_entropy
- test_classifier
- test_optimizers_by_linear_model

In commit c751063, CuPy kernels used in these tests are fixed and I've tested them with `nosetests --eval-attr='gpu<2'`.

tests/cupy_tests/core_tests/test_ndarray_reduction.py
```diff
@@ -20,6 +20,12 @@ def test_max_axis_large(self, xp, dtype):
         a = testing.shaped_random((3, 1000), xp, dtype)
         return a.max(axis=0)
 
+    @testing.for_dtypes(['?', 'b', 'h', 'e'])
+    @testing.numpy_cupy_allclose()
+    def test_max_axis_huge(self, xp, dtype):
+        a = testing.shaped_random((2, 1024, 1024), xp, dtype)
```
ir5: This array seems not huge. Did you actually intend to write `(2048, 1024, 1024)` instead? If so, we will need `@slow` decorator here.

ir5:
Thank you for the fix. Now, I also checked all the tests pass. The pull request looks almost good to me, but could you check my comment above?

tests/cupy_tests/core_tests/test_ndarray_reduction.py
uchida: Ops, this should be `(2048, 1024, 1024)` and marked as slow.

uchida:
> Thank you for the fix. Now, I also checked all the tests pass. The pull request looks almost good to me, but could you check my comment above?

checked and reflected your review comments.

okuta:
I tested on Windows. But, this PR does not work.
Please use `ptrdiff_t` instead of `ssize_t`?

uchida:
Due to [The New Data Types (Windows)](https://msdn.microsoft.com/en-us/library/windows/desktop/aa384264(v=vs.85).aspx), `#include <basetsd.h>` may work for Windows, as [CPython does it](https://github.com/python/cpython/blob/master/PC/pyconfig.h#L219-L223).
In Windows, CUDA 6.5-8.0 support at least Visual Studio 2010 (`_MSC_VER >= 1600`), so I think it seems no problem to include.

okuta:
`basetsd.h｀ have `SSIZE_T`.
Do you want to use it?
[CPython declares `ssize_t` here.](https://github.com/python/cpython/blob/master/PC/pyconfig.h#L172).

How about following code?
```
#ifdef _MSC_VER
typedef ptrdiff_t ssize_t;
#endif
```

uchida:
Ah, OK. In b570d6a, `ptrdiff_t` is used for `ssize_t`.

okuta:
It works on Windows.
Thank you!

okuta:
I heard from C ++ expert that `ptrdiff_t` should be used.
There is `ptrdiff_t` in C89. It seems be more standard.
(I misunderstood that `ssize_t` is standard.)

Please change `ssize_t` to `ptrdiff_t`.

uchida:
In current implementation, there exists some code using `Py_ssize_t`, so I think `ssize_t`/`Py_ssize_t` is preferable to show these are equivalent.

How about use include `#Include "Python.h"` and `Py_ssize_t`?

uchida:
sum up what I know about `ptrdiff_t` and `ssize_t`/`Py_ssize_t`:

Pros for `ptrdiff_t`:
- defined in C/C++ standard
- capable to represent pointer arithmetic (adequate to represent negative offset)

Cons for `ptrdiff_t`:
- Python/NumPy does not use them as index

Pros for `ssize_t`/`Py_ssize_t`:
- Python/NumPy use them as index

Cons for `ssize_t`/`Py_ssize_t`:
- not defined in C standard, but in POSIX standard
- in standard, capable of storing values at least in the range `[-1, SSIZE_MAX]`
- [`Py_ssize_t` may become `intptr_t`](https://github.com/python/cpython/blob/7c1d63a73b1807c8dc9d05a4cd63e1bca5598d01/Include/pyport.h#L76-L85)

okuta:
> How about use include #Include "Python.h" and Py_ssize_t?
I think that idea is bad. "Python.h" is very complex. I am worried about side effects.

CUDA only supports 64bit system.
Maybe `ptrdiff_t` is almost same as `Py_ssize_t` and `ssize_t`.

uchida:
> I think that idea is bad. "Python.h" is very complex. I am worried about side effects.

OK, I'll use `ptrdiff_t` in CUDA codes and keep `Py_ssize_t` in Cython codes.
Interoperability between Cython/Python and CUDA will work due to similarity of `ptrdiff_t` and `Py_ssize_t`.

uchida:
I've used to `ptrdiff_t` in CUDA codes in e52cf4d.
Some tests in Travis CI failed, but for local environment, all tests are passed and 
it seems something wrong in Travis build environment.

```
$ PYTHONWARNINGS='module::DeprecationWarning' nosetests -a '!gpu,!slow' --with-doctest chainer_tests
............................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................SSSSSSSSSSSSSSSSSS.....................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................S....................................................................................................................................................................................................................................................................................................../home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/chainer/link.py:449: DeprecationWarning: Link.zerograds is deprecated. Use Link.cleargrads instead.
  DeprecationWarning)
/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/chainer/variable.py:243: DeprecationWarning: Variable.zerograd is deprecated. Use Variable.cleargard instead.
  DeprecationWarning)
....................................................................../home/travis/build.sh: line 57:  2380 Killed                  PYTHONWARNINGS='module::DeprecationWarning' nosetests -a '!gpu,!slow' --with-doctest chainer_tests
The command "PYTHONWARNINGS='module::DeprecationWarning' nosetests -a '!gpu,!slow' --with-doctest chainer_tests" exited with 137.
```

okuta:
Please rebase or merge with current master.
That problem is fixed.

uchida:
@okuta Thank you to notifying me. I've merge current master.

uchida:
Tests failed and this is due to [Certificate Expired on PyPI](https://status.python.org/incidents/3fvl2mdpj8f4).

okuta:
LGTM!

uchida:
I pushed a merge commit 08b2edd to resolve merge conflict on `chainer/functions/loss/softmax_cross_entropy.py`.

okuta:
Thank you for your work.

This PR breaks compatibility. Thus, this PR will merge in [cupy](https://github.com/cupy/cupy) for Chainer v2. I will tackle this function.
Can I create new PR that use this PR? (And, close this PR)

uchida:
> Can I create new PR that use this PR? (And, close this PR)

Of course, go ahead.

okuta:
Thank you!

PR #1895, opened 2016-11-17
Multi-GPU training with NCCL

anaruse:
I'm sending a PR to improve performance of multi-GPU training using python **multiprocessing** module and [**NCCL**](https://github.com/NVIDIA/nccl). NCCL is optimized primitives of collective communication over multiple GPUs and it supports **all-reduce** collective. By just using NCCL's all-reduce collective, you can reduce gradient arrays on multiple GPU, that is required processing after backward propagation step in case of synchronous data parallel method.

This PR is related to #1848 that is also PR for performance improvement of multi-GPU training. It also uses python multiprocessing module but does not utilize NCCL. It uses **CUDA IPC APIs** for device memory sharing over multiple processes and all-reduce operation must be implemented explicitly.

There are several algorithms to realize all-reduce operation and rather naive algorithm is used in #1848 to keep the readability of the source code. You can implement fast all-reduce algorithms with python, but they are rather complicated, the best algorithm may change depending on topology of GPU connections, and I think coding fast all-reduce algorithms is not main focus of chainer.

Off course, you need to build and install NCCL to use it from chainer. However, once you decide to use NCCL, you don't need to think anymore on how to exchange data correctly among GPUs, when to allocate/free temporal memory, when to synchronize among GPUs and so on to implement all-reduce operation. Please see and compare the examples, **train_mnist_MP_IPC.py** and **train_mnist_MP_NCCL.py** in the directory of examples/mnist. You will see use of NCCL is much easier than use of CUDA IPC.


anaruse:
I'm attaching the results of performance measurement on DGX-1 with and w/o NCCL. It shows that near-linear scalability is possible on 8-GPU machine with NCCL.  **Gather & Bcast** is a naive all-reduce algorithm used in #1848. 

resnet-152

unnonouno:
This feature is merged in #2213 Thank you!

PR #1896, opened 2016-11-18
Implementation of r2_score function

mottodora:
I implemented the r2_score function that calculates R^2(coefficient of determination) regression score function.
As of now, the code is compatible with scikit-learn, only when the "sample_weight" argument is set to "none". 
Is this argument necessary for chainer? If necessary, I will implement the changes.
Additionally, I would like to know whether the type of this argument is Variable or ndarray?

### DIFF
--- a/chainer/functions/evaluation/r2_score.py
+++ b/chainer/functions/evaluation/r2_score.py
@@ -1,0 +1,42 @@
+from chainer import cuda
+from chainer import function
+from chainer.utils import type_check
+
+
+class R2_score(function.Function):
+    def __init__(self, sample_weight, multioutput):
+        if sample_weight is not None:
+            raise NotImplementedError()
+        if multioutput in ['uniform_average', 'raw_values']:
+            self.multioutput = multioutput
+        else:
+            raise ValueError("invalid multioutput argument")
+
+    def check_type_forward(self, in_types):
+        type_check.expect(in_types.size() == 2)
+        pred_type, true_type = in_types
+
+        type_check.expect(
+            pred_type.dtype.kind == 'f',
+            true_type.dtype.kind == 'f'
+        )
+
+        type_check.expect(
+            pred_type.ndim >= true_type.ndim,
+            pred_type.shape == true_type.shape,
+        )
+
+    def forward(self, inputs):
+        xp = cuda.get_array_module(*inputs)
+        pred, true = inputs
+        SS_res = xp.sum((pred-true)**2, axis=0)
+        SS_tot = xp.sum((true-xp.mean(true, axis=0))**2, axis=0)
+        if self.multioutput == 'uniform_average':
+            return xp.asarray((1 - SS_res / SS_tot).mean(), dtype=pred.dtype),
+        elif self.multioutput == 'raw_values':
+            return xp.asarray((1 - SS_res / SS_tot), dtype=pred.dtype),
+
+
+def r2_score(pred, true, sample_weight=None, multioutput='uniform_average'):
+    return R2_score(sample_weight=sample_weight, multioutput=multioutput)\
+            (pred, true)
--- a/chainer/functions/evaluation/r2_score.py
+++ b/chainer/functions/evaluation/r2_score.py
@@ -32,9 +32,17 @@ class R2_score(function.Function):
         SS_res = xp.sum((pred - true) ** 2, axis=0)
         SS_tot = xp.sum((true - xp.mean(true, axis=0)) ** 2, axis=0)
         if self.multioutput == 'uniform_average':
-            return xp.asarray((1 - SS_res / SS_tot).mean(), dtype=pred.dtype),
+            if xp.any(SS_tot == 0):
+                return xp.asarray(0.0, dtype=pred.dtype),
+            else:
+                return xp.asarray((1 - SS_res / SS_tot).mean(),
+                                  dtype=pred.dtype),
         elif self.multioutput == 'raw_values':
-            return xp.asarray((1 - SS_res / SS_tot), dtype=pred.dtype),
+            if xp.any(SS_tot == 0):
+                return xp.where(SS_tot != 0, 1 - SS_res / SS_tot, 0.0)\
+                        .astype(pred.dtype),
+            else:
+                return xp.asarray((1 - SS_res / SS_tot), dtype=pred.dtype),
 
 
 def r2_score(pred, true, sample_weight=None, multioutput='uniform_average'):
--- a/chainer/functions/evaluation/r2_score.py
+++ b/chainer/functions/evaluation/r2_score.py
@@ -49,18 +49,18 @@ def r2_score(pred, true, sample_weight=None, multioutput='uniform_average'):
     """Computes R^2(coefficient of determination) regression score function.
 
     Args:
-        pred(Variable): Variable holding a vector or matrix of estimated \
+        pred(Variable): Variable holding a vector or matrix of estimated
                 target values.
         true(Variable): Variable holding a vector or matrix of correct target \
                 values.
         sample_weight: None.
-        multioutput(string): ['uniform_average', 'raw_values']. if \
-                'uniform_average', this function return an average of R^2\
-                score of multiple output. If 'raw_average', this function \
+        multioutput(string): ['uniform_average', 'raw_values']. if
+                'uniform_average', this function return an average of R^2
+                score of multiple output. If 'raw_average', this function
                 return a set of R^2 score of multiple output.
     Returns:
-        Variable: A Variable holding a scalar array of the R^2 score if \
-                'multioutput' is 'uniform_average' or a vector of R^2 \
+        Variable: A Variable holding a scalar array of the R^2 score if
+                'multioutput' is 'uniform_average' or a vector of R^2
                 scores if 'multioutput' is 'raw_values'.
 
     .. note:: This function is non-differentiable.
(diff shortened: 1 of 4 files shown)


delta2323:
Please fix violation of the coding guideline by referencing [travis result](https://travis-ci.org/pfnet/chainer/jobs/176959744).

chainer/functions/evaluation/r2_score.py
```diff
@@ -0,0 +1,62 @@
...
+        type_check.expect(in_types.size() == 2)
+        pred_type, true_type = in_types
+
+        type_check.expect(
+            pred_type.dtype.kind == 'f',
+            true_type.dtype.kind == 'f'
+        )
+
+        type_check.expect(
+            pred_type.ndim >= true_type.ndim,
```
delta2323: We do not need this condition because `pred_type.shape == true_type.shape` implies the equality of `ndim`'s

chainer/functions/evaluation/r2_score.py
```diff
@@ -0,0 +1,62 @@
...
+
+    def forward(self, inputs):
+        xp = cuda.get_array_module(*inputs)
+        pred, true = inputs
+        SS_res = xp.sum((pred-true)**2, axis=0)
+        SS_tot = xp.sum((true-xp.mean(true, axis=0))**2, axis=0)
+        if self.multioutput == 'uniform_average':
+            return xp.asarray((1 - SS_res / SS_tot).mean(), dtype=pred.dtype),
+        elif self.multioutput == 'raw_values':
+            return xp.asarray((1 - SS_res / SS_tot), dtype=pred.dtype),
```
delta2323: scikit learn seems to set r2 value to 0 when `SS_tot` is 0.

chainer/functions/evaluation/r2_score.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+def r2_score(pred, true, sample_weight=None, multioutput='uniform_average'):
+    """Computes R^2(coefficient of determination) regression score function.
+
+    Args:
+        pred(Variable): Variable holding a vector or matrix of estimated \
```
delta2323: We do not need backslashes in breaking the description of an argument.

chainer/functions/evaluation/r2_score.py
```diff
@@ -0,0 +1,62 @@
...
+    Returns:
+        Variable: A Variable holding a scalar array of the R^2 score if \
+                'multioutput' is 'uniform_average' or a vector of R^2 \
+                scores if 'multioutput' is 'raw_values'.
+
+    .. note:: This function is non-differentiable.
+
+    """
+    return R2_score(sample_weight=sample_weight, multioutput=multioutput)\
+            (pred, true)
```
delta2323: You can break the line in the middle of parentheses as follows:
```python
return R2_score(sample_weight=sample_weight,
                multioutput=multioutput)(pred, true)
```

chainer/functions/evaluation/r2_score.py
mottodora: I deleted backslash.

chainer/functions/evaluation/r2_score.py
mottodora: I fixed this line.

chainer/functions/evaluation/r2_score.py
```diff
@@ -22,41 +23,48 @@ def check_type_forward(self, in_types):
...
 
 
 def r2_score(pred, true, sample_weight=None, multioutput='uniform_average'):
     """Computes R^2(coefficient of determination) regression score function.
 
     Args:
-        pred(Variable): Variable holding a vector or matrix of estimated \
+        pred(Variable): Variable holding a vector or matrix of estimated
                 target values.
         true(Variable): Variable holding a vector or matrix of correct target \
```
delta2323: Please remove backslash here, too.

chainer/functions/evaluation/r2_score.py
```diff
@@ -0,0 +1,70 @@
...
+        SS_res = xp.sum((pred - true) ** 2, axis=0)
+        SS_tot = xp.sum((true - xp.mean(true, axis=0)) ** 2, axis=0)
+        if self.multioutput == 'uniform_average':
+            if xp.any(SS_tot == 0):
+                return xp.asarray(0.0, dtype=pred.dtype),
+            else:
+                return xp.asarray((1 - SS_res / SS_tot).mean(),
+                                  dtype=pred.dtype),
+        elif self.multioutput == 'raw_values':
+            if xp.any(SS_tot == 0):
```
delta2323: If the reason for this branching is to avoid zero division by `SS_tot`, we cannot avoid it because `SS_res / SS_tot` is evaluated anyway in case `SS_tot == 0`. Are there other reason for this branching? If not, I recommend to remove the branching. By removing it, we can expect faster forward computation.

chainer/functions/evaluation/r2_score.py
```diff
@@ -0,0 +1,70 @@
...
+                return xp.where(SS_tot != 0, 1 - SS_res / SS_tot, 0.0)\
+                    .astype(pred.dtype),
+            else:
+                return xp.asarray((1 - SS_res / SS_tot), dtype=pred.dtype),
+
```
delta2323: We can merge `uniform_average` case and `raw_values` case as follows

```python
ret = xp.asarray((1 - SS_res / SS_tot), dtype=pred.dtype)
if self.multioutput == `uniform_average`
    return ret.mean()
else:
    return ret
```

chainer/functions/evaluation/r2_score.py
```diff
@@ -0,0 +1,70 @@
...
+
+def r2_score(pred, true, sample_weight=None, multioutput='uniform_average'):
+    """Computes R^2(coefficient of determination) regression score function.
+
+    Args:
+        pred(Variable): Variable holding a vector or matrix of estimated
+                target values.
```
delta2323: `pred` can be not only a vector or a matrix but a tensor of any dimension.

chainer/functions/evaluation/r2_score.py
```diff
@@ -0,0 +1,70 @@
...
+        true(Variable): Variable holding a vector or matrix of correct target \
```
delta2323: `true` can be not only a vector or a matrix but a tensor of any dimension.

chainer/functions/evaluation/r2_score.py
```diff
@@ -0,0 +1,70 @@
...
+                values.
+        sample_weight: None.
```
delta2323: How about writing the description in more detail like this:

Current implementation allow this argument to be `None` only.

chainer/functions/evaluation/r2_score.py
```diff
@@ -0,0 +1,70 @@
...
+        multioutput(string): ['uniform_average', 'raw_values']. if
+                'uniform_average', this function return an average of R^2
```
delta2323: return*s*

chainer/functions/evaluation/r2_score.py
```diff
@@ -0,0 +1,70 @@
...
+                score of multiple output. If 'raw_average', this function
+                return a set of R^2 score of multiple output.
+    Returns:
+        Variable: A Variable holding a scalar array of the R^2 score if
+                'multioutput' is 'uniform_average' or a vector of R^2
+                scores if 'multioutput' is 'raw_values'.
+
+    .. note:: This function is non-differentiable.
```
delta2323: As forward computation of this function consists of differentiable operations, we can back propagate an error if we implement `backward` method appropriately. Do you think it is beneficial to implement backward?

tests/chainer_tests/functions_tests/evaluation_tests/test_r2_score.py
```diff
@@ -0,0 +1,114 @@
...
+        self.assertEqual(y.data.dtype, self.dtype)
+        if self.multioutput == 'uniform_average':
+            self.assertEqual((), y.data.shape)
+        elif self.multioutput == 'raw_values':
+            self.assertEqual(x_data.shape[1:], y.data.shape)
+
+        expected = r2_score(self.x, self.t, sample_weight=None,
+                            multioutput=self.multioutput)
+        testing.assert_allclose(
+            expected, cuda.to_cpu(y.data), **self.check_forward_options)
```
delta2323: You need not transfer `y.data` to CPU because `testing.assert_allclose` does it.

chainer/functions/evaluation/r2_score.py
mottodora: scikit learn returns 0 when SS_tot is 0. If not brancing, this function will return NaN.

chainer/functions/evaluation/r2_score.py
mottodora: I think that there is no difference between mean squared error and r_squared as loss function.

chainer/functions/evaluation/r2_score.py
delta2323: I thought something like this:

```python
ret = xp.where(SS_tot != 0, 1 - SS_res / SS_tot, 0.0).astype(pred.dtype)
if self.multioutput == 'uniform_average':
    return ret.mean()
else:
    return ret
```

Does it make sense to you?

mottodora: Yes. I'm going to implement like this.

delta2323:
Could you merge the latest master branch?

delta2323:
LGTM

delta2323:

PR #1898, opened 2016-11-19
Add binary_accuracy function to the reference manual

ronekko:
`F.binary_accuracy` is implemented at #757 but missing in the manual.


rezoo:
Oops thank you for your PR. LGTM.

PR #1794, opened 2016-10-19
cupy matmul

boeddeker:
A starting point for #1765.

And an example starting Function:

```
class MatMul(function.Function):

    @property
    def label(self):
        return '_ @ _'

    def check_type_forward(self, in_types):
        type_check.expect(in_types.size() == 2)
        if not in_types[0].value.dtype in (
                numpy.float32, numpy.complex64, numpy.complex128):
            raise TypeError(in_types[0].value.dtype)
        assert in_types[1].value.dtype in (
            numpy.float32, numpy.complex64, numpy.complex128)
        type_check.expect(
            in_types[0].ndim >= 2,
            in_types[1].ndim >= 2,
            in_types[0].shape[:-2] == in_types[1].shape[:-2],
            in_types[0].shape[-1] == in_types[1].shape[-2],
        )
        # TODO: Allow complex64 combined with float32

    def forward(self, x):
        xp = cuda.get_array_module(*x)
        if xp.isnan(x[0]).any():
            print('fw x[0] is nan.')
        if xp.isnan(x[1]).any():
            print('fw x[1] is nan.')
        result = utils.force_array(x[0] @ x[1])

        if xp.isnan(result).any():
            print('fw result is nan.')

        return result,

    def forward_cpu(self, inputs):
        raise NotImplementedError()

    def forward_gpu(self, inputs):
        raise NotImplementedError()

    def backward(self, x, gy):
        grad_x_0_star = utils.force_array(gy[0] @ xp_hermitian(x[1]))
        grad_x_1_star = utils.force_array(xp_hermitian(x[0]) @ gy[0])
        return grad_x_0_star, grad_x_1_star
```

PR #1901, opened 2016-11-20
Simplify matmul function

okuta:
This PR uses `cupy.matmul`.
Rel #1794

### DIFF
--- a/chainer/functions/math/inv.py
+++ b/chainer/functions/math/inv.py
@@ -1,5 +1,4 @@
 import numpy.linalg
-import six
 
 from chainer import cuda
 from chainer import function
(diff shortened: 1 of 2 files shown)


boeddeker:
Is it not better to change chainer.matmul to be similar to numpy?

numpy.matmul: shapes:  (..., n, n) * (..., n, n) = (..., n, n)
numpy.dot: shapes:  (n, n) * (n, n) = (n, n)
chainer.matmul: shapes:  (n, n) * (n, n) = (n, n)
chainer.batch_matmul: shapes:  (m, n, n) * (m, n, n) = (m, n, n)

The behavior of chainer.batch_matmul for vectors and matrices (ndim in {1, 2}) is much more strange.

And one further point, in my opinion the operation should be equal to numpy.matmul or raising an exception.

Example code for python 3.5 and numpy > 1.10 with the constraints, no broadcasting and inputs have minimum 2 dims:
```
def xp_hermitian(x):
    xp = cuda.get_array_module(x)
    return xp.swapaxes(x.conj(), -1, -2)

class MatMul(function.Function):
        type_check.expect(in_types.size() == 2)
        if not in_types[0].value.dtype in (
                numpy.float32, numpy.complex64, numpy.complex128):
            raise TypeError(in_types[0].value.dtype)
        assert in_types[1].value.dtype in (
            numpy.float32, numpy.complex64, numpy.complex128)
        type_check.expect(
            in_types[0].ndim >= 2,
            in_types[1].ndim >= 2,
            in_types[0].shape[:-2] == in_types[1].shape[:-2],
            in_types[0].shape[-1] == in_types[1].shape[-2],
        )
    def forward(self, x):
        result = utils.force_array(x[0] @ x[1])
        return result,

    def backward(self, x, gy):
        grad_x_0_star = utils.force_array(gy[0] @ xp_hermitian(x[1]))
        grad_x_1_star = utils.force_array(xp_hermitian(x[0]) @ gy[0])
        return grad_x_0_star, grad_x_1_star
```

okuta:
Thank you for your advice.
I will fix.

okuta:
@boeddeker I thought about your advice.
I think that it is better to deal with another PR because the interface of matmul is complex (eg transa and transb).
Could you create issue about your proposal?

gwtnb:
LGTM

PR #1902, opened 2016-11-21
fix a doc error

soramichi:
TupleDataset.__init__() checks whether all items in the datasets argument
have the same length as datasets[0], but the doc says all datasets must
have (some) length, which should be a mistake.


okuta:
Thank you!
LGTM

PR #1907, opened 2016-11-22
Improve doc of element-wise sigmoid logistic function

zori:
This PR improves the documentation of the element-wise sigmoid logistic function by adding a usage example.

Adding plots using sphinx extensions will be left for future PRs :).

EDIT: previously `[WIP] Add example usage to docstrings of functions and links`:
PRs welcome!

Part of improving documentaion effort. See also https://github.com/pfnet/chainer/issues/1867

TODO (finish this list):

Functions:
- [ ] linear
- [x] sigmoid
- [ ] ...

Links:
- [ ] Bilinear
- [ ] Maxout
- [ ] ...

### DIFF
--- a/chainer/functions/activation/sigmoid.py
+++ b/chainer/functions/activation/sigmoid.py
@@ -67,7 +67,8 @@ def sigmoid(x, use_cudnn=True):
 
         >>> x = np.random.uniform(0, 1, (3, 4)).astype('f')
         >>> y = F.sigmoid(x)
-        >>> assert y.shape == (3, 4)
+        >>> y.shape
+        (3, 4)
 
     Args:
         x (:class:`~chainer.Variable` or :class:`numpy.ndarray` or \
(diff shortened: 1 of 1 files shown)


mitmul:
Very good stating point. Some points I got as feedbacks:

- You can check the output by just putting the estimated output after it instead of using `assert` as follows:
```
>>> y.shape
(3, 4)
```
Writing this line `(3, 4)` enables the doctest to check the output of `y.shape` is the same as that line.

- Let's think about adding plots to show the difference of functions effectively like:
    - [numpy.exp](https://docs.scipy.org/doc/numpy/reference/generated/numpy.exp.html)
    - [Transfer Function Layers (from torch.nn)](https://github.com/torch/nn/blob/master/doc/transfer.md#nn.transfer.dok)

We can add emmbedded plots using sphinx extensions: http://matplotlib.org/sampledoc/extensions.html#inserting-matplotlib-plots

zori:
This PR is ready for review.

mitmul:
LGTM

mitmul:
Let's split PRs for each function to improve docstrings for them.

PR #1908, opened 2016-11-23
Support copyparams to uninitialized parameters

beam2d:
`link_a.copyparams(link_b)` skips uninitialized parameters of `link_a`, even if the corresponding parameters of `link_b` are already initialized. With this PR, these parameters are also copied to `link_a`.


gwtnb:
LGTM

PR #1909, opened 2016-11-24
Remove cupy from the source tree

beam2d:
Chainer v2 will remove cupy, which will be moved to a separate repository. This PR removes cupy-related stuffs from the source tree of Chainer. Standalone cupy is available at `cupy` branch.


okuta:
LGTM!
Can I merge this?

beam2d:
Yes, please. Let's think about CI after the merge.

PR #1911, opened 2016-11-24
Use direct import

### LINKED ISSUE #1763: __init__ style
Is there a reason, why the __init__.py has the style (concat is only an example)

```
from chainer.functions.array import concat

...

Concat = concat.Concat
concat = concat.concat
```

and not 

`
from chainer.functions.array.concat import Concat, concat
`
?

The reason, why I ask is, that PyCharm is not able to analyse this import style and therefore PyCharm says always: 'concat' is not callable and an jump to the function goes to the file and not to the function.

okuta:
Fix #1763
This PR use direct import that imports class and function.

### DIFF
--- a/chainer/__init__.py
+++ b/chainer/__init__.py
@@ -7,24 +7,52 @@ import threading
 from chainer import cuda  # NOQA
 from chainer import dataset  # NOQA
 from chainer import datasets  # NOQA
-from chainer import flag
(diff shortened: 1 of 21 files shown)


gwtnb:
Please rebase.

gwtnb:
LGTM

PR #1912, opened 2016-11-24
Suppress cudaMalloc before fork in test

gwtnb:
This PR suppresses `cudaMalloc` before fork in test. I am not sure whether this change is acceptable, but we must not call `cudaMalloc` in global scope.


okuta:
LGTM!

PR #1915, opened 2016-11-24
Check contiguous to reduce memory usage

okuta:
`activation_forward` calls `ascontiguousarray`.
If input array is not c contiguous, Chainer uses ElementwiseKernel to reduce memory usage and calculation time.

### DIFF
--- a/chainer/functions/activation/relu.py
+++ b/chainer/functions/activation/relu.py
@@ -32,6 +32,7 @@ class ReLU(function.Function):
 
     def forward_gpu(self, x):
         if (cuda.cudnn_enabled and self.use_cudnn and
+                x[0].flags.c_contiguous and
                 (_cudnn_version >= 3000 or x[0].dtype != numpy.float16)):
             y = cudnn.activation_forward(x[0], _mode)
             self.y = y
@@ -44,6 +45,7 @@ class ReLU(function.Function):
 
     def backward_gpu(self, x, gy):
         if (cuda.cudnn_enabled and self.use_cudnn and
+                x[0].flags.c_contiguous and gy[0].flags.c_contiguous and
                 (_cudnn_version >= 3000 or x[0].dtype != numpy.float16)):
             gx = cudnn.activation_backward(x[0], self.y, gy[0], _mode)
         else:
--- a/chainer/functions/activation/sigmoid.py
+++ b/chainer/functions/activation/sigmoid.py
@@ -30,7 +30,7 @@ class Sigmoid(function.Function):
 
     def forward_gpu(self, inputs):
         x = inputs[0]
-        if (cuda.cudnn_enabled and self.use_cudnn and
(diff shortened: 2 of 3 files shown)


jekbradbury:
The `n_step_lstm` function currently takes only c-contiguous arrays, and throws a fatal error otherwise. Would it be better to add `ascontiguousarray` there or to fallback to the non-CUDNN implementation? This is important if your tensors are stored with (batch, timestep, feature) dimensions, which is sometimes the only possible order (because the batch dimension needs to be the first one in most other functions).

okuta:
I think depend on algorithm.
The activation functions are not complex calculation. Thus, ElementwiseKernel implementation is efficient.

There is a possibility that accept CHW-contiguous array to reduce memory usage.
We need benchmark for typical case.

ir5:
To check the effectiveness, I wrote the following code for evaluation of relu(forward) in terms of time efficiency. https://gist.github.com/ir5/208558ff5de7bcb1e1bf5063d4934946
I then observed that `cupy` is twice efficient than `cudnn`. The result was as follows.
```
Average time (C-contiguous, cudnn): 0.000291290283203125 \pm 1.5441468575976905e-05
Average time (F-contiguous, cudnn): 0.0002834987640380859 \pm 2.1259978143655337e-05
Average time (C-contiguous, cupy ): 0.00013306140899658204 \pm 2.037457996870514e-05
Average time (F-contiguous, cupy ): 0.0001317310333251953 \pm 9.069843864997235e-06
```
I'm not such confident if my benchmark is valid or not, but if this is valid, then using only `cupy.maximum` in relu(forward) seems the best eventually. How do you think about this? @okuta

okuta:
Please use CUDA event for checking time.
https://gist.github.com/okuta/1387d563024c8a5364358212babd2cbf

ir5:
Thank you for information. (And I noticed that my previous benchmark was invalid because it always used F-contiguous arrays...)
I rewrote benchmark code. https://gist.github.com/ir5/3a85683072d81dc0c63021926249dba0 The result is as follows.
```
C-cudnn forward :5.502048015594482
F-cudnn forward :19.85126495361328
C-cupy  forward :4.877376079559326
F-cupy  forward :14.312288284301758
C-cudnn backward:8.100031852722168
F-cudnn backward:38.25388717651367
C-cupy  backward:7.17033576965332
F-cupy  backward:51.579105377197266
```
As @okuta points out, in forward computation, F-cupy is more efficient than F-cudnn probably because cudnn calls `ascontiguousarray` in it.
However, in backward computation, F-cupy is worse than F-cudnn. I don't know the reason behind this result. When I reduced the size of the array, the result changed accordingly. Things seems complicated.

By the way, can you think of the reason for the result that C-cupy is slightly better than C-cudnn?

beam2d:
LGTM

PR #1919, opened 2016-11-25
Add scatter_update

yuyu2172:
This PR is related to #1844 that I closed.

The PR adds CuPy a new function `scatter_update`.
There is no compatible "function" in NumPy. However, you can think of it as `__setitem__` with one integer array in the slices argument (i.e. `x.scatter_update(idx, val, axis=1)` behaves like `x[:, idx] = val`).

There are two differences from NumPy's `__setitem__`.

+ `scatter_update` does not raise error when indices exceed size of axes. Instead, `scatter_update` wraps indices.
+ as stated in #1844, when indices with duplicate values are input, the behavior is undefined.

I prefer to name this function `scatter_update` instead of `put` as I originally did in #1844 because of the following two reasons.

+ `put` repeats `value` when it is shorter than `ind`. I am not planning to support that behavior.
+ This name is more consistent with `scatter_add` (#1920).

Note that the implementation is general (e.g. `_scatter_op` instead of more specific `_scatter_update`), because it makes more sense when adding `scatter_add` (#1920).


rezoo:
Thank you for your PR!
Discussing with Okuta-san (a creator of cupy), we concluded that such unique functions should be moved in the namespace of ``cupy.ext`` in order to avoid duplicates.
If possible, could you move this function to ``cupy.ext`` namespace? (test functions are also)

cupy/indexing/scatter.py
```diff
@@ -0,0 +1,28 @@
+def scatter_update(a, indices, v, axis=0):
+    """Replaces specified elements of an array with given values.
+
+    ``v`` needs to be broadcastable to shape
+        ``a.shape[:axis] + indices.shape + a.shape[axis+1:]``.
+
+    Args:
+        a (ndarray): An array that gets updated.
+        indices (array-like): Indices of elements that this function takes.
```
rezoo: It would be better to explain that each element in `indices` also accepts a negative value.

cupy/__init__.py
```diff
@@ -207,6 +207,9 @@
 ix_ = indexing.generate.ix_
 
 fill_diagonal = indexing.insert.fill_diagonal
+
+scatter_update = indexing.scatter.scatter_update
```
yuyu2172: OK!

cupy/indexing/scatter.py
```diff
@@ -0,0 +1,28 @@
+def scatter_update(a, indices, v, axis=0):
+    """Replaces specified elements of an array with given values.
+
+    ``v`` needs to be broadcastable to shape
+        ``a.shape[:axis] + indices.shape + a.shape[axis+1:]``.
```
yuyu2172: 1. I added your suggestion.
2. I added similarity with `numpy.ndarray.__setitem__` to the doc because `numpy.put` does not take `axis` as an argument. Is it OK?
3. I added another example.

cupy/indexing/scatter.py
yuyu2172: Fixed.

cupy/indexing/scatter.py
```diff
@@ -0,0 +1,28 @@
...
+
+    Args:
+        a (ndarray): An array that gets updated.
+        indices (array-like): Indices of elements that this function takes.
+        v (array-like): Values to place in ``a`` at target indices.
+        axis (int): The axis along which to select indices.
```
yuyu2172: Fixed.

cupy/indexing/scatter.py
```diff
@@ -0,0 +1,28 @@
...
+            them that is used to store value is undefined.
+
+        Examples
+        --------
+        >>> a = cupy.zeros((2,))
+        >>> i = cupy.arange(10001) % 2
+        >>> v = cupy.arange(10000).astype(np.float)
+        >>> cupy.scatter_update(a, i, v, axis=0)
+        >>> a
+        [9982. 9983.]
```
yuyu2172: As I said in the reply to another review comment, I would like to contrast `scatter_update` with `numpy.ndarray.__setitem__`, so I added an example of `numpy.ndarray.__setitem__`.

yuyu2172:
Thank you for commenting.
I fixed commented parts.
I also added a simple test to make sure new namespace is working.

rezoo:
Good. In order to induce people who want to use a ``put`` function on GPU, could you add a section of ``put`` to the reference, and explains that we provide an alternative function called ``scatter_update`` although its specification is a little bit different?

yuyu2172:
Thank you for replying.

I reflected your comment to the code.

yuyu2172:
In addition to what you pointed out, I added a test case when array is one dimensional.

yuyu2172:
Could you look at comment I posted at #1920?

yuyu2172:
After discussing with the reviewer, we concluded that this function is not necessary.
Instead, the `__setitem__` will be integrated with part of the code in PR in different PR.

I will close this PR.

PR #1920, opened 2016-11-25
Add scatter_add

yuyu2172:
Merge after #1919.

This PR adds CuPy a new function `scatter_add`.
The function adds values to an array given indices  like `a[indices] = a[indices] + v`.
Note that `scatter_add` only supports slices with one array.

However, it differs in important way when there are duplicating values in `indices`.
In `scatter_add`, all values of an array `v` gets added. In `a[indices] = a[indices] + v`, only one of the index in duplicating elements of  `indices` gets used for addition.

Here is an example.
```python
>>> a = np.array([0, 0, 0])
>>> i = np.array([0, 0, 1])
>>> v = np.array([1, 1, 1])
>>> a[i] = a[i] + v
>>> print a
[1, 1, 0]
```

```python
>>> a = cupy.array([0, 0, 0]).astype(np.float32)
>>> i = cupy.array([0, 0, 1]).astype(np.float32)
>>> v = cupy.array([1, 1, 1]).astype(np.float32)
>>> cupy.scatter_add(a, i, v, axis=0)           
>>> print a                                     
[2, 1, 0]
```

Note that `scatter_add` is a function that is necessary when implementing gradient of `cupy.take`.


yuyu2172:
On the name of `scatter_add`, I would like to discuss.

I was thinking about  functions relating to `scatter_add`, and how they should be named.
The functions I have in mind are ...
1. `scatter_add` with multiple arrays and `slice(None)` as `indices`. (i.e. `a[i1, :, i2] += v`)
2. `scatter_add` with arrays and basic indiexing  (i.e. `a[i1, 0] += v`)

Here some are possible scenarios.
It is possible not to implement the 2nd function because `b = a[:, 0]; b[i1] += v` can be called instead.
In that case, possible name choice would be following, or something else.

|  | one array        | multiple array |
| -----|-------- |:-------------:|
|choice 1|  scatter_add    | scatter_add_multi |
| choice 2 | scatter_add_single     | scatter_add   | 

If a function that `scatter_add` with arrays and basic indexing were to be added, names can be `scatter_add_single`, `scatter_add_multi`, `scatter_add`.

yuyu2172:
The reviewer and I concluded that API of this function should be changed.
We think that it should have same interface as `__setitem__`.

PR #1921, opened 2016-11-25
raise error when array is included in slices of __setitem__

yuyu2172:
`__setitem__` does not support slices with arrays, but no error is raised.
Previously, `__getitem__` raised error in such cases, but since #1832 it no longer raise error.

This PR makes `__setitem__` raise error without depending on `__getitem__`.

### DIFF
--- a/cupy/core/core.pyx
+++ b/cupy/core/core.pyx
@@ -1160,6 +1160,14 @@ cdef class ndarray:
 
     def __setitem__(self, slices, value):
         cdef ndarray v, x, y
+        if not isinstance(slices, tuple):
+            slices = (slices,)
+
+        for i, s in enumerate(slices):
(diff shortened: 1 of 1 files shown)


okuta:
Please add test case.

cupy/core/core.pyx
```diff
@@ -1160,6 +1160,14 @@ cdef class ndarray:
 
     def __setitem__(self, slices, value):
         cdef ndarray v, x, y
+        if not isinstance(slices, tuple):
+            slices = (slices,)
+
+        for i, s in enumerate(slices):
```
okuta: Please do not use `enumerate`. `for s in slices:`

okuta:
LGTM except comment.

yuyu2172:
Fixed. Thank you for commenting.

okuta:
LGTM!

PR #1923, opened 2016-11-26
Add MultiprocessIterator to the reference

rezoo:
The current reference does not contain ``MultiprocessIterator`` due to the top of ``_last_signal = object()``.


okuta:
LGTM!

PR #1924, opened 2016-11-27
Add MultiprocessParallelUpdater

jekbradbury:
As I mentioned in #1848, I have a patch that creates a multiprocess data parallel updater that's compatible with the existing ParallelUpdater but avoids the speed issue with spawning kernels on multiple GPUs from a single process. Unfortunately I used a somewhat different approach to @anaruse so our PRs are not currently compatible.

There are a few design choices here that might be discussed further:

- I use a TreeReduce gather/scatter strategy, which is intermediate in efficiency between gathering and scattering to a single master, as in #1848, and the optimized AllReduce implementation in NCCL used in #1895. This should allow strong scaling to high GPU counts, which is important given that AWS now has 16-GPU machines on the public cloud (I don't have benchmarks right now).
- I use the original Chainer process as the master process and only spawn `(n_devices - 1)` worker processes. This allows the ordinary optimizer and evaluator objects to be used unchanged.
- (edited) In order to support models with uninitialized parameters (i.e., with `Linear` links that have `None` as their first constructor parameter), I run the first batch before setting up the IPC handles (which requires the param arrays to be ready).
- Since I create persistent per-array IPC handles, `MultiprocessParallelUpdater` requires the use of `zerograds` rather than `cleargrads`, because the gradient arrays must be persistent.
- I modified **both `MultiprocessParallelUpdater` and `ParallelUpdater`** to make the changes to optimizer attributes (either `lr` or `eps`) that account for the smaller batch size that each GPU sees during data-parallel training. That means that running the same model with the same overall batch size on any number of GPUs using either `StandardUpdater`, `ParallelUpdater`, or `MultiprocessParallelUpdater` should now result in identical learning dynamics (within numerical error). I believe I've accounted for the various types of gradient magnitude dependence exhibited by all built-in optimizers.

I haven't written tests; please feel free to push contributions to my branch if you want to add those or make other changes (I will need to add you as a collaborator if you aren't a core dev, or you can send a PR).


t-abe:
Thank you very much for the PR! We are going to review it soon.

chainer/training/updater.py
```diff
@@ -323,3 +345,203 @@ def update_core(self):
...
+
+    def _nodes_at(self, depth):
+        return range(len(self._devices) - 1)[2 ** depth - 1::2 ** (depth + 1)]
+
+    def update_core(self):
+        optimizer = self.get_optimizer('main')
+        batch = self.get_iterator('main').next()
+
+        n = len(self._devices)
+        master_device = self._devices[0] if self._started else -1
```
t-abe: This line cause an error.
Forgot to define ``_started`` ?

jekbradbury:
@t-abe sorry for the error, I introduced a bug and just pushed a fix

chainer/training/updater.py
```diff
@@ -260,6 +273,15 @@ def __init__(self, iterator, optimizer, converter=convert.concat_examples,
             if devices['main'] >= 0:
                 optimizer.target.to_gpu(devices['main'])
 
+        # Correct optimizer parameters for new minibatch size
+        optim = optimizer.__class__.__name__
+        if optim in ('Adam', 'AdaGrad', 'RMSprop'):
+            optimizer.eps /= len(devices)
```
beam2d: I found eps is divided by the number of devices here. Should it be multiplication instead? I thought that eps has the same unit as the gradient, and thus it should be scaled linearly with the mini-batch size.

jekbradbury: Yes, you are right. `eps` should be multiplied for those three update rules. I was wrong about the change for AdaDelta, which might be impossible to fix by multiplying and dividing hyperparameters; RMSpropGraves should also multiply, not divide, `eps` by `n_gpus ** 2`.

t-abe:
Sorry for waiting you long.
I made PR #2213 which combines your and anaruse's (#1895) PRs just now.
It is based on your MuiltiprocessParallelUpdater but
- uses NCCL (https://github.com/NVIDIA/nccl) for interprocess communication instead of CUDA IPC,
  - CUDA IPC does not support communication between gpus on different CPUs
- the dataset is split and each worker has own iterator (no need to send a batch via pipe).
  - in my setting, sending a batch data via pipe was a bottlneck

Although this PR is still WIP, at lest ImageNet example works well. I am very happy if you comment on the PR.
Thank you.

PR #1925, opened 2016-11-27
Support cupy.where to take arguments that only has condition

yuyu2172:
Currently, `cupy.where` does not support arguments with only `condition`.
This PR handles that case by calling `ndarray.nonzeros`.

example
```python
>>> cond = cupy.array([True, False, True])  
>>> cupy.where(cond)
(array([0, 2]),)
```


okuta:
LGTM!

PR #1926, opened 2016-11-27
Fix _adv_getitem for py3

okuta:
The test is failed on Python3.
This PR fix it problem.


okuta:
This bug is already fixed.

PR #1927, opened 2016-11-27
Add graph node label function

Hiroshiba:
I want to add the name of `chainer.Variable` at the graph by created `computational_graph.py`, but there are no way to do easily.
This PR will provide the method to set the dot node label by adding the label function.

### DIFF
--- a/chainer/computational_graph.py
+++ b/chainer/computational_graph.py
@@ -12,18 +12,20 @@ class DotNode(object):
 
     """
 
-    def __init__(self, node, attribute=None):
+    def __init__(self, node, attribute=None, label_func=None):
         """Initializes DotNode.
 
         Args:
             node: :class: `Variable` object or :class: `Function` object.
             attribute (dict): Attributes for the node.
+            label_func (function): Label function.
 
         """
         assert isinstance(node, (variable.Variable, function.Function))
         self.node = node
         self.id_ = id(node)
-        self.attribute = {'label': node.label}
+        label = node.label if label_func is None else label_func(node)
+        self.attribute = {'label': label}
         if isinstance(node, variable.Variable):
             self.attribute.update({'shape': 'oval'})
         else:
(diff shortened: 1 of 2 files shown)


Hiroshiba:
> FAIL: test_statistics_cpu (test_batch_normalization.TestPopulationStatistics_param_0)  parameter: {'nx': 10, 'ny': 10}
I don't know why this error happened...

ir5:
Thank you for your PR and sorry for being late for review. I read your code, but I do not understand the situation where we want to use the feature you have implemented. Currently, a shape and dtype are displayed on each variable node. This may be not informative, so at least I understand your point.
* If you want to display the name of variables (say, `x` or `y`), I would like to say current implementation does not look good. It will be painful to define a function that maps variables to our original name. In this case, probably it is preferred to modify the implementation of `chainer.Variable` so that we can modify the `label` property arbitrary.
* If you want to display abstract information with another format (say, removing dtype information for readability), the current implementation looks good.

As for `function_label_func`, I don't understand the benefit for this. Is there situation where we want to rename the function node? If there isn't, I recommend to remove the feature so that we can keep the implementation simpler.

Hiroshiba:
Thank you your comment. Yes, I want to write the Variable name on graph, and remove one of the shape such as batch size.

> implementation of chainer.Variable so that we can modify the label property arbitrary
That's nice idea, but I can't think of how to implement pretty... (to add static method like setter of label_function to `chainer.Variable`?)

> As for function_label_func, I don't understand the benefit for this.
I agree with that.

My mind was changed and I think it's better to add label format on `chainer.Variable`. I'd appreciate it if you could teach me how to design.

ir5:
Thank you for reply. After I posted my comment, I noticed that `chainer.Variable` has `name` member, which namely represents the name of the chainer variable. I think we can use this. To preserve the backward compatibility, one possible option may be to add "print label or name" argument in `ComputationalGraph.dump` so that a user can choose which one to print. If we can print `name` member, we will be able to use it flexibly. (example: https://gist.github.com/ir5/9f10c9b95701c1cfa1fd23d4d8bbcb9c)

PR #1930, opened 2016-11-27
allow adding lambda functions as an extension

soramichi:
This PR allows adding a lambda function as an extension,
resulting in simpler coding for simple extensions such as printing some values for debugging.

Example usage:

    ...
    trainer.extend(lambda x: print(model.accuracy.data), trigger=(1, "iteration"))
    ...

### DIFF
--- a/chainer/training/extension.py
+++ b/chainer/training/extension.py
@@ -19,9 +19,9 @@ class Extension(object):
     flexibility (for example, it can have methods to configure the behavior).
     Using a lambda function allows one-line coding for simple purposes, but
     users have to specify the configurations in :meth:`Trainer.extend` by
-    themselves. If a callable class not inheriting this class is used, it must
-    have `name` attribute (for a lambda function, the name of the function is
-    automatically used).
+    themselves. For a callable class not inheriting this class, the default
+    configurations of this class are used unless the user explicitly specifies
+    them in :meth:`Trainer.extend` method.
 
     Attributes:
         trigger: Default value of trigger for this extension. It is set to
--- a/chainer/training/extension.py
+++ b/chainer/training/extension.py
@@ -12,11 +12,13 @@ class Extension(object):
     attributes, e.g. the default trigger and the default priority. This class
     provides a set of typical default values for these attributes.
 
-    There are two ways to define users' own extensions: inheriting this class,
-    or decorating closures by :func:`make_extension`. Decorator can slightly
-    reduce the overhead and is much easier to use, while this class provides
-    more flexibility (for example, it can have methods to configure the
-    behavior).
+    There are three ways to define users' own extensions: inheriting this class,
+    decorating closures by :func:`make_extension`, or using any callable including
+    lambda functions as extensions. Decorator can slightly reduce the overhead
+    and is much easier to use, while this class provides more flexibility
+    (for example, it can have methods to configure the behavior). Using a lambda
+    function allows one-line coding for simple purposes, but users have to specify
+    the configurations in :meth:`Trainer.extend` by themselves.
 
     Attributes:
         trigger: Default value of trigger for this extension. It is set to
--- a/chainer/training/extension.py
+++ b/chainer/training/extension.py
@@ -12,15 +12,16 @@ class Extension(object):
     attributes, e.g. the default trigger and the default priority. This class
     provides a set of typical default values for these attributes.
 
-    There are three ways to define users' own extensions: inheriting this class,
-    decorating closures by :func:`make_extension`, or using any callable including
-    lambda functions as extensions. Decorator can slightly reduce the overhead
-    and is much easier to use, while this class provides more flexibility
-    (for example, it can have methods to configure the behavior). Using a lambda
-    function allows one-line coding for simple purposes, but users have to specify
-    the configurations in :meth:`Trainer.extend` by themselves. If a callable
-    class not inheriting this class is used, it must have `name` attribute (for
-    a lambda function, the name of the function is automatically used).
+    There are three ways to define users' own extensions: inheriting this
+    class, decorating closures by :func:`make_extension`, or using any callable
+    including lambda functions as extensions. Decorator can slightly reduce the
+    overhead and is much easier to use, while this class provides more
+    flexibility (for example, it can have methods to configure the behavior).
+    Using a lambda function allows one-line coding for simple purposes, but
+    users have to specify the configurations in :meth:`Trainer.extend` by
+    themselves. If a callable class not inheriting this class is used, it must
+    have `name` attribute (for a lambda function, the name of the function is
+    automatically used).
 
     Attributes:
         trigger: Default value of trigger for this extension. It is set to
--- a/chainer/training/extension.py
+++ b/chainer/training/extension.py
@@ -18,10 +18,10 @@ class Extension(object):
     overhead and is much easier to use, while this class provides more
     flexibility (for example, it can have methods to configure the behavior).
     Using a lambda function allows one-line coding for simple purposes, but
-    users have to specify the configurations in :meth:`Trainer.extend` by
-    themselves. For a callable class not inheriting this class, the default
-    configurations of this class are used unless the user explicitly specifies
-    them in :meth:`Trainer.extend` method.
+    users have to specify the configurations as arguments to
+    :meth:`Trainer.extend`. For a callable not inheriting this class, the
+    default configurations of this class are used unless the user explicitly
+    specifies them in :meth:`Trainer.extend` method.
 
     Attributes:
         trigger: Default value of trigger for this extension. It is set to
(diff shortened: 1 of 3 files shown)


muupan:
Thanks for your PR!

`isinstance(extension, types.FunctionType)` is not enough for checking whether it is a lambda. A function that is wrapped by `make_extension` also has `types.FunctionType`. It is possible that it has a non-`None` `finalize` attribute, but this PR would overwrite it with `None`.

Currently, extensions are expected to have `name` (or `default_name`) and `finalize` attributes, which is the reason it doesn't accept `lambda`. Instead of checking whether it is lambda or not, I suggest loosening the requirements by handling the case when it doesn't have such attributes:

- When `extension` doesn't have `name` nor `default_name`, then it is possible to check whether it has `__name__` and use it if yes, otherwise raise an error.
- `entry.extension.finalize` can be replaced with `getattr(entry.extension, 'finalize', None)`

By doing so, it can accept not only lambda but also any arbitrary callable object whose argument is `Trainer`.

In addition, this PR will affect the definition of `Extension` in Chainer, which is written in the docstring of `Extension` class: https://github.com/pfnet/chainer/blob/master/chainer/training/extension.py. Could you modify it accordingly?

soramichi:
Thank you for the comments and suggestions.

I modified the code and the docstring in according to your comments. The docstring is bit confusing because if you add a lambda (or any callable that do **not** inherit Exntesion class), is is never fired unless you explicitly specify the `trigger` argument in the `extend` method.

It might be a good idea to change `get_trigger` function in `training/util.py` so that it returns (1, 'epoch') if `trigger` is None, but I didn't do it as it may break other constrains.

muupan:
`Extension`'s docstring in `training/extension.py` contains more detailed explanation on how to define custom extensions. Could you modify it too?

As you wrote, modifying `get_trigger` would affect the behavior of existing code that explicitly set the `trigger` attribute to `None`. How about `getattr(extension, 'trigger', None)` -> `getattr(extension, 'trigger', (1, 'iteration'))` in https://github.com/pfnet/chainer/pull/1930/files#diff-768d99d739390d9a9a936299d6fa7fceR216 instead? It would not affect the existing code since extensions that inherit `Extension` or use `make_extension` must have the `trigger` attribute.

soramichi:
Added fixes (docstring, trigger) as suggested and also added tests for lambda-based and callable class-based extensions.

soramichi:
Thanks for the detailed review.

I changed the docstring of extension.py as suggested, and for trainer.py I wrote:
`If it is ``None`` and the extension does not have the trigger attribute`
instead of:
`If the extension does not have the trigger attribute`
because checking if an extension has the trigger attribute or not is done only when the `trigger` parameter given to `extend` is `None`.

```Python
        if trigger is None:
            trigger = getattr(extension, 'trigger', (1, 'iteration'))
        trigger = trigger_module.get_trigger(trigger)
```

chainer/training/trainer.py
```diff
@@ -184,7 +185,10 @@ def extend(self, extension, name=None, trigger=None, priority=None,
             trigger (tuple or Trigger): Trigger object that determines when to
                 invoke the extension. If it is ``None``, ``extension.trigger``
                 is used instead. If the trigger is not callable, it is passed
-                to :class:`IntervalTrigger` to build an interval trigger.
+                to :class:`IntervalTrigger` to build an interval trigger. If
+                it is ``None`` and the extension does not have the trigger
+                attribute, the extension is triggered at every iteration as
```
delta2323: I think "by default" is better.

chainer/training/extension.py
```diff
@@ -12,11 +12,16 @@ class Extension(object):
...
-    reduce the overhead and is much easier to use, while this class provides
-    more flexibility (for example, it can have methods to configure the
-    behavior).
+    There are three ways to define users' own extensions: inheriting this
+    class, decorating closures by :func:`make_extension`, or using any callable
+    including lambda functions as extensions. Decorator can slightly reduce the
+    overhead and is much easier to use, while this class provides more
+    flexibility (for example, it can have methods to configure the behavior).
+    Using a lambda function allows one-line coding for simple purposes, but
+    users have to specify the configurations in :meth:`Trainer.extend` by
```
delta2323: What do you mean by "the configurations in Trainer.extend" ? Is it the argument of the method?

chainer/training/extension.py
```diff
@@ -12,11 +12,16 @@ class Extension(object):
...
+    themselves. For a callable class not inheriting this class, the default
```
delta2323: "a callable", not "a callable class" is better because there is a callable that is not a class (e.g. functions).

chainer/training/trainer.py
```diff
@@ -39,10 +39,11 @@ class Trainer(object):
...
 
     Extensions are callable objects that take the trainer object as the
-    argument. There are two ways to define custom extensions: inheriting the
-    :class:`Extension` class, and decorating functions by
-    :func:`make_extension`. See :class:`Extension` for more details on custom
-    extensions.
+    argument. There are three ways to define custom extensions: inheriting the
+    :class:`Extension` class, decorating functions by :func:`make_extension`,
+    and defining any callable including lambda functions. The default
+    configurations of an extension is used in the former two cases. See
```
delta2323: For complete explanation, It is better to write what happens in the third case.

chainer/training/trainer.py
```diff
@@ -184,7 +185,10 @@ def extend(self, extension, name=None, trigger=None, priority=None,
             trigger (tuple or Trigger): Trigger object that determines when to
                 invoke the extension. If it is ``None``, ``extension.trigger``
                 is used instead. If the trigger is not callable, it is passed
-                to :class:`IntervalTrigger` to build an interval trigger.
+                to :class:`IntervalTrigger` to build an interval trigger. If
```
delta2323: I think the sentence begin with "If it is None and the extension ... as default" should be placed that after the sentence "If it is None, extension.trigger is used instead." because it would make the document more readable.

tests/chainer_tests/training_tests/test_trainer.py
```diff
@@ -56,13 +79,26 @@ def test_elapsed_time_serialization(self):
...
+        self.trainer.run()
+        self.assertTrue(dummy_callable_class.is_called)
+        self.assertTrue(dummy_callable_class.is_finalized)
+
+    def test_add_lambda_extension(self):
+        dummy_class = DummyClass()
+        self.trainer.extend(lambda x: dummy_class.touch())
+        self.trainer.run()
+        self.assertTrue(dummy_class.is_touched)
+
```
delta2323: Could you test extending with a regular function just in case?

chainer/training/trainer.py
muupan: I agree that it is not complete. How about removing the sentence `The default configurations of an extension is used in the former two cases.` because the complete explanation is already written in the class `Extension`?

soramichi: Thank you for the comments, but I don't fully understand what you mean.
Does "the third case" mean using any callable as an extension? I'm not sure what to explain for this case because there's really nothing that happens in this case (maybe we should explain the fact that nothing happens).

soramichi: Modified the docstring as follows. The idea is not to write too much here (as @muupan suggested)  but just mention that how to configuration is also in the docstring of `Extension`.
```
     Extensions are callable objects that take the trainer object as the
     argument. There are three ways to define custom extensions: inheriting the
     :class:`Extension` class, decorating functions by :func:`make_extension`,
     and defining any callable including lambda functions. See
     :class:`Extension` for more details on custom extensions and how to
     configure them.
```

chainer/training/trainer.py
soramichi: Yes, that's what I also though a bit when I fixed originally. Changed the order as suggested.

chainer/training/trainer.py
soramichi: Fixed as suggested.

chainer/training/extension.py
soramichi: Sorry for a hard-to-parse sentence. Fixed as follows:
```
users have to specify the configurations as arguments to :meth:`Trainer.extend`.
```
(should be "the" arguments? I'll fix again if needed)

chainer/training/extension.py
soramichi: Fixed as suggested.

tests/chainer_tests/training_tests/test_trainer.py
soramichi: Yes, I added a similar test for a regular function.

soramichi:
@delta2323 @muupan Thank you for the very detailed check and suggestions. I added two commits and replied to all the reviews on by one, please check them out.

muupan:
LGTM.

@delta2323 Can you review the changes?

delta2323:
LGTM

PR #1840, opened 2016-11-06
Support advanced indexing with boolean array for getitem

wkentaro:
Related to https://github.com/pfnet/chainer/issues/487

PR #1934, opened 2016-11-28
Include non tuple to test cases of ndarray.__getitem__

yuyu2172:
This PR adds a test case where argument `slices` of `__getitem__` is **not** tuple.
This is already supported in the current code, but the test missed this case.


delta2323:
As discussed in #1840, the contents of this PR is included in #1840. So I close it. Thank you.

PR #1935, opened 2016-11-28
Add packbits and unpackbits

iwiwi:
This PR addresses two TODOs of CuPy by adding functions `packbits` and `unpackbits`. As confirmed in the tests, they exactly follows the behaviour of the counterparts of NumPy, except that `packbits` does not support `axis` option.

### DIFF
--- a/cupy/binary/packing.py
+++ b/cupy/binary/packing.py
@@ -34,8 +34,13 @@ def packbits(myarray):
     cupy.ElementwiseKernel(
         'raw T myarray, raw int32 myarray_size', 'uint8 packed',
         '''for (int j = 0; j < 8; ++j) {
-            int k = i * 8 + j;
-            packed |= (k < myarray_size && myarray[k] ? 1 : 0) << (7 - j);
+            int k = i * 8 + j, x;
+            if (k < myarray_size && myarray[k] != 0) {
+                x = 1;
+            } else {
+                x = 0;
+            }
+            packed |= x << (7 - j);
         }''',
         'packbits_kernel'
     )(myarray, myarray.size, packed)
--- a/cupy/binary/packing.py
+++ b/cupy/binary/packing.py
@@ -21,7 +21,8 @@ def packbits(myarray):
     .. seealso:: :func:`numpy.packbits`
     """
     if myarray.dtype.kind not in 'biu':
-        raise TypeError('Expected an input array of integer or boolean data type')
+        raise TypeError(
+            'Expected an input array of integer or boolean data type')
 
     if myarray.size == 0:
         return cupy.empty_like(myarray)
--- a/cupy/binary/packing.py
+++ b/cupy/binary/packing.py
@@ -31,11 +31,10 @@ def packbits(myarray):
     packed = cupy.zeros((packed_size,), dtype=cupy.uint8)
 
     cupy.ElementwiseKernel(
-        'raw T myarray, raw int32 myarray_size',
-        'uint8 packed',
+        'raw T myarray, raw int32 myarray_size', 'uint8 packed',
         '''for (int j = 0; j < 8; ++j) {
             int k = i * 8 + j;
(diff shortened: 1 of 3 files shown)


iwiwi:
I conducted easy benchmarking, which packs and unpacks 100M elements. The result is as follows:

```
NumPy packbits:   0.1074282169342041
CuPy packbits:    0.0015967607498168946
NumPy unpackbits: 0.04982085227966308
CuPy unpackbits:  0.0016769170761108398
```

My benchmark script is as follows:

```python
import time
import cupy as cp
import numpy as np

N = 100000000
T = 10

def bench_packbits(xp):
    a = xp.random.randint(0, 2, N).astype(cp.uint8)
    xp.packbits(a)  # Warm-up
    s = time.time()
    for _ in range(T):
        xp.packbits(a)
    return (time.time() - s) / T

def bench_unpackbits(xp):
    a = xp.packbits(xp.random.randint(0, 2, N).astype(cp.uint8))
    s = time.time()
    for _ in range(T):
        xp.unpackbits(a)
    return (time.time() - s) / T

print('NumPy packbits:  ', bench_packbits(np))
print('CuPy packbits:   ', bench_packbits(cp))
print('NumPy unpackbits:', bench_unpackbits(np))
print('CuPy unpackbits: ', bench_unpackbits(cp))
```

The environment is as follows:

* CPU: Intel(R) Xeon(R) CPU E5-2623 v3 @ 3.00GHz
* GeForce GTX TITAN X

cupy/binary/packing.py
```diff
@@ -1,7 +1,70 @@
...
+
+    myarray = myarray.ravel()
+    packed_size = (myarray.size + 7) // 8
+    packed = cupy.zeros((packed_size,), dtype=cupy.uint8)
+
+    cupy.ElementwiseKernel(
+        'raw T myarray, raw int32 myarray_size', 'uint8 packed',
+        '''for (int j = 0; j < 8; ++j) {
+            int k = i * 8 + j;
+            packed |= (k < myarray_size && myarray[k] ? 1 : 0) << (7 - j);
```
gwtnb: An if statement should be used to make more readable.

gwtnb:
LGTM except the above comment.

iwiwi:
@gwtnb-san, thank you for your quick review!

Following your comment, I updated the kernel as follows and conducted experiments:

```cpp
        for (int j = 0; j < 8; ++j) {
            int k = i * 8 + j;
            if (k < myarray_size && myarray[k] != 0) {
                packed |= 1 << (7 - j);
            }
        }
```

```
NumPy packbits:   0.10719270706176758
CuPy packbits:    0.002148294448852539
NumPy unpackbits: 0.04986889362335205
CuPy unpackbits:  0.0016822338104248047
```

Unfortunately, the new kernel seems slower. I tried another one, but the result was similar:

```
        for (int j = 0; j < 8; ++j) {
            int k = i * 8 + j;
            if (k < myarray_size) {
                packed |= (bool)myarray[k] << (7 - j);
            }
        }
```

```
NumPy packbits:   0.1073335886001587
CuPy packbits:    0.0020667314529418945
NumPy unpackbits: 0.049921488761901854
CuPy unpackbits:  0.0017202377319335937
```

This discussion might be a kind of 'premature optimization', but it would be nice if we can find a way to attain both performance and readability. Do you have any suggestion?

gwtnb:
How about like this?

```cuda
uint8 bit = k < myarray_size && myarray[k];
packed |= bit << 7-j;
```

gwtnb:
Note: a value of a logical AND expression is always `0` or `1` in [C](http://port70.net/~nsz/c/c89/c89-draft.html#3.3.13).

iwiwi:
Thank you for nice suggestion. I updated the kernel. The benchmark result is as fast as the original one:

```
CuPy packbits:    0.0015712347030639648
```

gwtnb:
LGTM!

PR #1938, opened 2016-11-28
Fix doc of function_hooks.PrintHook to have proper import

yuyu2172:
I found an example code in docstring of `chainer.function_hooks.PrintHook` to fail to work.

It fails to call `chainer.function_hooks.PrintHook()` because `chainer/__init__.py` does not contain `from chainer import function_hooks`.

I am not sure if the doc is wrong, or it is a bug and the `from chainer import function_hooks` is simply forgotten in `chainer/__init__.py`.


okuta:
LGTM!

PR #1939, opened 2016-11-28
Fix cleagrads and zerograds when use uninitialized params

okuta:
This bug causes a very confusing problem.
If `Chain` object contains `Link` object, `grad` of uninitialized parameter in the `Link` object will be initialized by `NaN`.


gwtnb:
LGTM

PR #1942, opened 2016-11-29
Add DeCov loss (https://arxiv.org/abs/1511.06068)

longjie:
This PR is for adding DeCov loss.
The most of the code is from cross_covariance.py, except the diagonal elements are set to zero.

### DIFF
--- a/chainer/functions/__init__.py
+++ b/chainer/functions/__init__.py
@@ -57,6 +57,7 @@ from chainer.functions.loss import contrastive  # NOQA
 from chainer.functions.loss import crf1d  # NOQA
 from chainer.functions.loss import cross_covariance  # NOQA
 from chainer.functions.loss import ctc  # NOQA
+from chainer.functions.loss import decov  # NOQA
 from chainer.functions.loss import hinge  # NOQA
 from chainer.functions.loss import huber_loss  # NOQA
 from chainer.functions.loss import mean_absolute_error  # NOQA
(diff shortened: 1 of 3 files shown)


mattya:
Thank you for your PR.
I checked the code and it looks good, but I could not reproduce the result written in the paper (cifar-10 dataset, quick cifar10 architecture, 79.72% test accuracy).
Though I think this is because of hyper parameters (the exact value of the hyper parameters are not written in the paper), 
if you have some experimental results successfully reproducing the paper, could you tell us how to achieve that?

longjie:
Sorry, I didn't try to reproduce the CIFAR10 result. I just only used it to chainer's MNIST example to check the effect. If needed I'm trying the CIFAR10. Or you can just ask the authors the parameters :-)

longjie:
I've tried to reproduce the cifar10 result. I noticed the test accuracy curve is maximized at the middle of iterations, so we need to stop early to get the highest test accuracy. If do so, I can get the test accuracy around 79.5% . The code is here: 

https://github.com/longjie/chainer/commit/1e7c0d0c8eec464ac4ca53d7bdf9a294451d7c7c

Try with the option: --decov 10.0 

Do you think the authors did same manner? Or something wrong with my implementation?

longjie:
You need to reproduce the CIFAR10 result for this PR? If so, I'll ask the author. Thanks for your effort.

mattya:
Thank you for doing experiments on Cifar-10!
By testing your result, I found the reason why I could not reproduce the accuracy was the optimizer.
Using MomentumSGD instead of Adam got the almost same result as yours and I confirmed the Decov loss improves the accuracy (though I'm not sure why Adam didn't work with Decov loss).

I don't think it's necessary to ask the author with the details of the experimental setting.
So, please make a small fix on `chainer/functions/__init__.py` to avoid conflicts, and then LGTM!

longjie:
Thank you, and happy new year!

mattya:
Happy new year!
Sorry, I have one more thing to ask you.
Could you use `six.moves.range` instead of `six.moves.xrange`? (more python3 like style)

longjie:
Here it is.

mattya:
Thank you!

PR #1945, opened 2016-12-01
Add nonzero and flatnonzero to the overview page of CuPy document

iwiwi:
I think `nonzero` and `flatnonzero` are missing from the list in the overview page of CuPy document. This PR adds them there.

http://docs.chainer.org/en/stable/cupy-reference/overview.html#sorting-searching-and-counting
image


okuta:
LGTM!

iwiwi:
Thanks!

Issue #1946, opened 2016-12-01
cuda.available is not enough for Optimizer's addhooks


ShigekiKarita:
I have a little bit rare environment with cuda.available=True but it raises cudaErrorInitializationError because I installed cuda on unstable GPU.

My issue is that some optimizer-addhooks cannot run without GPU regardless of CUDA availability. When I added `optimizer.add_hook(chainer.optimizer.GradientNoise(eta=0.01))` into `train_mnist.py` and ran `python train_mnist.py -g -1`
```
Traceback (most recent call last):
  File "./train_mnist.py", line 108, in <module>
    main()
  File "./train_mnist.py", line 105, in main
    trainer.run()
  File "/home/karita/.virtualenvs/v3.5.2/lib/python3.5/site-packages/chainer-1.18.0-py3.5-linux-x86_64.egg/chainer/training/trainer.py", line 292, in run
    entry.extension(self)
  File "/home/karita/.pythonz/pythons/CPython-3.5.2/lib/python3.5/contextlib.py", line 77, in __exit__
    self.gen.throw(type, value, traceback)
  File "/home/karita/.virtualenvs/v3.5.2/lib/python3.5/site-packages/chainer-1.18.0-py3.5-linux-x86_64.egg/chainer/reporter.py", line 90, in scope
    yield
  File "/home/karita/.virtualenvs/v3.5.2/lib/python3.5/site-packages/chainer-1.18.0-py3.5-linux-x86_64.egg/chainer/training/trainer.py", line 289, in run
    update()
  File "/home/karita/.virtualenvs/v3.5.2/lib/python3.5/site-packages/chainer-1.18.0-py3.5-linux-x86_64.egg/chainer/training/updater.py", line 170, in update
    self.update_core()
  File "/home/karita/.virtualenvs/v3.5.2/lib/python3.5/site-packages/chainer-1.18.0-py3.5-linux-x86_64.egg/chainer/training/updater.py", line 182, in update_core
    optimizer.update(loss_func, *in_vars)
  File "/home/karita/.virtualenvs/v3.5.2/lib/python3.5/site-packages/chainer-1.18.0-py3.5-linux-x86_64.egg/chainer/optimizer.py", line 408, in update
    self.call_hooks()
  File "/home/karita/.virtualenvs/v3.5.2/lib/python3.5/site-packages/chainer-1.18.0-py3.5-linux-x86_64.egg/chainer/optimizer.py", line 243, in call_hooks
    hook(self)
[...]

PR #1947, opened 2016-12-01
fix optimizer add-hooks with cuda kernels

ShigekiKarita:
As I reported in my issue #1946
I found checking cuda.available is not enough to use cuda kernels.
So I create this work-around that checks not only cuda availability but also GPU usage.

### DIFF
--- a/chainer/optimizer.py
+++ b/chainer/optimizer.py
@@ -28,7 +28,6 @@ def exponential_decay_noise(xp, shape, dtype, hook, opt):
 
 
 class Optimizer(object):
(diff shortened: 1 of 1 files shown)


okuta:
Thank you for your PR. Your PR behavior is good.

Please fix style of code.
https://travis-ci.org/pfnet/chainer/jobs/180460058#L743

ShigekiKarita:
Thanks for your comment. I've just fixed it!

okuta:
LGTM!

PR #1950, opened 2016-12-02
Add layer normalization

soskek:
This PR adds a new link, LayerNormalization from "Layer Normalization." by Jimmy Lei Ba, Jamie Ryan Kiros and Geoffrey E. Hinton on https://arxiv.org/abs/1607.06450.

The paper says Layer Normalization is successful specially for internal layers in RNN including LSTM and GRU. Such RNN units with incorporated Layer Normalization (like [this](https://github.com/pfnet/chainer/blob/4df382824b6ff95a012643b7aacbc9ad3f738887/chainer/links/connection/zoneoutlstm.py)) are not included in this PR. After this PR is merged, I will begin to make them.

### DIFF
--- a/chainer/links/normalization/layer_normalization.py
+++ b/chainer/links/normalization/layer_normalization.py
@@ -1,6 +1,7 @@
 from chainer import cuda
 from chainer import initializers
 from chainer import link
+from chainer import utils
 
 from chainer.functions.array import broadcast
 from chainer.functions.math import bias
@@ -54,6 +55,8 @@ class LayerNormalization(link.Chain):
             initial_beta = initializers.Zero()
         self._beta_initializer = initial_beta
         self.eps = eps
+        utils.experimental(
+            'chainer.links.normalization.layer_normalization.py')
 
     def _initialize_params(self, size):
         self.add_param('gamma', size)
--- a/chainer/links/normalization/layer_normalization.py
+++ b/chainer/links/normalization/layer_normalization.py
@@ -1,10 +1,10 @@
-import chainer
+from chainer import initializers
+from chainer import link
+
 from chainer.functions.array import broadcast
 from chainer.functions.math import sqrt
 from chainer.functions.math import square
 from chainer.functions.math import sum
-from chainer import initializers
-from chainer import link
 from chainer.links.connection import scale
 
 
@@ -12,8 +12,8 @@ class LayerNormalization(link.Chain):
 
     """Layer normalization layer on outputs of linear functions.
 
-    This is a link of "Layer Normalization". This layer
-    normalizes, scales and shifts input units with :link:`~chainer.links.Scale`.
+    This is a link of "Layer Normalization". This layer normalizes,
+    scales and shifts input units with :link:`~chainer.links.Scale`.
 
     Args:
         size (int): Size of input units.
--- a/chainer/links/normalization/layer_normalization.py
+++ b/chainer/links/normalization/layer_normalization.py
@@ -12,11 +12,17 @@ class LayerNormalization(link.Chain):
 
     """Layer normalization layer on outputs of linear functions.
 
-    This is a link of "Layer Normalization". This layer normalizes,
-    scales and shifts input units with :link:`~chainer.links.Scale`.
+    This is a link of "Layer Normalization". This layer
+    normalizes input units by statistics along the second axis,
+    scales and shifts them with :class:`~chainer.links.Scale`.
 
     Args:
         size (int): Size of input units.
+
+    Attributes:
+        scale (~chainer.links.Scale): Layer to scale and shift
+            units after normalization.
+        eps (float): Epsilon value for numerical stability.
 
     See: `Layer Normalization <https://arxiv.org/abs/1607.06450>`_
     """
--- a/chainer/links/normalization/layer_normalization.py
+++ b/chainer/links/normalization/layer_normalization.py
@@ -2,10 +2,11 @@ from chainer import initializers
 from chainer import link
 
 from chainer.functions.array import broadcast
+from chainer.functions.math import bias
+from chainer.functions.math import scale
 from chainer.functions.math import sqrt
 from chainer.functions.math import square
 from chainer.functions.math import sum
-from chainer.links.connection import scale
 
 
 class LayerNormalization(link.Chain):
@@ -15,13 +16,13 @@ class LayerNormalization(link.Chain):
     This link implements a "layer normalization" layer
     which normalizes the input units by statistics
     that are computed along the second axis,
-    scales and shifts them with :class:`~chainer.links.Scale`.
+    scales and shifts them.
 
 
     Args:
         size (int): Size of input units.
         eps (float): Epsilon value for numerical stability of normalization.
-        initial_gumma (~chainer.Initializer): Initializer for scaling vector.
+        initial_gamma (~chainer.Initializer): Initializer for scaling vector.
             If ``None``, then the vector is initialized
             by :class:`~chainer.initializers.HeNormal`.
             If a scalar, the vectors are filled by it.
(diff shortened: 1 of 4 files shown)


chainer/links/normalization/layer_normalization.py
```diff
@@ -0,0 +1,53 @@
...
+from chainer.functions.math import square
+from chainer.functions.math import sum
+from chainer.links.connection import scale
+
+
+class LayerNormalization(link.Chain):
+
+    """Layer normalization layer on outputs of linear functions.
+
+    This is a link of "Layer Normalization". This layer
```
bkvogel: Consider rewording:

This link implements a "layer normalization" layer which normalizes the input units by statistics that are computed along the second axis, scales and shifts them with :class:`~chainer.links.Scale`.

chainer/links/normalization/layer_normalization.py
```diff
@@ -0,0 +1,53 @@
...
+    normalizes input units by statistics along the second axis,
+    scales and shifts them with :class:`~chainer.links.Scale`.
```
bkvogel: Is it required that the input to this link be a two-dimensional array, such as the output of a linear layer? Can it be applied to a convolutional layer? (although, in the paper it seems it was claimed that layer normalization does not work so well with convolutional layers anyway). If it should only be applied to the output of a linear layer, consider mentioning that.

chainer/links/normalization/layer_normalization.py
```diff
@@ -0,0 +1,53 @@
...
+
+    Args:
+        size (int): Size of input units.
```
bkvogel: Is size required in here in `__init__()`? Consider using the "parameter shapeholder" feature so that size can be removed from here.

bkvogel: Document the additional args: `eps`, `initial_gamma`, `inital_beta`.

chainer/links/normalization/layer_normalization.py
```diff
@@ -0,0 +1,53 @@
...
+        )
+        if initial_gamma is None:
+            initial_gamma = initializers.One()
+        initializers.init_weight(self.scale.W.data, initial_gamma)
+        if initial_beta is None:
+            initial_beta = initializers.Zero()
+        initializers.init_weight(self.scale.bias.b.data, initial_beta)
+        self.eps = eps
+
+    def normalize(self, x):
```
bkvogel: Is this intended to be a private method? If so, consider renaming to something like `_normalize()` to give users a hint that it is not part of documented API.

bkvogel:
Thank you for the PR. I have added some comments.

chainer/links/normalization/layer_normalization.py
soskek: I reworded.

chainer/links/normalization/layer_normalization.py
soskek: I added them.

chainer/links/normalization/layer_normalization.py
soskek: I made it private.

chainer/links/normalization/layer_normalization.py
bkvogel: Yes, the `Linear` link is an example that implements the parameter shapeholder functionality. The basic idea is that if some parameters have a shape that depends on the size of the input activations, then we should defer initialization of such parameters until the first call of `__call__(x)` because the required size can be extracted from x at that time. This makes the link easier to use because the user does not need to specify this information as an argument to `__init__()`. 

In the case of `Linear` link, `None` can be used to turn on the parameter shapeholder feature. However, that is only for backward compatibility. In the case of your layer normalization link, I think it might be best to remove `size` argument completely from `__init__()`.

chainer/links/normalization/layer_normalization.py
```diff
@@ -0,0 +1,77 @@
...
+        eps (float): Epsilon value for numerical stability.
+
+    See: `Layer Normalization <https://arxiv.org/abs/1607.06450>`_
+    """
+
+    def __init__(self, size, eps=1e-6, initial_gamma=None, initial_beta=None):
+        super(LayerNormalization, self).__init__(
+            scale=scale.Scale(axis=1, W_shape=(size, ), bias_term=True),
+        )
+        if initial_gamma is None:
```
bkvogel: I suggest to use `chainer.utils.experimental.py` to mark this class as having an experimental API.  This is a new Chainer feature that was just merged into the master branch yesterday (12/5).  This will allow us to the flexibility to continue modifying the API of new Chainer features without having to worry so much about getting the API perfect the fist time. You can use the "experimental feature API" by calling `experimental('chainer.links.normalization.layer_normalization.py')` in the `__init__()` method which will mark the entire LayerNormalization class as having an experimental API.

chainer/links/normalization/layer_normalization.py
soskek: I got it. Great!
I will try it, however, current `links.connection.linear.Scale`, used in this Layer Normalization, have not yet supported the parameter shapeholder functionality. So, I will use directly own scaling vector `gamma` and shifting vector `beta` (and `functions.scale`) in place of `Scale` link, and support the parameter shapeholder functionality for them.

chainer/links/normalization/layer_normalization.py
```diff
@@ -0,0 +1,94 @@
...
+    Args:
+        eps (float): Epsilon value for numerical stability of normalization.
+        initial_gamma (~chainer.Initializer): Initializer for scaling vector.
+            If ``None``, then the vector is initialized
+            by :class:`~chainer.initializers.HeNormal`.
+            If a scalar, the vectors are filled by it.
+            If ``numpy.ndarray``, the vectors are set by it.
+        initial_beta (~chainer.Initializer): Initializer for shifting vector.
+            If ``None``, then the vector is initialized
+            by :class:`~chainer.initializers.HeNormal`.
```
bkvogel: If None, initialized to 0, not HeNormal.

chainer/links/normalization/layer_normalization.py
```diff
@@ -0,0 +1,94 @@
...
+    scales and shifts them.
+    Parameter initialization will be deferred until
+    the first forward data pass at which time the size will be determined.
+
+
+    Args:
+        eps (float): Epsilon value for numerical stability of normalization.
+        initial_gamma (~chainer.Initializer): Initializer for scaling vector.
+            If ``None``, then the vector is initialized
+            by :class:`~chainer.initializers.HeNormal`.
```
soskek: Thank you. I fixed it.

chainer/links/normalization/layer_normalization.py
soskek: Thank you. I also fixed it.

bkvogel:
Regarding the performance, the current implementation does not look very efficient. However, the code is also straightforward and readable, which is nice. I think this is fine for now, but as a future todo item, I suspect performance could be significantly improved with some optimization. After it is merged, if this layer turns out to be a performance bottleneck, consider submitting another PR to further optimize performance.

soskek:
OK, I got it! I will time this layer's processing when incorporating it into RNN and consider if it should be optimized.
Thank you for many reviews!

bkvogel:
LGTM!

loofahcus:
Is it better to add a parameter ``size`` in __init__  and do not use add_uninitialized_param?
When we use trainer, if add_uninitialized_param is used, it's not convenient to copy the model (as the test_model) before trainer.run

bkvogel:
I am not sure I understand the copy issue completely. Are you saying that in some cases the user might still want to be able to manually specify the `size` in init? If so, maybe we could consider to add an optional parameter `size` with default value `None`.

However, it is also possible to first force the model to initialize itself by supplying inputs (could be junk data) of the desired size and calling the model to trigger a forward data pass which would then initialize all of the parameters. Or make a model that supports both train and test modes?

loofahcus:
The 'copy issue' I mentioned before is in this example:
[train_ptb.py](https://github.com/pfnet/chainer/blob/master/examples/ptb/train_ptb.py) 

Because the model use ``dropout``, so we can't make the model support both train and test modes, and it copies a eval_model in [train_ptb.py#L207](https://github.com/pfnet/chainer/blob/master/examples/ptb/train_ptb.py#L207) , but ``copy`` doesn't know about any uninitialized_params. If we add a ``LayerNormalization`` to this model,  when it comes to eval procedure, it will crash.  

Feed the model with inputs to force it initialize should work, but it seems not that elegant

bkvogel:
Thanks for the explanation. I understand the issue with copying now. I think it is probably fine to have `size` be an optional parameter of `__init__()`. Please feel free to submit a PR for this if you are in a hurry. Otherwise, Chainer is planned to have a more uniform API for the parameter shape placeholder feature in v2.

loofahcus:
@bkvogel Thanks! I submit a PR [pull#1981](https://github.com/pfnet/chainer/pull/1981), please review it.
And thanks for your great job @soskek  : )

soskek:
I also understand the issue. Thanks for your notice and improvement!

PR #1954, opened 2016-12-03
Test and doc for MLP conv shapeplace holder.

fukatani:
In current version chainer, I think that MLP convolution shape place holder is supported.
I modified doc and added test for the shape place holder.


okuta:
LGTM!

PR #1959, opened 2016-12-04
fix bernoulli_nll doc

fukatani:
Especially, the sign of the second term is wrong.


rezoo:
? Replacing `B(x; p)` with `\\log B(x; p)` is correct, but It seems that the sign of the second term is valid.

fukatani:
I understood what you and math mean.
But it is a problem that braces are gone, isn't it?

rezoo:
Ah I see, I would like to merge it. Thx!

PR #1968, opened 2016-12-06
Add constant weights argument to softmax_cross_entropy function

mitmul:
This PR adds an option to softmax_cross_entropy function for weighting the softmax cross entropy loss with a constant vector.

### DIFF
--- a/chainer/functions/loss/softmax_cross_entropy.py
+++ b/chainer/functions/loss/softmax_cross_entropy.py
@@ -1,11 +1,12 @@
+import numpy
+import six
+
+import chainer
 from chainer import cuda
 from chainer import function
 from chainer.functions.activation import log_softmax
 from chainer.utils import type_check
 
-import chainer
-import numpy
-import six
 
 
 class SoftmaxCrossEntropy(function.Function):
@@ -207,7 +208,8 @@ def softmax_cross_entropy(
             with the second dimension. The shape of this array should be
             ``(x.shape[1],)``. If this is not ``None``, each class weight
             ``class_weight[i]`` is actually multiplied to the corresponding
-            softmax output ``y[i]`` before calculating the actual loss value.
+            log-softmax output ``y[i]`` before calculating the actual loss
+            value.
 
     Returns:
         Variable: A variable holding a scalar array of the cross entropy loss.
--- a/chainer/functions/loss/softmax_cross_entropy.py
+++ b/chainer/functions/loss/softmax_cross_entropy.py
@@ -24,8 +24,6 @@ class SoftmaxCrossEntropy(function.Function):
         if class_weight is not None:
             assert self.class_weight.ndim == 1
             assert self.class_weight.dtype.kind == 'f'
-            if isinstance(class_weight, chainer.Variable):
-                self.class_weight = class_weight.data
 
     def check_type_forward(self, in_types):
         type_check.expect(in_types.size() == 2)
@@ -199,10 +197,12 @@ def softmax_cross_entropy(
         cache_score (bool): When it is ``True``, the function stores result
             of forward computation to use it on backward computation. It
             reduces computational cost though consumes more memory.
-        class_weight (~numpy.ndarray or ~cupy.ndarray or ~chainer.Variable): An
-            array that contains constant weights that will be multiplied with
-            the loss values along with the second dimension. The shape of this
-            array should be ``(x.shape[1],)``.
+        class_weight (~numpy.ndarray or ~cupy.ndarray): An array that contains
+            constant weights that will be multiplied with the loss values along
+            with the second dimension. The shape of this array should be
+            ``(x.shape[1],)``. If this is not ``None``, each class weight
+            ``class_weight[i]`` is actually multiplied to the corresponding
+            softmax output ``y[i]`` before calculating the actual loss value.
 
     Returns:
         Variable: A variable holding a scalar array of the cross entropy loss.
--- a/chainer/functions/loss/softmax_cross_entropy.py
+++ b/chainer/functions/loss/softmax_cross_entropy.py
@@ -20,10 +20,10 @@ class SoftmaxCrossEntropy(function.Function):
         self.use_cudnn = use_cudnn
         self.normalize = normalize
         self.cache_score = cache_score
-        self._class_weight = class_weight
-        if self._class_weight is not None:
-            assert self._class_weight.ndim == 1
-            assert self._class_weight.dtype.kind == 'f'
+        self.class_weight = class_weight
+        if self.class_weight is not None:
+            assert self.class_weight.ndim == 1
+            assert self.class_weight.dtype.kind == 'f'
 
     def check_type_forward(self, in_types):
         type_check.expect(in_types.size() == 2)
@@ -54,12 +54,10 @@ class SoftmaxCrossEntropy(function.Function):
         log_y = log_softmax._log_softmax(x, self.use_cudnn)
         if self.cache_score:
             self.y = numpy.exp(log_y)
-        if self._class_weight is not None:
-            if self._class_weight.shape != x.shape:
-                shape = [1 if d != 1 else -1 for d in six.moves.range(x.ndim)]
-                self._class_weight = numpy.broadcast_to(
-                    self._class_weight.reshape(shape), x.shape)
-            log_y *= self._class_weight
+        if self.class_weight is not None:
+            shape = [1 if d != 1 else -1 for d in six.moves.range(x.ndim)]
+            log_y *= numpy.broadcast_to(
+                self.class_weight.reshape(shape), x.shape)
         log_yd = numpy.rollaxis(log_y, 1)
         log_yd = log_yd.reshape(len(log_yd), -1)
         log_p = log_yd[numpy.maximum(t.ravel(), 0), numpy.arange(t.size)]
@@ -85,10 +83,10 @@ class SoftmaxCrossEntropy(function.Function):
         log_y = log_softmax._log_softmax(x, self.use_cudnn)
         if self.cache_score:
             self.y = cupy.exp(log_y)
-        if self._class_weight is not None:
+        if self.class_weight is not None:
             shape = [1 if d != 1 else -1 for d in six.moves.range(x.ndim)]
             log_y *= cupy.broadcast_to(
-                self._class_weight.reshape(shape), x.shape)
+                self.class_weight.reshape(shape), x.shape)
         if self.normalize:
             coeff = cupy.maximum(1, (t != self.ignore_label).sum())
         else:
@@ -114,9 +112,11 @@ class SoftmaxCrossEntropy(function.Function):
         if y.ndim == 2:
             gx = y
             gx[numpy.arange(len(t)), numpy.maximum(t, 0)] -= 1
-            if self._class_weight is not None:
-                c = self._class_weight[
-                    numpy.arange(len(t)), numpy.maximum(t, 0)]
+            if self.class_weight is not None:
+                shape = [1 if d != 1 else -1 for d in six.moves.range(x.ndim)]
+                c = numpy.broadcast_to(
+                    self.class_weight.reshape(shape), x.shape)
+                c = c[numpy.arange(len(t)), numpy.maximum(t, 0)]
                 gx *= numpy.broadcast_to(numpy.expand_dims(c, 1), gx.shape)
             gx *= (t != self.ignore_label).reshape((len(t), 1))
         else:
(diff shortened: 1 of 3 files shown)


rezoo:
Good PR, but the point we have to discuss is a name ``c``. I think it is curtailed too much, and scikit-learn uses ``class_weight`` to represent this argument. How about using this one?

mitmul:
Yes, I think `class_weight` is better. I'll fix it soon.

mitmul:
Fixed it, thanks!

chainer/functions/loss/softmax_cross_entropy.py
```diff
@@ -48,6 +54,12 @@ def forward_cpu(self, inputs):
         log_y = log_softmax._log_softmax(x, self.use_cudnn)
         if self.cache_score:
             self.y = numpy.exp(log_y)
+        if self.class_weight is not None:
+            if self.class_weight.shape != x.shape:
+                shape = [1 if d != 1 else -1 for d in six.moves.range(x.ndim)]
+                self.class_weight = numpy.broadcast_to(
+                    self.class_weight.reshape(shape), x.shape)
+            log_y *= self.class_weight
```
rezoo: It seems that

1. l.58 is no needed.
2. It is undesirable to implicitly change the shape of the public variable `class_weight`. 
How about using the following?

```python
if self.class_weight is not None:
    shape = [1 if d != 1 else -1 for d in six.moves.range(x.ndim)]
    log_y *= numpy.broadcast_to(
        self.class_weight.reshape(shape), x.shape)
```

chainer/functions/loss/softmax_cross_entropy.py
```diff
@@ -48,6 +54,12 @@ def forward_cpu(self, inputs):
         log_y = log_softmax._log_softmax(x, self.use_cudnn)
         if self.cache_score:
             self.y = numpy.exp(log_y)
+        if self.class_weight is not None:
```
rezoo: Ditto.

chainer/functions/loss/softmax_cross_entropy.py
```diff
@@ -14,10 +15,15 @@ class SoftmaxCrossEntropy(function.Function):
...
     normalize = True
 
-    def __init__(self, use_cudnn=True, normalize=True, cache_score=True):
+    def __init__(self, use_cudnn=True, normalize=True, cache_score=True,
+                 class_weight=None):
         self.use_cudnn = use_cudnn
         self.normalize = normalize
         self.cache_score = cache_score
+        self.class_weight = class_weight
+        if class_weight is not None:
```
rezoo: How about using `if self.class_weight:` instead of `if self.class_weight is not None:`?

chainer/functions/loss/softmax_cross_entropy.py
```diff
@@ -73,6 +85,10 @@ def forward_gpu(self, inputs):
         log_y = log_softmax._log_softmax(x, self.use_cudnn)
         if self.cache_score:
             self.y = cupy.exp(log_y)
+        if self.class_weight is not None:
```
rezoo: ditto.

chainer/functions/loss/softmax_cross_entropy.py
```diff
@@ -99,18 +114,27 @@ def backward_cpu(self, inputs, grad_outputs):
         if y.ndim == 2:
             gx = y
             gx[numpy.arange(len(t)), numpy.maximum(t, 0)] -= 1
+            if self.class_weight is not None:
```
rezoo: ditto.

chainer/functions/loss/softmax_cross_entropy.py
```diff
@@ -99,18 +114,27 @@ def backward_cpu(self, inputs, grad_outputs):
...
         else:
             # in the case where y.ndim is higher than 2,
             # we think that a current implementation is inefficient
             # because it yields two provisional arrays for indexing.
+            n_unit = t.size // len(t)
             gx = y.reshape(y.shape[0], y.shape[1], -1)
             fst_index = numpy.arange(t.size) // n_unit
             trd_index = numpy.arange(t.size) % n_unit
             gx[fst_index, numpy.maximum(t.ravel(), 0), trd_index] -= 1
+            if self.class_weight is not None:
```
rezoo: ditto.

chainer/functions/loss/softmax_cross_entropy.py
```diff
@@ -1,11 +1,12 @@
...
-
-import chainer
 from chainer import cuda
 from chainer import function
 from chainer.functions.activation import log_softmax
 from chainer.utils import type_check
 
+import chainer
+import numpy
+import six
```
mitmul: Oh, that's automatically done by my atom-beautify plugin...will fix it!

tests/chainer_tests/functions_tests/loss_tests/test_softmax_cross_entropy.py
```diff
@@ -84,44 +94,58 @@ def check_forward(self, x_data, t_data, use_cudnn=True):
         testing.assert_allclose(
             loss_expect, loss_value, **self.check_forward_options)
 
+    @testing.with_requires('numpy>=1.10')
```
mitmul: Because numpy.broadcast_to is available only from numpy>=1.10. I will add a comment before the decorator.

tests/chainer_tests/functions_tests/loss_tests/test_softmax_cross_entropy.py
```diff
@@ -84,44 +94,58 @@ def check_forward(self, x_data, t_data, use_cudnn=True):
...
+        self.check_backward(
+            cuda.to_gpu(self.x), cuda.to_gpu(self.t),
+            None if not self.weight_apply else cuda.to_gpu(self.class_weight),
+            False)
 
 
 @testing.parameterize(
     {'t_value': -2, 'valid': False},
-    {'t_value': 3,  'valid': False},
+    {'t_value': 3, 'valid': False},
```
mitmul: This was intended to do that exactly. Is there another unnecessary space?

chainer/functions/loss/softmax_cross_entropy.py
mitmul: `if self.class_weight:` causes `ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()` when it is not `None`.

chainer/functions/loss/softmax_cross_entropy.py
mitmul: It requires the same computation (calculate `shape` and broadcast `class_weight` to `x.shape`) again in backward_cpu, so it leads a duplicate computation. So how about making `class_weight` private member like `_class_weight`?

mitmul:
@rezoo Thank you for reviewing! I replied to some comments.

chainer/functions/loss/softmax_cross_entropy.py
rezoo: I see.

chainer/functions/loss/softmax_cross_entropy.py
rezoo: I think that it's not particularly a problem because (i) the computational cost of `shape` is much lower than `log_y *=`, and (ii) both `broadcast_to` functions in `numpy` and `cupy` do not return an actual array but a readonly view.

tests/chainer_tests/functions_tests/loss_tests/test_softmax_cross_entropy.py
rezoo: Probably no.

chainer/functions/loss/softmax_cross_entropy.py
mitmul: Thanks, I see. I'll add them to backward_cpu too.

tests/chainer_tests/functions_tests/loss_tests/test_softmax_cross_entropy.py
mitmul: 👍

mitmul:
@rezoo Updated with chainer.Variable support as a type of class_weight

chainer/functions/loss/softmax_cross_entropy.py
```diff
@@ -125,38 +154,55 @@ def backward_gpu(self, inputs, grad_outputs):
...
         normalize (bool): If ``True``, this function normalizes the cross
             entropy loss across all instances. If ``False``, it only
             normalizes along a batch size.
         cache_score (bool): When it is ``True``, the function stores result
             of forward computation to use it on backward computation. It
             reduces computational cost though consumes more memory.
+        class_weight (~numpy.ndarray or ~cupy.ndarray or ~chainer.Variable): An
+            array that contains constant weights that will be multiplied with
+            the loss values along with the second dimension. The shape of this
+            array should be ``(x.shape[1],)``.
```
beam2d: Explain which element is multiplied to each loss value. Some users can guess it from the argument name, but it should be explicitly noted in the document.

chainer/functions/loss/softmax_cross_entropy.py
```diff
@@ -125,38 +154,55 @@ def backward_gpu(self, inputs, grad_outputs):
...
-            If ``t[i] == -1``, corresponding ``x[i]`` is ignored.
+        t (~chainer.Variable): Variable holding an int32 vector of ground truth
+            labels. If ``t[i] == -1``, corresponding ``x[i]`` is ignored.
         normalize (bool): If ``True``, this function normalizes the cross
             entropy loss across all instances. If ``False``, it only
             normalizes along a batch size.
         cache_score (bool): When it is ``True``, the function stores result
             of forward computation to use it on backward computation. It
             reduces computational cost though consumes more memory.
+        class_weight (~numpy.ndarray or ~cupy.ndarray or ~chainer.Variable): An
```
beam2d: If you accept Variable as an argument of a differentiable function, you should pass it as an argument of Function._\_call__ and support backprop if possible. I think the resulting loss value of this function is differentiable by the class weights, and so you should support backprop if you want to accept Variable. I don't think there is a use case that one wants to differentiate the loss by the class weights for now, and so I think it is better to only accept the class weights in ndarray and disallow Variable.

mitmul: I see. OK I will make a modification again to diallow Variable.

chainer/functions/loss/softmax_cross_entropy.py
```diff
@@ -125,38 +152,57 @@ def backward_gpu(self, inputs, grad_outputs):
...
             normalizes along a batch size.
         cache_score (bool): When it is ``True``, the function stores result
             of forward computation to use it on backward computation. It
             reduces computational cost though consumes more memory.
+        class_weight (~numpy.ndarray or ~cupy.ndarray): An array that contains
+            constant weights that will be multiplied with the loss values along
+            with the second dimension. The shape of this array should be
+            ``(x.shape[1],)``. If this is not ``None``, each class weight
+            ``class_weight[i]`` is actually multiplied to the corresponding
+            softmax output ``y[i]`` before calculating the actual loss value.
```
beam2d: Is this explanation correct? I thought that ``class_weight[i]`` is multiplied to the log loss of ``y[j]`` if ``y[j]==i``.

chainer/functions/loss/softmax_cross_entropy.py
```diff
@@ -14,10 +15,15 @@ class SoftmaxCrossEntropy(function.Function):
...
-    def __init__(self, use_cudnn=True, normalize=True, cache_score=True):
+    def __init__(self, use_cudnn=True, normalize=True, cache_score=True,
+                 class_weight=None):
         self.use_cudnn = use_cudnn
         self.normalize = normalize
         self.cache_score = cache_score
+        self.class_weight = class_weight
+        if class_weight is not None:
+            assert self.class_weight.ndim == 1
+            assert self.class_weight.dtype.kind == 'f'
```
beam2d: Dont' use assert for inputs passed by user. Raise an exception explicitly, instead.

beam2d: Also check that class_weight is not a Variable, otherwise user would see a confusing error message.

chainer/functions/loss/softmax_cross_entropy.py
mitmul: Hm, I'd like to make sure my understanding. First, `class_weight` is once broadcasted to the same shape as `log_y` that is the output of `_log_softmax` method, and then is multiplied to `log_y`. So it's actually multiplied to the log-softmax outputs, right? And why can `y[j]` be an index here: `y[j]==i`? Because `i` seems an index number, but I intended that `y` is the output vector from log-softmax operation `_log_softmax`.

chainer/functions/loss/softmax_cross_entropy.py
mitmul: Thanks! Will add a raise statement and test of that.

mitmul: Got it :+1:

mitmul:
@beam2d Thanks for adding new comments. I added 3 `raise` in the constructor and added tests for them.

chainer/functions/loss/softmax_cross_entropy.py
```diff
@@ -125,38 +157,58 @@ def backward_gpu(self, inputs, grad_outputs):
...
         cache_score (bool): When it is ``True``, the function stores result
             of forward computation to use it on backward computation. It
             reduces computational cost though consumes more memory.
+        class_weight (~numpy.ndarray or ~chainer.cuda.cupy.ndarray): An array
+            that contains constant weights that will be multiplied with the
+            loss values along with the second dimension. The shape of this
+            array should be ``(x.shape[1],)``. If this is not ``None``, each
+            class weight ``class_weight[i]`` is actually multiplied to the
+            corresponding log-softmax output ``y[i]`` before calculating the
+            actual loss value.
```
beam2d: (I start another thread since the old one is collapsed into the outdated diff)
OK, I should've written `t[i]==j` instead of `y[i]==j`. I maybe now understand your point, so then you should clarify what `y` is and its shape, and write `y[:, i]` instead of `y[i]` (if y is log(softmax(x))).

beam2d:
Thanks, LGTM now.

PR #1977, opened 2016-12-08
Suppress experimental feature warnings in test

gwtnb:
Experimental feature warning is not helpful for test.

### DIFF
--- a/tests/chainer_tests/links_tests/normalization_tests/test_layer_normalization.py
+++ b/tests/chainer_tests/links_tests/normalization_tests/test_layer_normalization.py
@@ -12,6 +12,15 @@ from chainer.testing import condition
 from chainer.utils import type_check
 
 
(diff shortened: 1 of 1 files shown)


bkvogel:
Is the reason for disabling the warning so that the test output will look nicer? That is, you don't want to see lots of experimental warnings when running the tests? Is there any other reason it is bad to print warnings in the tests?

gwtnb:
> Is the reason for disabling the warning so that the test output will look nicer? That is, you don't want to see lots of experimental warnings when running the tests? Is there any other reason it is bad to print warnings in the tests?

The reason is to make test output better. It is important to reduce noise from test output in order to extract meaningful output easily.

Note: When running tests with `--processes=16`, the experimental feature warning is printed 16 times.

bkvogel:
OK, I can agree with the reason. Thanks for the PR. LGTM!

PR #1978, opened 2016-12-08
Fix memory API

okuta:
This PR provides appropriate function name (`free_all_blocks` instead of `free_all_free` ).

`PinnedMemoryPool` already use that name.
https://github.com/pfnet/chainer/blob/8fb2902a852bfa2755da71fe6e394bbc0dbfd613/cupy/cuda/pinned_memory.pxd#L46

### DIFF
--- a/cupy/cuda/memory.pyx
+++ b/cupy/cuda/memory.pyx
@@ -2,6 +2,7 @@
 
 import collections
 import ctypes
+import warnings
 import weakref
 
 import six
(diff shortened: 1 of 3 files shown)


cupy/cuda/memory.pyx
```diff
@@ -431,11 +432,19 @@ cdef class MemoryPool(object):
...
         """Release free blocks."""
         dev = device.get_device_id()
         self._pools[dev].free_all_free()
 
+    cpdef free_all_free(self):
+        """Release free blocks."""
+        warnings.warn(
+            'free_all_free is deprecated. Use free_all_blocks instead.',
+            DeprecationWarning)
+        self.free_all_blocs()
```
rezoo: s/self.free_all_blocs()/self.free_all_blocks()/

okuta:
Thank you reviewing.
I fixed.

rezoo:
Good. Then I would like to merge.

PR #1981, opened 2016-12-09
Add parameter size for LayerNormalization

loofahcus:
Add size(an optional parameter) in ``__init__()`` as the discussion in [pull#1950](https://github.com/pfnet/chainer/pull/1950)


bkvogel:
Thank you for the PR. LGTM!

PR #1990, opened 2016-12-12
Add __init__.py for cifar example

gwtnb:
The cifar example does not run on Python 2 without this PR.


bkvogel:
Thanks. I added an `__init__.py` file when developing the example but then somehow forgot to check it in. LGTM.

PR #1995, opened 2016-12-13
Fix GIL problem in load module methods

### LINKED ISSUE #1973: train_mnist using a GPU hangs after training finishes when the model is (a bit) large
The train_mnist example works perfectly when the model size is set to the default,
but it stops when a bit larger model (e.g. -u 3072) is used.

```
$  python train_mnist.py -g 0 -e 2 -u 3072 # the models is bit larger than the default
GPU: 0
# unit: 3072
# Minibatch-size: 100
# epoch: 2

     total [####################..............................] 41.67%
this epoch [#########################################.........] 83.33%
       500 iter, 0 epoch / 2 epochs
    111.19 iters/sec. Estimated time to finish: 0:00:03.362567.
```

The execution hangs up here and there's no way to quit it other than sending a KILL signal.
Please note that the GPU memory is large enough and thus no spilling of data/model happens.

```
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 367.48                 Driver Version: 367.48                    |
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|===============================+======================+======================|
|   0  Tesla M40           On   | 0000:04:00.0     Off |                    0 |
| N/A   32C    P0   160W / 250W |    506MiB / 11443MiB |     93%      Default |
+-------------------------------+----------------------+----------------------+

+-----------------------------------------------------------------------------+
| Processes:                                                       GPU Memory |
|  GPU       PID  Type  Process name                               Usage      |
|=============================================================================|
|    0     24813    C   python                                         504MiB |
+-----------------------------------------------------------------------------+
```

Things I found (env1 and env2 are described below):
1. It happens in multiple environments using different GPUs and different CUDA versions.
2. In env1, removing extensions that touch the data in the GPU (namely snapshot and logreport) solves the problem.
3. However in env2, chainer hangs even when all extensions are removed. What is worse, it stops after training.run() finishes and there's nothing to do in the python level.
4. The same happens with my own chainer program that uses completely different data/model from the mnist example. It works when the model is small, but it hangs up when the number of nodes are increased.

Judging from these facts (especially from 2 and 3), **I guess there's a problem somewhere in an asynchronous cuda kernel, maybe the one that copies data back from GPU to CPU**.

||env1|env2|
|:--:|:--:|:--:|
|GPU|Tesla M40 (Maxwell, 12GB mem) |GRID K520 (Kepler, 4GB mem) |
|CPU|Intel Xeon E5-2690 v4 (Broadwell) |Intel Xeon E5-2670 (SandyBridge)*|
|SW versions|CUDA 8.0, cuDNN 5.1| CUDA 7.5, cuDNN 5.0|
|Misc||g2.2xlarge instance of AWS|

\* The CPU model of env2 might be reported from the virtual machine monitor and it can be different from actual physical CPU.

okuta:
Fix #1973.

### DIFF
--- a/cupy/cuda/driver.pyx
+++ b/cupy/cuda/driver.pyx
@@ -24,12 +24,13 @@ cdef extern from "cupy_cuda.h":
     int cuGetErrorString(Result error, const char** pStr)
 
     # Module load and kernel execution
-    int cuModuleLoad(Module* module, char* fname)
-    int cuModuleLoadData(Module* module, void* image)
-    int cuModuleUnload(Module hmod)
-    int cuModuleGetFunction(Function* hfunc, Module hmod, char* name)
+    int cuModuleLoad(Module* module, char* fname) nogil
+    int cuModuleLoadData(Module* module, void* image) nogil
+    int cuModuleUnload(Module hmod) nogil
+    int cuModuleGetFunction(Function* hfunc, Module hmod,
+                            char* name) nogil
     int cuModuleGetGlobal(Deviceptr* dptr, size_t* bytes, Module hmod,
-                          char* name)
+                          char* name) nogil
     int cuLaunchKernel(
         Function f, unsigned int gridDimX, unsigned int gridDimY,
         unsigned int gridDimZ, unsigned int blockDimX,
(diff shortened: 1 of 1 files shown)


gwtnb:
LGTM

PR #1998, opened 2016-12-13
Revert "Added cupy functions at indexing generate.py"

okuta:
Reverts pfnet/chainer#1018

Because that PR's document contains complete copy from NumPy.


beam2d:
Thanks.

@hisa0507 Sorry for the revert. I found that the original PR contains a large portion of copy of the document from NumPy. Can you make a PR again with fixed documents? The document should explain what the CuPy functionalities support. You can add a reference to the NumPy document to explain that a functionality is same as that of NumPy. If the support of the functionality is partial (it looks so), you should explicitly explain which part is supported and which part is not. I left the PR branch of this revert (revert-1018-add_cupy_func_generate), so if you already deleted your own PR branch, you can use it to recover the original changes.

PR #2000, opened 2016-12-13
Improve imagenet example performance

okuta:
The dtype of `mean` is `numpy.float32`.
Division operation is slow.


examples/imagenet/train_imagenet.py
```diff
@@ -61,7 +61,7 @@ def get_example(self, i):
 
         image = image[:, top:bottom, left:right]
         image -= self.mean[:, top:bottom, left:right]
-        image /= 255
+        image *= (1.0 / 255.0)
```
zori: Unrelated to  speed-up, could you add a comment above `# Scale to [0, 1]`. Although trivial, the other pre-processing steps are documented.

okuta: OK.

zori:
LGTM! #2000 :)

okuta:
Thank you!

PR #2001, opened 2016-12-13
add GraphReport

Kiikurage:
This PR adds GraphReport Extension. This extension outputs graph of logs. Example is [here](https://github.com/Kiikurage/graph_report).


mitmul:
Thank you for sending a PR! Could you add a test code for this extension under `chainer/tests/chaienr_tests/training_tests/extensions_tests` ?

mitmul:
@Kiikurage As for the name of this extension, I'd like to suggest "PlotReport" because the term "Graph" seems to be a bit confusing with the computational graph. I'd be happy if you could consider this naming :)

chainer/training/extensions/graph_report.py
```diff
@@ -0,0 +1,141 @@
+from chainer import reporter
+import chainer.serializer as serializer_module
+from chainer.training import extension
+import chainer.training.trigger as trigger_module
+import json
+import numpy as np
```
mitmul: Could you make it `import numpy` and replace all `np` with `numpy` in this source file?

chainer/training/extensions/graph_report.py
```diff
@@ -0,0 +1,141 @@
+from chainer import reporter
```
mitmul: Could you move this line to the bottom of the import block? Because Import statements must be organized into three parts in this order: standard libraries, third-party libraries, and internal imports. I'd be happy if you would check this contribution guide here :) http://docs.chainer.org/en/stable/contribution.html

chainer/training/extensions/graph_report.py
```diff
@@ -0,0 +1,141 @@
...
+import chainer.serializer as serializer_module
```
mitmul: ditto.

chainer/training/extensions/graph_report.py
```diff
@@ -0,0 +1,141 @@
...
+from chainer.training import extension
```
mitmul: ditto.

chainer/training/extensions/graph_report.py
```diff
@@ -0,0 +1,141 @@
...
+import chainer.training.trigger as trigger_module
```
mitmul: ditto.

chainer/training/extensions/graph_report.py
```diff
@@ -0,0 +1,141 @@
...
+import json
```
mitmul: Could you move this line to the top of this file? The reason is ditto to the former lines (according to the contribution guide).

chainer/training/extensions/graph_report.py
```diff
@@ -0,0 +1,141 @@
...
+import numpy as np
+from os import path
```
mitmul: Could you move this line to the top of this file? The reason is ditto to the former lines (according to the contribution guide).

chainer/training/extensions/graph_report.py
```diff
@@ -0,0 +1,141 @@
...
+import six
```
mitmul: ditto.

mitmul:
@Kiikurage Could you add a documentation index of this extension to `docs/source/reference/extensions.rst` ?

chainer/training/extensions/graph_report.py
```diff
@@ -0,0 +1,141 @@
...
+
+  $ pip install matplotlib'''
+        raise RuntimeError(msg)
+
+
+class GraphReport(extension.Extension):
+
+    def __init__(self, y_keys, x_key='iteration', trigger=(1, 'epoch'),
+                 postprocess=None, file_name='graph.png'):
+        """Trainer extension to output graph.
```
mitmul: Could you move this docstring to the right after the class name (outside the `__init__` function)?

mitmul: Could you move this docstring to the right after the class name (outside the `__init__` function)?

mitmul: How about "output plots" instead of "output graph"?

chainer/training/extensions/graph_report.py
```diff
@@ -0,0 +1,141 @@
...
+        if self._trigger(trainer):
+            stats = self._summary.compute_mean()
+            stats_cpu = {}
+            for name, value in six.iteritems(stats):
+                stats_cpu[name] = float(value)  # copy to CPU
+
+            updater = trainer.updater
+            stats_cpu['epoch'] = updater.epoch
+            stats_cpu['iteration'] = updater.iteration
+            x = stats_cpu[self._x_key]
```
mitmul: How about replacing these three lines (L100 to L102) with just one line as below?
```
x = updater.epoch if self._x_key == 'epoch' else updater.iteration
```

chainer/training/extensions/graph_report.py
```diff
@@ -0,0 +1,141 @@
...
+                    self._postprocess(f, a, summary)
+
+                f.savefig(path.join(trainer.out, self._file_name))
+
+            plot.close()
+            self._init_summary()
+
+    def serialize(self, serializer):
+        if isinstance(serializer, serializer_module.Serializer):
+            serializer('_graph_{}'.format(self._file_name),
```
mitmul: How about `_plot` instead of `_graph` if you would accept the name change suggested above?

chainer/training/extensions/graph_report.py
```diff
@@ -0,0 +1,141 @@
...
+                       json.dumps(self._data))
+
+        else:
+            self._data = json.loads(
+                serializer('_graph_{}'.format(self._file_name), ''))
```
mitmul: Ditto.

chainer/training/extensions/graph_report.py
```diff
@@ -0,0 +1,141 @@
...
+            data = self._data
+
+            for k in keys:
+                if k in stats_cpu:
+                    data[k].append((x, stats_cpu[k]))
+
+            f = plot.figure()
+            a = f.add_subplot(111)
+
+            flag_any_item = False
```
mitmul: It seems not to be necessary if you use `len(xy)` to check there's any array that has been drawn.

chainer/training/extensions/graph_report.py
```diff
@@ -0,0 +1,141 @@
...
+            for k in keys:
+                xy = data[k]
+                if len(xy) == 0:
+                    continue
+
+                flag_any_item = True
+                xy = np.array(xy)
+                a.plot(xy[:, 0], xy[:, 1], label=k)
+
+            if flag_any_item:
```
mitmul: It could be `len(xy) > 0` instead. The flag variable `flag_any_item` isn't needed in this case.

chainer/training/extensions/graph_report.py
```diff
@@ -0,0 +1,141 @@
...
+            f = plot.figure()
+            a = f.add_subplot(111)
+
+            flag_any_item = False
+            for k in keys:
+                xy = data[k]
+                if len(xy) == 0:
+                    continue
+
+                flag_any_item = True
```
mitmul: Ditto.

chainer/training/extensions/graph_report.py
```diff
@@ -0,0 +1,141 @@
...
+                xy = np.array(xy)
+                a.plot(xy[:, 0], xy[:, 1], label=k)
+
+            if flag_any_item:
+                if self._postprocess is not None:
+                    self._postprocess(f, a, summary)
+
```
mitmul: Could you remove this blank line?

Kiikurage:
Thank You for careful reviews, and I also think the proposed name "PlotReport" is much better than the current name. I'll change them.

chainer/training/extensions/graph_report.py
Kiikurage: In this case, I want to check whether one or more plot have been drawn or not. `len(xy)` can check whether each item should be drwan or not, but it not enough to my puporse. Or should I write like `numpy.any([len(data[k]) > 0 for k in keys])` ?

chainer/training/extensions/plot_report.py
```diff
@@ -0,0 +1,143 @@
...
+            for k in keys:
+                xy = data[k]
+                if len(xy) == 0:
+                    continue
+
+                flag_any_item = True
+                xy = numpy.array(xy)
+                a.plot(xy[:, 0], xy[:, 1], label=k)
+
+            if flag_any_item:
```
Kiikurage: > It could be len(xy) > 0 instead. The flag variable flag_any_item isn't needed in this case.

In this case, I want to check whether one or more plot have been drawn or not. len(xy) can check whether each item should be drwan or not, but it not enough to my puporse. Or should I write like numpy.any([len(data[k]) > 0 for k in keys]) ?

chainer/training/extensions/graph_report.py
mitmul: Ah, sorry, checking keys is also needed. How about this?
if keys and len(xy) > o:

chainer/training/extensions/plot_report.py
mitmul: Ah, sorry. How about using `a.has_data()` ? I think if `a` was used to plot `xy` once, `a.has_data()` will be True, otherwise (nothing has been drawn), it'll be False.

chainer/training/extensions/__init__.py
```diff
@@ -2,6 +2,7 @@
 from chainer.training.extensions import computational_graph  # NOQA
 from chainer.training.extensions import evaluator  # NOQA
 from chainer.training.extensions import exponential_shift  # NOQA
+from chainer.training.extensions import plot_report  # NOQA
```
mitmul: Thank you for changing the name. I think this name starts with "p", so it should be located one line before `print_report` at line 8.

chainer/training/extensions/plot_report.py
```diff
@@ -0,0 +1,143 @@
+import json
+import numpy
```
mitmul: Sorry for my confusing comments. `numpy` is a third party package, so it should be placed after this standard package part. Considering the alphabetical order, it's better to be placed right before `import six` :)

mitmul:
Thank you for updates. Could you merge this code with the master branch? There seems to be some conflicts.

Kiikurage:
ok, I fixed all points of your reviews.

mitmul:
Thank you for fixing all points! LGTM 👍

PR #2003, opened 2016-12-13
Add SetItem function for __setitem__ of Variable

wkentaro:
Inverse operation of GetItem function: https://github.com/pfnet/chainer/pull/1154

### DIFF
--- a/chainer/__init__.py
+++ b/chainer/__init__.py
@@ -129,5 +129,6 @@ class DebugMode(object):
 
 basic_math.install_variable_arithmetics()
 array.get_item.install_variable_get_item()
+array.set_item.install_variable_set_item()
 
 disable_experimental_feature_warning = False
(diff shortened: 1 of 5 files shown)


beam2d:
Thank you for the PR, but I don't think this setitem works correctly with backprop. For example, consider a statement `x[i] = v` for Variables x and v with some slices i. The resulting x has an updated array but the graph rooted with x is not modified (I mean, x does not depend on v on the graph even after the assignment). Thus, when we try to compute a gradient of a loss function `f(x)`, the error is not propagated to v, although the loss actually depends on v.

One possible idea to resolve this problem is to modify the graph by replacing the node x by its copy and inserting the SetItem node to the graph as a new creator of the original x.

Another concern is that some Functions (e.g. sigmoid, tanh, and cuDNN relu) keep its output array for backprop and assume that the array is not modified, and so applying setitem to the output of these Functions break the correctness of backprop. It would be very difficult to detect such a misuse.

PR #2024, opened 2016-12-16
call model.to_gpu() in StandardUpdater

td2sk:
Learning with GPU,  it is necessary to execute model.to_gpu() explicitly for StandardUpdater. (Otherwise we will get an array type error).
However, using ParallelUpdater, calling model.to_gpu() causes an CUDA error.

Since this behavior is difficult to understand, I think that StandardUpdater should also call model.to_gpu() automatically.

### DIFF
--- a/chainer/training/updater.py
+++ b/chainer/training/updater.py
@@ -127,6 +127,9 @@ class StandardUpdater(Updater):
             optimizer = {'main': optimizer}
         self._optimizers = optimizer
 
+        if device and device >= 0:
+            optimizer['main'].target.to_gpu(device)
+
         self.converter = converter
         self.loss_func = loss_func
         self.device = device
--- a/chainer/training/updater.py
+++ b/chainer/training/updater.py
@@ -127,7 +127,7 @@ class StandardUpdater(Updater):
             optimizer = {'main': optimizer}
         self._optimizers = optimizer
 
-        if device and device >= 0:
+        if device is not None and device >= 0:
             optimizer['main'].target.to_gpu(device)
 
         self.converter = converter


chainer/training/updater.py
```diff
@@ -127,6 +127,9 @@ def __init__(self, iterator, optimizer, converter=convert.concat_examples,
             optimizer = {'main': optimizer}
         self._optimizers = optimizer
 
+        if device and device >= 0:
```
ir5: This line does not work correctly if `device == 0`. I believe that this should be `if device is not None and device >= 0:`.

td2sk:
The only way to calculate on GPU with ParallelUpdater is to pass the GPU ID to the devices argument of ParallelUpdater.
(In fact, ParallelUpdater has an internal copy of the models to deploy to each GPU, we have to access the private variable to put these on the GPU without using the devices option.)
If we call explicitly calls model.to_gpu(), a CUDA error will occur.

On the other hand, StandardUpdater uses the device option only for processing mini batches, so the user must explicitly call model.to_gpu ().
(Otherwise type error will occur.)


|  | StandardUpdater | ParallelUpdater |
|-----------|------------|------------|
| call model.to_gpu() in user code | OK | CUDA error |
| not call model.to_gpu() in user code | type error | OK |


It is a cause of misunderstanding that it is different whether the user should call to_model () or not.
When my patches are merged, we are always required to specify the device option and the explicit model.to_gpu () is unnecessary.
Behavior changes as follows.

|  | StandardUpdater | ParallelUpdater |
|-----------|------------|------------|
| call model.to_gpu() in user code | CUDA error | CUDA error |
| not call model.to_gpu() in user code | OK | OK |

ir5:
Thank you for the detailed comment. Now I understood your point, and the PR looks good to me.

beam2d:
LGTM

PR #2025, opened 2016-12-16
Add 'with nogil' to cudaStreamWaitEvent call

gwtnb:
Ref: https://github.com/pfnet/chainer/issues/1973, https://github.com/pfnet/chainer/pull/1995


okuta:
LGTM!

PR #2027, opened 2016-12-16
Out-of-core training

anaruse:
I'm sending a PR for out-of-core training on GPU. This enables you to perform training of very deep models such as ResNet, DenseNet and etc with suitable min-batch size on GPU.

One concern about GPU is its memory size. The amount of GPU memory is usually smaller than that of CPU memory, especially on server machines. And it is not so uncommon that you see a kind of out-of-memory error while training very large models on GPU.

A basic idea behind this PR is to make use of CPU memory as a swap area. So, this PR provides a new method named _**to_swap()**_. You can use the  _**to_swap()**_ to move any variables on GPU memory to CPU memory. A method _**to_gpu()**_ is enhanced and you can use the _**to_gpu()**_ to move variables from CPU memory back to GPU memory. And this PR also provides several methods such as _**ancestors_to_swap()**_, _**ancestors_to_gpu()**_, _**interrupt_backward()**_, _**resume_backward()**_, _**enable_swapout()**_, _**disable_swapout()**_ and so on that are useful to realize appropriate data movement during forward and backward propagation.

For its usage, please refer to a file **examples/example_OOC.py**.


anaruse:
I've added a few methods so that it is easier to use OOC feature.
The most important fix with this patch is a **set_end_of_sub_graph()** method added to the Variable class. This is a kind of hint from developers to a Chainer's runtime to let the runtime knows when to move data safely between CPU and GPU. Still, you need to modify existing models, but, thanks to the **set_end_of_sub_graph()** method, the amount of modification is considerably reduced, and I think now it is kind of acceptable for many developers.
For its usage, please refer to files examples/example_OOC_(simple/VGG16/ResNet152).py.

anaruse:
Added a method _**forget()**_ to the Variable class and modified back-prop methods in order to reduce memory usage. This was implemented based on the result of discussion with Chainer developers such as Okuta-san, Tokui-san, Unno-san. Thanks for the discussion!

The attached is the effect of the out-of-core training combined with the forget() method. As you can see, GPU memory usage is greatly reduced by this and very deep network like ResNet152 can be trained with large batch size. In other words, you can deal with much deeper and larger networks than ResNet152, on a single GPU.
image

For its usage, please refer to files examples/example_OOC_(simple/VGG16/ResNet152).py.

Issue #2006, opened 2016-12-15
Add tests on Trainer's extension handling


muupan:
Currently there is no tests on  how `Trainer` handles extensions via `Trainer.extend` and `Trainer.run`. Such tests are needed to confirm changes like https://github.com/pfnet/chainer/pull/1930 are safe.

One can use a mock updater to write such tests as in https://github.com/pfnet/chainer/blob/master/tests/chainer_tests/training_tests/test_trainer.py.

muupan:
Solved by #2032

PR #2032, opened 2016-12-18
Add tests for  how extensions are executed (issue #2006)

soramichi:
This PR adds tests for how extensions are executed (in response to issue #2006)

What this test checks:
- An extension inheriting `extension.Extension` is properly called and finalized.
- An extension created by `make_extension` is properly called and finalized.
- Two extensions created by `make_extension` are executed in the correct order when the default priority is used.
- Two extensions created by `make_extension` are executed in the correct order when the priorities are specified in `make_extension`.

Because there are many ways to set configurations of an extension (specifying in `trainer.extend`, specifying in `make_extension`, and having configs as class members of an extension), testing all possible paths in a clean test code is difficult.  It might be a good idea to change the original code to provide a single way to specify configurations of an extension.

I can also add tests for extensions created with lambda functions and callables not inheriting `extension.Extension`, once PR #1930 is merged.

### DIFF
--- a/tests/chainer_tests/training_tests/test_trainer.py
+++ b/tests/chainer_tests/training_tests/test_trainer.py
@@ -11,6 +11,19 @@ from chainer import testing
 from chainer import training
 
 
+class DummyExtension(training.extension.Extension):
+
+    def __init__(self):
+        self.is_called = False
+        self.is_finalized = False
+
+    def __call__(self, trainer):
+        self.is_called = True
+
+    def finalize(self):
+        self.is_finalized = True
+
+
 class TestTrainerElapsedTime(unittest.TestCase):
 
     def setUp(self):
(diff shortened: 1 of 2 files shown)


muupan:
Great, thanks for the PR! Do you mind if we first merge this PR and then merge #1930 with its tests?

soramichi:
Sure, then I'll add some tests to #1930 after this PR is merged.

muupan:
Since these new test cases are added to check `Trainer`'s functionality of correctly handling extensions, I suggest moving them to `tests/chainer_tests/training_tests/test_trainer.py`. The test cases themselves look good.

soramichi:
Hi, thank you for the suggestion. I move the newly added tests to `test_trainer.py` from `test_extension.py`

tests/chainer_tests/training_tests/test_trainer.py
```diff
@@ -11,6 +11,19 @@
...
+        self.is_finalized = False
+
+    def __call__(self, trainer):
+        self.is_called = True
+
+    def finalize(self):
+        self.is_finalized = True
+
+
 class TestTrainerElapsedTime(unittest.TestCase):
```
muupan: Now that this class not only for testing elapsed time, could you rename the class name, say, to `TestTrainer`?

muupan:

soramichi:
Renamed the test class as suggested. Thanks for the comment and for checking the CI result.

PR #2037, opened 2016-12-22
Fix test code for convolution_2d

zori:
'assert_called_once' is not a method, instead directly check the 'call_count' property. Use 'autospec'.

Related to https://github.com/pfnet/chainer/pull/2011/


okuta:
LGTM!

Issue #2042, opened 2016-12-22
Support advanced indexing for chainer.Variable's __getitem__


yuyu2172:
I would like to discuss how to support advanced indexing of `chainer.Variable` and some addition that is necessary for `cupy`'s API.

## Current Status
+ `chainer.Variable.__getitem__` supports basic indexing (combination of `slice(None)` (i.e. `:`), `None`, integer, slice, and Ellipsis).
+ It does not support integer array as index. So, `x[[1,2]]` is not supported. Also, it does not support multiple array as indexing. (e.g.. `x[[1, 2], [3, 1]]`)
+ It does not support boolean mask indexing.
+ It does not support combination of basic indexing and array indexing.

## What needs to be added
For forward pass, it is pretty much done because `cupy.__getitem__` supports most of the advanced indexing and combination of basic/adv indexing.

For backward pass, additional `cupy` function is necessary.
Putting the topic in context, here is an example for basic indexing. When we have forward pass as `y = x[0]`, then backward pass would be `gx = zeros_like(x); gx[0] = gy`. 

The basic indexing uses `__setitem__` to propagate gradients, however it is not the case when we use integer array index and the index has duplicate elements.
For example, `y = x[[1, 0, 1]]` has backward pass `gx[0] = gy[1]; gx[1] = gy[0] + gy[2]`, and it can not be expressed by `__setitem__`.
`gx` is calculated using `i=[1, 0, 1]` and `gy` by `gx[i[k]] += gy[k]` for `k=1,2,3`.
I would like to call a function that does the operation `scatter_add`, and this function is missing in `cupy`. 

The backward pass of `Variable.__getitem__` is calculated similarly in the case when there are multiple arrays or combination of basic/advanced indexing.
Here is a prototype for forward/backward pass for `Variable.__getitem__`.

```python
class GetItem(function.Fucntion): 
    def __init__(self, slices):
        self.slices = slices

    def forward(self, xs):
         ary = xs[0]
         return ary[self.slices],

    def backward(self, xs, gys):
         gy = gys[0]
         gx = zeros_like(xs[0])
         scatter_add(gx, self.slices, gy)
         return gx,
```

## Addition to CuPy API

In order to implement above functionality, `scatter_add` needs to be implemented. It should have identical interface as `__setitem__`, which takes tuple `slices` as an argument.

muupan:
Resolved by #2203

PR #2040, opened 2016-12-22
make ndarray.__setitem__ to handle one integer array

yuyu2172:
Related to #1919 which the reviewer and I concluded unnecessary addition to API.

In this PR, I modified `ndarray.__setitem__` to accept argument `slices` with following items.
+ one integer array
+ combination of basic and an integer array

I did not support multiple integer arrays in `slices` in this PR, but I am planning to send PR that supports them.

Mechanism to handle one dimensional array is same as #1919.
The part of the code that processes `slices` is mostly same as #1863.

PR #2043, opened 2016-12-22
Add cupy.scatter_add with interface identical to ndarray.__setitem__

yuyu2172:
Merge after #2040.

Related to #1920. I found the interface of `scatter_add` in #1920 to be insufficient for implementing backward pass of `chainer.Variable.__getitem__` (#2042).

This implements the function discussed in #2042.
This PR adds
+ `scatter_add` that handles one integer array
+ `scatter_add` that handles basic indexing
+ `scatter_add` that handles combination of basic indexing and an integer array.

This does not support slices with multiple integer arrays.

### DIFF
--- a/cupy/core/core.pyx
+++ b/cupy/core/core.pyx
@@ -2211,9 +2211,12 @@ cpdef _scatter_op_single(ndarray a, ndarray indices, v, int axis=0, op=''):
         _scatter_update_kernel(
             v, indices, cdim, rdim, adim, a.reduced_view())
     elif op == 'add':
-        if not issubclass(v.dtype.type, (numpy.int32, numpy.float32)):
+        if not issubclass(v.dtype.type,
+                          (numpy.int32, numpy.float32,
+                           numpy.uint32, numpy.uint64, numpy.ulonglong)):
             raise TypeError(
-                'scatter_add only supports int32 and float32 as data type')
+                'scatter_add only supports int32, float32, uint32, uint64 as'
+                'data type')
         _scatter_add_kernel(
             v, indices, cdim, rdim, adim, a.reduced_view())
     else:
--- a/cupy/core/core.pyx
+++ b/cupy/core/core.pyx
@@ -2211,6 +2211,8 @@ cpdef _scatter_op_single(ndarray a, ndarray indices, v, int axis=0, op=''):
         _scatter_update_kernel(
             v, indices, cdim, rdim, adim, a.reduced_view())
     elif op == 'add':
+        # There is constraints on types because atomicAdd() in CUDA 7.5
+        # only supports int32, uint32, uint64, and float32.
         if not issubclass(v.dtype.type,
                           (numpy.int32, numpy.float32,
                            numpy.uint32, numpy.uint64, numpy.ulonglong)):
--- a/cupy/core/core.pyx
+++ b/cupy/core/core.pyx
@@ -1247,106 +1247,16 @@ cdef class ndarray:
             array([9998., 9999.])
 
         """
-        cdef Py_ssize_t i, ndim, n_newaxes, n_ellipses, ellipsis, axis
-        cdef Py_ssize_t ellipsis_size
-        cdef ndarray a, v, x, y
+        _scatter_op(self, slices, value, 'update')
 
-        if not isinstance(slices, tuple):
-            slices = [slices]
-        else:
-            slices = list(slices)
+    def scatter_add(self, slices, value):
+        """Adds given values to specified elements of an array.
 
-        # Expand ellipsis into empty slices
-        ellipsis = -1
-        n_newaxes, n_ellipses = 0, 0
-        for i, s in enumerate(slices):
-            if s is None:
-                n_newaxes += 1
-            elif s is Ellipsis:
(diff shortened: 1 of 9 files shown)


rezoo:
Thank you for your commits. I would like to check this after merging #2040.

yuyu2172:
I rebased master branch.

cupy/ext/scatter.py
```diff
@@ -0,0 +1,42 @@
+def scatter_add(a, slices, value):
+    """Adds given values to specified elements of an array.
+
+    This function adds given ``value`` to specified elements of ``a``.
+    When all of the indices target different locations, :func:`scatter_add`
+    does operation equivalent to ``a[slices] = a[slices] + value``.
+    When there are indices targeting the same location, :func:`scatter_add`
+    uses all of the values for addition. On the other hand,
```
rezoo: If there are multiple elements indicating a same location, :func:`scatter_add` uses all of these values for addition.

cupy/ext/scatter.py
```diff
@@ -0,0 +1,42 @@
...
+    ``a[slices] = a[slices] + value`` only adds the contribution from one of
+    the indices targeting the same location.
+
+    Similarly to array indexing, negative indices are interpreted as counting
+    from the end of an axis.
```
rezoo: Note that just like an array indexing, negative indices are interpreted as counting from the end of an array.

cupy/core/core.pyx
```diff
@@ -2287,10 +2210,127 @@ cpdef _scatter_op_single(ndarray a, ndarray indices, v, int axis=0, op=''):
     if op == 'update':
         _scatter_update_kernel(
             v, indices, cdim, rdim, adim, a.reduced_view())
+    elif op == 'add':
+        if not issubclass(v.dtype.type, (numpy.int32, numpy.float32)):
+            raise TypeError(
+                'scatter_add only supports int32 and float32 as data type')
+        _scatter_add_kernel(
+            v, indices, cdim, rdim, adim, a.reduced_view())
```
yuyu2172: This is because those types are not supported by CUDA's atomicAdd. http://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#atomicadd
Since CuPy targets CUDA 7.5 and above, I removed support for double type even though it is supported by Pascal Architecture.

By the way, I added support for uint32 and uint64 that can be supported, but I forgot to.

tests/cupy_tests/core_tests/test_ndarray_scatter.py
```diff
@@ -0,0 +1,95 @@
...
+    {'shape': (3, 4, 5),
+     'slices': (slice(0, 1), None, [1, -1]), 'value': 1},
+    {'shape': (3, 4, 5),
+     'slices': (slice(0, 1), slice(1, 2), [1, -1]), 'value': 1},
+    {'shape': (3, 4, 5),
+     'slices': (slice(0, 1), None, slice(1, 2), [1, -1]), 'value': 1},
+    # broadcasting
+    {'shape': (3, 4, 5), 'slices': (slice(None), [[1, 2], [0, -1]],),
+     'value': numpy.arange(3 * 2 * 2 * 5).reshape(3, 2, 2, 5)},
+)
```
yuyu2172: There is a test case that handles `cupy.core.core.ndarray` indices. It is `test_scatter_add_cupy_argument` locating at line 75 of `test_ndarray_scatter.py`.

cupy/ext/scatter.py
yuyu2172: How about

If there are multiple elements targeting the same location, :func:`scatter_add` uses all of these values for addition.

cupy/ext/scatter.py
```diff
@@ -0,0 +1,42 @@
+def scatter_add(a, slices, value):
+    """Adds given values to specified elements of an array.
+
+    This function adds given ``value`` to specified elements of ``a``.
+    When all of the indices target different locations, :func:`scatter_add`
+    does operation equivalent to ``a[slices] = a[slices] + value``.
```
yuyu2172: Added

cupy/ext/scatter.py
yuyu2172: added

yuyu2172:
Thanks for comments. I fixed the code.

tests/cupy_tests/core_tests/test_ndarray_scatter.py
rezoo: I see.

cupy/ext/scatter.py
rezoo: OK.

cupy/core/core.pyx
rezoo: I see, but I still recommend you to add the following comment before the ``if`` statement:
```
# give some constraints on types because atomicAdd() in CUDA 7.5 only supports int32, int64, and float32.
```

rezoo:
I added some comments. Can you check them?

yuyu2172:
Thanks. I fixed the code.

rezoo:
Good. After passing all the tests I'd like to merge it.

PR #2049, opened 2016-12-26
Optimize cupy.asfortranarray

yuyu2172:
This PR optimizes `cupy.asfortranarray` by using cuBLAS's kernel when possible.
The kernel is called when input array is c_contiguous matrix and its dtype can be handled by cuBLAS.
I did not included support for complex arrays in this PR.

Here is an benchmark. The new implementation is 23x faster than the previous implementation.
The function defined in the code is roughly equivalent to what I wrote in the PR.

## Benchmark code

```python
# benchmark_asfortran.py

import cupy
import numpy as np
from cupy import cuda


def asfortranarray_old(a, dtype=None):
    ret = cupy.empty(a.shape[::-1], a.dtype if dtype is None else dtype).T
    ret[...] = a
    return ret


def asfortranarray_new(x):
    assert x.flags.c_contiguous
    m, n = x.shape
    out = cupy.empty(x.shape[::-1], x.dtype).T
    cupy.cuda.cublas.sgeam(
        cupy.cuda.Device().cublas_handle, 1, 1,
        m, n,
        alpha=1., A=x.data.ptr, lda=n, beta=0., B=x.data.ptr, ldb=n,
        C=out.data.ptr, ldc=m)
    return out



m = 2**15
n = 2**13
x = cupy.arange(m * n, dtype=np.float32).reshape(m, n)
with cuda.profile():
    for i in range(10):
        y_new = asfortranarray_new(x)
    for i in range(10):
        y_old = asfortranarray_old(x)

np.testing.assert_equal(y_new.get(), y_old.get())
```

## Result
Tested on Pascal TitanX, CUDA 8.0.
This can be checked by `nvprof --profile-from-start off python benchmark_asfortran.py`.
```
 ==27382== Profiling result:                                                                               
 95.99%  1.45644s        10  145.64ms  144.67ms  145.92ms  cupy_copy                                      
  4.01%  60.768ms        10  6.0768ms  6.0756ms  6.0791ms  void transpose_readWrite_alignment_kernel<float, float, int=1, bool=0, int=6, int=5, int=3>(cublasTransposeParams<float>, float const *, float*, float const *)                                                                                                    
  0.00%  1.2160us         1  1.2160us  1.2160us  1.2160us  [CUDA memcpy HtoD]                             
````

`cupy_copy` is the kernel called in the previous implementation, and `transpose_readWrite_alignment_kernel` is the kernel called in the new one.
You can see 23x speed up by looking at average execution times.
The [CUDA memcpy HtoD] is called when `cupy.cuda.Device().cublas_handle` is called for the first time.

### CODE REVIEW

File: cupy/manipulation/kind.py [hunk 1]
```diff
@@ -19,6 +21,27 @@ def asfortranarray(a, dtype=None):
...
+    if (a.flags.c_contiguous and
+            (a.dtype == numpy.float32 or a.dtype == numpy.float64) and
+            a.ndim == 2 and
+            dtype is None):
+        m, n = a.shape
+        if a.dtype == numpy.float32:
+            cupy.cuda.cublas.sgeam(
+                cupy.cuda.Device().cublas_handle,
+                1,  # transa
+                1,  # transb
```
rezoo: s/transa/transpose a/
s/transb/transpose ret/

cupy/manipulation/kind.py
```diff
@@ -19,6 +21,27 @@ def asfortranarray(a, dtype=None):
...
+                m, n, 1., a.data.ptr, n, 0., a.data.ptr, n,
+                ret.data.ptr, m)
+        elif a.dtype == numpy.float64:
+            cupy.cuda.cublas.dgeam(
+                cupy.cuda.Device().cublas_handle,
+                1,  # transa
+                1,  # transb
```
rezoo: Ditto.

cupy/manipulation/kind.py
```diff
@@ -19,6 +21,27 @@ def asfortranarray(a, dtype=None):
...
+                m, n, 1., a.data.ptr, n, 0., a.data.ptr, n,
+                ret.data.ptr, m)
+        return ret
```
rezoo: I like
```python
else:
    ret[...] = a
    return ret
```
because it is easier to understand the above function considers two cases.

rezoo:
@okuta If you do not find any problems, I would like to merge this PR.

okuta:
LGTM!

PR #2050, opened 2016-12-26
Make cuda.get_array_module support Variable

muupan:
This PR will make `cuda.get_array_module` support `Variable` arguments so that it will return the array module of `Variable.data`. It will also add tests for `cuda.get_array_module`.

### DIFF
--- a/chainer/cuda.py
+++ b/chainer/cuda.py
@@ -29,6 +29,7 @@ import warnings
 import numpy
 import six
 
+import chainer
 
 available = False
 cudnn_enabled = False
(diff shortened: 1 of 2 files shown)


chainer/cuda.py
```diff
@@ -555,6 +559,8 @@ def get_array_module(*args):
 
     """
     if available:
+        args = (arg.data if isinstance(arg, chainer.Variable) else arg
```
okuta: Please use list for performance.
`args = [arg.data if instance(arg, chainer.Variable) else arg for arg in args]`.

chainer/cuda.py
```diff
@@ -543,8 +544,11 @@ def reduce(in_params, out_params, map_expr, reduce_expr, post_map_expr,
 def get_array_module(*args):
     """Gets an appropriate one from :mod:`numpy` or :mod:`cupy`.
 
-    This is almost equivalent to :func:`cupy.get_array_module`. The only
-    difference is that this function can be used even if CUDA is not available.
+    This is almost equivalent to :func:`cupy.get_array_module`. The differences
+    are that this function can be used even if CUDA is not available and that
+    it will return their data arrays' array module for
```
okuta: typo?

muupan: You mean `.` at line 551?

okuta: Sorry, I misunderstood.

muupan:
Removed the unnecessary period and used list instead of tuple.

okuta:
LGTM!

PR #2062, opened 2016-12-27
Fix max pooling backward cpu

ronekko:
`MaxPooling2D.backward_cpu()` causes an error when I call the method more than once.

You can reproduce the error by the following code:
```
import numpy as np

x = np.ones((1, 1, 4, 4), dtype=np.float32)
h = F.max_pooling_2d(x, 2)
h.grad = np.ones_like(h.data)

h.backward()

try:
    h.backward()
except IndexError as e:
    print e  # index 16 is out of bounds for axis 1 with size 16

try:
    h.backward()
except IndexError as e:
    print e  # index 24 is out of bounds for axis 1 with size 16
```

We expect the code will print nothing, but the message as the comment will be shown.

The actual error message is like this:
```
  File "C:\Users\sakurai\Anaconda2\lib\site-packages\chainer\functions\pooling\max_pooling_2d.py", line 91, in backward_cpu
    gcol[indexes] = gy[0].ravel()
    
IndexError: index 16 is out of bounds for axis 1 with size 16
```

The current implementation of `forward_cpu()` has a probably unintended side effect at [L88-89](https://github.com/pfnet/chainer/blob/d85f9e501a183a3c39fdb4c3ba6e1c4d2d492be5/chainer/functions/pooling/max_pooling_2d.py#L88-L89)

```
indexes = self.indexes.ravel()
indexes += numpy.arange(0, indexes.size * kh * kw, kh * kw)
```

`ndarray.ravel()` might return the view of `self.indexes` instead of the copy and it will be broken by `+=` .
To avoid this, I replaced `ravel()` with `flatten()` which always return the copy.

`MaxPoolingND.backward_cpu()` has the same problem so I fixed it as well.

### DIFF
--- a/chainer/functions/pooling/max_pooling_2d.py
+++ b/chainer/functions/pooling/max_pooling_2d.py
@@ -85,7 +85,7 @@ class MaxPooling2D(pooling_2d.Pooling2D):
         gcol = numpy.zeros(
             (n * c * out_h * out_w * kh * kw), dtype=x[0].dtype)
 
-        indexes = self.indexes.ravel()
+        indexes = self.indexes.flatten()
         indexes += numpy.arange(0, indexes.size * kh * kw, kh * kw)
 
         gcol[indexes] = gy[0].ravel()
--- a/chainer/functions/pooling/max_pooling_nd.py
+++ b/chainer/functions/pooling/max_pooling_nd.py
@@ -87,7 +87,7 @@ class MaxPoolingND(pooling_nd._PoolingND):
 
         gcol = numpy.zeros(n * c * prod_outs * prod_ksize, dtype=x[0].dtype)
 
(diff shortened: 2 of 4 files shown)


ronekko:
There is a test failed but it seems that it is not related to this PR.

https://travis-ci.org/pfnet/chainer/jobs/186992459#L784-L808
`FAIL: test_backward_cpu (test_theano_function.TestTheanoFunctionTwoOutputs_param_1)  parameter: {'inputs': [{'shape': (3, 2), 'type': 'float32'}, {'shape': (2,), 'type': 'float32'}], 'outputs': [{'shape': (3, 2), 'type': 'float32'}, {'shape': (3, 2), 'type': 'float32'}]}`

okuta:
LGTM!

PR #1956, opened 2016-12-03
[WIP] Add Separable conv2d

fukatani:
This PR adds Separable convolution 2d Link. This is used by Xception model.

PR #2067, opened 2017-01-03
Supports depthwise convolution

fukatani:
This is PR for supporting depthwise convolution 2D.
Main usage for this layer is separable convolution.

(http://www.di.ens.fr/data/publications/papers/phd_sifre.pdf)
(https://arxiv.org/abs/1610.02357)

After merge this PR I wil restart separable convolution #1956.


chainer/functions/connection/depthwise_convolution_2d.py
```diff
@@ -0,0 +1,164 @@
...
+
+        xp = cuda.get_array_module(*x)
+        if xp is numpy:
+            self.col = conv.im2col_cpu(
+                x, kh, kw, self.sy, self.sx, self.ph, self.pw)
+        else:
+            self.col = conv.im2col_gpu(
+                x, kh, kw, self.sy, self.sx, self.ph, self.pw)
+
+        arys = [xp.tensordot(self.col[:, i, :, :, :, :], W[:, i, :, :],
```
mattya: Hello,
I think list comprehensions or for-loop is slow,
so how about using xp.matmul like following?

```python

# col.shape == (B,C,KY,KX,IY,IX)
# w.shape == (D,C,KY,KX)
a_ = col.transpose(1,0,4,5,2,3).reshape((C, B*IY*IX, KY*KX))
b_ = w.transpose(1,2,3,0).reshape((C, KY*KX, D))

c_ = xp.matmul(a_, b_)
# c_.shape == (C, B*IY*IX, D)
y = c_.reshape((C, B, IY, IX, D)).transpose(1,0,4,2,3).reshape((B, C*D, IY, IX))

```

chainer/functions/connection/depthwise_convolution_2d.py
```diff
@@ -0,0 +1,164 @@
...
+        gy = grad_outputs[0]
+        h, w = x.shape[2:]
+
+        xp = cuda.get_array_module(*x)
+        gy = xp.rollaxis(gy, 1, 4)
+        garys = xp.split(gy, W.shape[1], axis=3)
+        gW = xp.empty_like(W)
+        gcol = xp.empty_like(self.col)
+        gcol = xp.rollaxis(self.col, 0, 4)
+        for i in moves.range(W.shape[1]):
```
mattya: Also we can get better performance when using xp functions instead of for-loop here.

fukatani:
Thanks for your advice. :smiley: 
I have struggled to implement this PR without for-loop.

Now, I completed implementation.
But the part of GPU test need `cupy.matmul` for float16 and float64. (#2151)

fukatani:
I confirmed GPU test.
Could you check this PR?

docs/source/reference/links.rst
```diff
@@ -46,6 +46,10 @@ DeconvolutionND
 ~~~~~~~~~~~~~~~
 .. autoclass:: DeconvolutionND
 
+DepthwiseConvolutionND
```
mattya: Is it ```DepthwiseConvolution2D```?

docs/source/reference/links.rst
```diff
@@ -46,6 +46,10 @@ DeconvolutionND
...
+~~~~~~~~~~~~~~~~~~~~~~
+.. autoclass:: DepthwiseConvolutionND
```
mattya: ditto

fukatani:
Thank you for your comments. I fixed them.

chainer/functions/connection/depthwise_convolution_2d.py
```diff
@@ -0,0 +1,183 @@
...
+
+    Returns:
+        ~chainer.Variable:
+            Output variable. The shape is :math:`(n, c_I * c_M, h_O, w_O)`.
+
+    Like ``Convolution2D``, ``DepthwiseConvolution2D`` function computes
+    correlations between filters and patches of size :math:`(k_H, k_W)` in
+    ``x``.
+    But unlike ``Convolution2D``, ``DepthwiseConvolution2D`` not adds up
+    between input channels of filters but concatenates.
```
delta2323: concatenates them.

chainer/functions/connection/depthwise_convolution_2d.py
```diff
@@ -0,0 +1,183 @@
...
+    the output of convolution.
+
+    See: `L. Sifre. Rigid-motion scattering for image classification\
+          <http://www.di.ens.fr/data/publications/papers/phd_sifre.pdf>`_
+
+    .. seealso:: :class:`~chainer.links.DepthwiseConvolution2D`
+
+    .. admonition:: Example
+
+        >>> x = np.random.uniform(0, 1, (2, 3, 4, 7)).astype('f')
```
delta2323: Is `astype('f')` necessary? `np.random.uniform` returns `numpy.float64` and it should pass `check_type_forward` of this function.

chainer/functions/connection/depthwise_convolution_2d.py
```diff
@@ -0,0 +1,183 @@
...
+            gy = xp.rollaxis(gy, 1, 4)
+            gb = gy.sum(axis=(0, 1, 2))
+            return gx, gW, gb
+
+
+def depthwise_convolution_2d(x, W, b=None, stride=1, pad=0):
+    """Two-dimensional depthwise convolution function.
+
+    This is an implementation of two-dimensional depthwise convolution.
+    It takes three variables: the input image ``x``, the filter weight ``W``,
```
delta2323: two or three variables

chainer/functions/connection/depthwise_convolution_2d.py
```diff
@@ -0,0 +1,183 @@
...
+    and the bias vector ``b``.
```
delta2323: and optionally, the bias vector

chainer/functions/connection/depthwise_convolution_2d.py
```diff
@@ -0,0 +1,183 @@
...
+
+    Notation: here is a notation for dimensionalities.
+
+    - :math:`n` is the batch size.
+    - :math:`c_I` is the number of the input.
+    - :math:`c_M` is the number of the channel multiplier.
```
delta2323: Remove "the number of"

chainer/functions/connection/depthwise_convolution_2d.py
```diff
@@ -0,0 +1,183 @@
...
+            Bias variable of length :math:`c_M * c_I` (optional).
+        stride (int or pair of ints): Stride of filter applications.
+            ``stride=s`` and ``stride=(s, s)`` are equivalent.
+        pad (int or pair of ints): Spatial padding width for input arrays.
+            ``pad=p`` and ``pad=(p, p)`` are equivalent.
+
+
+    Returns:
+        ~chainer.Variable:
+            Output variable. The shape is :math:`(n, c_I * c_M, h_O, w_O)`.
```
delta2323: Its shape is ...

chainer/functions/connection/depthwise_convolution_2d.py
delta2323: remove "between"

chainer/functions/connection/depthwise_convolution_2d.py
```diff
@@ -0,0 +1,183 @@
...
+
+    Like ``Convolution2D``, ``DepthwiseConvolution2D`` function computes
+    correlations between filters and patches of size :math:`(k_H, k_W)` in
+    ``x``.
+    But unlike ``Convolution2D``, ``DepthwiseConvolution2D`` not adds up
+    between input channels of filters but concatenates.
+    For that reason, the shape of outputs of depthwise convolution are
+    :math:`(n, c_I * c_M, h_O, w_O)`, :math:`c_M` called channel_multiplier.
```
delta2323: For that reason, the shape of outputs of depthwise convolution are :math:`(n, c_I * c_M, h_O, w_O)`. :math:`c_M`  is called *channel multiplier*.

chainer/functions/connection/depthwise_convolution_2d.py
```diff
@@ -0,0 +1,183 @@
...
+
+    :math:`(h_O, w_O)` is determined by the equivalent equation with
+    ``Convolution2D``.
```
delta2323: by the equivalent equation of Convolution2D.

fukatani:
Thanks for your attentive review.
It will be helpful if you can see the identity with the paper.

fukatani:
Is this sentence natural English?
```
If ``None``, then this function uses to initialize ``bias``.
```
I think the following phrase is more correct.
```
If ``None``, then this link uses ``bias`` to initialize bias.
```

If so, I fix depthwise doc.
And some doc of links use same phrase, I'll fix them.

delta2323:
Ah, yes. It must be a typo.

chainer/functions/connection/depthwise_convolution_2d.py
```diff
@@ -0,0 +1,183 @@
...
+                x, kh, kw, self.sy, self.sx, self.ph, self.pw)
+        else:
+            self.col = conv.im2col_gpu(
+                x, kh, kw, self.sy, self.sx, self.ph, self.pw)
+
+        B, C, KY, KX, IY, IX = self.col.shape
+        D = W.shape[0]  # (D, C, KY, KX)
+        c_ = self.col.transpose(1, 0, 4, 5, 2, 3) \
+            .reshape((C, B * IY * IX, KY * KX))
+        w_ = W.transpose(1, 2, 3, 0).reshape((C, KY * KX, D))
```
delta2323: How do you ditermine the dimension ordering of W?  If the shape of W is (C, KY, KX, D), you.do not have to transpose W here.

chainer/functions/connection/depthwise_convolution_2d.py
```diff
@@ -0,0 +1,183 @@
...
+
+        # (C, B*IY*IX, KY*KX), (C, KY*KX, D)-> (C, B*IY*IX, D)
+        y = xp.matmul(c_, w_).astype(x.dtype, copy=False)
+
+        # (C, B*IY*IX, D) -> (B, C*D, IY, IX)
+        y = y.reshape((C, B, IY, IX, D)).transpose(1, 0, 4, 2, 3) \
```
delta2323: y.reshape(C, B IY*IX, D) is enough because you will separate IY and IX afterwards

chainer/functions/connection/depthwise_convolution_2d.py
fukatani: Current order is same as other convolutional function. (convolution_2d and dilated_convolution_2d).

Should I change it?

chainer/functions/connection/depthwise_convolution_2d.py
fukatani: Thanks! I fixed it.

chainer/functions/connection/depthwise_convolution_2d.py
delta2323: As far as I recognize, the benefit of aligning the dimension ordering of the depthwise convolution to ordinal convolution is to copy the trained weights to one link to the other. Do you  think there is such a use case? Also, let me know if there are other good points of the aligning their ordering.

The reason I would like to remove the transpose operation is mainly computation speed. So, if there is no significance performance difference even we change the dimension ordering, I prefer to keep the ordering as it is.

fukatani: As far as I know,  copy depthwise weight to original's don't exist.

1) Users may be confused by misunderstanding weight shape during debugging.
depthwise_convolution_2d and convolution_2d usage are similar, so user may expect similar order.

2) When visualizing weight, user can use common scripts. 

I know both of them is not big advantage, user can clear them by reading doc and reshape.
It seems better not to stick to align order.

fukatani:
I measured forward computation performance with the following script.

with transpose: 0.00181772509765625 sec
without transpose: 0.00179154296875 sec

The impact on backward computation time is even less.

```python
import cupy
from chainer.utils import conv

sy, sx = 1, 1
ph, pw = 0, 0

x = cupy.arange(300000).reshape(10, 3, 100, 100)
col = conv.im2col_gpu(x, 100, 100, sy, sx, ph, pw)
W1 = cupy.arange(90000).reshape(3, 3, 100, 100)
W2 = cupy.arange(90000).reshape(3, 100, 100, 3)

out_c, _, kh, kw = W1.shape
n, c, h, w = x.shape

out_h = conv.get_conv_outsize(h, kh, sy, ph)
out_w = conv.get_conv_outsize(w, kw, sx, pw)


def with_transpose():
    col = conv.im2col_gpu(x, kh, kw, sy, sx, ph, pw)
    B, C, KY, KX, IY, IX = col.shape
    D = W1.shape[0]  # (D, C, KY, KX)
    c_ = col.transpose(1, 0, 4, 5, 2, 3).reshape((C, B * IY * IX, KY * KX))
    w_ = W1.transpose(1, 2, 3, 0).reshape((C, KY * KX, D))

    y = cupy.matmul(c_, w_).astype(x.dtype, copy=False)
    return y.reshape((C, B, IY * IX, D)).transpose(1, 0, 3, 2).reshape((B, C * D, IY, IX))


def without_transpose():
    col = conv.im2col_gpu(x, kh, kw, sy, sx, ph, pw)
    B, C, KY, KX, IY, IX = col.shape
    D = W2.shape[3]  # (D, C, KY, KX)
    c_ = col.transpose(1, 0, 4, 5, 2, 3).reshape((C, B * IY * IX, KY * KX))
    w_ = W2.reshape((C, KY * KX, D))

    y = cupy.matmul(c_, w_).astype(x.dtype, copy=False)
    return y.reshape((C, B, IY * IX, D)).transpose(1, 0, 3, 2).reshape((B, C * D, IY, IX))


with_transpose()  # avoid initialize
without_transpose()  # avoid initialize

cnt = 1000
stream = cupy.cuda.Stream.null
start = stream.record()
for i in range(cnt):
    with_transpose()
end = stream.record()
end.synchronize()

print(cupy.cuda.get_elapsed_time(start, end) / 1000 / cnt)

start = stream.record()
for i in range(cnt):
    without_transpose()
end = stream.record()
end.synchronize()
print(cupy.cuda.get_elapsed_time(start, end) / 1000 / cnt
```

delta2323:
Thank you for your experiment! As I confirm that the introduction of transpose does not affect too much, I agree with your original implementation.

delta2323:
LGTM, I am running GPU tests. 

@mattya Should I merge this PR if I confirm this PR is merge-ready?

delta2323:
As you can see, currently Travis tests of this PR fail. But this error seems independent of this PR as I encountered the same problem in another PR #1754. I am now investigating the problem.

mattya:
@delta2323 I think it's ok to merge after solving the Travis problem.

delta2323:
As you can see, there occurs the following errors in Travis. This is because old NumPy (< 1.10) does not have `matmul` (see [reference](https://docs.scipy.org/doc/numpy/reference/generated/numpy.matmul.html)). As Chainer support NumPy older than 1.10, we should make this function work with it. We have several choices:

* Implement this function without `numpy.matmul`
* Check the version of NumPy and switch the implementation depending on the version.
* Check the version of NumPy and raise an error if NumPy is older than 1.10.

Could you consider the implement this function without `matmul`? If it is difficult, I think the workaround is to take the third option and support older NumPy in another PR.

```
ERROR: test_forward_cpu (test_simplified_dropconnect.TestSimplifiedDropconnect_param_0)  parameter: {'train': True, 'ratio': 0.0, 'W_dtype': <type 'numpy.float16'>, 'x_dtype': <type 'numpy.float16'>}
  File "/home/travis/build/pfnet/chainer/tests/chainer_tests/functions_tests/noise_tests/test_simplified_dropconnect.py", line 56, in test_forward_cpu
    self.check_forward(self.x, self.W, self.b)
  File "/home/travis/build/pfnet/chainer/tests/chainer_tests/functions_tests/noise_tests/test_simplified_dropconnect.py", line 52, in check_forward
    self.train)
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/chainer/functions/noise/simplified_dropconnect.py", line 148, in simplified_dropconnect
    return SimplifiedDropconnect(ratio, mask)(x, W, b)
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/chainer/function.py", line 199, in __call__
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/chainer/functions/noise/simplified_dropconnect.py", line 61, in forward
    y = xp.matmul(W, x[:, :, None])
AttributeError: 'module' object has no attribute 'matmul'
```

fukatani:
`einsum` can be a substitute for `matmul`.

Performance is almost the same.
As long as I benchmarked, `matmul` is only 0.5% faster (in numpy.1.11).

delta2323:
Thank you very much! I will run GPU tests and will merge this PR if all tests pass.

delta2323:
GPU test failed because of unit tests that seem irrelevant to this PR. Could you try to rebase or merge the latest master branch?

fukatani:
OK! I merged master branch.

delta2323:
Thank you for the update. I rerun the CI.

delta2323:

PR #2068, opened 2017-01-04
Add @testing.gpu

gwtnb:
Add `@testing.gpu` to ignore GPU tests.


delta2323:
LGTM. But I am wondering why non-existence of `@gpu` had not caused problems.

gwtnb:
It seems that test without GPU does not run cupy_tests on CI.

https://github.com/pfnet/chainer-test/blob/2ad03f1d1e0c52005fa4bdeb9c24f689b3dc4df2/test_cpu.sh#L8

delta2323:
I see. I was considering how we should detect such a mistake. Currently, there is no way to detect this except manual check.

Issue #1615, opened 2016-09-08
Parameter shapeholder does not work with link copy.


delta2323:
When we defer the initialization of parameters with the parameter shapeholder, they are created by the first forward propagation. If we copy the model before the first forward propagation, it does not create the parameter in the copied model. But as parameters are "marked" as initialized, there is no chance for copied model to create ones. This is problematic when we copy the link for evaluation, as we do in the examples.

Following code illustrates the problem

```
import chainer
from chainer import links as L
import numpy

x = numpy.random.uniform(size=(8, 16)).astype(numpy.float32)
x = chainer.Variable(x)

l = L.Linear(None, 4)
l_c = l.copy()  # (1)
l(x)  # (2)
l_c(x)
```

Traceback

```
Traceback (most recent call last):
  File "test.py", line 11, in <module>
    l_c(x)
  File "/Users/oonokenta/dev/chainer/chainer/links/connection/linear.py", line 86, in __call__
    return linear.linear(x, self.W, self.b)
AttributeError: 'Linear' object has no attribute 'W'
```

If we switch `(1)` and `(2)`, the error does not occur.

zori:
Can confirm. deepcopy-ing the model will create it's own set of uninitialized parameters. See https://github.com/pfnet/chainer/pull/1576

niboshi:
Closing as this has been fixed in #2072.

PR #2072, opened 2017-01-06
Support uninitialized variable and use it for uninitialized parameters

### LINKED ISSUE #1967: Uninitialized variable and parameter
I wrote a proposal of uninitialized variable and parameter. The uninitialized variable is a variable that does not have the data array. It can be used to represent an uninitialized parameter, which appears when using parameter shape placeholder (i.e., the weight matrix with unknown input size that will be inferred at the first iteration). It makes the handling of uninitialized parameters as easy as handling initialized parameters.

The details are explained in the following design document: https://docs.google.com/document/d/1-3fG8zs7jq3kM9jie1_No6beWn2jHtX1WqXWeXlhpCU/edit?usp=sharing. I don't think it breaks the compatibility, but it changes the way to implement parameter shape placeholder support, so it would be better to introduce it in v2.

beam2d:
Fix #1967. It includes the following changes:

- Variable supports uninitialized state, in which the data array is None.
- Link supports uninitialized variable as a parameter. Link.add_uninitialized_param and Link.has_uninitialied_params are **removed**.
- The implementations of links that use parameter shape placeholder (Convolution2D, Deconvolution2D, DilatedConvolution2D, LayerNormalization, Linear, and LSTM) are updated to use the new uninitialized parameters. The interface of them is not changed.

The two methods of Link are removed. I was first thinking about leaving them with deprecation warning, but I found that the existing code is broken even if I leave them (because the way of initializing the uninitialized parameter is changed: we had to call `link.add_param('param', shape)` in the previous design, while we have to call `link.param.initialize(shape)` in the new design). In this case, it is better to raise an error on using these methods (because the related code cannot run as expected), and so I decided to remove them.

Some methods of Variable are also updated to support the uninitialized parameter. Some of them are not explained in the original design document. I also updated the document of them, so I hope the changes are obvious.

### DIFF
--- a/chainer/variable.py
+++ b/chainer/variable.py
@@ -1,4 +1,5 @@
 import collections
+import copy
 import heapq
 import traceback
 import warnings
@@ -123,6 +124,11 @@ Actual: {0}'''.format(type(data))
         self.creator = None
 
         self.name = name
+
+    def __copy__(self):
+        copied = Variable()
+        copied.__dict__ = copy.copy(self.__dict__)
+        return copied
 
     def __reduce__(self):
         return Variable, (self.data, self.volatile, self.name, self._grad)
--- a/chainer/variable.py
+++ b/chainer/variable.py
@@ -10,6 +10,7 @@ import chainer
 from chainer import cuda
 from chainer import flag
 from chainer import utils
+from chainer import initializers
 
 
 def _check_grad_type(func, x, gx):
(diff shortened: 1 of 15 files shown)


takagi:
Does the implementation of Variable work as expected with copying when an uninitialized Variable has been initialized?

Let `W1` is an uninitialized Variable and `W2` is its copy:

    W1 = Variable()
    W2 = some_copy(W1)  # Some copy operation that expects the data is shared

then initialize `W1`:

    W1.initialize(some_shape)

As the result, the data of `W1` and that of `W2` would not be identical anymore:

    W1.data is W2.data  # False
    W2.data is None  # True

beam2d:
What is the "copy" operation that you expect the data to be shared? I thought that the variable should be initialized before applying such an operation.

takagi:
It is because I saw some issues relating to uninitialized parameters and their initialization timing, such as #1576, #1615 and #1691. What I mentioned might be too much. If a Variable is ensured to be initialized before it is copied as a part of a link, as you say, the latest implementation is enough.

bkvogel:
Should this feature be marked as an experimental API? Or should v2 features just be implicitly considered experimental until the beta or final release?

bkvogel:
Since this feature will break backward compatibility anyway, should we also consider cleaning up the constructor API by having the arguments that potentially depend on the input sizes to default to `None`?
For example, in `Linear` link, `in_size` could have default value `None` (its position would also need to change). Likewise, in `Convolution2D`, `in_channels` could default to `None`. Does this seem reasonable, or would you prefer to defer these kind of changes to a separate PR?

beam2d:
OK, I understand your point, and we should consider such use cases. It could be resolved by the following approach:

- Introduce `_uninitialized_copies` attribute to Variable, which is a list of (weak references to) variables copied from one uninitialized variable. This attribute is None if the variable is initialized. This list is shared by all copies (including the original one).
- When one of the variable in the list is initialized, all other copies are also initialized by the same array so that it behaves identical to variables copied after the initialization.

I will try to implement this feature.

beam2d:
@bkvogel We would need some discussions on the point if we should mark it as an experimental API.

As for cleaning up the constructors, I think it should be done by a separate PR.

takagi:
@beam2d I hava an implementation idea for it and will write the code with following your branch. Would you check it later?

beam2d:
Good! OK, then would you send a PR to this branch? Then I can check it.

takagi:
Sure~

takagi:
@beam2d I've send a PR, please check it. I hope I would have no misunderstanding about the expected behavior of Variable data and gradient arrays.

beam2d:
Thanks, I'll check that.

chainer/variable.py
```diff
@@ -76,26 +78,45 @@ class Variable(object):
...
+        if data is None:
+            self.initializer = (
+                initializers.NaN() if initializer is None else initializer)
+            dtype = getattr(self.initializer, 'dtype', numpy.float32)
+            self._grad_initializer = initializers.NaN(dtype)
+        elif not isinstance(data, (numpy.ndarray, cuda.ndarray)):
             msg = '''numpy.ndarray or cuda.ndarray are expected.
 Actual: {0}'''.format(type(data))
-            raise TypeError(msg)
+                raise TypeError(msg)
```
bkvogel: There seems to be an indentation error here.

bkvogel:
LGTM!

PR #2073, opened 2017-01-06
Improve PlotReport

mitmul:
This PR makes the plots created by PlotReport extension richer and adds this extension to the trainer in the MNIST example to show how to use this extension.

This PR actually adds legend to plots, and a label to x-axis, and grid to the figure.

After running `examples/mnist/train_mnist.py`, the below two plots will be saved in the result dir.

loss
accuracy

### DIFF
--- a/chainer/training/extensions/plot_report.py
+++ b/chainer/training/extensions/plot_report.py
@@ -1,13 +1,13 @@
 import json
-import warnings
 from os import path
+import warnings
 
 import numpy
 import six
 
 from chainer import reporter
-from chainer.training import extension
 import chainer.serializer as serializer_module
+from chainer.training import extension
 import chainer.training.trigger as trigger_module
 
 try:
--- a/chainer/training/extensions/plot_report.py
+++ b/chainer/training/extensions/plot_report.py
@@ -1,13 +1,12 @@
-import json
+from chainer import reporter
+from chainer.training import extension
 from os import path
 
+import chainer.serializer as serializer_module
+import chainer.training.trigger as trigger_module
+import json
 import numpy
 import six
-
-from chainer import reporter
-import chainer.serializer as serializer_module
-from chainer.training import extension
-import chainer.training.trigger as trigger_module
 
 try:
     import matplotlib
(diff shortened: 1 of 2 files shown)


chainer/training/extensions/plot_report.py
```diff
@@ -53,25 +54,27 @@ class PlotReport(extension.Extension):
...
+            object, Axes object, and all plot data are passed to this callback
+            in this order. This callback can modify the figure.
+        file_name (str): Name of the figure file under the output directory.
+            It can be a format string.
 
     """
 
     def __init__(self, y_keys, x_key='iteration', trigger=(1, 'epoch'),
-                 postprocess=None, file_name='graph.png'):
+                 postprocess=None, file_name='plot.png'):
```
ir5: It may not be a big deal, but I wonder whether we need to change the default file name here. Keeping the previous name to preserve backward compatibility may be better.

chainer/training/extensions/plot_report.py
```diff
@@ -23,28 +24,28 @@
 
 def _check_available():
     if not _available:
-        msg = '''matplotlib is not installed on your environment.
-Please install matplotlib to plot figure.
-
-  $ pip install matplotlib'''
-        raise RuntimeError(msg)
+        warnings.warn('matplotlib is not installed on your environment, '
+                      'so that nothing will be plotted at this time. '
```
ir5: (Not confident) "so that" -> so?

chainer/training/extensions/plot_report.py
```diff
@@ -23,28 +24,28 @@
...
+                      'Please install matplotlib to plot figures.\n\n'
+                      '$ pip install matplotlib')
```
ir5: (Minor comment) Adding two spaces at the first position and \n at the end of the line may make the message more readable.

chainer/training/extensions/plot_report.py
```diff
@@ -23,28 +24,28 @@
...
 
 
 class PlotReport(extension.Extension):
+
     """Trainer extension to output plots.
 
     This extension accumulates the observations of the trainer to
-    :class:`~chainer.DictSummary` at a regular interval specified by a
```
ir5: I think you did not have to change these docstrings.

chainer/training/extensions/plot_report.py
mitmul: I suggested to change the class name from "GraphReport" to "PlotReport" when I reviewed this PR: https://github.com/pfnet/chainer/pull/2001
But I forgot to suggest to change this filename. So this is a follow-up modification for consistency of the naming of this functionality.

chainer/training/extensions/plot_report.py
mitmul: Basically yes, but these lines have unnecessary spaces before changing lines.
So I removed those spaces to shrink the number of lines, but resulted in the same number of lines.

chainer/training/extensions/plot_report.py
ir5: Understood. Now it looks valid to me.

ir5:
LGTM

beam2d:
LGTM

PR #2075, opened 2017-01-07
Make astype handle conversion b/w np.longlong and np.int64

yuyu2172:
This PR fixes a bug in cupy.ndarray.astype that does not handle conversion between `np.dtype('q')` and `np.dtype('l')`. 

A simple failing case is as follows. The assertions pass for NumPy, but fails for CuPy.

```python 
import numpy as np
import cupy


def check_dtype_type(xp):
    x = xp.array([1,2,3], np.dtype('q'))
    assert x.dtype.type != np.dtype('l').type
    assert x.dtype.type == np.dtype('q').type

    xl = x.astype(np.dtype('l'))
    assert xl.dtype.type == np.dtype('l').type
    assert xl.dtype.type != np.dtype('q').type

check_dtype_type(np)
check_dtype_type(cupy)
```

This happens because equality operator is True for objects of type `numpy.dtype` (e.g. `x.dtype`, `np.dtype('q')`), but fails for type objects (`np.dtype('q').type`):

```python
>>> np.dtype('l') == np.dtype('q')
True
>>> np.dtype('l').type != np.dtype('q').type
False
```

### DIFF
--- a/cupy/core/core.pyx
+++ b/cupy/core/core.pyx
@@ -276,7 +276,7 @@ cdef class ndarray:
         """
         # TODO(beam2d): Support ordering, casting, and subok option
         dtype = numpy.dtype(dtype)
-        if dtype == self.dtype:
+        if dtype.type == self.dtype.type:
             if copy:
                 return self.copy()
             else:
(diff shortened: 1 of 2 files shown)


rezoo:
Good. LGTM.

PR #2084, opened 2017-01-11
Fix deadlock issue on multi-threaded application

koji123:
This PR is a simple fix for deadlock issue on multi-threaded application.

I added "with nogil" to the cudaStreamAddCallback API call and I moved mod.load(cubin) outside the file lock. My application needs this fix to make it work.

Maybe we should add "with nogil" to all CUDA API call but I'm not sure what influence they have.


okuta:
LGTM!

PR #2091, opened 2017-01-12
Fix test for cupy.pad by passing lists instead of passing ndarrays

gwtnb:
Old numpy does not accept array_like. [1.9](https://docs.scipy.org/doc/numpy-1.9.3/reference/generated/numpy.pad.html) [1.10](https://docs.scipy.org/doc/numpy-1.10.4/reference/generated/numpy.pad.html)


beam2d:
LGTM

PR #2097, opened 2017-01-13
Change name of kernels for scatter_* to be more consistent with other kernels

yuyu2172:
I changed name of kernels used for `scatter_*` to `cupy_scatter_*`.
These names are more consistent with other kernels called by CuPy (e.g. `cupy_take`).


gwtnb:
LGTM!

PR #2099, opened 2017-01-14
Support boolean array for cupy.ndarray.__setitem__

yuyu2172:
This PR supports boolean array for `cupy.ndarray.__setitem__`.

+ It supports boolean array indexing on axes (e.g. `a[:, np.array([True, False])]`).
+ It does not support combination of boolean arrays (e.g. `a[np.array([True, False]), np.array([False, True])]`).

Also, I changed name of internal function `_boolean_array_indexing` which was used for boolean array indexing of `__getitem__` to `_getitem_mask`. I did so because the name was vague, and it needed to be distinguishable from functions for `__setitem__`.

### DIFF
--- a/cupy/core/core.pyx
+++ b/cupy/core/core.pyx
@@ -2255,7 +2255,7 @@ cpdef _scatter_op_mask_single(ndarray a, ndarray mask, v, int axis, op):
 
     # broadcast v to shape determined by the mask
     mask_scanned = scan(mask.astype(numpy.int32).ravel())  # starts with 1
-    n_true = int(mask_scanned.max())
+    n_true = int(mask_scanned[-1])
     lshape = a.shape[:axis]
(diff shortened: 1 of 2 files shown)


rezoo:
Thank you! After reading the code, I concluded that there is no problem to merge this PR.
@okuta Could you briefly read the code and confirm that there is no problem? Just in case.

yuyu2172:
I restarted CI.

yuyu2172:
@rezoo 
I found the current implementation to fail in the case when `value` can not be broadcasted to `a`'s shape.
The failing cases are illustrated in 27cad5.

I fixed the code to work in those cases, so could you check it?

rezoo:
Good. Then I'll check this one.

rezoo:
Good. For computational efficiency this PR uses the two `scan` operators, and I think it is a good choice in the field of GPU community. LGTM.

PR #2101, opened 2017-01-14
Make copied Variable instances to share its data and gradient arrays.

takagi:
Rel. #2072.

I've modified `Variable` class to share its actual data and gradient arrays across its copied instances including the initialized/uninitialized state. The code may not be complete yet, but would be enough to get the idea.

I guess this behavior from the docstring of `link.copy` method:

    The whole hierarchy rooted by this link is copied. The copy is
    basically shallow, except that the parameter variables are also
    shallowly copied. It means that the parameter variables of copied one
    are different from ones of original link, while they share the data and
    gradient arrays.

If I had some misunderstanding and the copied Variable instances should still have mutually independent data and gradient arrays, please point it out. If what I guess is right and the change is reasonable, I will complete the change.

### DIFF
--- a/chainer/variable.py
+++ b/chainer/variable.py
@@ -111,7 +111,10 @@ class Variable(object):
 Actual: {0}'''.format(type(data))
                 raise TypeError(msg)
 
-        self.data = data
+        # Use a list as a data structure to hold the data array indirectly to
+        # abstract its initialized/uninitialized state.
+        self._data = [data]
+
         self.rank = 0
         self._volatile = flag.Flag(volatile)
 
@@ -200,6 +203,14 @@ Actual: {0}'''.format(type(data))
                              str(self.data.dtype))
 
     @property
+    def data(self):
+        return self._data[0]
+
+    @data.setter
+    def data(self, d):
+        self._data[0] = d
+
+    @property
     def grad(self):
         return self._grad
 
@@ -230,7 +241,7 @@ Actual: {0}'''.format(type(data))
         if self.data is None:
             self._initial_device = -1
         else:
-            self.data = cuda.to_cpu(self.data)
+            self._data = [cuda.to_cpu(self.data)]
             if self._grad is not None:
                 self._grad = cuda.to_cpu(self._grad)
 
@@ -247,7 +258,7 @@ Actual: {0}'''.format(type(data))
             self._initial_device = current if device is None else device
         else:
             with cuda.get_device(device):
-                self.data = cuda.to_gpu(self.data)
+                self._data = [cuda.to_gpu(self.data)]
                 if self._grad is not None:
                     self._grad = cuda.to_gpu(self._grad)
 
@@ -533,7 +544,7 @@ Actual: {0}'''.format(type(data))
             if grad is not None:
                 grad = cuda.to_gpu(grad, device=self._initial_device)
 
-        self.data = data
+        self._data[0] = data
         self.grad = grad
 
     def __lt__(self, other):
--- a/tests/chainer_tests/test_variable.py
+++ b/tests/chainer_tests/test_variable.py
@@ -1,3 +1,4 @@
+import copy
 import inspect
 import unittest
 
@@ -594,6 +595,13 @@ class TestUninitializedVariable(unittest.TestCase):
         x.to_gpu()
         x.to_cpu()
         self.check_constant_initialization(x, self.a, np)
+
+    def test_copy_to_initialize(self):
+        # This test intends the use case of link.copy() method.
+        x = chainer.Variable()
+        y = copy.copy(x)
+        x.initialize((3, 2))
+        self.assertIs(x.data, y.data)
 
     def test_cleargrad(self):
         x = chainer.Variable()


beam2d:
Thank you for the PR. It changes the behavior of some existing codes. For example, consider the following case.
```
l1 = Link(w=1)
l2 = l1.copy()
l2.to_gpu()
```
In the current Chainer, l1 is not copied to GPU, i.e., after to_gpu l1 and l2 have distinct parameter arrays. If my understanding is correct, after this PR being merged, l1 is also copied to GPU, since l1 and l2 refer the same parameter array even after to_gpu. I don't decide yet if this behavior should be kept as is, but at least the above code is useful for copying a parameter to another device without memory overheads.

takagi:
I got the case. It would be possible to think that there are two mixed problems: copy in host memory and transfer across host and device memory.

Although compatibility issues might be remained, I think we would be able to separate the case like your example from link copy operation. With delegating such operation to `Link.to_gpu` and `Variable.to_gpu` methods and providing `copied` keyword parameter, we could transfer its arrays to device memory with leaving the original ones on the host memory:

    l1 = Link(w=1)
    l2 = l1.to_gpu(copied=True)

beam2d:
OK, that sounds good. I think it's better to make the default behavior same as the current one, i.e., make to_gpu clones the _vdata by default so that the data on the original device is left.

takagi:
I have implemented `to_gpu` method to work with `copied` keyword parameter. Also I have illustrated what I have in mind: https://docs.google.com/presentation/d/1JY4n4yFpmnds55rswg_zUCaeBJ_207roZVeebPlWsXA/edit?usp=sharing

beam2d:
Thank you, I think this design is good. I recommend to rename `copied` to `copy` since it is the standard naming in NumPy and maybe it is easier to guess.

takagi:
Okay!

chainer/link.py
```diff
@@ -257,18 +257,23 @@ def to_cpu(self):
         """
```
delta2323: Returned value is not `self` if `copy` is `True`.

chainer/link.py
```diff
@@ -257,18 +257,23 @@ def to_cpu(self):
...
-            d[name].to_cpu()
+            d[name] = d[name].to_cpu(copy=copy)
         for name in self._persistent:
             value = d[name]
             if isinstance(value, cuda.ndarray):
                 d[name] = value.get()
-        self._cpu = True
-        self._device_id = None
-        return self
+        link._cpu = False
```
delta2323: Should it be `True`?

chainer/variable.py
```diff
@@ -50,6 +51,134 @@ def make_message(message):
...
+            self._initializer = (
+                initializers.NaN() if initializer is None else initializer)
+            dtype = getattr(self.initializer, 'dtype', numpy.float32)
+            self._grad_initializer = initializers.NaN(dtype)
+        elif not isinstance(data, (numpy.ndarray, cuda.ndarray)):
+                msg = '''numpy.ndarray or cuda.ndarray are expected.
+Actual: {0}'''.format(type(data))
+                raise TypeError(msg)
+
+        self.data = data
```
delta2323: How about changing `data` to `_data` as we do not expect users to access this attribute directly.

chainer/link.py
takagi: Fixed the docsrings for `to_gpu` and `to_gpu`!

chainer/link.py
takagi: Fixed!

chainer/variable.py
takagi: Good. I have fixed them!

takagi:
I will check if this change affects serialization/deserialization of links.

chainer/link.py
```diff
@@ -245,30 +245,35 @@ def copy(self):
...
-    def to_cpu(self):
+    def to_cpu(self, copy=False):
         """Copies parameter variables and persistent values to CPU.
 
         This method does not handle non-registered attributes. If some of such
         attributes must be copied to CPU, the link implementation must
         override this method to do so.
 
-        Returns: self
+        Returns: self if `copy`` is False, otherwise a copied link
```
bkvogel: From a user's perspective, I wonder if this API change might cause the following confusion: Without looking at the code, I might look at this documentation and wonder when is it appropriate to use`copy=True`? That is, "When or why should I use this new feature?" I might also wonder what is the difference between first copying a link and then calling `to_cpu()` versus using `copy=True`. The same thought applies to `to_gpu()`. 

Maybe the documentation is fine as is, but if you can think of a way to simply state the motivation for being able to return a copy and when or how it should be used, that might avoid any confusion to the user.

takagi: Thanks! Compared with `Variable` implementation of Chainer v1 and #2072, this change introducing `copy=True` essentially does the same thing with `link.copy()` followed by `link.to_cpu()` and looks making no sense.

Under the change of `Variable` implementation in this PR, introducing `copy=True` makes it possible to efficiently copy an array from CPU to GPU with leaving a link accessible  to the original array on CPU without extra CPU to CPU copy. This requirement is caused by the change that how `Variable` has an array internally and it would not be quite clear for users, as you say.

In this PR, `Variable` has the following three states and it would cause this problem:
- uninitialized
- initialized, array on CPU
- initialized, array on GPU

I will think another model where `Variable` has only two states:
- uninitialized
- initialized

and Variable that has its array on CPU is another instance from one that has its array on GPU, which may be more natural for Chainer.

Let me implement the second model.

bkvogel: Thanks for the explanation! I think that sounds good.

takagi:
Thanks to @bkvogel, I have found that the change in this PR was too much and I have implemented another model based on #2072 again.

The latest commit, 6218428, would be enough to handle Variable's initialized/uninitialized state accompanying `link.copy()` method, where a copied link would be able to reference the initialized array even when initialized is the original link's Variable.

delta2323:
> I have implemented another model based on #2072 again.
Have you already sent it as a PR?

takagi:
The latest commit e5439fe is it. I will clean the commits in this branch or make another PR later because this branch has outdated commits which are all reverted in f66da46.

takagi:
I have cleaned up the commits to keep only effective one. I used force pushing and I leave the original branch here [takagi:_v2-uninit-var.bak](https://github.com/takagi/chainer/tree/_v2-uninit-var.bak) . Then I will remove the WIP mark.

beam2d:
Would you add some tests for this change? The tests should check the behavior changing from v1.

takagi:
Sure!

takagi:
I've added the test that check the change of this PR.

beam2d:
Sorry for being late, it looks good to me. Thank you!

PR #2104, opened 2017-01-15
Add order option to cupy.zeros

yuyu2172:
This PR adds order option to `cupy.zeros`.
Also, I changed `cupy.zeros` to stop calling `cupy.empty`. This is more efficient and simpler.


okuta:
LGTM!

PR #2105, opened 2017-01-15
Add decorators to parameterize tests with order arg

yuyu2172:
Merge after #2103, #2104.

This PR adds a decorator that parameterizes tests with order options.
In addition to that, the new decorator is used to simplify tests written for `cupy.empty` and `cupy.zeros`.


okuta:
LGTM!

PR #2106, opened 2017-01-15
use for_dtypes_combination decorator for array indexing tests

yuyu2172:
This PR changes a decorator that parameterized combination of dtypes to `for_dtypes_combination`.
This is better because you can speedup tests by reducing number of combinations to tests. Also, you can test all combinations with `for_dtypes_combination` by setting an environment variable properly.


okuta:
LGTM!

PR #2110, opened 2017-01-16
Fix test_arithmetic for numpy 1.12.0

gwtnb:
numpy 1.12.0 raises when calculating negative power of integers.

```python
# raises
numpy.power(numpy.ones((), dtype=numpy.int32), -1)
```


okuta:
LGTM!

PR #2112, opened 2017-01-16
Fix the simulated bug behaviour of packbits

iwiwi:
`packbits` and `unpackbits` of NumPy <= 1.11 have bugs against empty arrays (see https://github.com/numpy/numpy/issues/8324). I intentionally implemented this behaviour in CuPy's counterparts. These bugs are fixed at NumPy 1.12 (https://github.com/numpy/numpy/pull/8327), and thus this PR also fixes the 'bugs' in CuPy.


gwtnb:
LGTM

Issue #2120, opened 2017-01-18
Error exmples for DCGAN


owruby:
I' tried to run `python train_dcgan.py --gpu 0` and recieved this error.

``` python
Traceback (most recent call last):
  File "train_dcgan.py", line 115, in <module>
    main()
  File "train_dcgan.py", line 84, in main
    device=args.gpu)
  File "/home/s1311320/projects/practice/chainer-dev/examples/dcgan/updater.py", line 14, in __init__
    super(DCGANUpdater, self).__init__(*args, **kwargs)
  File "/home/s1311320/.pyenv/versions/chainer-dev/lib/python3.5/site-packages/chainer/training/updater.py", line 131, in __init__
    optimizer['main'].target.to_gpu(device)
KeyError: 'main'
```
I fixed `update.py`

from
``` python
if device is not None and device >= 0:
    optimizer['main'].target.to_gpu(device)
```

to
``` python
for opt in self._optimizers.values():
    opt.target.to_gpu(device)
```

okuta:
This issue was fixed at #2121

PR #2121, opened 2017-01-18
Fix update.py for multi optimizers

owruby:
#2120 
I fixed `updater.py` for multi optimizers.

### DIFF
--- a/chainer/training/updater.py
+++ b/chainer/training/updater.py
@@ -128,7 +128,8 @@ class StandardUpdater(Updater):
         self._optimizers = optimizer
 
         if device is not None and device >= 0:
-            optimizer['main'].target.to_gpu(device)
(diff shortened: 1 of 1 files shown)


chainer/training/updater.py
```diff
@@ -128,7 +128,8 @@ def __init__(self, iterator, optimizer, converter=convert.concat_examples,
         self._optimizers = optimizer
 
         if device is not None and device >= 0:
-            optimizer['main'].target.to_gpu(device)
+            for opt in self._optimizers.values():
```
rezoo: How about using `six.itervalues(self._optimizers)` instead of `self._optimizers.values()`?

rezoo:
Good. After passing all tests I'd like to merge

PR #1928, opened 2016-11-27
Simplify _adv_getitem function

okuta:

PR #2126, opened 2017-01-18
Support multiple integer arrays for cupy.ndarray.__setitem__

yuyu2172:
This PR supports multiple integer arrays for `cupy.ndarray.__setitem__`. I rebased #1928.

I had to change interface of an internal function `_scatter_op_single` to support broader inputs.

After the change,  `_scatter_op_single` supports the input below. Also I left a comment in the code.
```
a.shape= (shape_0, ..., shape_{li-1}, shape_li, ..., shpe_ri, ..., shape_{a.ndim-1})
indices:  an integer array indexing axes from axis_li to axis_ri
li: id of the first axis that get indexed
ri: id of the last axis that get indexed
```
For example, when working on `a[[1,0], [0, 1]], a.shape=(2, 3, 4)`, you have `li=0, ri=1, indices=[3, 1]`.

Originally, `_scatter_op_single` only supported the special case of `li=ri`. In that case, I had to call `a._reshape` before passing to `_scatter_op_single` when working on multiple array indexing. `cupy.ndarray._reshape` may invoke copy, which is a problem because I need to store values to a view and not to a copy of the original array.

The new `_scatter_op_single` makes it possible to avoid calling `cupy.ndarray._reshape` thus never risking to call copy.

### DIFF
--- a/cupy/core/core.pyx
+++ b/cupy/core/core.pyx
@@ -1213,15 +1213,12 @@ cdef class ndarray:
     def __setitem__(self, slices, value):
         """x.__setitem__(slices, y) <==> x[slices] = y
 
-        Supported ``slices`` consists of
+        Supports both basic and advanced indexing.
 
-        * basic indexing
+        .. note::
 
-        * one integer array
-
-        * combination of basic indexing and integer arrays
-
-        * a boolean array
+            Currently, it does not support ``slices`` that consists of more
+            than one boolean arrays
 
         .. note::
 
--- a/cupy/core/core.pyx
+++ b/cupy/core/core.pyx
@@ -1161,7 +1161,7 @@ cdef class ndarray:
             if sum(arr_slices_mask) == 1:
                 axis = arr_slices_mask.index(True)
                 return a.take(adv_slices[axis], axis)
-            return _adv_getitem(a, adv_slices)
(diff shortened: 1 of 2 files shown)


rezoo:
Good, but it seems that `TestArrayAdvancedIndexingVectorValue` does not cover the case where `'indexes'` contains multiple integer arrays. Is it OK?

yuyu2172:
Thank you for comments.
I added tests to `TestArrayAdvancedIndexingVectorValue`. It was my mistake.
I also added tests on broadcasting of indices to some other test classes.

rezoo:
Thank you! After merging #1928, I would like to continue the review.

rezoo:
I merged #1928 Could you resolve the conflicts?

yuyu2172:
Thanks. I resolved the conflicts.

rezoo:
Thanks. I would like to review.

rezoo:
Thanks to you, now `cupy.ndarray` supported a various kinds of advanced indexing.
The next thing we need is PR; currently there does not exist documents of `__getitem__` and `__setitem__` since sphinx ignores all of the special methods.
It seems better to add them into the document, but what do you think?

rezoo:
Good. LGTM.

PR #2131, opened 2017-01-18
Remove fs attribute from CaffeFunction

### LINKED ISSUE #2130: `fs` attribute of CaffeFunction is missing
`CaffeFunction` should have `fs` attribute which is a `FunctionSet` that has a set of layers, according to its [document](https://github.com/pfnet/chainer/blob/master/chainer/links/caffe/caffe_function.py#L121). But it is missing. I think we should treat it as a document bug because (1) `forwards` attribute essentially same information and (2) `FunctionSet` is not used any longer.

delta2323:
Fix #2130


beam2d:
LGTM

PR #2137, opened 2017-01-19
Implement UpdateRule and use it in optimizers

### LINKED ISSUE #2008: Variable-wise update rule
This is a proposal for v2 to modify Optimizer so that we can choose different update equations and hyperparameters for different parameter variables. In this proposal, UpdateRule class is introduced, which represents the update equations used for one parameter variable. An instance of it is set to a target parameter, and Optimizer uses it to update the parameter. Each variable in the target link has its own update rule, and so we can modify its hyperparameters and add a hook function dedicated to the update rule.

Here is the design document for this proposal: https://docs.google.com/document/d/1t1UrpsKzoDZGK7fHDbN0by93253V37Kkg-7qFpOTytc/edit#.

beam2d:
Fix #2008. See the design docs for the original design. It enables us to configure the hyperparameters of optimization methods differently for each parameter (e.g. using different learning rates, different weight decay coefficients). This PR contains slight modifications to that design, including the following things.

- Hyperparameter object is introduced to let the optimizer provide the default values for hyperparameters of each update rule. It enables us to configure the hyperparameter at once for all parameters for which a hyperparameter is not overridden. **Note:** I actually don't satisfy with this design, since it introduces some complexity to the implementation of any optimizer and would require some care on handling hyperparameters. Any discussions are welcome.
- Most optimizer hook functions are now implemented as update rule hook functions. The `call_for_each_param` flag is introduced to these hook functions, with which the optimizer treats the hook functions as optimizer hook functions and invoke them for all parameters at their update.

### DIFF
--- a/chainer/link.py
+++ b/chainer/link.py
@@ -423,6 +423,39 @@ class Link(object):
         for name in self._params:
             dst[name].addgrad(src[name])
 
+    def enable_update(self):
+        """Enables update rules of all parameters under the link hierarchy.
+
+        This method sets the :attr:`~chainer.UpdateRule.enabled` flag of the
+        update rule of each parameter variable to ``True``.
+
+        """
+        for param in self.params():
+            rule = param.update_rule
+            if rule is not None:
+                rule.enabled = True
+
+    def disable_update(self):
+        """Disables update rules of all parameters under the link hierarchy.
+
+        This method sets the :attr:~chainer.UpdateRule.enabled` flag of the
+        update rule of each parameter variable to ``False``.
+
+        """
+        for param in self.params():
+            rule = param.update_rule
+            if rule is not None:
+                rule.enabled = False
+
+    @property
+    def update_enabled(self):
+        """``True`` if at least one parameter has an update rule enabled."""
+        for param in self.params():
+            rule = param.update_rule
+            if rule is not None and rule.enabled:
+                return True
+        return False
+
     def serialize(self, serializer):
         """Serializes the link object.
 
--- a/chainer/optimizer.py
+++ b/chainer/optimizer.py
@@ -51,6 +51,7 @@ class Hyperparameter(object):
         parent (Hyperparameter): Parent hyperparameter.
 
     """
+
     def __init__(self, parent=None):
         self._parent = parent
 
@@ -104,7 +105,7 @@ class UpdateRule(object):
     the first update. The values of the state dictionary are automatically
     copied to the appropriate device before the update based on the data and
     grad arrays.
-    
+
     Args:
         parent_hyperparam (Hyperparameter): Hyperparameter that provides the
             default values.
@@ -117,6 +118,7 @@ class UpdateRule(object):
         t (int): Number of updates made by this update rule.
 
     """
+
     def __init__(self, parent_hyperparam=None):
         self._hooks = collections.OrderedDict()
         self._state = None
@@ -305,6 +307,7 @@ class Optimizer(object):
             method.
 
     """
+
     def setup(self, link):
         """Sets a target link and initializes the optimizer states.
 
@@ -539,6 +542,7 @@ class GradientMethod(Optimizer):
        :class:`GradientMethod` object for efficiency.
 
     """
+
     def __init__(self):
         super(GradientMethod, self).__init__()
         self.hyperparam = Hyperparameter()
--- a/chainer/optimizer.py
+++ b/chainer/optimizer.py
@@ -1,10 +1,13 @@
 import collections
+import copy
 
 import numpy
 import six
 
+import chainer
 from chainer import cuda
 import chainer.link as link_module
+import chainer.serializer as serializer_module
 
 
 def _sum_sqnorm(arr):
@@ -27,6 +30,249 @@ def exponential_decay_noise(xp, shape, dtype, hook, opt):
     return xp.random.normal(0, std, shape).astype(dtype)
 
 
+class Hyperparameter(object):
+
+    """Set of hyperparameter entries of an optimizer.
+
+    This is a utility class to provide a set of hyperparameter entries for
+    update rules and an optimizer. Each entry can be set as an attribute of a
+    hyperparameter object.
+
+    A hyperparameter object can hold a reference to its parent hyperparameter
+    object. When an attribute does not exist in the child hyperparameter, it
+    automatically refers to the parent. We typically set the hyperparameter of
+    the gradient method as the parent of the hyperparameter of each update
+    rule. It enables us to centralize the management of hyperparameters (e.g.
+    we can change the learning rate of all update rules just by modifying the
+    hyperparameter of the central optimizer object), while users can freely
+    customize the hyperparameter of each update rule if needed.
+
+    Args:
+        parent (Hyperparameter): Parent hyperparameter.
+
+    """
+    def __init__(self, parent=None):
+        self._parent = parent
+
+    def __getattr__(self, name):
+        return getattr(self._parent, name)
+
+    def __repr__(self):
+        d = self.get_dict()
+        keys = sorted(d.keys())
+        values_repr = ', '.join('%s=%s' % (k, d[k]) for k in keys)
+        return 'Hyperparameter(%s)' % values_repr
+
+    def get_dict(self):
+        """Converts the hyperparameter into a dictionary.
+
+        Returns:
+            Dictionary containing all entries that can be referred by this
+            hyperparameter object.
+
+        """
+        d = {} if self._parent is None else self._parent.get_dict()
+        for k, v in six.iteritems(self.__dict__):
+            if k != '_parent':
+                d[k] = v
+        return d
+
+
+class UpdateRule(object):
+
+    """Base class of all update rules.
+
+    Update rule is an object that implements how to update one parameter
+    variable using the gradient of a loss function. This class provides the
+    interface and the common features of any update rules.
+
+    An update rule can be set to a :class:`~chainer.Variable` object that
+    represents a parameter array of a model. An :class:`~chainer.Optimizer`
+    instance defines which parameters to update, and the update rule instance
+    of each parameter defines how to update it.
+
+    Hook functions can be set to any update rule instance. The hook function is
+    called just before any updates in the order of registrations.
+
+    An implementation of update rule should override :meth:`update_core` or
+    its device-dependent variants (i.e., :meth:`update_core_cpu` and
+    :meth:`update_core_gpu`).
+
+    The state (e.g. a moving average of the gradient) of the update rule is
+    stored into the state dictionary. An implementation of update rule using
+    state should also override :meth:`init_state` to initialize the state at
+    the first update. The values of the state dictionary are automatically
+    copied to the appropriate device before the update based on the data and
+    grad arrays.
+    
+    Args:
+        parent_hyperparam (Hyperparameter): Hyperparameter that provides the
+            default values.
+
+    Attributes:
+        enabled (bool): Flag to configure if this update rule is active. If the
+            update rule is not active (i.e., ``enabled = False``), the
+            :meth:`update` method does not update the parameter.
+        hyperparam (Hyperparameter): Hyperparameter of the update rule.
+        t (int): Number of updates made by this update rule.
+
+    """
+    def __init__(self, parent_hyperparam=None):
+        self._hooks = collections.OrderedDict()
+        self._state = None
+        self.enabled = True
+        self.hyperparam = Hyperparameter(parent_hyperparam)
+        self.t = 0
+
+    @property
+    def state(self):
+        """State dictionary."""
+        return self._state
+
+    def add_hook(self, hook, name=None):
+        """Adds a hook function.
+
+        The hook function is called before any updates.
+
+        Args:
+            hook (callable): Hook function to be added. It takes two
+                arguments: the update rule object and the parameter variable.
+            name (str): Name of the hook function. The name attribute of the
+                hook function is used by default.
+
+        """
+        if not callable(hook):
+            raise TypeError('hook function must be callable')
+
+        if name is None:
+            name = getattr(hook, 'name', getattr(hook, '__name__', None))
+            if name is None:
+                raise ValueError(
+                    'the name of the hook function is not specified')
+        if name in self._hooks:
+            raise ValueError('hook "{}" already exists'.format(name))
+
+        self._hooks[name] = hook
+
+    def remove_hook(self, name):
+        """Removes the specified hook function.
+
+        Args:
+            name (str): Name of the hook function to be removed. The hook
+                function registered with this name will be removed.
+
+        """
+        del self._hooks[name]
+
+    def update(self, param):
+        """Invokes hook functions and updates the parameter.
+
+        Args:
+            param (~chainer.Variable): Variable to be updated.
+
+        """
+        if not self.enabled:
+            return
+
+        self.t += 1
+        self._prepare(param)
+        for hook in six.itervalues(self._hooks):
+            hook(self, param)
+        self.update_core(param)
+
+    def update_core(self, param):
+        """Updates the parameter.
+
+        Implementation of UpdateRule should override this method or both of
+        :meth:`_update_core_cpu` and :meth:`_update_core_gpu`.
+
+        Args:
+            param (~chainer.Variable): Variable to be updated.
+
+        """
+        with cuda.get_device(param.data) as dev:
+            if int(dev) == -1:
+                self.update_core_cpu(param)
+            else:
+                self.update_core_gpu(param)
+
+    def update_core_cpu(self, param):
+        """Updates the parameter on CPU.
+
+        See :meth:`update_core` for details.
+
+        Args:
+            param (~chainer.Variable): Variable to be updated.
+
+        """
+        raise NotImplementedError
+
+    def update_core_gpu(self, param):
+        """Updates the parameter on GPU.
+
+        See :meth:`update_core` for details.
+
+        Args:
+            param (~chainer.Variable): Variable to be updated.
+
+        """
+        raise NotImplementedError
+
+    def init_state(self, param):
+        """Initializes the state.
+
+        Any implementations that use the state should override this mehtod.
+        This method is called at the first update.
+
+        Args:
+            param (~chainer.Variable): Parameter variable. It can be used to
+                extract the shape and the data type of the parameter.
+
+        """
+        pass
+
+    def serialize(self, serializer):
+        """Serializes the update rule state.
+
+        Be careful that this method only saves/loads the state of the update
+        rule.
+
+        Args:
+            serializer (~chainer.AbstractSerializer): Serializer object.
+
+        """
+        if self.state is None:
+            if isinstance(serializer, serializer_module.Deserializer):
+                # try to initialize the state to retrieve state entries
+                self._state = {}
+                self_copy = copy.copy(self)
+                arr = numpy.empty(1, dtype=numpy.float32)
+                self_copy.init_state(chainer.Variable(arr, grad=arr))
+
+                for key in self._state:
+                    self._state[key] = serializer(key, None)
+        else:
+            for key in self._state:
+                self._state[key] = serializer(key, self._state[key])
+
+    def _prepare(self, param):
+        with cuda.get_device(param.data) as device:
+            state = self.state
+            if state is None:
+                state = self._state = {}
+                self.init_state(param)
+
+            for name, value in six.iteritems(state):
+                if not isinstance(value, (numpy.ndarray, cuda.ndarray)):
+                    continue
+                value_device = cuda.get_device(value)
+                if value_device.id != device.id:
+                    if device.id >= 0:
+                        state[name] = cuda.to_gpu(value)
+                    else:
+                        state[name] = cuda.to_cpu(value)
+
+
 class Optimizer(object):
     """Base class of all numerical optimizers.
 
@@ -36,20 +282,20 @@ class Optimizer(object):
     parameters based on a given loss function.
 
     Each optimizer implementation must be defined as a child class of
-    Optimizer. It must override :meth:`update` method. An optimizer can use
-    *internal states* each of which is tied to one of the parameters. State is
-    a dictionary of serializable values (typically arrays of size same as
-    the corresponding parameters). In order to use state dictionaries, the
-    optimizer must override :meth:`init_state` method (or its CPU/GPU versions,
-    :meth:`init_state_cpu` and :meth:`init_state_gpu`).
+    Optimizer. It must override :meth:`update` method.
 
     If the optimizer is based on single gradient computation (like
     most first-order methods), then it should inherit :class:`GradientMethod`,
-    which adds some features dedicated for the first order methods.
+    which adds some features dedicated for the first order methods, including
+    the support of :class:`~chainer.UpdateRule`.
 
     Optimizer instance also supports *hook functions*. Hook function is
     registered by the :meth:`add_hook` method. Each hook function is called
-    in registration order in advance of the actual parameter update.
+    in registration order in advance of the actual parameter update. If the
+    hook function has an attribute ``call_for_each_param`` of a truth value,
+    the hook function is used as a hook function of all update rules (i.e., it
+    is invoked for every parameter by passing the corresponding update rule and
+    the parameter).
 
     Attributes:
         target: Target link object. It is set by the :meth:`setup` method.
@@ -59,7 +305,6 @@ class Optimizer(object):
             method.
 
     """
-
     def setup(self, link):
         """Sets a target link and initializes the optimizer states.
 
@@ -76,104 +321,20 @@ class Optimizer(object):
         self.target = link
         self.t = 0
         self.epoch = 0
-        self._states = {}
         self._hooks = collections.OrderedDict()
 
-        self.prepare()
-
-    def prepare(self):
-        """Prepares for an update.
-
-        This method initializes missing optimizer states (e.g. for newly added
-        parameters after the set up), and copies arrays in each state
-        dictionary to CPU or GPU according to the corresponding parameter
-        array.
-
-        """
-        states = self._states
-        for name, param in self.target.namedparams(False):
-            if name not in states:
-                state = {}
-                self.init_state(param, state)
-                states[name] = state
-            else:
-                state = states[name]
-                with cuda.get_device(param.data) as dev:
-                    if int(dev) == -1:  # cpu
-                        for key, value in six.iteritems(state):
-                            if isinstance(value, cuda.ndarray):
-                                state[key] = value.get()
-                    else:  # gpu
-                        cupy = cuda.cupy
-                        for key, value in six.iteritems(state):
-                            if isinstance(value, numpy.ndarray):
-                                state[key] = cuda.to_gpu(value)
-                            elif (isinstance(value, cupy.ndarray) and
-                                  value.device != dev):
-                                state[key] = cupy.copy(value)
-
-    def init_state(self, param, state):
-        """Initializes the optimizer state corresponding to the parameter.
-
-        This method should add needed items to the ``state`` dictionary. Each
-        optimizer implementation that uses its own states should override this
-        method or CPU/GPU dedicated versions (:meth:`init_state_cpu` and
-        :meth:`init_state_gpu`).
-
-        Args:
-            param (~chainer.Variable): Parameter variable.
-            state (dict): State dictionary.
-
-        .. seealso:: :meth:`init_state_cpu`, :meth:`init_state_gpu`
-
-        """
-        with cuda.get_device(param.data) as dev:
-            if int(dev) == -1:
-                self.init_state_cpu(param, state)
-            else:
-                self.init_state_gpu(param, state)
-
-    def init_state_cpu(self, param, state):
-        """Initializes the optimizer state on CPU.
-
-        This method is called from :meth:`init_state` by default.
-
-        Args:
-            param (~chainer.Variable): Parameter variable. Its data array is
-                of type :class:`numpy.ndarray`.
-            state (dict): State dictionary.
-
-        .. seealso:: :meth:`init_state`
-
-        """
-        pass
-
-    def init_state_gpu(self, param, state):
-        """Initializes the optimizer state on GPU.
-
-        This method is called from :meth:`init_state` by default.
-
-        Args:
-            param (~chainer.Variable): Parameter variable. Its data array is
-                of type :class:`cupy.ndarray`.
-            state (dict): State dictionary.
-
-        .. seealso:: :meth:`init_state`
-
-        """
-        pass
-
     def update(self, lossfun=None, *args, **kwds):
-        """Updates the parameters and optimizer states.
+        """Updates the parameters.
 
-        This method updates the parameters of the target link and corresponding
-        optimizer states. The behavior of this method is different for the
-        cases either ``lossfun`` is given or not.
+        This method updates the parameters of the target link. The behavior of
+        this method is different for the cases either ``lossfun`` is given or
+        not.
 
-        If ``lossfun`` is given, then this method initializes the gradients by
-        zeros, calls it with given extra arguments, and calls the
+        If ``lossfun`` is given, this method typically clears the gradients,
+        calls the loss function with given extra arguments, and calls the
         :meth:`~chainer.Variable.backward` method of its output to compute the
-        gradients. The implementation might call ``lossfun`` more than once.
+        gradients. The actual implementation might call ``lossfun`` more than
+        once.
 
         If ``lossfun`` is not given, then this method assumes that the
         gradients of all parameters are already computed. An implementation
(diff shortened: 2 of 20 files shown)


chainer/initializers/constant.py
```diff
@@ -78,3 +80,16 @@ def One(dtype=None):
 
     """
     return Constant(1.0, dtype=dtype)
+
+
+def NaN(dtype=None):
```
delta2323: Is it possible extract `NaN` initializer as another PR from this one?

chainer/link.py
```diff
@@ -142,31 +141,37 @@ def xp(self):
...
         If ``initializer`` is not ``None``, the data is initialized by
-        ``initializer``.
+        ``initializer``. The initializer can be an array, in which case the
+        parameter array is directly initialized by it.
 
-        If the supplied ``name`` argument corresponds to an uninitialized
-        parameter (that is, one that was added with the
-        :meth:`add_uninitialized_param` method), ``name`` will be removed
-        from the set of uninitialized parameters.
+        If ``shape`` is None and the initializer is not an array, this method
```
delta2323: ``None``

delta2323:
It seems to me that this PR does not only introduce `UpdateRule` but also refactor how to handle uninitialized variables in links. Are these two tightly coupled? Otherwise, I would like to separate this into smaller PRs.

chainer/variable.py
```diff
@@ -75,24 +76,44 @@ class Variable(object):
             'auto') or boolean values can be used, too.
         name (str): Name of the variable.
         grad (array): Initial gradient array.
+        initializer (~chainer.Initializer): Initializer of the data array.
+            If `data` is None, this object is used for initializing the data
```
delta2323: ``None``

delta2323:
Also it is better to separate changes applied to concrete Optimizer's and UpdateRule's. One suggestion is to move changes all but one concrete Optimizer and UpdateRule (e.g. SGD) and create other PR(s).

chainer/variable.py
```diff
@@ -75,24 +76,44 @@ class Variable(object):
...
+            array in the :meth:`initialize` method.
 
     Attributes:
         data: Data array of type either :class:`numpy.ndarray` or
-            :class:`cupy.ndarray`.
+            :class:`cupy.ndarray`. If it is None, the variable is left in an
```
delta2323: ``None``

chainer/variable.py
```diff
@@ -75,24 +76,44 @@ class Variable(object):
...
         creator: The function who creates this variable. It is ``None`` if the
             variable is not created by any function.
+        update_rule: :class:`~chainer.optimizer.UpdateRule` instance that
+            updates this variable as a parameter.
         volatile: Ternary :class:`~chainer.Flag` object. If ``'ON'``, the
             variable does not keep track of any function applications. See
             :class:`~chainer.Flag` for the detail of ternary flags.
+        initializer: Initializer of the data array. It is used for initializing
+            the data array of an uninitialized variable.
+        update_rule (~chainer.optimizer.UpdateRule): Update rule attached to
```
delta2323: Is `update_rule` appeared twice?

beam2d:
Ah, sorry, I forgot to point out that this PR depends on #2072, and thus it includes a part of the codes. We should first merge #2072 and then we can merge this PR. The review might proceed before #2072 is merged, though.

beam2d:
Should I separate the PR for the core implementation of UpdateRule and each optimizer's implementation? I think we should include at least one example that uses the state dictionary in this PR. I agree that I can separate others from this PR, if needed.

chainer/variable.py
```diff
@@ -241,6 +273,12 @@ def zerograd(self):
         warnings.warn(
             'Variable.zerograd is deprecated. Use Variable.cleargard instead.',
             DeprecationWarning)
+
+        if self.data is None:
+            dtype = getattr(self.initializer, 'dtype', None)
+            self._grad_initializer = initializers.Zero(dtype)
+            return
```
delta2323: As `zerograd` is deprecated, I think it is better to remove `zerograd` in major version up, rather than fixing it.

tests/chainer_tests/test_variable.py
```diff
@@ -540,6 +541,223 @@ def test_pickle_gpu(self):
         cp.testing.assert_array_equal(x.data, d.data)
         cp.testing.assert_array_equal(x.grad, d.grad)
 
+<<<<<<< HEAD
```
delta2323: Conflict message remains.

delta2323:
OK, I will review #2072 first.

delta2323:
> I think we should include at least one example that uses the state dictionary in this PR. 

Yes, this is what I wanted to mean.

takagi:
FYI, you can change the comparison base from `_v2` to `_v2-uninit-var` with the edit button above of the issue, which make it easier to review. After #2072 will be merged, it will be enough to change the base back to `_v2`. I often use this feature.

beam2d:
Since #2072 has been merged, I will resume this PR.

chainer/optimizer.py
```diff
@@ -27,153 +30,313 @@ def exponential_decay_noise(xp, shape, dtype, hook, opt):
...
+    variable using the gradient of a loss function. This class provides the
+    interface and the common features of any update rules.
+
+    An update rule can be set to a :class:`~chainer.Variable` object that
+    represents a parameter array of a model. An :class:`~chainer.Optimizer`
+    instance defines which parameters to update, and the update rule instance
+    of each parameter defines how to update it.
+
+    Hook functions can be set to any update rule instance. The hook function is
+    called just before any updates.
```
delta2323: What is the order of hooks called.

delta2323:
It is not directly related to this PR but should we extract the examples of hooks (e.g. `WeightDecay`) and make a new directory?

beam2d:
It might be better to do so.

chainer/optimizers/ada_delta.py
```diff
@@ -4,43 +4,83 @@
...
+    def __init__(self, rho=_default_hyperparam.rho,
+                 eps=_default_hyperparam.eps):
+        super(AdaDelta, self).__init__()
+        self.hyperparam = optimizer.Hyperparameter()
+        self.hyperparam.rho = rho
+        self.hyperparam.eps = eps
+
+    def setup_update_rule(self, param):
+        param.update_rule = AdaDeltaRule()
+        param.update_rule.hyperparam.set_parent(self.hyperparam)
```
delta2323: It seems what optimizers do in `setup_update_rule` is same (i.e. it adds an instance of `UpdateRule` to the parameter  and set an instance of `HyperParameter` which acts as default hyper parameters). The difference is the type of `UpdateRule` only. Do you expect the developer of custom optimizers would like to do other than the typical procedure? If not, I think just providing an instance of UpdateRule is simpler and not confusing.

chainer/optimizer.py
```diff
@@ -27,153 +30,313 @@ def exponential_decay_noise(xp, shape, dtype, hook, opt):
...
+    rule. It enables us to centralize the management of hyperparameters (e.g.
+    we can change the learning rate of all update rules just by modifying the
+    hyperparameter of the central optimizer object), while users can freely
+    customize the hyperparameter of each update rule if needed.
 
-    Optimizer instance also supports *hook functions*. Hook function is
-    registered by the :meth:`add_hook` method. Each hook function is called
-    in registration order in advance of the actual parameter update.
+    Args:
+        parent (Hyperparameter): Parent hyperparameter.
```
delta2323: Do you expect that the hierarchical structure of `HyperParameter`'s which has more than two stages (i.e. parent-child relationship). I do not come up with any drawback for allowing "deeper" hierarchy. But I am just curious of it.

delta2323:
> It is not directly related to this PR but should we extract the examples of hooks (e.g. WeightDecay) and make a new directory?

I sent it as PR #2397

delta2323:
I added comments. Could you take a look?

chainer/optimizer.py
beam2d: Maybe there are few examples for using a deeper hierarchy of hyperparameters. One possible example is that one might want to customize a hyperparameter only for a part of the model, and it might happen in multiple levels of model granularity (e.g. one might want to customize the hyperparameter of a part A of the model, and further customize that of a subpart of A).

chainer/optimizers/ada_delta.py
beam2d: It is up to the implementation of GradientMethod whether it uses a hyperparameter or not, and thus I implemented `setup_update_rule` in this way. All implementations of gradient methods in Chainer eventually use hyperparameters, and so their implementation is almost identical. I'm wondering if we can unify the implementation. Do you think it's enough to make GradientMethod use a hyperparameter by default, or make another abstraction layer like `GradientMethodWithHyperparmeter`?

beam2d:
Thank you, and sorry for late response. I updated the branch according to your comments, and also rebase it onto the current v2 branch.

chainer/optimizer.py
delta2323: Thank you for explanation. I thought we should mention to a deeper hierarchy if there is much need for such a deeper hierarchy. But I come to think we can leave it undocumented. If we allow a deeper hierarchy in the document, we must test this specification, which is low priority considering low demand for it. On the other hand, it is somewhat strange to forbid a deeper hierarchy even if we can expect it is possible (although not tested).

chainer/optimizers/ada_delta.py
delta2323: IMO, adding the abstraction seems a bit overkilling, considering that `GradientMethod` itself is already an abstraction of several optimizers. I am thinking if there is a better solution.

delta2323:
Thank you updating the PR. I replied to you first. I will take a look the change later.

chainer/optimizer.py
```diff
@@ -27,153 +30,313 @@ def exponential_decay_noise(xp, shape, dtype, hook, opt):
...
 
-    Each optimizer implementation must be defined as a child class of
-    Optimizer. It must override :meth:`update` method. An optimizer can use
-    *internal states* each of which is tied to one of the parameters. State is
-    a dictionary of serializable values (typically arrays of size same as
-    the corresponding parameters). In order to use state dictionaries, the
-    optimizer must override :meth:`init_state` method (or its CPU/GPU versions,
-    :meth:`init_state_cpu` and :meth:`init_state_gpu`).
+    This is a utility class to provide a set of hyperparameters for update
+    rules and an optimizer. It can be used as a usual object.
```
delta2323: What do you mean by "It can be used as a usual object"?

chainer/optimizer.py
```diff
@@ -27,153 +30,313 @@ def exponential_decay_noise(xp, shape, dtype, hook, opt):
...
-    in registration order in advance of the actual parameter update.
+    Args:
+        parent (Hyperparameter): Parent hyperparameter.
+
+    """
+    def __init__(self, parent=None):
+        self._parent = parent
+
+    def __getattr__(self, name):
+        return getattr(self._parent, name)
```
delta2323: I am not sure why we need not to retrieve data from `self` and directly access to its parents.

chainer/optimizer.py
```diff
@@ -27,153 +30,313 @@ def exponential_decay_noise(xp, shape, dtype, hook, opt):
...
 
     Attributes:
-        target: Target link object. It is set by the :meth:`setup` method.
-        t: Number of update steps. It must be incremented by the
-            :meth:`update` method.
-        epoch: Current epoch. It is incremented by the :meth:`new_epoch`
-            method.
+        enabled (bool): Flag to configure if this update rule is active. If the
+            update rule is not active (i.e., ``enabled = False``), the
+            :meth:`update` method does not update the parameter.
```
delta2323: What kind of use cases are you thinking of in your mind in which `enabled=False`?

chainer/optimizers/ada_delta.py
beam2d: Agree. OK, then I will make the boilerplate included in `GradientMethod` instead of writing it every time when adding a new method.

tests/chainer_tests/serializers_tests/test_npz.py
```diff
@@ -252,8 +255,8 @@ def _check_optimizer_group(self, file, state, prefix=''):
                 'child/linear/b/msdx',
                 'child/Wc/msg',
                 'child/Wc/msdx') + state
-        self.assertSetEqual(set(file.keys()),
-                            {prefix + x for x in keys})
+        self.assertEqual(set(file.keys()),
+                         {prefix + x for x in keys})
```
beam2d: Ah, maybe not relevant, it just stops using an obsolete assert function. Should I revert it?

beam2d: Ah no, maybe not relevant. It just stops using an obsolete assert function. Should I revert it?

chainer/optimizer.py
beam2d: I wanted to mean that users can add attributes to this object.

chainer/optimizer.py
beam2d: Because `__getattr__` is called only if the access to the attribute of `name` failed. If there is an attribute `name`, this method won't be called. See: https://docs.python.org/3.6/reference/datamodel.html#object.__getattr__

chainer/optimizer.py
beam2d: I often get requests that users want to freeze a part of the model on a specific update without editing the link hierarchy. Using `enabled` flag, we can stop updating parameters without modifying the model definition. I now think it might be better to add a convenient method to `Link` that disables update rules of all parameters under the hierarchy.

tests/chainer_tests/serializers_tests/test_npz.py
delta2323: I see. If we should change `v1`, too. It is better to make another PR whose destination is the master branch.

chainer/optimizer.py
delta2323: Thank you. Then, I think it is more kind to explain the expected usage like "You can add hyperparameters as attributes as a usual object".

chainer/optimizer.py
delta2323: I see. Thank you.

chainer/optimizer.py
delta2323: You mean to provide a method to a link that switches `enabled` flag of all `UpdateRule` of variables, don't you. It is a good idea.

beam2d: Yes, I meant that. I come to think it would be better to include it in this PR, so I will add it.

beam2d:
@delta2323 I updated the code.

chainer/optimizer.py
```diff
@@ -357,33 +534,40 @@ class GradientMethod(Optimizer):
...
+    def __init__(self):
+        super(GradientMethod, self).__init__()
+        self.hyperparam = Hyperparameter()
+
+    def setup(self, link):
+        super(GradientMethod, self).setup(link)
+        for param in link.params():
+            rule = self.create_update_rule()
+            rule.hyperparam.set_parent(self.hyperparam)
+            param.update_rule = rule
```
niboshi: I think it's an unnecessary coupling between `Optimizer` and `UpdateRule` to set parent hyperparameter here.
How about setting it via `UpdateRule`'s constructor, like this?

```
class UpdateRule:
    def __init__(self, parent_hyperparam=None):
        self.hyperparam = Hyperparameter(parent_hyperparam)
```

```
class AdamRule:
    def __init__(self, parent_hyperparam=None, ...):
        super().__init__(parent_hyperparam or _default_hyperparam)
```

```
class Adam:
    def create_update_rule(self):
        return AdamRule(self.parent_hyperparam)
```

Then `Hyperparameter.set_parent()` could be no longer needed.

chainer/optimizer.py
```diff
@@ -27,153 +30,316 @@ def exponential_decay_noise(xp, shape, dtype, hook, opt):
     return xp.random.normal(0, std, shape).astype(dtype)
 
 
-class Optimizer(object):
-    """Base class of all numerical optimizers.
+class Hyperparameter(object):
```
beam2d: Hmm, I used the word "hyperparameter" to represent a single realization of configuration for each update rule/optimizer (i.e. this is the smallest unit of managing hyperparameters in Chainer). Note that we call an array "a parameter" (instead of calling each element of the array a parameter) because array is the smallest unit of managing parameters in Chainer. I don't have a strong opinion in this regard, but it might be better to ask native speakers.

chainer/optimizer.py
```diff
@@ -481,27 +641,26 @@ class WeightDecay(object):
...
-                if int(dev) == -1:
-                    g += rate * p
-                else:
-                    self.kernel()(p, rate, g)
+        p, g = param.data, param.grad
+        with cuda.get_device(p) as dev:
+            if int(dev) == -1:
+                g += self.rate * p
+            else:
+                kernel(p, self.rate, g)
```
beam2d: I'll fix it.

chainer/variable.py
```diff
@@ -594,6 +598,16 @@ def initialize(self, shape):
...
+    def update(self):
+        """Updates the data array using the gradient and the update rule.
+
+        This method updates the variable using the update rule attached to this
+        variable. If the gradient is ``None``, it does nothing.
+
+        """
+        if self.update_rule is not None and self._grad is not None:
+            self.update_rule.update(self)
+
```
beam2d: That makes sense. I will fix this part as well as each update rule implementation.

delta2323:
Now optimizers can accept `None` to indicate to use default hyperparameters, we should test this behavior. If we test all possible combinations (`None` or concrete floating values) of default hyperparameters, it would be a combinatorial explosion. So, I think it is acceptable only part of them (e.g. all `None` case and all non-`None` case).

tests/chainer_tests/test_link.py
```diff
@@ -292,6 +295,23 @@ def test_serialize_deserialize_to_uninitialized_param(self):
...
+        self.assertFalse(self.link.x.update_rule.enabled)
+        self.assertFalse(self.link.u.update_rule.enabled)
+
+    def test_update_enabled(self):
+        self.assertTrue(self.link.update_enabled)
+        self.link.disable_update()
+        self.assertFalse(self.link.update_enabled)
+        self.link.enable_update()
+        self.assertTrue(self.link.update_enabled)
+
```
beam2d: Only a portion of `update_rule`s are enabled at the initial state and so I thought the first `assertTrue` does that test. Isn't that enough?

tests/chainer_tests/test_optimizer.py
```diff
@@ -230,85 +401,67 @@ def test_hardclipping_gpu(self):
...
+        self.optimizer.setup(self.target)
 
-    @attr.gpu
-    def test_init_state_gpu(self):
-        self.check_init_state(True)
+        self.assertEqual(create_update_rule.call_count, 2)
+        self.assertEqual(create_update_rule.call_args_list[0][0], ())
+        self.assertEqual(create_update_rule.call_args_list[0][1], {})
+        self.assertEqual(create_update_rule.call_args_list[1][0], ())
+        self.assertEqual(create_update_rule.call_args_list[1][1], {})
```
beam2d: Sounds good. I will fix it.

chainer/optimizer.py
beam2d: It looks nice. I will fix it.

chainer/optimizer.py
niboshi: Neither do I have a strong opinion. It's OK if you have an intent to choose this one. But I agree that it's better to ask native speakers.

tests/chainer_tests/test_variable.py
```diff
@@ -775,6 +775,23 @@ def test_addgrad_to_uninitialized_variable_gpu_to_another_gpu(self):
...
         cp.testing.assert_array_equal(x.grad, self.b)
 
+    def test_update_rule(self):
+        update_rule = mock.MagicMock()
+        g = self.a.copy()
+        x = chainer.Variable(self.a, grad=g)
+        x.update_rule = update_rule
+        x.update()
+        self.assertEqual(update_rule.update.call_count, 1)
+        args = update_rule.update.call_args_list[0][0]
```
delta2323: It is better to check the length of the argument list is 1.

tests/chainer_tests/test_link.py
niboshi: Ah you're right, I missed that. Thank you for pointing out.

beam2d:
@delta I think the optimizers of v1 also have default values for each hyperparameter entry and so nothing has changed on this point. Or am I missing something on this topic?

chainer/link.py
```diff
@@ -426,6 +426,39 @@ def addgrads(self, link):
...
+    def enable_update(self):
+        """Enables update rules of all parameters under the link hierarchy.
+
+        This method sets the :attr:`~chainer.UpdateRule.enabled` flag of the
+        update rule of each parameter variable to ``True``.
+
+        """
+        for param in self.params():
+            rule = param.update_rule
+            if rule is not None:
```
beam2d: This is already tested by `test_enable_update` (note that parameters `y` and `v` do not have update rules).

chainer/link.py
```diff
@@ -426,6 +426,39 @@ def addgrads(self, link):
...
+    def disable_update(self):
+        """Disables update rules of all parameters under the link hierarchy.
+
+        This method sets the :attr:~chainer.UpdateRule.enabled` flag of the
+        update rule of each parameter variable to ``False``.
+
+        """
+        for param in self.params():
+            rule = param.update_rule
+            if rule is not None:
```
beam2d: same as above

chainer/link.py
```diff
@@ -426,6 +426,39 @@ def addgrads(self, link):
...
+                rule.enabled = False
+
+    @property
+    def update_enabled(self):
+        """``True`` if at least one parameter has an update rule enabled."""
+        for param in self.params():
+            rule = param.update_rule
+            if rule is not None and rule.enabled:
```
beam2d: same as above

tests/chainer_tests/optimizers_tests/test_optimizers_by_linear_model.py
```diff
@@ -111,8 +111,10 @@ def test_model_setup_multi_gpu(self):
             optimizer = self.model.optimizer
             model.to_gpu(1)
             optimizer.setup(model)
-        for name, param in optimizer.target.namedparams(False):
-            for v in six.itervalues(optimizer._states[name]):
+        for param in optimizer.target.params(False):
+            param.zerograd()
+            param.update()
```
beam2d: Yes. Will write.

tests/chainer_tests/test_optimizer.py
```diff
@@ -11,6 +11,167 @@
...
+    def test_add_hook_with_function_name(self):
+        hook_body = mock.MagicMock()
+
+        def foo(update_rule, data, grad):
+            hook_body(update_rule, data, grad)
+
+        self.update_rule.add_hook(foo)
+        self.update_rule.remove_hook('foo')
+        self.update_rule.update(self.var)
+        self.assertEqual(hook_body.call_count, 0)
```
beam2d: If or not `add_hook` works correctly is tested in another method (`test_add_hook`) and so I thought it's ok to assume the correctness. Should I check it in this test method, too?

tests/chainer_tests/test_optimizer.py
```diff
@@ -11,6 +11,167 @@
...
+        self.update_rule.update(self.var)
+        self.assertEqual(self.update_rule.update_core.call_count, 1)
+
+    def setup_state(self):
+        def init_state(data):
+            state = self.update_rule.state
+            state['a'] = 0
+            state['b'] = np.array([1, 2, 3], dtype=np.float32)
+        self.update_rule.init_state = init_state
+
```
beam2d: I think the CPU/GPU correctness is checked by the series of `test_state_*` methods as you wrote.

tests/chainer_tests/test_optimizer.py
```diff
@@ -11,6 +11,167 @@
...
+    def test_state_copy_to_another_gpu(self):
+        self.setup_state()
+
+        def update_core(param):
+            self.assertIsInstance(self.update_rule.state['b'], cuda.ndarray)
+            self.assertEqual(self.update_rule.state['b'].device.id, 1)
+
+        self.update_rule.update_core = lambda param: None
+        self.update_rule.update(chainer.Variable(
+            cuda.to_gpu(self.data), grad=cuda.to_gpu(self.grad)))
```
beam2d: This `update` lets the update rule to copy the state to GPU 0. This method actually tests if the second `update` correctly copies the state from GPU 0 to GPU 1.

tests/chainer_tests/test_variable.py
beam2d: I will fix it.

beam2d:
@niboshi @delta2323 Updated based on your comments. Could you check them?

chainer/link.py
delta2323: Sorry, I missed it. Then it is OK.

chainer/link.py
delta2323: OK

tests/chainer_tests/test_optimizer.py
delta2323: Strictly speaking, `test_add_hook` checks `add_hook` with `name` argument different from the default argument. So, adding the test covers different cases. But I also feel I am worried details too much.

tests/chainer_tests/test_optimizer.py
delta2323: OK. Then you do not have to add tests.

tests/chainer_tests/test_optimizer.py
delta2323: I see. Then could you add its "raison d'être" to the comment?

delta2323:
My intension was to check if optimizers work with both default and non-default hyperparameters. In that sense, even current test cases are not sufficient. But I think its priority is relatively low.

beam2d:
Maybe now I got it. You mean we should check if the handling of `Hyperparameter` is correct for each optimizer implementation, right? It is better to directly check that instead of extending the current tests checking the correctness of actual optimization on linear models. I will try to write tests after rebasing the branch onto the current v2.

chainer/optimizers/rmsprop.py
```diff
@@ -4,33 +4,86 @@
...
+        lr (float): Learning rate.
+        alpha (float): Exponential decay rate of the second order moment.
+        eps (float): Small value for the numerical stability.
 
-    def init_state(self, param, state):
+    """
+    def __init__(self, parent_hyperparam=None, lr=None, alpha=None, eps=None):
+        super(RMSpropRule, self).__init__(
+            parent_hyperparam or _default_hyperparam)
+        self.hyperparam.set_parent(_default_hyperparam)
```
beam2d: Ooops, I'll remove it. Thanks!

chainer/optimizer.py
```diff
@@ -27,153 +30,320 @@ def exponential_decay_noise(xp, shape, dtype, hook, opt):
...
+        return d
+
+    def set_parent(self, parent):
+        """Sets the parent hyperparameter.
+
+        Args:
+            parent (Hyperparameter): The parent hyperparameter to be set.
+
+        """
+        self._parent = parent
```
beam2d: Oh, you're right. I'll remove it too.

delta2323:
> Maybe now I got it. You mean we should check if the handling of Hyperparameter is correct for each optimizer implementation, right?

Yes, right.

niboshi:
Thank you.
LGTM w.r.t. my reviews.

beam2d:
Rebased to the current v2 and squashed the changes. I also added tests as suggested by Oono-san.

delta2323:
I think it is better to add examples in [the original design document](https://docs.google.com/document/d/1t1UrpsKzoDZGK7fHDbN0by93253V37Kkg-7qFpOTytc/edit#heading=h.xir0e9d6pkm5) to the docstring because (1) I am fearing users could confuse that we can hook functions to both `UpdateRule` and `Optimizer` and (2) users could have difficulty in changing how `UpdateRule` or hyperparameters of some specific variables without examples.

delta2323:
I am wondering the difference between an `Optimizer`'s hook with `call_for_each_param` being `True` and an `UpdateRule`'s hook. Is the order hooks are called matter?

delta2323:
LGTM except comments. Could you rebase this branch?

tests/chainer_tests/optimizers_tests/test_optimizers.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def create(self, *args, **kwargs):
+        self.optimizer = self.impl(*args, **kwargs)
+        self.optimizer.setup(self.target)
+
+    def get_hyperparam(self, name):
+        return getattr(self.target.w.update_rule.hyperparam, name)
+
+    def test_hyperparams(self):
+        self.create()
```
beam2d: Yes, it's required to extract the default hyperparameter of the optimizer.

chainer/optimizer.py
```diff
@@ -27,153 +30,314 @@ def exponential_decay_noise(xp, shape, dtype, hook, opt):
...
         pass
 
-    def init_state_gpu(self, param, state):
-        """Initializes the optimizer state on GPU.
+    def serialize(self, serializer):
+        """Serializes the update rule state.
 
-        This method is called from :meth:`init_state` by default.
+        Be careful that this method only saves/loads the state of the update
+        rule.
```
beam2d: The parameters of the target link is not saved/loaded by this method. I'll add a note.

tests/chainer_tests/optimizers_tests/test_optimizers.py
delta2323: OK, thank you for your explanation.

chainer/optimizer.py
```diff
@@ -27,153 +30,314 @@ def exponential_decay_noise(xp, shape, dtype, hook, opt):
...
+                self._state = {}
+                self_copy = copy.copy(self)
+                arr = numpy.empty(1, dtype=numpy.float32)
+                self_copy.init_state(chainer.Variable(arr, grad=arr))
+
+                for key in self._state:
+                    self._state[key] = serializer(key, None)
+        else:
+            for key in self._state:
+                self._state[key] = serializer(key, self._state[key])
```
beam2d: We have to assign it here. We cannot assume that `serializer` here is `Serializer`. It can be a deserializer (be careful that there are two if statements nested here).

chainer/optimizer.py
```diff
@@ -27,153 +30,314 @@ def exponential_decay_noise(xp, shape, dtype, hook, opt):
...
+
+    If the optimizer is based on single gradient computation (like
+    most first-order methods), then it should inherit :class:`GradientMethod`,
+    which adds some features dedicated for the first order methods, including
+    the support of :class:`~chainer.UpdateRule`.
+
+    Optimizer instance also supports *hook functions*. Hook function is
+    registered by the :meth:`add_hook` method. Each hook function is called
+    in registration order in advance of the actual parameter update. If the
+    hook function has an attribute ``call_for_each_param`` of a truth value,
```
beam2d: Thanks. Will fix.

chainer/optimizer.py
```diff
@@ -357,33 +532,39 @@ class GradientMethod(Optimizer):
...
-    An implementation of a gradient method must override the following methods:
-
-    - :meth:`init_state` or both :meth:`init_state_cpu` and
-      :meth:`init_state_gpu`
-    - :meth:`update_one` or both :meth:`update_one_cpu` and
-      :meth:`update_one_gpu`
+    This class uses :class:`~chainer.UpdateRule` to manage the update rule of
+    each parameter. A child class of GradientMethod should override
+    :meth:`setup_update_rule` to set up the default update rule to each
+    parameter.
```
beam2d: I'll fix it.

chainer/optimizer.py
delta2323: Sorry, it was a stupid mistake of me.

beam2d:
Fixed for the comments and rebased to the current v2.

delta2323:
LGTM. I will merge this PR after I confirm all GPU tests pass.

delta2323:
Why don't you change `observe_lr` instead of adding `HyperParameterProxy`? It is simpler to me.

beam2d:
I just wanted to keep the interface of each optimizer same as much as possible to v1. Some users would have written a script that directly modifies a hyperparameter via an attribute of the optimizer, which will be broken without these proxy properties.

delta2323:
I agree that we should keep backward compatibility as much as possible even though v2 is a good timing for rethinking interfaces.
My concern is that developers who make new optimizers must set `HyperParameterProxy` by themselves to make it work with the learning-rate observer.
Further, if the only reason for `HyperParameterProxy` is to keep backward compatibility, it is not beneficial for users who start to use Chainer from v2, whose number we expect will increase. I agree with your design If there would be benefit for them.

beam2d:
My concern is that `hyperparam` is not a part of the interface of `Optimizer`. Users can create an optimizer which is not a gradient method without using hyperparameters (possibly because the method cannot be broken down into update rules for each parameter). We can still let them use `Hyperparameter` to represent the learning rate, but it would be more natural to just set it to an attribute.

beam2d:
Is it better to make it mandatory for any optimizers to use `Hyperparameter` with the name `hyperparam`? If so, we can fix `observe_lr` to use it.

delta2323:
> My concern is that `hyperparam` is not a part of the interface of Optimizer.

I see. It is reasonable for me.

PR #2138, opened 2017-01-19
[v2] Merge v1.20

beam2d:


beam2d:
mitmul confirmed that it passed the tests, so I will merge it.

PR #2147, opened 2017-01-20
Add FindConvolution* wrappers to cupy

yuyu2172:
This PR adds cudnn wrappers around FindConvolution* functions.

cuDNN v3 and above support `cudnnFindConvolution*Algorithm`, and cuDNN v5 and above support `cudnnFindConvolution*AlogrithmEx`.

On implementation, I could not find a Cython feature to avoid a flake8 error at https://github.com/yuyu2172/chainer/blob/28eca7525e663ee9f25a6655cfecf1cca6826043/cupy/cuda/cudnn.pxd#L13.

The error is 
```
./cupy/cuda/cudnn.pxd:13:9: E125 continuation line with same indent as next logical line  
```
Basically, the error says that I should not change lines in the middle of an expression. Usually, in Python, there is a workaround, but I could not find any for this cython code.
Also, I needed to change lines because the expression exceeds the 80 characters limit.

I would like to know if it is fine to use NOQA to avoid flake8 error in this case?
I would also like to know if it is possible to avoid the error.

### DIFF
--- a/cupy/cuda/cudnn.pxd
+++ b/cupy/cuda/cudnn.pxd
@@ -10,7 +10,7 @@ cdef extern from *:
     ctypedef int ConvolutionBwdDataPreference \
         'cudnnConvolutionBwdDataPreference_t'
     ctypedef struct ConvolutionBwdDataAlgoPerf \
-        'cudnnConvolutionBwdDataAlgoPerf_t':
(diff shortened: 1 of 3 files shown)


okuta:
LGTM!

PR #2148, opened 2017-01-20
Add global variables introduced in cuDNN v5.1

yuyu2172:
This PR adds cuDNN related global variables that were added from cuDNN v5.1


okuta:
LGTM!

PR #2151, opened 2017-01-21
Support all real types cupy.matmul

fukatani:
All real types `matmul` calculation is supported with casting to `float32` or `float64` in core.pyx.
Especially, I need `float16` and `float64` support to implement `depthwise_conv2d`.

Since the parts of data types are no good calculation accuracy, I modified test.


cupy/core/core.pyx
```diff
@@ -2657,6 +2661,11 @@ cpdef ndarray matmul(ndarray a, ndarray b):
     out = ndarray(out_shape, dtype=dtype)
     out.data.memset(0, out.nbytes)
 
+    if dtype == ret_dtype:
+        ret = out
+    else:
+        ret = ndarray(out_shape, ret_dtype)
+
```
muupan: `ret` is not used until L2741. How about merging L2664-2667 to L2741-2743?

tests/cupy_tests/math_tests/test_matmul.py
```diff
@@ -39,30 +39,40 @@ class TestMatmul(unittest.TestCase):
...
     def test_operator_matmul(self, xp, dtype1, dtype2):
-        if not numpy.result_type(dtype1, dtype2) == numpy.float32:
-            return xp.array([])
-        x1 = xp.array(self.x1, dtype=dtype1)
-        x2 = xp.array(self.x2, dtype=dtype2)
+        x1 = testing.shaped_arange(self.shape_pair[0], xp, dtype1)
+        x2 = testing.shaped_arange(self.shape_pair[1], xp, dtype2)
         return operator.matmul(x1, x2)
 
+    # Since calculation accuracy is bad for (float16, uint8).
```
muupan: Do you mean the calculation accuracy is bad only between float16 and uint8? If so, how about testing the cases between float16 and non-uint8 types and between uint8 and non-float16 types? And is it still possible to include low-accuracy cases with low tolerance parameters?

fukatani:
Thank you for your comments.
I fixed them.

tests/cupy_tests/math_tests/test_matmul.py
```diff
@@ -39,30 +39,54 @@ class TestMatmul(unittest.TestCase):
...
+    @testing.for_all_dtypes(name='dtype2', no_float16=True)
+    @testing.numpy_cupy_allclose(rtol=1e-3, atol=1e-3)  # required for uint8
     def test_cupy_matmul(self, xp, dtype1, dtype2):
-        if not numpy.result_type(dtype1, dtype2) == numpy.float32:
-            return xp.array([])
-        x1 = xp.array(self.x1, dtype=dtype1)
-        x2 = xp.array(self.x2, dtype=dtype2)
+        x1 = testing.shaped_arange(self.shape_pair[0], xp, dtype1)
+        x2 = testing.shaped_arange(self.shape_pair[1], xp, dtype2)
+        return xp.matmul(x1, x2)
```
muupan: I think `test_cupy_matmul` should be also fixed in the same way as `test_operator_matmul` as for float16 and uint8.

fukatani: Sorry, I missed it.
I fixed `test_cupy_matmul`.

tests/cupy_tests/math_tests/test_matmul.py
```diff
@@ -39,30 +39,66 @@ class TestMatmul(unittest.TestCase):
...
+            return xp.array([])
+        if dtype2 == numpy.float16 and dtype1 == numpy.int8:
             return xp.array([])
-        x1 = xp.array(self.x1, dtype=dtype1)
-        x2 = xp.array(self.x2, dtype=dtype2)
+        x1 = testing.shaped_arange(self.shape_pair[0], xp, dtype1)
+        x2 = testing.shaped_arange(self.shape_pair[1], xp, dtype2)
+        return xp.matmul(x1, x2)
+
+    # Since calculation accuracy is bad for (float16, uint8).
```
muupan: This should also be `int8`?

tests/cupy_tests/math_tests/test_matmul.py
```diff
@@ -39,30 +39,66 @@ class TestMatmul(unittest.TestCase):
...
+    @testing.with_requires('numpy>=1.10')
+    @testing.numpy_cupy_allclose(rtol=1, atol=1)
+    def test_cupy_matmul2(self, xp):
+        x1 = testing.shaped_arange(self.shape_pair[0], xp, numpy.float16)
+        x2 = testing.shaped_arange(self.shape_pair[1], xp, numpy.int8)
+        return xp.matmul(x1, x2)
+
+    # Since calculation accuracy is bad for (float16, uint8).
```
muupan: Ditto

muupan:
Could you explain why you replaced `uint8` with `int8`?

fukatani:
I'm really sorry, It's simple misconception.

`uint8` has larger value range than `int8`,
So I thought that `uint8` would have greater error.

I don't know the reason, but the fact is inverse.

muupan:
Thanks for the fixes and your explanation. I ran test_matmul.py 100 times without no failure, so the current test cases seem stable enough.

fukatani:
Thanks for your validation :smiley: 
So this PR can be merged?

muupan:
Sorry for the delay. Now it can be merged.

fukatani:
Thanks for merging! :smile:

PR #2153, opened 2017-01-23
Change the behaviour of Variable.__len__

mitmul:
Change for https://github.com/pfnet/chainer/issues/1792

### DIFF
--- a/chainer/variable.py
+++ b/chainer/variable.py
@@ -167,7 +167,7 @@ Actual: {0}'''.format(type(data))
         """
         if self.data.ndim == 0:
             raise TypeError('len() of unsized object')
-        return self.data.shape[0]
+        return len(self.data)
 
     @property
     def volatile(self):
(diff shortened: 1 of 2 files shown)


beam2d:
Would you fix the docstring as well?

mitmul:
Yes, thanks for pointing out. I followed the numpy's condition for exception https://github.com/numpy/numpy/blob/master/numpy/core/src/multiarray/mapping.c#L51

chainer/variable.py
```diff
@@ -159,13 +159,15 @@ def __pos__(self):
...
 
         Returns:
-            int: Number of elements of the data array.
+            int: Number of the first dimension of the data array.
 
         """
-        return self.data.size
+        if self.data.ndim == 0:
+            raise TypeError('len() of unsized object')
+        return self.data.shape[0]
```
beam2d: Is it enough to just call `len(self.data)`?

beam2d:
LGTM

PR #2156, opened 2017-01-23
Fix 'OverflowError: Python int too large to convert to C long'

kmatsuura:
Fix the difference between Python int and C long in chainer_tests.


okuta:
LGTM!

Issue #2020, opened 2016-12-16
Remove wscale options


beam2d:
Linear, Convolution2D, and many similar links have an initialization option `wscale`, which is used to scale the initial weight matrices relatively from the default initialization. This option is now just confusing, since we now have weight initializers that can effectively select arbitrary scaling of the initial parameters. This is even broken in some cases (see #1806). I think it's time to remove this option, and just use a non-default initializer to select the appropriate scaling.

delta2323:
After removing `wscale` option, I would like to clean up the API of `Linear.__init__` etc., especially, `bias`, `nobias`, and `initial_bias` option.

delta2323:
Do you think it is reasonable to remove `scale` option from `_get_initializer` as well?

beam2d:
Yes, I think so.

delta2323:
OK, I will try to do it, either in this PR or possibly as another one.

PR #2164, opened 2017-01-24
Remove scale option from init_weight and _get_initializer

delta2323:
This PR removes `scale` option from `init_weight` and `_get_initializer`, which is discussed in #2020. We should feed scaled initializer instead.

Some links uses this option. As this dependency was removed in #2020, please merge the PR before this PR.


beam2d:
LGTM

PR #2167, opened 2017-01-24
Do not use isinstance to select array module

delta2323:
Generally speaking, `isinstance` should not be used if possible because it is slow in some case. This PR avoid to use it to determine an array module used.

### DIFF
--- a/chainer/functions/pooling/unpooling_2d.py
+++ b/chainer/functions/pooling/unpooling_2d.py
@@ -1,3 +1,5 @@
+import numpy
+
 from chainer import cuda
 from chainer.functions.pooling import pooling_2d
 from chainer.utils import conv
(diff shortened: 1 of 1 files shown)


delta2323:
Just a comment. We already use `isinstance` via `get_array_module` in the same function. So we still cannot completely remove `isinstance` with this PR.

chainer/functions/pooling/unpooling_2d.py
```diff
@@ -43,7 +43,7 @@ def forward(self, x):
         xp = cuda.get_array_module(*x)
         col = xp.tile(x[0][:, :, None, None],
                       (1, 1, self.kh, self.kw, 1, 1))
-        if isinstance(x[0], cuda.ndarray):
```
gwtnb: `cuda.cupy` is unavailable on environments without a GPU. How about `xp is not numpy`?

delta2323: I fixed. I use `xp is numpy` and switch the order for readbility.

gwtnb:
LGTM

PR #2169, opened 2017-01-24
Support float array type loc and scale for cupy.random.normal

fukatani:
Like numpy.random.normal.

Please see https://docs.scipy.org/doc/numpy/reference/generated/numpy.random.normal.html .

### DIFF
--- a/tests/cupy_tests/random_tests/test_distributions.py
+++ b/tests/cupy_tests/random_tests/test_distributions.py
@@ -1,16 +1,14 @@
 import unittest
 
-import numpy
-
 import cupy
 from cupy.random import distributions
 from cupy import testing
 
 
 @testing.parameterize(*testing.product({
-    'shape': [(3, 2)],
+    'shape': [(4, 3, 2), (3, 2)],
     'loc_shape': [(), (3, 2)],
-    'scale_shape': [(), (3, 2)]
+    'scale_shape': [(), (3, 2)],
 })
 )
 @testing.gpu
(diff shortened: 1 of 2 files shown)


tests/cupy_tests/random_tests/test_distributions.py
```diff
@@ -1,9 +1,32 @@
...
+
+    def check_distribution(self, dist_func, dtype):
+        loc = cupy.ones(self.loc_shape)
+        scale = cupy.ones(self.scale_shape)
+        out = dist_func(loc, scale, self.shape, dtype)
+        self.assertEqual(self.shape, out.shape)
+        self.assertEqual(out.dtype, dtype)
+
+    def test_normal_float32(self):
+        self.check_distribution(distributions.normal, numpy.float32)
```
delta2323: How about using [`for_float_dtypes`](http://docs.chainer.org/en/stable/cupy-reference/testing.html#cupy.testing.for_float_dtypes) (or other variants) to parameterize dtype?

An example usage is [here](https://github.com/pfnet/chainer/blob/8edf6b601ce2669d2f4c812d1bc8a62e78451eb0/tests/cupy_tests/core_tests/test_ndarray_scatter.py#L75).

tests/cupy_tests/random_tests/test_distributions.py
```diff
@@ -1,9 +1,32 @@
...
+import cupy
+from cupy.random import distributions
 from cupy import testing
 
 
+@testing.parameterize(*testing.product({
+    'shape': [(3, 2)],
+    'loc_shape': [(), (3, 2)],
+    'scale_shape': [(), (3, 2)]
+})
```
delta2323: How about adding more cases in which some broadcastings occur (some of the tests you have implemented do it)?

tests/cupy_tests/random_tests/test_distributions.py
```diff
@@ -1,9 +1,32 @@
...
+)
```
delta2323: How about parameterizing the `dtype` of `loc` and `scale` as well?

fukatani:
Thanks. I added tests.

delta2323:
LGTM. I am checking if GPU tests will pass. If it is successful, I will merge this PR.

PR #2173, opened 2017-01-25
Add unit tests for `_get_initializer`

delta2323:
This PR adds unit tests for `chainer.initializers._get_initializer`. It does not test `scale` option as I expect this option will be removed in v2 (as I did in PR #2164)


gwtnb:
LGTM

PR #2183, opened 2017-01-26
add write_trigger argument for LogRecord extension

uchida:
In the current `LogRecord` extension implementation, it aggregate and write at *the same time*.
This cause performance penalty in long time training, because log getting bigger and bigger.
One can save writings with following code:

```python
log_report = extensions.LogReport(log_name=None) # to disable log write
trainer.extend(log_report)
trainer.run()
with open(os.path.join(trainer.out, "log"), "w") as f:
    json.dump(log_report.log, f, indent=4)
```

but this seems less flexible, so we use a `LogRecord` implementation in this pull request.

### DIFF
--- a/chainer/training/extensions/log_report.py
+++ b/chainer/training/extensions/log_report.py
@@ -19,14 +19,14 @@ class LogReport(extension.Extension):
     :class:`~chainer.DictSummary` at a regular interval specified by a supplied
     trigger, and writes them into a log file in JSON format.
 
-    There are two triggers to handle this extension. One is the trigger to
+    There are three triggers to handle this extension. One is the trigger to
     invoke this extension, which is used to handle the timing of accumulating
-    the results. It is set to ``1, 'iteration'`` by default. The other is the
+    the results. It is set to ``1, 'iteration'`` by default. The second is the
     trigger to determine when to emit the result. When this trigger returns
     True, this extension appends the summary of accumulated values to the list
-    of past summaries, and writes the list to the log file. Then, this
-    extension makes a new fresh summary object which is used until the next
-    time that the trigger fires.
+    of past summaries. Then, this extension makes a new fresh summary object
+    which is used until the next time that the trigger fires. Last is the
+    trigger to determine when to write the list to the log file.
 
     It also adds some entries to each result dictionary.
 
@@ -38,10 +38,13 @@ class LogReport(extension.Extension):
     Args:
         keys (iterable of strs): Keys of values to accumulate. If this is None,
             all the values are accumulated and output to the log file.
-        trigger: Trigger that decides when to aggregate the result and output
-            the values. This is distinct from the trigger of this extension
+        trigger: Trigger that decides when to aggregate the result values.
+            This is distinct from the trigger of this extension
             itself. If it is a tuple in the form ``<int>, 'epoch'`` or
             ``<int>, 'iteration'``, it is passed to :class:`IntervalTrigger`.
+        write_trigger: Trigger that decides when to write the results the list
+            to the log file. if `None` is given, default for this argument,
+            use ``triger`` arguments, invoked after each aggregations.
         postprocess: Callback to postprocess the result dictionaries. Each
             result dictionary is passed to this callback on the output. This
             callback can modify the result dictionaries, which are used to
@@ -54,10 +57,14 @@ class LogReport(extension.Extension):
 
     """
 
-    def __init__(self, keys=None, trigger=(1, 'epoch'), postprocess=None,
-                 log_name='log'):
+    def __init__(self, keys=None, trigger=(1, 'epoch'), write_trigger=None,
+                 postprocess=None, log_name='log'):
         self._keys = keys
         self._trigger = trigger_module.get_trigger(trigger)
+        if write_trigger:
+            self._write_trigger = trigger_module.get_trigger(write_trigger)
+        else:
+            self._write_trigger = self._trigger
         self._postprocess = postprocess
         self._log_name = log_name
         self._log = []
@@ -75,7 +82,7 @@ class LogReport(extension.Extension):
         else:
             summary.add({k: observation[k] for k in keys if k in observation})
 
-        if self._trigger(trainer):
+        if self._trigger(trainer) or self._write_trigger(trainer):
             # output the result
             stats = self._summary.compute_mean()
             stats_cpu = {}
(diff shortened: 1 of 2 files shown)


delta2323:
Thank you for your PR and sorry for our late response. Is it possible for you to write unit tests for this feature?

delta2323:
@uchida Hi, is it possible to work on this PR ? Thanks in advance.

uchida:
Sorry for being late, I've added testcase for LogReport extension.

chainer/training/extensions/log_report.py
```diff
@@ -90,21 +97,25 @@ def __call__(self, trainer):
...
-                fd, path = tempfile.mkstemp(prefix=log_name, dir=trainer.out)
-                with os.fdopen(fd, 'w') as f:
-                    json.dump(self._log, f, indent=4)
+            if self._trigger(trainer):
+                self._log.append(stats_cpu)
 
-                new_path = os.path.join(trainer.out, log_name)
-                shutil.move(path, new_path)
+            if self._write_trigger(trainer):
+                self._write(trainer)
```
delta2323: It seems the argument is inconsistent with function declaration of `_write`. The error is not raised in the unit test because `_write` method is substituted with a mock object.

delta2323: So, could you check `_write` method is called with appropriate arguments?

delta2323:
Is it possible to write unittests that call and check `_write` method directly?

chainer/training/extensions/log_report.py
uchida: Ops, sorry for my bad, fixed

uchida:
> Is it possible to write unittests that call and check _write method directly?

wrote unittest and fixed inconsistent arguments.

delta2323:
Thanks, I will take a look.

tests/chainer_tests/training_tests/extensions_tests/test_logreport.py
```diff
@@ -9,29 +11,32 @@
...
             trainer.run()
             mocked.assert_called()
-            self.assertEqual(mocked.call_count, 1)
-            self.assertEqual(len(log_report._log), 10)
+            self.assertEqual(mocked.call_count, stops // writes)
+            self.assertEqual(len(log_report._log), stops // aggregates)
 
 
-def _get_mocked_trainer(stop_trigger=(10, 'iteration')):
+def _get_mocked_trainer(out, stop_trigger=(10, 'iteration')):
```
delta2323: It is easier to read to align the argument order to that of `Trainer` (i.e. swap `out` and `stop_trigger`. You can do it because the default argument of `stop_trigger` is not used).

tests/chainer_tests/training_tests/extensions_tests/test_logreport.py
```diff
@@ -0,0 +1,60 @@
...
+
+class TestLogReport(unittest.TestCase):
+
+    def setUp(self):
+        self.temp_dir = tempfile.mkdtemp()
+
+    def tearDown(self):
+        shutil.rmtree(self.temp_dir)
+
+    def test_trigger(self):
```
delta2323: It is certainly true that this test passes if `_write` is called with appropriate arguments. But it also passes if arguments are wrong but `_write` is never called. What I intended is to create directly call `log_report._write` with expected argument to check arguments fed to the mock object in `test_write_trigger`. You can check it with [assert_called_with](https://docs.python.org/3.6/library/unittest.mock.html#unittest.mock.Mock.assert_called_with). I am sorry my comment was insufficient.

delta2323: Sorry, I read my comment again and found that the sentence was inconsistent. My intension was two-fold:

* Write a unit test that directly calls `log_report._write` with expected arguments (without mock object)
* Check appropriate arguments are correctly fed to the the mock object of `_write` method.

You did the second one already. Could you the first one?

uchida: Method names starts with `_` is for internal use, so I'm hesitating to write a unit test that directly calls `log_report._write`.
And making `_write()` to `write()` seems not so good, I'm expecting most users never call.

How about checking content of `{out_dir}/log` after `trainer.run()`?

delta2323: I got the point. It sounds good to check via the output directory.

delta2323:
Could you fix [the flake8 validation](https://travis-ci.org/pfnet/chainer/jobs/224237120)

delta2323:
LGTM, I am running GPU tests.

chainer/training/extensions/log_report.py
```diff
@@ -54,10 +57,14 @@ class LogReport(extension.Extension):
 
     """
 
-    def __init__(self, keys=None, trigger=(1, 'epoch'), postprocess=None,
-                 log_name='log'):
+    def __init__(self, keys=None, trigger=(1, 'epoch'), write_trigger=None,
+                 postprocess=None, log_name='log'):
```
delta2323: I noticed that adding `write_trigger` option in the middle breaks users' code. Although we can break compatibility theoretically (cf. [our compatibility policy](https://docs.chainer.org/en/stable/compatibility.html#breaking-the-compatibility)), we need not unnecessarily do that if we can avoid it, thought the order of options is a bit natural. Could you add the option to the last position?

tests/chainer_tests/training_tests/extensions_tests/test_logreport.py
```diff
@@ -0,0 +1,64 @@
...
+                write_trigger=(writes, 'iteration')
+            )
+            trainer.extend(log_report)
+            trainer.run()
+            mocked.assert_called_with(log_report.log[-1], self.temp_dir)
+            self.assertEqual(mocked.call_count, stops // writes)
+            self.assertEqual(len(log_report.log), stops // aggregates)
+
+
+def _get_mocked_trainer(stop_trigger=(10, 'iteration'), out='result'):
```
delta2323: Chainer offers [chainer.testing.get_trainer_with_mock_updater](https://github.com/chainer/chainer/blob/master/chainer/testing/training.py#L8) from v2 for testing trainers and related modules. Could you use that instead of the custom one if possible?

delta2323:
@uchida Hi, could you let me know if you could spare time to work on this PR?

PR #2163, opened 2017-01-24
Remove wscale

delta2323:
Fix #2020

PR #2185, opened 2017-01-26
Clean APIs of Linear and convolution-like links related to bias argument

### LINKED ISSUE #2180: Clean up APIs of linear and convolution-like links
Currently `Linear` and convolution-like links (e.g. `Convolution2D`, `Deconvolution2D` etc.) have `bias`, `nobias` and `initial_bias` arguments. Although these arguments have different role, they have overlaps and depend on each other to some extent (e.g. if `initial_bias` is `None`, `bias` is used for initialization). As this situation is messy for users, I would like to simplify the interface.

`ConvolutionND` and `DeconvolutionND` have simpler `__init__` method. It has `initial_bias` only which is in charge of all of their responsibilities (it is not related directly to this issue, it caused an unexpected bug as discussed in #2018). I would like to align old-style links to newer ones.

delta2323:
Fix #2180. As this PR is implemented on top of #2163, please merge it before this PR.


delta2323:
The design principle is to use single argument `initial_bias`, whose values are passed to `init_weight` or `_get_initializer` and converted to `Initializer` instance. One thing that needs discussion is how to omit bias parameters. There are two options for this

* When `initial_bias` is ``None``, bias parameters are omitted (this is what we take in `ConvolutionND` and `DeconvolutionND`). It makes simpler API, but it greatly affects user code because if `initial_bias` is ``None``, we does make bias parameters initialized by default initializer (namely `HeNormal`) in v1.
* Use an argument other than `initial_bias` (like `nobias`) to turn on/off bias parameters. There will be more compatibility with v1, but APIs of links will be more complicated.

delta2323:
I have decided take the former option (i.e. use `nobias` option to enable/disable bias parameter).

bkvogel:
I think it looks good and agree with your design decision. Could you fix travis errors (flake8, I think)?

delta2323:
@bkvogel Thank you for your review. I fixed the code and is waiting for CI.

bkvogel:
It looks like CI failed. Could you check again? Thanks.

bkvogel:
LGTM!

Issue #2182, opened 2017-01-26
Improve docs of functions and links


shiba24:
(Last update: 23 August, 2017)
# About
This is a part of documentation improvement. This issue is about API document, in particular, functions and links. __Contributions are welcomed! If you have any questions, please feel free to comment in this issue.__ See also: #1867 
# Current document
Functions: http://docs.chainer.org/en/stable/reference/functions.html
Links: http://docs.chainer.org/en/stable/reference/links.html
# How to contribute
Before modifying files, please see http://docs.chainer.org/en/stable/contribution.html#coding-guidelines .
For really lazy guys, you need to run below commands to confirm your codes following the coding guideline!
```
autopep8 --global-config .pep8 path/to/your/code.py
flake8 path/to/your/code.py
```

## Preparation
Set the developer mode for chainer.
 1. Type `pip uninstall chainer` multiple times, until you see the error of `Cannot uninstall requirement chainer, not installed`.
 2. Go to your (forked-and-cloned) chainer working directory.
 3. Type `python setup.py develop`. Now you can use `import chainer` in the python.

## Modification
Please see each function/link template and how-to described in the latter part of this issue.

### Notation rule
We are now creating notation rules in the comment of this issue! Please see: https://github.com/pfnet/chainer/issues/2182#issuecomment-275901211

## Test on your local machine
 1. Run doctest
After modifying the files, you need to run doctest before sending PRs, using `sphinx` (please `pip install sphinx` for the first time). To test only one file, you need to add these lines at the end of the file you modified, and type `python FILENAME`.
```
if __name__ == "__main__":
    import doctest
    import numpy as np
    # import cupy  # comment out if you don' have GPU
    import chainer
    from chainer import cuda, Function, gradient_check, training, utils, Variable
    from chainer import datasets, iterators, optimizers, serializers
    from chainer import Link, Chain, ChainList
    import chainer.functions as F
    import chainer.links as L
    from chainer.training import extensions
    np.random.seed(0)
    doctest.testmod()
```
 2. Confirm compiled files
Make sure to delete these parts after the doctest! If the codes passes it, then you can compile the documentation and see the result.
```
cd docs
make html
```
Now you can access the compiled documentations in your browser (e.g. google chrome). URL is:
```
file://PATH-TO-YOUR-CHAINER-DIRECTORY/docs/build/html/index.html
```
Please check your modification seems good!

 3. (Optional) Run doctest for all the documentations
[...]

PR #2189, opened 2017-01-27
Improve docs of functions/connection/linear

shiba24:

Modified functions/links:
 - linear (functions)

One possible issue to discuss: should the example include all the optional variables (in this case, `b`)?

### DIFF
--- a/chainer/functions/connection/linear.py
+++ b/chainer/functions/connection/linear.py
@@ -71,21 +71,21 @@ def linear(x, W, b=None):
 
     Args:
         x (:class:`~chainer.Variable` or :class:`numpy.ndarray` or \
-        :class:`cupy.ndarray`): Input variable, which is :math:`(b, s_1, \
-            s_2, ..., s_n)`-shaped float array. Its first dimension `b`
+        :class:`cupy.ndarray`): Input variable, which is a :math:`(s_B, s_1, \
+            s_2, ..., s_n)`-shaped float array. Its first dimension :math:`(s_B)`
             is assumed to be the *minibatch dimension*. The other dimensions
             are treated as concatenated one dimension whose size must be
             :math:`(s_1 * ... * s_n = n)`.
         W (:class:`~chainer.Variable` or :class:`numpy.ndarray` or \
         :class:`cupy.ndarray`): Weight variable of shape :math:`(m, n)`,
-            where :math:`(n == s_1 * ... * s_n)`.
+            where :math:`(n = s_1 * ... * s_n)`.
         b (:class:`~chainer.Variable` or :class:`numpy.ndarray` or \
         :class:`cupy.ndarray`): Bias variable (optional) of shape
             :math:`(m,)`.
 
     Returns:
         ~chainer.Variable: Output variable. A float array with shape
-        of :math:`(b, m)`.
+        of :math:`(s_B, m)`.
 
     .. seealso:: :class:`~chainer.links.Linear`
 
--- a/chainer/functions/connection/linear.py
+++ b/chainer/functions/connection/linear.py
@@ -70,16 +70,22 @@ def linear(x, W, b=None):
         (3, 5)
 
     Args:
-        x (~chainer.Variable): (b, s1, s2, .., sn) shaped float array. Its
-            first dimension `b` is assumed to be the *minibatch dimension*.
-            The other dimensions are treated as concatenated one dimension
-            whose size must be ``s1 * ... * sn = n``.
-        W (~chainer.Variable): Weight variable of shape ``(m, n)``, where
-            ``n == s1 * ... * sn``
-        b (~chainer.Variable): Bias variable (optional) of shape ``(m,)``.
+        x (:class:`~chainer.Variable` or :class:`numpy.ndarray` or \
+        :class:`cupy.ndarray`): Input variable, which is :math:`(b, s_1, \
+            s_2, ..., s_n)`-shaped float array. Its first dimension `b` 
+            is assumed to be the *minibatch dimension*. The other dimensions
+            are treated as concatenated one dimension whose size must be
+            :math:`(s_1 * ... * s_n = n)`.
+        W (:class:`~chainer.Variable` or :class:`numpy.ndarray` or \
+        :class:`cupy.ndarray`): Weight variable of shape :math:`(m, n)`,
+            where :math:`(n == s_1 * ... * s_n)`.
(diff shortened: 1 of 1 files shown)


chainer/functions/connection/linear.py
```diff
@@ -58,17 +58,34 @@ def linear(x, W, b=None):
 
     It accepts two or three arguments: an input minibatch ``x``, a weight
     matrix ``W``, and optionally a bias vector ``b``. It computes
-    :math:`Y = xW^\\top + b`.
+     .. math:: Y = xW^\\top + b.
+
+    Example::
```
mitmul: Could you use `.. admonition:: Example` instead of `Example::` ?

shiba24:
Thank you for the quick review. I modified it.

chainer/functions/connection/linear.py
```diff
@@ -58,17 +58,34 @@ def linear(x, W, b=None):
...
+        (3, 5)
 
     Args:
-        x (~chainer.Variable): Input variable. Its first dimension is assumed
-            to be the *minibatch dimension*. The other dimensions are treated
-            as concatenated one dimension whose size must be ``N``.
-        W (~chainer.Variable): Weight variable of shape ``(M, N)``.
-        b (~chainer.Variable): Bias variable (optional) of shape ``(M,)``.
+        x (:class:`~chainer.Variable` or :class:`numpy.ndarray` or \
+        :class:`cupy.ndarray`): Input variable, which is :math:`(b, s_1, \
```
mitmul: Could you change the notation to indicate the first dimension of the input variable? Because `b` is already used as the name for the bias, an input argument, so it might be a little confusing for someone to see the definitions of linear link at the first time.

chainer/functions/connection/linear.py
```diff
@@ -58,17 +58,34 @@ def linear(x, W, b=None):
...
+            s_2, ..., s_n)`-shaped float array. Its first dimension `b`
```
mitmul: Ditto. (another name for the first dimension `b` is preferable)

chainer/functions/connection/linear.py
```diff
@@ -58,17 +58,34 @@ def linear(x, W, b=None):
...
+            is assumed to be the *minibatch dimension*. The other dimensions
+            are treated as concatenated one dimension whose size must be
+            :math:`(s_1 * ... * s_n = n)`.
+        W (:class:`~chainer.Variable` or :class:`numpy.ndarray` or \
+        :class:`cupy.ndarray`): Weight variable of shape :math:`(m, n)`,
+            where :math:`(n == s_1 * ... * s_n)`.
```
mitmul: I think `=` is OK instead of `==` here.

chainer/functions/connection/linear.py
```diff
@@ -58,17 +58,34 @@ def linear(x, W, b=None):
...
+        b (:class:`~chainer.Variable` or :class:`numpy.ndarray` or \
+        :class:`cupy.ndarray`): Bias variable (optional) of shape
+            :math:`(m,)`.
 
     Returns:
-        ~chainer.Variable: Output variable.
+        ~chainer.Variable: Output variable. A float array with shape
+        of :math:`(b, m)`.
```
mitmul: If you accept my suggestion to change the notation for the first dimension, please change `b` here also.

shiba24:
I see. I changed the first dimension from `b` to `s_B`.

mitmul:
Thanks for changing ti! It's basically LGTM, but I'm wondering why did you change the notations to indicate the weight variable shape from the uppercase ones `(M, N)` to `(m, n)`? I feel when `s_B` denotes the batch dimension, it's natural that the uppercase `M` and `N` are also used to denote other dimensions. Please consider this :)

shiba24:
Hm, I agree with you. Thank you for your patient review. I think that this kind of 'rule' for the notation may be good to be set!

shiba24:
So, which of upper and lower cases do you think is appropriate for the shape notation? I mean, `s_b, m, n, etc` or `s_B, M, N`? @mitmul

chainer/functions/connection/linear.py
```diff
@@ -58,17 +58,34 @@ def linear(x, W, b=None):
...
 
     Args:
-        x (~chainer.Variable): Input variable. Its first dimension is assumed
-            to be the *minibatch dimension*. The other dimensions are treated
-            as concatenated one dimension whose size must be ``N``.
-        W (~chainer.Variable): Weight variable of shape ``(M, N)``.
-        b (~chainer.Variable): Bias variable (optional) of shape ``(M,)``.
+        x (:class:`~chainer.Variable` or :class:`numpy.ndarray` or \
+        :class:`cupy.ndarray`): Input variable, which is a :math:`(s_B, s_1, \
+            s_2, ..., s_n)`-shaped float array. Its first dimension :math:`(s_B)`
```
mitmul: This line length exceed 79 characters. So the flake8 test failed. Please make a line break before :math:`(s_B)`

shiba24:
I changed too-long-line part, and made the notations of shape upper case.

shiba24:
> We need to unify the notation rules and describe it explicitly somewhere...

It might be temporally, but I started describing possible rules here.
https://github.com/pfnet/chainer/issues/2182#issuecomment-275901211

chainer/functions/connection/linear.py
```diff
@@ -58,17 +58,34 @@ def linear(x, W, b=None):
...
 
     Args:
-        x (~chainer.Variable): Input variable. Its first dimension is assumed
-            to be the *minibatch dimension*. The other dimensions are treated
-            as concatenated one dimension whose size must be ``N``.
-        W (~chainer.Variable): Weight variable of shape ``(M, N)``.
-        b (~chainer.Variable): Bias variable (optional) of shape ``(M,)``.
+        x (:class:`~chainer.Variable` or :class:`numpy.ndarray` or \
+        :class:`cupy.ndarray`): Input variable, which is a :math:`(s_B, s_1, \
+            s_2, ..., s_n)`-shaped float array. Its first dimension
```
mitmul: This line containes whitespaces after the last character, then the below flake8 error occurs:
```
./chainer/functions/connection/linear.py:75:68: W291 trailing whitespace
```

shiba24: Yes I noticed just now. After now I will test flake8 for the modification on my local machine... Sorry!

mitmul:
Thanks! LGTM

PR #2191, opened 2017-01-28
Implement inplace ReLU

gwtnb:
Implement inplace ReLU.

### DIFF
--- a/chainer/functions/activation/relu.py
+++ b/chainer/functions/activation/relu.py
@@ -2,7 +2,6 @@ import numpy
 
 from chainer import cuda
 from chainer import function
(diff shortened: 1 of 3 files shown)


gwtnb:
We should consider "inplace mode". https://github.com/pfnet/chainer/pull/2188#issuecomment-275993112

mattya:
LGTM!

beam2d:
I have to note that we should carefully make the design of in-place function because it overwrites the variable array which would break the backward of other functions. I will explain the detail in another issue, so wait for that.

okuta:
I think this functionality is covered by #2368 .

PR #2194, opened 2017-01-30
Add unit test asserting that pooling are (left) inverse of unpooling

delta2323:
It is natural that pooling functions (`max_pooling` and `average_pooling`) should be the left inverse of the unpooling function (i.e. `pooling(unpooling(x)) = x`). This PR adds unit tests that checks this consistency.

I found this equality does not hold in some combination of parameters (window size(`ksize`), stride, and pad). Although such a consistency is natural, I am not certain it should hold for any parameters other than the default one. It is a problem of how we determine the specification of these functions.


gwtnb:
LGTM

PR #2201, opened 2017-01-30
Support mask for cupy.ndarray.__getitem__ when mask and array's shape differ

yuyu2172:
This PR adds support for `cupy.ndarray.__getitem__` to handle a boolean mask array even when the mask's shape and array's shape differ.

Most of the implementation follows/extends that of `_scatter_op_mask_single`.

Note: In addition to that, here are some other things that are added.
+ consistency among names of the masking function for getitem and setitem
+ add tests for cupy.__getitem__ that uses `cupy.ndarray` indices. I restored tests that were removed in #1912.


rezoo:
Good. LGTM.

PR #2175, opened 2017-01-25
Support boolean array indices for scatter_add

yuyu2172:
merge after #2099

This PR supports boolean array indices for `cupy.scatter_add`.

PR #2202, opened 2017-01-30
Support multiple integer array indexing for cupy.scatter_add

yuyu2172:
Merge after #2175 and #2126.

This PR supports multiple integer array indexing for `cupy.scatter_add`.


rezoo:
Sorry for my late reply. I confirmed and LGTM.

PR #2206, opened 2017-01-31
Add to the documentation that cupy.scatter_add can substitute add.at

yuyu2172:
This PR adds to the documentation that `cupy.scatter_add` can substitute `add.at`.


rezoo:
LGTM.

PR #2207, opened 2017-01-31
Add remove_variable option to build_computational_graph

### LINKED ISSUE #156: Add remove_variable option to build_computational_graph
In current implementation, the computational graph is a bipartite graph consisting in `Variable`s and `Function`s. Sometimes we do not need `Variable`s in the visualization of the computational graphs (e.g. Generated graph is too large. We just need to know how `Function`s are composed). To satisfy this requirement, it would be nice to extract `Functions` only from the computational graph and visualize them.

One of the possible implementation is to add `remove_variable` option to `build_computational_graph`.

mitmul:
Fix #156


muupan:
Could you add test cases for this mode?

mitmul:
@muupan Thank you for reviewing this PR! I fixed points you pointed out and added a test for this remove_variable option.

muupan:
Thanks for the fixes. They look good.

I noticed this PR will not only add a `remove_variable` option but also introduce additional two changes:
- adding variable names to labels https://github.com/pfnet/chainer/pull/2207/files#diff-54541860b0345091e1a50777c35c730bR31
- adding default styles https://github.com/pfnet/chainer/pull/2207/files#diff-54541860b0345091e1a50777c35c730bR6

I think these changes will break backward compatibility to some extent because the appearance of a resulting graph will be changed. For example, the figure in this page should be replaced: http://docs.chainer.org/en/stable/reference/graph.html What do you think?

mitmul:
Ah, right. So I'll update the documentation later. Should it be done at the same time as this?

mitmul:
From the offline discussion, I need to add a compatible mode that can output the completely same dot file as the previous version.

mitmul:
Fixed to allow backward compatible mode.

As I documented in the docstring, there are roughly three modes to make a graph. For example, let `y` be the output of MLP model defined in the official MNIST example,

1. Backward compatible mode
```
cg = build_computational_graph(y, variable_style=None, function_style=None, show_name=False)
```
test_0

2. Richest mode (as the dump_graph extension does)
```
cg = build_computational_graph(y)
```
test_no_option

3. Remove variables
```
cg = build_computational_graph(y, remove_variable=True)
```
test_remove

muupan:
@mitmul Can you resolve the conflicts?

chainer/computational_graph.py
```diff
@@ -59,18 +67,35 @@ class ComputationalGraph(object):
...
+            added to the label of the node. Default is `True`.
+
+    .. note::
+
+        To make the same graph as the previous versions, specify
+        `variable_style=None`, `function_style=None`, and `show_name=False`
+        explicitly. The previous versions than `v1.21.0` outputs the same
+        results as when it takes these arguments, but after `v1.22.0`, the
+        output has the richest representation as default, namely, styles are
+        set and names are shown.
```
muupan: > To make the same graph as the previous versions
"the previous versions" is not clear in this sentence. I think it is better to change the order of explanations so that we first explain the default behavior has changed, then show how to reproduce the previous results.

> The previous versions than `v1.21.0`
This sounds like v1.21.0 is not included, is it correct?

And we need to fix the versions because this PR would be merged as v1.23.0.

chainer/computational_graph.py
```diff
@@ -166,6 +237,15 @@ def build_computational_graph(outputs, remove_split=True,
...
         See :class:`TestGraphBuilder` for details.
 
+    .. note::
+
+        To make the same graph as the previous versions, specify
+        `variable_style=None`, `function_style=None`, and `show_name=False`
+        explicitly. The previous versions than `v1.21.0` outputs the same
+        results as when it takes these arguments, but after `v1.22.0`, the
+        output has the richest representation as default, namely, styles are
+        set and names are shown.
```
muupan: Ditto

muupan: Ditto

mitmul:
@muupan Thank you for your comments! I fixed the note. Could you confirm it's appropriate?

muupan:
Thanks for the fixes. LGTM.

PR #2213, opened 2017-02-01
Add NCCL-based MultiprocessParallelUpdater

t-abe:
This PR is the combination of @jekbradbury (#1924) and @anaruse 's (#1895) PRs.
Thank you very much @jekbradbury and  @anaruse  !

It is based on #1924 MuiltiprocessParallelUpdater but
- uses NCCL (https://github.com/NVIDIA/nccl) for interprocess communication instead of IPC,
- the dataset is split and each worker has own iterator (no need to send a batch via pipe).

Unlike #1895, it uses reduce+bcast for implementation easiness.

The imagenet example (train_imagenet_data_parallel.py) seems work to well.

Known issues are
- [x] when a main process uses a gpu other than gpu-0, NCCL throws an error: NCCL_INVALID_RANK
- [x] the reported values (loss, etc.) in subprocesses  are not transferred to a main process
- [x] deadlocking happens when you only use one gpu.
- [x] sometimes does not exit gracefully
- failed when a model has no parameters

also need
- [x] train an ImageNet model and check its accuracy is similar to non-parallel one
- [x] include NCCL to install at the same time or add how to install NCCL into README
- [x] check if NCCL is installed
- [x] raise an error or warning if NCCL is not installed


jekbradbury:
Looks great! Make sure to fix my bug from #1924 with divide versus multiply on the optimizer attributes, though (fixed in the latest version of #1924).
Also, maybe putting the observation through the pipe wouldn't be too slow if you first converted it to ordinary floats?

t-abe:
Thank you! I am going to pick the fix.

It needs to_cpu to convert the reported values to floats.
to_cpu blocks until a memcpy back to CPU is performed.
So python is blocked until all GPU computations (foward/backward) are finished.
Without reporting, python does not wait forwad/backward computation and continue to prepare a next batch.

okuta:
Could you split NCCL files? @t-abe

ir5:
It seems we still need to fix flake8 issues.

chainer/link.py
```diff
@@ -454,6 +457,143 @@ def zerograds(self):
             for param in link._uninitialized_params.values():
                 param._zeroed = True
 
+    def size_num_grads(self):
```
ir5: I believe that the implementation of `Link` class should be as light as possible because it is one of the core classes in chainer. If the core part contains methods that almost no one will use, they can obscure other useful methods. This method is only for MultiprocessParallelUpdater. We can put this method in the outside of `Link` class, say, in in `multiprocess_parallel_updater.py` as a function. This opinion holds for entire part of this PR in `Link` class. To reflect this opinion will require the modification of many parts, so rebuttal is welcome.

chainer/link.py
```diff
@@ -454,6 +457,143 @@ def zerograds(self):
...
+        """Put together all gradient arrays and make a single array
+
+        Return:
+            cupy.ndarray
+        """
+        if self._cpu:
+            raise RuntimeError('Link.gather_grads works only on GPU.')
+
+        size, num = self.size_num_grads()
+        # print("size:{}, num:{}".format(size, num))
```
ir5: Please remove unused comment-outed print.

ir5: This holds for other many parts.

chainer/training/updater.py
```diff
@@ -326,3 +316,16 @@ def update_core(self):
 
         for model in six.itervalues(models_others):
             model.copyparams(model_main)
+
+
+def _calc_loss(model, in_arrays):
```
ir5: This function should be placed in `multiprocess_parallel_updater.py`.

chainer/training/updaters/multiprocess_parallel_updater.py
```diff
@@ -0,0 +1,223 @@
...
+from chainer.training.updater import StandardUpdater, _calc_loss
+
+try:
+    from cupy.cuda import nccl
+    _available = True
+except ImportError:
+    _available = False
+
+
+class _Worker(multiprocessing.Process):
```
ir5: Is it common to start a class name with underscore? I haven't seen this before. I believe that this style applies for only methods/variables in class. c.f., https://docs.python.org/3.4/tutorial/classes.html#tut-private

chainer/variable.py
```diff
@@ -249,6 +249,28 @@ def zerograd(self):
             else:
                 self._grad.fill(0)
 
+    def scatter_grad(self, array, offset):
```
ir5: Ditto. Please consider to implement these methods in outside of `Variable` class.

examples/imagenet/train_imagenet_data_parallel.py
```diff
@@ -0,0 +1,132 @@
...
+        description='Learning convnet from ILSVRC2012 dataset')
+    parser.add_argument('train', help='Path to training image-label list file')
+    parser.add_argument('val', help='Path to validation image-label list file')
+    parser.add_argument('--arch', '-a', choices=archs.keys(),
+                        default='nin', help='Convnet architecture')
+    parser.add_argument('--batchsize', '-B', type=int, default=32,
+                        help='Learning minibatch size')
+    parser.add_argument('--epoch', '-E', type=int, default=10,
+                        help='Number of epochs to train')
+    parser.add_argument('--gpus', '-g', type=int, nargs="*",
```
ir5: `--gpus` should not be an int.

examples/imagenet/train_imagenet_data_parallel.py
```diff
@@ -0,0 +1,132 @@
+#!/usr/bin/env python
+"""Example code of learning a large scale convnet from ILSVRC2012 dataset.
```
ir5: Please explain that this code uses MultiprocessParallelUpdater, and thus this code requires the installation of NCCL.

chainer/training/updaters/multiprocess_parallel_updater.py
```diff
@@ -0,0 +1,223 @@
...
+                del gg
+                self.model.cleargrads()
+                if pp is None:
+                    pp = self.model.gather_params()
+                self.comm.bcast(pp.data.ptr, pp.size, nccl.NCCL_FLOAT, 0,
+                                null_stream.ptr)
+                self.model.scatter_params(pp)
+                pp = None
+
+                # Sending observation via pipe is too slow.
```
ir5: I believe that this is related the issue "the reported values (loss, etc.) in subprocesses are not transferred to a main process". If this issue can't be solved, please state this explicitly in the docstring of `MultiprocessParallelUpdater` (possibly explaining the reason, that communication via pipe is too slow), removing these comments, and mark the issue as solved.

chainer/training/updaters/multiprocess_parallel_updater.py
```diff
@@ -0,0 +1,223 @@
...
+        for iterator in iterators[1:]:
+            assert len(iterator.dataset) == len(iterators[0].dataset)
+
+        # Correct optimizer parameters for new minibatch size
+        optim = optimizer.__class__.__name__
+        if optim in ('Adam', 'AdaGrad', 'RMSprop'):
+            optimizer.eps *= len(devices)
+        elif optim in ('RMSpropGraves', 'AdaDelta'):
+            optimizer.eps *= len(devices) ** 2  # not quite right for AdaDelta
+        else:
```
ir5: This looks a little dangerous since newly added optimizer may not have `.lr` instance variable. How about replacing this with `elif optim in (<all the currently known optimizers>):` and then adding `else: Raise ...`? This may not be the best solution though.

examples/imagenet/train_imagenet_data_parallel.py
jekbradbury: It's a space-separated list of ints.

ir5: Oh sorry, I didn't know this usage. Now this part looks good.

tests/chainer_tests/training_tests/updaters_tests/test_multiprocess_parallel_updater.py
```diff
@@ -0,0 +1,115 @@
...
+
+        cupy.testing.assert_array_equal(model0.conv.W.data, model1.conv.W.data)
+        cupy.testing.assert_array_equal(model0.conv.b.data, model1.conv.b.data)
+        cupy.testing.assert_array_equal(model0.fc.W.data, model1.fc.W.data)
+        cupy.testing.assert_array_equal(model0.fc.b.data, model1.fc.b.data)
+
+    def test_gather_params_raise_on_cpu(self):
+        model = SimpleNet()
+        with self.assertRaises(RuntimeError):
+            mpu.gather_params(model)
```
mitmul: I thinks adding the blow line at the bottom of this file is needed.
```
testing.run_module(__name__, __file__)
```

t-abe: Thanks, I did.

chainer/datasets/sub_dataset.py
```diff
@@ -132,6 +132,49 @@ def split_dataset_random(dataset, first_size, seed=None):
     return split_dataset(dataset, first_size, order)
 
 
+def split_dataset_n(dataset, n, order=None):
+    """Splits a dataset into ``n`` subsets.
```
t-abe: done

ir5:
CI result shows that some test in `test_multiprocess_parallel_updater.py` fails. We need to investigate it.

chainer/training/updaters/multiprocess_parallel_updater.py
```diff
@@ -0,0 +1,376 @@
...
+    num = 0
+    for param in link.params():
+        if param.size == 0:
+            continue
+        size += param.size
+        num += 1
+    return size, num
+
+
+_batch_memcpy = cuda.cupy.ElementwiseKernel(
```
ir5: It is possible that `cupy` is unavailable. This statement should be executed only when `_available` is True. Is this correct?

chainer/training/updaters/__init__.py
```diff
@@ -0,0 +1,3 @@
+from chainer.training.updaters import multiprocess_parallel_updater  # NOQA
+
```
ir5: (Minor comment) Why there is a blank line here?

chainer/training/updaters/multiprocess_parallel_updater.py
```diff
@@ -0,0 +1,376 @@
...
+        self.proc_id = proc_id
+        self.pipe = pipe
+        self.converter = master.converter
+        self.model = master._master
+        self.device = master._devices[proc_id]
+        self.iterator = master._mpu_iterators[proc_id]
+        self.n_devices = len(master._devices)
+
+    def setup(self):
+        print("setup %d" % self.device)
```
ir5: Print message here seems unclear for users. How about explain what is happening here more clearly, or just removing this print message if it is actually unnecessary?

chainer/training/updaters/multiprocess_parallel_updater.py
```diff
@@ -0,0 +1,376 @@
...
+        _, comm_id = self.pipe.recv()
+        self.comm = nccl.NcclCommunicator(self.n_devices, comm_id,
+                                          self.proc_id)
+
+        self.model.to_gpu(self.device)
+        self.reporter = chainer.reporter.Reporter()
+        self.reporter.add_observer('main', self.model)
+        print("%d setup done" % self.device)
```
ir5: ditto.

chainer/training/updaters/multiprocess_parallel_updater.py
```diff
@@ -0,0 +1,376 @@
...
+            if self.comm is not None:
+                gp = gather_params(self._master)
+                self.comm.bcast(gp.data.ptr, gp.size, nccl.NCCL_FLOAT,
+                                0, null_stream.ptr)
+
+    def finalize(self):
+        self._send_message(('finalize', None))
+
+        for worker in self._workers:
+            print("join", worker)
```
ir5: ditto

chainer/training/updaters/multiprocess_parallel_updater.py
```diff
@@ -0,0 +1,376 @@
...
+            worker = _Worker(i, worker_end, self)
+            worker.start()
+            self._workers.append(worker)
+            self._pipes.append(pipe)
+
+        with cuda.Device(self._devices[0]):
+            self._master.to_gpu(self._devices[0])
+            if len(self._devices) > 1:
+                comm_id = nccl.get_unique_id()
+                self._send_message(("set comm_di", comm_id))
```
mitmul: typo of "comm_id"?

chainer/training/updaters/__init__.py
t-abe: I followed style of other `__init__.py`. They separate `import namespace` and `import classes`.

chainer/training/updaters/multiprocess_parallel_updater.py
```diff
@@ -0,0 +1,376 @@
...
+            assert len(iterator.dataset) == len(iterators[0].dataset)
+
+        # Correct optimizer parameters for new minibatch size
+        optim = optimizer.__class__.__name__
+        if optim in ('Adam', 'AdaGrad', 'RMSprop'):
+            optimizer.eps *= len(devices)
+        elif optim in ('RMSpropGraves', 'AdaDelta'):
+            optimizer.eps *= len(devices) ** 2  # not quite right for AdaDelta
+        elif hasattr(optimizer, 'lr'):
+            optimizer.lr /= len(devices)
```
t-abe: I thought it should let users know when parameters are changed rather than "else case".

chainer/training/updaters/__init__.py
ir5: Understood.

ir5:
It seems Travis CI still fails. It looks like setup.py tries to build NCCL related things though no-NCCL setting is configured? I'm currently not sure what is exactly happening.

chainer/training/updaters/multiprocess_parallel_updater.py
ir5: Okay. It looks sensible.

ir5:
LGTM

PR #2214, opened 2017-02-01
Add eps to the variance in fixed_batch_normalization

### LINKED ISSUE #855: eps is a redundant argument of fixed_batch_normalization

delta2323:
Fix #855. See also #2149


gwtnb:
LGTM

PR #2217, opened 2017-02-01
Add snapshot frequency option to mnist example

johshisha:
I changed mnist example to specify frequency of taking a snapshot, because previous scripts had mismatch between comment and script.

### DIFF
--- a/examples/mnist/train_mnist.py
+++ b/examples/mnist/train_mnist.py
@@ -89,7 +89,7 @@ def main():
     trainer.extend(extensions.dump_graph('main/loss'))
 
     # Take a snapshot for each specified epoch
-    frequency = args.epoch if args.frequency == -1 else args.frequency
+    frequency = args.epoch if args.frequency == -1 else max(1, args.frequency)
     trainer.extend(extensions.snapshot(), trigger=(frequency, 'epoch'))
 
     # Write a log of evaluation statistics for each epoch
--- a/examples/mnist/train_mnist.py
+++ b/examples/mnist/train_mnist.py
@@ -40,6 +40,8 @@ def main():
                         help='Number of images in each mini-batch')
     parser.add_argument('--epoch', '-e', type=int, default=20,
                         help='Number of sweeps over the dataset to train')
+    parser.add_argument('--frequency', '-f', type=int, default=20,
+                        help='Frequency of taking a snapshot')
     parser.add_argument('--gpu', '-g', type=int, default=-1,
                         help='GPU ID (negative value indicates CPU)')
     parser.add_argument('--out', '-o', default='result',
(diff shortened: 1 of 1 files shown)


delta2323:
Thank you for your PR! The MNIST example had originally taken a snapshot at each epoch. But its behavior was changed to save only at the end of the training process at some point (since snapshot takes too much time so that Chainer's MNIST example looks slower than that of other frameworks). That is the reason of the inconsistency.

I think it is a bit redundant to offer both `epoch` and `frequency` options. 
As the MNSIT example is not so large, it is rare to take snapshots on the halfway of training. Rather, it is more natural to do it only at the end. If we want to take snapshot at the end of the training process, we have to change both `epoch` and `frequency` option.

So, how about fixing the document only in this PR?

johshisha:
In MNIST example, I know it is redundant to offer both `epoch` and `frequency` options.  
But, as a example script, I think it is more useful to show that frequency of taking a snapshot can be specified for complicated implementation.

delta2323:
This PR would change the behavior of the existing script that invokes the MNIST example with the total epoch different from default one (with `-e` option). To keep the behavior unchanged, I would like to set the default value of `frequency` some special value (e.g. -1). If we set the frequency to this value, the frequency will equal to the total epoch. How do you think of it?

johshisha:
I agree with  you.
I defined the default value of the frequency as `-1`, and the frequency will equal to the total epoch if we set the default value.
For the implementation, I used `if` statements but I think it is not cool.
Is there a better way?

examples/mnist/train_mnist.py
```diff
@@ -86,8 +88,9 @@ def main():
     # The "main" refers to the target link of the "main" optimizer.
     trainer.extend(extensions.dump_graph('main/loss'))
 
-    # Take a snapshot at each epoch
-    trainer.extend(extensions.snapshot(), trigger=(args.epoch, 'epoch'))
+    # Take a snapshot for each specified epoch
+    frequency = args.frequency if args.frequency is not -1 else args.epoch
```
delta2323: Although it is a matter of tastes but I think `frequency = args.epoch if args.frequency == -1 else args.frequency` is better because negation could make the if statement more complicated.

delta2323:
Thank you for the update. I could not come up with simpler solution.

examples/mnist/train_mnist.py
johshisha: I updated.
Please review again.

johshisha:
By the way, I found a small bug that `ZeroDivisionError` occurs when setting frequency to `0`. I can solve it with an if statement, but I think this is a common problem of the `IntervalTrigger` class. So I think we should make a new PR. How do you think of it?

delta2323:
I agree with you in that we should consider zero-interval problem in `IntervalTrigger`. Another choice is to allow only positive values as an interval (i.e. we forbid negative values as well as 0). We have [`_never_fire_trigger`](https://github.com/pfnet/chainer/blob/master/chainer/training/trigger.py#L8) although it is not a part of public APIs currently. So, we can make it public if we want a trigger that is  never ignited.

johshisha:
I updated to allow only positive values as an interval except `-1`.
Please review again. 6841612

delta2323:
Thank you for updating the PR. I am now running all tests. We will merge it when they all pass.

delta2323:
LGTM

PR #2226, opened 2017-02-03
Fix GIL bug in cublas

okuta:


gwtnb:
Does `cublasGetVersion` block? If you added `with nogil:`s defensively, I think it is acceptable.

okuta:
You are right.
Maybe `cublasGetVersion` doesn't block.
But I `with nogil:`s defensively, because this method is only called several times.

PR #2229, opened 2017-02-03
Improved doc of crelu, elu, tanh, broadcast, broadcast_to, concat

shiba24:

Modified functions/links:
 - functions/activation/crelu
 - functions/activation/elu
 - functions/activation/tanh
 - functions/array/broadcast (including broadcast_to)
 - functions/array/concat

I am not sure that this example for elu is good or not. We need a figure in the future.

Minor change on:
 - functions/activation/sigmoid

I found the shape description of this is small letter (s_n).

### DIFF
--- a/chainer/functions/activation/elu.py
+++ b/chainer/functions/activation/elu.py
@@ -50,7 +50,7 @@ class ELU(function.Function):
 def elu(x, alpha=1.0):
     """Exponential Linear Unit function.
 
-    This function is expressed as
+    For a parameter :math:`\\alpha`, it is expressed as
 
     .. math::
         f(x) = \\left \\{ \\begin{array}{ll}
@@ -58,15 +58,30 @@ def elu(x, alpha=1.0):
         \\alpha (\\exp(x) - 1) & {\\rm if}~ x < 0,
         \\end{array} \\right.
 
-    where :math:`\\alpha` is a parameter.
     See: https://arxiv.org/abs/1511.07289
 
     Args:
-        x (~chainer.Variable): Input variable.
-        alpha (float): Parameter :math:`\\alpha`.
+        x (:class:`~chainer.Variable` or :class:`numpy.ndarray` or \
+        :class:`cupy.ndarray`):
+            Input variable. A :math:`(s_1, s_2, ..., s_n)`-shaped float array.
+        alpha (float): Parameter :math:`\\alpha`. Default is 1.0.
 
     Returns:
-        ~chainer.Variable: Output variable.
+        ~chainer.Variable: Output variable. A
+        :math:`(s_1, s_2, ..., s_n)`-shaped float array.
+
+    .. admonition:: Example
+
+        >>> x = np.random.randint(-10, 10, (3, 2)).astype('f')
+        >>> x
+        array([[  2.,   5.],
+               [-10.,  -7.],
+               [ -7.,  -3.]], dtype=float32)
+        >>> y = F.elu(x, alpha=1.)
+        >>> y.data
+        array([[ 2.        ,  5.        ],
+               [-0.99995458, -0.99908811],
+               [-0.99908811, -0.95021296]], dtype=float32)
 
     """
     return ELU(alpha=alpha)(x)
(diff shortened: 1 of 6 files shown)


chainer/functions/activation/elu.py
```diff
@@ -50,23 +50,38 @@ def backward_gpu(self, x, gy):
...
 
-    where :math:`\\alpha` is a parameter.
     See: https://arxiv.org/abs/1511.07289
 
     Args:
-        x (~chainer.Variable): Input variable.
-        alpha (float): Parameter :math:`\\alpha`.
+        x (:class:`~chainer.Variable` or :class:`numpy.ndarray` or \
+        :class:`cupy.ndarray`):
+            Input variable. A :math:`(s_1, s_2, ..., s_n)`-shaped float array.
```
mitmul: `s_n` should be `s_N`?

chainer/functions/activation/sigmoid.py
```diff
@@ -66,13 +66,13 @@ def sigmoid(x, use_cudnn=True):
...
+            Input variable. A :math:`(s_1, s_2, ..., s_N)`-shaped float array.
         use_cudnn (bool): If ``True`` and cuDNN is enabled, then this function
             uses cuDNN as the core implementation.
 
     Returns:
         ~chainer.Variable: Output variable. A
-        :math:`(s_1, s_2, ..., s_n)`-shaped float array.
+        :math:`(s_1, s_2, ..., s_N)`-shaped float array.
 
     .. admonition:: Example
```
mitmul: How about adding an example to show the concrete values of the outputs? For example,

```
It maps the input values into the range of :math:`[0, 1]`.

>>> x = np.random.uniform(-3, 3, (3,)).astype('f')
>>> x
array([-2.43635702,  0.45567897,  2.57577729], dtype=float32)
>>> F.sigmoid(x).data
array([ 0.08044198,  0.6119886 ,  0.92928624], dtype=float32)
```

And before the below line (L.79), how about adding an explanation line to say the shape consistency like:

```
It is an element-wise operation, so the shape of the output array is the same as the input array.
```

chainer/functions/activation/tanh.py
```diff
@@ -57,13 +57,25 @@ def backward_gpu(self, x, gy):
...
         use_cudnn (bool): If ``True`` and cuDNN is enabled, then this function
             uses cuDNN as the core implementation.
 
     Returns:
-        ~chainer.Variable: Output variable.
+        ~chainer.Variable: Output variable. A
+        :math:`(s_1, s_2, ..., s_n)`-shaped float array.
+
+    .. admonition:: Example
+
```
mitmul: How about adding an example to show the concrete values of the outputs? For example,

```
It maps the input values into the range of :math:`[-1, 1]`.

>>> x = np.random.uniform(-3, 3, (3,)).astype('f')
>>> x
array([-2.43635702,  0.45567897,  2.57577729], dtype=float32)
>>> F.tanh(x).data
array([-0.98481113,  0.42655593,  0.98848587], dtype=float32)
```

chainer/functions/activation/tanh.py
```diff
@@ -57,13 +57,25 @@ def backward_gpu(self, x, gy):
...
+        >>> x = np.random.uniform(0, 1, (3, 4)).astype('f')
```
mitmul: How about adding an explanation line before this? e.g.,

```
It is an element-wise operation, so the shape of the output array is the same as the input array.

>>> x = np.random.uniform(0, 1, (3, 4)).astype('f')
```

shiba24: Do we need this? `It is an element-wise operation, so the shape of the output array is the same as the input array.`
Because, we have already shown that the shapes of Args and Returns are same, (L65 and L71). But if it is better to describe it here as well, no problem to add this sentence in the example!

shiba24:
Umm, I thought I had merged, but it seems like I couldn't merge correctly. It didn't pass the tests.
May I ask you directly how to do that, maybe my next working day (Monday, 13)?

shiba24:
Oh I am sorry, the reason why it hadn't passed the test was not because of merging master, but because of flake8.

PR #2236, opened 2017-02-06
Add reshape method to chainer.Variable

yuyu2172:
This PR adds a reshape method to `chainer.Variable`.

The interface follows that of `numpy.ndarray.reshape`.

This PR is useful in a situation when you want to declare reshape at the end of a line rather than at the beginning. A simple example to demonstrate the visual difference is as follows.

```python
import numpy
import chainer
import chainer.functions as F

x = chainer.Variable(numpy.array([1,2,3,4]))

y = F.reshape(F.tile(x, 4), (4, 4))

# an alternative way to write the same thing
y = F.tile(x, 4).reshape(4, 4)
```

### DIFF
--- a/chainer/variable.py
+++ b/chainer/variable.py
@@ -441,7 +441,7 @@ Actual: {0}'''.format(type(data))
            :func:`chainer.functions.reshape` for full documentation,
 
         """
-        if len(shape) == 1 and isinstance(shape[0], tuple):
+        if len(shape) == 1 and isinstance(shape[0], (tuple, list)):
             shape = shape[0]
         return chainer.functions.reshape(self, shape)
 
--- a/chainer/variable.py
+++ b/chainer/variable.py
@@ -434,6 +434,17 @@ Actual: {0}'''.format(type(data))
                             x._grad += gx
             del gxs  # to reduce memory usage
 
+    def reshape(self, *shape):
+        """Returns a variable of a different shape and the same content.
+
+        .. seealso::
+           :func:`chainer.functions.reshape` for full documentation,
+
+        """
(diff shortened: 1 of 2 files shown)


chainer/variable.py
```diff
@@ -434,6 +434,17 @@ def add_cand(cand):
...
 
+    def reshape(self, *shape):
+        """Returns a variable of a different shape and the same content.
+
+        .. seealso::
+           :func:`chainer.functions.reshape` for full documentation,
+
+        """
+        if len(shape) == 1 and isinstance(shape[0], tuple):
+            shape = shape[0]
```
rezoo: Could you describe why you decided to slightly change the specification of `reshape` as above?

yuyu2172: The condition is necessary in the case when shape is passed as a tuple.

For example, the following code would raise error without the condition.

```python
import chainer
import numpy

x = chainer.Variable(numpy.array([1,2,3] ))
x.reshape((3,))
```

By the way, this API is consistent with `numpy.ndarray.reshape`.

rezoo: Ah I overlooked. Thanks!

BTW, as this PR slightly changes API, I will ask the chainer team whether it is OK to merge or not. Please wait for a while.

rezoo: Sorry, one comment: could you add the case where `shape[0]` is `list`?

yuyu2172: Thanks. I fixed the code.

rezoo:
Thank you! Then I would like to merge.

chainer/variable.py
beam2d: @yuyu2172 Would you also make it support list as well as tuple at shape[0]?

PR #2245, opened 2017-02-07
Ignore cudaErrorNoDevice on making device backup

niboshi:
Even if CUDA is available, cudaGetDevice() can return
cudaErrorNoDevice when CUDA_VISIBLE_DEVICE is set to -1.

In such case, this has resulted in CUDARuntimeError.


okuta:
LGTM!

PR #2247, opened 2017-02-07
Improve usage of ndarray.ctypes

okuta:
`ndarray.ctypes.data_as` is slower than `ndarray.ctypes.get_as_parameter`.
```
% python -m timeit -s 'import numpy,ctypes;a = numpy.zeros(10) ' 'a.ctypes.data_as(ctypes.c_void_p)'
100000 loops, best of 3: 2.95 usec per loop

% python -m timeit -s 'import numpy,ctypes;a = numpy.zeros(10) ' 'a.ctypes.get_as_parameter()'
100000 loops, best of 3: 1.9 usec per loop
```


gwtnb:
LGTM

PR #2253, opened 2017-02-09
Pad function

KotaroSetoyama:


KotaroSetoyama:
I've almost finished the implementation. However, I don't have any ideas about what should be checked in `check_type_forward`. If you notice some shortage, I'll implement them.

chainer/functions/array/pad.py
```diff
@@ -0,0 +1,65 @@
...
+    def check_type_forward(self, in_types):
+        type_check.expect(in_types.size() == 1)
+
+    def forward(self, inputs):
+        xp = cuda.get_array_module(*inputs)
+        if len(self.keywords) == 0:
+            return xp.pad(inputs[0], self.pad_width, mode=self.mode),
+        else:
+            return xp.pad(inputs[0], self.pad_width, mode=self.mode,
+                          **self.keywords),
```
delta2323: I thought it works even if `self.keywords` is an empty dictionary and we do not need the branching. Doesn't it work?

chainer/functions/array/pad.py
```diff
@@ -0,0 +1,65 @@
...
+    def __init__(self, pad_width, mode, **keywords):
+        self.mode = mode
+        self.keywords = keywords
+        self.pad_width = pad_width
+        self.pad_bw = numpy.asarray(pad_width)
+        if self.pad_bw.size == 1:
+            self.pad_bw = numpy.repeat(self.pad_bw, 2)
+
+    def check_type_forward(self, in_types):
+        type_check.expect(in_types.size() == 1)
```
delta2323: Could you add a comment that we can check the validity of the input in `xp.pad` (I understand so. Correct me if it is wrong).

chainer/functions/array/pad.py
```diff
@@ -0,0 +1,65 @@
...
+def pad(x, pad_width, mode, **keywords):
+    """Pad an input variable.
+
+    Args:
+        x (chainer.Variable or :class:``numpy.ndarray`` or cupy.ndarray):
+            Input data.
+        pad_width (int or array-like):
+            Number of values padded to the edges of each axis.
+        mode (str):
+            `constant`
```
delta2323: Please add a document what the option `mode` itself indicates, not only does each value of `mode` option (e.g. `constant`) indicates. (e.g. "specifies how the function fills the periphery of the array")

delta2323:
LGTM

PR #2255, opened 2017-02-09
warn when numpy and cupy are used together

soramichi:
This PR is an improvement over #2076. My apologies to making a new PR, but I mistakenly deleted the whole repository once so I cannot push to the branch used in #2076.

This version considers the case where W is a sub-class of `numpy.array` so that this case is also accepted.

### DIFF
--- a/chainer/functions/connection/embed_id.py
+++ b/chainer/functions/connection/embed_id.py
@@ -27,6 +27,10 @@ class EmbedIDFunction(function.Function):
     def forward(self, inputs):
         x, W = inputs
 
+        if(not type_check.same_types(*inputs)):
+            raise ValueError('numpy and cupy must not be used together\n'
+                             'type(W): %s, type(x): %s' % (type(W), type(x)))
+
         xp = cuda.get_array_module(*inputs)
         if chainer.is_debug():
             valid_x = xp.logical_and(0 <= x, x < len(W))
@@ -35,11 +39,6 @@ class EmbedIDFunction(function.Function):
             if not valid_x.all():
                 raise ValueError('Each not ignored `x` value need to satisfy'
                                  '`0 <= x < len(W)`')
-
-        if (issubclass(type(W), numpy.ndarray) !=
-                issubclass(type(x), numpy.ndarray)):
-            raise ValueError('numpy and cupy must not be used together\n'
-                             'type(W): %s, type(x): %s' % (type(W), type(x)))
 
         if self.ignore_label is not None:
             mask = (x == self.ignore_label)
--- a/chainer/functions/connection/embed_id.py
+++ b/chainer/functions/connection/embed_id.py
@@ -36,15 +36,15 @@ class EmbedIDFunction(function.Function):
                 raise ValueError('Each not ignored `x` value need to satisfy'
                                  '`0 <= x < len(W)`')
 
-        if self.ignore_label is not None:
-            mask = (x == self.ignore_label)
-            return xp.where(
-                mask[..., None], 0, W.take(xp.where(mask, 0, x), axis=0)),
-
         if (issubclass(type(W), numpy.ndarray) !=
                 issubclass(type(x), numpy.ndarray)):
             raise ValueError('numpy and cupy must not be used together\n'
                              'type(W): %s, type(x): %s' % (type(W), type(x)))
+
+        if self.ignore_label is not None:
+            mask = (x == self.ignore_label)
+            return xp.where(
+                mask[..., None], 0, W.take(xp.where(mask, 0, x), axis=0)),
 
         return W.take(x, axis=0),
 
--- a/chainer/functions/connection/embed_id.py
+++ b/chainer/functions/connection/embed_id.py
@@ -41,6 +41,10 @@ class EmbedIDFunction(function.Function):
             return xp.where(
                 mask[..., None], 0, W.take(xp.where(mask, 0, x), axis=0)),
 
+        if not issubclass(type(W), type(x)):
+            raise ValueError('numpy and cupy must not be used together\n'
+                             'type(W): %s, type(x): %s' % (type(W), type(x)))
+
         return W.take(x, axis=0),
 
     def backward(self, inputs, grad_outputs):
--- a/chainer/functions/connection/embed_id.py
+++ b/chainer/functions/connection/embed_id.py
@@ -41,7 +41,8 @@ class EmbedIDFunction(function.Function):
             return xp.where(
                 mask[..., None], 0, W.take(xp.where(mask, 0, x), axis=0)),
 
-        if not issubclass(type(W), type(x)):
+        if (issubclass(type(W), numpy.ndarray) !=
+                issubclass(type(x), numpy.ndarray)):
             raise ValueError('numpy and cupy must not be used together\n'
                              'type(W): %s, type(x): %s' % (type(W), type(x)))
 
--- a/chainer/utils/type_check.py
+++ b/chainer/utils/type_check.py
@@ -3,10 +3,12 @@ import operator
 import sys
 import threading
 
-import cupy
 import numpy
 
 from chainer import cuda
+
+if cuda.available:
+    import cupy
 
 
 _thread_local = threading.local()
@@ -491,9 +493,13 @@ def expect(*bool_exprs):
 def same_types(*arrays):
     are_numpy_arrays = map(lambda x: issubclass(type(x), numpy.ndarray),
                            [a for a in arrays])
-    are_cupy_arrays = map(lambda x: issubclass(type(x), cupy.ndarray),
-                          [a for a in arrays])
-    return all(are_numpy_arrays) or all(are_cupy_arrays)
+    all_numpy_arrays = all(are_numpy_arrays)
+    if cuda.available:
+        are_cupy_arrays = map(lambda x: issubclass(type(x), cupy.ndarray),
+                              [a for a in arrays])
+        return all_numpy_arrays or all(are_cupy_arrays)
+    else:
+        return all_numpy_arrays
 
 
 prod = Variable(numpy.prod, 'prod')
--- a/chainer/utils/type_check.py
+++ b/chainer/utils/type_check.py
@@ -7,9 +7,6 @@ import numpy
 
 from chainer import cuda
 
-if cuda.available:
-    import cupy
-
 
 _thread_local = threading.local()
 
@@ -495,7 +492,7 @@ def same_types(*arrays):
                            arrays)
     all_numpy_arrays = all(are_numpy_arrays)
     if cuda.available:
-        are_cupy_arrays = map(lambda x: issubclass(type(x), cupy.ndarray),
+        are_cupy_arrays = map(lambda x: issubclass(type(x), cuda.cupy.ndarray),
                               arrays)
         return all_numpy_arrays or all(are_cupy_arrays)
     else:
--- a/chainer/utils/type_check.py
+++ b/chainer/utils/type_check.py
@@ -3,6 +3,7 @@ import operator
 import sys
 import threading
 
+import cupy
 import numpy
 
 from chainer import cuda
@@ -485,6 +486,14 @@ def expect(*bool_exprs):
     for expr in bool_exprs:
         assert isinstance(expr, Testable)
         expr.expect()
+
+
+def same_types(*arrays):
+    are_numpy_arrays = map(lambda x: issubclass(type(x), numpy.ndarray),
+                           [a for a in arrays])
+    are_cupy_arrays = map(lambda x: issubclass(type(x), cupy.ndarray),
+                          [a for a in arrays])
+    return all(are_numpy_arrays) or all(are_cupy_arrays)
 
 
 prod = Variable(numpy.prod, 'prod')
--- a/chainer/utils/type_check.py
+++ b/chainer/utils/type_check.py
@@ -492,11 +492,11 @@ def expect(*bool_exprs):
 
 def same_types(*arrays):
     are_numpy_arrays = map(lambda x: issubclass(type(x), numpy.ndarray),
-                           [a for a in arrays])
+                           arrays)
     all_numpy_arrays = all(are_numpy_arrays)
     if cuda.available:
         are_cupy_arrays = map(lambda x: issubclass(type(x), cupy.ndarray),
-                              [a for a in arrays])
+                              arrays)
         return all_numpy_arrays or all(are_cupy_arrays)
     else:
         return all_numpy_arrays
--- a/tests/chainer_tests/utils_tests/test_type_check.py
+++ b/tests/chainer_tests/utils_tests/test_type_check.py
@@ -360,11 +360,16 @@ class TestProd(unittest.TestCase):
 
 class TestSameTypes(unittest.TestCase):
 
-    @attr.gpu
     def test_all_numpy_array(self):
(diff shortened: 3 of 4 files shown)


delta2323:
Thank you for PR. Could you write unit tests that checks this change?

chainer/functions/connection/embed_id.py
```diff
@@ -41,6 +41,10 @@ def forward(self, inputs):
             return xp.where(
                 mask[..., None], 0, W.take(xp.where(mask, 0, x), axis=0)),
 
+        if not issubclass(type(W), type(x)):
```
delta2323: Is it enough to check `W` is a subclass of `type(x)`? I mean, should we check the case `x` is a subclass of `numpy.ndarray`?

soramichi: Thanks for the review. You mean something like this right?
```
>>> W = np.array([[1,2], [3,4]])
>>> x = np.matrix("0; 1")
>>> W.take(x) # works
array([[1],
       [2]])
>>> issubclass(type(W), type(x))
False
>>>
```

I will fix it soon to accept this case as well and add unit tests in a couple of days.

soramichi:
Hi, I updated three things:
1. Changed the type-checking condition. Now it passes when both `x` and `W` are sub-classes of `numpy.array`, or neither one of  `x` and `W` is a sub-class of `numpy.array` (which means both `x` and `W` are sub-classes of `cupy.array`).
2. Moved the type-checking code to a place before `mask[..., None], 0, W.take(xp.where(mask, 0, x), axis=0))`, because I found the `take` here also causes the same error in some cases.
3. Added two test cases that check if the exception added by this PR is properly raised when `numpy` and `cupy` are mixed by mistake.

delta2323:
I come to think that this checking mechanism is useful for functions other than `embed_id`. So how about extracting it to somewhere outside of `embed_id` and call it from the Chainer function?

If we use this checking function from various Chainer functions, I think it is better to generalize this function to accept more than two Chainer variables.

soramichi:
>  this checking mechanism is useful for functions other than embed_id

Yes, you are absolutely correct.

I added a generalized version of the type-checking code as `type_check.same_types()` and also some unit-tests for the generalized type-checking. The `same_type()` function returns `True` if and only if (all inputs are sub-classes of `numpy.ndarray`) or (all inputs are sub-classes of `cupy.ndarray`).

Any further discussions are welcome especially for
- Some corner cases I might miss (I haven't yet checked if this generalization suits in any kind of `forward` functions).
- Another approach for generalized type-checking. For example, `Variable()` constructor can save a state to a global variable that remembers all types of past inputs. I mean like this:
```
x = Variable(numpy.array[]) # Variable function saves the fact it received numpy.ndarray
y = Variable(cupy.array[]) # Now Variable knows it created two instances from different array types, which must be an error.
```

A very nice point of this modified Variable approach is we don't need to put any type-checking code outside of the `Variable()` constructor. The disadvantages of it is, first it does not cover `W`, and secondly even if `Variable()` receives two different types of arrays we cannot be 100% sure it's a mistake.

soramichi:
Hi,

I inserted this **type-checking at the beginning all `forward` and `backward` functions in `functions/connection/*.py`, and confirmed all the tests passed** (not yet for chainer functions in other directories). So in this meaning, this generalized version is feasible and correct.

For feasibility of additional coding cost, this method does require adding it to every `forward` and `backward` functions, so the cost might be bit high.

delta2323:
Thank you for your effort! I will reply to your comments below.

> Another approach for generalized type-checking. For example, Variable() constructor can save a state to a global variable that remembers all types of past inputs.

I agree that we can safely forbid to feed `cupy.ndarray` and `numpy.ndarray` together to Chainer functions. But I think it is too restrictive to impose that every function to obey the rule because we sometimes want to feed mixed arrays to Chainer functions. For example, (1) when CuPy does not have some functionalities and we have to calculate with NumPy (or vise versa) (2) when it is costly to transfer the part of input arrays to GPU in typical use case. So I think it is reasonable as an initial step to add the checking mechanism manually if necessary.

delta2323:
> Should I change the code not to use cupy explicitly by replacing issubclass(x, cupy.ndarray) with not issubclass(x, numpy.ndarray)? 

We can check if GPU is available with `chainer.cuda.available`. If this is `True`, we can use CuPy by means of `chainer.cuda.cupy`. Otherwise, all input arrays should be (the subclass) of `numpy.ndarray`. So one of the solution would be

```
def same_types(*arrays):
    are_numpy_arrays = map(lambda x: issubclass(type(x), numpy.ndarray),  [a for a in arrays])
    all_numpy_arrays = all(are_numpy_arrays)
    if chainer.cuda.available:
        are_cupy_arrays = map(lambda x: issubclass(type(x), cupy.ndarray), [a for a in arrays])
        return all_numpy_arrays or all(are_cupy_arrays)
    else:
        return all_numpy_arrays
```

delta2323:
I am concerning additional overhead introduced by this type checking. Is it possible for you to measure the elapsed times with/without inserting `same_types`? If it prohibitively takes times, I think we should use `same_types` only in debug mode (which can be checked by `chainer.is_debug`).

soramichi:
Hi,

I ran the mnist example with type-checking inserted at the beginning of `forward` and `backward` of `functions/connections/linear.py` (which is the only link used in the mnist example). 

The below is the elapsed time of functions included in `type_check.py`, sorted by the `tottime`. The training was done using GRID K520 GPU equipped in a g2.2xlarge AWS instance. The elapsed time was taken for the CPU-side using `cProfile` because `same_types` is executed in the CPU even when the training is offloaded into the GPU.

The result shows `same_types` spent only 0.333 seconds. I believe this is negligible so we don't need to put it specifically in the debug mode. 

```
         56069484 function calls (52802191 primitive calls) in 135.129 seconds

   Ordered by: internal time
   List reduced from 1735 to 34 due to restriction <'type_check.py'>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
3724000/1344000    3.885    0.000    8.983    0.000 /home/ec2-user/src/anaconda3/lib/python3.5/site-packages/chainer/utils/type_check.py:234(_eval_expr)
  1512000    1.576    0.000    2.356    0.000 /home/ec2-user/src/anaconda3/lib/python3.5/site-packages/chainer/utils/type_check.py:263(__init__)
   882000    1.296    0.000    2.657    0.000 /home/ec2-user/src/anaconda3/lib/python3.5/site-packages/chainer/utils/type_check.py:296(__init__)
882000/742000    1.225    0.000    4.386    0.000 /home/ec2-user/src/anaconda3/lib/python3.5/site-packages/chainer/utils/type_check.py:311(eval)
   644000    1.108    0.000   11.108    0.000 /home/ec2-user/src/anaconda3/lib/python3.5/site-packages/chainer/utils/type_check.py:443(expect)
  2170000    1.101    0.000    1.101    0.000 /home/ec2-user/src/anaconda3/lib/python3.5/site-packages/chainer/utils/type_check.py:147(__init__)
   658000    1.093    0.000    1.414    0.000 /home/ec2-user/src/anaconda3/lib/python3.5/site-packages/chainer/utils/type_check.py:390(__init__)
   210000    1.030    0.000    2.742    0.000 /home/ec2-user/src/anaconda3/lib/python3.5/site-packages/chainer/utils/type_check.py:66(_get_type)
   882000    1.000    0.000    3.657    0.000 /home/ec2-user/src/anaconda3/lib/python3.5/site-packages/chainer/utils/type_check.py:159(__getattr__)
   644000    0.765    0.000    2.717    0.000 /home/ec2-user/src/anaconda3/lib/python3.5/site-packages/chainer/utils/type_check.py:91(f)
   224000    0.754    0.000   12.044    0.000 /home/ec2-user/src/anaconda3/lib/python3.5/site-packages/chainer/utils/type_check.py:474(expect)
    98000    0.581    0.000    3.596    0.000 /home/ec2-user/src/anaconda3/lib/python3.5/site-packages/chainer/utils/type_check.py:56(get_types)
   308000    0.576    0.000    1.119    0.000 /home/ec2-user/src/anaconda3/lib/python3.5/site-packages/chainer/utils/type_check.py:282(__init__)
   644000    0.570    0.000    1.952    0.000 /home/ec2-user/src/anaconda3/lib/python3.5/site-packages/chainer/utils/type_check.py:439(__init__)
   658000    0.448    0.000    6.681    0.000 /home/ec2-user/src/anaconda3/lib/python3.5/site-packages/chainer/utils/type_check.py:403(_eval_left)
658000/644000    0.421    0.000    3.147    0.000 /home/ec2-user/src/anaconda3/lib/python3.5/site-packages/chainer/utils/type_check.py:406(_eval_right)
   280000    0.377    0.000    0.769    0.000 /home/ec2-user/src/anaconda3/lib/python3.5/site-packages/chainer/utils/type_check.py:338(__init__)
   280000    0.354    0.000    2.503    0.000 /home/ec2-user/src/anaconda3/lib/python3.5/site-packages/chainer/utils/type_check.py:347(eval)
    78000    0.333    0.000    0.912    0.000 /home/ec2-user/src/anaconda3/lib/python3.5/site-packages/chainer/utils/type_check.py:490(same_types)
   966000    0.315    0.000    0.315    0.000 /home/ec2-user/src/anaconda3/lib/python3.5/site-packages/chainer/utils/type_check.py:290(eval)
   280000    0.313    0.000    1.083    0.000 /home/ec2-user/src/anaconda3/lib/python3.5/site-packages/chainer/utils/type_check.py:162(__getitem__)
   210000    0.260    0.000    0.307    0.000 /home/ec2-user/src/anaconda3/lib/python3.5/site-packages/chainer/utils/type_check.py:32(__init__)
   308000    0.250    0.000    2.992    0.000 /home/ec2-user/src/anaconda3/lib/python3.5/site-packages/chainer/utils/type_check.py:60(<genexpr>)
    98000    0.245    0.000    0.767    0.000 /home/ec2-user/src/anaconda3/lib/python3.5/site-packages/chainer/utils/type_check.py:47(size)
   196000    0.193    0.000    0.267    0.000 /home/ec2-user/src/anaconda3/lib/python3.5/site-packages/chainer/utils/type_check.py:15(get_function_check_context)
    42000    0.176    0.000    1.859    0.000 /home/ec2-user/src/anaconda3/lib/python3.5/site-packages/chainer/utils/type_check.py:363(eval)
   156000    0.128    0.000    0.164    0.000 /home/ec2-user/src/anaconda3/lib/python3.5/site-packages/chainer/utils/type_check.py:493(<lambda>)
    78000    0.095    0.000    0.127    0.000 /home/ec2-user/src/anaconda3/lib/python3.5/site-packages/chainer/utils/type_check.py:491(<lambda>)
    42000    0.082    0.000    0.155    0.000 /home/ec2-user/src/anaconda3/lib/python3.5/site-packages/chainer/utils/type_check.py:353(__init__)
    78000    0.063    0.000    0.063    0.000 /home/ec2-user/src/anaconda3/lib/python3.5/site-packages/chainer/utils/type_check.py:492(<listcomp>)
    42000    0.054    0.000    0.208    0.000 /home/ec2-user/src/anaconda3/lib/python3.5/site-packages/chainer/utils/type_check.py:165(__call__)
    78000    0.035    0.000    0.035    0.000 /home/ec2-user/src/anaconda3/lib/python3.5/site-packages/chainer/utils/type_check.py:494(<listcomp>)
    14000    0.024    0.000    0.155    0.000 /home/ec2-user/src/anaconda3/lib/python3.5/site-packages/chainer/utils/type_check.py:398(eval)
    14000    0.020    0.000    0.051    0.000 /home/ec2-user/src/anaconda3/lib/python3.5/site-packages/chainer/utils/type_check.py:85(f)
```

chainer/utils/type_check.py
```diff
@@ -466,4 +469,17 @@ def expect(*bool_exprs):
         assert isinstance(expr, Testable)
         expr.expect()
 
+
+def same_types(*arrays):
+    are_numpy_arrays = map(lambda x: issubclass(type(x), numpy.ndarray),
+                           [a for a in arrays])
```
delta2323: I think `map(lambda x: ..., arrays)` works.

chainer/utils/type_check.py
```diff
@@ -466,4 +469,17 @@ def expect(*bool_exprs):
...
+    all_numpy_arrays = all(are_numpy_arrays)
+    if cuda.available:
+        are_cupy_arrays = map(lambda x: issubclass(type(x), cupy.ndarray),
+                              [a for a in arrays])
```
delta2323: ditto (replace `[a for a in arrays]` with `arrays`)

delta2323:
Thank you for detailed profiling. Then it seems that we do not have to do the check only in debug mode.

soramichi:
Thank you for the review. Removed the redundant list comprehension as suggested, and improved the tests; removed @attr.gpu from `test_all_numpy_array()`, added a test to use a 1D array, a 2-D array, and a matrix togehter.

delta2323:
Thank you for the update. It is better to make another PR for inserting the type-checking. Because this PR is almost finished and smaller PRs are better.

I am running all GPU tests. I will merge this PR if all tests will be passed.

delta2323:
I ran the GPU tests and found that it fails because of the bug, which is not related to this PR and already fixed in the latest master branch. Could you merge the latest master branch?

soramichi:
I rebased this PR with the latest master as of today. Please check it out again.

delta2323:
@soramichi Thank you. I am running the GPU test again.

chainer/utils/type_check.py
```diff
@@ -7,6 +7,9 @@
 
 from chainer import cuda
 
+if cuda.available:
+    import cupy
```
delta2323: use `cuda.cupy`. It is available regardless of availability of `cupy` (i.e. the value of `cuda.available`).

chainer/utils/type_check.py
```diff
@@ -486,4 +489,17 @@ def expect(*bool_exprs):
         assert isinstance(expr, Testable)
         expr.expect()
 
+
+def same_types(*arrays):
+    are_numpy_arrays = map(lambda x: issubclass(type(x), numpy.ndarray),
+                           arrays)
+    all_numpy_arrays = all(are_numpy_arrays)
+    if cuda.available:
+        are_cupy_arrays = map(lambda x: issubclass(type(x), cupy.ndarray),
```
delta2323: use `cuda.cupy` instead of `cupy`.

tests/chainer_tests/utils_tests/test_type_check.py
```diff
@@ -3,9 +3,14 @@
 
 import numpy
 
+from chainer import cuda
 from chainer import testing
+from chainer.testing import attr
 from chainer.utils import type_check as T
 
+if cuda.available:
+    import cupy
```
delta2323: Use `cuda.cupy`.

chainer/utils/type_check.py
delta2323: Note that we must not remove `cuda.available` here because `cupy.ndarray` is not available when `cuda.available` is `False`.

soramichi:
@delta2323 Sorry for the late response. I changed the code so that it uses `cuda.cupy` instead of bare `cuda`. I also rebased it again because there happened a conflict in `chainer/utils/type_check.py` with the latest master.

delta2323:
LGTM. As the code is frozen for tomorrow's release, I will merge this PR after v1.22.0 is released. Thank you for your contribution!

delta2323:

PR #2258, opened 2017-02-10
2244 fix interval trigger

### LINKED ISSUE #2244: IntervalTrigger may become True at the improper timing when resuming learning
For example,
`IntervalTrigger(3, 'epoch')` will return True at the initial iteration when resuming from epoch 5.
This is probably because 
https://github.com/pfnet/chainer/blob/master/chainer/training/triggers/interval.py#L26
is not taken over from snapshot.

Hakuyume:
Fix #2244 
This code assumes that the batchsize does not change during training.

### DIFF
--- a/chainer/training/triggers/interval.py
+++ b/chainer/training/triggers/interval.py
@@ -1,6 +1,3 @@
-from __future__ import division
-
-
 class IntervalTrigger(object):
 
     """Trigger based on a fixed interval.
@@ -30,7 +27,6 @@ class IntervalTrigger(object):
         self.period = period
         assert unit == 'epoch' or unit == 'iteration'
         self.unit = unit
-        self.count = None
 
     def __call__(self, trainer):
         """Decides whether the extension should be called on this iteration.
@@ -46,18 +42,12 @@ class IntervalTrigger(object):
 
         """
         updater = trainer.updater
-        iteration = updater.iteration
-
         if self.unit == 'epoch':
             epoch = updater.epoch_detail
-            if self.count is None:
-                if iteration > 0:
-                    prev = (epoch * (iteration - 1) / iteration) // self.period
-                else:
-                    prev = 0
-            else:
-                prev = self.count
-            self.count = epoch // self.period
-            return prev != self.count
+            previous_epoch = updater.previous_epoch_detail
+            if previous_epoch is None:
+                return False
+            return previous_epoch // self.period != epoch // self.period
         else:
+            iteration = updater.iteration
             return iteration > 0 and iteration % self.period == 0
--- a/chainer/training/triggers/interval.py
+++ b/chainer/training/triggers/interval.py
@@ -1,4 +1,4 @@
-import math
+from math import floor
 
 
 class IntervalTrigger(object):
@@ -45,6 +45,6 @@ class IntervalTrigger(object):
         epoch = updater.epoch_detail
         if self.unit == 'epoch':
             prev = epoch * (iteration - 1) / iteration
-            return math.floor(prev / self.period) < math.floor(epoch / self.period)
+            return floor(prev / self.period) < floor(epoch / self.period)
         else:
             return iteration > 0 and iteration % self.period == 0
--- a/chainer/training/triggers/interval.py
+++ b/chainer/training/triggers/interval.py
@@ -1,5 +1,4 @@
 from __future__ import division
-from math import floor
 
 
 class IntervalTrigger(object):
@@ -27,6 +26,7 @@ class IntervalTrigger(object):
         self.period = period
         assert unit == 'epoch' or unit == 'iteration'
         self.unit = unit
+        self.count = None
 
     def __call__(self, trainer):
         """Decides whether the extension should be called on this iteration.
@@ -44,8 +44,13 @@ class IntervalTrigger(object):
         updater = trainer.updater
         iteration = updater.iteration
         epoch = updater.epoch_detail
+
         if self.unit == 'epoch':
-            prev = epoch * (iteration - 1) / iteration
-            return floor(prev / self.period) < floor(epoch / self.period)
+            if self.count is None:
+                prev = (epoch * (iteration - 1) / iteration) // self.period
+            else:
+                prev = self.count
+            self.count = epoch // self.period
+            return prev != self.count
         else:
             return iteration > 0 and iteration % self.period == 0
--- a/chainer/training/triggers/interval.py
+++ b/chainer/training/triggers/interval.py
@@ -43,9 +43,9 @@ class IntervalTrigger(object):
         """
         updater = trainer.updater
         iteration = updater.iteration
-        epoch = updater.epoch_detail
 
         if self.unit == 'epoch':
+            epoch = updater.epoch_detail
             if self.count is None:
                 prev = (epoch * (iteration - 1) / iteration) // self.period
             else:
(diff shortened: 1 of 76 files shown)


tests/chainer_tests/training_tests/triggers_tests/test_minmax_trigger.py
```diff
@@ -26,6 +26,10 @@ def epoch(self):
         return 1
 
     @property
+    def epoch_detail(self):
```
delta2323: I am wondering why we modify the unit test of `MinMaxTrigger` as this PR is about `IntervalTrigger`?

Hakuyume: This is because `MinMaxTrigger` calls `IntervalTrigger` internally.
https://github.com/pfnet/chainer/blob/master/chainer/training/triggers/minmax_value_trigger.py#L24
https://github.com/pfnet/chainer/blob/master/chainer/training/util.py#L39

Without this modification, the tests of `MinMaxTrigger` failled.

Hakuyume: Sorry, I could pass the tests of `MinMaxTrigger` without modifying the test code.

delta2323:
> This code assumes that the batchsize does not change during training.

Could you add the assumption to the document of `IntervalTrigger`?

Hakuyume:
I added note.

delta2323:
Could you run `examples/mnist/train_mnist.py` with `--epoch=1`? Our example test failed with division-by-zero error.

```
15:13:46 + coverage run -a --branch examples/mnist/train_mnist.py --epoch=1 --unit=10
15:13:46 Running mnist example
15:16:51 Traceback (most recent call last):
15:16:51   File "examples/mnist/train_mnist.py", line 115, in <module>
15:16:51     main()
15:16:51   File "examples/mnist/train_mnist.py", line 112, in main
15:16:51     trainer.run()
15:16:51   File "/home/user/.local/lib/python2.7/site-packages/chainer/training/trainer.py", line 292, in run
15:16:51     while not stop_trigger(self):
15:16:51   File "/home/user/.local/lib/python2.7/site-packages/chainer/training/triggers/interval.py", line 50, in __call__
15:16:51     prev = (epoch * (iteration - 1) / iteration) // self.period
15:16:51 ZeroDivisionError: float division by zero
```

Hakuyume:
I missed the case that `IntervalTrigger` is used as stop trigger.
I added some test codes and fixed `IntervalTrigger`.

delta2323:
I am a little worried ()
I come to think

delta2323:
I am a little concern about the inconsistent behavior caused by the rounding error. For example, I have such a case in my mind: suppose that mathematically `updater.epoch_detail * (iteration - 1) / iteration` equals to some integer (say 1000) exactly, but because of the rounding problem, the value could be less than the integer. Then, the actual `prev` value could be different from expected one.

But as I have not checked it yet, It could be unnecessary worry. Do you have an opinion on this problem?

Hakuyume:
As you pointed out, the rounding error causes an undesired behavior.
If `(updater.epoch_detail * (iteration - 1) / iteration) // period` is less than `prev // period`, this trigger fires twice (before and after resuming).
I could easily reproduce this behavior using `round()` in `DummyUpdater.epoch_detail`.

To fix this bug, I think we have to modify `Updater` to save the previous epoch in the snapshot.

delta2323:
Thank you for considering the problem. It seems reasonable to me to have `Updater` serialize the value of previous `epoch_detail` is one of the option. 

The other choice I come up with is to implement `serialize` method to `IntervalTrigger` to save `prev` value and to call the method in `Updater.serialize`. But the former solution seems better to me because (1) `Updater` have to check if the trigger is `IntervalTrigger` or has the `serialize` method (2) it is natural for `IntervalTrigger` to get the previous value of `epoch_detail` as well as current `epoch_detail`.

In either case, we must be careful to keep the [model format compatibility](http://docs.chainer.org/en/latest/compatibility.html#model-format-compatibility).

Hakuyume:
I also think the former solution is better because all resumable triggers need the previous value of `epoch_detail`.
It is redundant that each trigger keeps its own value of previous `epoch_detail`.

I will implement the former solution.

Hakuyume:
I updated `SerialIterator`, `MultiProcessIterator`, and `StandardUpdater` to support `previous_epoch_detail`.
For the model format compatibility, an iterator guesses `previous_epoch_detail` if its information is absent.

chainer/training/triggers/interval.py
```diff
@@ -40,9 +39,11 @@ def __call__(self, trainer):
         """
         updater = trainer.updater
         if self.unit == 'epoch':
-            prev = self.count
-            self.count = updater.epoch_detail // self.period
-            return prev != self.count
+            epoch = updater.epoch_detail
+            previous_epoch = updater.previous_epoch_detail
```
delta2323: I think `getattr(updater, previous_epoch_detail, None)` is safer.

chainer/training/triggers/interval.py
```diff
@@ -40,9 +39,11 @@ def __call__(self, trainer):
         """
         updater = trainer.updater
         if self.unit == 'epoch':
-            prev = self.count
-            self.count = updater.epoch_detail // self.period
-            return prev != self.count
+            epoch = updater.epoch_detail
```
delta2323: It is confusing as Chainer clearly distinguish `epoch` with `epoch_detail`.

tests/chainer_tests/iterators_tests/test_multiprocess_iterator.py
```diff
@@ -185,4 +244,94 @@ def test_copy_not_repeat(self):
         for _ in range(2):
             self.assertRaises(StopIteration, copy_it.next)
 
+
+@testing.parameterize(*testing.product({
+    'n_prefetch': [1, 2],
+    'shared_mem': [None, 1000000],
+}))
+class TestMultiprocessIteratorSerialize(unittest.TestCase):
```
delta2323: Is it possible to send this unittest as another PR?

tests/chainer_tests/iterators_tests/test_serial_iterator.py
```diff
@@ -147,4 +211,83 @@ def test_iterator_shuffle_nondivisible(self):
         self.assertNotEqual(out[0:10], out[10:20])
 
 
+class TestSerialIteratorSerialize(unittest.TestCase):
```
delta2323: ditto (could you extract it and make another PR?)

Hakuyume:
I created another PR for serialization tests #2346.

Hakuyume:
I also divide `previous_epoch_detail` feature into another PR #2347.

chainer/training/triggers/interval.py
Hakuyume: `getattr` is safer in the terms of not raising any exception.
But setting `previous_epoch(_detail)` to `None` causes undesired behavior (same as #2244).
I think it is better to raise exception (or warning) when `updater` does not have `previous_epoch_detail` property.

chainer/training/triggers/interval.py
Hakuyume: I will fix it.

delta2323:
I will review this PR after #2347 is done.

Hakuyume:
This PR has a lot of unused commits because I changed the direction of this PR. I am planning to rebase this PR on #2347 and #2472. If it is better to keep all history, I will merge instead.

delta2323:
@Hakuyume It is OK to close this PR and create another PR on top of them.

Hakuyume:
@delta2323 Thank you. I made another PR (#2484).

PR #2263, opened 2017-02-12
Partial support 64bit address

okuta:
rel #1894.


okuta:
I checked about performance.
The calculation takes four times longer.

before
```
0.0242472839355
0.00388062866211
```

#1894
```
0.037744418335
0.017447015625
```

This PR.
```
0.0343400909424
0.0140911513672
```

code
```
import cupy
from chainer import cuda

def f(cnt):
    stream = cuda.Stream.null
    start = stream.record()
    a = cupy.ones((3000, 2, 2, 2, 1)).T
    b = cupy.ones((3000, 2, 2, 2, 1))
    for i in range(cnt):
        a + b
    end = stream.record()
    end.synchronize()
    elapsed_time = cupy.cuda.get_elapsed_time(start, end) / cnt / 1000
    print(elapsed_time)

f(10)
f(30000)
```

uchida:
A bottleneck seems `_ind.set(i)` 64bit integer calculation in `CUPY_FOR`.
How about divide `n`, upper limit for `CUPY_FOR`, into smaller integer in 32bit ranges.

okuta:
@uchida  Thank you for your advice.
I fixed `_ind.set(i)`. This PR is as fast as master branch.

After optimized result.
```
0.024789302063
0.0038293059082
```

tests/cupy_tests/core_tests/test_ndarray_reduction.py
```diff
@@ -20,6 +20,15 @@ def test_max_axis_large(self, xp, dtype):
...
 
+    """
+    @testing.attr.slow
+    @testing.for_dtypes(['?', 'b', 'h', 'e'])
+    @testing.numpy_cupy_allclose()
+    def test_max_axis_huge(self, xp, dtype):
+        a = testing.shaped_random((2048, 1024, 1024), xp, dtype)
+        return a.max(axis=0)
+    """
+
```
niboshi: Why is this test commented out? Also, I think the size of array should be greater than 2^31 to test huge integers.

(applies to other tests as well)

cupy/core/carray.cuh
```diff
@@ -226,19 +246,35 @@ private:
...
-      a /= s;
+  __device__ void set(ptrdiff_t i) {
+    if (ndim == 0) {
+      index_[0] = i;
+      return;
     }
-    if (ndim > 0) {
-      index_[0] = a;
+    if (size_ > 1LL << 31) {
+      size_t a = i;
```
niboshi: I think signedness of `a` won't affect the behavior. If so, wouldn't it be more intuitive (easier to read) if every integral types are signed?

(But this is not relevant to this PR)

cupy/core/carray.cuh
```diff
@@ -226,19 +246,35 @@ private:
...
+      for (int dim = ndim; --dim > 0; ) {
+        size_t s = shape_[dim];
+        index_[dim] = (a % s);
+        a /= s;
+      }
+      if (ndim > 0) {
+        index_[0] = a;
+      }
+    } else {
+      unsigned int a = i;
```
niboshi: Ditto (signedness of `a`)

cupy/core/carray.cuh
okuta: I use sinedness type for performance.

niboshi: Ah I see. It's ok then.

niboshi:
Don't you need test cases?
In my environment, the `xp.empty(2**31)` works but `xp.ones(2**31)` causes an error:

```
:
  File "cupy/core/core.pyx", line 361, in cupy.core.core.ndarray.fill (cupy/core/core.cpp:9979)
    cpdef fill(self, value):
  File "cupy/core/core.pyx", line 373, in cupy.core.core.ndarray.fill (cupy/core/core.cpp:9925)
    elementwise_copy(value, self, dtype=self.dtype)
  File "cupy/core/core.pyx", line 1533, in cupy.core.core.elementwise_copy (cupy/core/core.cpp:57243)
    return _elementwise_copy(*args, **kwargs)
  File "cupy/core/elementwise.pxi", line 780, in cupy.core.core.ufunc.__call__ (cupy/core/core.cpp:47680)
    kern.linear_launch(indexer.size, inout_args)
  File "cupy/cuda/function.pyx", line 143, in cupy.cuda.function.Function.linear_launch (cupy/cuda/function.cpp:4260)
    _launch(self.ptr,
  File "cupy/cuda/function.pyx", line 107, in cupy.cuda.function._launch (cupy/cuda/function.cpp:3601)
    cp = _pointer(a)
  File "cupy/cuda/function.pyx", line 71, in cupy.cuda.function._pointer (cupy/cuda/function.cpp:2710)
    return (<core.Indexer>x).get_pointer()
  File "cupy/core/carray.pxi", line 69, in cupy.core.core.Indexer.get_pointer (cupy/core/core.cpp:32749)
    return to_cindexer(self.size, self.shape)
  File "cupy/core/carray.pxi", line 49, in cupy.core.core.to_cindexer (cupy/core/core.cpp:32300)
    return CIndexer(size, shape)
  File "cupy/core/carray.pxi", line 43, in cupy.core.core.CIndexer.__init__ (cupy/core/core.cpp:32232)
    self.val.shape_and_index[i] = shape[i]
OverflowError: value too large to convert to int
```

okuta:
That is correct.
This PR supports 64bit address partially. So, CArray type supports only 31bit size.
I will support 64bit address completely other PR.

niboshi:
Mmm, I don't still understand.
What is enabled by this PR?
How about merging this PR into v2 together with breaking changes?

okuta:
I close this PR. Please see https://github.com/cupy/cupy/pull/31.
I will create another PR for Chainer v2.

PR #2266, opened 2017-02-13
Remove matplotlib.use from PlotReport

### LINKED ISSUE #2146: matplotlib.use('Agg') conflicts with user code
In this line in `chainer/chainer/training/extensions/plot_report.py` the `matplotlib` backend is changed [Source](https://github.com/pfnet/chainer/blob/master/chainer/training/extensions/plot_report.py#L16):

    matplotlib.use('Agg')

Unfortunately, this can interfere with users code. For example, when the user sets the backend himself anywhere, it is not known, whether his setting or the Chainer settings wins (is imported first).

The `plot_report` gets imported, when `extensions` is imported. For now, I just removed from the corresponding `__init__` file locally, which is definitely not a clean solution.

mitmul:
Fix #2146

### DIFF
--- a/chainer/training/extensions/plot_report.py
+++ b/chainer/training/extensions/plot_report.py
@@ -47,7 +47,6 @@ class PlotReport(extension.Extension):
     It also adds ``'epoch'`` and ``'iteration'`` entries to each result
     dictionary, which are the epoch and iteration counts at the output.
 
-
     .. warning::
 
         If your environment needs to specify a backend of matplotlib
--- a/chainer/training/extensions/plot_report.py
+++ b/chainer/training/extensions/plot_report.py
@@ -47,14 +47,22 @@ class PlotReport(extension.Extension):
     It also adds ``'epoch'`` and ``'iteration'`` entries to each result
     dictionary, which are the epoch and iteration counts at the output.
 
-    If your environment needs to specify a backend of matplotlib explicitly,
-    please call ``matplotlib.use`` before importing Chainer. For example,
 
-    .. code-block:: python
-        import matplotlib
-        matplotlib.use('Agg')
+    .. warning::
 
-        import chainer
+        If your environment needs to specify a backend of matplotlib
+        explicitly, please call ``matplotlib.use`` before importing Chainer.
+        For example:
+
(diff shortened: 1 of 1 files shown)


delta2323:
Do you have an idea on [this problem](https://github.com/pfnet/chainer/issues/2146#issuecomment-278374592) about the import of `matplotlib.pyplot`

chainer/training/extensions/plot_report.py
```diff
@@ -50,6 +47,15 @@ class PlotReport(extension.Extension):
...
     dictionary, which are the epoch and iteration counts at the output.
 
+    If your environment needs to specify a backend of matplotlib explicitly,
+    please call ``matplotlib.use`` before importing Chainer. For example,
+
+    .. code-block:: python
+        import matplotlib
+        matplotlib.use('Agg')
+
+        import chainer
```
delta2323: Please also write that `matplotlib.use` is disabled once the user imports `PlotReport`.

chainer/training/extensions/plot_report.py
```diff
@@ -50,6 +47,22 @@ class PlotReport(extension.Extension):
...
+
+        .. code-block:: python
+
+            import matplotlib
+            matplotlib.use('Agg')
+
+            import chainer
+
+        Then, once ``chainer.training.extensions`` is imported,
+        ``matplotlib.use`` will never be available.
```
delta2323: I think "``matplotlib.use`` will have no effect" is more appropriate because ``matplotlib.use`` does be available in the sense that we can call the function anyway.  Also, could you add the link to the corresponding part of the document of matplotlib [here](http://matplotlib.org/faq/usage_faq.html#what-is-a-backend) ?

mitmul:
Thank you for your review. I fixed the points. Please check it again!

delta2323:
Thank you for the update. Basically, your change is OK. Please fix flake8 and autopep8 error.

https://travis-ci.org/pfnet/chainer/jobs/203288906

mitmul:
@delta2323 Thanks. I fixed the flake8 errors. Please check it again~

delta2323:
LGTM

PR #2272, opened 2017-02-14
Add spatial transformer sampler

yuyu2172:
This PR adds a bilinear interpolation of an image, which is described in a paper "Spatial Transformer Networks" (https://arxiv.org/abs/1506.02025).

### DIFF
--- a/chainer/functions/array/spatial_transformer_sampler.py
+++ b/chainer/functions/array/spatial_transformer_sampler.py
@@ -100,16 +100,20 @@ class SpatialTransformerSampler(function.Function):
         w3 = w3.astype(x.dtype)
         w4 = w4.astype(x.dtype)
 
-        ys = []
-        for b in range(B):
-            elem = (w1[b, :, None] * x[b, :, v0[b], u0[b]] +
-                    w2[b, :, None] * x[b, :, v0[b], u1[b]] +
-                    w3[b, :, None] * x[b, :, v1[b], u0[b]] +
-                    w4[b, :, None] * x[b, :, v1[b], u1[b]])
-            elem = elem.reshape(out_H, out_W, C)
-            elem = elem.transpose(2, 0, 1)
-            ys.append(elem)
-        y = xp.concatenate([xp.expand_dims(y, axis=0) for y in ys], axis=0)
+        x_indexed_1 = xp.concatenate([xp.expand_dims(
+            x[b, :, v0[b], u0[b]], axis=0) for b in range(B)], axis=0)
+        x_indexed_2 = xp.concatenate([xp.expand_dims(
+            x[b, :, v0[b], u1[b]], axis=0) for b in range(B)], axis=0)
+        x_indexed_3 = xp.concatenate([xp.expand_dims(
+            x[b, :, v1[b], u0[b]], axis=0) for b in range(B)], axis=0)
+        x_indexed_4 = xp.concatenate([xp.expand_dims(
+            x[b, :, v1[b], u1[b]], axis=0) for b in range(B)], axis=0)
+        y = w1[:, :, None] * x_indexed_1
+        y += w2[:, :, None] * x_indexed_2
+        y += w3[:, :, None] * x_indexed_3
+        y += w4[:, :, None] * x_indexed_4
+
+        y = y.reshape(B, out_H, out_W, C).transpose(0, 3, 1, 2)
         return y,
 
     def backward_cpu(self, inputs, grad_outputs):
@@ -185,25 +189,27 @@ class SpatialTransformerSampler(function.Function):
         wv1 = wv1.astype(gy.dtype)
 
         # --- gu, gv
-        gus = []
-        gvs = []
-        for b in range(B):
-            gu_elem = (-wv1[b, :, None] * x[b, :, v0[b], u0[b]] +
-                       wv1[b, :, None] * x[b, :, v0[b], u1[b]] -
-                       wv0[b, :, None] * x[b, :, v1[b], u0[b]] +
-                       wv0[b, :, None] * x[b, :, v1[b], u1[b]])
-            gu_elem = gu_elem.reshape(out_H, out_W, C)
-            gu_elem = gu_elem.transpose(2, 0, 1)
-            gus.append(gu_elem)
-            gv_elem = (-wu1[b, :, None] * x[b, :, v0[b], u0[b]] -
-                       wu0[b, :, None] * x[b, :, v0[b], u1[b]] +
-                       wu1[b, :, None] * x[b, :, v1[b], u0[b]] +
-                       wu0[b, :, None] * x[b, :, v1[b], u1[b]])
-            gv_elem = gv_elem.reshape(out_H, out_W, C)
-            gv_elem = gv_elem.transpose(2, 0, 1)
-            gvs.append(gv_elem)
-        gu = xp.concatenate([xp.expand_dims(g, axis=0) for g in gus], axis=0)
-        gv = xp.concatenate([xp.expand_dims(g, axis=0) for g in gvs], axis=0)
+        x_indexed_1 = xp.concatenate([xp.expand_dims(
+            x[b, :, v0[b], u0[b]], axis=0) for b in range(B)], axis=0)
+        x_indexed_2 = xp.concatenate([xp.expand_dims(
+            x[b, :, v0[b], u1[b]], axis=0) for b in range(B)], axis=0)
+        x_indexed_3 = xp.concatenate([xp.expand_dims(
+            x[b, :, v1[b], u0[b]], axis=0) for b in range(B)], axis=0)
+        x_indexed_4 = xp.concatenate([xp.expand_dims(
+            x[b, :, v1[b], u1[b]], axis=0) for b in range(B)], axis=0)
+
+        gu = -wv1[:, :, None] * x_indexed_1
+        gu += wv1[:, :, None] * x_indexed_2
+        gu -= wv0[:, :, None] * x_indexed_3
+        gu += wv0[:, :, None] * x_indexed_4
+
+        gv = -wu1[:, :, None] * x_indexed_1
+        gv -= wu0[:, :, None] * x_indexed_2
+        gv += wu1[:, :, None] * x_indexed_3
+        gv += wu0[:, :, None] * x_indexed_4
+
+        gu = gu.reshape(B, out_H, out_W, C).transpose(0, 3, 1, 2)
+        gv = gv.reshape(B, out_H, out_W, C).transpose(0, 3, 1, 2)
 
         gu *= gy
         gv *= gy
--- a/chainer/functions/array/spatial_transformer_sampler.py
+++ b/chainer/functions/array/spatial_transformer_sampler.py
@@ -36,31 +36,32 @@ class SpatialTransformerSampler(function.Function):
         return self._forward(inputs)
 
     def forward_gpu(self, inputs):
-        if cuda.cudnn_enabled and self.use_cudnn and _cudnn_version >= 5000:
-            x, grid = inputs
-            out_shape = x.shape[:2] + grid.shape[2:]
-            y = cuda.cupy.empty(out_shape, dtype=x.dtype)
-            shape = numpy.array(out_shape, dtype=numpy.int32)
-            x = cuda.cupy.ascontiguousarray(x)
-            grid_t = cuda.cupy.transpose(grid, (0, 2, 3, 1))
-            grid_t = cuda.cupy.ascontiguousarray(grid_t)
+        if not (cuda.cudnn_enabled and self.use_cudnn and
+                _cudnn_version >= 5000):
+            return self._forward(inputs)
+        x, grid = inputs
+        out_shape = x.shape[:2] + grid.shape[2:]
+        y = cuda.cupy.empty(out_shape, dtype=x.dtype)
+        shape = numpy.array(out_shape, dtype=numpy.int32)
+        x = cuda.cupy.ascontiguousarray(x)
+        grid_t = cuda.cupy.transpose(grid, (0, 2, 3, 1))
+        grid_t = cuda.cupy.ascontiguousarray(grid_t)
 
-            handle = cudnn.get_handle()
-            x_desc = cudnn.create_tensor_descriptor(x)
-            y_desc = cudnn.create_tensor_descriptor(y)
+        handle = cudnn.get_handle()
+        x_desc = cudnn.create_tensor_descriptor(x)
+        y_desc = cudnn.create_tensor_descriptor(y)
 
-            self.st_desc =\
-                cuda.cupy.cudnn.create_spatial_transformer_descriptor(
-                    _sampler_type, grid.dtype, len(shape), shape.ctypes.data)
+        self.st_desc =\
+            cuda.cupy.cudnn.create_spatial_transformer_descriptor(
+                _sampler_type, grid.dtype, len(shape), shape.ctypes.data)
 
-            one = numpy.array(1, dtype=x.dtype).ctypes
-            zero = numpy.array(0, dtype=x.dtype).ctypes
-            libcudnn.spatialTfSamplerForward(
-                handle, self.st_desc.value, one.data,
-                x_desc.value, x.data.ptr, grid_t.data.ptr, zero.data,
-                y_desc.value, y.data.ptr)
-            return y,
-        return self._forward(inputs)
+        one = numpy.array(1, dtype=x.dtype).ctypes
+        zero = numpy.array(0, dtype=x.dtype).ctypes
+        libcudnn.spatialTfSamplerForward(
+            handle, self.st_desc.value, one.data,
+            x_desc.value, x.data.ptr, grid_t.data.ptr, zero.data,
+            y_desc.value, y.data.ptr)
+        return y,
 
     def _forward(self, inputs):
         x, grid = inputs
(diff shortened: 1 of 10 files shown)


mitmul:
Hi, thank you very much for sending this nice PR! The shortened names like `spatial_tf_sampler` and `SpatialTfSampler` and some other abbreviation for `transformer`, possibly, seem not to be so necessary for me. Could you consider to replace them with the original verbose name?

yuyu2172:
Thanks for your comments.

The abbreviation followed cudnn API.
I agree that it is somewhat inconsistent there are mix of functions with and without abbreviation.

I didn't change names of functions that wrap cudnn. Is that O.K.?

ir5:
Thank you for your PR. This PR is a little long, so I would like to review the test code at first.
About the output array, the current test code checks only its dimension, but I'm afraid that this is insufficient. I believe that it would be better to contain a test code that checks the values of the output array as well.
One way for this is to pass a trivial transformation (identity, 90 degree rotation) for the `grid` argument, then check if each pixel is correct.
Another way may be to create a hand-created input and compute the correct output (the result of bilinear interpolation) manually, then check if the output of the function matches to the intended output. But this may be too much.

ir5:
Just for check, I applied the current function implementation to an image with manually set transformation (grid), and got a desired result.
Input:
c
Output:
c_trans

For 2x2 image in [wikipedia of bilinear interpolation](https://en.wikipedia.org/wiki/Bilinear_interpolation#/media/File:Bilininterp.png), I got a similar result.
test_trans
So, although I haven't checked the code thoroughly yet, the code seems to work correctly.

chainer/functions/array/spatial_transformer_sampler.py
```diff
@@ -0,0 +1,288 @@
...
+    def __init__(self, use_cudnn=True):
+        self.use_cudnn = use_cudnn
+
+    def check_type_forward(self, in_types):
+        n_in = in_types.size()
+        type_check.expect(2 == n_in)
+
+        x_type = in_types[0]
+        grid_type = in_types[1]
+        type_check.expect(
```
ir5: Don't you have to check `x_type.shape[1]` is equal to 2?

chainer/functions/array/spatial_transformer_sampler.py
```diff
@@ -0,0 +1,288 @@
...
+        x (~chainer.Variable):  Input variable of shape :math:`(n, c_I, h, w)`.
+        grid (~chainer.Variable): Coordinate variable of shape
+            :math:`(n, 2, h_O, w_O)`. Each coordinate defines the spatial
+            location in the input where a sampling kernel is applied to get
+            the value at a particular pixel in the output.
+            ``grid[idx, :, i, j]`` corresponds to the coordinate that is used
+            to sample the values for an output pixel at location
+            :math:`(i, j)`.
+
+            In the 2nd dimension, the first coordinate corresponds to the
```
ir5: It might be worth noting that the coordinate (-1, -1) corresponds to the upper-left corner of the input image.

chainer/functions/array/spatial_transformer_sampler.py
```diff
@@ -0,0 +1,288 @@
...
+            location along the horizontal axis, and the second coordinate
+            corresponds to the location along the vertical axis.
+
+            The values of this variable is clipped in range :math:`[-1, 1]`.
+            For example, a coordinate with value :math:`(-1, 1)` samples
+            most of the values from a pixel at the left bottom of the input
+            image.
+        use_cudnn (bool): If ``True``, then this function uses cuDNN if
```
ir5: It might be worth noting that cuDNN is available for version >= 5.0.0.

chainer/functions/array/spatial_transformer_sampler.py
yuyu2172: Right. I needed to check the dimension of the second axis of the grid.

chainer/functions/array/spatial_transformer_sampler.py
yuyu2172: I added what you suggested.

chainer/functions/array/spatial_transformer_sampler.py
yuyu2172: I added what you suggested.

yuyu2172:
Thanks for reviewing.
I will add tests for forward pass in some simple cases.
Note that the tests for backward pass somewhat guarantee the correctness of the forward pass.

yuyu2172:
Here is an example code to check spatial transformer if someone wants to try out quickly.

```python
import matplotlib.pyplot as plt
import numpy as np
from skimage.data import astronaut

import chainer.functions as F


img = astronaut()
img_c = img.transpose(2, 0, 1)[None].astype(np.float32)  # (1, 3, H, W)

mesh = np.meshgrid(
    np.linspace(-1., -0., num=128), np.linspace(-1., 0., num=128))
grid = np.stack(mesh)[None]  # (1, 2, h_o, w_o)  

out = F.spatial_transformer_sampler(img_c, grid)
out_img = out.data[0].transpose(1, 2, 0)  # (H, W, 3)

plt.subplot(1, 2, 1)
plt.imshow(img)
plt.subplot(1, 2, 2)
plt.imshow(out_img.astype(np.uint8))
plt.show()
```

yuyu2172:
@ir5 
I added test cases checking forward map that should result in the identity map and the counter clockwise rotation.

tests/chainer_tests/functions_tests/array_tests/test_spatial_transformer_sampler.py
```diff
@@ -0,0 +1,125 @@
...
+from chainer import testing
+from chainer.testing import attr
+from chainer.testing import condition
+
+
+def _identiy_grid(in_shape):
+    mesh = numpy.meshgrid(
+        numpy.linspace(-1., 1., num=in_shape[2]),
+        numpy.linspace(-1., 1., num=in_shape[3]))
+    grid = numpy.stack(mesh).astype(numpy.float32)  # (2, H, W)
```
ir5: Test results indicate that numpy.stack is not available in numpy v1.9. Please use another method.

tests/chainer_tests/functions_tests/array_tests/test_spatial_transformer_sampler.py
```diff
@@ -0,0 +1,125 @@
...
+    grid = numpy.repeat(grid[None], in_shape[0], axis=0)
+    return grid
+
+
+def _rotate_grid(in_shape):
+    mesh = numpy.meshgrid(
+        numpy.linspace(-1., 1., num=in_shape[2]),
+        numpy.linspace(-1., 1., num=in_shape[3]))
+    mesh = [numpy.rot90(mesh[0]), numpy.rot90(mesh[1])]
+    grid = numpy.stack(mesh).astype(numpy.float32)
```
ir5: ditto.

tests/chainer_tests/functions_tests/array_tests/test_spatial_transformer_sampler.py
```diff
@@ -0,0 +1,125 @@
...
+    return grid
+
+
+def _rotate_BCHW(x):
+    tmp = []
+    for i in range(x.shape[0]):
+        x_i = x[i].transpose(1, 2, 0)
+        x_i = numpy.rot90(x_i)
+        tmp.append(x_i.transpose(2, 0, 1))
+    return numpy.stack(tmp)
```
ir5: ditto.

chainer/functions/array/spatial_transformer_sampler.py
```diff
@@ -0,0 +1,289 @@
...
+        w3 = (u1 - u) * (v - v0)
+        w4 = (u - u0) * (v - v0)
+        w1 = w1.astype(x.dtype)
+        w2 = w2.astype(x.dtype)
+        w3 = w3.astype(x.dtype)
+        w4 = w4.astype(x.dtype)
+
+        ys = []
+        for b in range(B):
+            tmp = (w1[b, :, None] * x[b, :, v0[b], u0[b]] +
```
ir5: `tmp` looks ambiguous as a variable name. How about rename it to another name? (e.g., `elem`, `y`, etc.)

chainer/functions/array/spatial_transformer_sampler.py
```diff
@@ -0,0 +1,289 @@
...
+        wu0 = wu0.astype(gy.dtype)
+        wu1 = wu1.astype(gy.dtype)
+        wv0 = wv0.astype(gy.dtype)
+        wv1 = wv1.astype(gy.dtype)
+
+        # --- gu, gv
+        gus = []
+        gvs = []
+        for b in range(B):
+            gu_tmp = (-wv1[b, :, None] * x[b, :, v0[b], u0[b]] +
```
ir5: ditto.

chainer/functions/array/spatial_transformer_sampler.py
```diff
@@ -0,0 +1,289 @@
...
+            if xp is numpy:
+                scatter_add = numpy.add.at
+            else:
+                scatter_add = xp.scatter_add
+            scatter_add(dydx, (v0[b], u0[b], samples_arange), wu1[b] * wv1[b])
+            scatter_add(dydx, (v0[b], u1[b], samples_arange), wu0[b] * wv1[b])
+            scatter_add(dydx, (v1[b], u0[b], samples_arange), wu1[b] * wv0[b])
+            scatter_add(dydx, (v1[b], u1[b], samples_arange), wu0[b] * wv0[b])
+
+            gy_tmp = gy[b].reshape(C, -1)
```
ir5: ditto.

tests/chainer_tests/functions_tests/array_tests/test_spatial_transformer_sampler.py
```diff
@@ -0,0 +1,125 @@
...
+        numpy.linspace(-1., 1., num=in_shape[2]),
+        numpy.linspace(-1., 1., num=in_shape[3]))
+    mesh = [numpy.rot90(mesh[0]), numpy.rot90(mesh[1])]
+    grid = numpy.stack(mesh).astype(numpy.float32)
+    grid = numpy.repeat(grid[None], in_shape[0], axis=0)
+    return grid
+
+
+def _rotate_BCHW(x):
+    tmp = []
```
ir5: Renaming this to other name may be better. (e.g., `x_new`, `x_rotated`, ...)

yuyu2172:
@ir5 
Thanks for commenting.
Sorry for the Travis error.
I reflected your comments to the code.

ir5:
Thank you, but there seems to be still some errors... (Maybe PEP8 related things?)

ir5:
I noticed that this is the problem of autopep8, not this PR. https://github.com/pfnet/chainer/pull/2363

ir5:
To solve autopep-1.3.0 issue, we need to merge or rebase current `master` branch. Could you do this?

ir5:
Thank you. LGTM.

chainer/functions/array/spatial_transformer_sampler.py
```diff
@@ -214,23 +214,23 @@ def _backward(self, inputs, grad_outputs):
...
         # --- gx
+        if xp is numpy:
+            scatter_add = numpy.add.at
+        else:
+            scatter_add = xp.scatter_add
         gxs = []
-        samples_arange = xp.arange(out_H * out_W, dtype=numpy.int32)
+        gy = gy.reshape(B, C, -1)
         for b in range(B):
-            dydx = xp.zeros((H, W, out_H * out_W), dtype=gy.dtype)
```
yuyu2172: This consumes large amount of memory even when dealing with ordinary images (e.g. 256 * 256 * 128 * 128 for resizing image from (256, 256) to (128, 128))

yuyu2172:
@ir5 
I found inefficient code in backward pass of cudnn mode.
I am sorry, but could you check the fix?

chainer/functions/array/spatial_transformer_sampler.py
```diff
@@ -0,0 +1,289 @@
...
+            grid_type.ndim == 4,
+            grid_type.shape[1] == 2,
+            x_type.shape[0] == grid_type.shape[0],
+        )
+
+    def forward_cpu(self, inputs):
+        return self._forward(inputs)
+
+    def forward_gpu(self, inputs):
+        if cuda.cudnn_enabled and self.use_cudnn and _cudnn_version >= 5000:
```
beam2d: It's better to reverse the `if` condition for readability as follows.
```
if not (the condition for using cuDNN):
    return self._forward(inputs)
...  # (code using cuDNN)
```

chainer/functions/array/spatial_transformer_sampler.py
```diff
@@ -0,0 +1,289 @@
...
+        grid = grid.reshape(grid.shape[:2] + (-1,))
+
+        u = grid[:, 0]
+        v = grid[:, 1]
+
+        # clip coordinates to [-1, 1]
+        u = u.clip(-1, 1)
+        v = v.clip(-1, 1)
+
+        # rescale coordiantes from [-1, 1] to [0, width or height - 1]
```
beam2d: typo: coordinates

chainer/functions/array/spatial_transformer_sampler.py
```diff
@@ -0,0 +1,289 @@
...
+        ys = []
+        for b in range(B):
+            elem = (w1[b, :, None] * x[b, :, v0[b], u0[b]] +
+                    w2[b, :, None] * x[b, :, v0[b], u1[b]] +
+                    w3[b, :, None] * x[b, :, v1[b], u0[b]] +
+                    w4[b, :, None] * x[b, :, v1[b], u1[b]])
+            elem = elem.reshape(out_H, out_W, C)
+            elem = elem.transpose(2, 0, 1)
+            ys.append(elem)
+        y = xp.concatenate([xp.expand_dims(y, axis=0) for y in ys], axis=0)
```
beam2d: Can we reduce the number of kernel launches by rephrasing the for loop and concatenate into the following code?
```
x_indexed = xp.concatenate([xp.expand_dims(x[b, :, v0[b], u0[b]], axis=0) for b in range(B)], axis=0)
y = w1[:, :, None] * x_indexed
x_indexed = xp.concatenate([xp.expand_dims(x[b, :, v0[b], u1[b]], axis=0) for b in range(B)], axis=0)
y += w2[:, :, None] * x_indexed
...
y = y.reshape(B, out_H, out_W, C).transpose(0, 3, 1, 2)
```
The idea is simple: concatenate x first and then compute the multiply-add. I expect that it slightly improves the performance, but I didn't try it, so if it is not effective (or the code is wrong), please ignore this comment.

chainer/functions/array/spatial_transformer_sampler.py
```diff
@@ -0,0 +1,289 @@
...
+        return y,
+
+    def backward_cpu(self, inputs, grad_outputs):
+        return self._backward(inputs, grad_outputs)
+
+    def backward_gpu(self, inputs, grad_outputs):
+        if cuda.cudnn_enabled and self.use_cudnn and _cudnn_version >= 5000:
```
beam2d: Reverse the if condition (the same comment as forward_gpu)

chainer/functions/array/spatial_transformer_sampler.py
```diff
@@ -0,0 +1,289 @@
...
+            gus.append(gu_elem)
+            gv_elem = (-wu1[b, :, None] * x[b, :, v0[b], u0[b]] -
+                       wu0[b, :, None] * x[b, :, v0[b], u1[b]] +
+                       wu1[b, :, None] * x[b, :, v1[b], u0[b]] +
+                       wu0[b, :, None] * x[b, :, v1[b], u1[b]])
+            gv_elem = gv_elem.reshape(out_H, out_W, C)
+            gv_elem = gv_elem.transpose(2, 0, 1)
+            gvs.append(gv_elem)
+        gu = xp.concatenate([xp.expand_dims(g, axis=0) for g in gus], axis=0)
+        gv = xp.concatenate([xp.expand_dims(g, axis=0) for g in gvs], axis=0)
```
beam2d: See the comment on _forward_gpu. The same rephrasing might be applied to this for-concat.

chainer/functions/array/spatial_transformer_sampler.py
```diff
@@ -0,0 +1,289 @@
...
+    Notatition: here is a notation for dimensionalities.
+
+    - :math:`n` is the batch size.
+    - :math:`c_I` is the number of the input channels.
+    - :math:`h` and :math:`w` are the height and width of the input image,
+      respectively.
+    - :math:`h_O` and :math:`w_O` are the height and width of the output
+      image.
+
+    See detail in a paper: `Spatial Transformer Networks \
```
beam2d: a paper -> the following paper (or the original paper)

chainer/functions/array/spatial_transformer_sampler.py
```diff
@@ -0,0 +1,289 @@
...
+        x (~chainer.Variable):  Input variable of shape :math:`(n, c_I, h, w)`.
+        grid (~chainer.Variable): Coordinate variable of shape
+            :math:`(n, 2, h_O, w_O)`. Each coordinate defines the spatial
+            location in the input where a sampling kernel is applied to get
+            the value at a particular pixel in the output.
+            ``grid[idx, :, i, j]`` corresponds to the coordinate that is used
+            to sample the values for an output pixel at location
+            :math:`(i, j)`.
+
+            In the 2nd dimension, the first coordinate corresponds to the
```
beam2d: 2nd -> second

chainer/functions/array/spatial_transformer_sampler.py
ir5: You are true. This modification looks good to me.

chainer/functions/array/spatial_transformer_sampler.py
yuyu2172: Thanks for a nice idea.

yuyu2172:
Thanks for reviewing. I updated the code

yuyu2172:
Here is a benchmark code for forward pass.

result

original: 13.65(ms)
changed: 13.59(ms)

```python
import numpy as np
from skimage.data import astronaut

import chainer.functions as F
import time
import cupy


img = astronaut()
img_c = img.transpose(2, 0, 1)[None].astype(np.float32)  # (1, 3, H, W)
img_c = np.repeat(img_c, 36, axis=0)
print img_c.shape

mesh = np.meshgrid(
    np.linspace(-1., 1., num=128), np.linspace(-1., 1., num=128))
mesh = [np.rot90(mesh[0]), np.rot90(mesh[1])]
grid = np.stack(mesh)[None]  # (1, 2, h_o, w_o)  
grid = np.repeat(grid, 36, axis=0)

for i in range(5):
    out = F.spatial_transformer_sampler(
        cupy.array(img_c), cupy.array(grid))
    out.data.get()


start = cupy.cuda.Event()
end = cupy.cuda.Event()
start.record()
out = F.spatial_transformer_sampler(
    cupy.array(img_c), cupy.array(grid))
end.record()
end.synchronize()
time = cupy.cuda.get_elapsed_time(start, end)  # milliseconds
print time
```

yuyu2172:
Sorry, I found a mistake in my benchmark. I was using cuDNN mode in the benchmark code I used in the previous comments.

The updated benchmark showed some improvement in the performance of code after reducing kernel calls.
On my laptop with Quadro M2000M, here are the time it took to execute the benchmark.

Before reducing number of kernels (before cf74099):   67.4360641479  (ms)
After reducing number of kernels :  63.8747703552(ms)


```python

import numpy as np
from skimage.data import astronaut

import chainer.functions as F
import time
import cupy


img = astronaut()
img_c = img.transpose(2, 0, 1)[None].astype(np.float32)  # (1, 3, H, W)
img_c = np.repeat(img_c, 36, axis=0)
print img_c.shape   # (36, 3, 512, 512)

mesh = np.meshgrid(
    np.linspace(-1., 1., num=128), np.linspace(-1., 1., num=128))
mesh = [np.rot90(mesh[0]), np.rot90(mesh[1])]
grid = np.stack(mesh)[None]  # (1, 2, h_o, w_o)  
grid = np.repeat(grid, 36, axis=0).astype(np.float32)   # (36, 2, 128, 128)

# dry run
for i in range(5):
    out = F.spatial_transformer_sampler(
        cupy.array(img_c), cupy.array(grid), use_cudnn=False)
    out.data.get()

times = []
for i in range(5):
    start = cupy.cuda.Event()
    end = cupy.cuda.Event()
    start.record()
    out = F.spatial_transformer_sampler(
        cupy.array(img_c), cupy.array(grid), use_cudnn=False)
    end.record()
    end.synchronize()
    times.append(cupy.cuda.get_elapsed_time(start, end))  # milliseconds)
print np.mean(times)

```

mitmul:
@beam2d Could you approve this PR? I confirmed that this PR has already passed all Jenkis tests: https://ci-chainer.pfidev.jp/job/chainer-sakura/1326/

beam2d:
Sorry for being late. Approved.

PR #2274, opened 2017-02-15
Add cupy.meshgrid

yuyu2172:
This PR adds cupy.meshgrid.
*sparse* option is not supported in this PR.

### DIFF
--- a/cupy/creation/ranges.py
+++ b/cupy/creation/ranges.py
@@ -132,9 +132,10 @@ def meshgrid(*xi, **kwargs):
     Given one-dimensional coordinate arrays x1, x2, ..., xn, this function
     makes N-D grids.
 
-    For ndarrays x1, x2, ..., xn with lengths ``Ni = len(xi)``, return
-    ``(N1, N2, N3, ..., Nn)`` shaped arrays if indexing='ij' or
-    ``(N2, N1, N3, ..., Nn)`` shaped arrays if indexing='xy'.
+    For one-dimensional arrays x1, x2, ..., xn with lengths ``Ni = len(xi)``,
+    this function returns ``(N1, N2, N3, ..., Nn)`` shaped arrays
+    if indexing='ij' or ``(N2, N1, N3, ..., Nn)`` shaped arrays
+    if indexing='xy'.
 
     Unlike NumPy, CuPy currently only supports 1-D arrays as inputs.
     Also, CuPy does not support ``sparse`` option yet.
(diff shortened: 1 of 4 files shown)


cupy/creation/ranges.py
```diff
@@ -126,7 +126,66 @@ def logspace(start, stop, num=50, endpoint=True, base=10.0, dtype=None):
...
 
 
-# TODO(okuta): Implement meshgrid
+def meshgrid(*xi, **kwargs):
+    """Return coordinate matrices from coordinate vectors.
+
+    Given one-dimensional coordinate arrays x1, x2, ..., xn, this function
+    makes N-D grids.
+
+    For ndarrays x1, x2, ..., xn with lengths ``Ni = len(xi)``, return
```
ir5: For consistency with the previous sentence, describing "For one-dimensional arrays ..." looks better to me. Further, writing "this function returns ..." seems better than just "return".

cupy/creation/ranges.py
```diff
@@ -126,7 +126,66 @@ def logspace(start, stop, num=50, endpoint=True, base=10.0, dtype=None):
...
+
+    .. seealso:: :func:`numpy.meshgrid`
+
+    """
+
+    indexing = kwargs.pop('indexing', 'xy')
+    copy = bool(kwargs.pop('copy', True))
+    if indexing not in ['xy', 'ij']:
+        raise ValueError('Valid values for `indexing` are \'xy\' and \'ij\'.')
+
```
ir5: This may be not so important, but how about check if `kwargs` contains unnecessary arguments as is in numpy implementation? https://github.com/numpy/numpy/blob/v1.12.0/numpy/lib/function_base.py#L4504

yuyu2172:
Thanks for commenting.
I fixed what you pointed out.

cupy/creation/ranges.py
```diff
@@ -126,7 +126,71 @@ def logspace(start, stop, num=50, endpoint=True, base=10.0, dtype=None):
...
+    .. seealso:: :func:`numpy.meshgrid`
+
+    """
+
+    indexing = kwargs.pop('indexing', 'xy')
+    copy = bool(kwargs.pop('copy', True))
+    if kwargs:
+        raise TypeError(
+            'meshgrid() got an unexpected keyword argument \'{}\''.format(
+                        list(kwargs)[0]))
```
ir5: Oh, there is still a minor error. According to the CI result, this line violates flake8 with `--config .flake8.cython` option.

ir5: Please note that we need to install both flake8 and hacking for checking codes. The installation of hacking changes the behavior of flake8.

yuyu2172: Thanks! I did not know that.

beam2d:
LGTM

beam2d:
Thanks!

PR #2278, opened 2017-02-15
Add Tox21 dataset

delta2323:
This PR adds a dataset loader of [Tox21](https://tripod.nih.gov/tox21/challenge/), which is one of the most widely used dataset in bioinformatics. Tox21 is a dataset of Toxicology provided by [The Toxicology in the 21st Century (Tox21)](https://ncats.nih.gov/tox21). It consists of the chemical information of compounds and their assessment of toxicity.

As raw files is [SDF](https://en.wikipedia.org/wiki/Chemical_table_file#SDF) format, we extract feature from them with [RDKit](https://github.com/rdkit/rdkit). So users have to install it manually.

PR #2279, opened 2017-02-15
[WIP] Add multitask example with Tox21

delta2323:
This PR adds an example of does multitask learning with Chainer. As this PR uses Tox21 dataset which is added by another PR #2278, please review it before.


delta2323:
close this PR to resend from the forked repository

PR #2287, opened 2017-02-18
improve forget bias to 1

icoxfog417:
The default forget bias will be better to set 1.

Let me introduce the following paper.

[An Empirical Exploration of Recurrent Network Architectures](http://jmlr.org/proceedings/papers/v37/jozefowicz15.pdf)

> Thus we recommend adding a bias of 1 to the forget gate of every LSTM in every application; it is easy to do often results in better performance on our tasks. This adjustment is the simple improvement over the LSTM that we set out to discover.

And TensorFlow's forget bias is set to 1.

[BasicLSTMCell implementation](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/rnn/python/ops/core_rnn_cell_impl.py#L138)

### DIFF
--- a/chainer/links/connection/lstm.py
+++ b/chainer/links/connection/lstm.py
@@ -16,7 +16,7 @@ class LSTMBase(link.Chain):
 
     def __init__(self, in_size, out_size,
                  lateral_init=None, upward_init=None,
-                 bias_init=0, forget_bias_init=0):
+                 bias_init=0, forget_bias_init=1):
         super(LSTMBase, self).__init__(
             upward=linear.Linear(in_size, 4 * out_size, initialW=0),
             lateral=linear.Linear(out_size, 4 * out_size,


delta2323:
Thank you for sending the PR and sorry for our late response. Although the default value of forget bias vector is not documented, I think it is better to merge it to v2 to keep backward compatibility because it could affect users' code who uses this link with the default value and we expect many users do so.

icoxfog417:
I think that to merge this PR in v2 makes sense.

It's the truth that the change of default parameter affects the user's reproducibility. And according to the Chainer compatibility policy, minor/revision release must keep the backward compatibility.

So it's ok to wait the comming of v2!

delta2323:
Thank you. Then, could you change the destination branch to `_v2` ?

icoxfog417:
I changed the target branch to `_v2`, but then PR includes some commits. Should I recreate the PR?

delta2323:
> Should I recreate the PR?

Thank you very much. Then could you do that?

icoxfog417:
Sure. I close this pull request and transported to #2425

PR #2288, opened 2017-02-18
PlotReport takes non-negligible portion of time in small networks

soramichi:
When a figure is saved with `bbox_inches='tight'`, `matplotlib.backend_bases.print_method()` is called twice per each `savefig()`, as a comment in the matplotlib source says: https://github.com/matplotlib/matplotlib/blob/master/lib/matplotlib/backend_bases.py#L2156_L2166

Due to this, when the network is small (like the MNIST example) and the latency of storage IO is high, `savefig()` takes non-negligible portion of the total elapsed time. For example, when the MNIST example is executed in a g2.2xlarge AWS instance with a magnetic device used as the storage, **`write_png()` function (the main part of `print_method()`) takes 5.9 seconds out of 134.7 seconds of total execution time**.

Given this, this PR deletes `bbox_inches='tight'` and reduces the number of calls to `print_method()` to only once per each `savefig()`, resulting in 2.8 seconds used for `write_png()` in the aforementioned environment (g2.2xlarge, magnetic device used as the storage). Because PlotReport uses the default font size and shape, removing `bbox_inches='tight'` does not affect the output image.

### DIFF
--- a/chainer/training/extensions/plot_report.py
+++ b/chainer/training/extensions/plot_report.py
@@ -140,7 +140,7 @@ class PlotReport(extension.Extension):
                     self._postprocess(f, a, summary)
                 l = a.legend(bbox_to_anchor=(1.05, 1), loc=2, borderaxespad=0.)
                 f.savefig(path.join(trainer.out, self._file_name),
-                          bbox_extra_artists=(l,), bbox_inches='tight')
+                          bbox_extra_artists=(l,))
 
             plot.close()
             self._init_summary()


mitmul:
@soramichi Thank you for sending this PR. Now I understand the performance problem due to the frequent call of `savefig()`. However, I found that if the legend is drawn outside of the plotting area as the current implementation does, omitting `bbox_inches='tight'` leads incomplete display of legend box like:

loss
(in the rightmost part, legend box can't appear with the whole view)

I don't know much about matplotlib, so I'd appreciate it if you could suggest an alternative way to avoid both performance reduction and incorrect display of legend box.

soramichi:
@mitmul Hi, thanks for trying it out. I retried plotting without `bbox_inches='tight'`, and confirmed that the legend does not fit in the figure as you pointed out. I might have checked a wrong figure when I opened this PR.

Speaking about alternative ways, calling `f.tight_layout()` before `f.save()` yields a proper figure but it incurs the same amount of overhead as `bbox_inches='tight'`. Explicitly specifying the size for the  `bbox_inches` option (instead of using 'tight') might reduce the overhead, but I don't think it's feasible given that PlotReport is a library but not a one-time code.

So given these I decided to close this PR. My deep apologies to taking your extra time by my mistake.

Issue #2252, opened 2017-02-09
Training utilities and examples that are independent of trainer


bkvogel:
The trainer feature offers useful abstractions that can often simplify the process of implementing a model in Chainer. However, in some cases users might find it more convenient to write their own training loop manually instead of using trainer. 

For some models such as certain types of RNNs, it might sometimes be more convenient to write a custom training loop. Trainer comes with a learning curve and so new users could start being productive with Chainer quicker if they can implement models without trainer, at least in the beginning. Related to this, new users might also be able to more quickly understand the Chainer included examples if we could also provide an implementation of many of the examples without using trainer (of course we would also want to still include the trainer versions).

That is, trainer should be considered as an option that can often be useful, but that is not required to implement a training/testing loop in Chainer. Unfortunately, there is currently one drawback to not using trainer: without it, the user cannot access the many trainer extensions such as ProgressBar, LogReport, etc.. To resolve this inconvenience, I propose to write some training utilities that provide similar functionality to the trainer extensions but that can also be easily used in a custom training loop. I also propose to write additional examples (based on some of the current examples) that use a custom training loop with these utilities. I am currently experimenting with such utilities and would like to submit a corresponding WIP PR soon, especially if others think such utilities could be useful. 

I expect that this feature should not break backward compatibility, but I might like to use some of the other v2 features in the utilities and examples and so am thinking to label this as a v2 feature.

Crissman:
Part of future roadmap.

PR #2301, opened 2017-02-22
[WIP] Training utilities and examples using a custom training loop

bkvogel:
Add some training utilities for #2252. This PR currently contains a timer utility and a progress utility that can optionally display a progress bar (borrowing code with some duplication from the existing `ProgressBar`). Other utilities can be added later as separate PRs.

Additional examples of using custom training loops have been added for MNIST, CIFAR, and ptb examples.  These examples now contain both a Trainer version and a "custom training loop" version. These new examples also show how to use the new v2 `chainer.config.train` flag to disable train mode during model evaluation.


delta2323:
"No compat" label is attached to this PR. But does it really break compatibility?

bkvogel:
You are right, "no compat" was not needed so I removed it. I was initially considering to break compatibility but then decided against it.

delta2323:
OK, thanks.

bkvogel:
Since this did not make it into v2, I will close, update the PR and resubmit from a forked repo.

PR #2302, opened 2017-02-22
Batch renormalization

tkerola:
This PR adds a batch renormalization function and link, as described in the following paper: https://arxiv.org/abs/1702.03275

Training works, and gradient checks pass, so the implementation is probably ok, but there are some issues I would like to discuss before merging this PR:
* There is no cuDNN support for batch renormalization, as the current cuDNN function for batch normalization does not allow applying the correction parameters `r` and `d` before `gamma` and `beta`.
* This is an extension of `BatchNormalization`, but it has additional overhead, so I put it in a separate file. The downside is that there is some code that is both appearing in `BatchNormalization` and `BatchRenormalization`, which goes against the DRY principle.

Since `BatchNormalization` is very popular, I suggest to keep it separate from `BatchRenormalization` in order to keep down the overhead (perhaps they could be merged later in a separate PR if deemed appropriate). We could also refactor both `BatchNormalization` and `BatchRenormalization` to share the similar code parts, but that would require more careful checking of this PR.

Please let me know what you think.
Once these issues are solved, we can brush up the documentation etc. for this PR.


delta2323:
Sorry for our late response. I agree with you in that we separate `BatchRenormalization` from `BatchNormalization` from the same reasons you wrote (overhead and popularity of `BatchNormalization`).

As for cuDNN support, I think it cannot be helped we cannot support cuDNN as its API is not expect to implement Batch Renormalization. Even if there could be some hack to realize it with cuDNN, we can do it as another PR.

delta2323:
Hi, can we start to review this PR?

tkerola:
Hi, yes I think that it is ok. I just realized that since I removed cuDNN support for BatchRenormalization, some cuDNN checks still exist in this PR. So I will try to remove them soon.

delta2323:
Could you fix flake8 and autopep8 violation?

delta2323:
I expect we can merge this feature after v2 release. So, I propose that we change the destination of this PR to `_v2`. I also want to know how much difficult to fix the code v2-compatible. Of course, I will help you to fix this PR.

tkerola:
Hi, I made the code pass flake8 and autopep8.
I am not that familiar with v2 yet, so I would appreciate if you could help me, or point me to some document that says how to port the code to v2.

delta2323:
First of all, could you change the destination from the master branch to `_v2`? Also, I will add the PR to the internal Jenkins CI to find out where we have to change.

We are preparing how to update the code to work with Chainer v2, but is not completed yet (acutually, I am thinking to update the document through this review).

chainer/functions/normalization/batch_renormalization.py
```diff
@@ -0,0 +1,236 @@
...
+                d = xp.clip((mean - self.running_mean) / running_sigma,
+                            -self.dmax, self.dmax)
+            else:
+                r = xp.ones_like(gamma)
+                d = xp.zeros_like(gamma)
+
+            if self.keep_r_d_fixed:
+                # Hack for making gradient check treat r and d as true
+                # constants with respect to the batch.
+                print("Warning: self.keep_r_d_fixed is True, which is needed "
```
delta2323: Use `warnings.warn`.

delta2323:
One thing I noticed is that `L.BatchRenormalization.__call__` takes `test` argument and `F.BetchRenormalization.__init__` takes `train` argument. In v2, training/test modes are managed by `chainer.configuration.train` and function that behaves differently should reference the value. See [BatchNormalization for v2](https://github.com/pfnet/chainer/blob/_v2/chainer/links/normalization/batch_normalization.py#L130).

tkerola:
Hi, I fixed the PR according to your comments. It is now rebased onto `_v2` and passes tests.

tkerola:
Hmm, Travis CI fails due to a type check in `chainer_tests/functions_tests/normalization_tests/test_batch_renormalization.py`: ` dtype('float32') != <class 'numpy.float16'>`. But the test passes on my machine. Strange, I don't understand why.

chainer/functions/normalization/batch_renormalization.py
```diff
@@ -0,0 +1,232 @@
...
+                mean_type.shape == gamma_type.shape,
+                var_type.dtype == x_type.dtype,
+                var_type.shape == gamma_type.shape,
+            )
+
+    def forward(self, inputs):
+        xp = cuda.get_array_module(*inputs)
+        x, gamma, beta = inputs[:3]
+        if configuration.config.train:
+            if self.running_mean is None:
```
delta2323: It seems the only case `self.running_mean` is `None` is that it is initialized by `None` in `__init__` and the `forward` is called for the first time. So how about initializing `self.running_mean` with zero in `__init__` if `running_mean` is `None`?

chainer/functions/normalization/batch_renormalization.py
```diff
@@ -0,0 +1,232 @@
...
+                print("Set running_mean,var to zero")
```
delta2323: I do not think it is good idea that this function outputs this message because user could use stdout for other purpose and the message could contaminate the stdout.

chainer/functions/normalization/batch_renormalization.py
```diff
@@ -0,0 +1,232 @@
...
+                self.running_mean = xp.array(self.running_mean)
+                self.running_var = xp.array(self.running_var)
+        elif len(inputs) == 5:
+            self.fixed_mean = inputs[3]
+            self.fixed_var = inputs[4]
+
+        head_ndim = gamma.ndim + 1
+        expander = (None, Ellipsis) + (None,) * (x.ndim - head_ndim)
+
+        # NOTE(tommi): Removed cuDNN support since it does not support
```
delta2323: It is not appropriate to say "removed" because users do not know the time cuDNN implementation existed. Simply "Do not support cuDNN implementation" is enough

chainer/functions/normalization/batch_renormalization.py
```diff
@@ -0,0 +1,232 @@
...
+        # batch renormalization
+        if configuration.config.train:
+            axis = (0,) + tuple(range(head_ndim, x.ndim))
+            mean = x.mean(axis=axis)
+            var = x.var(axis=axis)
```
delta2323: I think `var = x.var(axis=axis) + self.eps` is simpler.

chainer/functions/normalization/batch_renormalization.py
```diff
@@ -0,0 +1,232 @@
...
+            var += self.eps
+        else:
+            mean = self.fixed_mean
+            var = self.fixed_var + self.eps
```
delta2323: When `configuration.config.train` is `False` and `len(inputs) == 3`, `fixed_mean` is not defined in the first call of `forward`. If we expect the length of `inputs` is 3 in the test phase, we should check it as an input validation of the `forward` method.  Maybe it is related to #2515.

chainer/functions/normalization/batch_renormalization.py
```diff
@@ -0,0 +1,232 @@
...
+        self.std = xp.sqrt(var, dtype=var.dtype)
+        if not self.keep_r_d_fixed or self.r is None:
+            if configuration.config.train:
+                running_sigma = xp.sqrt(self.running_var + self.eps)
+                r = xp.clip(self.std / running_sigma,
```
delta2323: Why not assign the value directly to `self.r`?

chainer/functions/normalization/batch_renormalization.py
```diff
@@ -0,0 +1,232 @@
...
+                            1.0 / self.rmax, self.rmax)
+                d = xp.clip((mean - self.running_mean) / running_sigma,
```
delta2323: ditto (why not assign the value directly to `self.d`?)

chainer/functions/normalization/batch_renormalization.py
```diff
@@ -0,0 +1,232 @@
...
+                            -self.dmax, self.dmax)
+            else:
+                r = xp.ones_like(gamma)
```
delta2323: ditto

chainer/functions/normalization/batch_renormalization.py
```diff
@@ -0,0 +1,232 @@
...
+                d = xp.zeros_like(gamma)
```
delta2323: ditto

chainer/functions/normalization/batch_renormalization.py
```diff
@@ -0,0 +1,232 @@
...
+            x_hat = _xhat(x, mean, std, expander)
+            ggamma = (gy * x_hat).sum(axis=axis)
+            gmean = -gs * gbeta
+            gvar = -0.5 * gamma / var * ggamma
+            gx = gs[expander] * gy
+            return gx, ggamma, gbeta, gmean, gvar
+
+        # Note: If length of inputs is not 5, we must be in train mode.
+        assert configuration.config.train
+        # NOTE(tommi): Removed cuDNN support since it does not support
```
delta2323: ditto

chainer/functions/normalization/batch_renormalization.py
```diff
@@ -0,0 +1,232 @@
...
+        if configuration.config.train:
+            if self.running_mean is None:
+                print("Set running_mean,var to zero")
+                self.running_mean = xp.zeros_like(gamma)
+                self.running_var = xp.zeros_like(gamma)
+            else:
+                self.running_mean = xp.array(self.running_mean)
+                self.running_var = xp.array(self.running_var)
+        elif len(inputs) == 5:
+            self.fixed_mean = inputs[3]
```
delta2323: I do not think `fixed_mean` need not to be an attribute as it is used in `forward` only.

chainer/functions/normalization/batch_renormalization.py
```diff
@@ -0,0 +1,232 @@
...
+            self.fixed_var = inputs[4]
```
delta2323: ditto (`fixed_var` need not to be an attribute).

chainer/functions/normalization/batch_renormalization.py
tkerola: I am not sure if this is possible to do in `__init__`, since we would need to know the shape of `gamma` in order to initialize `self.running_mean` with zeros. I think this information is not available until `forward` is called.

chainer/functions/normalization/batch_renormalization.py
tkerola: Good find. Will remove.

chainer/functions/normalization/batch_renormalization.py
tkerola: Good find. This should ideally be fixed for `BatchNormalizationFunction` as well.

chainer/functions/normalization/batch_renormalization.py
tkerola: Fixed

chainer/functions/normalization/batch_renormalization.py
tkerola: Fixed

chainer/functions/normalization/batch_renormalization.py
tkerola: Fixed. Should ideally be fixed for `BatchNormalizationFunction` as well.

chainer/functions/normalization/batch_renormalization.py
tkerola: Fixed.

chainer/functions/normalization/batch_renormalization.py
tkerola: Fixed

chainer/functions/normalization/batch_renormalization.py
tkerola: Fixed

chainer/functions/normalization/batch_renormalization.py
tkerola: Fixed

chainer/functions/normalization/batch_renormalization.py
tkerola: Fixed

chainer/functions/normalization/batch_renormalization.py
tkerola: Fixed

tkerola:
I fixed the PR according to your comments.

chainer/functions/normalization/batch_renormalization.py
delta2323: I see. Then it is OK to initialize `running_mean` lazily.

delta2323:
Thanks. Your change looks good. I will go on to `L.BatchRenormalization` later.

delta2323:
Sorry for putting aside this PR. As _v2 is merged to the master branch, could you change the destination branch to `master`?

tkerola:
Hi, I changed the destination to master. (Didn't check if tests pass yet though.)

delta2323:
I think I understand how `keep_r_d_fixed` works. However it would be better not to provide users with an argument that is not unnecessary for them if possible. I thought we can remove the option by devising the function we provide to `check_backward` as follows. Could you consider its feasibility?

```
def check_backward(...):
    _f = BatchRenormalizationFunction(...)
    r = None

    def f(x):
        if r is not None:
            _f.r = r
        ret = _f(x)
        if r is None:
            r = f.r
        return ret

    check_backward(f, ...)
```

tkerola:
I have been thinking about if it is possible to remove `keep_r_d_fixed`, but I am not sure if it is possible. I agree with you that this argument should not be provided to the users though.
The reason why this argument is needed is because `BatchRenormalization` does not calculate the true gradient, but rather clips the gradient using `rmax` and `dmax`.
I considered your suggested idea, but I don't think this will not work as `_f.r` will become updated during each pass of finite differences during the gradient check, which will cause the gradient check to fail as we need to consider `r` and `d` as constants when calculating the gradients, while these are actually calculated from the forward pass of the function.
If we remove `keep_r_d_fixed`, then the condition `self.r is None` needs to be remove here as well:
https://github.com/chainer/chainer/pull/2302/files#diff-6006b6aaf49c53da5f5554c46346c285R102
If we don't do this, the function will not work properly when it is used directly as a `Function` (it will work when being used as a `Link` though, since the `Link` re-creates the `Function` during each forward pass).
But if we remove it, then the gradient check will fail, as `r` and `d` will no longer be considered as constants wrt. to the first forward pass during finite differences. Please let me know if you have any idea of how we can resolve this.

delta2323:
Then, how about instantiating `BatchRenormalizationFunction` at every call of `f`, as we do in the link?

delta2323:
If we add options to "hack" `BatchRenormalizationFunction`, I would suggest `freeze_running_statistics` that updates neither of `self.running_average` nor `self.r`.

There are two reasons for this. First, that could be useful for not only unit tests but also users who just want to do forward propagation.

The second reason is consistency of `self.r` and `self.running_average`. As `r` is deterministically computed from the running average and the average of this batch, there should be some relation between the two. If we keep only `self.r` fixed while updating `self.running_average`, there will be inconsistency between the two.

tkerola:
> Then, how about instantiating BatchRenormalizationFunction at every call of f, as we do in the link?
I think we should not do this, since a user could create a `BatchRenormalizationFunction` object and call it repeatedly, which would then cause the function to not behave correctly.

Unrelated to the above issue, I added `freeze_running_statistics` and made `r` and `running_mean` consistent, as you suggested.

delta2323:
`freeze_running_statistics` sounds better as it is more explanatory.

delta2323:
Some travis tests failed, could you fix that?

tkerola:
Hmm, the tests pass on my computer. I am not sure why it fails on TravisCI.

delta2323:
jenkins, test this please.

delta2323:
@tkerola OK, I am running Jenkins to see if current code works.

delta2323:
GPU test passed. So the error seems specific to the Travis environment.

tkerola:
How can this be debugged? I am not familiar with Travis.

delta2323:
I am not sure it is related but I found NumPy version is 1.9, which is relatively old, although we declare its support.

delta2323:
I could reproduce the error with NumPy 1.9.0.

tkerola:
I tracked down why the tests fail with numpy 1.9.0.
```
# numpy 1.13.0
In [1]: import numpy as np
In [2]: np.__version__
Out[2]: '1.13.0'
In [3]: np.sqrt(np.float16(1) + np.float16(2)).dtype
Out[3]: dtype('float16')

# numpy 1.9.0
In [1]: import numpy as np
In [2]: np.__version__
Out[2]: '1.9.0'
In [3]: np.sqrt(np.float16(1) + np.float16(2)).dtype
Out[3]: dtype('float32')
```
Seems to be some bug in numpy 1.9.0. I put in a workaround.

delta2323:
jenkins, test this please

delta2323:
Jenkins didn't seem to run correctly. I reran the test.

tkerola:
It seems like the tests are passing now.

delta2323:
jenkins, test this please.

yuyu2172:
From the paper, parameter `rmax` and `dmax` should be gradually increased during training.
Do you have any recommended way to do scheduling?

By myself, I scheduled these parameters using `LinearShift` like below.
```python
# rn is BatchRenormalization Link.
trainer.extend(
    extensions.LinearShift('rmax', (1, 3), (5000, 40000), rn),
    trigger=(10, 'iteration'))
```

tkerola:
The `LinearShift` solution looks nice!

I originally used a custom trainer extension, but it doesn't look as nice as yours:
```
@chainer.training.make_extension()
def batch_renorm_schedule(trainer):
    # For the first 5000 iterations, rmax = 1, dmax = 0 (simply batch norm)
    # Gradually relaxed after 40000 iterations to rmax = 3
    # Gradually relaxed after 25000 iterations to dmax = 5

    model = trainer.updater.get_optimizer('main').target
    current_iteration = trainer.updater.iteration

    stage0 = 5000
    stage1_rmax = 40000
    stage1_dmax = 25000
    rmax_start = 1
    rmax_final = 3
    dmax_start = 0
    dmax_final = 5

    if current_iteration < stage0:
        return
    if current_iteration > max(stage1_rmax, stage1_dmax):
        return

    from chainer import link
    from myssd.nets.batch_renormalization import BatchRenormalization

    # Need to parse the whole model, find BRN layers, and update params.
    def parse_chain(src):
        assert isinstance(src, link.Chain)
        for child in src.children():
            if isinstance(child, link.Chain):
                parse_chain(child)
            if isinstance(child, BatchRenormalization):
                if current_iteration < stage1_rmax:
                    child.rmax = rmax_start + (rmax_final - rmax_start) * (current_iteration - stage0) / float(stage1_rmax - stage0)
                    if current_iteration % 1000 == 0:
                        print('[{}] Updated rmax={} for {}'.format(current_iteration, child.rmax, child.name))
                else:
                    child.rmax = rmax_final
                if current_iteration < stage1_dmax:
                    child.dmax = dmax_start + (dmax_final - dmax_start) * (current_iteration - stage0) / float(stage1_dmax - stage0)
                    if current_iteration % 1000 == 0:
                        print('[{}] Updated dmax={} for {}'.format(current_iteration, child.dmax, child.name))
                else:
                    child.dmax = dmax_final

    parse_chain(model)

```

tkerola:
The `LinearShift` solution looks nice!

I originally used a custom trainer extension. But it doesn't look as nice as yours:
```
@chainer.training.make_extension()
def batch_renorm_schedule(trainer):
    # For the first 5000 iterations, rmax = 1, dmax = 0 (simply batch norm)
    # Gradually relaxed after 40000 iterations to rmax = 3
    # Gradually relaxed after 25000 iterations to dmax = 5

    model = trainer.updater.get_optimizer('main').target
    current_iteration = trainer.updater.iteration

    stage0 = 5000
    stage1_rmax = 40000
    stage1_dmax = 25000
    rmax_start = 1
    rmax_final = 3
    dmax_start = 0
    dmax_final = 5

    if current_iteration < stage0:
        return
    if current_iteration > max(stage1_rmax, stage1_dmax):
        return

    from chainer import link
    from chainer.functions.normalization.batch_renormalization import BatchRenormalization

    # Need to parse the whole model, find BRN layers, and update params.
    def parse_chain(src):
        assert isinstance(src, link.Chain)
        for child in src.children():
            if isinstance(child, link.Chain):
                parse_chain(child)
            if isinstance(child, BatchRenormalization):
                if current_iteration < stage1_rmax:
                    child.rmax = rmax_start + (rmax_final - rmax_start) * (current_iteration - stage0) / float(stage1_rmax - stage0)
                    if current_iteration % 1000 == 0:
                        print('[{}] Updated rmax={} for {}'.format(current_iteration, child.rmax, child.name))
                else:
                    child.rmax = rmax_final
                if current_iteration < stage1_dmax:
                    child.dmax = dmax_start + (dmax_final - dmax_start) * (current_iteration - stage0) / float(stage1_dmax - stage0)
                    if current_iteration % 1000 == 0:
                        print('[{}] Updated dmax={} for {}'.format(current_iteration, child.dmax, child.name))
                else:
                    child.dmax = dmax_final

    parse_chain(model)

```

yuyu2172:
One disadvantage with `LinearShift` is that you need to create N extension objects (`N=number of RNs`).
Thanks for sharing your code :+1:

tkerola:
@delta2323 Github says that you are requesting changes, but this seems to be for an older commit. I believe these have been fixed. Please let me know if there are any issues left with this PR. Thank you.

delta2323:
I confirm that all tests passed. Thank you for your contribution and sorry for making you wait.

PR #2316, opened 2017-02-23
Update plot_report.py

mitmul:
remove self


mitmul:
I close this because it's included here: #2317

PR #2319, opened 2017-02-23
Add persistent() function to Link, Chain and ChainList class

takerum:
Add persistent() function to Link, Chain and ChainList,  in order to access the persistent parameters from outside of the instance of each class.


muupan:
Could you add tests cases to `tests/chainer_tests/test_link.py`?

PR #2327, opened 2017-02-25
fix epoch_detail

Hakuyume:
`epoch_detail` takes wrong value if the iterator is not repeated.
Here is an example.
```
dataset = [1, 2, 3]
it = chainer.iterators.SerialIterator(dataset, 2, repeat=False)
it.next() # [3, 1]
it.next() # [2]
it.epoch # 1
it.epoch_detail # 2.0
```
I think `epoch_detail` should be 1.0 in this case.


mattya:
Thanks for fixing and adding tests.
LGTM.

PR #2332, opened 2017-02-26
Implement cupy.msort.

takagi:
This PR implements `cupy.msort`. Please merge #2057 first.

`numpy.msort(a)` is equivalent to `numpy.sort(a, axis=0)`. For implementation reasons, `cupy.sort` currently supports only sorting an array with its rank of one, so `cupy.msort(a)` would be actually equivalent to `cupy.sort(a)` for now.


takagi:
This PR is reposted to cupy/cupy#150 after CuPy separation from Chainer, so I close this.

PR #2333, opened 2017-02-26
L.Linear is preferable to F.Linear in LSTM document.

dsanno:
I fixed the LSTM document.
`L.Linear` is preferable to `F.Linear` because it has parameters (weight and bias).


okuta:
Thanks!
LGTM.

PR #2337, opened 2017-02-27
Add ConvNet tutorials

mitmul:
- [x] Convolutional Network for Visual Recognition Tasks
  - [x] How to write a convolutional network that has a large number of components (ResNet)


docs/source/tutorial/convnet.rst
```diff
@@ -0,0 +1,428 @@
...
+.. doctest::
+
+    model = L.Classifier(LeNet5())
+
+    # Foward & Loss calculation
+    loss = model(x, t)
+
+This class takes a model object as an iput argument and registers it to
+``predictor`` property as a trained parameter. Then calling the returned object
+as a function like the above code with feeding ``x`` and ``t`` as the input
```
bkvogel: As shown above, the returned object can then be called like a function in which we pass `x`` and ``t`` as the input arguments and the resulting loss value (which we recall is a `Variable`) is returned.

docs/source/tutorial/convnet.rst
```diff
@@ -0,0 +1,428 @@
...
+arguments, yield a loss value calculated from those two input variables.
+
+See the detailed implementation of :class:`~chainer.links.Classifier` from
+here: :class:`chainer.links.Classifier` and check the implementation by looking
+at the source.
+
+Chainer is flexible, so we can write our original network in many different
+ways. It would give users an intuitive way to design brand-new and complex
```
bkvogel: From the above examples, we can see that Chainer provides the flexibility to write our original network in many different ways. Such flexibility intends to make it intuitive for users to design new and complex models.

docs/source/tutorial/convnet.rst
```diff
@@ -0,0 +1,428 @@
...
+models.
+
+VGG16
+'''''
+
+Next, let's write more large models like VGG16 [Simonyan14]_ in Chainer.
+When you write a large network consisted of several building block
+networks, :class:`~chainer.ChainList` is useful. First, let's see how to write
+a VGG16 model:
```
bkvogel: Next, let's write some larger models in Chainer. When you write a large network consisting of several building block networks, :class:`~chainer.ChainList` is useful. First, let's see how to write
 +a VGG16 [Simonyan14] model

docs/source/tutorial/convnet.rst
```diff
@@ -0,0 +1,428 @@
...
+            h = F.max_pooling_2d(h, 2, 2)
+            if self.fc:
+                h = F.dropout(F.relu(self.fc4(h)), train=train)
+                h = F.dropout(F.relu(self.fc5(h)), train=train)
+                h = self.fc6(h)
+            return h
+
+That's it. VGG16 is a model which won the 1st place in
+`classification + localization task at ILSVRC 2014 <http://www.image-net.org/challenges/LSVRC/2014/results#clsloc>`_,
+and since then, became one of standard models for many different tasks as a
```
bkvogel: and since then, has become one of the standard models for many different tasks

docs/source/tutorial/convnet.rst
```diff
@@ -0,0 +1,428 @@
...
+pre-trained model. This has 16-layers, so it's called "VGG-16", but we can
+write this model without writing all layers independently. Because this model
+is consisted of several building blocks that have the same architecture,
+so that we can build the whole network by re-using the building block
+definition. Each part of the network is consisted of 2 or 3 convolutional
```
bkvogel: Since this model consists of several building blocks that have the same architecture, we can build the whole network by re-using the building block definition.

docs/source/tutorial/convnet.rst
```diff
@@ -0,0 +1,428 @@
...
+them, and :meth:`~chainer.functions.max_pooling_2d` operations. This block
+is written as :class:`VGGBlock` in the above example code. And the whole
+network just calls this block one by one in sequential manner.
+
+ResNet152
+'''''''''
+
+How about ResNet? ResNet [He16]_ came in the next year's ILSVRC. It is way deeper
+model than VGG16, so the depth is up to 152 layers. This sounds super laboring
+to build, but it can be implemented in almost same manner as VGG16. In the
```
bkvogel: How about ResNet? ResNet [He16]_ came in the following year's ILSVRC. It is a much deeper model than VGG16, having up to 152 layers. This sounds super laborious to build, but it can be implemented in almost same manner as VGG16.

docs/source/tutorial/convnet.rst
```diff
@@ -0,0 +1,428 @@
...
+            h = F.relu(self.bn_b(self.conv3x3b(x), test=not train))
+            h = self.bn_c(self.conv1x1c(x), test=not train)
+            if self.proj:
+                x = self.bn_r(self.conv1x1r(x), test=not train)
+            return F.relu(h + x)
+
+In :class:`BottleNeck` class, it switches whther ``conv1x1r`` to fit the
+channel dimension of the input ``x`` and the output ``h`` and ``bn_r`` for it
+should be added or not with respect to :attr:`~BottleNeck.proj` attribute.
+Writing the building block in this way gains the re-usability of a class a lot.
```
bkvogel: In the :class:`BottleNeck` class, depending on the value of the ``proj`` argument supplied to the initializer, it will conditionally also compute the batch normalization of a ``conv1x1r`` layer before the final ReLU layer. Writing the building block in this way improves the re-usability of a class.

docs/source/tutorial/convnet.rst
```diff
@@ -0,0 +1,428 @@
...
+It switches not only the behavior in :meth:`__class__` by flags but also the
+parameter registration. In this case, when :attr:`proj` is ``False``, the
+:class:`BottleNeck` doesn't have `conv1x1r` and `bn_r` layers, so the memory
+usage would be efficient compared to the case when it registers both anyway and
+just ignore them if :attr:`proj` is ``False``.
+
+Using nesting :class:`~chainer.Chain` s and :class:`~chainer.ChainList` for
```
bkvogel: Using nested

docs/source/tutorial/convnet.rst
```diff
@@ -0,0 +1,428 @@
...
+sequential part enables us to write complex and very deep models easily.
+
+Use Pre-trained Models
+''''''''''''''''''''''
+
+Various ways to write your models is described above. And VGG16 and ResNet are
+very useful as general feature extractor for many different tasks from image
+classification. So Chainer provides you the pre-trained models of VGG16 and
+ResNet-50/101/152 models with a simple API. You can use those models in this
+way:
```
bkvogel: Various ways to write your models were described above. It turns out that VGG16 and ResNet are very useful as general feature extractors for many different image classification tasks. So, Chainer provides you with the pre-trained  VGG16 and ResNet-50/101/152 models with a simple API. You can use these models as follows:

docs/source/tutorial/convnet.rst
```diff
@@ -0,0 +1,428 @@
...
+
+.. doctest::
+
+    from chainer.links import VGG16Layers
+
+    model = VGG16Layers()
+
+When :class:`~chainer.links.VGG16Layers` is instantiated, the pre-trained
+parameters is automatically downloaded from the author's server. So you can
```
bkvogel: the pre-trained parameters are automatically downloaded

docs/source/tutorial/convnet.rst
```diff
@@ -0,0 +1,428 @@
...
+immediately start to use VGG16 with pre-trained weight as a good image feature
+extractor. See the details of this model here:
+:class:`chainer.links.VGG16Layers`.
+
+In the case of ResNet models, there are 3 variation in terms of the number of
+layers. We have :class:`chainer.links.ResNet50`,
```
bkvogel: In the case of ResNet models, there are three variations differing in the number of layers.

docs/source/tutorial/convnet.rst
mitmul: @bkvogel Thank you for great re-wording! In this sentence, I'd like to say the computation of `conv1x1r` itself is also conditional. So can I say as the below? :

it will conditionally also compute `conv1x1r` and the following batch normalization before the final ReLU layer

I'd appreciate it if you could check this part again :)

mitmul: Sorry, additionally, I'd like to say the `conv1x1r` is used to make the number of channels of the input `x` same as the number of channels of the output of `conv1x1c`. This part "`conv1x1r` to fit the channel dimension of the input `x` and the output `h`" intended to say that. Could you advice me how to add this information to this paragraph?

docs/source/tutorial/convnet.rst
mitmul: Sorry, I intended to say "those models are useful not only for classification tasks but also for many different tasks other than classification tasks". So can I say this as follows?:

It turns out that VGG16 and ResNet are very useful as general feature extractors noy only for image classification tasks but also for many different tasks other than classification.

I'd appreciate it if you could check this part again :)

mitmul:
@bkvogel Thank you for reviewing! I'd like to confirm two points replied in the above comments. I'd appreciate it if you could check them again :)

docs/source/tutorial/convnet.rst
bkvogel: How about:

it will conditionally compute a convolutional layer conv1x1r which will extend the number of channels of the input x to be equal to the number of channels of the output of conv1x1c, and followed by a batch normalization layer before the final ReLU layer.

docs/source/tutorial/convnet.rst
```diff
@@ -0,0 +1,428 @@
...
+            if proj:
+                self.add_link('conv1x1r', L.Convolution2D(
+                    n_in, n_out, 1, stride, 0, initialW=w, nobias=True))
+                self.add_link('bn_r', L.BatchNormalization(n_out))
+            self.proj = proj
+
+        def __call__(self, x, train):
+            h = F.relu(self.bn_a(self.conv1x1a(x), test=not train))
+            h = F.relu(self.bn_b(self.conv3x3b(x), test=not train))
+            h = self.bn_c(self.conv1x1c(x), test=not train)
```
bkvogel: Is this code correct? It looks like you keep overwriting h because the value is not used in the following lines (it keeps reading from x only).

docs/source/tutorial/convnet.rst
bkvogel: How about:

Various ways to write your models were described above. It turns out that VGG16 and ResNet are very useful as general feature extractors for many kinds of tasks, including but not limited to image classification. So, Chainer provides you with the pre-trained VGG16 and ResNet-50/101/152 models with a simple API. You can use these models as follows:

docs/source/tutorial/convnet.rst
mitmul: Sorry, this code is wrong. I'll fix it. Thank you for pointing out it!

mitmul:
Thank you for replying my comments. I fixed all points by following your suggestions :)

bkvogel:
LGTM!

PR #2339, opened 2017-02-27
Add examples that use a custom training loop

bkvogel:
Add examples for MNIST, CIFAR, and ptb that use a custom training loop instead of trainer.
Such examples were in Chainer prior to the Trainer feature but were removed when that feature was added. This PR adds these examples back again.

Although these examples do not use trainer, they do use dataset iterators.


mitmul:
Thank you for this informative PR! I confirmed that these custom training examples successfully trained the models.

The final loss and accuracy on CIFAR10 (with modification on batchsize=64)
```
train mean loss=0.00535679790019989, accuracy=0.99872
test mean loss=0.4296657483577728, accuracy=0.9199
```

The final perplexities of PTB example (without any modification)
```
training perplexity:  3.5897860527038574
test
test perplexity: 4.462698653402887
```

So it definitely works correctly, but the batchsize in the CIFAR10 example might be too large. How about reducing the size to 64 (the current default is 128)?

bkvogel:
OK, I reduced the batchsize to 64. I also made the learning rate adjustable as an argparse argument since for very small batchsizes, the user might need to also reduce the learning rate.

examples/cifar/train_cifar_custom_loop.py
```diff
@@ -0,0 +1,120 @@
...
+            for batch in copy.copy(test_iter):
+                x_array, t_array = convert.concat_examples(batch, args.gpu)
+                x = chainer.Variable(x_array)
+                t = chainer.Variable(t_array)
+                loss = model(x, t)
+                sum_loss += float(loss.data) * len(t.data)
+                sum_accuracy += float(model.accuracy.data) * len(t.data)
+
+            model.predictor.train = True
+            print('test mean loss={}, accuracy={}'.format(
```
ir5: Current code outputs train results and test results alternately, but it seems that they are not aligned (zigzag). It would be better if there is one more space after `test` output.
default

examples/mnist/train_mnist_custom_loop.py
```diff
@@ -0,0 +1,102 @@
+#!/usr/bin/env python
+from __future__ import print_function
+
+try:
```
ir5: Do you use matplotlib in this code? As far as I could see, there is no part where matplotlib is used. This is different from `examples/mnist/train_mnist.py`, where matplotlib may be used in PlotReport.

examples/cifar/train_cifar_custom_loop.py
```diff
@@ -0,0 +1,120 @@
...
+
+        x_array, t_array = convert.concat_examples(batch, args.gpu)
+        x = chainer.Variable(x_array)
+        t = chainer.Variable(t_array)
+        optimizer.update(model, x, t)
+        sum_loss += float(model.loss.data) * len(t.data)
+        sum_accuracy += float(model.accuracy.data) * len(t.data)
+
+        if train_iter.is_new_epoch:
+            print('train mean loss={}, accuracy={}'.format(
```
ir5: It might be helpful if we output the current epoch near here. The same thing may hold for other two example codes.

examples/mnist/train_mnist_custom_loop.py
```diff
@@ -0,0 +1,102 @@
...
+            sum_loss = 0
+            for batch in copy.copy(test_iter):
+                x_array, t_array = convert.concat_examples(batch, args.gpu)
+                x = chainer.Variable(x_array)
+                t = chainer.Variable(t_array)
+                loss = model(x, t)
+                sum_loss += float(loss.data) * len(t.data)
+                sum_accuracy += float(model.accuracy.data) * len(t.data)
+
+            print('test mean loss={}, accuracy={}'.format(
```
ir5: ditto.

examples/cifar/train_cifar_custom_loop.py
```diff
@@ -0,0 +1,120 @@
+from __future__ import print_function
```
ir5: The meaning of the word "custom loop" may not be clear for beginners, so it might be helpful if we add some docstrings that explains this code in detail. (e.g., "This code is a custom loop version of train_cifar.py. That is, we train models not using Trainer class in chainer, but we manually compute the loss of minibatches and apply an optimizer to update the models.")

ir5: The same thing may hold for mnist version.

examples/mnist/train_mnist_custom_loop.py
```diff
@@ -0,0 +1,102 @@
+#!/usr/bin/env python
```
ir5: docstrings?

examples/mnist/train_mnist_custom_loop.py
```diff
@@ -0,0 +1,102 @@
...
+        batch = train_iter.next()
+        x_array, t_array = convert.concat_examples(batch, args.gpu)
+        x = chainer.Variable(x_array)
+        t = chainer.Variable(t_array)
+        optimizer.update(model, x, t)
+        sum_loss += float(model.loss.data) * len(t.data)
+        sum_accuracy += float(model.accuracy.data) * len(t.data)
+
+        if train_iter.is_new_epoch:
+            print('train mean loss={}, accuracy={}'.format(
```
ir5: (ditto) How about printing current epoch?

examples/ptb/train_ptb_custom_loop.py
```diff
@@ -0,0 +1,136 @@
...
+            sum_perp += loss.data
+            count += 1
+
+        optimizer.target.cleargrads()  # Clear the parameter gradients
+        loss.backward()  # Backprop
+        loss.unchain_backward()  # Truncate the graph
+        optimizer.update()  # Update the parameters
+
+        if iteration % 10 == 0:
+            print('training perplexity: ', sum_perp / (count*args.batchsize))
```
ir5: (ditto) How about printing current iter?

examples/ptb/train_ptb_custom_loop.py
```diff
@@ -0,0 +1,136 @@
...
+            sum_perp = 0
+            count = 0
+
+        if train_iter.is_new_epoch:
+            perp = evaluate(model, val_iter)
+            print('validation perplexity: ', perp)
```
ir5: (ditto) How about printing current epoch?

examples/cifar/train_cifar_custom_loop.py
bkvogel: Agreed. I made the changes.

examples/cifar/train_cifar_custom_loop.py
bkvogel: Fixed

examples/cifar/train_cifar_custom_loop.py
bkvogel: Fixed formatting.

examples/mnist/train_mnist_custom_loop.py
bkvogel: Added docstring.

examples/mnist/train_mnist_custom_loop.py
bkvogel: OK, I removed it.

examples/mnist/train_mnist_custom_loop.py
bkvogel: I made the change.

examples/mnist/train_mnist_custom_loop.py
bkvogel: fixed.

examples/ptb/train_ptb_custom_loop.py
bkvogel: fixed

bkvogel:
Thanks for your comments. I have made the changes.

examples/cifar/train_cifar_custom_loop.py
```diff
@@ -0,0 +1,127 @@
...
+                sum_loss / test_count, sum_accuracy / test_count))
+            sum_accuracy = 0
+            sum_loss = 0
+
+    # Save the model and the optimizer
+    print('save the model')
+    serializers.save_npz('mlp.model', model)
+    print('save the optimizer')
+    serializers.save_npz('mlp.state', optimizer)
+
```
ir5: We need one more blank line here.

examples/ptb/train_ptb_custom_loop.py
```diff
@@ -0,0 +1,139 @@
...
+
+        sum_perp += loss.data
+        optimizer.target.cleargrads()  # Clear the parameter gradients
+        loss.backward()  # Backprop
+        loss.unchain_backward()  # Truncate the graph
+        optimizer.update()  # Update the parameters
+
+        if iteration % 10 == 0:
+            print('iteration: ', iteration)
+            print('training perplexity: ', sum_perp / count)
```
ir5: I noticed that the scale of perplexities in the result of `train_ptb.py` and `train_ptb_custom_loop.py` are somehow different.
In `train_ptb.py`, the output is like below.
```
$ python examples/ptb/train_ptb.py -g 1
#vocab = 10000
epoch       iteration   perplexity  val_perplexity............]  0.95%
0           500         651.99
0           1000        277.979
1           1500        216.014     201.447
```
On the other hand, the result of `train_ptb_custom_loop.py` is like below.
```
$ python examples/ptb/train_ptb_custom_loop.py -g 0
#vocab = 10000
iteration:  10
training perplexity:  9.557374954223633
...(omitted)...
iteration:  1490
training perplexity:  5.3896355628967285
iteration:  1500
training perplexity:  5.4502482414245605
iteration:  1510
training perplexity:  5.413826942443848
```
I haven't check the code for computing the perplexities yet, so I'm not sure if these computation results are consistent or not. Do you think this is problematic or not?

ir5:
Thank you for changes. I added another comments.

examples/ptb/train_ptb_custom_loop.py
bkvogel: I fixed the custom loop version to compute the perplexity in the same way as the trainer version.

examples/cifar/train_cifar_custom_loop.py
bkvogel: fixed

ir5:
Thank you for fixes. There seems to be autopep8 related errors. To fix this, could you try merge or rebase current master branch?  (c.f. #2363)

ir5:
Thank you. Now LGTM.

mitmul:
From the offline discussion, this is the `reset` method of SerialIterator:
https://github.com/pfnet/chainer/pull/2387

This has been merged to `master` branch, and will be released with the next version of Chainer (v1.23.0).

bkvogel:
Thank you for suggesting `reset()` method of `SerialIterator`. I have modified the examples to use it.

PR #2344, opened 2017-02-28
Add batched beam search

soskek:
This is PR to add beam search which is based on batch processing.

### DIFF
--- a/examples/seq2seq/seq2seq.py
+++ b/examples/seq2seq/seq2seq.py
@@ -2,6 +2,7 @@ import argparse
 import collections
 
 from nltk.corpus import comtrans
+from nltk.translate import bleu_score
 import numpy
 
 import chainer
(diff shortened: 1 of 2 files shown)


examples/seq2seq/seq2seq.py
```diff
@@ -26,6 +26,23 @@ def sequence_embed(embed, xs):
     return exs
 
 
+def get_topk(x, k=5, axis=1):
+    ids_list = []
+    scores_list = []
+    xp = cuda.get_array_module(x)
+    for i in range(k):
+        ids = xp.argmax(x, axis=axis).astype('i')
```
soskek: In my environment (chainer 1.21.0), `cupy.argmax` exists. It was implemented before I knew.

soskek:
I agree to detaching this from the current example for simplicity!

PR #2346, opened 2017-03-01
Add tests for iterator serialization

Hakuyume:
I added some tests for serialization of `SerialIterator` and `MultiprocessIterator`.


delta2323:
Thank you very much. I am now running GPU tests.

PR #2347, opened 2017-03-01
Add previous_epoch_detail property

Hakuyume:
I added new property `previous_epoch_detail` to `SerialIterator`, `MultiprocessIterator`, and `StandardUpdater`.
This property is similar to `epoch_detail` but it keeps the value of `epoch_detail` at the previous iteration.
Before the first iteration, `previous_epoch_detail` is `None`.


Hakuyume:
It lacks tests for serialization currently.
After #2346 is merged, I will add tests.

delta2323:
I would like to write the background of this PR.

The motivation of `previous_epoch_detail` is to make `IntervalTrigger` support the interval whose length is a float multiple of epochs (which was introduced by #1423). As reported in issue #2244, when we serialize an `Extension`, its interval trigger forgets `self.count`, which is a value of `epoch_detail` of the previous iteration. This value is used to determine if `IntervalTrigger` fires (i.e. to check if the current iteration is the first time that exceeds a multiple of the predefined interval).

`Updater.is_new_epoch` attribute represents if the current epoch is the first iteration of the new epoch. Using it seems a straightforward solution at first sight. That is true if we support only integer intervals. But we sometimes want to use multiple of some floating value (e.g. 2.5 epochs). In that case, we cannot use `is_new_epoch` as it fires only the beginning of each epoch (i.e. can only support integer interval).

As all resumable triggers need the previous value of epoch_detail, we decided to make `Updater` store this value. As `Updater` borrows values related to iterations and epochs from the main `Iterator` it holds by default. We add `previous_epoch_detail` to `Iterator`s.

delta2323:
The design seems OK to me. But as I know the context of this PR, I am not confident my opinion is objective. I will ask some core members to double-check the design.

Hakuyume:
Thank you for checking my code.
I am worried that saving `previous_epoch_detail` is too specialized for this purpose.
I wonder if saving `previous_epoch` and `previous_position` is better for future extensibility.

delta2323:
I merged #2346. So, you can add tests for serialization.

delta2323:
Thank you for your suggestion of `previous_epoch` and `previous_position`. I am wondering we can define `previous_position` well.

In ordinal iterators like `SerialIterator`, the concept of `position` is obvious. But I am not sure same is true for almost all iterators (if not all). Actually, as you can find out from the definition of [`Iterator`](https://github.com/pfnet/chainer/blob/master/chainer/dataset/iterator.py#L17), currently `Iterator` assumes  the concept of `epoch`.

Hakuyume:
I agree that some kind of iterators do not have obvious `position`.
It is better to use `previous_epoch_detail` for the common APIs of `Iterator`.

My concern is that we cannot compute `previous_epoch` or `previous_position` from `previous_epoch_detail`.
If someone want to use these information, he/she have to modify `Iterator`.
My idea is to keep `previous_epoch` and `previous_position` internally.

delta2323:
Do you have some use case of `previous_epoch` and `previous_position` in mind? I would suggest you that it is better to add these attributes after we have some concrete use cases and are confident these attributes are actually necessary.

Hakuyume:
Actually, I have no examples using `previous_epoch` and `previous_position`. It may be only a needless fear.
As you suggested, I leave `previous_epoch` and `previous_position` unimplemented.

Hakuyume:
Thank you for merging. I added serialization tests.

chainer/iterators/multiprocess_iterator.py
```diff
@@ -62,13 +62,17 @@ def __init__(self, dataset, batch_size, repeat=True, shuffle=True,
 
         self._finalized = None
 
+        self._previous_epoch_detail = -1.
```
delta2323: I thought it is more straightforward to initialize `_previous_epoch_detail` with `None` directly, rather than setting `-1` and convert it to `None` in its getter method. Is there any intention for this design?

Hakuyume: I wondered if all serializers can save `None` as is.
I thought it is safer to save integer instead.
If all seralizers support saving `None`, `previous_epoch_detail` should be initialized with `None`.

Hakuyume: Here is an example.
```
import chainer


class Dummy:

    def __init__(self):
        self.value = None

    def serialize(self, serializer):
        self.value = serializer('value', self.value)


dummy = Dummy()
chainer.serializers.save_hdf5('tmp', dummy)  # TypeError: Object dtype dtype('O') has no native HDF5 equivalent
```
It worked for npz.

Hakuyume: There is another solution. We initialize `previous_epoch_detail` with `None` and convert it to `-1` in `serialize` method. In this case, we can remove getter method.

delta2323: OK, I understand it. It is OK to set the initial value of `previous_epoch_detail` to `-1`. In that case, we should explicitly write this specification in the document. Personally, I prefer it as it makes the implementation simpler.

Hakuyume: Do I need modify the docs?
Or simply add a comment as follows.
```       
        # use -1 instead of None internally,
        # since some serializers do not support None.
        self._previous_epoch_detail = -1.
```

Hakuyume:
I updated some docs.

chainer/iterators/multiprocess_iterator.py
delta2323: Thank you very much. The comment looks good to me.

delta2323: Sorry for bringing up the problem again, but I am wondering why we choose `None` instead of `-1` as the initial value of `previous_epoch_detail`. I thought the other alternative is to define `-1` as an invalid value of `previous_epoch_detail` (e.g. in the document of `Iterator`) and not to convert the value internally. I am sorry if I miss some previous discussion and make a non-sense question.

delta2323:
Thank you for your work. I added some comments to your reply. In parallel, I am running GPU tests in the internal Jenkins.

chainer/iterators/multiprocess_iterator.py
Hakuyume: Actually, there is no need to use `None` as the invalid value. I chose `None` because we can easily understand it is invalid. We can use `-1` instead of `None` with enough documents.

Hakuyume:
@delta2323 Thank you for your comment. I replied to it.
I merged master in order to solve conflicts.

chainer/iterators/multiprocess_iterator.py
Hakuyume: Is it better to use `-1` instead of `None`?

delta2323: I think it is somewhat a matter of tastes. I would like the opinion of others.

delta2323: I come to think `None` is better than `-1` for indicating invalid `previous_epoch_detail` because if someone calculates some values from invalid `previous_epoch_detail` with some mathematical operations, it will raise error if we use `None`, while return an invalid value without error if `-1`. So, `None` is safer, I think.

delta2323:
I confirmed all GPU tests pass. So, I merge this PR.

delta2323:
Thank you for enduring discussion with me!

PR #2348, opened 2017-03-01
ignore .idea directory for pycharm users

corochann:
.idea is config files directory which pycharm automatically creates
This PR is to ignore .idea directory from git commit.


okuta:
LGTM!

PR #2349, opened 2017-03-01
Re-compute

anaruse:
This PR is a split from #2027 and focusing solely on re-compute features in order to reduce memory usage.

With this branch, you can specify layer types whose outputs will be forgotten during forward propagation phase and recomputed during backward propagation phase, by using **chainer.use_recompute()** as below, 

    with chainer.use_recompute("RELU", "BN"):
        loss = model(x, t)

Now, ReLU and BatchNormalization layers are supported by this branch, and "RELU" stands for ReLU, "BN" stands for BatchNormalization.

You also need to tell chainer runtime when and where are safe and good to do re-compute during backward propagation using **set_end_of_sub_graph()** method.  Please see [a example of resnet152](https://github.com/anaruse/chainer/blob/re-compute/examples/imagenet/resnet152.py) for the usage.

### DIFF
--- a/chainer/__init__.py
+++ b/chainer/__init__.py
@@ -33,6 +33,8 @@ from chainer.flag import ON  # NOQA
 from chainer.function import force_backprop_mode  # NOQA
 from chainer.function import Function  # NOQA
 from chainer.function import no_backprop_mode  # NOQA
+from chainer.function import use_recompute  # NOQA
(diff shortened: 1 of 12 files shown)


anaruse:
Added feature to decide break-points in a graph automatically so that users may not need to explicitly put **set_break_point()** (formally named as set_end_of_sub_graph()) in their models.

PR #2352, opened 2017-03-02
CTC repeat char

### LINKED ISSUE #2292: Making the CTC function to be the same as Graves 06
In Graves 06, eq.(6) tells that the first line should be taken if l'\_s = b or l'\_{s-2} = l'\_s.
However, in ConnectionistTemporalClassification class of the chainer implementation, only the l'\_s = b condition seems to be implemented.
This makes the CTC probability larger when the same label is repeated, then the CTC loss goes smaller, and the network tend to give repeated labels excessively.  We observed slowness of the training process too.
(Supposing that the repeated labels are 'aa', the current implementation allows both 'aaa' and  'a_a' output, however, the Graves 06's paper supposes that only 'a_a' output should be allowed.)
The blog post https://blog.wtf.sg/2014/10/06/connectionist-temporal-classification-ctc-with-theano/ , on which your code is based, seems to have the same problem.

jnishi:
Fix #2292 

I added the case of  l'_s = b or l'_{s-2} = l'_s. in creation of recurrence relation.

### DIFF
--- a/chainer/functions/loss/ctc.py
+++ b/chainer/functions/loss/ctc.py
@@ -87,16 +87,24 @@ class ConnectionistTemporalClassification(function.Function):
             res = create_recurrence_relation(x, self.zero_padding)
         return res.astype(numpy.float32)
 
-    def recurrence_relation(self, path_length, max_length, dtype, xp):
+    def recurrence_relation(self, label, path_length, max_length, dtype, xp):
         """Transition in forword and backword algorithms is represented as matrix.
 
         See also
         https://blog.wtf.sg/2014/10/06/connectionist-temporal-classification-ctc-with-theano/
         """
-        rr = (xp.eye(max_length, dtype=dtype) +
-              xp.eye(max_length, k=1, dtype=dtype) +
-              xp.eye(max_length, k=2, dtype=dtype) *
-              (xp.arange(max_length, dtype=dtype) % dtype(2)))
+        batch, lab = label.shape
+        repeated = xp.ones((batch, lab * 2 + 1))
+        repeated[:, 1::2] = (label -
+                             xp.take(label, (xp.arange(lab) - 1) % lab
+                                     + xp.arange(batch)[:, None] * batch)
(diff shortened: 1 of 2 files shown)


chainer/functions/loss/ctc.py
```diff
@@ -192,28 +201,30 @@ def calc_trans(self, path, yseq, xp):
     def forward(self, inputs):
         xp = cuda.get_array_module(inputs[0])
         self.input_length = inputs[0]
-        label_length = inputs[1]
+        self.label_length = inputs[1]
```
okuta: `label_length` is used by only this method.
How about using local variable?

chainer/functions/loss/ctc.py
```diff
@@ -87,16 +87,24 @@ def log_matrix(self, x, xp):
...
         See also
         https://blog.wtf.sg/2014/10/06/connectionist-temporal-classification-ctc-with-theano/
         """
-        rr = (xp.eye(max_length, dtype=dtype) +
-              xp.eye(max_length, k=1, dtype=dtype) +
-              xp.eye(max_length, k=2, dtype=dtype) *
-              (xp.arange(max_length, dtype=dtype) % dtype(2)))
+        batch, lab = label.shape
+        repeat_mask = xp.ones((batch, lab * 2 + 1))
+        repeat_mask[:, 1::2] = (label -
```
okuta: How about following code?
```python
repeat_mask[:, 1::2] = (label != xp.take(label, xp.arange(-1, lab - 1) % lab + xp.arange(0, batch * lab, lab)[:, None]))
```

okuta:
Is this bug fix? or enhancement?

jnishi:
Thanks for comments. I applied them. Please review again.

jnishi:
> Is this bug fix? or enhancement?

This is bug fix. As it is mentioned in #2352, this case is considered in the original paper.

okuta:
OK.
LGTM!

PR #2359, opened 2017-03-03
Add transpose method to chainer.Variable

yuyu2172:
In a similar intent as #2236, this PR adds a function to `chainer.Variable` for convenience.


rezoo:
LGTM!

PR #2361, opened 2017-03-03
Format compatibility of iterators

Hakuyume:
Currently, `SerialIterator` and `MultiprocessIterator` serialize `self._order` in different names.
So we cannot change the type of iterator during training.
``` 
dataset = [1, 2, 3]

iterator = chainer.iterators.SerialIterator(dataset, 2)
chainer.serializers.save_npz('tmp', iterator)

iterator = chainer.iterators.MultiprocessIterator(dataset, 2)
chainer.serializers.load_npz('tmp', iterator) #  KeyError: 'order is not a file in the archive'
```

I think this behavior is inconvenient.
I unified the keys of `self._order` to `order`.
For backward compatibility, iterators try to load `_order` if they cannot find `order`.


Hakuyume:
I will add tests based on #2346.

delta2323:
I merged #2346. So, please go ahead with this PR.

Hakuyume:
Thank you for merging #2346. I added tests for backward compatibility and inter-iterator compatibility.

tests/chainer_tests/iterators_tests/test_iterator_compatibility.py
```diff
@@ -0,0 +1,93 @@
...
+    def __init__(self, target):
+        super(DummyDeserializer, self).__init__()
+        self.target = target
+
+    def __getitem__(self, key):
+        raise NotImplementedError
+
+    def __call__(self, key, value):
+        if isinstance(value, numpy.ndarray):
+            value[:] = self.target[key]
```
okuta: Please fix this code. Maybe following code is correct.
```python
if isinstance(value, numpy.ndarray):
    value[:] = self.target[key]
    return value
return self.target[key]
```

Hakuyume:
Thank you for reviewing. I fixed `DummyDeserializer`.

okuta:
Thank you!

PR #2364, opened 2017-03-06
Remove unnecessary ReLU operator in VGG16Layers

rezoo:
It seems that the ReLU operator in the final layer of VGG is not needed.


gwtnb:
LGTM

PR #2370, opened 2017-03-07
Fix BN function channel size support

bkvogel:
Fix batch normalization function to handle multi-dimensional parameters correctly and update documentation of the function to be consistent with how it is documented in the corresponding batch normalization link. Previously, the function assumed that the second dimension of the input was always the only channel dimension.


muupan:
LGTM. I think it is better to add test cases with multiple channel dimensions to the BatchNormalization link as well.

PR #2371, opened 2017-03-07
Add chainer.functions.resize_images

yuyu2172:
This PR adds a function that resizes image like variables.
This is similar to #2272, but `resize_images` does not backpropagate to coordinate variables and has nicer API.

### example

```python
import matplotlib.pyplot as plt
import numpy as np
from skimage.data import astronaut

from chainer.functions import resize_images

img = astronaut().astype(np.float32)
img = np.transpose(img, (2, 0, 1))

out = resize_images(img[None], (512, 256))[0].data.transpose(1, 2, 0)
plt.imshow(out.astype(np.uint8))
plt.show()
```

figure_1-1

### DIFF
--- a/tests/chainer_tests/functions_tests/array_tests/test_resize_images.py
+++ b/tests/chainer_tests/functions_tests/array_tests/test_resize_images.py
@@ -31,7 +31,7 @@ class TestResizeImagesForwardIdentity(unittest.TestCase):
         self.check_forward(cuda.to_gpu(self.x), output_shape=self.in_shape[2:])
 
 
-class TestResizeImagesForwardScale(unittest.TestCase):
+class TestResizeImagesForwardDownScale(unittest.TestCase):
 
     in_shape = (2, 2, 4, 4)
     output_shape = (2, 2, 2, 2)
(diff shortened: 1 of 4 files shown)


taizan:
This is just what I want !!

tests/chainer_tests/functions_tests/array_tests/test_resize_images.py
```diff
@@ -0,0 +1,94 @@
...
+from chainer import functions
+from chainer import gradient_check
+from chainer import testing
+from chainer.testing import attr
+from chainer.testing import condition
+
+
+@testing.parameterize(*testing.product({
+    'in_shape': [(2, 3, 8, 6), (2, 1, 4, 6)],
+}))
```
rezoo: Are there any reasons to limit the tests to `numpy.float32`?
It seems that your proposed function also accepts `float64` and `float16`.

tests/chainer_tests/functions_tests/array_tests/test_resize_images.py
```diff
@@ -0,0 +1,94 @@
...
+
+    def test_forward_cpu(self):
+        self.check_forward(self.x, output_shape=self.in_shape[2:])
+
+    @attr.gpu
+    def test_forward_gpu(self):
+        self.check_forward(cuda.to_gpu(self.x), output_shape=self.in_shape[2:])
+
+
+class TestResizeImagesForwardScale(unittest.TestCase):
```
rezoo: Ditto. It seems that your proposed function also accepts `float64` and `float16`.

tests/chainer_tests/functions_tests/array_tests/test_resize_images.py
```diff
@@ -0,0 +1,94 @@
...
+    @attr.gpu
+    def test_forward_gpu(self):
+        self.check_forward(
+            cuda.to_gpu(self.x), output_shape=self.output_shape[2:])
+
+
+@testing.parameterize(*testing.product({
+    'in_shape': [(2, 3, 8, 6), (2, 1, 4, 6)],
+    'output_shape': [(10, 5), (3, 4)]
+}))
```
rezoo: Ditto. It seems that your proposed function also accepts `float64` and `float16`.

tests/chainer_tests/functions_tests/array_tests/test_resize_images.py
```diff
@@ -0,0 +1,94 @@
...
+
+    @attr.gpu
+    def test_forward_gpu(self):
+        self.check_forward(cuda.to_gpu(self.x), output_shape=self.in_shape[2:])
+
+
+class TestResizeImagesForwardScale(unittest.TestCase):
+
+    in_shape = (2, 2, 4, 4)
+    output_shape = (2, 2, 2, 2)
```
rezoo: Just in case is it better to add the test case where `output_shape` is larger than `in_shape`?

tests/chainer_tests/functions_tests/array_tests/test_resize_images.py
yuyu2172: Thanks.
`float64` and `float16` are not supported by the backward pass of this function because `cupy.scatter_add` does not support them.

I changed the type-check to only accept `float32`.

tests/chainer_tests/functions_tests/array_tests/test_resize_images.py
yuyu2172: Nice. I added that case as `TestResizeImagesForwardUpScale`

yuyu2172:
Thanks for commenting!
I fixed the code.

rezoo:
Good. then I would like to merge.

PR #2379, opened 2017-03-08
Add axis option to l2_normalization

yuyu2172:
This PR adds `axis` option to `chainer.functions.normalization` so that it can normalize arrays along arbitrary axis.

Previously, the function only supported the case when an input array is of dimension two and normalization happens in the second axis.
The default value of `axis` is set to behave identical to the previous implementation when `axis` is not supplied by the user.

### DIFF
--- a/tests/chainer_tests/functions_tests/normalization_tests/test_l2_normalization.py
+++ b/tests/chainer_tests/functions_tests/normalization_tests/test_l2_normalization.py
@@ -27,9 +27,10 @@ class TestL2Normalization(unittest.TestCase):
         self.gy = numpy.random.uniform(-1, 1, self.shape).astype(numpy.float32)
 
     def check_forward(self, x_data, axis):
+        eps = 1e-5
         x = chainer.Variable(x_data)
 
-        y = functions.normalize(x, axis=axis)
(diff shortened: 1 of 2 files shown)


tests/chainer_tests/functions_tests/normalization_tests/test_l2_normalization.py
```diff
@@ -10,57 +11,69 @@
...
 from chainer.testing import condition
-from chainer.utils import type_check
 
 
 @testing.parameterize(
-    {'shape': (4, 15)},
+    {'shape': (4, 15), 'axis': 1},
+    {'shape': (4, 3, 2, 5), 'axis': 0},
+    {'shape': (4, 3, 2, 5), 'axis': 1},
+    {'shape': (4, 3, 2, 5), 'axis': 2},
```
mattya: This test case frequently fails by exceeding the tolerance.

yuyu2172:
Thanks for commenting.

There was error in forward pass of test. The expected values were not calculated correctly.
I also made tests for backward pass more robust.

The new tests have never failed after executing tests for 100 times.

mattya:
Thanks for the update.  LGTM!

PR #2386, opened 2017-03-09
Static-type CUDA kernel argument pointers

niboshi:
This PR improves performance of linear_launch() by static-typing kernel argument pointers.

One concern is that dependency from function.pyx to core.pyx is introduced,
because core.ndarray and core.Indexer must be visible to function.pyx in order to do static-typing.

In my benchmark, linear_launch() is sped up by about 30%.

### DIFF
--- a/cupy/core/core.pxd
+++ b/cupy/core/core.pxd
@@ -1,4 +1,3 @@
-from libc cimport stdint
 from libcpp cimport vector
 from cupy.cuda cimport memory
 
--- a/cupy/core/core.pxd
+++ b/cupy/core/core.pxd
@@ -1,0 +1,73 @@
+from libc cimport stdint
+from libcpp cimport vector
+from cupy.cuda cimport memory
+
+from cupy.cuda.function cimport CPointer
+
+
+cdef class ndarray:
+    cdef:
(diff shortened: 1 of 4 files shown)


cupy/core/core.pxd
```diff
@@ -0,0 +1,73 @@
+from libc cimport stdint
```
okuta: This line is redundant.

cupy/cuda/function.pyx
```diff
@@ -64,8 +67,10 @@ cdef inline CPointer _pointer(x):
     cdef Py_ssize_t itemsize
     if x is None:
         return CPointer()
-    if hasattr(x, 'cstruct'):
```
niboshi: Currently only core.ndarray and core.Indexer have cstruct attribute (in my understanding), so within the scope of cupy, it doesn't break backward compatibility. It would be a condition that never be satisfied while costly. Is the statement still needed?

okuta: OK, I understand. Please remove it.

niboshi:
Thank you for reviewing! I followed your review.

okuta:
OK!

PR #2387, opened 2017-03-09
Stop copying iterators for efficiency

iwiwi:
This PR aims to improve the performance of common use cases with ChainerMN by reducing the frequency of process spawn.

In `Evaluator`, currently iterators are copied for every evaluation time. For most cases, this is not very time consuming. However, we found that this is very slow when we are using `MultiProcessIterator` with `forkserver` or `spawn` modes (c.f., https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods). Therefore, to avoid spawning new processes for each evaluation time, this PR proposes to introduce `reset` method to iterators and use it in `Evaluator`.

### DIFF
--- a/chainer/iterators/multiprocess_iterator.py
+++ b/chainer/iterators/multiprocess_iterator.py
@@ -45,22 +45,16 @@ class MultiprocessIterator(iterator.Iterator):
         self.dataset = dataset
         self.batch_size = batch_size
         self._repeat = repeat
-        if shuffle:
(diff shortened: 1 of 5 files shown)


iwiwi:
I talked with Okuta-san, and I was requested to add tests checking `NotImplementedError`

iwiwi:
Added tests for checking `NotImplementedError`

okuta:
LGTM!

iwiwi:
Thanks!! 👏

PR #2390, opened 2017-03-10
Support cleargrad in optimizer hook

### LINKED ISSUE #2389: cleargrad in optimizer hooks
If `cleargrad` method is used in optimizer hooks, the optimizer raises `TypeError` exception.
```
class Freeze(object):
    name = 'Freeze'

    def __call__(self, opt):
        for param in opt.target.params():
            param.cleargrad()


model = chainer.links.Linear(1, 1)
optimizer.setup(model)
optimizer.add_hook(Freeze())

optimizer.update()  # TypeError: unsupported operand type(s) for *: 'float' and 'NoneType'
```
If we use `param.grad[:] = 0` instead of `param.cleargrad`, it works well.

This is because `GradientMethod`  replace 'None' only before calling hooks.
https://github.com/pfnet/chainer/blob/master/chainer/optimizer.py#L400-L404

Hakuyume:
Fix #2389

### DIFF
--- a/chainer/optimizer.py
+++ b/chainer/optimizer.py
@@ -369,15 +369,18 @@ class GradientMethod(Optimizer):
 
     """
 
+    def _replace_cleared_grads(self):
+        for name, param in self.target.namedparams():
+            if param.grad is None:
+                with cuda.get_device(param.data):
+                    xp = cuda.get_array_module(param.data)
+                    param.grad = xp.zeros_like(param.data)
+
     def call_hooks(self):
         """Invokes hook functions in registration order."""
         for hook in six.itervalues(self._hooks):
             hook(self)
-            for name, param in self.target.namedparams():
-                if param.grad is None:
-                    with cuda.get_device(param.data):
-                        xp = cuda.get_array_module(param.data)
-                        param.grad = xp.zeros_like(param.data)
+            self._replace_cleared_grads()
 
     def update(self, lossfun=None, *args, **kwds):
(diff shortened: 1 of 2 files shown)


chainer/optimizer.py
```diff
@@ -369,6 +369,19 @@ class GradientMethod(Optimizer):
 
     """
 
+    def _replace_cleared_grads(self):
+        for name, param in self.target.namedparams():
+            if param.grad is None:
+                with cuda.get_device(param.data):
+                    xp = cuda.get_array_module(param.data)
+                    param.grad = xp.zeros_like(param.data)
+
```
niboshi: Can you make this method public? By doing so, derived optimizers can override it to skip reallocation if not needed. (See the TODO below)
```
# TODO(unno): Some optimizers can skip this process if they does not
# affect to a parameter when its gradient is zero.
```

Then you can also remove this TODO.

As for the method name, `_replace_cleared_grads()` can be a bit ambiguous as a public method. Can you rename it to `reallocate_cleared_grads()`?

Hakuyume:
@niboshi Thank you for your review.
I changed the name of method and removed TODO.

niboshi:
Thank you for the fix!
Finally, could you add brief docstring to that method, mentioning that it can be overridden to skip allocation as needed?

Hakuyume:
I added a docstring to `reallocate_cleared_grads`.

niboshi:
LGTM. Thank you for contribution!

PR #2369, opened 2017-03-06
Fix error with advanced indexing when a zero length array-like is the index

yuyu2172:
This resolves issue #2366.

Three bugs are fixed

+ index is `[]`
+ index is `[[]]`
+ index is a boolean array of size 0

Issue #2366, opened 2017-03-06
Error in cupy's advanced indexing when an empty list is given


muupan:
When an empty list is given as a selection object to numpy's advanced indexing, no item is selected for that axis.
```
>>> numpy.zeros((4,3,2))[[]]
array([], shape=(0, 3, 2), dtype=float64)
```

However, cupy's advanced indexing raises an error.
```
>>> cupy.zeros((4,3,2))[[]]
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "cupy/core/core.pyx", line 1142, in cupy.core.core.ndarray.__getitem__ (cupy/core/core.cpp:23915)
    raise IndexError(
IndexError: arrays used as indices must be of integer or boolean type. (actual: <class 'numpy.float64'>)
```

yuyu2172:
Related to this bug, I found more.

+ [[]]  (double nested list)
+ `numpy.array([], dtype=numpy.bool)`   (a boolean array of size 0)

yuyu2172:
#2369 solved this issue, so I think we can close it.

niboshi:
Confirmed that it's fixed now. Thank you for reporting!

PR #2393, opened 2017-03-10
Support zero length list for cupy.scatter_add and cupy.ndarray.__setitem__

yuyu2172:
Related #2369, #2366.

Three bugs are fixed

+ index is `[]`
+ index is `[[]]`
+ index is a boolean array of size 0


rezoo:
LGTM!

PR #2395, opened 2017-03-10
Stop creating cuda.Device in Variable.backward when cupy is not used

### LINKED ISSUE #2394: cuda.Device should not be created in Variable.backward when cupy is not used
From chainer==1.21.0, `cuda.Device()` is created always in `Variable.backward` whenever `cuda.available`, even when `Variable.data` is `numpy.ndarray`.

https://github.com/pfnet/chainer/blob/d91ba89da9bbef0b316006648b162d9f4bffbce9/chainer/variable.py#L357

Not only being unnecessary, it can result in an error when `Variable.backward` is called by another process.

The code below runs without problem with chainer==1.20.0.1, but raises `cupy.cuda.runtime.CUDARuntimeError: cudaErrorInitializationError: initialization error` at `cuda.Device()` with chainer==1.21.0.
```
import multiprocessing as mp
import chainer
print(chainer.__version__)
import numpy as np
chainer.cuda.get_device(0).use()

def hoge():
    # Note that this function uses numpy, not cupy, but still it results in an error
    x = chainer.Variable(np.asarray([0], dtype=np.float32)) + 1
    x.backward()

p = mp.Process(target=hoge)
p.start()
p.join()
```

muupan:
Resolves #2394

### DIFF
--- a/chainer/variable.py
+++ b/chainer/variable.py
@@ -352,7 +352,7 @@ Actual: {0}'''.format(type(data))
         if self.creator is None:
             return
         initial_device = None
-        if cuda.available:
+        if cuda.available and isinstance(self.data, cuda.cupy.ndarray):
             try:
                 initial_device = cuda.Device()
             except cuda.cupy.cuda.runtime.CUDARuntimeError as e:


gwtnb:
LGTM

PR #2398, opened 2017-03-11
Improve docs of sigmoid, hard sigmoid

keisuke-umezawa:

Modified functions/links:
- functions/activation/sigmoid
- functions/activation/hard_sigmoid

### DIFF
--- a/chainer/functions/activation/hard_sigmoid.py
+++ b/chainer/functions/activation/hard_sigmoid.py
@@ -70,11 +70,12 @@ def hard_sigmoid(x):
     .. admonition:: Example
 
         It maps the input values into the range of :math:`[0, 1]`.
-        >>> x = np.arange(-2, 3, 2).astype('f')
+
+        >>> x = np.array([-2.6, -1, 0, 1, 2.6])
         >>> x
-        array([-2.,  0.,  2.], dtype=float32)
+        array([-2.6, -1. ,  0. ,  1. ,  2.6])
         >>> F.hard_sigmoid(x).data
-        array([ 0.09999999,  0.5       ,  0.89999998], dtype=float32)
+        array([ 0. ,  0.3,  0.5,  0.7,  1. ])
 
     """
     return HardSigmoid()(x)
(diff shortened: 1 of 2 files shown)


chainer/functions/activation/sigmoid.py
```diff
@@ -76,7 +76,6 @@ def sigmoid(x, use_cudnn=True):
 
     .. admonition:: Example
 
-        It maps the input values into the range of :math:`[0, 1]`.
```
mitmul: Could you tell me why you removed this line?

chainer/functions/activation/hard_sigmoid.py
```diff
@@ -59,10 +59,21 @@ def hard_sigmoid(x):
...
+            Input variable. A :math:`(s_1, s_2, ..., s_N)`-shaped float array.
 
     Returns:
-        ~chainer.Variable: Output variable.
+        ~chainer.Variable: Output variable. A
+        :math:`(s_1, s_2, ..., s_N)`-shaped float array.
+
+    .. admonition:: Example
+
+        >>> x = np.arange(-2, 3, 2).astype('f')
```
mitmul: How about using `x = np.array([-2.6, -1, 0, 1, 2.6])` instead? Because I thinks it's better to show all different cases that the above condition includes.

chainer/functions/activation/sigmoid.py
keisuke-umezawa: I deleted this line because the output html was broken in my environment.
example

chainer/functions/activation/hard_sigmoid.py
keisuke-umezawa: ok. I will fix them!

chainer/functions/activation/sigmoid.py
keisuke-umezawa: Ok, I added blank lines before and after this line.

mitmul:
LGTM

chainer/functions/activation/sigmoid.py
mitmul: @keisuke-umezawa Sorry, could you check this again? Somehow it seems to be reverted before you added lines here.

PR #2405, opened 2017-03-13
Fix repeat test

okuta:
Current code calls `setup` and `teardown` twice. So, test was finished not cleanly.


chainer/testing/condition.py
```diff
@@ -52,11 +52,13 @@ def fail():
 
             for _ in six.moves.range(times):
                 suite = unittest.TestSuite()
+                # create new instance for once call setup and teardown
```
delta2323: Suggestion: "Create new instance to call the setup and the teardown only once."

okuta: Thanks!

delta2323:
LGTM

PR #2406, opened 2017-03-13
Suppress feature warning

okuta:
This PR improve test output readability.
Suppress feature warning that is printed by experimental function.

### DIFF
--- a/chainer/testing/helper.py
+++ b/chainer/testing/helper.py
@@ -1,5 +1,6 @@
 import pkg_resources
 import unittest
+import warnings
 
 
 def with_requires(*requirements):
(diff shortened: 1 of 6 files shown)


delta2323:
It seems this PR is made on top of #2405. So should I check #2405 first?

okuta:
Yes, please.

delta2323:
As #2405 is merged, please merge the latest master branch if necessary.

delta2323:
`chainer.links.TheanoFunction` is also experimental. Please annotate its test with `@testing.suppress_feature_warning()`

okuta:
@delta2323 I fixed.

chainer/testing/helper.py
```diff
@@ -29,3 +30,43 @@ def with_requires(*requirements):
...
+        self._warn.__exit__(None, None, None)
+        original_teardown(self)
+
+    klass.setUp = setUp
+    klass.tearDown = tearDown
+
+    return klass
+
+
+def suppress_feature_warning():
```
delta2323: Is feature should be future?

chainer/testing/helper.py
```diff
@@ -29,3 +30,43 @@ def with_requires(*requirements):
...
+    """Decorator that suppress feature warning.
```
delta2323: ditto (feature -> future?)

delta2323:
It seems to me chainer-test/#209 also tries to solve the same problem. I think it is better if we can solve it with environment variables because it is easier and we need not to define new functions.

Could you check the PR chainer-test/#209 to see if my understanding is correct and we can apply the same changes to Travis as well?

okuta:
OK, I understand.
That is good idea.

PR #2407, opened 2017-03-13
Stop checking that optimizer is an instance of Optimizer class

iwiwi:
This is necessary workaround for compatibility with ChainerMN's `MultiNodeOptimizer`.

ChainerMN's `MultiNodeOptimizer` wraps an actual optimizer. There is no good reason for `MultiNodeOptimizer` to inherit `Optimizer`, as it would not use any methods nor variables of `Optimizer`. And, for some technical reasons from ChainerMN side, it is not preferable for `MultiNodeOptimizer` to inherit `Optimizer`. Therefore, this PR proposes Chainer's `Updater` to change the way to determine whether `optimizer` argument is a `dict` or not. (This is the conclusion of discussion with @beam2d -san.)

### DIFF
--- a/chainer/training/updater.py
+++ b/chainer/training/updater.py
@@ -123,7 +123,7 @@ class StandardUpdater(Updater):
             iterator = {'main': iterator}
         self._iterators = iterator
 
-        if isinstance(optimizer, optimizer_module.Optimizer):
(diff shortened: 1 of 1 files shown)


okuta:
LGTM!

Issue #1816, opened 2016-10-27
Use initializer in NStepLSTM


delta2323:
The initialization method of current `NStepLSTM` link is fixed to sampling from Gaussian distribution (see [here](https://github.com/pfnet/chainer/blob/master/chainer/links/connection/n_step_lstm.py#L62)). One of the possible extension of this link is to add initializers to the argument and use it in `__init__`.

MannyKayy:
Reopening this issue:

![image](https://user-images.githubusercontent.com/7952854/39865715-b23fcaf6-5445-11e8-8060-f50848978401.png)


Runtime Info:

Chainer: 5.0.0a1
NumPy: 1.13.0
CuPy:
  CuPy Version          : 5.0.0a1
  CUDA Root             : /usr/local/cuda
  CUDA Build Version    : 8000
  CUDA Driver Version   : 9020
  CUDA Runtime Version  : 8000
  cuDNN Build Version   : 7102
  cuDNN Version         : 7102
  NCCL Build Version    : 2104

PR #2409, opened 2017-03-14
Accept initializers for weights and biases in NStepLSTM

niboshi:
This PR is to allow arbitrary initializers in `NStepLSTM.__init__` (See issue #1816).

### DIFF
--- a/chainer/links/connection/n_step_lstm.py
+++ b/chainer/links/connection/n_step_lstm.py
@@ -24,12 +24,14 @@ class NStepLSTMBase(link.ChainList):
         in_size (int): Dimensionality of input vectors.
         out_size (int): Dimensionality of hidden states and output vectors.
         dropout (float): Dropout ratio.
-        initialW: Value used to initialize the weight matrices. May be an
-            initializer instance or another value that
-            :func:`~chainer.init_weight` helper function can take.
-        initial_bias: Value used to initialize the bias vectors. May be an
-            initializer instance or another value that
-            :func:`~chainer.init_weight` helper function can take.
+        initialW (2-D array): Initial weight value. If ``None``, then the
+            default initializer is used.
+            May also be a callable that takes ``numpy.ndarray`` or
+            ``cupy.ndarray`` and edits its value.
+        initial_bias (1-D array): Initial bias value. If ``None``, the bias
+            vector is initialized to zero.
+            May also be a callable that takes ``numpy.ndarray`` or
+            ``cupy.ndarray`` and edits its value.
         use_bi_direction (bool): if ``True``, use Bi-directional LSTM.
 
     .. seealso::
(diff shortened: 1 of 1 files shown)


unnonouno:
Thank you, but @aonotas is currently working for refactoring n-step-lstm. It is better to merge this PR after @aonotas 's work will be finished.

niboshi:
OK, I got it.

niboshi:
Rebased to the current master.

Note that `initializers.init_weight()` defaults to the scaled normal distribution when the second argument (`initial_bias`) is `None`.

unnonouno:
Merge the master. Many __init__ method for Function are fixed.

niboshi:
Rebased to master.

chainer/links/connection/n_step_lstm.py
```diff
@@ -53,12 +61,12 @@ def __init__(self, n_layers, in_size, out_size, dropout, use_bi_direction,
...
                             w_in = out_size
-                        w = variable.Parameter(
-                            normal.Normal(numpy.sqrt(1. / w_in)),
-                            (out_size, w_in))
-                        b = variable.Parameter(0, (out_size,))
-                        setattr(weight, 'w%d' % j, w)
-                        setattr(weight, 'b%d' % j, b)
+                        name_w = 'w{}'.format(j)
+                        name_b = 'b{}'.format(j)
+                        w = variable.Parameter(initialW, (out_size, w_in))
```
unnonouno: Do we need to use `initializers._get_initializer` for `initialW`?

niboshi:
Fixed to call `_get_initializer` for `initialW`. Also updated docstring.

I understand the only point of `_get_initializer` is to convert `None` to the "default" initializer, which is `LeCunNormal`. Is that correct?

I think it could be done in better way.
If this understanding is correct, I will make another PR to refactor.

unnonouno:
Yes, and it is the same behavior as links.Linear.

unnonouno:
LGTM

MannyKayy:
This PR was overwritten during the [rnn code refactor].

(https://github.com/chainer/chainer/commit/f8e99052b84375faf86bce88c2b86b57db1f6885#diff-38391e4f2027e281d76ad22b877bd019).

image


Runtime Info:

Chainer: 5.0.0a1
NumPy: 1.13.0
CuPy:
  CuPy Version          : 5.0.0a1
  CUDA Build Version    : 8000
  CUDA Driver Version   : 9020
  CUDA Runtime Version  : 8000
  cuDNN Build Version   : 7102
  cuDNN Version         : 7102
  NCCL Build Version    : 2104

delta2323:
Thank you very much for your report. What you pointed out seems to be correct and should be considered as a bug.

delta2323:
I made an issue dealing with it (#4755). By the way, this PR was merged in v3.0.0rc1, but is not written in [the release note](https://github.com/chainer/chainer/releases/tag/v3.0.0rc1) of this version. I think we add this PR to it.

PR #2416, opened 2017-03-17
[WIP] Strict dtype check

niboshi:
This PR aims to fix dtype.char inconsistency between numpy and cupy filed in issue #1646.

Currently some tests will fail.
(for example: tests/cupy_tests/math_tests/test_misc.py)


niboshi:
Closing the PR for now, because the issue is not that urgent and I don't have much time for this.

Issue #2418, opened 2017-03-17
Bugs in advanced indexing of Cupy's __getitem__ when indexes are list


yuyu2172:
There is a inconsistency between NumPy and CuPy's advanced indexing when input is a list.

```python
import numpy
import cupy
x = numpy.arange(24).reshape(2, 3, 4)
x_gpu = cupy.arange(24).reshape(2, 3, 4)
print(x[[1]].shape)   # (1, 3, 4)
print(x[[[1]]].shape)   # (1, 3, 4)

print(x_gpu[[1]].shape)  # (1, 3, 4)
print(x_gpu[[[1]]].shape)  # (1, 1, 3, 4)
```

This happens because NumPy handles list type indices differently based on its contents.
When `indexes = [int_1, ..., int_n]`, the indexing is equivalent to the case when `eq_indexes = (indexes,)`.

However, when `indexes` contain anything other than integers (e.g. `indexes=[[1]]`, which is `isinstance(indexes[0], list) == True`), the indexing is equivalent to the case when `eq_indexes = (indexes[0], indexes[1], ..., indexes[n-1])`.

yuyu2172:
Solved

PR #2419, opened 2017-03-17
Fix Advanced indexing of Cupy's __getitem__ when an index argument is a list

yuyu2172:
This PR fixes a bug found in #2418.

### DIFF
--- a/cupy/core/core.pyx
+++ b/cupy/core/core.pyx
@@ -4,6 +4,7 @@ from __future__ import division
 import ctypes
 import sys
 
+import copy
 import numpy
 import six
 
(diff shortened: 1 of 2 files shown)


cupy/core/core.pyx
```diff
@@ -1098,10 +1098,18 @@ cdef class ndarray:
...
+        if isinstance(slices, tuple):
             slices = list(slices)
+        elif isinstance(slices, list):
+            slices = list(slices)  # copy list
+            single_integer_array_indexing = True
+            for s in slices:
+                if not isinstance(s, int):
+                    single_integer_array_indexing = False
+            if single_integer_array_indexing:
+                slices = [slices]
```
niboshi: How about replacing this part with `if all(isinstance(s, int) for s in slices)`? It's more idiomatic and readable. Your code is slightly faster, but the difference is negligible.
The same goes for PR #2420.

niboshi:
Thank you very much for the PR!
Could you consider my review comment?

yuyu2172:
Thanks for commenting.
I fixed both this PR and #2420.

niboshi:
LGTM. Thank you!

PR #2420, opened 2017-03-17
Fix Advanced indexing of Cupy's setitem and scatter_add when an index argument is a list

yuyu2172:
This PR fixes a bug similar to #2418.
The solution is similar to #2419.


niboshi:
LGTM. Thank you!

PR #2423, opened 2017-03-17
Improve Windows NVTX support

takagi:
This PR improves NVTX support for Windows environments, following @kmatsuura 's change.

### DIFF
--- a/cupy/cuda/cupy_cuda.h
+++ b/cupy/cuda/cupy_cuda.h
@@ -24,7 +24,9 @@
 #include <cuda_profiler_api.h>
 #include <cuda_runtime.h>
 #include <curand.h>
+#ifndef CUPY_NO_NVTX
 #include <nvToolsExt.h>
+#endif // #ifndef CUPY_NO_NVTX
 
 extern "C" {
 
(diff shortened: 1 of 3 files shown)


okuta:
How about using `nvToolsExt64_1.lib` to avoid coping dll file.

kmatsuura:
Add to check existence of nvToolsExt64_1.dll
https://github.com/kmatsuura/chainer/commit/c30521c4cc95f0d2876b1c5a42561ed4dc0d1b65

takagi:
I will follow the fix.

cupy/cuda/cupy_cuda.h
```diff
@@ -24,9 +24,7 @@
 #include <cuda_profiler_api.h>
 #include <cuda_runtime.h>
 #include <curand.h>
-#ifndef _WIN32
```
okuta: This modify assumes `os.path.exists(nvtoolsext_path) == True`.
Is it correct?

takagi: It does fail if nvToolsExt.h is not found.

okuta: OK.

okuta:
LGTM!

PR #2424, opened 2017-03-17
Add GoogLeNet link

ronekko:
I added pretrained GoogLeNet link to chainer.links.model.vision .

### DIFF
--- a/chainer/links/model/vision/googlenet.py
+++ b/chainer/links/model/vision/googlenet.py
@@ -1,0 +1,401 @@
+from __future__ import print_function
+from collections import OrderedDict
+import os
+
+import numpy
+try:
+    from PIL import Image
+    available = True
+except ImportError as e:
+    available = False
+    _import_error = e
+
+from chainer.dataset.convert import concat_examples
+from chainer.dataset import download
+from chainer import flag
+from chainer.functions.activation.relu import relu
+from chainer.functions.activation.softmax import softmax
+from chainer.functions.array.reshape import reshape
+from chainer.functions.math.sum import sum as sum_function
+from chainer.functions.pooling.average_pooling_2d import average_pooling_2d
+from chainer.functions.pooling.max_pooling_2d import max_pooling_2d
+from chainer.functions.normalization.local_response_normalization import (
+    local_response_normalization)
+from chainer.functions.noise.dropout import dropout
+from chainer.initializers import constant
+from chainer.initializers import uniform
+from chainer import link
+from chainer.links.connection.convolution_2d import Convolution2D
+from chainer.links.connection.inception import Inception
+from chainer.links.connection.linear import Linear
+from chainer.serializers import npz
+from chainer.utils import imgproc
+from chainer.variable import Variable
+
+
+class GoogLeNet(link.Chain):
+
+    """A pre-trained GoogLeNet model provided by BVLC [1].
+
+    When you specify the path of the pre-trained chainer model serialized as
+    a ``.npz`` file in the constructor, this chain model automatically
+    initializes all the parameters with it.
+    This model would be useful when you want to extract a semantic feature
+    vector per image, or fine-tune the model on a different dataset.
+
+    If you want to manually convert the pre-trained caffemodel to a chainer
+    model that can be specified in the constructor,
+    please use ``convert_caffemodel_to_npz`` classmethod instead.
+
+    .. [1] `<https://github.com/BVLC/caffe/tree/master/models/bvlc_googlenet>`_
+
+    Args:
+        pretrained_model (str): the destination of the pre-trained
+            chainer model serialized as a ``.npz`` file.
+            If this argument is specified as ``auto``,
+            it automatically downloads the caffemodel from the internet.
+            Note that in this case the converted chainer model is stored
+            on ``$CHAINER_DATASET_ROOT/pfnet/chainer/models`` directory,
+            where ``$CHAINER_DATASET_ROOT`` is set as
+            ``$HOME/.chainer/dataset`` unless you specify another value
+            as a environment variable. The converted chainer model is
+            automatically used from the second time.
+            If the argument is specified as ``None``, all the parameters
+            are not initialized by the pre-trained model, but the default
+            initializer used in the original paper, i.e.,
+            ``chainer.initializers.GlorotUniform(scale=1.0)``.
+
+    Attributes:
+        available_layers (list of str): The list of available layer names
+            used by ``__call__`` and ``extract`` methods.
+
+    """
+
+    image_mean = numpy.array([104, 117, 123], dtype=numpy.float32)  # BGR
+
+    def __init__(self, pretrained_model='auto'):
+        if pretrained_model:
+            # As a sampling process is time-consuming,
+            # we employ a zero initializer for faster computation.
+            kwargs = {'initialW': constant.Zero()}
+        else:
+            # employ default initializers used in the original paper
+            kwargs = {'initialW': uniform.GlorotUniform(scale=1.0)}
+        super(GoogLeNet, self).__init__(
+            conv1=Convolution2D(3, 64, 7, stride=2, pad=3, **kwargs),
+            conv2_reduce=Convolution2D(64, 64, 1, **kwargs),
+            conv2=Convolution2D(64, 192, 3, stride=1, pad=1, **kwargs),
+            inc3a=Inception(192, 64, 96, 128, 16, 32, 32),
+            inc3b=Inception(256, 128, 128, 192, 32, 96, 64),
+            inc4a=Inception(480, 192, 96, 208, 16, 48, 64),
+            inc4b=Inception(512, 160, 112, 224, 24, 64, 64),
+            inc4c=Inception(512, 128, 128, 256, 24, 64, 64),
+            inc4d=Inception(512, 112, 144, 288, 32, 64, 64),
+            inc4e=Inception(528, 256, 160, 320, 32, 128, 128),
+            inc5a=Inception(832, 256, 160, 320, 32, 128, 128),
+            inc5b=Inception(832, 384, 192, 384, 48, 128, 128),
+            loss3_fc=Linear(1024, 1000, **kwargs),
+
+            loss1_conv=Convolution2D(512, 128, 1, **kwargs),
+            loss1_fc1=Linear(2048, 1024, **kwargs),
+            loss1_fc2=Linear(1024, 1000, **kwargs),
+
+            loss2_conv=Convolution2D(528, 128, 1, **kwargs),
+            loss2_fc1=Linear(2048, 1024, **kwargs),
+            loss2_fc2=Linear(1024, 1000, **kwargs)
+        )
+        if pretrained_model == 'auto':
+            _retrieve(
+                'bvlc_googlenet.npz',
+                'http://dl.caffe.berkeleyvision.org/bvlc_googlenet.caffemodel',
+                self)
+        elif pretrained_model:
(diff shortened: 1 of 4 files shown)


rezoo:
Thank you for your PR, and sorry for the late review.
Now I would like to review your PR.

chainer/links/model/vision/googlenet.py
```diff
@@ -0,0 +1,401 @@
...
+        else:
+            x = x[:, :, 16:240, 16:240]
+        # Set volatile option to ON to reduce memory consumption
+        x = Variable(self.xp.asarray(x), volatile=flag.ON)
+        y = self(x, layers=['prob'])['prob']
+        if oversample:
+            n = y.data.shape[0] // 10
+            y_shape = y.data.shape[1:]
+            y = reshape(y, (n, 10) + y_shape)
+            y = sum_function(y, axis=1) / 10
```
rezoo: I suppose that you decided to use `sum_function` to avoid overriding.
How about using `from chainer.functions.math import sum` and replacing it with `sum.sum(y, axis=1)`?

docs/source/reference/links.rst
```diff
@@ -191,6 +191,13 @@ where ``fc7`` denotes a layer before the last fully-connected layer.
 Unlike the usual links, these classes automatically load all the
 parameters from the pre-trained models during initialization.
 
+GoogLeNet
+~~~~~~~~~
+.. autoclass:: GoogLeNet
+   :members:
+
+.. autofunction:: chainer.links.model.vision.googlenet.prepare
+
```
rezoo: I'm curious why you decided to place `GoogLeNet` at the top instead of `VGG16Layers`.
My feeling is that VGG model is the most commonly used, and GoogLeNet is the next.

If you think that these models should be arranged alphabetically, it seems better to rearrange the other models.

rezoo:
I have one additional question: did you check whether the output is equal to the original one?

chainer/links/model/vision/googlenet.py
```diff
@@ -0,0 +1,401 @@
...
+from chainer.links.connection.inception import Inception
+from chainer.links.connection.linear import Linear
+from chainer.serializers import npz
+from chainer.utils import imgproc
+from chainer.variable import Variable
+
+
+class GoogLeNet(link.Chain):
+
+    """A pre-trained GoogLeNet model provided by BVLC [1].
```
ronekko: I appended a draft of the description about GoogLeNet.

chainer/links/model/vision/googlenet.py
```diff
@@ -0,0 +1,401 @@
...
+            it automatically downloads the caffemodel from the internet.
+            Note that in this case the converted chainer model is stored
+            on ``$CHAINER_DATASET_ROOT/pfnet/chainer/models`` directory,
+            where ``$CHAINER_DATASET_ROOT`` is set as
+            ``$HOME/.chainer/dataset`` unless you specify another value
+            as a environment variable. The converted chainer model is
+            automatically used from the second time.
+            If the argument is specified as ``None``, all the parameters
+            are not initialized by the pre-trained model, but the default
+            initializer used in the original paper, i.e.,
```
ronekko: I'm sorry that I'm not careful about the default initialization.
The provider of the caffe model says that the original GoogLeNet used Gaussian for initialization, however I cannot find such description in the original paper.
So I want to have the default initilizer according to the caffe model's one, however it is neither Gaussian nor Glorot.

In [train_val.prototxt](https://github.com/BVLC/caffe/blob/master/models/bvlc_googlenet/train_val.prototxt#L63) of the caffe model, "xavier" with default parameter is used for `weight_filler` .
When "xavier" is used with `variance_norm: AVERAGE` parameter, it means that [the weights are initialized by `Uniform(-s, s), s = sqrt(6 / (fan_in + fan_out))` ](https://github.com/BVLC/caffe/blob/master/include/caffe/filler.hpp#L154), which is the Glorot's initializer.
However, [the default parameter is `variance_norm: FAN_IN` ](https://github.com/BVLC/caffe/blob/master/src/caffe/proto/caffe.proto#L61) and then [the weights are initialized by `s = sqrt(3 / fan_in)` ](https://github.com/BVLC/caffe/blob/master/include/caffe/filler.hpp#L160).

chainer/links/model/vision/googlenet.py
```diff
@@ -0,0 +1,401 @@
...
+            ``chainer.initializers.GlorotUniform(scale=1.0)``.
+
+    Attributes:
+        available_layers (list of str): The list of available layer names
+            used by ``__call__`` and ``extract`` methods.
+
+    """
+
+    image_mean = numpy.array([104, 117, 123], dtype=numpy.float32)  # BGR
```
ronekko: For the first question, I sometimes need the mean values of RGB.
So it is useful if `image_mean` was explicitly defined when someone want to use it.
Although we provide `prepare` function, I think there is a case that we cannot use it.
For example in my case of fine-tuning, I make an HDF5 file of dataset and store images into it with some preprocessing, in order to avoid repetition of the same preprocessing every time the images are loaded.
And in this situation I manually have to subtract the mean RGB from partially preprocessed images.

For the second point I agree with you, but I am not sure which of `[104.0, 117.0, 123.0]` or `[103.939, 116.779, 123.68]` is appropriate. It seems that the former is correct, since in [train_val.prototxt](https://github.com/BVLC/caffe/blob/master/models/bvlc_googlenet/train_val.prototxt#L34-L36) the mean RGB is defined as `[104, 117, 123]` .
I checked the accuracy for both cases on ILSVRC 2012 validation set, however the difference was negligible.
Here is the top-1, top-5 correct rate without oversampling and top-1, top-5 with oversampling for each case.

- `[104.0, 117.0, 123.0]`
    - 0.67376, 0.88168, 0.68622 0.89128
- `[103.939, 116.779, 123.68]`
    - 0.67378, 0.88168, 0.68622 0.89116

chainer/links/model/vision/googlenet.py
```diff
@@ -0,0 +1,401 @@
...
+
+    def __init__(self, pretrained_model='auto'):
+        if pretrained_model:
+            # As a sampling process is time-consuming,
+            # we employ a zero initializer for faster computation.
+            kwargs = {'initialW': constant.Zero()}
+        else:
+            # employ default initializers used in the original paper
+            kwargs = {'initialW': uniform.GlorotUniform(scale=1.0)}
```
ronekko: I investigated closely the initialization of the caffemodel.
It seems that they actually used Uniform(-s, s) with `s = sqrt(3 / fan_in)` .
Since [ `variance_norm` of `weight_filler` is not specified](https://github.com/BVLC/caffe/blob/master/models/bvlc_googlenet/train_val.prototxt#L63) and [the default value of `VrianceNorm` is `FAN_IN` ](https://github.com/BVLC/caffe/blob/master/src/caffe/proto/caffe.proto#L61), then

```
    weight_filler {
      type: "xavier"
    }
```

turns out [s = sqrt(3 / fan_in)](https://github.com/BVLC/caffe/blob/master/include/caffe/filler.hpp#L160) and this is `LeCunUniform(scale=1.0)` in Chainer.

docs/source/reference/links.rst
ronekko: I moved GoogLeNet next to VGG.
(Since the original order of VGG and ResNet can be seen as from weak model to powerful one for me, so I put GoogLeNet first as the weakest model.)

ronekko:
> I have one additional question: did you check whether the output is equal to the original one?

Yes I did if you mean comparison between outpus of chainer's CaffeFunction and this PR's. These outputs are exactly same (i.e. `np.array_equal()` is True).
On the other hand, I did not compare between the outputs of caffe's and this PR's, because I did not have caffe environment.
Alternatively, I tested the accuracy on ILSVRC 2012 validation set by the model of this PR.
The results of top-1 and top-5 correct rate for single center crop (i.e. `oversample=False` of `model.prepare()` ) are 67.4% and 88.2%, which is slightly worse than the [officially reported performance](https://github.com/BVLC/caffe/tree/master/models/bvlc_googlenet) of 68.7% and 88.9%.

ronekko:
Thank you so much for your review.
I updated the code and answered your points.
Could you check them?

ronekko:
I rebased this branch to the master in order to include the latest code.

chainer/links/model/vision/googlenet.py
rezoo: > For example in my case of fine-tuning, I make an HDF5 file of dataset and store images into it with some preprocessing, in order to avoid repetition of the same preprocessing every time the images are loaded.

In this case, to reduce the size and IO access, I think it would be better to store the image with uint8 into hdf5, and write the code as follows:
```python
from chainer.links.model.vision.googlenet import prepare

class Dataset(dataset_mixin):
    def __init__(self, path):
        self.db = h5py.File(path, 'r')
        self.dataset = self.db["image"]

    def get_example(self, i):
        return prepare(self.dataset[i])
```
Although I did not have the opportunity to use image_mean, but I think there is no problem to add `image_mean` to the class variable. However, if you really want to do this, it would be better to add `image_mean` to the other classes like VGG and ResNet, and explicitly describe these variables to the document.

Regarding the average, there is no problem if the original uses discrete values.

chainer/links/model/vision/googlenet.py
rezoo: I understood the situation. At least we cannot say that the same initializer is used in the caffemodel. How about replacing the sentence as follows?

If the argument is specified as ``None``, all the parameters are not initialized by the pre-trained model, but the default initializer denoted as ~

chainer/links/model/vision/googlenet.py
rezoo: Thanks for the investigation. How about using `LeCunUniform`, and describing it to the document?

chainer/links/model/vision/googlenet.py
rezoo: I see.

rezoo:
I finished the review. In my opinion, if you want to add `image_mean` to googlenet, it would be better to add them to the other models in the same manner. (In that case it would be appropriate that we first merge a PR without `image_mean`, and make another PR that adds `image_mean` to all the pre-trained models).

chainer/links/model/vision/googlenet.py
ronekko: I fixed the sentence.

chainer/links/model/vision/googlenet.py
ronekko: I added docs and comments in order to explain why LeCunUniform is used here.

chainer/links/model/vision/googlenet.py
ronekko: > In this case, to reduce the size and IO access, I think it would be better to store the image with uint8 into hdf5

Thank you for sharing your example code. I've been also doing the similar way (although I'm using `fuel` instead of chainer's dataset & iterator) and therefore I cannot subtract beferehand the mean pixel values, which are float value in general. In my observation, when the storage is HDD, IO bottleneck is problematic. However, when I use SSD, CPU bottleneck is dominant (especially jpg decoding and resizing).

Anyway, I do not use `image_mean` this time. As you pointed out, it should be added in another PR.

ronekko:
Thank you. I reflected your points.

chainer/links/model/vision/googlenet.py
rezoo: Good.

docs/source/reference/links.rst
rezoo: I see.

Issue #1963, opened 2016-12-04
should chainer matmul have an equal behavior as numpy matmul?


boeddeker:
Currently chainer has matmul and batch_matmul, but both have a different behavior as numpy:

Here some shapes:
```
matmul: (2, 3),	(3, 4)	-> (2, 4)
matmul: (2,),	(1, 4)	-> (2, 4)  # different from numpy
matmul: (2, 3),	(3,)	-> (2, 1)  # different from numpy
np.matmul: (2, 3), (3, 4) -> (2, 4)
np.matmul: (2,),   (2, 4) -> (4,)
np.matmul: (2, 3), (3,)	  -> (2,)

batch_matmul: (5, 2, 3), (5, 3, 4) -> (5, 2, 4)
batch_matmul: (5, 2, 3), (5, 3)	   -> (5, 2, 1)  # different from numpy
batch_matmul: (5, 3),	 (5, 1, 3) -> (5, 3, 3)  # different from numpy
np.matmul: (5, 2, 3),	(5, 3, 4) -> (5, 2, 4)
np.matmul: (5, 2, 3),	(3, 4)	  -> (5, 2, 4)
np.matmul: (5, 3),	(5, 3, 4) -> (5, 5, 4)
```

In my opinion, the behavior of chainer matmul should be equal to numpy matmul or raise an Exception.

In #1901 @okuta has started to rewrite matmul's internal code and I asked to also rewrite the interface.
Therefore I want to start a discussion, if the interface should be changed.

A possible starting point for a new matmul is the following code:

``` python
def xp_hermitian(x):
    xp = cuda.get_array_module(x)
    return xp.swapaxes(x.conj(), -1, -2)

class MatMul(function.Function):
    def check_type_forward(self, in_types):
        type_check.expect(in_types.size() == 2)
        if not in_types[0].value.dtype in (
                numpy.float32, numpy.complex64, numpy.complex128):
            raise TypeError(in_types[0].value.dtype)
        assert in_types[1].value.dtype in (
            numpy.float32, numpy.complex64, numpy.complex128)
        type_check.expect(
            in_types[0].ndim >= 2,
            in_types[1].ndim >= 2,
            in_types[0].shape[:-2] == in_types[1].shape[:-2],
            in_types[0].shape[-1] == in_types[1].shape[-2],
        )
    def forward(self, x):
        xp = cuda.get_array_module(*x)
        result = utils.force_array(x[0] @ x[1])
        return result,

    def backward(self, x, gy):
[...]

PR #2426, opened 2017-03-18
Numpy like matmul

fukatani:
This PR is proposal for numpy like matmul.
As reffered as #1963, current `matmul` and `batch_matmul` Implementation do different behavior as `numpy.matmul`.

Except for the following two points, `NumpyLikeMatmul` is consistent with `numpy.matmul`.
1. Disallow broadcast as other basic_math function.
2. Arguments `transa` and `transb` exists.


fukatani:
I want to rethink and make this PR WIP.
Sorry for inconvinience.

boeddeker:
The cupy matmul function has the the code line `a = ascontiguousarray(a, dtype=dtype)` and the cuda functions have the options transa and transb. Maybe it is better to make these parameters accessible from outside? Then is the swapaxis no longer required.

fukatani:
Thank you for your comment!

I think `asfortranarray` and `ascontiguousarray` don't affect calculation result.

ex.
```python
import numpy as np

x = np.arange(6).reshape(2,3)
print(x.flags['C_CONTIGUOUS'])  # TRUE
print(x)
y = np.asfortranarray(x)
print(y.flags['C_CONTIGUOUS'])  # FALSE
print(y)

z = np.arange(6).reshape(3,2)
assert (np.matmul(x, z) == np.matmul(y, z)).all()  # PASS this assertion
```

boeddeker:
Yes that is true. What I meant was, that the input to the matmul function may be ccontiguous and the user use the transa or transb option and has the idea, that the code is optimised to do this calculation.
But what happens is, that the variable matmul do the transpose and the matmul use ascontiguous and generate a new array. But in this example it is not necessary to create a new array.

So the idea for my previous comment was: 
   The call to ascontiguousarray may generate a copy and when variable matmul do the transpose the probability that the cupy function generates a copy is higher.

fukatani:
OK.
So you mean call `ascontigiousarray` at `cupy.matmul` and `swapaxes` at `chainer.matmul` are redundant.
That's a good point.

I think that we should measure the impact on performance before starting optimization.
And merge will be smoother if optimization is done by different PR.

fukatani:
Of course, if you are willing to send PR, I'll close this PR. It was originally your issue.

boeddeker:
Yes my thought go in that direction. The call to swapaxis will result in my mind in most cases, that ascontignous do a copy operation. And when the cuda argument is used, it can handel this transpose inside of the dot operation.

But you are right, this could be done in a different PR.

boeddeker:
Currently, I have not the time to write a PR, that is also the reason, why I have not written a PR like this one.

fukatani:
I investigate each matmul behavior.
Here are the result calculated when `a.ndim == b.ndim`.

|method        | 1 dim vs 1 dim  | 2 dim vs 2 dim  | 3 dim vs 3 dim |4 or more|
| --------------- |---------------| -------------------- | --- | -------|
| NumpyLikeMatmul| as numpy.matmul  | as numpy.matmul | as numpy.matmul | as numpy.matmul   |
| Matmul | as numpy.matmul  | as numpy.matmul  | N/A  | N/A  |
| BatchedMatmul | N/A | original behavior (using for loop)   | as numpy.matmul (using for loop)   | N/A  |



Thanks to @boeddeker, we already know each matmul broadcast behavior.

|method        | 1 dim vs 2 dim  | 2 dim vs 3 dim  |
| --------------- |---------------|  -------|
| NumpyLikeMatmul| N/A  | N/A  |
| Matmul | original behavior  | original behavior  | 
| BatchedMatmul | original behavior | original behavior  |



Here is behavior when `transa == True`.

|method        | 1 dim(vector)          | 2 dim                 | 3 dim|4 or more|
| --------------- |---------------| -------------------- | --- | -------|
| NumpyLikeMatmul| a (do nothing) | a.T | a.swapaxes(-1, -2)  |a.swapaxes(-1, -2)  |
| Matmul | asmat(a).T | a.T | N/A  | N/A  |
| BatchedMatmul | N/A | a.reshape(a.shape[:2] + (-1,)).swapaxes(-1, -2)   | a.swapaxes(-1, -2)   | N/A  |


In my opinion, Calculations desired by most users can be covered only with `NumyLikeMatmul`, so it is better to replace in v2.

Although some calculations such as those involving broadcasts can not be done alone on `NumpyLikeMatmul`, they can be reproduced by doing explicit array manipulation.
Not to do implicit broadcasting is chainer-like-API.

tests/chainer_tests/functions_tests/math_tests/test_matmul.py
```diff
@@ -272,4 +280,99 @@ def setUp(self):
...
+class TestNumpyLikeMatMulInvalid(unittest.TestCase):
+
+    def test_invalid_shape(self):
+        x_data = numpy.zeros((2, 3, 4), dtype=numpy.float32)
+        y_data = numpy.zeros((1, 4, 3), dtype=numpy.float32)
+        x = chainer.Variable(x_data)
+        y = chainer.Variable(y_data)
+
+        with self.assertRaises(type_check.InvalidType):
+            F.matmul(x, y)
```
niboshi: Did you mean `F.numpy_like_matmul`?

tests/chainer_tests/functions_tests/math_tests/test_matmul.py
```diff
@@ -272,4 +280,99 @@ def setUp(self):
...
+
+    def test_invalid_ndim(self):
+        x_data = numpy.zeros((3, 2, 5), dtype=numpy.float32)
+        y_data = numpy.zeros((3, 5), dtype=numpy.float32)
+        x = chainer.Variable(x_data)
+        y = chainer.Variable(y_data)
+
+        with self.assertRaises(type_check.InvalidType):
+            F.matmul(x, y)
```
niboshi: Ditto. (Did you mean `F.numpy_like_matmul`?)

fukatani:
Thanks! I fixed test.

niboshi:
I'm sorry for being late.
As v2 has already been released, and we have revised the release policy so that we can make more frequent compatibility-breaking releases, it's more straight-forward to include this PR directly into the next alpha release. (Please see [Contribution Guide](https://docs.chainer.org/en/stable/contribution.html) and [Compatibility Policy](https://docs.chainer.org/en/stable/compatibility.html) for details)

So could you rename function/links in this PR to override the existing ones?

fukatani:
Sure, I overrided existing methods.

fukatani:
I fixed test. But part of appveyor test failed. Could you help me?

niboshi:
Thank you for the fix! I will review again.

>I fixed test. But part of appveyor test failed. Could you help me?

It's not relevant to this PR. So please don't care.
It should be fixed by #2855, but it's not merged yet.

chainer/functions/math/matmul.py
```diff
@@ -145,81 +168,23 @@ def matmul(a, b, transa=False, transb=False):
         a (Variable): The left operand of the matrix multiplication.
             A 1-D array of shape ``(N,)`` is considered as an
             :math:`N \\times 1` matrix.
```
niboshi: Can you describe the behavior of ndim>2 case?

chainer/functions/math/matmul.py
```diff
@@ -145,81 +168,23 @@ def matmul(a, b, transa=False, transb=False):
...
-            A 2-D array of shape ``(M, N)`` is considered as an
-            :math:`M \\times N` matrix.
         b (Variable): The right operand of the matrix multiplication.
             Its array is treated as a matrix in the same way as ``a``'s array.
         transa (bool): If ``True``, transpose ``a``.
+            If ``a.ndim == 1``, do nothing.
```
niboshi: I think this description is not precise.
How about this?

> If ``True``, each matrices in ``a`` will be transposed.

chainer/functions/math/matmul.py
```diff
@@ -145,81 +168,23 @@ def matmul(a, b, transa=False, transb=False):
...
         transb (bool): If ``True``, transpose ``b``.
+            If ``b.ndim == 1``, do nothing.
```
niboshi: Ditto

tests/chainer_tests/functions_tests/math_tests/test_matmul.py
```diff
@@ -46,230 +53,99 @@ def test_matmul_backward_gpu(self):
...
+            x1 = self.x1.swapaxes(-1, -2) if self.transa else self.x1
+
+        if self.x2.ndim == 1:
+            x2 = self.x2
+        else:
+            x2 = self.x2.swapaxes(-1, -2) if self.transb else self.x2
+        if x1.ndim <= 2:
+            self.forward_answer = numpy.dot(x1, x2)
+        else:
+            self.forward_answer = numpy.einsum('...ij,...jk->...ik', x1, x2)
```
niboshi: Please factor out from line 115 (`if self.x1.ndim == 1:`) to line 127 (`self.forward_answer = numpy.einsum('...ij,...jk->...ik', x1, x2)`) as a method, something like:
```
def _get_forward_answer(self, x1, x2, transa, transb):
    ...
```

fukatani:
Thank you for your attentive review.
How about this version?

tests/chainer_tests/functions_tests/math_tests/test_matmul.py
```diff
@@ -10,24 +9,44 @@
 from chainer import testing
 from chainer.testing import attr
 from chainer.testing import condition
+from chainer.utils import type_check
 
 
 class _TestMatMul(unittest.TestCase):
```
niboshi: `_TestMatMul` is no longer needed, because it's subclassed only by `TestMatMul`. How about merging them?

tests/chainer_tests/functions_tests/math_tests/test_matmul.py
```diff
@@ -10,24 +9,44 @@
...
 
-    def check_forward(self, x1_data, x2_data):
+    def _get_forward_answer(self, x1, x2, transa, transb):
+        if x1.ndim >= 2:
+            x1 = x1.swapaxes(-1, -2) if transa else x1
```
niboshi: You can write `if transa and x1.ndim >= 2: ...`

tests/chainer_tests/functions_tests/math_tests/test_matmul.py
```diff
@@ -10,24 +9,44 @@
...
+
+        if x2.ndim >= 2:
+            x2 = x2.swapaxes(-1, -2) if transb else x2
```
niboshi: Ditto

fukatani:
You are absolutely right.
I fixed them.

niboshi:
LGTM! Thank you very much for a big contribution!

chainer/functions/__init__.py
```diff
@@ -204,8 +204,8 @@
 from chainer.functions.math.logarithm_1p import log1p  # NOQA
 from chainer.functions.math.logsumexp import logsumexp  # NOQA
 from chainer.functions.math.logsumexp import LogSumExp  # NOQA
-from chainer.functions.math.matmul import batch_matmul  # NOQA
-from chainer.functions.math.matmul import BatchMatMul  # NOQA
+from chainer.functions.math.matmul import matmul  # NOQA
```
okuta: This line and L209 are same. Is it mistake?

okuta:
Can we provide `batch_matmul` as deprecated function in version 3. We can use `FutureWarning`.

delta2323:
I agree with @okuta in that we restore `batch_matmul` and declare deprecation.

fukatani:
So what should I do?
This PR will be revirted or create new PR to reimplement `batch_matmul` with deprecation?

PR #2431, opened 2017-03-18
Support numpy like to-string function for variable.

fukatani:
This PR breaks backward compatibility.

Numpy like repr function is useful for console debugging.

ex.
```
>>> a = numpy.array([1., 2.])
>>> a
array([ 1.,  2.])
>>> v = Variable(a)
>>> v
variable([ 1.,  2.])
>>> v = Variable(a, name='x')
>>> v
variable x([ 1.,  2.])
```

I also implemented __str__ function.
ex.
```
>>> a = numpy.array([1., 2.])
>>> print(a)
[ 1.  2.]
>>> v = Variable(a)
>>> print(v)
variable([ 1.  2.])
>>> v = Variable(a, name='x')
>>> print(v)
variable x([ 1.  2.])
```


mitmul:
Thank you for sending this PR! It's useful PR and LGTM, but I'm thinking how we should deal with the backward compatibility break. Then, how about merging this to Chainer v2?
@beam2d,  Do you have any comments?

fukatani:
Merging to v2 is acceptable for me.

beam2d:
I think it's better to merge it to v2. IMHO, how a variable is printed can be treated as an implementation detail, but it does not mean this change won't break any existing code.

fukatani:
Thank you for your comments.
I close this PR and reopen to send merge request to _v2.

PR #2442, opened 2017-03-21
Suppress feature warnings in cupy

okuta:
rel #2406


okuta:
We have another method. See also https://github.com/pfnet/chainer-test/pull/209/files

PR #2454, opened 2017-03-23
Add type check to functions/connection/*

soramichi:
Continued from #2255.

Ths PR adds type checking codes to `functions/connection/*.py` to see if numpy and cupy are mixed (and outputs a graceful warning when they are).

I confirmed that all the related tests with @attr.gpu pass in my local environment.

### DIFF
--- a/chainer/functions/connection/bilinear.py
+++ b/chainer/functions/connection/bilinear.py
@@ -51,10 +51,10 @@ class BilinearFunction(function.Function):
         e2 = array.as_mat(inputs[1])
         W = inputs[2]
 
-        if(not type_check.same_types(*inputs)):
+        if not type_check.same_types(*inputs):
             raise ValueError('numpy and cupy must not be used together\n'
-                             'type(W): %s, type(e1): %s, type(e2): %s'
-                             % (type(W), type(e1), type(e2)))
+                             'type(W): {0}, type(e1): {1}, type(e2): {2}'
+                             .format(type(W), type(e1), type(e2)))
 
         xp = cuda.get_array_module(*inputs)
         if xp is numpy:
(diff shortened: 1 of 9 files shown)


chainer/functions/connection/bilinear.py
```diff
@@ -51,6 +51,11 @@ def forward(self, inputs):
         e2 = array.as_mat(inputs[1])
         W = inputs[2]
 
+        if(not type_check.same_types(*inputs)):
+            raise ValueError('numpy and cupy must not be used together\n'
+                             'type(W): %s, type(e1): %s, type(e2): %s'
+                             % (type(W), type(e1), type(e2)))
```
niboshi: Formatting with `%` operator is discouraged. Could you use `str.format()` instead?

soramichi:
Hi, thank you for the review.

I fixed the coding style as suggested, applied the same fix to `embed_id.py` (which had been already merged in #2255), and inserted the type checking codes to `dilated_convolution_2d.py` that I missed previously.

chainer/functions/connection/bilinear.py
```diff
@@ -77,6 +82,12 @@ def backward(self, inputs, grad_outputs):
         e1 = array.as_mat(inputs[0])
         e2 = array.as_mat(inputs[1])
         W = inputs[2]
+
+        if(not type_check.same_types(*inputs)):
+            raise ValueError('numpy and cupy must not be used together\n'
+                             'type(W): %s, type(e1): %s, type(e2): %s'
+                             % (type(W), type(e1), type(e2)))
+
```
niboshi: Please fix this section as well.

soramichi:
oops, sorry for the mistake. fixed.

niboshi:
LGTM. Thank you for your contribution!

PR #2455, opened 2017-03-23
Support numpy like repr function

fukatani:
Reopen #2431 .
This PR breaks backward compatibility.

I think that numpy like repr function is useful for console debugging.

ex.
```
>>> a = numpy.array([1., 2.])
>>> a
array([ 1.,  2.])
>>> v = Variable(a)
>>> v
variable([ 1.,  2.])
>>> v = Variable(a, name='x')
>>> v
variable x([ 1.,  2.])
```

I also implemented __str__ function.
ex.
```
>>> a = numpy.array([1., 2.])
>>> print(a)
[ 1.  2.]
>>> v = Variable(a)
>>> print(v)
variable([ 1.  2.])
>>> v = Variable(a, name='x')
>>> print(v)
variable x([ 1.  2.])
```


chainer/variable.py
```diff
@@ -52,6 +52,52 @@ def make_message(message):
...
+        if var.name:
+            prefix = 'variable ' + var.name + '('
+        else:
+            prefix = 'variable('
+        lst = numpy.array2string(arr, None, None, None, ', ', prefix)
+    else:  # show zero-length shape unless it is (0,)
+        lst = '[], shape=%s' % (repr(arr.shape),)
+        if var.name:
+            lst = var.name + lst
+    return '%s(%s)' % (prefix[:-1], lst)
```
niboshi: `prefix` will be undefined if `shape == (0,0)` etc.
I think how `prefix` is determined should be independent from `shape`.

Also, the operation `prefix[:-1]` seems redundant for me. 

In summary, how about the following logic?

```
    if var.name:
        prefix = 'variable ' + var.name
    else:
        prefix = 'variable'

    if arr.size > 0 or arr.shape == (0,):
        lst = numpy.array2string(arr, None, None, None, ', ', prefix + '(')
    else:  # show zero-length shape unless it is (0,)
        lst = '[], shape=%s' % (repr(arr.shape),)

    return '%s(%s)' % (prefix, lst)
```

niboshi: Or, as for `prefix[:-1]`, it could also be reasonable to leave `'('` in `prefix` and `return prefix + lst + ')'`.
In this way it will be similar to the formatting method in `variable_str()`.

fukatani:
Thanks! Your proposal is a nice implementation.
And I added testcase when  `shape==(0, 0)`,

fukatani:
Only at Python 2.7 in Windows, shape is displayed in the form of `shape(0L, 0L)`.
Is it better to branch test by `os.platform`?

And I don't understand why travis failed. Could you help me?

niboshi:
> Only at Python 2.7 in Windows, shape is displayed in the form of `shape(0L, 0L)`.
> Is it better to branch test by `os.platform`?

Yes, please do that.
As for travis failure, I think you should have double blank lines instead of just one (at line 1131 of test_variable.py).

fukatani:
Thanks for your advise.

I don't know the reason why longint is used, and cupy may have different results.

fukatani:
Thank you for your review.
I fixed test name inappropriate test names.

Sorry for inconvinience.

niboshi:
Thank you for the fix!
LGTM!

PR #2458, opened 2017-03-25
Add spatial_transformer_grid

yuyu2172:
Merge after #2272.

This PR adds a function that generates sampling points described in https://arxiv.org/abs/1506.02025.

### DIFF
--- a/chainer/functions/array/spatial_transformer_grid.py
+++ b/chainer/functions/array/spatial_transformer_grid.py
@@ -1,0 +1,160 @@
+import numpy
+
+from chainer import cuda
+from chainer import function
+from chainer.utils import type_check
+
+
+if cuda.cudnn_enabled:
+    cudnn = cuda.cudnn
+    libcudnn = cudnn.cudnn
+    _cudnn_version = libcudnn.getVersion()
+    _sampler_type = libcudnn.CUDNN_SAMPLER_BILINEAR
+
+
+class SpatialTransformerGrid(function.Function):
+
+    def __init__(self, output_shape, use_cudnn=True):
+        self.output_shape = output_shape
+        self.use_cudnn = use_cudnn
+
+    def check_type_forward(self, in_types):
+        n_in = in_types.size()
+        type_check.expect(n_in == 1)
+
+        theta_type = in_types[0]
+        type_check.expect(
+            theta_type.dtype.kind == 'f',
+            theta_type.ndim == 3,
+            theta_type.shape[1] == 2,
+            theta_type.shape[2] == 3,
+        )
+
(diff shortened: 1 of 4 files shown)


mattya:
Thank you for the PR.
I think this function is assumed to be used with `spatial_transformer_sampler`.
Current `spatial_transformer_sampler` assumes the grid value is inside of the range [-1, 1] 
(and it seems `spatial_transformer_sampler` clip the value when cudnn=False while it doesn't clip when cudnn=True),
but this function may return outside of the range [-1, 1].

yuyu2172:
@mattya 

Thanks for commenting. Do you think it is ok to clip output of `spatial_transformer_grid` to range [-1, 1]?

Speaking of the difference between cuDNN and non-cuDNN implementations of `spatial_transformer_sampler`, the specification of cuDNN is not well documented in the case when grid is outside of range [-1, 1].
I am guessing that the images are padded by zeros outside of the image, but I haven't confirmed it.
I think so because the sampler returns non zero values when a grid is slightly outside of the range, but returns zeros when the grid is sufficiently exceeding the limit.

yuyu2172:
I will check if I can fix `spatial_transformer_sampler` to fix the discrepancy between cuDNN and non cuDNN.

yuyu2172:
@mattya 
I confirmed that indeed the cuDNN implementation samples points from a zero padded image.

I fixed `spatial_transformer_sampler` in #2648.

mattya:
Thank you for fixing.
I've checked the consistency between cudnn and non-cudnn mode.

chainer/functions/array/spatial_transformer_grid.py
```diff
@@ -0,0 +1,164 @@
...
+        use_cudnn (bool): If ``True``, then this function uses cuDNN if
+            available. Note that, cuDNN supports SpatialTransformerGrid
+            from version 5.0.0.
+
+    Returns:
+        ~chainer.Variable:  A variable of shape :math:`(n, 2, h_O, w_O)`.
+        In the 2nd dimension, the first element is the coordinate along the
+        x axis, and the second element is the coordinate along the y axis.
+        All the coordinates are scaled to fit range :math:`[-1, 1]`.
+        This means that the coordinate :math:`(-1, 1)` corresponds to
```
mattya: Is it (-1, -1) ?

mattya: Is it (-1, -1) ?

yuyu2172:
Thanks for the review.
I fixed the documentation.

mattya:
LGTM!

PR #2460, opened 2017-03-26
Improve docs of lstm, slstm

keisuke-umezawa:

Modified functions/links:
 - functions/activation/lstm
 - functions/activation/slstm


keisuke-umezawa:
Jenkins test has also passed.
https://ci-chainer.pfidev.jp/job/chainer-sakura/1160/

mitmul:
@unnonouno @soskek @kiyukuta Could anyone check these docmentation updates especially the new example code for `slstm`? All tests have been passed. If the examples are enough informative and correct, I'd like to merge this PR.

soskek:
I read the docs. I noticed two issues.
1. Both for lstm and slstm, I feel strange that calling input arrays with shape `4 * n_units` as "incoming signals" or "input signals". In the essential formulation of LSTM, "input" is an array before linear transformation into `4 * n_units` from `x_{t}` rather than both `x_{t}` and `h_{t-1}`, such as input of `L.LSTM`, e.g., a word vector or output from lower RNN layer.  However, array with shape `4 * n_units` might have no name in related papers unfortunately...  
So, it looks good to modify the example in doc more informative for disambiguation. In the current example, the formal "input" array (`y` in the example) is called "current input signal" and it is also confusing compared to the prior names. Could their names be modified consistently or less confusing?
2. Second issue does not obligate to this PR. Current implementation of slstm seems different from formulation in the cited paper and little similar to other TreeLSTM model (https://arxiv.org/pdf/1503.00075.pdf) but also different from it. If this is also confirmed by others, it looks good to fix or re-implement as TreeLSTM, which is more cited and popular in NLP, toward Chainer v2. What do you (and other reviewers) think?

chainer/functions/activation/slstm.py
```diff
@@ -218,23 +218,58 @@ def slstm(c_prev1, c_prev2, x1, x2):
...
+        Each of ``c1``, ``c2``, ``h1`` and ``h2`` has ``n_units`` channels.
+        Most typical preparation of ``x1``, ``x2`` is:
+
+        >>> n_units = 100
+        >>> h1 = chainer.Variable(np.zeros((1, n_units), 'f'))
+        >>> h2 = chainer.Variable(np.zeros((1, n_units), 'f'))
+        >>> h = chainer.Variable(np.zeros((1, n_units), 'f'))
+        >>> c1 = chainer.Variable(np.zeros((1, n_units), 'f'))
+        >>> c2 = chainer.Variable(np.zeros((1, n_units), 'f'))
+        >>> c = chainer.Variable(np.zeros((1, n_units), 'f'))
```
soskek: Defining `h` and `c` (L256, L259) is not effected below the lines. They can be removed.

keisuke-umezawa: Thank you. I will remove them.

keisuke-umezawa:
> Both for lstm and slstm, I feel strange that calling input arrays with shape 4 * n_units as "incoming signals" or "input signals". In the essential formulation of LSTM, "input" is an array (n_units) before linear transformation into 4 * n_units from x_{t} and excluding h_{t-1}'s information, such as input of L.LSTM, e.g., a word vector or output from lower RNN layer. However, pre-activation array with shape 4 * n_units might have no name in related papers unfortunately...
> So, it looks good to modify the example in doc more informative for disambiguation. In the current example, the formal "input" array (y in the example) is called "current input signal" and it is also confusing compared to the prior names. Could their names be modified consistently or less confusing?

How about using "input array" for x, "incoming signal" for y and "outgoing signal" for h.
I will fix the comment to improve consistency.

keisuke-umezawa:
> Second issue does not obligate to this PR. Current implementation of slstm seems different from formulation in the cited paper (https://arxiv.org/pdf/1503.04881.pdf) and little similar to other TreeLSTM model (https://arxiv.org/pdf/1503.00075.pdf) but also different from it. (I skipped the details here, but I can describe them if needed.) If this is also confirmed by others, it looks good to fix or re-implement as TreeLSTM, which is more cited and popular in NLP, toward Chainer v2. What do you (and other reviewers) think?

@others
How about this? If needed, I will make issue about this comment. And if possible, I will make the another PR for the issue.

soskek:
>How about using "input array" for x, "incoming signal" for y and "outgoing signal" for h. I will fix the comment to improve consistency.
Looks good to me! This issue may have no best solution, so I would like another reviewer to agree to these names just in case.

kiyukuta:
>How about using "input array" for x, "incoming signal" for y and "outgoing signal" for h. I will fix the comment to improve consistency.

Also lgtm!

In addition, for `lstm`, I think the difference of input between `functions.lstm` (this function) and `links.lstm` is a little confusing.
The former receives `input array`, The latter receives `incoming signal`.
How about adding a notation about the difference ? 

How do you (or other reviewers) think ? 


>to others,
How about this? If needed, I will make issue about this comment. And if possible, I will make the another PR for the issue.

Yes, let's separate it and make another issue or PR!

soskek:
>How about adding a notation about the difference ?
I agree to this whether it will be done in this PR or a new one.

keisuke-umezawa:
Ok. I will add the difference in this PR.

soskek:
Thanks!
As you may know in #2592 #2606 , I implemented new TreeLSTM link for v2 and we will make slstm function deprecated until v2. So, I think the modification for slstm can be minimum.

keisuke-umezawa:
Thank you for the information. So, I will add the difference note for lstm, not for slstm.

keisuke-umezawa:
I added the notation of difference.

chainer/functions/activation/lstm.py
```diff
@@ -220,9 +228,24 @@ def lstm(c_prev, x):
...
+        the previous outgoing signal ``h``. Different parameters are used for
+        different kind of input sources.
+
+    .. note::
+
+        We use the naming rule below.
+
+        - incoming signal
+            The formal input of the formulation of LSTM (e.g. in NLP, word
+            vector or output of previous RNN). The input of
```
soskek: "output of previous RNN" is little confusing and seems to represent "previous h of the unit itself". Could you modify it to "output of lower RNN layer" or something like that?

chainer/functions/activation/lstm.py
```diff
@@ -220,9 +228,24 @@ def lstm(c_prev, x):
...
+            :class:`chainer.links.LSTM` is the *incoming signal*.
+        - input array
+            The array which is linear transformed from *incoming signal*. The
```
soskek: I think `input array = (incoming signal).dot(W_{upward}) + (previous h).dot(W_{literal}) + bias`.
So, "from \*incoming signal\* and the previous outgoing signal." can be more suitable.

chainer/functions/activation/lstm.py
```diff
@@ -157,28 +157,29 @@ def lstm(c_prev, x):
...
-    cell state :math:`c_{\\text{prev}}` and the incoming signal :math:`x`.
+    cell state ``c_prev`` and the input array ``x``.
 
-    First, the incoming signal :math:`x` is split into four arrays
-    :math:`a, i, f, o` of the same shapes along the second axis.
-    It means that :math:`x` 's second axis must have 4 times the length of
-    :math:`c_{\\text{prev}}`.
+    First, the input array ``x`` is split into four arrays
+    :math:`a, i, f, o` of the same shapes along the second axis. It means that
+    ``x`` 's second axis must have 4 times the length of ``c_prev``.
```
soskek: Looks good.

soskek:
Thank you! LGTM!

keisuke-umezawa:
Who should merge this PR?

PR #2461, opened 2017-03-26
Reduce test time

okuta:
This PR reduce 30% test time.

### DIFF
--- a/tests/chainer_tests/functions_tests/connection_tests/test_convolution_nd.py
+++ b/tests/chainer_tests/functions_tests/connection_tests/test_convolution_nd.py
@@ -17,7 +17,7 @@ from chainer.utils import conv
 
 
 @testing.parameterize(*(testing.product({
-    'dims': [(6,), (5, 4), (4, 3, 3)],
+    'dims': [(5,), (4, 3), (3, 4, 3)],
     'cover_all': [True, False],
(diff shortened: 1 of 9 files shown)


gwtnb:
LGTM

tests/chainer_tests/functions_tests/connection_tests/test_convolution_nd.py
```diff
@@ -17,15 +17,15 @@
...
+    'dims': [(5,), (4, 3), (3, 4, 3)],
     'cover_all': [True, False],
     'c_contiguous': [True],
     'x_dtype': [numpy.float32],
     'W_dtype': [numpy.float32],
 }) + testing.product({
     'dims': [(4,)],
     'cover_all': [False],
-    'c_contiguous': [True, False],
+    'c_contiguous': [False],
```
delta2323: Why we can remove `c_contiguous=True` case?

okuta: Because `c_contiguous=True` is easy case. And, above test case includes it.

delta2323: ok, got it. thanks.

PR #2462, opened 2017-03-26
Fix strict option in deserializer to support recursive calls

rezoo:
Current `strict` options in `NpzDeserializer` and `HDF5Deserializer` work well when the link does not have a hierarchical structure, however, it automatically changes `strict` option to `True` after the 2nd level even if users explicitly specify `False`. This PR aims to fix this problem.

### DIFF
--- a/chainer/serializers/hdf5.py
+++ b/chainer/serializers/hdf5.py
@@ -95,7 +95,8 @@ class HDF5Deserializer(serializer.Deserializer):
 
     def __getitem__(self, key):
         name = self.group.name + '/' + key
-        return HDF5Deserializer(self.group.require_group(name))
+        return HDF5Deserializer(
+            self.group.require_group(name), strict=self.strict)
 
     def __call__(self, key, value):
         if not self.strict and key not in self.group:
--- a/chainer/serializers/hdf5.py
+++ b/chainer/serializers/hdf5.py
@@ -97,11 +97,11 @@ class HDF5Deserializer(serializer.Deserializer):
         name = self.group.name + '/' + key
         try:
             group = self.group.require_group(name)
-            return HDF5Deserializer(group, strict=self.strict)
         except ValueError:
             # require_group raises ValueError if there does not exist
             # the given group and the file is read mode.
-            return HDF5Deserializer(None, strict=self.strict)
+            group = None
+        return HDF5Deserializer(group, strict=self.strict)
 
     def __call__(self, key, value):
         if self.group is None:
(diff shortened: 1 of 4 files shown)


okuta:
Can you write test?

rezoo:
@okuta I added several tests. Can you check them?

chainer/serializers/hdf5.py
```diff
@@ -95,13 +95,21 @@ def __init__(self, group, strict=True):
 
     def __getitem__(self, key):
         name = self.group.name + '/' + key
-        return HDF5Deserializer(self.group.require_group(name))
+        try:
```
okuta: How about following code to minimize try section?
```python
try:
    group = self.group.require_group(name)
except ValueError:
    group = None
return HDF5Deserializer(group, strict=self.strict)
```

rezoo: Thanks! I fixed.

tests/chainer_tests/serializers_tests/test_hdf5.py
```diff
@@ -178,9 +181,20 @@ def tearDown(self):
...
-        y = numpy.empty((2, 3), dtype=numpy.float32)
-        ret = self.deserializer('y', y)
-        self.assertIs(ret, y)
+        child = link.Chain(linear2=links.Linear(2, 3))
+        target = link.Chain(linear=links.Linear(3, 2), child=child)
+        target_child_W = numpy.copy(child.linear2.W.data)
+        target_child_b = numpy.copy(child.linear2.b.data)
+        self.deserializer.load(target)
+
+        self.assertTrue(numpy.array_equal(
```
rezoo: Fixed.

tests/chainer_tests/serializers_tests/test_npz.py
```diff
@@ -163,9 +165,20 @@ def tearDown(self):
...
-        y = numpy.empty((2, 3), dtype=numpy.float32)
-        ret = self.deserializer('y', y)
-        self.assertIs(ret, y)
+        child = link.Chain(linear2=links.Linear(2, 3))
+        target = link.Chain(linear=links.Linear(3, 2), child=child)
+        target_child_W = numpy.copy(child.linear2.W.data)
+        target_child_b = numpy.copy(child.linear2.b.data)
+        self.deserializer.load(target)
+
+        self.assertTrue(numpy.array_equal(
```
rezoo: Fixed.

okuta:
LGTM!

PR #2471, opened 2017-03-27
Fix typo in docstring of elu function

ronekko:
I removed an unnecessary line in the example code.
http://docs.chainer.org/en/latest/reference/functions.html#chainer.functions.elu


delta2323:
LGTM

delta2323:
Thank you very much!

PR #2472, opened 2017-03-28
fix mocked trainer

Hakuyume:
Currently, the design of updater used in `_get_mocked_trainer` is wrong.
- `epoch_detail` before the first iteration is `1`.
    It should be `0` because no samples are processed before the first iteration.
-  `epoch_detail` does not change during iterations.
    This means batchsize is zero or the size of dataset is infinity. However, some test cases use epoch-base triggers.


delta2323:
Thank you. I am running GPU test and will merge it if all tests pass.

delta2323:
I confirmed all GPU tests passed. I will merge this PR after tomorrow's release.

delta2323:

PR #2476, opened 2017-03-28
Fix Windows NVTX support.

takagi:
This PR follows #2423, which is already merged to the master.

In #2423, reference to nvToolsExt.h is alive even when NVTX is not found on a Windows environment and it prevents `cupy.cuda` module from being built when `check_library` function is called.

This PR removes the reference to nvToolsExt.h in such a case.


okuta:
LGTM!

PR #2482, opened 2017-03-31
Fix an exception handling problem during importing matplotlib

t-abe:
related to https://github.com/pfnet/chainer/issues/2146

`import matplotlib` sometimes throws `TypeError`. So `TypeError` also should be caught.


mitmul:
LGTM

PR #2484, opened 2017-03-31
interval_trigger using previous_epoch_detail

### LINKED ISSUE #2244: IntervalTrigger may become True at the improper timing when resuming learning

Hakuyume:
Fix #2244 
I modified `interval_trigger` to use `previous_epoch_detail`. This PR is a refactored version of #2258. Previous discussions are in #2258.

This PR is based on #2472. I will rebase this PR after #2472 is merged.


tests/chainer_tests/training_tests/triggers_tests/test_interval_trigger.py
```diff
@@ -41,6 +47,13 @@ def _test_trigger(self, updater, trigger, expecteds):
         self.assertEqual(trigger(trainer), expected)
 
 
+def _test_trigger_stop(self, updater, trigger, expecteds):
```
delta2323: If I understand correctly, this function is for stop triggers. If so, could you change the function name to make its role clearer.

delta2323:
Is it possible to write a test that the original bug #2244 is fixed. I mean `IntervalTrigger` works well even it is serialized and restored.

tests/chainer_tests/training_tests/triggers_tests/test_interval_trigger.py
Hakuyume: I renamed this function.

tests/chainer_tests/training_tests/triggers_tests/test_interval_trigger.py
```diff
@@ -76,5 +107,137 @@ def test_unaligned_epoch_interval_trigger(self):
...
+        _test_stop_trigger(self, updater, trigger, expected)
+
+
+class TestResumedEpochIntervalTrigger(unittest.TestCase):
+
+    def test_resumed_epoch_interval_trigger(self):
+        updater = DummyUpdater(iters_per_epoch=1, initial_iteration=4)
+        trigger = training.trigger.IntervalTrigger(3, 'epoch')
+        expected = [False, True, False, False, True, False]
+        _test_trigger(self, updater, trigger, expected)
```
Hakuyume: This test corresponds to #2244

Hakuyume:
I modified one of tests to fit to #2244.

tests/chainer_tests/training_tests/triggers_tests/test_interval_trigger.py
delta2323: I see. I meant the test in which we do serialize the trigger to the temporary file, deserialize it and confirm it works correctly. Is it possible to write such a test?

Hakuyume: OK. Can I use `StandardUpdater` in serialization tests? Is it better to add serialization method to `DummyUpdater`?

delta2323: Yes, it is OK to use `StandardUpdater`.

Hakuyume: I add serialization method (it is more simple than using `StandardUpdater`).

Hakuyume:
I updated serialization tests to use `load_npz` and `save_npz`.

chainer/training/triggers/interval.py
```diff
@@ -23,7 +23,6 @@ def __init__(self, period, unit):
         self.period = period
         assert unit == 'epoch' or unit == 'iteration'
         self.unit = unit
-        self.count = 0
```
delta2323: Although `count` is not a documented attribute of `IntervalTrigger`, some users could use this attribute. If it is not heavy, could you restore this attribute with a comment that it is for keeping backward compatibility.

tests/chainer_tests/training_tests/triggers_tests/test_interval_trigger.py
```diff
@@ -30,51 +33,81 @@ def epoch_detail(self):
...
+        'expected': [False, False, True, False, False, True, False]},
+    # unaligned epoch
+    {
+        'iters_per_epoch': 2.5, 'interval': (1, 'epoch'), 'resume': 3,
+        'expected': [False, False, True, False, True, False, False]},
+    # tiny epoch
+    {
+        'iters_per_epoch': 0.5, 'interval': (1, 'epoch'), 'resume': 4,
+        'expected': [True, True, True, True, True, True, True]},
+]))
```
delta2323: Using `testing.parameterize` simplifies and strengthens the test. Great!

tests/chainer_tests/training_tests/triggers_tests/test_interval_trigger.py
```diff
@@ -30,51 +33,81 @@ def epoch_detail(self):
...
+
+    def test_stop_trigger(self):
+        trigger = training.trigger.IntervalTrigger(*self.interval)
+        updater = DummyUpdater(self.iters_per_epoch)
+        trainer = training.Trainer(updater)
+        for expected in [False] + self.expected:
+            self.assertEqual(trigger(trainer), expected)
+            if expected:
+                break
+            updater.update()
```
delta2323: It seems to me what `test_stop_trigger` checks is essentially same as `test_trigger` does.

`test_trigger`
```
[update] [assert(expected[0])] | [update] [assert(expected[1])] | [update] ....
```

`test_stop_trigger`
```
[assert(False)] [update] | [assert(expected[0])] [update] | [assert(expected[1])] ....
```

tests/chainer_tests/training_tests/triggers_tests/test_interval_trigger.py
```diff
@@ -30,51 +33,81 @@ def epoch_detail(self):
...
+
+    def test_resumed_trigger(self):
+        temp_dir = tempfile.mkdtemp()
+        temp_file = os.path.join(temp_dir, 'temp.npz')
```
delta2323: I think [`NamedTemporaryFile`](http://docs.python.jp/2/library/tempfile.html#tempfile.NamedTemporaryFile) is simpler, if it is allowed that the temporary file does not have `npz` suffix, because you do not have to create a temporary directory and the temporary file is automatically removed.

tests/chainer_tests/training_tests/triggers_tests/test_interval_trigger.py
Hakuyume: Thank you. I am planning to update the tests for `ManualScheduleTrigger` using this.

chainer/training/triggers/interval.py
Hakuyume: OK. I will restore it.

tests/chainer_tests/training_tests/triggers_tests/test_interval_trigger.py
Hakuyume: `test_stop_trigger` calls the trigger before the first iteration. It tests the case of `previous_epoch_detail` is `None`. As you pointed out, those two test can be merged. I will remove `test_trigger` and rename `test_stop_trigger` -> `test_trigger`

tests/chainer_tests/training_tests/triggers_tests/test_interval_trigger.py
Hakuyume: I will try to use it.

Hakuyume: According to the doc of `NamedTemporaryFile`, this file cannot be open twice under Windows environment. Windows tests failed with `NamedTemporaryFile`.

Hakuyume: I'm sorry. It worked with `delete=False`.

delta2323:
LGTM, I am running GPU tests now.

tests/chainer_tests/training_tests/triggers_tests/test_interval_trigger.py
Hakuyume: I made another PR for `ManualScheduleTrigger` (#2557).

delta2323:
Could you fix [`ParallelSequentialIterator`](https://github.com/pfnet/chainer/blob/master/examples/ptb/train_ptb.py#L51) in the PTB example to add `previous_epoch_detail` attribute?

Here is a part of the error log that we ran in our CI server.

```
17:11:09 + coverage run -a --branch examples/ptb/train_ptb.py --epoch=1 --unit=10 --test
17:11:14 Downloading from https://raw.githubusercontent.com/tomsercu/lstm/master/data/ptb.train.txt...
17:11:14 Downloading from https://raw.githubusercontent.com/tomsercu/lstm/master/data/ptb.valid.txt...
17:11:14 Downloading from https://raw.githubusercontent.com/tomsercu/lstm/master/data/ptb.test.txt...
17:11:14 #vocab = 10000
17:11:14 Traceback (most recent call last):
17:11:14   File "examples/ptb/train_ptb.py", line 240, in <module>
17:11:14     main()
17:11:14   File "examples/ptb/train_ptb.py", line 229, in main
17:11:14     trainer.run()
17:11:14   File "/home/user/.local/lib/python2.7/site-packages/chainer/training/trainer.py", line 292, in run
17:11:14     while not stop_trigger(self):
17:11:14   File "/home/user/.local/lib/python2.7/site-packages/chainer/training/triggers/interval.py", line 46, in __call__
17:11:14     previous_epoch_detail = updater.previous_epoch_detail
17:11:14   File "/home/user/.local/lib/python2.7/site-packages/chainer/training/updater.py", line 148, in previous_epoch_detail
17:11:14     return self._iterators['main'].previous_epoch_detail
17:11:14 AttributeError: 'ParallelSequentialIterator' object has no attribute 'previous_epoch_detail'
```

Hakuyume:
I modified PTB example.

delta2323:
Thank you for the update. I am running the GPU test again.

delta2323:
Sorry, I missed that `WindowIterator` in the PTB example also need `previous_epoch_detail`

chainer/training/triggers/interval.py
```diff
@@ -40,9 +42,16 @@ def __call__(self, trainer):
         """
         updater = trainer.updater
         if self.unit == 'epoch':
-            prev = self.count
-            self.count = updater.epoch_detail // self.period
-            return prev != self.count
+            epoch_detail = updater.epoch_detail
+            previous_epoch_detail = updater.previous_epoch_detail
```
delta2323: As `previous_epoch_detail` is a newly-added attribute, some users could create custom updaters that does not have the `previous_epoch_detail` attribute. So, could you consider to replace the right hand side with `getattr(updater, 'previous_epoch_detail', <something>)` so that it works with updaters without `previous_epoch_detail`?

delta2323: If we solve the problem, we need not to add `previoud_epoch_detail` to custom iterators in the examples.

Hakuyume: I agree that it is inconvenient to prohibit iterators/updaters without `previous_epoch_detail` because some users already made custom iterators/updaters without it. However, if the iterator/updater does not have `previous_epoch_detail`, there is a risk of resuming issue (#2244). It is danger to ignore the absence of `previous_epoch_detail`.

How about this solution?
If `IntervalTrigger` can not get `previous_epoch_detail`, it fallbacks to previous behaviour. In addition, it raises a warning saying "Without previous_epoch_datail, IntervalTrigger has a resuming issue. Please implement this property to your iterator/updater".

delta2323: > However, if the iterator/updater does not have previous_epoch_detail, there is a risk of resuming issue (#2244). 
I see. Thank you for making the problem clearer. I will ask other core members how to deal with non-existence of `previous_epoch_detail`. Aside from that, showing a warning message seems a good idea to me.

Hakuyume:
I used `NpzSerializer` in the tests. But it has no effect on tests of `IntervalTrigger` because `IntervalTrigger` does not have `serialize` method.

delta2323:
The initial motivation of this test is to confirm the original issue #2244 is solved. So I wanted to reproduce the situation the bug occurred as a unit test. One thing came up to me is that we should think of this test as a one for `IntervalTrigger` but for other modules (maybe `Trianer`?). So it could be better to move the test to another place.

Hakuyume:
In this test, we can assume `Trainer` and `Updater` can serialize/deserialize themselves properly because this test is not for `Trainer` or `Updater`. Serialization/deserialization is not necessary to reproduce the bug repoted in #2244  although the bug is usually observed after deserialization.

delta2323:
Finally, I understand what you are thinking. I accept your change to remove serializer.

chainer/training/triggers/interval.py
Hakuyume: > I will ask other core members how to deal with non-existence of previous_epoch_detail. 

Did you get any answers about this?

delta2323: I am sorry to make you wait. We discussed the matter and will share you the summary within few days.

delta2323: Hi @Hakuyume

Sorry for our late response. I will share you the discussion as I promised. Although, we have found some negative points of the idea of controlling `IntervalTrigger` with  `Updater.previous_epoch_detail`.

IntervalTrigger cannot rely on Updater.previous_epoch_detail if it is **not called at every epoch**.
Suppose we have a instance of IntervalTrigger whose interval is 1 epoch and 1 epoch consists of 4 iterations. Consider the following situation:

```
 1  2   3   4    5   6   7   8   9
---|---|---|---||---|---|---|---||---|
 x       x              o       x
```

The numbers are iteration counts and x and o indicate the iteration the trigger is *called* (not *fired*. The trigger is fired only when `trigger(trainer)` is `True`). We want to pay attention to the 6th iteration (indicated as o). As this is the first iteration of the second epoch, the trigger should be fired. But, as updater.previous_epoch_detail is 1.25 (=5/4), updater.previous_epoch_detail is 1.5, the trigger cannot fire itself. In order to make the trigger fire at the iteration, the trigger must memorize the last time it is called by ourselves, otherwise, trainers are responsible for managing the last time each trigger is called.

We encountered this problem when we use `IntervalTrigger`s to collect the statistics of data nodes in the computational graph. We associate a trigger for each data node in the computational graph. But because of the define-by-run nature of Chainer, not all nodes appeared at every forward propagation (we somehow associate data nodes in the computational graphs of different iterations somehow).

If we assume that triggers are called every iteration, this problem does not happen. Actually it is true of triggers for controlling the invocation of extensions, that are fed with `trigger` option of the `Trainer.extend` method. But this assumption restricts the usage of triggers. Actually one of our members find it when he uses triggers to collect the statistics of Chainer variables.

At first, I thought it is somewhat strange the `IntervalTrigger` is stateful. But we found that we already have stateful triggers (e.g. `BestValueTrigger`). So, we need to serialize triggers anyway.

Of course this it is not the final decision, we are open to discussion with you and of course anyone who is interested in this topic. I would like to know your opinion and I promise I will reply to you as quickly as possible in the next time.

Hakuyume: @delta2323 Thank you for sharing the discussion.

> IntervalTrigger cannot rely on Updater.previous_epoch_detail if it is not called at every epoch.

I agree this (maybe `every epoch` -> `every iteration`). Current design assumes that triggers are called for each iteration.

> We encountered this problem when we use IntervalTriggers to collect the statistics of data nodes in the computational graph. We associate a trigger for each data node in the computational graph.

What is `node(s)`? Could you share me some examples? 

> Actually it is true of triggers for controlling the invocation of extensions, that are fed with trigger option of the Trainer.extend method. But this assumption restricts the usage of triggers. 

So will you change the arguments of `trigger.__call__()`? Currently, it takes a trainer. I think this is specific. 

Even if we save `previous_epoch_detail` in triggers, the issue, how to treat the cases that updaters/triggers were serialized by older version and do not have `previous_epoch_detail` in it (https://github.com/chainer/chainer/pull/2484#discussion_r112811180). How do you think of it.

Hakuyume: I found a new issue that current `IntervalTrigger` does not works even for `iterations` if the trigger is not called for each iteration.

delta2323: > What is node(s)? Could you share me some examples?

Sorry, this was an undefined term. What I have in mind is that we have a abstract computational graph that consists of **nodes** and **operators** and we regard a computational graph (as a module of Chainer) made by forward propagation as a realization of the part of the abstract graph. Nodes in the abstract computational graph correspond to `chainer.Variable`s and operators to `chainer.Function`s. 

For example, suppose we change operations to be applied to the intermediate variable depending on its value, abstract computational graph will have a branching and each forward propagation will realize either of the branches.

delta2323: > So will you change the arguments of trigger.__call__()? Currently, it takes a trainer. I think this is specific.

No, we do not think of it. I am not sure I understand your question but what I wanted to say here is that developers of `trigger`s must assume that they are called every iteration, that restricts the usage of `trigger`s.

Hakuyume: > Nodes in the abstract computational graph correspond to chainer.Variables and operators to chainer.Functions.

I understand some nodes are not called for each iteration. However, from my understanding, `Variable` and `Function` do not have any `Trainer` objects. How do they call a trigger internally?

Hakuyume: > developers of triggers must assume that they are called every iteration, that restricts the usage of triggers.

I understand that you do not want to restrict the usage of trigger. To be honest, it seems strange to me that triggers are called from non-Trainer objects. because it takes Trainer as its argument. My suggestion is making the APIs more general If you want to support such usages officially.

delta2323: > Even if we save previous_epoch_detail in triggers, the issue, how to treat the cases that updaters/triggers were serialized by older version and do not have previous_epoch_detail in it

Sorry, my reply was about the difficulty to use `Updater.previous_epoch_detail` and did not fully answer to your original question.

As `IntervalTrigger` cannot rely on `Updater.previous_epoch_detail`, it should store and serialize the previous value of `epoch_detail` (as e.g. `IntervalTrigger.previous_epoch_detail`).  When `previous_epoch_detail` attribute is missing in deserializing `IntervalTrigger`, there are two choices:
1. Do not rely on `Updater.previous_epoch_detail`. Set some special value that indicates (e.g. `None`, `-1`) and to `IntervalTrigger.previous_epoch_detail` and raise a warning that notifies users that there could be a resuming problem.
2. Try to retrieve the value `Updater.previous_epoch_detail`, assuming that the trigger is called at every iteration. If it is missing, raise a warning of the resuming problem.

If there is a way to infer the previous value of `epoch_detail` from trainer, 
It can take invalid values for indicating the first iteration (e.g `None` or `-1`).

1.  Try to get the value from `IntervalTrigger.previous_epoch_detail`.
2. If `IntervalTrigger.previous_epoch_detail` is missing, try to get the value from `Updater.previous_epoch_detail`, assuming that the trigger is called at every iteration. If it is successful, we store the value to `IntervalTrigger.previous_epoch_detail`.
3. If `Updater.previous_epoch_detail`, we set previous_epoch_detail to

Hakuyume: If you want to support the cases that triggers are not called for each iteration, `IntervalTrigger` requires not only `previous_epoch_detail` but also `previous_iteration`.

niboshi: > >     Nodes in the abstract computational graph correspond to chainer.Variables and operators to chainer.Functions.
>
> 
> I understand some nodes are not called for each iteration. However, from my understanding, Variable and Function do not have any Trainer objects. How do they call a trigger internally?

This is outside the standard features of `Variable(Node)` and `Function`, but an extension does this.

In an extension I'm developing, it holds "logical" variable and function nodes(*) and triggers associated with them. Not necessarily all logical nodes are instantiated in each iteration, because of Define-by-Run nature. Only when the associated logical nodes are instantiated, the extension call these triggers with the trainer instance (which are given to the extension).

(*) Corresponding variables and functions across multiple iterations are considered a single "logical" variables and functions, despite distinct physical `Variable(Node)` and `Function` objects.

Hakuyume: > In an extension I'm developing, it holds "logical" variable and function nodes(*) and triggers associated with them. Not necessarily all logical nodes are instantiated in each iteration, because of Define-by-Run nature. Only when the associated logical nodes are instantiated, the extension calls these triggers with the trainer instance (which was given to the extension).

Thank you. This is exactly what I want to know.

Hakuyume: I understand the situation and it is not so strange. I think it is a good solution that each trigger keeps `previous_iteration` and `previous_epoch_detail` in itself.

Hakuyume:
I updated `IntervalTrigger` according to discussion in https://github.com/chainer/chainer/pull/2484#pullrequestreview-34158881.

@delta2323 Could you check my codes?

delta2323:
Thank you for updating the code. OK, I will check #2824 first.

delta2323:
Now #2824 is merged, I will continue the review of this PR.

delta2323:
Could you merge the master branch as #2824 has been merged to it.

tests/chainer_tests/training_tests/triggers_tests/test_interval_trigger.py
```diff
@@ -1,80 +1,136 @@
...
+import tempfile
 import unittest
 
+from chainer import serializers
 from chainer import testing
 from chainer import training
 
 
-class DummyUpdater(training.Updater):
+def get_trainer_with_mock_updater(iter_per_epoch):
```
delta2323: Use `chainer.testing.get_trainer_with_mock_updater`

Hakuyume: `chainer.testing.get_trainer_with_mock_updater` uses `IntervalTrigger` internally. I think using `chainer.testing.get_trainer_with_mock_updater` for testing `IntervalTrigger` is not good.

Hakuyume: I noticed that we can change the trigger used in `get_trainer_with_mock_updater` by `stop_trigger` argument.

Hakuyume: And I want to change the number of iterations per epoch for testing. `chainer.testing.get_trainer_with_mock_updater` does not support it.

Hakuyume: I noticed we can avoid using `IntervalTrigger` in `get_trainer_with_mock_updater` by setting `stop_trigger=None`. To solve later issue, I added `iter_per_epoch` option to `get_trainer_with_mock_updater` #2913.

mitmul:
This PR passed all CI test jobs.

delta2323:
LGTM

delta2323:
LGTM, I will merge #2913 firs.

delta2323:
Jenkins, test this please.

delta2323:
Jenkins test passed.

Hakuyume:
@delta2323 Isn't this PR `to-be-backported`? This bug also exists in v2.

delta2323:
In current guide line, as we wrote in (the contribution guide](https://docs.chainer.org/en/stable/contribution.html#issue-pr-labels), minor revisions should not add any feature, or documented API, while I agree that the bug in v2 should be fixed. In that sense, the guide does anticipate this situation.

Hakuyume:
@delta2323 Let me confirm.
- This bug is in v2 and you agree that it should be fixed.
- This PR cannot be backported because it contains additional features

I think the code before you changed the decision (https://github.com/chainer/chainer/pull/2484#discussion_r119837941) is suitable to backporting. It has no additional features and simply fixes bug. (Note that it can't treat the case that IntervalTrigger does not called for each iteration).

Hakuyume:
> In that sense, the guide does anticipate this situation.

`does anticipate` -> `does not anticipate`?
According to current guideline, we cannot fix bugs that require additional features. How about rethinking this guideline?

delta2323:
> This bug is in v2 and you agree that it should be fixed.
> This PR cannot be backported because it contains additional features
Yes, these are what I am thinking.

> does anticipate -> does not anticipate?
Sorry, it was a typo. 

Thank you for suggestion, We are discussing the matter in the slack channel.

Hakuyume:
> Thank you for suggestion, We are discussing the matter in the slack channel.

OK, I will wait for your decision.

delta2323:
Two possible solution we can come up easily are:

* Backport modified version of this PR, in which the code raises an error if users touches newly-added API (`IntervalTrigger.serialize`).
* Make the newly-added API private but adding `_` as a prefix.

Hakuyume:
> Backport modified version of this PR, in which the code raises an error if users touches newly-added API (IntervalTrigger.serialize).

Is it possible to distinguish users from `Trainer`? 

> Make the newly-added API private but adding _ as a prefix.

We cannot make `serialize` private because it should be called by `Trainer`.

delta2323:
Hmm.. Then, the third option is that, as suggested by @beam2d  in the channel, simply we do not register `serialize` method to the document.

Hakuyume:
@delta2323 Your suggestion looks good to me. What should I do?

delta2323:
Thank you for your cooperation. You don't have to do anything. I made a backport PR #3151.

PR #2491, opened 2017-04-03
Improve docs of stack, vstack, hstack

keisuke-umezawa:

Modified functions/links:
 - functions/array/stack
 - functions/array/hstack
 - functions/array/vstack

doctest result is
```
Document: reference/functions
-----------------------------
1 items passed all tests:
 117 tests in default
117 tests in 1 items.
117 passed and 0 failed.
Test passed.
```

### DIFF
--- a/chainer/functions/array/hstack.py
+++ b/chainer/functions/array/hstack.py
@@ -53,10 +53,30 @@ def hstack(xs):
     """Concatenate variables horizontally (column wise).
 
     Args:
-        xs (list of chainer.Variable): Variables to be concatenated.
+        xs (list of :class:`~chainer.Variable` or :class:`numpy.ndarray` or \
+        :class:`cupy.ndarray`):
+            Input variables to be concatenated. The variables must have the
+            same shape, except in the horizontal (column wise) dimension.
 
     Returns:
         ~chainer.Variable: Output variable.
+
+    .. admonition:: Example
+
+        >>> x1 = np.arange(0, 12).reshape(3, 4)
+        >>> x1
+        array([[ 0,  1,  2,  3],
+               [ 4,  5,  6,  7],
+               [ 8,  9, 10, 11]])
+        >>> x2 = np.arange(12, 18).reshape(3, 2)
+        >>> x2
+        array([[12, 13],
+               [14, 15],
+               [16, 17]])
+        >>> F.hstack([x1, x2]).data
+        array([[ 0,  1,  2,  3, 12, 13],
+               [ 4,  5,  6,  7, 14, 15],
+               [ 8,  9, 10, 11, 16, 17]])
 
     """
 
--- a/chainer/functions/array/hstack.py
+++ b/chainer/functions/array/hstack.py
@@ -57,6 +57,8 @@ def hstack(xs):
         :class:`cupy.ndarray`):
             Input variables to be concatenated. The variables must have the
             same shape, except in the horizontal (column wise) dimension.
+            Otherwise, it throws :class:`chainer.utils.type_check.InvalidType`
+            exception.
 
     Returns:
         ~chainer.Variable: Output variable.
--- a/chainer/functions/array/hstack.py
+++ b/chainer/functions/array/hstack.py
@@ -63,7 +63,14 @@ def hstack(xs):
             the same shape.
 
     Returns:
-        ~chainer.Variable: Output variable.
+        ~chainer.Variable:
+            Output variable. When the input variables have the second axis
+            (i.e. :math:`ndim \\geq 2`), the shapes of inputs and output are
+            the same along all but the second axis. The length of second axis
+            is the sum of the lengths of inputs' second axis.
+            When the variables do not have the second axis (i.e.
+            :math:`ndim < 2`), the shape of output is ``(N, )`` (``N`` is the
+            sum of the input variables' size).
 
     .. admonition:: Example
 
--- a/chainer/functions/array/hstack.py
+++ b/chainer/functions/array/hstack.py
@@ -56,9 +56,9 @@ def hstack(xs):
         xs (list of :class:`~chainer.Variable` or :class:`numpy.ndarray` or \
         :class:`cupy.ndarray`):
             Input variables to be concatenated. The variables must have the
-            same shape, except in the horizontal (column wise) dimension.
-            Otherwise, it throws :class:`chainer.utils.type_check.InvalidType`
-            exception.
+            same ``ndim``. Also, when the variables have horizontal dimension
+            (i.e. :math:`ndim \\geq 2`), the variables must have the same
+            shape, except in the horizontal (column wise) dimension.
 
     Returns:
         ~chainer.Variable: Output variable.
--- a/chainer/functions/array/hstack.py
+++ b/chainer/functions/array/hstack.py
@@ -67,6 +67,17 @@ def hstack(xs):
 
     .. admonition:: Example
 
+        >>> x1 = np.array((1, 2, 3))
+        >>> x1.shape
+        (3,)
+        >>> x2 = np.array((2, 3, 4))
+        >>> x2.shape
+        (3,)
+        >>> y = F.hstack((x1, x2))
+        >>> y.shape
+        (6,)
+        >>> y.data
+        array([1, 2, 3, 2, 3, 4])
         >>> x1 = np.arange(0, 12).reshape(3, 4)
         >>> x1.shape
         (3, 4)
--- a/chainer/functions/array/hstack.py
+++ b/chainer/functions/array/hstack.py
@@ -56,9 +56,11 @@ def hstack(xs):
         xs (list of :class:`~chainer.Variable` or :class:`numpy.ndarray` or \
         :class:`cupy.ndarray`):
             Input variables to be concatenated. The variables must have the
-            same ``ndim``. Also, when the variables have horizontal dimension
-            (i.e. :math:`ndim \\geq 2`), the variables must have the same
-            shape, except in the horizontal (column wise) dimension.
+            same ``ndim``. When the variables have the second axis (i.e.
+            :math:`ndim \\geq 2`), the variables must have the same shape
+            along all but the second axis. When the variables do not have the
+            second axis(i.e. :math:`ndim < 2`), the variables need not to have
+            the same shape.
 
     Returns:
         ~chainer.Variable: Output variable.
--- a/chainer/functions/array/hstack.py
+++ b/chainer/functions/array/hstack.py
@@ -64,16 +64,23 @@ def hstack(xs):
     .. admonition:: Example
 
         >>> x1 = np.arange(0, 12).reshape(3, 4)
+        >>> x1.shape
+        (3, 4)
         >>> x1
         array([[ 0,  1,  2,  3],
                [ 4,  5,  6,  7],
                [ 8,  9, 10, 11]])
         >>> x2 = np.arange(12, 18).reshape(3, 2)
+        >>> x2.shape
+        (3, 2)
         >>> x2
         array([[12, 13],
                [14, 15],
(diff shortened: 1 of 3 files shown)


delta2323:
How about adding the shape of output arrays to examples? It could be more intuitive than actual data.

chainer/functions/array/hstack.py
```diff
@@ -53,11 +53,31 @@ def hstack(xs):
     """Concatenate variables horizontally (column wise).
 
     Args:
-        xs (list of chainer.Variable): Variables to be concatenated.
+        xs (list of :class:`~chainer.Variable` or :class:`numpy.ndarray` or \
+        :class:`cupy.ndarray`):
+            Input variables to be concatenated. The variables must have the
+            same shape, except in the horizontal (column wise) dimension.
```
delta2323: Please describe the behavior when input arrays does not have horizontal dimension (i.e. `ndim` == 0 or 1).

chainer/functions/array/vstack.py
```diff
@@ -49,11 +49,32 @@ def vstack(xs):
     """Concatenate variables vertically (row wise).
 
     Args:
-        xs (list of chainer.Variable): Variables to be concatenated.
+        xs (list of :class:`~chainer.Variable` or :class:`numpy.ndarray` or \
+        :class:`cupy.ndarray`):
+            Input variables to be concatenated. The variables must have the
+            same shape, except in the vertical (row wise) dimension.
```
delta2323: Please describe the behavior when input arrays does not have horizontal dimension (i.e. ndim == 0).

chainer/functions/array/stack.py
```diff
@@ -6,12 +6,61 @@ def stack(xs, axis=0):
...
 
     Args:
-        xs (list of chainer.Variable): Variables to be concatenated.
-        axis (int): Axis of result along which variables are stacked.
+        xs (list of :class:`~chainer.Variable` or :class:`numpy.ndarray` or \
+        :class:`cupy.ndarray`):
+            Input variables to be concatenated. The variables must have the
+            same shape.
+        axis (int): The axis along which the arrays will be stacked. Default
+            is 0.
```
delta2323: Please write the restriction of the value of `axis`. At least, `0<= axis < ndim` is acceptable and `axis >= ndim` is not allowed. I did not investigated negative axis yet.

keisuke-umezawa: As shown in the example, it seems to allow `0<= axis <= ndim is acceptable and `axis > ndim` is not allowed. (ndim is inputted array's dimension).
And if you input the `axis < 0`, it is the same result with `ndim + 1 + axis`.

chainer/functions/array/hstack.py
keisuke-umezawa: I tried this case and get the result below.
```
>>> x1 = np.arange(0, 12).reshape(3, 4)
>>> x2 = np.arange(0).reshape(3, 0)
>>> x2.shape
(3, 0)
>>> y = F.hstack([x1, x2])
>>> y.shape
(3, 4)
>>> y.data
array([[ 0,  1,  2,  3],
       [ 4,  5,  6,  7],
       [ 8,  9, 10, 11]])
```

I think that the result is intuitive and I think it is not necessary to add the additional description.
But I do not have enough knowledge about chainer, so please advise what to write.

delta2323: Thank you for investigation. In your case, `x2.ndim` equals to `2`, not `0` nor `1` (`ndim` is a dimension of the ndarray, which is equal to `len(x.shape)`). As `hstack` concatenate input arrays along second axes, it is not trivial for users what happens if input arrays do not have the axes.

chainer/functions/array/stack.py
delta2323: Thank you for explanation. Then, could you write this restriction to the description of `axis`? Aside from that, I noticed that we do not check `axis=-(ndim + 1)` case in the unit test. I will add the case as another PR.

delta2323: I am checking the specification of the negative axis in NumPy (i.e. NumPy guarantees the behavior or it happens to behave correctly), as pointed out in #2532. So, could you wait applying this change? Or I would appreciate it if you investigate it if you have a time.

keisuke-umezawa: Thank you for the investigations. I will write the restrictions of the axis.

chainer/functions/array/hstack.py
keisuke-umezawa: I recognized my misunderstand. I will check the result and write the descriptions.

keisuke-umezawa: I added the comment below,
```
Otherwise, it throws chainer.utils.type_check.InvalidType exception.
```

chainer/functions/array/vstack.py
keisuke-umezawa: Same treatment with hstack.

keisuke-umezawa:
@delta2323 plz reveiw after the test.

chainer/functions/array/hstack.py
delta2323: It seems `F.stack` does not raise `InvalidType` error when array is NumPy and `ndim` is 0. Could you check it does work other cases?

```
In [1]: import numpy as np
In [2]: from chainer import functions as F
In [3]: a = np.array(4)
In [4]: b = np.array(4)
In [5]: F.stack((a, b))
Out[5]: <variable at 0x113acd860>
In [6]: F.stack((a, b)).data
Out[6]: array([4, 4])
```

keisuke-umezawa: So, sorry... I have just recognized what you mean.

How about this one? I will rewrite like this.

chainer/functions/array/hstack.py
```diff
@@ -53,11 +53,40 @@ def hstack(xs):
     """Concatenate variables horizontally (column wise).
 
     Args:
-        xs (list of chainer.Variable): Variables to be concatenated.
+        xs (list of :class:`~chainer.Variable` or :class:`numpy.ndarray` or \
+        :class:`cupy.ndarray`):
+            Input variables to be concatenated. The variables must have the
+            same ``ndim``. Also, when the variables have horizontal dimension
+            (i.e. :math:`ndim \\geq 2`), the variables must have the same
+            shape, except in the horizontal (column wise) dimension.
```
delta2323: I am afraid this is not the right behavior, unfortunately. When `ndim=1`, this function concatenates along the first axis and the input variables need not have same shape.

```
In [2]: a = np.array([3, 4])
In [3]: b = np.array([5, 6, 7, 8])
In [6]: chainer.functions.hstack((a, b)).data
Out[6]: array([3, 4, 5, 6, 7, 8])
```

chainer/functions/array/vstack.py
```diff
@@ -52,10 +52,11 @@ def vstack(xs):
         xs (list of :class:`~chainer.Variable` or :class:`numpy.ndarray` or \
         :class:`cupy.ndarray`):
             Input variables to be concatenated. The variables must have the
-            same ``ndim``. Also, when the variables have vertical dimension
-            (i.e. :math:`ndim \\geq 1`), the variables must have the same
-            shape, except in the vertical (row wise) dimension.
-
+            same ``ndim``. When the variables have the second axis (i.e.
```
delta2323: Is `second` should be `first`?

chainer/functions/array/vstack.py
```diff
@@ -52,10 +52,11 @@ def vstack(xs):
...
+            :math:`ndim \\geq 2`), the variables must have the same shape
```
delta2323: Should it be `\\geq 1`?

chainer/functions/array/vstack.py
```diff
@@ -52,10 +52,11 @@ def vstack(xs):
...
+            along all but the first axis. When the variables do not have the
+            second axis(i.e. :math:`ndim < 2`), the variables must have the
```
delta2323: Should it be the first axis (ndim < 1)?

delta2323: If `ndim < 1`, then input variables must be scalar (i.e. their shape is `()`). So, their shapes agree automatically.

chainer/functions/array/vstack.py
keisuke-umezawa: When concatenating the 1dim array, the variables must have the same shape along all ~~but the first axis~~... Are there my misunderstanding?

```
>>> import numpy as np
>>> import chainer.functions as F
>>> x1 = np.arange(0, 12)
>>> x2 = np.arange(0, 12)
>>> F.vstack([x1, x2]).data
array([[ 0,  1,  2,  3,  4,  5,  6,  7,  8,  9, 10, 11],
          [ 0,  1,  2,  3,  4,  5,  6,  7,  8,  9, 10, 11]])
>>> x2 = np.arange(0, 11)
>>> F.vstack([x1, x2]).data
  File "/home/kumezawa/dev/chainer/chainer/functions/array/vstack.py", line 59, in vstack
    return Vstack()(*xs)
  File "/home/kumezawa/dev/chainer/chainer/function.py", line 189, in __call__
  File "/home/kumezawa/dev/chainer/chainer/function.py", line 273, in _check_data_type_forward
  File "/home/kumezawa/dev/chainer/chainer/functions/array/vstack.py", line 23, in check_type_forward
    type_check.expect(in_types[0].shape == in_types[i].shape)
  File "/home/kumezawa/dev/chainer/chainer/utils/type_check.py", line 487, in expect
    expr.expect()
  File "/home/kumezawa/dev/chainer/chainer/utils/type_check.py", line 449, in expect
Invalid operation is performed in: Vstack (Forward)

Expect: in_types[0].shape == in_types[1].shape
Actual: (12,) != (11,)
```

keisuke-umezawa: Ok, maybe I understand. When using 1dim array, the array is column-wide (or 2nd axis wide) and the size of vertical dimension(or 1st axis) always 1.

keisuke-umezawa: So, I misunderstood thant 1-dim array is column vector, but actually it is row vector.
Is it right?

chainer/functions/array/vstack.py
delta2323: Sorry, I did not check the behavior of the corner case well and that would confuse you. Current description seems correct to me.

keisuke-umezawa: the reason why the behavior is confusing us is maybe below.
When (vstack|hstack)ing 2-dim array, the first axis is vertical axis and the second is horizontal axis.
But when (vstack|hstack)ing 1-dim array, the first axis is regarded as horizontal axis.
So, 1-dim array is treated as row vector.

This is the reason why the description of 1-dim array seems not to be consistent with 2-dim array.

delta2323: Thank you for your explanation. Finally, I understand it.

delta2323:
If you do not mind, I would appreciate it if you add the shape of output variables in details.

keisuke-umezawa:
I added examples of 1-dim array and detail descriptions about the output variables.

chainer/functions/array/stack.py
```diff
@@ -17,7 +17,13 @@ def stack(xs, axis=0):
             is the same with :math:`ndim + 1 - |axis|`.
 
     Returns:
-        ~chainer.Variable: Output variable.
+        ~chainer.Variable:
+            Output variable. Let ``x_1, x_2, ..., x_n`` and ``y`` be the input
+            variables and the output variable,
+            ``y[:, ..., 0, ..., :]`` is ``x_1``,
+            ``y[:, ..., 1, ..., :]`` is ``x_2``
+            and ``y[:, ..., n, ..., :]`` is ``x_n`` (The indexed axis indicates
```
delta2323: Index is not `n` but `n-1`.

PR #2511, opened 2017-04-06
Skip tests that use axis option of count_nonzero

delta2323:
`axis` option was introduced to `count_nonzero` in NumPy 1.12.0 . So we have to skip tests that uses this option if NumPy is older.


okuta:
Thank you!
LGTM!

PR #2513, opened 2017-04-06
Add TransformDataset

yuyu2172:
This PR adds a general dataset class called `TransformDataset`.
This class lets users to apply a function to a indexed data.
It becomes handy in the case when a user wants to write a custom data augmentation code on top of the prepared dataset class.

This class decouples implementation of data augmentation code and a code to retrieve data from filesystem or from web, making it a lot easier to reuse datasets across different tasks.

The class is already used in ChainerCV, an extension library to Chainer.

### DIFF
--- a/chainer/datasets/transform_dataset.py
+++ b/chainer/datasets/transform_dataset.py
@@ -9,6 +9,8 @@ class TransformDataset(dataset_mixin.DatasetMixin):
     dataset's :meth:`__getitem__`. Arrays returned by :meth:`__getitem__` of
     the base dataset with integer as an argument are transformed by the given
     function :obj:`transform`.
+    Also, :meth:`__len__` returns the integer returned by the base dataset's
+    :meth:`__len__`.
 
     The function :obj:`transform` takes, as an argument, :obj:`in_data`, which
     is the output of the base dataset's :meth:`__getitem__`, and returns
@@ -25,7 +27,9 @@ class TransformDataset(dataset_mixin.DatasetMixin):
 
     Args:
         dataset: The underlying dataset. The index of this dataset corresponds
-            to the index of the base dataset.
+            to the index of the base dataset. This object needs to support
+            functions :meth:`__getitem__` and :meth:`__len__` as described
+            above.
         transform (callable): A function that is called to transform values
             returned by the underlying dataset's :meth:`__getitem__`.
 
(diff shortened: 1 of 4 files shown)


chainer/datasets/transform_dataset.py
```diff
@@ -0,0 +1,43 @@
...
+    >>> from chainer.datasets import TransformDataset
+    >>> dataset, _ = get_mnist()
+    >>> def transform(in_data):
+    >>>     img, label = in_data
+    >>>     img -= 0.5  # scale to [-0.5, -0.5]
+    >>>     return img, label
+    >>> dataset = TransformDataset(dataset, transform)
+
+    Args:
+        dataset: The underlying dataset. The index of this dataset corresponds
```
mitmul: The dataset object given here should have `__len__` method besides `__getitem__` method because `len(self._dataset)` is called in the `__len__` method of this TransoformDataset. So I think it would be better to describe explicitly about what are required for "The underlying dataset". As for `__getitem__` method of the underlying dataset, it has been already mentioned in the beginning of this docstring, so please add an explanation about `__len__` method here or the first paragraph of the docstring.

chainer/datasets/transform_dataset.py
```diff
@@ -0,0 +1,43 @@
...
+
+    The function :obj:`transform` takes, as an argument, :obj:`in_data`, which
+    is the output of the base dataset's :meth:`__getitem__`, and returns
+    the transformed arrays as output. Please see the following example.
+
+    >>> from chainer.datasets import get_mnist
+    >>> from chainer.datasets import TransformDataset
+    >>> dataset, _ = get_mnist()
+    >>> def transform(in_data):
+    >>>     img, label = in_data
```
mitmul: When the line is continued without evaluation by the interpreter, you need to begin the line with `...` instead of `>>>` here.

chainer/datasets/transform_dataset.py
```diff
@@ -0,0 +1,43 @@
...
+    >>>     img -= 0.5  # scale to [-0.5, -0.5]
```
mitmul: Ditto.

chainer/datasets/transform_dataset.py
```diff
@@ -0,0 +1,43 @@
...
+    >>>     return img, label
```
mitmul: Ditto.

yuyu2172:
Thanks for reviewing!
I fixed points you pointed out.

mitmul:
LGTM

PR #2532, opened 2017-04-11
Strengthen the unittest of F.stack

delta2323:
Current unittest of `F.stack` does not check `axis=-(ndim+1)` case, which is valid. As we would like to add this case to the document (#2491), I add this case to confirm it works.


okuta:
Is this correct behavior?
I am worried about this is undefined behavior.

delta2323:
I check the behavior is expected one in NumPy v1.12.1 with small examples. [NumPy document](https://docs.scipy.org/doc/numpy/reference/generated/numpy.stack.html) does not explicitly describe the behavior of negative axis (except `-1`, which it takes as an example). But looking at [the implementation](https://github.com/numpy/numpy/blob/v1.12.0/numpy/core/shape_base.py#L350) it accepts `-(ndim+1) <= axis <= ndim`.

okuta:
LGTM!

PR #2534, opened 2017-04-11
Support batch-common-mask dropconnect

fukatani:
Current dropconnect support only batchwise mask. (i.e. masked element is different with each sample in mini-batch).
This mask specification is same with the original implementation.

One of purpose of this option is memory conservation.

### DIFF
--- a/chainer/functions/noise/simplified_dropconnect.py
+++ b/chainer/functions/noise/simplified_dropconnect.py
@@ -53,13 +53,13 @@ class SimplifiedDropconnect(function.Function):
     def forward(self, inputs):
         scale = inputs[1].dtype.type(1. / (1 - self.ratio))
         xp = cuda.get_array_module(*inputs)
-        if self.batchwise_mask:
-            mask_shape = (inputs[0].shape[0], inputs[1].shape[0],
-                          inputs[1].shape[1])
-        else:
-            mask_shape = (inputs[1].shape[0], inputs[1].shape[1])
 
         if self.mask is None:
+            if self.batchwise_mask:
+                mask_shape = (inputs[0].shape[0], inputs[1].shape[0],
+                              inputs[1].shape[1])
+            else:
+                mask_shape = (inputs[1].shape[0], inputs[1].shape[1])
             if xp == numpy:
                 self.mask = xp.random.rand(*mask_shape) >= self.ratio
             else:
--- a/chainer/functions/noise/simplified_dropconnect.py
+++ b/chainer/functions/noise/simplified_dropconnect.py
@@ -25,7 +25,7 @@ class SimplifiedDropconnect(function.Function):
 
     """Linear unit regularized by simplified dropconnect."""
 
-    def __init__(self, ratio: object, mask: object = None, batchwise_mask: object = True) -> object:
+    def __init__(self, ratio, mask=None, batchwise_mask=True):
         self.ratio = ratio
         self.mask = mask
         self.batchwise_mask = batchwise_mask
--- a/chainer/functions/noise/simplified_dropconnect.py
+++ b/chainer/functions/noise/simplified_dropconnect.py
@@ -16,7 +16,7 @@ def _matmul(a, b, xp):
     if xp is numpy:
         # numpy 1.9 does not support matmul.
         # So we use numpy.einsum instead of numpy.matmul.
-        return xp.einsum('ijk,ikl->ijl', a, b)
+        return xp.einsum('...jk,...kl->...jl', a, b)
     else:
         return xp.matmul(a, b)
 
@@ -71,12 +71,8 @@ class SimplifiedDropconnect(function.Function):
         x = _as_mat(inputs[0])
         W = inputs[1] * scale * self.mask
 
-        if self.batchwise_mask:
-            # ijk,ik->ij
-            y = _matmul(W, x[:, :, None], xp)
-        else:
-            y = x.dot(W.T)
-
+        # (i)jk,ik->ij
+        y = _matmul(W, x[:, :, None], xp)
         y = y.reshape(y.shape[0], y.shape[1]).astype(x.dtype, copy=False)
 
         if len(inputs) == 3:
@@ -91,18 +87,12 @@ class SimplifiedDropconnect(function.Function):
         gy = grad_outputs[0]
         xp = cuda.get_array_module(*inputs)
 
-        if self.batchwise_mask:
-            # ij,ijk->ik
-            gx = _matmul(gy[:, None, :], W, xp).reshape(inputs[0].shape)
-        else:
-            gx = gy.dot(W)
+        # ij,(i)jk->ik
+        gx = _matmul(gy[:, None, :], W, xp).reshape(inputs[0].shape)
         gx = gx.astype(x.dtype, copy=False)
 
-        if self.batchwise_mask:
-            # ij,ik,ijk->jk
-            gW = (gy[:, :, None] * x[:, None, :] * self.mask).sum(0) * scale
-        else:
-            gW = gy.T.dot(x) * scale
+        # ij,ik,ijk->jk
+        gW = (gy[:, :, None] * x[:, None, :] * self.mask).sum(0) * scale
         gW = gW.astype(W.dtype, copy=False)
 
         if len(inputs) == 3:
--- a/chainer/functions/noise/simplified_dropconnect.py
+++ b/chainer/functions/noise/simplified_dropconnect.py
@@ -134,7 +134,8 @@ def simplified_dropconnect(x, W, b=None, ratio=.5, batchwise_mask=True,
         ratio (float):
             Dropconnect ratio.
         batchwise_mask (bool):
-            If ``True``, dropped value depends on each sample in mini-batch.
+            If ``True``, dropped connections depend on each sample in
+            mini-batch.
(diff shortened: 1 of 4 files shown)


chainer/functions/noise/simplified_dropconnect.py
```diff
@@ -25,9 +25,10 @@ class SimplifiedDropconnect(function.Function):
 
     """Linear unit regularized by simplified dropconnect."""
 
-    def __init__(self, ratio, mask=None):
+    def __init__(self, ratio, batchwise_mask, mask=None):
```
delta2323: I am afraid it breaks compatibility because if a user code that feeds `mask` as a positional argument accept that value as `batchwise_mask` in this method.

chainer/functions/noise/simplified_dropconnect.py
```diff
@@ -127,6 +133,8 @@ def simplified_dropconnect(x, W, b=None, ratio=.5, train=True, mask=None):
         b (~chainer.Variable): Bias variable (optional) of shape ``(M,)``.
         ratio (float):
             Dropconnect ratio.
+        batchwise_mask (bool):
+            If ``True``, dropped value depends on each sample in mini-batch.
```
delta2323: It seems to me that `value` indicates the activation, or the output of the function. So, how about changing it with "dropped connections depend on ..."?

tests/chainer_tests/functions_tests/noise_tests/test_simplified_dropconnect.py
```diff
@@ -98,4 +98,55 @@ def test_backward_gpu_nobias(self):
...
+
+    def check_forward(self, x_data, W_data, b_data):
+        # Check only data type, y is tested by SimplifiedDropconnect link test.
+        x = chainer.Variable(x_data)
+        W = chainer.Variable(W_data)
+        b = chainer.Variable(b_data)
+        func = simplified_dropconnect.SimplifiedDropconnect(0.5, False)
+        y = func(x, W, b)
+        self.assertEqual(y.data.dtype, numpy.float32)
+        self.assertEqual(func.mask.shape, W.shape)
```
delta2323: Could you add this shape-check to the test case that uses sample-wise masks (i.e. `TestSimplifiedDropconnect`)?

tests/chainer_tests/links_tests/activation_tests/test_simplified_dropconnect.py
```diff
@@ -181,6 +181,42 @@ def test_serialization(self):
...
+
+        x_shape = (4,) + self.in_shape
+        self.x = numpy.ones(x_shape).astype(numpy.float32)
+
+    def check_forward(self, x_data):
+        x = chainer.Variable(x_data)
+        y = self.link(x, False, True)
+
+        # check mask equality here.
+        self.assertTrue((y.data[0] == y.data[1]).all())
```
delta2323: use `numpy.testing.assert_allclose`

fukatani:
Thanks for your review.
I fixed them.

`batchwise_mask` argument lasted in all APIs.

tests/chainer_tests/links_tests/activation_tests/test_simplified_dropconnect.py
```diff
@@ -181,6 +183,42 @@ def test_serialization(self):
...
+        self.x = numpy.ones(x_shape).astype(numpy.float32)
+
+    def check_forward(self, x_data):
+        x = chainer.Variable(x_data)
+        y = self.link(x, False, True)
+
+        # check mask equality here.
+        testing.assert_allclose(y.data[0], y.data[1])
+        testing.assert_allclose(y.data[0], y.data[2])
+        testing.assert_allclose(y.data[0], y.data[3])
```
delta2323: If possible, I would like you to check that the values satisfies not only the constraint that they are same but also their values are expected ones.

fukatani:
Thanks. I added test with `y.creator.mask`.

I want to strengthen other tests, so I will send another PR if this is merged.

delta2323:
Could you fix flake8 validation?

chainer/functions/noise/simplified_dropconnect.py
```diff
@@ -52,8 +53,12 @@ def check_type_forward(self, in_types):
     def forward(self, inputs):
         scale = inputs[1].dtype.type(1. / (1 - self.ratio))
         xp = cuda.get_array_module(*inputs)
-        mask_shape = (inputs[0].shape[0], inputs[1].shape[0],
-                      inputs[1].shape[1])
+        if self.batchwise_mask:
+            mask_shape = (inputs[0].shape[0], inputs[1].shape[0],
+                          inputs[1].shape[1])
+        else:
+            mask_shape = (inputs[1].shape[0], inputs[1].shape[1])
```
delta2323: We do not have to create `mask_shape` unless `self.mask` is `None`. So please move it after l.62

tests/chainer_tests/functions_tests/noise_tests/test_simplified_dropconnect.py
```diff
@@ -98,4 +103,88 @@ def test_backward_gpu_nobias(self):
...
+    def test_forward_cpu(self):
+        self.check_forward(self.x, self.W, self.b)
+
+    @attr.gpu
+    def test_forward_gpu(self):
+        self.check_forward(
+            cuda.to_gpu(self.x), cuda.to_gpu(self.W), cuda.to_gpu(self.b))
+
+
+class TestSimplifiedDropconnectNotBatchwiseMask(unittest.TestCase):
```
delta2323: I would like to confirm my understanding. The difference of `TestSimplifiedDropconnect` and `TestSimplifiedDropconnectBatchwiseMaskShape`  is to use the default argument of `batchwise_mask` or not, isn't it?

fukatani: Finally, I removed `TestSimplifiedDropconnectBatchwiseMaskShape`.


To test the mask shape,  I thought I had to use `simplified_dropconnect.SimplifiedDropconnect` to access mask. I made `TestSimplifiedDropconnectBatchwiseMaskShape`.
And I want to test the API `functions.simplified_dropconnect`, so I introduced testSimplifiedDropconnect`

However, I found that I can access the `mask` via `creator`, `TestSimplifiedDropconnectBatchwiseMaskShape` is unneccessary.

fukatani:
Thanks for your review! I fixed them.

tests/chainer_tests/functions_tests/noise_tests/test_simplified_dropconnect.py
delta2323: OK, I understand it

delta2323:
LGTM. I am running GPU tests.

delta2323:
Could you merge or rebase the master branch?

fukatani:
OK! I rebased.

delta2323:
Thank you. I am running GPU tests.

tests/chainer_tests/functions_tests/noise_tests/test_simplified_dropconnect.py
```diff
@@ -71,11 +78,12 @@ def test_forward_gpu_nobias(self):
     def check_backward(self, x_data, W_data, b_data, y_grad):
         args = x_data, W_data
         if b_data is not None:
-            args = args + (b_data,)
+            args += (b_data,)
```
delta2323: Parentheses can be removed.

delta2323:
Please test the case where `mask` is not `None` and `batchwise_mask` is False.

chainer/functions/noise/simplified_dropconnect.py
```diff
@@ -136,6 +142,9 @@ def simplified_dropconnect(x, W, b=None, ratio=.5, train=True, mask=None):
             Otherwise, The mask must be ``(n, M, N)`` shaped array.
```
delta2323: I think the same of `mask` should be `(M, N)` when `batchwise_mask` is `False`. Is that correct? If so, could you fix the description of `mask`?

delta2323:
It seems to me that the option name `batchwise_mask` is a bit confusing because we already have `mask` option. So, how about changing the name more explanatory like `use_batchwise_mask`?

chainer/functions/noise/simplified_dropconnect.py
fukatani: In current implementation, `mask` shape input by user is not checked. So user can input any shape. Should we check `mask` shape? And expected shape should be changed by `use_batchwise_mask`?

delta2323: So, users can provide different mask by putting a ndarray of shape `(n, M, N)` to `mask` even if `use_backwise_mask` is true, can't you? I would like to confirm that this specification is on purpose. Even so, I think it is better to check the shape of `mask` if it is broadcastable. You can do it with e.g. [`xp.broadcast_to`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.broadcast_to.html)

fukatani: OK,
I think again, we had better check mask shape.

And I added test for mask shape.

delta2323:
jenkins, test this please.

delta2323:
The GPU test passed. LGTM

Issue #2547, opened 2017-04-13
Allow cupy.fuse decorator without parentheses


niboshi:
Currently `cupy.fuse` decorator needs parentheses even if no argument is supplied.

Allowed:
```
@cupy.fuse()
def f(x, y):
    return x + y
````

Not allowed:
```
@cupy.fuse
def f(x, y):
    return x + y
````

`()` can be easily forgotten and the generated error is not very helpful.

```
Traceback (most recent call last):
  File "test_fuse.py", line 9, in <module>
    z = f(x, y)
TypeError: <lambda>() takes 1 positional argument but 2 were given
```

It would be better to allow `cupy.fuse` without parentheses.

niboshi:
Moved to https://github.com/cupy/cupy/pull/43

PR #2548, opened 2017-04-13
Allow cupy.fuse decorator without parentheses

niboshi:
See #2547.


niboshi:
Moved to https://github.com/cupy/cupy/pull/43

PR #2549, opened 2017-04-13
Use cleargrads by default

gwtnb:
Fix https://github.com/pfnet/chainer/issues/2521.

### DIFF
--- a/chainer/optimizer.py
+++ b/chainer/optimizer.py
@@ -461,10 +461,6 @@ class GradientMethod(Optimizer):
     provide such an alias to each attribute. It can be done by only adding one
     line for each attribute using :class:`HyperparameterProxy`.
 
-    .. note::
-       It is recommended to call :meth:`use_cleargrads` after creating a
-       :class:`GradientMethod` object for efficiency.
-
     Attributes:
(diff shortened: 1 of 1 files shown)


chainer/optimizer.py
```diff
@@ -530,10 +530,11 @@ def use_cleargrads(self, use=True):
...
         .. note::
-           Note that :meth:`update` calls :meth:`~Link.zerograds` by default
-           for backward compatibility. It is recommended to call this method
-           before first call of `update` because `cleargrads` is more
-           efficient than `zerograds`.
+           Note that :meth:`update` calls :meth:`~Link.cleargrads` by default.
+           :meth:`~Link.cleargrads` is more efficient than
+           :meth:`~Link.cleargrads`, so one do not have to call
+           :meth:`use_cleargrads`. This method remains for backward
+           compatibility.
```
niboshi: 1. `is more efficient than Link.cleargrads` -> `... than Link.zerograds`
2.  `one do not` -> `one does not`
3. I think this method can be marked as deprecated, like `link.zerograds()`.

niboshi:
Thank you for PR! Could you check my review?

PR #2550, opened 2017-04-13
Fix duplicated stats printing for resumed execution

soramichi:
When a training is resumed from the point when N epochs have finished, Chainer mistakenly prints N+1 stats because the `LogReport` extension is triggered right after a resume and it adds a new stat to the original N stats.

The below is a sample output when a training is resumed from the point when 2 epochs have finished. The stat for epoch 2 is shown twice (with weird values).

```
1           0.19225     0.112377              0.9415         0.9641                    15.1585       
2           0.0614427   0.097964              0.980733       0.96865                   31.9467       
2           0.0228196   0.0861454             1              0.9714                    33.387        
     total [######################################............] 77.78%
this epoch [################..................................] 33.33%
      1400 iter, 2 epoch / 3 epochs
    45.962 iters/sec. Estimated time to finish: 0:00:08.702801
```

This PR fixes this issue by adding a member named `resumed` to `trainer`. This variable is set to `True`  when a snapshot is taken so that the `LogReport` extension can check it to know whether the current state is right after a resume or not.

The blow is a sample output for the same setting as the above, after applying this PR. The stat for epoch 2 is shown only once.

```
1           0.19225     0.112377              0.9415         0.9641                    15.1585       
2           0.0614427   0.097964              0.980733       0.96865                   31.9467       
     total [############################################......] 88.89%
this epoch [#################################.................] 66.67%
      1600 iter, 2 epoch / 3 epochs
    45.515 iters/sec. Estimated time to finish: 0:00:04.394147.
```

### DIFF
--- a/chainer/training/extensions/_snapshot.py
+++ b/chainer/training/extensions/_snapshot.py
@@ -73,9 +73,7 @@ def snapshot(savefun=npz.save_npz,
     """
     @extension.make_extension(trigger=(1, 'epoch'), priority=-100)
     def snapshot(trainer):
-        trainer.resumed = True
         _snapshot_object(trainer, trainer, filename.format(trainer), savefun)
-        trainer.resumed = False
 
     return snapshot
 
--- a/chainer/training/extensions/_snapshot.py
+++ b/chainer/training/extensions/_snapshot.py
@@ -73,7 +73,9 @@ def snapshot(savefun=npz.save_npz,
     """
     @extension.make_extension(trigger=(1, 'epoch'), priority=-100)
     def snapshot(trainer):
+        trainer.resumed = True
         _snapshot_object(trainer, trainer, filename.format(trainer), savefun)
+        trainer.resumed = False
 
     return snapshot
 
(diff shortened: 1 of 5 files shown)


niboshi:
Thank you for PR!
However, introducing a flag seems a bit adhoc to me.
The real problem behind this issue is that the state of `LogReport._trigger` is not saved in the snapshots.
It would be better to make the trigger serialized correctly.

Here is my suggested implementation.
https://github.com/niboshi/chainer/commit/9047e7bb0742412778d07e3be09c6c21107bdf3d

Could you consider this and merge it to your branch if you feel it's OK?
(I tried to make a PR to your branch, but I couldn't find your fork in the base fork candidates)

soramichi:
@niboshi 

Thank you very much for your suggestion! I totally agree that `trigger` should be responsible for this issue instead of `LogReport` itself.

I checked your modification on my local environment and it looks working, so I merged your commit and also added test cases for the newly added `serialize()` method.

PR #2557, opened 2017-04-14
Improve tests of manual schedule trigger

Hakuyume:
I rewrote tests of `ManualScheduleTrigger` with `testing.parameterize`. I think this is better than a lot of test classes.


mitmul:
LGTM. All jenkins tests have been passed.

PR #2561, opened 2017-04-14
SigmoidCrossEntropy without reduction (v2)

Hakuyume:
`SigmoidCrossEntropy` without reduction. This PR is related to #2558.


chainer/functions/loss/sigmoid_cross_entropy.py
```diff
@@ -13,12 +13,7 @@ class SigmoidCrossEntropy(function.Function):
 
     ignore_label = -1
 
-    def __init__(self, normalize=True):
-        self.normalize = normalize
-
     def check_type_forward(self, in_types):
-        type_check.expect(in_types.size() == 2)
```
delta2323: Why did you remove this constraint?

chainer/functions/loss/sigmoid_cross_entropy.py
```diff
@@ -30,50 +25,39 @@ def forward(self, inputs):
...
         # stable computation of the cross entropy.
-        loss = -xp.sum(
-            self.ignore_mask * (x * (t - (x >= 0)) -
-                                xp.log1p(xp.exp(-xp.abs(x)))))
-        return utils.force_array(xp.divide(loss, self.count, dtype=x.dtype)),
+        loss = -(
+            self.ignore_mask *
+            (x * (t - (x >= 0)) - xp.log1p(xp.exp(-xp.abs(x)))))
+
+        return utils.force_array(loss.astype(x.dtype)),
```
delta2323: I think we can safely remove `utils.force_array` as `loss` is no longer 0-dimensional array (if we assume inputs are not scalar). My concern is if it works even the size of the output is 1. Could you add test cases `'shape' : (1,)` and `'shape': (1, 1)` to `TestSigmoidCrossEntropy`?

chainer/functions/loss/sigmoid_cross_entropy.py
```diff
@@ -30,50 +25,39 @@ def forward(self, inputs):
...
-        normalize (bool): Variable holding a boolean value which
-            determines the normalization constant. If true, this function
-            normalizes the cross entropy loss across all instances. If else,
-            it only normalizes along a batch size.
+        x (Variable): Variable holding a float32 array whose elements
+            indicate the unnormalized log probabilities.
+        t (Variable): Variable holding an int32 array of ground truth labels.
+            If ``t[i0, i1, ... ] == -1``, corresponding ``x[i0, i1, ... ]`` is
+            ignored. In this case, the loss value of the ignored instance is
+            set to ``0``.
```
delta2323: Do you think we should accept a scalar-input case (i.e. `x.shape` and `t.shape` are `()`)? In that case, `utils.force_array` could be needed.

chainer/functions/loss/sigmoid_cross_entropy.py
```diff
@@ -30,50 +25,39 @@ def forward(self, inputs):
...
 
     Returns:
-        Variable: A variable object holding a scalar array of the cross entropy
-            loss.
+        Variable: Variable holding a float32 array of the cross entropy.
+            The shape is same as ``x``.
```
delta2323: ... same as *that of* x.

chainer/functions/loss/sigmoid_cross_entropy.py
Hakuyume: Sorry, it was my misunderstanding. (I thought `2` means two-dimentional input).

chainer/functions/loss/sigmoid_cross_entropy.py
Hakuyume: I added those tests. In order to support scalar input, I kept `utils.force_array`.

chainer/functions/loss/sigmoid_cross_entropy.py
Hakuyume: It seems good to support a scalar input. In this case, the shape of output should be `()`.

delta2323:
LGTM, I am running GPU tests.

delta2323:
Sorry for not responding so long. GPU test failed due to the error that does not relevant to this PR. Could you merge the latest `_v2` branch?

Hakuyume:
I merged `_v2`.

delta2323:

PR #2562, opened 2017-04-14
SquaredError (v2)

Hakuyume:
`MeanSquaredError` without reduction. The function is renamed to `SquaredError`. This PR is related to #2558.


delta2323:
I thought `mean_squared_error` is still useful for users as a alias of `F.average(F.squared_error(x), axes=...)`. Or do you think the existence of such a function is confusing for users because loss functions in v2 outputs element wise values by default, while `mean_squared_error` single scalar or sample wise loss values?

tests/chainer_tests/functions_tests/loss_tests/test_squared_error.py
```diff
@@ -11,27 +11,30 @@
 from chainer.testing import condition
 
 
-class TestMeanSquaredError(unittest.TestCase):
+@testing.parameterize(
+    {'shape': (4, 3)},
+    {'shape': (4, 3, 2)},
+    {'shape': (4,)},
```
delta2323: Please test cases where `'shape'` is `()`, `(1,)`, and `(1, 1)`.

Hakuyume:
I add tests and modified `SquaredError` to support a scalar input.

Hakuyume:
I think we do not have to implement `MeanSquaredError`. It is confusing and `F.average(F.squared_error(x), axes=...)` is not so complicated.

delta2323:
Thank you for your update. I am running GPU test now.

delta2323:
From a marketing point of view, `MeanSquaredError` is so common in (statistical) machine learning community that lacking it could be unnatural for users. But when it comes to implement it, the choice of reduce strategy (`sum` or `mean`) can be problematic again.

Hakuyume:
When we implement `mean_squared_error`, it seems natural to return a `mean` value.

Hakuyume:
I agree that `MeanSquaredError` is very common.

delta2323:
Thank you for your opinion. We discussed offline and have reached a consensus that we leave `mean_squared_error` but implement it with `squared_error`.

delta2323:
Regarding reduction strategy, I guess users sometimes want the function to output sum along minibatch, not mean. Considerting the situation in which we compute the mean of loss values of whole training dataset (not minibatch). If we are given mean value for each minibatch, we have to multiply it by minibatch size (be aware that the last minibatch could be less than the other minibatches).

Hakuyume:
OK, I will restore `MeanSquaredError` and make another PR to rewrite it with `SquaredError`.

I understand the situation in which sum of squared errors is required. But users can combine `Sum` and `SquaredError` to compute sum of squared errors. Do we need to implement `reduce` option to `SquaredError`?

delta2323:
Thank you for your suggestion. I understand by your comment that we can realize sample-wise squared error with the combination of `Sum` and `SquaredError`, instead of adding `reduce` option to `MeanSquaredError`.

As it is easier to adding option afterwards than removing it. I come to think `MeanSquaredError` does not need `reduce` option (at least for now).

Hakuyume:
I found `SquaredDifference` already exists.

Hakuyume:
I restored `MeanSquaredError`. To rewrite it with `SquaredError`, we need `Average` function. Can I cherry-pick #2377 for _v2?

Hakuyume:
I found #2607 contains #2377. Is it better to wait for #2607 to be merged?

delta2323:
Sorry I could not answer to you before merging #2607. As it is merged, I believe you can continue to work on the problem.

Hakuyume:
I think this PR is finished. Could you review it?

delta2323:
> I found SquaredDifference already exists. It also computes element-wise L2 error.

Oh, sorry I missed the function... It is confusing for users there are two functions that do the same things.

delta2323:
I am sorry we have made you modify the implementation again and again. I would like to propose the following changes regarding `(Mean)SquaredError` and `(Mean)AbsoluteError`.

@unnonouno @beam2d Could you check if my proposal is appropriate?

**Problem**
* There already exists `F.SquaredDifference`, whose functionality is same as  `F.SquaredError` which has been added to the master branch (#2566). 

**Solution**
* Apply the following changes to the master branch (v1) 
  * Remove `F.SquaredError` (as it is confusing for users both `SquaredError` and `SquaredDifference` exist).
  * Leave `F.MeanSquaredError` and `F.mean_squared_error`as it is (as it is better to leave `F.MeanSquaredError` to keep backward compatibility , we should implement `F.mean_squared_error` with `F.MeanSquaredError`, rather than `F.SquaredDifference` because we can expect better performance). 
  * Add `F.AbsoluteDifference` and `F.absolute_difference`under `chainer/functions/math/` to align it with `F.SquaredDifference` and `F.squared_difference`.
* We do not do any special thing to `_v2` branch and wait v1.24.0 (will be released on 9th May) to be merged to `_v2`.

Hakuyume:
@delta2323 I can understand it is confusing that both `SquaredError` and `SquaredDifference` exist. And I agree with you that it is better to remove `SquaredError` for v1 because `SquaredDifference` was already released.
However, for v2, we can choose `SquaredError` or `SquaredDifference`. I think `SquaredError` is better because it will be mainly used as a loss function. How do you think of it?

delta2323:
Thank you for your opinion. I agree that `SquaredError` is more popular than `SquaredDifference`. My  concern is about backward compatibility. Although we can break it when upgrading to v2, we do not have to do it unnecessarily. But I am also thinking v2 could be a good timing for renaing.

@unnonouno How do you think of it?

Hakuyume:
In the term of a good timing for renaming, `hinge` and `huber_loss` seem inconsistent to me. `hinge_loss/huber_loss` or `hinge/huber` are better. It is not a big problem, but it is a good time to discuss.

delta2323:
> In the term of renaming, hinge and huber_loss seem inconsistent to me.

Thank you for suggestion. I agree with fixing this inconsistency. One suggestion is to create `hinge_loss`  and issue `DeprecationWarning` when `hinge` is used (or do the same thing to `huber` and `huber_loss`).

delta2323:
Regarding the `SquaredError`. We discussed internally and came up the idea that we make `SquaredDifference` deprecated and encourage users to use `SquaredLoss` instead. How do you think of it?

Hakuyume:
@delta2323 from my understanding, your idea is shown as below. Is it right?
 
|| current | newer v1  | v2 | 
|:-:|:--:|:-:|:-:|
|`MeanSquaredError`|present|present|present but rewritten with`Average` + `SquaredError` https://github.com/pfnet/chainer/pull/2562#issuecomment-295076714|
|`SquaredError`|absent|present|present|
|`SquaredDifference`|present|present|present but deprecated https://github.com/pfnet/chainer/pull/2562#issuecomment-298429326|
|`MeanAbsoluteError`|present|present|present but rewritten with`Average` + `AbsoluteError`
|`AbsoluteError`|absent|present|present|
|`AbsoluteDifference`|absent|present https://github.com/pfnet/chainer/pull/2562#issuecomment-297619767|present but deprecated|

I think this is better.

|| current | newer v1  | v2 | 
|:-:|:--:|:-:|:-:|
|`MeanSquaredError`|present|present|present but rewritten with`Average` + `SquaredError`|
|`SquaredError`|absent|present|present|
|`SquaredDifference`|present|**present but deprecated**|present but deprecated|
|`MeanAbsoluteError`|present|present|present but rewritten with`Average` + `AbsoluteError`
|`AbsoluteError`|absent|present|present|
|`AbsoluteDifference`|absent|**absent**|**absent**|

delta2323:
Thank you for organizing information. I prefer your proposal, too.

Hakuyume:
I think rewriting `MeanSquaredError` and making `SquaredDifference` deprecated should be in other PRs. This PR is finished.

delta2323:
`SquaredError`, which is implemented by #2566, will be available in v1.24.0 and v.1.24.0 will be merged to v2 after its release. As this PR is identical to #2566, we need not to merge this PR to v2. We can close this PR and wait v1.24.0 to be merged to v2. Let me know if I misunderstand something I will reopen the PR if needed. I really appreciate your corporation.

PR #2566, opened 2017-04-15
SquaredError

Hakuyume:
`MeanSquaredError` without reduction. The new function is named `SquaredError`. The original `MeanSquaredError` is kept as is. This PR is related to #2558.

### DIFF
--- a/chainer/functions/__init__.py
+++ b/chainer/functions/__init__.py
@@ -73,6 +73,7 @@ from chainer.functions.loss import mean_squared_error  # NOQA
 from chainer.functions.loss import negative_sampling  # NOQA
 from chainer.functions.loss import sigmoid_cross_entropy  # NOQA
 from chainer.functions.loss import softmax_cross_entropy  # NOQA
+from chainer.functions.loss import squared_error  # NOQA
 from chainer.functions.loss import triplet  # NOQA
 from chainer.functions.loss import vae  # NOQA  # NOQA
 from chainer.functions.math import average  # NOQA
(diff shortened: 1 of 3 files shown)


delta2323:
One of the motivation of the repair work of loss functions is inconsistency of strategies for reduction  along samples (i.e. axis=0). Some functions take sum of loss functions of each samples, while others take their mean. That is confusing for users because they how to refer documents or implementation which strategy is taken.

So, could you consider adding the option `reduce` to `MeanSquaredError` (and `MeanAbsoluteError`) to switch how to reduce along the first axis? It could be better to send it as another PR and to change the name of this PR.

delta2323:
Aside from my previous comment, `SquaredError` itself looks good to me. I am running GPU test.

Hakuyume:
@delta2323 I made three designs for `reduce` option. I guess you meant the third design. Is it right?

1. `reduce` option takes only `'mean'`.
    `reduce` is only for consistency. Users can know the reduction strategy from this option. I mentioned this idea at https://github.com/pfnet/chainer/issues/2558#issuecomment-294256374.
1. `reduce` option takes `'mean'` or `'sum'`.
    Users can choose reduction strategy. However, It looks odd because this is *mean*_squared_error.
1. `reduce` option takes `mean_along_samples` and `mean_samplewise`.
    This is the idea you mentioned at https://github.com/pfnet/chainer/issues/2558#issuecomment-294271426.

delta2323:
> I guess you meant the third design. Is it right?

Yes. And I think we need to take care the option name so that it is understandable to users.

delta2323:
I ran GPU tests and confirmed all tests passed. So, I merge`SquaredError`. Aside from that let's keep on discussing `MeanSquaredError` and `MeanAbsoluteError`.

PR #2573, opened 2017-04-17
Improve the document of Hinge loss

delta2323:
This PR fixes the document of `F.hinge_loss`. Mainly, inconsistency between function's arguments and mathematical definition in the document and several markups.


mitmul:
LGTM

PR #2585, opened 2017-04-18
[WIP] Supply -fopenmp flag on compile time.

takagi:
This PR follows #2487. In #2487, `-fopenmp` flag for cusolver is supplied on link time. This PR fixes it to supply on compile time.

Without this fix, MSVC warns as following:

    LINK : warning LNK4044: オプション '/openmp' は無効です。無視されます。


takagi:
It is said that with this change, a linux environment causes a link error on its run time.

takagi:
Generated .cpp files need to be deleted to re-build when the compiler flag is changed.

takagi:
I close this.

PR #2590, opened 2017-04-18
Improve docs of convolution_nd

keisuke-umezawa:

Modified functions/links:
 - functions/connection/convolution_nd

### DIFF
--- a/chainer/functions/connection/convolution_nd.py
+++ b/chainer/functions/connection/convolution_nd.py
@@ -397,8 +397,8 @@ def convolution_nd(x, W, b=None, stride=1, pad=0, use_cudnn=True,
         - The number of spatial dimensions is more than one.
         - ``cover_all`` is ``False``
         - The input's ``dtype`` is equal to the filter weight's.
-        - The ``dtype`` is FP16 (available when cuDNN version :math:`\\geq`
-          v3), FP32 or FP64
+        - The ``dtype`` is FP16, FP32 or FP64. (FP16 is only available when
+          cuDNN version :math:`\\geq` v3.)
 
     .. seealso:: :class:`~chainer.links.ConvolutionND`, :func:`convolution_2d`
 
--- a/chainer/functions/connection/convolution_nd.py
+++ b/chainer/functions/connection/convolution_nd.py
@@ -391,7 +391,7 @@ def convolution_nd(x, W, b=None, stride=1, pad=0, use_cudnn=True,
 
         This function uses cuDNN implementation for its forward and backward
         computation if ALL of the following conditions are satisfied:
-    
+
         - ``cuda.cudnn_enabled`` is ``True``
         - ``use_cudnn`` is ``True``
         - The number of spatial dimensions is more than one.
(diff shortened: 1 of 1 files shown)


chainer/functions/connection/convolution_nd.py
```diff
@@ -349,21 +384,56 @@ def convolution_nd(x, W, b=None, stride=1, pad=0, use_cudnn=True,
...
+        This function uses cuDNN implementation for its forward and backward
+        computation if ALL of the following conditions are satisfied:
+
+        - ``cuda.cudnn_enabled`` is ``True``
+        - ``use_cudnn`` is ``True``
+        - The number of spatial dimensions is more than one.
+        - ``cover_all`` is ``False``
+        - The input's ``dtype`` is equal to the filter weight's.
+        - The ``dtype`` is FP32, FP64 or FP16(cuDNN version is equal to or
+          greater than v3)
```
delta2323: Does it mean if dtype is FP16, cuDNN version should be >= v3 and if FP32 and FP64, its version need not to be?

chainer/functions/connection/convolution_nd.py
```diff
@@ -328,18 +328,53 @@ def convolution_nd(x, W, b=None, stride=1, pad=0, use_cudnn=True,
...
+    Patches are extracted at positions shifted by multiples of ``stride`` from
+    the first position ``(-p_1, -p_2, ..., -p_N)`` for each spatial axis.
+
+    Let :math:`(s_1, s_2, ..., s_N)` be the stride of filter application.
+    Then, the output size :math:`(l_1, l_2, ..., l_N)` is determined by the
+    following equations:
+
+    .. math::
+
+       l_n = (d_n + 2p_n - p_n) / s_n + 1 \ \ (n = 1, ..., N)
```
delta2323: It seems strange to me because it does not depend on `k_n`. Maybe second `p_n` should be replaced with `k_n`.

delta2323:
GPU tests failed because of the test that is not seemingly relevant to this PR. Could you try to merge the latest master branch? The problem can be solved there.

chainer/functions/connection/convolution_nd.py
keisuke-umezawa: Yes, see the below codes.
https://github.com/pfnet/chainer/blob/master/chainer/functions/connection/convolution_2d.py#L20-L22.
It checks `_cudnn_version >= 3000 or x_dtype != numpy.float16`.

By your comments, I recognized the issue. This document has the note for limitation of using cuDNN. But the doc of convolution_2d does not have the note for limitation.
Can I add the same note into the doc of convolution_2d?

delta2323: Thank you. The current description could be misleading because user could think that the condition cuDNN >= v3 must be hold even FP32 or FP64 cases.

delta2323:
Jenkins passed. So, merging the latest branch seems effective.

keisuke-umezawa:
@delta2323 How about this?

delta2323:
LGTM

PR #2598, opened 2017-04-19
Fix fuse with *args without input_num

### LINKED ISSUE #2594: Cupy fusion bug
(not sure if this is supposed to go in github.com/cupy/cupy yet or if this is still the right place)

This code works:
```python
@cupy.fuse()
def sigsbwd(gy, y1, y2):
    for y in (y1, y2):
        gy = gy * y * (1.0 - y)
    return gy
sigsbwd(gy, *ys) # ys is a list of 2 ndarrays
```
but this doesn't (the code generator makes two `v1` variables, one of which is a `&float` and the other a `float`, and nvcc throws an error):
```python
@cupy.fuse()
def sigsbwd(gy, *ys):
    for y in ys:
        gy = gy * y * (1.0 - y)
    return gy
sigsbwd(gy, *ys) # ys is a list of 2 ndarrays
```
I thought these were syntactically equivalent at the Python level, but I guess fusion is doing some fairly low-level Python manipulation and is tripped up by the varargs.

niboshi:
Fix #2594

### DIFF
--- a/tests/cupy_tests/core_tests/test_fusion.py
+++ b/tests/cupy_tests/core_tests/test_fusion.py
@@ -543,8 +543,14 @@ class TestFusionUfunc(unittest.TestCase):
         return numpy.random.rand(10, 10) * (higher - lower) + lower
 
     def check(self, func, n, gen, *args):
+        self._check(func, n, gen, *args, omit_nin=True)
+        self._check(func, n, gen, *args, omit_nin=False)
 
-        @cupy.fuse(input_num=n)
+    def _check(self, func, n, gen, *args, omit_nin=False):
+
+        nin = n if not omit_nin else None
+
+        @cupy.fuse(input_num=nin)
         def f(*x):
             return func(*x)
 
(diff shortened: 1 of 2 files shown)


tests/cupy_tests/core_tests/test_fusion.py
```diff
@@ -543,8 +543,16 @@ def random_real(self, lower=-1000, higher=1000):
         return numpy.random.rand(10, 10) * (higher - lower) + lower
 
     def check(self, func, n, gen, *args):
+        self._check(func, n, gen, *args, omit_nin=True)
+        self._check(func, n, gen, *args, omit_nin=False)
 
-        @cupy.fuse(input_num=n)
+    def _check(self, func, n, gen, *args, **kwargs):
```
okuta: How about this code?
```python
def check(self, func, n, gen, *args):
    self._check(func, n, gen, args, True)

def _check(self, func, n, gen, args, omit_nin):
    nin = n if not omit_nin else None
```

tests/cupy_tests/core_tests/test_fusion.py
```diff
@@ -569,8 +577,16 @@ def f(*x):
             numpy.testing.assert_array_almost_equal(n, fc.get())
 
     def check_reduce(self, func, n, reduce_f, gen, *args):
+        self._check_reduce(func, n, reduce_f, gen, *args, omit_nin=True)
+        self._check_reduce(func, n, reduce_f, gen, *args, omit_nin=False)
+
+    def _check_reduce(self, func, n, reduce_f, gen, *args, **kwargs):
```
okuta: ditto

niboshi:
Thank you for review! Fixed it.

PR #2616, opened 2017-04-22
Improve docs of deconvolution_2d

keisuke-umezawa:

Modified functions/links:
 - functions/connection/deconvolution_2d

### DIFF
--- a/chainer/functions/connection/deconvolution_2d.py
+++ b/chainer/functions/connection/deconvolution_2d.py
@@ -325,8 +325,15 @@ def deconvolution_2d(x, W, b=None, stride=1, pad=0,
                      outsize=None, use_cudnn=True, deterministic=False):
     """Two dimensional deconvolution function.
 
-    This is an implementation of two-dimensional deconvolution (alse known as
-    transposed convolution).
+    This is an implementation of two-dimensional deconvolution. In most of deep
+    learning frameworks and papers, this function is called
+    **transposed convolution**. But because of historical reasons (e.g. paper
+    by Ziller `Deconvolutional Networks`_) and backward compatibility, this
+    function is called **deconvolution** in Chainer.
+
+    .. _Deconvolutional Networks: \
+http://www.matthewzeiler.com/pubs/cvpr2010/cvpr2010.pdf
+
     It takes three variables: input image ``x``,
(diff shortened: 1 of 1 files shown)


chainer/functions/connection/deconvolution_2d.py
```diff
@@ -325,21 +325,49 @@ def deconvolution_2d(x, W, b=None, stride=1, pad=0,
                      outsize=None, use_cudnn=True, deterministic=False):
     """Two dimensional deconvolution function.
 
-    This is an implementation of two-dimensional deconvolution.
+    This is an implementation of two-dimensional deconvolution (alse known as
```
delta2323: typo: also

chainer/functions/connection/deconvolution_2d.py
```diff
@@ -325,21 +325,49 @@ def deconvolution_2d(x, W, b=None, stride=1, pad=0,
...
+    transposed convolution).
```
delta2323: According to [the tutorial in Theano](http://deeplearning.net/software/theano/tutorial/conv_arithmetic.html#about-this-tutorial) it is a misusage to call transposed convolution deconvolution. TensorFlow also [mentions the misusage](https://github.com/tensorflow/tensorflow/blob/4c3b060b9b063a247172c939e0e1314577b581e3/tensorflow/python/ops/nn_ops.py#L46).

keisuke-umezawa: Yes, they said that name of deconvolution is not good.
TensorFlow have named this functions as `conv2d_transpose`.
So, it is better to rename deconvolution functions, but it will break
the backward compatiblity.
Please ask if we should rename the functions or not.

delta2323: Yes, we cannot do that in at least `v1`. But it is confusing for users if we write transposed convolution is an alias of deconvolution. So, it is better to write explicitly that in most of deep learning frameworks and papers, this is called transposed convolution but for historical reasons,  (e.g. paper by Ziller [here](http://www.matthewzeiler.com/pubs/cvpr2010/cvpr2010.pdf)) we call it deconvolution.

delta2323:
LGTM

PR #2621, opened 2017-04-24
Recompute (v2)

anaruse:
This PR implements a recompute feature proposed in #2349 for v2, and is related to the PR #2452.

As an aggressive buffer release feature has been implemented in #2452, memory usage of v2 are greatly reduced compared with v1 w/o performance penalties. However, you can further reduce memory usage by using the recompute feature but at expense of some performance penalties.

Like #2349, this branch allow you to select target functions, and outputs of specified target functions will be forgotten during forward phase (in order to reduce memory usage) and re-computed soon before they are necessary during backward phase. You can select target functions using a method `use_recompute()` as below.

    with chainer.use_recompute("BN", "RELU", "CONV"):
        trainer.run()

Now, the recompute feature supports four functions. When "BN" is specified in use_recompute(), **batch normalization** functions become target of the recompute feature. When "RELU", **relu and leaky_relu** functions. When "CONV", **convolution_2d** functions.

The graph below shows the effect of the recompute feature measured on Tesla P100. The vertical axis shows training performance (images per second) and the horizontal axis shows batch size. As you can see, when recompute feature is not used (blue line), ResNet152 can be trained at batch size of up-to 80, but it can not be trained at batch size more than 80 due to lack of memory. When recompute feature is used and "BN", "RELU" and "CONV" are selected as target functions (green line), ResNet152 can be trained at batch size of up-to 190 but at expense of some performance penalties. This indicates that memory usage can be greatly reduced with the recompute feature and training of much larger models might be feasible by the recompute feature. 

image

Note: this branch is implemented on top of #2586.


anaruse:
The base branch was changed from _v2 to master and several conflicts to the base branch was resolved.

niboshi:
I'm sorry for being late to review.

@beam2d I want to consult about the design.

- I feel the way to configure the feature (`use_recompute`) is not straight-forward.

1. abbreviations like `"BN"` is a bit arbitrary (it doesn't correspond to actual function name `BatchNormalizationFunction`)
2. It's not 1-vs-1 correspondence (`"RELU"` corresponds to both `LeakyReLU` and `ReLU`)
3. Can't select specific function instance from a network with multiple functions of same type.

Are these acceptable?
To fix (3), I think we need to pass around a flag (like old `train` flag) to each of link/function instances, which might be cumbersome, though.
If we take this flag approach, IMO it's better to introduce a named argument to `Function` base class, and contain related modification within the base class. In this case, existing functions need to be modified to call `Function.__init__()` explicitly in order to enable this feature.

anaruse:
Thank you for the feedback, @niboshi !

Regarding the feedback 1. and 2., I also feel it is ambiguous to use `"BN"` for `BatchNormalizationFunction` and `"RELU"` for `ReLU` and `LeakyReLU`. What do you think it is better to use function names themselves like `BatchNormalizationFunction` as an input to `use_recompute()`?

Regarding the feedback 3., I think the suggestion to add a flag to each link/function class is a bit against a policy of Chainer v2 that you should use global/local configuration and should not add the flag as much as possible.

Even with current design, you can select whether to enable or disable recompute feature for each function by using `use_recompute()` and `no_recompute()` as below.

    with chainer.use_recompute("CONV"):
        ...
        h = conv1(x)
        with chainer.no_recompute("CONV"):
            h = conv2(h)
        y = conv3(h)
        ...

As you can see, `use_recompute()` and `no_recompute()` can be used in a nested way.  In the case above, recompute is enabled kinda globally but disabled locally, so that outputs of `conv1()` and `conv3()` are recomputed and output of `conv2()` is not recomputed.

BTW, I now feel that the names of `use_recompute()` and `no_recompute()` are not intuitive. `enable_recompute()` and `disable_recompute()` might be better?

niboshi:
>Regarding the feedback 1. and 2., I also feel it is ambiguous to use `"BN"` for `BatchNormalizationFunction` and `"RELU"` for `ReLU` and `LeakyReLU`. What do you think it is better to use function names themselves like `BatchNormalizationFunction` as an input to `use_recompute()`?

How about using class object instead of string? This way there's no ambiguity and you can take advantage of IDE support as well.

>Even with current design, you can select whether to enable or disable recompute feature for each function by using `use_recompute()` and `no_recompute()` as below.

Oh, I didn't notice it's possible with current design. Then it's no problem here.

>BTW, I now feel that the names of `use_recompute()` and `no_recompute()` are not intuitive. `enable_recompute()` and `disable_recompute()` might be better?

Possibly. Or how about switching enable/disable with a parameter?
e.g. 
```
with chainer.enable_recompute(True, "CONV"):
     ...
```

I also wonder if the term "recompute" is appropriate or not. The term implies `backward` will be called, but I guess this feature is also useful without actual recomputation, isn't it?
In my understanding, this feature is effectively cancellation of `retain_inputs`. So I feel some word that explains this behavior might be better. I don't come up with good wording, though... do you have any idea?
Please point out if my understanding is not correct.

anaruse:
> How about using class object instead of string? This way there's no ambiguity and you can take advantage of IDE support as well.

Basically, I think it is a good idea to use a class object. But, it is annoying what object (or instance) to use.

From an ease-of-maintenance viewpoint, function class object should be used as an argument of `use/no_recompute()` mainly because it is decided in a function class whether to enable recompute feature or not.

However, I think it is not intuitive for users. For example, a name of convolution function class is `Convolution2DFunction` and a name of relu function class is `ReLU`, but these names are not typically used in model description. The following is a snippet of AlexNet in Chainer's example and `L.Convolution2D` is used to represent convolution layer and `F.relu` is used to represent relu layer in it. From an ease-of-use viewpoint, these name should be used as an argument of `use/no_recompute()`, but there is some difficulty in terms of ease-of-maintenance or program-ability. As you can see, `L.Convolution2D` is a link object (this might be acceptable) and `F.relu` is just a function and not a class object (this is problematic).

    def __init__(self):
        super(Alex, self).__init__()
        with self.init_scope():
            self.conv1 = L.Convolution2D(None,  96, 11, stride=4)
            self.conv2 = L.Convolution2D(None, 256,  5, pad=2)
            ...

    def __call__(self, x, t):
        h = F.max_pooling_2d(F.local_response_normalization(
            F.relu(self.conv1(x))), 3, stride=2)
        h = F.max_pooling_2d(F.local_response_normalization(
            F.relu(self.conv2(h))), 3, stride=2)
        ...

So, I think it would be good trade-off point to use a name or short name of function class as below.

* `Convolution2D` for 2d convolution layers
* `BatchNormalization` for batch-normalization layers
* `ReLU` for relu layers
* `LeakyReLU` for leaky-relu layers

What would you think?

niboshi:
>However, I think it is not intuitive for users. For example, a name of convolution function class is `Convolution2DFunction` and a name of relu function class is `ReLU`, but these names are not typically used in model description.

Ah, you're right. Function class objects are not suitable here as you described.

>So, I think it would be good trade-off point to use a name or short name of function class as below.
>
>* `Convolution2D` for 2d convolution layers
>* `BatchNormalization` for batch-normalization layers
>* `ReLU` for relu layers
>* `LeakyReLU` for leaky-relu layers

This suggestion looks good to me.

anaruse:
The APIs have been changed based on the discussion.
Methods names are also altered from `use/no_recompute` to `enable/disable_recompute`.

beam2d:
Is this feature able to generalize to arbitrary functions? I guess that the essential feature that this PR tries to realize is setting `None` to the data array of a specified VariableNode and recomputing it when the data array is required during backward. I feel it is more direct to let users specify which variables should be recomputed (e.g., either by wrapping a code by `with` block or by explicitly calling a method of the variable, maybe called `forget`). If a certain function is often used with this forgetting feature, it is also ok to add an option (like `relu(x, forget=True)`) or add a variant of the function (like `relu_forget(x)`). I will continue thinking about the possible design improvement.

anaruse:
Please hold this PR until the issue discussed in #2586 and #2912 is resolved, because recompute feature highly depends on it...

PR #2633, opened 2017-04-24
Improve docs of accuracy

keisuke-umezawa:

Modified functions/links:
 - functions/evaluation/accuracy

### DIFF
--- a/chainer/functions/evaluation/accuracy.py
+++ b/chainer/functions/evaluation/accuracy.py
@@ -78,7 +78,7 @@ def accuracy(y, t, ignore_label=None):
 
     .. admonition:: Example
 
-        We show the most common case, when ``y`` is two dimensinal array.
+        We show the most common case, when ``y`` is the two dimensional array.
 
         >>> y = np.array([[0.1, 0.7, 0.2], # prediction label is 1
         ...               [8.0, 1.0, 2.0], # prediction label is 0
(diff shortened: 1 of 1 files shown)


chainer/functions/evaluation/accuracy.py
```diff
@@ -59,9 +59,15 @@ def accuracy(y, t, ignore_label=None):
...
     Args:
-        y (Variable): Variable holding a matrix whose (i, j)-th element
-            indicates the score of the class j at the i-th example.
-        t (Variable): Variable holding an int32 vector of ground truth labels.
+        y (:class:`~chainer.Variable` or :class:`numpy.ndarray` or \
+        :class:`cupy.ndarray`):
+            Array whose (i, j, k, ...)-th element indicates the score of
+            the class j at the (i, k, ...)-th sample.
+            The prediction label :math:`\\hat t` is calculated by the formula
+            :math:`\\hat t_(i, k, ...) = argmax_j y(i, j, k, ...)`.
```
delta2323: Remove the underscore after `t`

delta2323: I think we usually do not make `argmax` italic. How about changing it to  `\operatorname{\mathrm{argmax}}` ?

keisuke-umezawa: Thank you for the advice. I will fix them.

keisuke-umezawa:
By the way, the input `t` must be int32 type. I think it is better to hold other integer type. How do you think? If you think the priority of this issue is higher, I will fix them.

delta2323:
I believe it is only because initial version was set and left like this, thinking that we can broaden its dtype afterwards. But I am not perfectly confident.

keisuke-umezawa:
Ok. So, in this PR, I just describe the current functionality.

delta2323:
LGTM

PR #2637, opened 2017-04-25
Fix broken alex_fp16 example.

keisukefukuda:
Since "bias" keyword parameter  in L.Convolution2D.__init__() is removed, alex_fp16 example is broken.

```
  File "examples/imagenet/train_imagenet.py", line 205, in <module>
    main()
  File "examples/imagenet/train_imagenet.py", line 132, in main
    model = archs[args.arch]()
  File "/home/kfukuda/chainermn/examples/imagenet/alex_v2.py", line 58, in __init__
    stride=4, initialW=W, bias=bias),
TypeError: __init__() got an unexpected keyword argument 'bias'
```

I think the `bias` arguments can be simply removed in general cases.
In this case, however, FP16 dtype is used explicitly. So I replaced it with `initial_bias`.
(I'm not confident...)

### DIFF
--- a/examples/imagenet/alex.py
+++ b/examples/imagenet/alex.py
@@ -54,14 +54,18 @@ class AlexFp16(Alex):
         chainer.Chain.__init__(
             self,
             conv1=L.Convolution2D(None, 96, 11,
-                                  stride=4, initialW=W, bias=bias),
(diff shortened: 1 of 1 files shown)


okuta:
LGTM!

PR #2642, opened 2017-04-25
Modify docstring in connection

ysekky:
In some module, in_channel/in_size parameter accepts None, but only int is described in docstring.

This PR add "or None" to docstring.

```
$ env/bin/pep8 chainer/links/connection/*
$
```


mitmul:
LGTM

PR #2644, opened 2017-04-26
Add resnet50 example

keisukefukuda:
* Added ResNet50 example, originally written by Kudo-san
* Modified train_imagenet.py accordingly.

The result is:

    epoch       iteration   main/loss   validation/main/loss  main/accuracy  validation/main/accuracy  lr
    2           1000        1.79846                           0.377313                                 0.01
    4           2000        1.28585                           0.525031                                 0.01
    7           3000        1.13306                           0.581438                                 0.01
    9           4000        1.02884                           0.61925                                  0.01

### DIFF
--- a/examples/imagenet/resnet50.py
+++ b/examples/imagenet/resnet50.py
@@ -109,15 +109,14 @@ class ResNet50(chainer.Chain):
 
     def __call__(self, x, t):
         self.clear()
-        with chainer.using_config('train', self.train):
-            h = self.bn1(self.conv1(x))
-            h = F.max_pooling_2d(F.relu(h), 3, stride=2)
-            h = self.res2(h)
-            h = self.res3(h)
-            h = self.res4(h)
-            h = self.res5(h)
-            h = F.average_pooling_2d(h, 7, stride=1)
-            h = self.fc(h)
+        h = self.bn1(self.conv1(x))
+        h = F.max_pooling_2d(F.relu(h), 3, stride=2)
+        h = self.res2(h)
+        h = self.res3(h)
+        h = self.res4(h)
+        h = self.res5(h)
+        h = F.average_pooling_2d(h, 7, stride=1)
+        h = self.fc(h)
 
         loss = F.softmax_cross_entropy(h, t)
         chainer.report({'loss': loss, 'accuracy': F.accuracy(h, t)}, self)
--- a/examples/imagenet/resnet50.py
+++ b/examples/imagenet/resnet50.py
@@ -1,0 +1,124 @@
+# Original author: yasunorikudo
+# (https://github.com/yasunorikudo/chainer-ResNet)
+
+import chainer
+import chainer.functions as F
+from chainer import initializers
+import chainer.links as L
+
+
+class BottleNeckA(chainer.Chain):
+
+    def __init__(self, in_size, ch, out_size, stride=2):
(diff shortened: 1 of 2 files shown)


examples/imagenet/resnet50.py
```diff
@@ -0,0 +1,123 @@
...
+
+    def __call__(self, x):
+        h = F.relu(self.bn1(self.conv1(x)))
+        h = F.relu(self.bn2(self.conv2(h)))
+        h = self.bn3(self.conv3(h))
+
+        return F.relu(h + x)
+
+
+class Block(chainer.Chain):
```
beam2d: I feel it's better to inherit `ChainList` instead of `Chain`, and just pass each bottleneck child link to `add_link` one by one in `__init__`. Then, the `__call__` is also made simple.

examples/imagenet/resnet50.py
```diff
@@ -0,0 +1,123 @@
...
+                                  initialW=initializers.HeNormal(),
+                                  nobias=True),
+            bn1=L.BatchNormalization(64),
+            res2=Block(3, 64, 64, 256, 1),
+            res3=Block(4, 256, 128, 512),
+            res4=Block(6, 512, 256, 1024),
+            res5=Block(3, 1024, 512, 2048),
+            fc=L.Linear(2048, 1000),
+        )
+        self.train = True
```
beam2d: Remove this unused attribute.

examples/imagenet/resnet50.py
```diff
@@ -0,0 +1,123 @@
...
+
+    def clear(self):
+        self.loss = None
+        self.accuracy = None
+
+    def __call__(self, x, t):
+        self.clear()
```
beam2d: Remove this line because `self.loss` and `self.accuracy` are not used. The `clear` method itself can also be removed.

examples/imagenet/resnet50.py
mitmul: @keisukefukuda This tutorial might be informative to know how to use `ChainList` for this case: http://docs.chainer.org/en/latest/tutorial/convnet.html#resnet152

keisukefukuda:
I reflected the comments from @beam2d -san and @mitmul -san!

beam2d:
LGTM. It runs in my environment. @mitmul Could you confirm the fix and merge it if it's ok?

examples/imagenet/resnet50.py
```diff
@@ -0,0 +1,111 @@
...
+
+class ResNet50(chainer.Chain):
+
+    insize = 224
+
+    def __init__(self):
+        super(ResNet50, self).__init__(
+            conv1=L.Convolution2D(3, 64, 7, 2, 3,
+                                  initialW=initializers.HeNormal(),
+                                  nobias=True),
```
mitmul: I don't know why (and feel it's strange) but the first convolution has bias term in the original implementation. See the original prototext here: https://github.com/KaimingHe/deep-residual-networks/blob/master/prototxt/ResNet-50-deploy.prototxt

keisukefukuda: Thank you for the advice. I fixed it!

mitmul:
Somehow the last commit wasn't merged (why? it's creepy...) so I submit a PR to fix that #2655

PR #2646, opened 2017-04-26
[WIP] Add on_trigger option to trainer.extend

mitmul:
This PR adds a new option to `trainer.extend` method for an extension to be a signal handler.
Will add tests.

### DIFF
--- a/chainer/training/trainer.py
+++ b/chainer/training/trainer.py
@@ -1,5 +1,6 @@
 import collections
 import os
+import signal
 import time
 
 import six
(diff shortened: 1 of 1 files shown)


mitmul:
Hmm, should `trigger` argument of `trainer.extend` support a SIGNAL additionally to a tuple?

beam2d:
I feel it's better to just add a new method rather than extending the `extend` method because they have different roles (the name might be `signal`, `call_on_signal`, `add_signal_handler`, etc.). It would also be better to set up the real signal handler at the beginning of `run` method and recover the original handler at the end of the training. Is it also possible to allow multiple extensions to be registered for the same signal?

PR #2648, opened 2017-04-27
Fix spatial_transformer_sampler to behave consistently among cuDNN and non-cuDNN mode

yuyu2172:
This PR follows a discussion in #2458.

The current implementation of `spatial_transformer_sampler` in cuDNN mode and non-cuDNN mode are different.
This happens in the case when grid is outside the range [-1, 1].
In the implementation of cuDNN, points are sampled from a zero padded image when this happens.
However, in the current implementation of non cuDNN mode, grid is clipped to range [-1, 1].
Since creating difference between cuDNN and non-cuDNN was not intended, I would like to fix it with this PR.

This also fixes a problem that the documentation is not reflecting the implementation of cuDNN.

The tests became less stable so the tolerance for error is loosened even for cuDNN implementation when grid is outside the range [-1, 1].
Also, I added a test to verify that cuDNN and non cuDNN implementation produce almost same backward and forward passes.

### DIFF
--- a/chainer/functions/array/spatial_transformer_sampler.py
+++ b/chainer/functions/array/spatial_transformer_sampler.py
@@ -74,7 +74,7 @@ class SpatialTransformerSampler(function.Function):
         u = grid[:, 0]
         v = grid[:, 1]
 
-        # Pad the image so that points locating outside of the original
+        # Pad the image so that pixels locating outside of the original
         # image's size can be sampled.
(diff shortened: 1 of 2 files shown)


yuyu2172:
It is a little bit problematic that implementation changes from the already merged one.
However, the original `spatial_transformer_sampler` has not been included in any release yet as it was merged after the latest version (1.23) was released.

mattya:
LGTM!

Issue #1970, opened 2016-12-06
IDE-friendly registration of links, parameters, and persistent values


beam2d:
This is a proposal of new interface of registering links, parameters, and persistent values to links and chains. The current interface (add_param, add_persistent, and add_link) automatically adds an attribute to the link object, which is harmful for writing codes with IDE like PyCharm, since the attribute is not recognized by them. I want to support these objects explicitly substituted to the attributes so that PyCharm can recognize them, and then we can utilize the code-completion and type checking (though the latter needs type hinting annotations to the Chainer's internal code).

The detailed design is described here: https://docs.google.com/document/d/1SS-N8kT-2tZpKnrKBKeVf4tSE_QM9YhsbC4SRe4Klf0/edit?usp=sharing. I'm wondering whether we should deprecate the existing methods or not. I'm planning to make this change for v2, though the compatibility-breaking change is small and it might be ok to include other parts to v1.

jekbradbury:
There's another option: just override `Chain.__setattr__` so it checks if the object being added is an instance of `Link`; if it is, add it as a sub-link; if it isn't, add it as an ordinary Python attribute. This way it would be possible to write `__init__` functions like this:

```python
def __init__(self, size):
    super().__init__()
    self.l1 = L.Linear(size, size)
    self.l2 = L.Linear(size, 2)
```

beam2d:
That looks much more natural than one I gave. One concern I have is whether there is a use case that one wants to add a link/variable object as an attribute without putting it into the link/parameter hierarchy. A workaround exists for this case: wrapping the link/variable object by another object which is not an instance of Link/Variable.

If the above option is acceptable, is it enough to just iterate over `__dict__` when one wants to enumerate all child links or all parameters? I'm thinking about reorganizing the optimization stuffs (in parallel to this issue), and if we can avoid enumerating all parameters in each iteration, this option might be acceptable, too.

jekbradbury:
That sounds like it should work, although another option is as follows:

Keep using a `_children` attribute for iteration, and add to that during `__setattr__`, but don’t add to it if the Link is added using a `add_without_registration` method.

beam2d:
I'm thinking about a slight modification to my original proposal that removes the redundancy of writing the name twice. The link/parameter is first registered without names (e.g. the name is set to the empty string by default). The `__setattr__` method checks if the argument has been registered without names, and in that case it changes the name to the attribute name.

delta2323:
Fixed via #2657.

PR #2651, opened 2017-04-27
Add Parameter

beam2d:
This PR is a part of #1970. In order to distinguish a parameter variable from a non-parameter variable without relying on explicit `add_param` call, I wrote a subclass of `Variable` named `Parameter` (a design similar to PyTorch). I think this is a good timing, so I also refactored the code and move features of `Variable` specific to parameters to `Parameter`. It makes the implementation and API of `Variable` much simpler.

I will work on #1970 using this Parameter class. Any comments are welcome.

### DIFF
--- a/chainer/__init__.py
+++ b/chainer/__init__.py
@@ -50,6 +50,7 @@ from chainer.reporter import Summary  # NOQA
 from chainer.serializer import AbstractSerializer  # NOQA
 from chainer.serializer import Deserializer  # NOQA
 from chainer.serializer import Serializer  # NOQA
+from chainer.variable import Parameter  # NOQA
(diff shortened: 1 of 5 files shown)


beam2d:
I found that I mistakenly added changes to the Parameter in #2657. Maybe it is better to directly review the Parameter part in #2657 as well.

delta2323:
Do we need this PR if we adopt `Link.init_scope()` and `Link.forward_scope()` suggested in #2657?

delta2323:
Could you close this PR if it is not needed.

beam2d:
This change is included in #2657 and that one is more advanced than this PR, so I'll close this PR.

Issue #2052, opened 2016-12-26
Allow duplicated registration of params/links to links/chains


beam2d:
Registering a parameter/link to multiple links/chains is currently not allowed to simplify the management of the link hierarchy. It makes implementations of some complicated networks difficult and less portable, since in some cases it is essential to reuse the same network multiple times from different models. I think it's time to relax that condition with some best practices which should be explicitly noted in the document. See also #1967 and #1970, which are also related to the design of links and parameters.

PR #2657, opened 2017-04-28
Add new APIs of parameter/link registration to Link/Chain

### LINKED ISSUE #1970: IDE-friendly registration of links, parameters, and persistent values

beam2d:
Fix #1970. This PR implements `Link.__setattr__` and `Chain.__setattr__` to automatically register parameters and child links. It also includes changes to existing links both in the chainer package and the examples.

It uses the `Parameter` class introduced by #2651, so please review and merge that first.


niboshi:
If a user put a `Parameter` as a private (maybe temporary) attribute, will it be treated as a new parameter?
```
self._tmp = self.param1
```
Perhaps we can check for duplicate parameter in `__setattr__` and ignore if found. But then, what if a user want shared parameters?
```
self.param1 = self.param2 = Parameter(...)
```

but

niboshi:
Mmm, I feel it's too much exploit of python mechanism.
The concern is, if a user put a parameter in a private (maybe temporary) attribute, the new parameter will be unexpectedly created.
```
self._tmp = self.param1
```
Perhaps we can check for duplicate parameter in `__setattr__` and ignore if found.
But then, what if a user wants shared parameters?
```
self.param1 = param2 = Parameter(...)
```

IMO, I'd rather prefer your original proposal, which is more verbose for users, but less concern.
How do you think?

jekbradbury:
The MXNet team is working on adding define-by-run functionality with a Chainer-inspired API. They have yet another different idea (I think a pretty clever one) for parameter/link registration.
The concept (introduced [here](https://github.com/dmlc/mxnet/pull/5705#discussion_r116655131)) is to add a new global/thread-local configuration object that defines the link that things are currently being registered to, and enter this state with a `with` block:

```
class Net(Chain):
    def __init__(self):
        super().__init__()
        with self:
            self.lin1 = L.Linear(...)
```
or wrap the entire init function in this block automatically like this:
```
class Link(object):
    def __init__(self, *args, **kwargs):
        with self:
            self.init(*args, **kwargs)
    ...

class Net(Chain):
    def init(self):
        self.lin1 = L.Linear(...)
```
The benefit this has over PyTorch's automatic registration by overloading `__setattr__` is that it automatically works for lists of links/params now too (this frequently [trips up](https://discuss.pytorch.org/t/generate-label-score-results-for-nllloss2d/2880/4) new PyTorch users):
```
class Net(Link):
    def init(self):
        self.params = [Parameter(...), Parameter(...)]
```

niboshi:
Wow, I love that idea. Maybe we should give `init` more special name, to not users think it's for general initialization. (If a user does `self._tmp = self.param1` in this method, the same problem happens.)

beam2d:
The idea given by @jekbradbury looks good! Thank you for the suggestion. I think adding a method like `init` would be confusing (regardless of the naming) because users still have to write their own `__init__` to customize the argument list. I feel it's better to use a decorator if we don't want to use `with` statement explicitly in `__init__`.

jekbradbury:
In my proposal `__init__` would forward arguments to `init` much like `__call__` does to `forward` in PyTorch; of course, this would introduce a new IDE-unfriendly behavior because IDEs would display (*args, **kwargs) as the signature of `__init__` rather than the user-created `init` signature.

beam2d:
Oh, I see. I misunderstood that point; so Link itself also provides `init` to emulate the current `Link.__init__`.  That makes sense.

beam2d:
Using a metaclass, we can remove the need of using `init` and automatically turn on the registration only in `__init__`. Is that preferable? A simple implementation is like follows:

```py
class LinkMeta(type):
    def __new__(cls, name, bases, namespace):
        if '__init__' in namespace:
            init = namespace['__init__']
            @functools.wraps(init)
            def __init__(self, *args, **kwargs):
                with self:
                    init(self, *args, **kwargs)
            namespace['__init__'] = __init__
        return type.__new__(cls, name, bases, namespace)

@six.add_metaclass(LinkMeta)
class Link(object):
    ...

class MyLink(Link):
    def __init__(self):
        self.W = Parameter(...)  # automatically register W

    def __call__(self, x):
        self.x = x  # does not register x even if x is a Parameter object
        ...
```

Then, `Link.__init__` as well as `__init__` of all subclasses will be wrapped by `with self`.

jekbradbury:
Either way (magic metaclass or user defined method with different name) works, but you should pick one and use it for both `__init__` and `__call__` (i.e., do the same thing here https://github.com/pfnet/chainer/issues/2748). I think I personally prefer magic metaclass.

niboshi:
I'd prefer explicit way (like `with ...` or `init`).
If the mechanism is hidden from users, they can make a mistake unless they read documentation very carefully.
Explicit notation will be a alarm to users that "this is something special, be careful".

niboshi:
And if we adopt `with ...` way, i'd prefer more explanatory notation than `with self`, like the one suggested here: https://github.com/dmlc/mxnet/pull/5705#discussion_r116725561

beam2d:
That also makes sense. OK, let's take a way that is easy to understand for everyone.
```py
def __init__(self, args...):
    with self.registration_scope():
        self.W = Parameter(...)
```

niboshi:
Looks good to me. `registration_scope` might be a bit difficult to spell, but i think it's acceptable.

beam2d:
Okey, I also think it's better to use an easier name. I think it's also better to align this API with #2748 (I now think it's also good to use an explicit `with` statement to wrap `__call__` as well with a different scope method). How about using the following names for them: `self.init_scope()` and `self.forward_scope()`?

takagi:
Another idea for this PR is introducing a Chainer-specific convention that automatic registration works only for (conventional) public attributes and not for (conventional) private attributes, whose name begin with one or more underscores.

    class MyLink(Link):
        def __init__(self, x):
            self.W = Parameter(...)  # automatically register W
            self._x = x              # does not register _x even if x is a Parameter object
            ...

This might well suit to what users intend implicitly in their mind. Of course, registration scope with `with` block is more explicit and might be better in that point.

beam2d:
Such a kind of "Chainer-specific rule" is handy for users that got used to it. What we have to take is the trade-off between such a handiness and the hurdle of learning the API; it is easier to learn an approach that is more verbose and less magical, whereas such a verbose API will be a bit cumbersome for professional users who want to quickly write their own research code that they think does not need to be readable for others. I personally think a good framework should provide a standard way to write a code which is easy to understand for any readers (regardless of the authors' intention), and so I'm choosing the explicit API now.

takagi:
I agree to your addressing and respect your policy. Also, I understand you must make decision on design. My comment was off the point for the latest context.

beam2d:
No problem :) It is good to have as many ideas as possible. Thank you for giving an idea!

delta2323:
Memo: We need to fix tutorials to align with the changes applied in this PR.

chainer/link.py
```diff
@@ -43,10 +44,16 @@ class Link(object):
 
     Link is the primitive structure for the model definitions. It supports
     management of parameter variables and *persistent values* that should be
-    incorporated to serialization. Parameters are variables registered via
-    the :meth:`add_param` method, or given to the initializer method.
+    incorporated to serialization.
+
+    Parameter is an instance of :class:`~chainer.Parameter` registered to a
+    link. A :class:`~chainer.Parameter` object can be registered as a
+    parameter of the link by substituting it to an attribute within *an
```
delta2323: Is assigning rather than substituting easier to understand?

chainer/link.py
```diff
@@ -139,25 +149,83 @@ def xp(self):
...
+        return getattr(self, '_within_init_scope', False)
+
+    @contextlib.contextmanager
+    def init_scope(self):
+        """Creates an initialization scope.
+
+        This method returns a context manager object that enables registration
+        of parameters (and links for :class:`~chainer.Chain`) by substitution.
+        A :class:`~chainer.Parameter` object can be automatically registered
+        by substituting it to an attribute under this context manager.
```
delta2323: ditto (is assigning more appropriate?)

chainer/link.py
```diff
@@ -171,28 +239,21 @@ def add_param(self, name, shape=None, dtype=numpy.float32,
                 initializer. Note that in these cases, ``dtype`` argument is
                 ignored.
 
+        Returns:
+            ~chainer.Parameter: The registered parameter object.
```
delta2323: Why do you change the specification of `add_param` to return the parameter to be registered. As `add_param` is deprecated, I thought we do not have to do that. And actually no code in this PR assumes that ` add_param` returns an instance of `Parameter`.

chainer/link.py
```diff
@@ -43,10 +44,16 @@ class Link(object):
...
+
+    Parameter is an instance of :class:`~chainer.Parameter` registered to a
+    link. A :class:`~chainer.Parameter` object can be registered as a
+    parameter of the link by substituting it to an attribute within *an
+    initialization scope*, which is a code surrounded by a
+    :meth:`init_scope` context manager using the ``with`` statement.
+
     Persistent values are arrays, scalars, or any other serializable values
-    registered via the :meth:`add_persistent` method.
+    registered via the :meth:`register_persistent` method.
```
delta2323: We should add `add_persistent` because `add_persistent` is still a valid method to register persistent, although it is not recommended.

chainer/link.py
```diff
@@ -205,14 +266,36 @@ def add_persistent(self, name, value):
                 for the attribute name.
             value: Value to be registered.
 
+        Returns:
+            The registered value.
```
delta2323: Same as `add_param`. Why do you change the function to return the registered value?

chainer/link.py
```diff
@@ -205,14 +266,36 @@ def add_persistent(self, name, value):
...
             raise AttributeError(
                 'cannot register a new persistent value %s: attribute exists'
                 % name)
-        self._persistent.append(name)
+        self._persistent.add(name)
+        self._params.discard(name)
         d[name] = value
+        return value
+
+    def register_persistent(self, name):
```
delta2323: Just for confirmation. You do not make `add_persistent` deprecated on purpose differently from `add_param`. Is it right?

chainer/link.py
```diff
@@ -500,6 +583,18 @@ class Chain(Link):
     like a file path in UNIX, consisting of names of nodes on the path, joined
     by slashes ``/``.
 
+    A child link can be added just by substituting it to an attribute of the
```
delta2323: ditto (Is assigning more appropriate?)

chainer/link.py
```diff
@@ -552,35 +650,57 @@ def __getitem__(self, name):
...
     def add_link(self, name, link):
         """Registers a child link to this chain.
 
-        The registered link is saved and loaded on serialization and
-        deserialization, and involved in the optimization. The registered link
-        is called a child. The child link is set to an attribute of the chain
-        with the given name.
+        .. deprecated:: v2.0.0
+
+           Substitute the child link directly to an attribute within
```
delta2323: ditto (Is assigning more appropriate?)

chainer/variable.py
```diff
@@ -796,6 +771,142 @@ def add_cand(cand):
...
+    __array_priority__ = 200
+
+
+class Parameter(Variable):
+
+    """Parameter variable that can be registered to a link.
+
+    Parameter is a subclass of :class:`Variable`. It almost behaves as same
+    as a usual variable except that a parameter can be registered to a
+    :class:`~chainer.Link` object just by substituting it to an attribute of
```
delta2323: ditto (Is assigning more appropriate?)

delta2323: ditto (Is assigning more appropriate?)

chainer/variable.py
```diff
@@ -796,6 +771,142 @@ def add_cand(cand):
...
+    the link.
```
delta2323: ... within `Link.init_scope()` context.

chainer/variable.py
```diff
@@ -796,6 +771,142 @@ def add_cand(cand):
...
+        initializer (~chainer.Initializer or numpy.ndarray or cupy.ndarray):
+            Initializer of the data array. If ``shape`` is given, this
+            initializer is immediately used to initialize the data array.
+            Otherwise, if it is an array, it is immediately used as the data
+            array, and otherwise the data array is left uninitialized and will
+            be initialized by this initializer in :meth:`initialize`. It can
+            also be a scalar, in which case the data array will be filled by
+            this scalar. Note that float32 is used in this case.
+        shape (int or tuple of int or None): Shape of the parameter. If it is
+            omitted, the initialization is deferred to the call of
```
delta2323: "If it is `None`" is more precise.

tests/chainer_tests/test_variable.py
```diff
@@ -567,24 +567,58 @@ def test_pickle_gpu(self):
...
 
+    def test_initialize_by_scalar(self):
+        x = chainer.Parameter(2., (3,))
+        np.testing.assert_array_equal(x.data, np.array([2., 2., 2.]))
+
+    def test_initialize_by_initializer(self):
+        x = chainer.Parameter(initializers.One(), (3,))
+        np.testing.assert_array_equal(
+            x.data, np.array([1., 1., 1.], dtype='f'))
+
```
delta2323: Add an test that initialization by `None` and `numpy.ndarray`.

tests/chainer_tests/test_variable.py
```diff
@@ -567,24 +567,58 @@ def test_pickle_gpu(self):
...
+    def test_update_rule(self):
+        update_rule = mock.MagicMock()
+        g = self.a.copy()
+        x = chainer.Parameter()
```
delta2323: `x = chainer.Parameter(self.a)` is simpler.

tests/chainer_tests/test_variable.py
```diff
@@ -567,24 +567,58 @@ def test_pickle_gpu(self):
...
+        x.data = self.a
+        x.grad = g
+        x.update_rule = update_rule
+        x.update()
+        self.assertEqual(update_rule.update.call_count, 1)
+        self.assertEqual(update_rule.update.call_args_list[0], [(x,), {}])
+
+    def test_update_rule_without_grad(self):
+        update_rule = mock.MagicMock()
+        x = chainer.Parameter()
```
delta2323: ditto (`x = chainer.Parameter(self.a)`)

tests/chainer_tests/test_variable.py
```diff
@@ -641,90 +675,98 @@ def check_zerograd(self, x, xp):
...
         x.zerograd()
         x.initialize((3, 2))
         self.assertEqual(x.grad.dtype, x.data.dtype)
 
-    def test_copydata_to_uninitialized_variable(self):
-        x = chainer.Variable()
-        y = chainer.Variable(self.a)
+    def test_copydata_to_uninitialized_parameter(self):
+        x = chainer.Parameter()
+        y = chainer.Parameter()
```
delta2323: ditto (`y = chainer.Parameter(self.a)`)

tests/chainer_tests/test_variable.py
```diff
@@ -641,90 +675,98 @@ def check_zerograd(self, x, xp):
...
         x.copydata(y)
         np.testing.assert_array_equal(x.data, self.a)
 
     @attr.gpu
-    def test_copydata_to_uninitialized_variable_gpu(self):
-        x = chainer.Variable()
-        y = chainer.Variable(self.a)
+    def test_copydata_to_uninitialized_parameter_gpu(self):
+        x = chainer.Parameter()
+        y = chainer.Parameter()
```
delta2323: ditto (`y = chainer.Parameter(self.a)`)

tests/chainer_tests/test_variable.py
```diff
@@ -641,90 +675,98 @@ def check_zerograd(self, x, xp):
...
         cp = cuda.cupy
         self.assertIsInstance(x.data, cp.ndarray)
         cp.testing.assert_array_equal(x.data, self.a)
 
-    def test_copydata_from_uninitialized_variable(self):
+    def test_copydata_from_uninitialized_parameter(self):
         initializer = initializers.Zero()
-        x = chainer.Variable(self.a)
-        y = chainer.Variable(initializer=initializer)
+        x = chainer.Parameter()
```
delta2323: ditto (`y = chainer.Parameter(self.a)`)

tests/chainer_tests/test_variable.py
```diff
@@ -641,90 +675,98 @@ def check_zerograd(self, x, xp):
...
         self.assertIsInstance(y.data, np.ndarray)
         np.testing.assert_array_equal(x.data, y.data)
 
     @attr.gpu
-    def test_copydata_from_uninitialized_variable_gpu(self):
+    def test_copydata_from_uninitialized_parameter_gpu(self):
         initializer = initializers.Zero()
-        x = chainer.Variable(self.a)
-        y = chainer.Variable(initializer=initializer)
+        x = chainer.Parameter()
```
delta2323: ditto (`x = chainer.Parameter(self.a)`)

tests/chainer_tests/test_variable.py
```diff
@@ -641,90 +675,98 @@ def check_zerograd(self, x, xp):
...
         x.copydata(y)
         self.assertIsNone(x.data)
         self.assertIsNone(y.data)
 
-    def test_addgrad_to_uninitialized_variable(self):
-        x = chainer.Variable()
-        y = chainer.Variable(self.a, grad=self.b)
+    def test_addgrad_to_uninitialized_parameter(self):
+        x = chainer.Parameter()
+        y = chainer.Parameter()
```
delta2323: ditto

tests/chainer_tests/test_variable.py
```diff
@@ -641,90 +675,98 @@ def check_zerograd(self, x, xp):
...
         self.assertIsInstance(x.grad, np.ndarray)
         np.testing.assert_array_equal(x.grad, self.b)
 
     @attr.gpu
-    def test_addgrad_to_uninitialized_variable_cpu_to_gpu(self):
-        x = chainer.Variable()
-        y = chainer.Variable(self.a, grad=self.b)
+    def test_addgrad_to_uninitialized_parameter_cpu_to_gpu(self):
+        x = chainer.Parameter()
+        y = chainer.Parameter()
```
delta2323: ditto

tests/chainer_tests/test_variable.py
```diff
@@ -734,9 +776,11 @@ def test_addgrad_to_uninitialized_variable_cpu_to_gpu(self):
         cp.testing.assert_array_equal(x.grad, self.b)
 
     @attr.gpu
-    def test_addgrad_to_uninitialized_variable_gpu_to_cpu(self):
-        x = chainer.Variable()
-        y = chainer.Variable(self.a, grad=self.b)
+    def test_addgrad_to_uninitialized_parameter_gpu_to_cpu(self):
+        x = chainer.Parameter()
+        y = chainer.Parameter()
```
delta2323: ditto

tests/chainer_tests/test_variable.py
```diff
@@ -745,9 +789,11 @@ def test_addgrad_to_uninitialized_variable_gpu_to_cpu(self):
         np.testing.assert_array_equal(x.grad, self.b)
 
     @attr.gpu
-    def test_addgrad_to_uninitialized_variable_gpu_to_gpu(self):
-        x = chainer.Variable()
-        y = chainer.Variable(self.a, grad=self.b)
+    def test_addgrad_to_uninitialized_parameter_gpu_to_gpu(self):
+        x = chainer.Parameter()
+        y = chainer.Parameter()
```
delta2323: ditto

tests/chainer_tests/test_variable.py
```diff
@@ -758,9 +804,11 @@ def test_addgrad_to_uninitialized_variable_gpu_to_gpu(self):
         cp.testing.assert_array_equal(x.grad, self.b)
 
     @attr.multi_gpu(2)
-    def test_addgrad_to_uninitialized_variable_gpu_to_another_gpu(self):
-        x = chainer.Variable()
-        y = chainer.Variable(self.a, grad=self.b)
+    def test_addgrad_to_uninitialized_parameter_gpu_to_another_gpu(self):
+        x = chainer.Parameter()
+        y = chainer.Parameter()
```
delta2323: ditto

chainer/link.py
```diff
@@ -139,25 +149,83 @@ def xp(self):
...
+                   super().__init__()
+                   with self.init_scope():
+                       self.W = chainer.Parameter(0, (10, 5))
+                       self.b = chainer.Parameter(0, (5,))
+
+        """
+        old_flag = self.within_init_scope
+        self._within_init_scope = True
+        try:
+            yield
```
beam2d: Because otherwise the code after `yield` will not be executed when an exception is raised in the `with` block.

chainer/link.py
beam2d: Oh, it seems an old code. I will remove it.

chainer/link.py
beam2d: Yes. I did not make it deprecated because the persistent value support is not mature (it is currently only used by `BatchNormalization` and `SimplifiedDropConnect` AFAIK).

tests/chainer_tests/test_variable.py
beam2d: added

beam2d:
@delta2323 I updated the code. Could you check it?

chainer/link.py
delta2323: OK

chainer/link.py
delta2323: OK, I agree with not making it deprecated.

delta2323:
The CI failed due to the violation of flake8 and autopep8. Could you fix it?

chainer/link.py
```diff
@@ -172,27 +240,16 @@ def add_param(self, name, shape=None, dtype=numpy.float32,
                 ignored.
 
         """
-        d = self.__dict__
-        if name in d:
-            raise AttributeError(
-                'cannot register a new parameter %s: attribute exists'
-                % name)
```
delta2323: The removal of this code changes the behavior of `add_param`.  Consider the following code:

```
l = chainer.Link()
l.add_persistent('w', np.array([1, 2, 3]))
l.add_param('w')
```

Previous code raised `AttributeError` but it does not if we apply this PR.

beam2d: Good catch! I will fix `__setattr__` to raise an error in the above case.

beam2d: Good catch! I will fix it.

delta2323: Is it possible to add test cases for the duplicated registration of `add_persistent` and `add_param`?

beam2d: added

chainer/link.py
```diff
@@ -211,9 +272,27 @@ def add_persistent(self, name, value):
...
+    def register_persistent(self, name):
+        """Registers an attribute of a given name as a persistent value.
+
+        This is a convenient method to register an existing attribute as a
+        persistent value.
+
+        Args:
+            name (str): Name of the attribute to be registered.
+
+        """
```
delta2323: I think it is more kind to write to the document that if `name` is already registered as a parameter (i.e. in `self._params`), this method  "hijacks" the corresponding parameter, that is, unregisters it from `_params` and reregisters it as a persistent value.

chainer/link.py
```diff
@@ -552,35 +646,57 @@ def __getitem__(self, name):
...
         Args:
             name (str): Name of the child link. This name is also used as the
                 attribute name.
             link (Link): The link object to be registered.
 
         """
-        if link.name is not None:
-            raise ValueError(
-                'given link is already registered to another chain by name %s'
-                % link.name)
```
delta2323: Almost same as the problem in `Link.add_param`, removal of this code changes the behavior of `Link.add_link`.

chainer/link.py
```diff
@@ -552,35 +646,57 @@ def __getitem__(self, name):
         """Equivalent to getattr."""
         return getattr(self, name)
 
+    def __setattr__(self, name, value):
```
delta2323: Differently from `Link.add_link`, `__setattr__` allows duplicate registration of single link to multiple parent links, although the name of the child link is overwritten. Do you intentionally do that, or do you simply forget it? I thought it might be a specification because there is an issue (#2052) addressing this problem.

chainer/link.py
```diff
@@ -714,27 +830,25 @@ def __len__(self):
...
+        """Registers a child link and adds it to the tail of the list.
 
         Args:
             link (Link): The link object to be registered.
 
         """
-        if link.name is not None:
-            raise ValueError(
-                'given link is already registered to another chain by name %s'
-                % link.name)
```
delta2323: Did you change the specification on purpose?

chainer/link.py
```diff
@@ -714,27 +830,25 @@ def __len__(self):
...
-        deserialization, and involved in the optimization. The registered link
-        is called a child. The child link is accessible via :meth:`children`
-        generator, which returns a generator running through the children in
-        registered order.
+        """
+        self.add_link(link)
 
-        This method also sets the :attr:`~Link.name` attribute of the
-        registered link. If the given link already has the name attribute set,
-        then it raises an error.
```
delta2323: Why did you remove these descriptions?

chainer/link.py
delta2323: If it is not on purpose, could you fix it and add corresponding test cases?

chainer/link.py
beam2d: I found that this behavior is strange (basically we do not need to register a parameter as a persistent value). I will fix this point.

chainer/link.py
beam2d: It is intentional, but I forgot about that till now. I will add this point to the upgrade guide.

chainer/link.py
beam2d: yes

chainer/link.py
beam2d: Because it does not raise an error anymore.

tests/chainer_tests/test_variable.py
delta2323: OK

chainer/link.py
delta2323: OK, it would be desirable if we add tests that ensures the correctness of backprop even though there are multiple registrations.

delta2323: Also, please write to the document that the `name` attribute of `link` registered is overwritten.

chainer/link.py
delta2323: OK.

chainer/link.py
delta2323: OK, then please write the specification to the document and the upgrade guide, as I asked you in other comment box.

chainer/link.py
beam2d: wrote

chainer/link.py
delta2323: OK for the second paragraph. But I thought the first paragraph is still valid after this PR is applied.

chainer/variable.py
```diff
@@ -818,6 +792,142 @@ def add_cand(cand):
...
+                # uninitialized parameter
+                super(Parameter, self).__init__(name=name)
+                self.initializer = initializer
+                dtype = getattr(initializer, 'dtype', numpy.float32)
+                self._grad_initializer = constant.NaN(dtype)
+        else:
+            # parameter initialized with a given shape
+            if isinstance(initializer, (numpy.ndarray, cuda.ndarray)):
+                initializer = constant.Constant(initializer)
+            data = initializers.generate_array(initializer, shape, numpy)
```
delta2323: Should we notify users that parameters are initialized as `numpy.ndarray` even if `initializer` option is `cupy.ndarray`?

chainer/link.py
beam2d: Wrote to the docstring of Link

chainer/link.py
beam2d: fixed

chainer/link.py
beam2d: see the comment above

chainer/link.py
beam2d: see the comment above

chainer/variable.py
beam2d: oh it looks a bug. I will fix it.

chainer/link.py
delta2323: > The registered link is saved and loaded on serialization and deserialization, and involved in the optimization. The registered link is called a child. The child link is accessible via :meth:`children` generator, which returns a generator running through the children in	registered order.

Sorry, I was talking about these sentences. Are these are obsolete, too?

delta2323:
Please write a test for `Link.init_scope` to check that (1) `Paramerter` assigned outside of `init_scope` is  not registered (2) `Variable` assigned inside of `init_scope` is not registered.
Also, please write a test for `Chain.init_scope` to check that `Link` assigned outside of `init_scope` is not registered.

chainer/link.py
beam2d: Oh, is this part removed? It's still remained in my copy...

beam2d:
Added those tests.

chainer/link.py
delta2323: It seems so, if my understanding is correct.

beam2d: Ah, ok, I didn't know that you are talking about the docstring of `add_link` (the diff only shows the old line number which is totally different from the current code). That part is moved to the docstring of `Link`.

examples/word2vec/train_word2vec.py
```diff
@@ -64,11 +64,12 @@
 class ContinuousBoW(chainer.Chain):
 
     def __init__(self, n_vocab, n_units, loss_func):
-        super(ContinuousBoW, self).__init__(
-            embed=L.EmbedID(
-                n_vocab, n_units, initialW=I.Uniform(1. / n_units)),
-            loss_func=loss_func,
-        )
+        super(ContinuousBoW, self).__init__()
+        self.loss_func = loss_func
```
delta2323: Why loss_func is not registerd?

examples/word2vec/train_word2vec.py
```diff
@@ -81,11 +82,12 @@ def __call__(self, x, context):
 class SkipGram(chainer.Chain):
 
     def __init__(self, n_vocab, n_units, loss_func):
-        super(SkipGram, self).__init__(
-            embed=L.EmbedID(
-                n_vocab, n_units, initialW=I.Uniform(1. / n_units)),
-            loss_func=loss_func,
-        )
+        super(SkipGram, self).__init__()
+        self.loss_func = loss_func
```
delta2323: ditto

examples/word2vec/train_word2vec.py
beam2d: oooops, it looks a mistake. I'll quickly fix.

chainer/link.py
delta2323: OK, finally, I found it. Then, I confirm that your change is OK .

tests/chainer_tests/test_link.py
```diff
@@ -398,6 +409,11 @@ def test_add_link_to_existing_attribute(self):
         with self.assertRaises(AttributeError):
             self.l1.add_link('z', chainer.Link())
 
+    def test_assign_link_outside_of_init_scope(self):
+        l = chainer.Link()
+        self.l1.l = l
+        self.assertTrue(l is not link for link in self.l1.links())
```
niboshi: Did you forget `all()`?

delta2323:
LGTM. Now, I want to rely on Travis CI and Jenkins test.

niboshi:
Please check my last review, too

beam2d:
@niboshi I fixed that point.

niboshi:
@beam2d Ah, OK

PR #2665, opened 2017-04-30
make variable names consistent

soramichi:
A tensor descriptor for `xij` should be named `xji_desc`, but not `x_desc`, to be consistent with the names used in other places like [here](https://github.com/pfnet/chainer/blob/master/chainer/functions/connection/dilated_convolution_2d.py#L248).


delta2323:
LGTM, I'm running GPU test.

delta2323:
Thank you!

PR #2669, opened 2017-05-01
Add path option for load_npz method

sergeant-wizard:
I came across a situation where I want to load the serialized parameters under a _certain path_.
If my understanding is correct, the only way to achieve the behavior is to write something like the following:

```py
serializers.NpzDeserializer(np.load('saved_model'))['path_of_interest'].load(obj)
```

which would produce quite a lot of the same code written in the `load_npz` method.
By applying the change I propose here, the user can obtain the same result by:

```py
serializers.load_npz('saved_model', obj, path='path_of_interest')
```

which I think is more straightforward.

### DIFF
--- a/chainer/serializers/hdf5.py
+++ b/chainer/serializers/hdf5.py
@@ -124,7 +124,7 @@ class HDF5Deserializer(serializer.Deserializer):
         return value
 
 
-def load_hdf5(filename, obj, path=''):
+def load_hdf5(filename, obj):
     """Loads an object from the file in HDF5 format.
 
     This is a short-cut function to load from an HDF5 file that contains only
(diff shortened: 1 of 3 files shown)


delta2323:
Could you add tests for verifying this feature?

sergeant-wizard:
Sorry about the delay.
I think these tests should cover the feature that is added.
I found out that hdf5 currently doesn't support the path option so I omitted it for now.

delta2323:
LGTM, I am running GPU tests.

sergeant-wizard:
Oh I forgot to add docstring for the option. I am adding it now

delta2323:
> Oh I forgot to add docstring for the option. I am adding it now

Good catch. Thank you.

delta2323:
Test passed.

PR #2678, opened 2017-05-01
Improve docs of flatten, reshape

keisuke-umezawa:

Modified functions/links:
 - functions/array/flatten
 - functions/array/reshape

### DIFF
--- a/chainer/functions/array/flatten.py
+++ b/chainer/functions/array/flatten.py
@@ -16,7 +16,7 @@ def flatten(x):
     """Flatten a given array into one dimension.
 
     Args:
-        x (list of :class:`~chainer.Variable` or :class:`numpy.ndarray` or \
+        x (:class:`~chainer.Variable` or :class:`numpy.ndarray` or \
         :class:`cupy.ndarray`): Input variable.
 
     Returns:
--- a/chainer/functions/array/flatten.py
+++ b/chainer/functions/array/flatten.py
@@ -13,12 +13,34 @@ class Flatten(function.Function):
 
 
 def flatten(x):
-    """Flatten a given array.
+    """Flatten a given array into one dimension.
 
     Args:
-        x (~chainer.Varaiable): Input variable.
+        x (list of :class:`~chainer.Variable` or :class:`numpy.ndarray` or \
+        :class:`cupy.ndarray`): Input variable.
 
     Returns:
-        ~chainer.Variable: Output variable.
+        ~chainer.Variable: Output variable flatten to one dimension.
+
+    .. admonition:: Example
+
+        >>> x = np.array([[1, 2], [3, 4]])
+        >>> x.shape
+        (2, 2)
+        >>> y = F.flatten(x)
+        >>> y.shape
+        (4,)
+        >>> y.data
+        array([1, 2, 3, 4])
+
+        >>> x = np.arange(8).reshape(2, 2, 2)
+        >>> x.shape
+        (2, 2, 2)
+        >>> y = F.flatten(x)
+        >>> y.shape
+        (8,)
+        >>> y.data
+        array([0, 1, 2, 3, 4, 5, 6, 7])
+
     """
     return Flatten()(x)
--- a/chainer/functions/array/flatten.py
+++ b/chainer/functions/array/flatten.py
@@ -22,6 +22,11 @@ def flatten(x):
     Returns:
         ~chainer.Variable: Output variable flatten to one dimension.
 
+    .. note::
+
+        When you input a scalar array (i.e. the shape is ``()``),
+        you can also get the one dimension array whose shape is ``(1,)``.
+
     .. admonition:: Example
 
         >>> x = np.array([[1, 2], [3, 4]])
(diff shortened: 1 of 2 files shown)


chainer/functions/array/flatten.py
```diff
@@ -13,12 +13,34 @@ def backward(self, inputs, grads):
 
 
 def flatten(x):
-    """Flatten a given array.
+    """Flatten a given array into one dimension.
```
delta2323: Do you think we should mention the situation that input array is a scalar (i.e. shape `()`)? In that case, the resulting array has a shape `(1,)`. I think it is better to add it to the document. But I am also thinking users could think it is natural and need not to treat as an exceptional case.

chainer/functions/array/reshape.py
```diff
@@ -51,12 +51,36 @@ def reshape(x, shape):
...
+
+    .. admonition:: Example
+
+        >>> x = np.array([[1, 2, 3, 4], [5, 6, 7, 8]])
+        >>> y = F.reshape(x, (8,))
+        >>> y.shape
+        (8,)
+        >>> y.data
+        array([1, 2, 3, 4, 5, 6, 7, 8])
+        >>> y = F.reshape(x, (4,-1))
```
delta2323: Add a space after the comma.

chainer/functions/array/reshape.py
```diff
@@ -51,12 +51,36 @@ def reshape(x, shape):
     """Reshapes an input variable without copy.
 
     Args:
-        x (~chainer.Variable): Input variable.
-        shape (tuple of ints): Target shape.
+        x (:class:`~chainer.Variable` or :class:`numpy.ndarray` or \
+        :class:`cupy.ndarray`): Input variable.
+        shape (:class:`tuple` of :class:`int` s):
+            The **size** of shape (**size** means the number of elements) must
```
delta2323: Add an brief explanation of this argument before its constraint (e.g. "Expected shape of the output array").

chainer/functions/array/reshape.py
```diff
@@ -51,12 +51,36 @@ def reshape(x, shape):
...
+        >>> y.data
+        array([1, 2, 3, 4, 5, 6, 7, 8])
+        >>> y = F.reshape(x, (4,-1))
+        >>> y.shape
+        (4, 2)
+        >>> y.data
+        array([[1, 2],
+               [3, 4],
+               [5, 6],
+               [7, 8]])
```
delta2323: Is it possible to add an example that the size of the input array and `shape` argument are not consistent?

chainer/functions/array/reshape.py
delta2323: It seems to me the explanation of `size` could be misleading because we can also interpret it as the *length* of the `shape` (for example, some user could think the size of `(4, 5, 6)` is 3, rather than 4 * 5 * 6 = 120)

chainer/functions/array/flatten.py
keisuke-umezawa: Ok. I will add the document about 0-dim, but I think the case is exceptional, so I will add the document as a note.

chainer/functions/array/reshape.py
keisuke-umezawa: you mean that the case like this?
```
>>> x = np.array([[1, 2, 3, 4], [5, 6, 7, 8]])
>>> y = F.reshape(x, (4, 3))
  File "/home/kumezawa/dev/chainer/chainer/functions/array/reshape.py", line 86, in reshape
    return Reshape(shape)(x)
  File "/home/kumezawa/dev/chainer/chainer/function.py", line 189, in __call__
  File "/home/kumezawa/dev/chainer/chainer/function.py", line 273, in _check_data_type_forward
  File "/home/kumezawa/dev/chainer/chainer/functions/array/reshape.py", line 32, in check_type_forward
    type_check.prod(x_type.shape) == type_check.prod(self.shape))
  File "/home/kumezawa/dev/chainer/chainer/utils/type_check.py", line 487, in expect
    expr.expect()
  File "/home/kumezawa/dev/chainer/chainer/utils/type_check.py", line 449, in expect
Invalid operation is performed in: Reshape (Forward)

Expect: prod(in_types[0].shape) == prod((4, 3))
Actual: 8 != 12
```

I can add the example, but the error message is a little verbose.

delta2323: Yes, that is what I was thinking. We can use [ELLIPSIS](http://www.sphinx-doc.org/ja/stable/ext/doctest.html#confval-doctest_default_flags) to skip the part of output messages. The tests of [chainer.utils.experimental](https://github.com/pfnet/chainer/blob/master/chainer/utils/experimental.py#L46) could be helpful. But let me know if you find it is difficult to write example for exceptional cases.

keisuke-umezawa: Ok. I added example. When testing exception, we do not need to use ELLIPSIS.
http://stackoverflow.com/questions/12592/can-you-check-that-an-exception-is-thrown-with-doctest-in-python

delta2323: Look good.

keisuke-umezawa: @delta2323 how about this PR? Are there anything to improve?

delta2323:
GPU test passed. It is ready to be merged.

chainer/functions/array/reshape.py
delta2323: Sorry for my late reply. This PR is LGTM. We are working to update the workflow to merge PRs and blocking merging them. Could you wait until it is ready?

keisuke-umezawa:
Thank you fort the reply! I will wait for the merge.

PR #2683, opened 2017-05-01
contrastive without reduction

delta2323:
This PR removes `reduce` option from `F.Contrastive` and `F.contrastive`. Different from v1, they output sample wise loss values in v2. Related to #2558 #2603


delta2323:

PR #2689, opened 2017-05-02
Remove obsolete document in DilatedConvolution2D

delta2323:
I remove an obsolete document about `wscale`, which is removed by #2163.


niboshi:
LGTM

Issue #2691, opened 2017-05-02
Clean up APIs of L.DepthwiseConvolution2D and L.SimplifiedDropConnect


delta2323:
In PRs #2163 and #2180, we reorganized the APIs of convolution-like links (e.g. `L.Convolution2D`) for the v2, that involves compatibility break. But as `L.DepthwiseConvolution2D` and `L.SimplifiedDropConnect` are newly-added to the master branch and will be released in v1.24.0, we can change their APIs during v1. Since we do not have to break compatibility in the major version up unnecessarily, we should apply this change before the release date of v1.24.0.

delta2323:
Resolved by #2699 and #2700

PR #2699, opened 2017-05-03
Clean up the API of L.DepthwiseConvolution2D

delta2323:
This PR removes `wscale` and `bias` option from `L.DepthwiseConvolution2D`.  It also removes `initialW` attribute because it does not used in the link. Further, the reason why `L.Convolution2D` is for the backward compatibility, but this link does not have a compatibility we should keep.

Although this PR changes the API of the link, it does not break compatibility because this link is not released in any version.

Related #2691


beam2d:
LGTM

PR #2700, opened 2017-05-03
Clean up the API of L.SimplifiedDropConnect

delta2323:
This PR removes `wscale` option from and add `nobias` option to `L.DepthwiseConvolution2D`.

Although this PR changes the API of the link, it does not break compatibility because this link is not released in any version.

Related #2691


beam2d:
LGTM

PR #2708, opened 2017-05-05
softmax cross entropy without reduction

delta2323:
This PR removes reduce option from `SoftmaxCrossEntropy` and `softmax_cross_entropy`. Different from v1, they output elementwise loss values in v2. Related to #2558 #2357


delta2323:
Should we change `L.Classifier` from
```
self.loss = self.lossfun(self.y, t)
```
to
```
self.loss = F.sum(self.lossfun(self.y, t))
```
because the default value of `lossfun` is `F.softmax_cross_entropy`

delta2323:

PR #2709, opened 2017-05-05
triplet without reduction

delta2323:
This PR removes `reduce` option from `Triplet` and `triplet`. Different from v1, they output elementwise loss values in v2. Related to #2558 #2681


delta2323:

PR #2713, opened 2017-05-06
Add observe_value and observe_lr to extension.rst

mitmul:
Somehow both `observe_value` and `observe_lr` are missing from the documentation. Those are useful, so should be added.


delta2323:
LGTM. But we need to add examples of their usage.

PR #2716, opened 2017-05-07
Fix the bug that F.SoftmaxCrossEntropy does not work properly when ignore_label is other than -1

### LINKED ISSUE #2715: ignore_label in Softmax.forward_gpu does not work when it is not -1
`F.SoftMaxCrossEntropy` can specify `ignore_label` in its `__init__` method. But -1 is hard-coded in its `forward_gpu` and `backward_gpu` methods ([here](https://github.com/pfnet/chainer/blob/master/chainer/functions/loss/softmax_cross_entropy.py#L113) and [here](https://github.com/pfnet/chainer/blob/master/chainer/functions/loss/softmax_cross_entropy.py#L195)). Therefore it does not calculate loss value if `ignore_label` is other than `-1`.

This bug is reported by @musyoku. Thank you!

delta2323:
Fix #2715

### DIFF
--- a/chainer/functions/loss/softmax_cross_entropy.py
+++ b/chainer/functions/loss/softmax_cross_entropy.py
@@ -109,11 +109,10 @@ class SoftmaxCrossEntropy(function.Function):
         log_y = cupy.rollaxis(log_y, 1, log_y.ndim)
         if self.reduce == 'mean':
             ret = cuda.reduce(
-                'S t, raw T log_y, int32 n_channel, raw T coeff', 'T out',
-                't == {} ? T(0) : log_y[_j * n_channel + t]'.format(
-                    self.ignore_label),
+                'S t, raw T log_y, int32 n_channel, raw T coeff, S ignore_label', 'T out',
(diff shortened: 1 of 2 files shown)


okuta:
How about following code?

```python
ret = cuda.reduce(
  'S t, raw T log_y, int32 n_channel, raw T coeff, S ignore_label', 'T out',
  't == ignore_label ? T(0) : log_y[_j * n_channel + t]',
  'a + b', 'out = a * -coeff[0]', '0', 'crossent_fwd'
)(t, log_y.reduced_view(), log_y.shape[-1], self._coeff, self.ignore_label)
```

delta2323:
Thank you for suggestion. I would like to know the reason you prefer this implementation. I noticed that some kernels in this function are implemented in the way you suggested. Is the reason that we should align the implementation to existing ones?

okuta:
Your code generates cubin for each `self.ignore_label`.

delta2323:
Got it. Thank you.

delta2323:
I fixed.

okuta:
LGTM!

PR #2726, opened 2017-05-11
Avoid adding non-zero offset to zero-pointer

### LINKED ISSUE #2722: Avoid adding non-zero offset to an empty pointer
Currently there're cases non-zero offset value is added to an empty pointer.
(https://github.com/pfnet/chainer/blob/44c0fc4f1c4a629effb5ba2928f28c6cd59e1a92/cupy/core/core.pyx#L1237)

While not affecting the external behaviour, it can be a source of future bugs and should be avoided.

Such cases include:
```
cupy.ones((2,0))[1,:]
cupy.ones((0,0)[[],[]]
```

niboshi:
Fixes #2722.

### DIFF
--- a/cupy/core/core.pyx
+++ b/cupy/core/core.pyx
@@ -1216,10 +1216,16 @@ cdef class ndarray:
                 else:
                     dim = (s_stop - s_start + 1) // s_step + 1
 
-                shape.push_back(dim)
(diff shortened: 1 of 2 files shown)


niboshi:
Moved to cupy: https://github.com/cupy/cupy/pull/42

PR #2727, opened 2017-05-11
Make Variable unhashable

### LINKED ISSUE #2670: Variable defines __hash__ but does not define __eq__
`Variable` defines `__hash__` but doesn't define `__eq__`:
```
class Variable(object):
    # ...

    def __eq__(self, other):
        raise NotImplementedError()

    # ...

    def __hash__(self):
```

But, the python documentation (https://docs.python.org/3.1/reference/datamodel.html#object.__hash__) says:
> If a class does not define an __eq__() method it should not define a __hash__() operation either;

Is there a reason `Variable` is designed like this?

niboshi:
Fixes #2670


delta2323:
Could you check the travis result?

niboshi:
Closing PR to clean up branches.

PR #2729, opened 2017-05-11
Update the installation guide for v2.0.0

beam2d:
This change updates the installation guide for v2.0.0. In particular, the dependencies are slightly updated.


docs/source/install.rst
```diff
@@ -1,89 +1,101 @@
...
-  $ apt-get install g++
+The following versions of Python can be used: 2.7.6+, 3.4.3+, 3.5.1+, and 3.6.0+.
 
-  # CentOS 7
-  $ yum install gcc-c++
+.. note::
 
-If you use old ``setuptools``, upgrade it::
+   We are testing Chainer automatically with Jenkins, where all the above *recommended* environments are tested.
+   We cannot guarantee that Chainer works on other environments, even if Chainer looks running correctly.
```
delta2323: Duplicated description, but might be OK to emphasize it. If you prefer, this is suggested change:

```
.. note:: 
... We cannot guarantee that Chainer works on other environments like Windows and macOS platforms (especially, CUDA support in these environments).

File: docs/source/install.rst [hunk 2]
```diff
@@ -1,89 +1,101 @@
...
+  $ pip install -U setuptools
 
-Caffe model support
+The following Python packages are required to install Chainer.
+The latest version of each package will automatically be installed if missing.
 
-* `Protocol Buffers <https://developers.google.com/protocol-buffers/>`_
-* protobuf>=3.0.0 is required for Py3
+* `NumPy <http://www.numpy.org/>`_ 1.9, 1.10, 1.11, 1.12
+* `Six <https://pythonhosted.org/six/>`_ 1.9.0 or higher
```
delta2323: The reason why you do not use "1.9 or higher" in NumPy is that it is hard for use to ensure Chainer will work in the future NumPy versions?

delta2323: Write "1.9.0+" to align with the notation for other packages (or the other way round).

beam2d: As for NumPy, yes. It sometimes breaks the compatibility.
As for the notation, I'll fix it into `1.9+`.

docs/source/install.rst
```diff
@@ -1,89 +1,99 @@
...
-* `Protocol Buffers <https://developers.google.com/protocol-buffers/>`_
-* protobuf>=3.0.0 is required for Py3
+The following packages are optional dependencies.
+Chainer can be installed without them, in which case the corresponding features are not available.
 
-All of the above libraries are automatically installed with ``pip`` or ``setup.py``.
+* CUDA/cuDNN support
 
-CUDA/cuDNN support
+  * `cupy <http://docs.cupy.chainer.org/>`_
```
delta2323: Do you intentionally remove the version of CuPy?

beam2d: I though it is not needed now because Chainer 2.0.0 will support any version of CuPy, but now I think it is better to clarify the version. I will fix it.

beam2d:
Fixed the above point and merged the current v2.

delta2323:
LGTM, I will merge after resolving the problem of current `_v2` branch.

delta2323:
Sorry the conflict occurred again before I merged the PR. Could you resolve the conflict again? Thanks in advance.

beam2d:
Merged the latest v2.

delta2323:
LGTM

PR #2731, opened 2017-05-11
Remove TOC from examples/README.md

niboshi:
Table of contents in this short list is rather confusing.
Maybe we can get it back when the list grows longer.


beam2d:
LGTM

PR #2740, opened 2017-05-16
Dynamically import matplotlib.pyplot in PlotReport

rezoo:
Related to #2146.

`PlotReport` in the current branch does affect the order of modules. This PR aims to solve this issue by using dynamic import.


delta2323:
If my understanding is correct, `matplotlib.use()` after the call of `PlotReport.__call__` does not take effect. Is it right? If so, it is better to rewrite [the warning](https://github.com/pfnet/chainer/blob/master/chainer/training/extensions/plot_report.py#L50) in the document as well.

rezoo:
@delta2323 Sorry for being late. Yes, you are right. I have rewrote the comments and the warnings.

delta2323:
Is it possible to write unit tests that check matplotlib is imported after first call of the extension?

rezoo:
@delta2323 I added the simple test.

tests/chainer_tests/training_tests/extensions_tests/test_plot_report.py
```diff
@@ -23,5 +23,14 @@ def test_available(self):
         else:
             self.assertEqual(len(w), 1)
 
+    def test_lazy_import(self):
+        if extensions.PlotReport.available():
```
delta2323: Why don't you use `unittest.skipUnless`?

rezoo: This is because we want to test the case where matplotlib is not installed.

delta2323: Isn't is possible to `@unittest.skipUnless(extensions.PlotReport.available())` ?

delta2323: Isn't it possible to write as `@unittest.skipUnless(extensions.PlotReport.available(), 'matplotlib is not installed')`

rezoo:
OK I understand what you want. I've fixed.

rezoo:
@delta2323 Hi, is there any progress?

chainer/training/extensions/plot_report.py
```diff
@@ -51,17 +51,20 @@ class PlotReport(extension.Extension):
...
             matplotlib.use('Agg')
 
-            import chainer
+            trainer.extend(
+                extensions.PlotReport(['main/loss', 'validation/main/loss'],
+                                      'epoch', file_name='loss.png'))
+            trainer.run()
 
-        Then, once ``chainer.training.extensions`` is imported,
+        Then, once ``trainer.run`` is called,
```
delta2323: I think "once one of instances of this extension is called" is more accurate (and I believe is not complicated)

tests/chainer_tests/training_tests/extensions_tests/test_plot_report.py
```diff
@@ -17,14 +17,16 @@ def test_available(self):
...
             self.assertEqual(extensions.PlotReport.available(), available)
 
-        # It shows warning only when matplotlib.pyplot is not available
+        # It shows warning only when matplotlib is not available
         if available:
             self.assertEqual(len(w), 0)
         else:
             self.assertEqual(len(w), 1)
 
+    # In Python 2 the above test does not raise UserWarning,
```
delta2323: Sorry, I could not find out the relationship between the comment and the actual unit test.

rezoo: `available()` raises `UserWaning` in the first time, but it does not produce it in the second time, and therefore `self.assertEqual(len(w), 1)` always fails. As unittest does not guarantee the order of calling test methods, in python2 the above test method fails

rezoo: "the above" is correct.

rezoo:
I fixed.

delta2323:
jenkins, test this please.

delta2323:
test passed

PR #2752, opened 2017-05-19
Add strict option to load_npz and load_hdf5

t-abe:
I think `load_xxx` also needs `strict` option.


gwtnb:
Please send to the _v2 branch.

PR #2755, opened 2017-05-22
Remove unnecessary imports from functions and links

beam2d:
This PR removes unnecessary imports of modules in `chainer/functions/__init__.py` and `chainer/links/__init__.py`. In particular, the changes on `chainer/functions/__init__.py` improves the PyCharm support (it will correctly analyses the usage of functions in `chainer.functions`). This PR also fixes the use of symbols under `chainer.links` to meet our coding style guideline.


delta2323:
Jenkins build fails. Could you check # 475 of v2 ?

beam2d:
I think the failure is not due to this change. See the tests of the `_v2` branch.

okuta:
LGTM!

PR #2761, opened 2017-05-23
Make Hyperparameter copyable in Py3.6

beam2d:
This PR fixes a bug of Hyperparameter on deep copy (or, strictly speaking, on unpickling).


gwtnb:
LGTM

Issue #2537, opened 2017-04-12
Inconsistent naming convention between LSTM and GRU


delta2323:
Chainer has both stateful and stateless RNN units (e.g. LSTM) as links. But there naming convention is not consistent among different kinds of RNN. We should 

* `L.StelessLSTM` and `L.LSTM` (stateful)
* `L.StatefulPeepholeLSTM`
* `L.StatefulZoneoutLSTM`
* `L.GRU`(stateless) and `L.StatefulGRU`

I am considering to remove or change the implementation of at least one of `L.LSTM` and `L.GRU` because it should be confusing for users that default implementations of how states are handled (stateless/stateful) are different. So, this issue should be solved in v2.

delta2323:
Close it because it is duplicated to #2285

delta2323:
Sorry, what this issue wants to solve is different form #2285.

delta2323:
Take the LSTM case for example. My idea is to make both `StatefulLSTM` and `StatelessLSTM` and also prepare an `LSTM` as an alias for the stateful one. I (and also some development team members) think that the abbreviation should point a more commonly-used object, which is, I think, the stateful version.

So, what we do is
* Add `L.PeepholeLSTM`, which is alias of `L.StatefulPeepholeLSTM`
* Add `L.ZoneoutLSTM`, which is an alias of `L.StatefulZoneoutLSTM`
* Add `L.StatelessGRU` whose implementation is almost identical to that of current `L.GRU`.
* Change `L.GRU` from stateless to stateful (todo that, we must declare changes).

PR #2768, opened 2017-05-25
Add L.StatelessGRU and change the implementation of L.GRU

delta2323:
This PR adds `L.StatelessGRU` and change `L.GRU` from stateless to stateful as proposed in #2537.

As the semantics of `L.GRU` is changed, this change breaks users' code. To notify users the change, we determine which API (v1/v2) users assume by the number of argument passed to `__call__` and invalidate the former one.

TODO: Add this change to the upgrade guide #2741.


delta2323:
I make a naming convention of the branch name (should be prefixed with `_v2`). I created another identical PR and close it.

PR #2770, opened 2017-05-25
Add Variable.creator setter

beam2d:
The current v2 broke some users' code unintentionally. This PR fixes it.

### DIFF
--- a/chainer/variable.py
+++ b/chainer/variable.py
@@ -161,6 +161,12 @@ class VariableNode(object):
         """Function node that created this variable node."""
         return self._creator
 
+    @creator.setter
+    def creator(self, func):
+        self._creator = func
+        if func is not None:
+            self._rank = func.rank + 1
+
     @property
     def data(self):
         """Data array of the corresponding variable.
@@ -206,16 +212,21 @@ class VariableNode(object):
     def set_creator(self, creator):
         """Sets a :class:`Function` object that created this node.
 
+        This method is equivalent to ``self.creator = creator``.
+
         Args:
             creator (Function): Function object that created this node.
 
         """
-        self._creator = creator
-        self._rank = creator.rank + 1
+        self.creator = creator
 
     def unchain(self):
-        """Deletes the reference to the creator of this variable node."""
-        self._creator = None
+        """Deletes the reference to the creator of this variable node.
+
+        This method is equivalent to ``self.creator = None``.
+
+        """
+        self.creator = None
 
     def retain_data(self):
         """Lets the node hold a reference to the underlying data array.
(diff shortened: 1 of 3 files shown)


chainer/variable.py
```diff
@@ -161,6 +161,12 @@ def creator(self):
         """Function node that created this variable node."""
         return self._creator
 
+    @creator.setter
+    def creator(self, func):
+        self._creator = func
+        if func is not None:
+            self._rank = func.rank + 1
+
```
niboshi: Don't you have to set `self._rank = 0` if `func is None`?

tests/chainer_tests/test_variable.py
```diff
@@ -183,6 +183,23 @@ def test_backward_gpu_retain_grad(self):
         ret = self.create_linear_chain(2, True)
         self.check_backward((ret[0], ), (ret[1], ), (ret[2], ), True)
 
+    def test_unchain(self):
+        ret = self.create_linear_chain(3, False)
+        ret[1].unchain()
+        self.check_backward((ret[1],), (ret[2],), (ret[3],), False)
```
niboshi: How about adding test to check `ret[1].creator` and `.rank`?

tests/chainer_tests/test_variable.py
```diff
@@ -183,6 +183,23 @@ def test_backward_gpu_retain_grad(self):
...
+
+    def test_set_none_to_creator(self):
+        ret = self.create_linear_chain(3, False)
+        ret[1].creator = None
+        self.check_backward((ret[1],), (ret[2],), (ret[3],), False)
```
niboshi: Ditto (test for `.creator` and `.rank`)

tests/chainer_tests/test_variable.py
```diff
@@ -183,6 +183,23 @@ def test_backward_gpu_retain_grad(self):
...
+
+    def test_set_none_and_original_to_creator(self):
+        ret = self.create_linear_chain(2, False)
+        creator = ret[1].creator
+        ret[1].creator = None
+        ret[1].creator = creator
+        self.check_backward((ret[0],), (ret[1],), (ret[2],), False)
```
niboshi: Ditto (test for `.creator` and `.rank`)

delta2323:
Not the setter of `Variable.creator` but `VariableNode.creator`?

chainer/variable.py
beam2d: No, because `_rank` is used for sorting nodes in backward, and the leaf node (including a variable that does not have a creator) is not involved in the sorting.

beam2d:
Both are added for avoiding code duplications.

tests/chainer_tests/test_variable.py
beam2d: OK, will add.

chainer/variable.py
niboshi: OK.

niboshi:
`Function.rank` is set only if `configuration.config.enable_backprop` is `True`. That means, user could get `AttributeError` on accessing `Function.rank`, which is not user-friendly.

I  suggest:
- Raise some error on entrance of `Variable.backward()` if `not enable_backprop`.
- Check if `Function.rank` is set in `VariableNode.creator` setter and `set_creator`.

beam2d:
In most cases, `Function` object used with `enable_backprop=True` is not recorded (I mean, the Function object is not attached to any variable as a creator), and so `backward` does not reach the Function object and no error occurs. If users manually attach a `Function` object to a variable, the error may occur. The error message for this case should be improved. Good catch! I will fix `creator` setter.

beam2d:
I chose to let `Function.rank` return 0 if not set.

niboshi:
LGTM!

PR #2771, opened 2017-05-25
Expose VariableNode as a documented API

delta2323:
This PR makes `chainer.variable.VariableNode`, which was introduced from v2 documented.

`VariableNode` corresponds to a part of documented API of `Variable` in v1.  Also, the upgrade guide we are working on in #2741 mentioned this object. Therefore, it is natural that we expose `VariableNode` to users in v2 as well.

But I do not create an alias `chainer.VariableNode` as I think we do not expect users  use this object through `Variable` and do not touch it directly.


niboshi:
LGTM. Could you change the base branch to master?

delta2323:
Thank you I changed the destination branch.

niboshi:
Sorry, it looks like this has been solved in #2827.

delta2323:
OK, no problem

PR #2773, opened 2017-05-25
Update contribution guide

beam2d:
I wrote an updated contribution guide for v2.

### DIFF
--- a/docs/source/contribution.rst
+++ b/docs/source/contribution.rst
@@ -1,9 +1,14 @@
-Chainer Contribution Guide
-==========================
+Contribution Guide
+==================
 
 This is a guide for all contributions to Chainer.
 The development of Chainer is running on `the official repository at GitHub <https://github.com/pfnet/chainer>`_.
 Anyone that wants to register an issue or to send a pull request should read through this document.
+
+.. note::
+
+   Many points of this document are updated at v2.
+   We strongly recommend all contributors of v1 to read through the document again.
 
 Classification of Contributions
 -------------------------------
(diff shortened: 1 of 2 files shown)


docs/source/contribution.rst
```diff
@@ -1,48 +1,129 @@
...
+ 16 weeks    --          Y.0.0      Z.0.0a1
+========== =========== =========== ============
+
+The dates shown in the left-most column are relative to the release of ``X.0.0rc1``.
+In particular, each revision release is made four weeks after the previous one of the same major version, and the pre-release of the upcoming major version is made at the same time.
+
+Note that the development of ``X.0.x`` stops at ``X.0.2``.
+During the parallel development of ``Y.0.0`` and ``Z.0.0a1``, the version ``Y`` is treated as an **almost-stable version** and ``Z`` is treated as a development version.
+
+If there is a critical bug found in ``X.0.2`` after stopping the development of version ``X``, we may release a hot-fix for this version at any time.
```
delta2323: Does it mean we do not release any hot-fix version for X.0.0 and X.0.1?

docs/source/contribution.rst
```diff
@@ -1,48 +1,129 @@
...
+
+.. note::
+
+   The release cycle of ``2.0.x`` and ``3.0.0x`` are slightly different from this table because we do not have ``3.0.0a1`` at the timing of the release of ``2.0.0``.
+   In this case, the releases of ``3.0.0x`` are shifted four weeks behind the usual timeline, that is, ``3.0.0a1`` will be released at the same time with ``2.0.1``.
+
+As you can see in the above table, we basically do not have any minor releases from v2.
```
delta2323: ", that means the second digit is always 0."

docs/source/contribution.rst
```diff
@@ -1,48 +1,129 @@
...
+- Otherwise, choose the ``dev`` branch.
 
-Issues and PRs
---------------
+For example, a change to fix a bug of the stable version should basically be sent to the ``master`` branch.
+A change to fix a bug caused by newly-introduced APIs in the development version should be sent to the ``dev`` branch.
+A change to introduce a new function/class, add a new argument to an existing function, and/or modify the existing arguments of a function should be sent to the ``dev`` branch.
 
-Issues and PRs are classified into following categories:
+*Note: a change that can be applied to both branches should be sent to the* ``master`` *branch.*
```
delta2323: Use `.. note::`

delta2323:
We need to define the term *major update* and *revision update* somewhere in the contribution guide because it could not be obvious from users. Is the following definition correct?

* Major update is an update form X.0.0rc1 to X.0.0.
* Revision update is one of the following updates: 
   * from X.0.0 to Y.0.0a1
   * from Y.0.0a1 to Y.0.0b1
   * from Y.0.0b1 to Y.0.0rc1

delta2323:
If we narrow the supported versions of CUDA and cuDNN (which is done in cupy/cupy#34), we should notify it users. I am not sure it is appropriate but, the update guide is one of the possible place to write it.

PR #2776, opened 2017-05-26
Remove get_device

mitmul:
get_device has been marked as deprecated, so we can remove it in v2. It takes over this PR: #2243

### DIFF
--- a/chainer/cuda.py
+++ b/chainer/cuda.py
@@ -301,7 +301,7 @@ def copy(array, out=None, out_device=None, stream=None):
     if out is None:
         if out_device is None:
             out_device = array
-        with get_device_from_id(out_device):
+        with _get_device(out_device):
             out = cupy.empty_like(array)
 
     with get_device_from_array(array):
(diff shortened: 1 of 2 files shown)


chainer/cuda.py
```diff
@@ -337,10 +301,10 @@ def copy(array, out=None, out_device=None, stream=None):
     if out is None:
         if out_device is None:
             out_device = array
-        with get_device(out_device):
+        with get_device_from_id(out_device):
```
niboshi: Should this be `_get_device`? I think `out_device` can be either array or id.

mitmul:
@niboshi Thank you very much for your quick review. I fixed them.

niboshi:
Thank you. You also need to fix 2 more `get_device()` lines in `optimizer.py`.

niboshi:
@unnonouno Can you check and merge?

beam2d:
We decided to postpone this change. 1) We need a longer period to remove the deprecated `get_device`, so the removal will be done in the next major. 2) Removing the use of `get_device` from Chainer's internal code can be done in the revision update, so we will make it in 2.0.1 or later.

mitmul:
It's been replaced with #2890

PR #2778, opened 2017-05-27
Update API Compatibility Policy

beam2d:
This change updates API Compatibility Policy for v2. Since it includes a link to some sections of the updated contribution guide, this PR depends on #2773. Please merge #2773 before this one.

### DIFF
--- a/docs/source/compatibility.rst
+++ b/docs/source/compatibility.rst
@@ -28,7 +28,7 @@ See :ref:`contrib` for details of versioning.
 
 The backward compatibility is kept for **revision updates**, which are applied to the latest stable version.
 A **major update** from the latest release candidate basically keeps the backward compatibility, although it is not guaranteed.
-Any **pre-releases** may break the bakward compatibility.
+Any **pre-releases** may break the backward compatibility.
 
 
 Breaking the Compatibility
@@ -52,7 +52,7 @@ The following list shows an example of what we can do to reduce the cost (*Note:
 .. note::
 
    Since Chainer v2, we have stopped adopting any solid processes to break backward compatibilities (e.g. a solid schedule for deprecating and removing a feature) in order to keep the development fast enough to support the cutting-edge research.
-   **It does not mean we stop taking care of mantenability of user codes.**
+   **It does not mean we stop taking care of maintenability of user codes.**
    We are still paying much attention to not breaking user codes.
 
 
--- a/docs/source/compatibility.rst
+++ b/docs/source/compatibility.rst
@@ -1,3 +1,5 @@
+.. _compatibility:
+
 API Compatibility Policy
 ========================
 
--- a/docs/source/compatibility.rst
+++ b/docs/source/compatibility.rst
@@ -26,7 +26,7 @@ Versioning and Backward Compatibility
 The versioning of Chainer follows the `PEP 440 <https://www.python.org/dev/peps/pep-0440/>`_ and a part of `Semantic versioning <http://semver.org/>`_.
 See :ref:`contrib` for details of versioning.
 
-The backward compatibility is kept for **revision updates**, which are applied to the latest stable version.
+The backward compatibility is kept for **revision updates**, which are applied to the stable version.
 A **major update** from the latest release candidate basically keeps the backward compatibility, although it is not guaranteed.
 Any **pre-releases** may break the backward compatibility.
 
--- a/docs/source/compatibility.rst
+++ b/docs/source/compatibility.rst
@@ -3,8 +3,8 @@
 API Compatibility Policy
 ========================
 
-This document expresses the design policy on compatibilities of Chainer APIs.
-Development team should obey this policy on deciding to add, extend, and change APIs and their behaviors.
+This document explains the design policy on compatibilities of Chainer APIs.
+Development team should follow this policy on deciding to add, extend, and change APIs and their behaviors.
 
 This document is written for both users and developers.
 Users can decide the level of dependencies on Chainer’s implementations in their codes based on this document.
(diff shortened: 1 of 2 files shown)


docs/source/compatibility.rst
```diff
@@ -13,59 +15,45 @@ Note that this document may contain ambiguities on the level of supported compat
...
-Processes to Break Backward Compatibilities
--------------------------------------------
+The versioning of Chainer follows the `PEP 440 <https://www.python.org/dev/peps/pep-0440/>`_ and a part of `Semantic versioning <http://semver.org/>`_.
+See :ref:`contrib` for details of versioning.
 
-Deprecation, Dropping, and Its Preparation
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+The backward compatibility is kept for **revision updates**, which are applied to the latest stable version.
+A **major update** from the latest release candidate basically keeps the backward compatibility, although it is not guaranteed.
+Any **pre-releases** may break the bakward compatibility.
```
delta2323: typo: backward compatibility.

docs/source/compatibility.rst
```diff
@@ -13,59 +15,45 @@ Note that this document may contain ambiguities on the level of supported compat
...
-The actual change should be done in the following steps:
+.. note::
 
-- Announce that the API will be changed in the future.
-  At this point, the actual version of change need not be accurate.
-- After the announcement, mark the API as *to be changed in the future* with version number of planned changes.
-  At this point, users should not use the marked API in their new application codes.
-- At the major update announced in the above update, change the API.
+   Since Chainer v2, we have stopped adopting any solid processes to break backward compatibilities (e.g. a solid schedule for deprecating and removing a feature) in order to keep the development fast enough to support the cutting-edge research.
+   **It does not mean we stop taking care of mantenability of user codes.**
```
delta2323: typo: maintenability

docs/source/compatibility.rst
```diff
@@ -13,59 +15,45 @@ Note that this document may contain ambiguities on the level of supported compat
...
+-------------------------------------
 
-Processes to Break Backward Compatibilities
--------------------------------------------
+The versioning of Chainer follows the `PEP 440 <https://www.python.org/dev/peps/pep-0440/>`_ and a part of `Semantic versioning <http://semver.org/>`_.
+See :ref:`contrib` for details of versioning.
 
-Deprecation, Dropping, and Its Preparation
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+The backward compatibility is kept for **revision updates**, which are applied to the latest stable version.
```
delta2323: I think we should remove "latest" because that could mistakenly make users think that there are multiple stable versions, which is not true.

docs/source/compatibility.rst
```diff
@@ -105,97 +91,95 @@ and how users enable or disable the warnings practically.
...
-On the other hand, warnings may be added at any minor updates for any APIs.
-It means minor updates do not keep backward compatibility of warnings.
+On the other hand, warnings may be added at any revision updates for any APIs.
+It means revision updates do not keep backward compatibility of warnings.
 
 Model Format Compatibility
 --------------------------
 
-Objects serialized by official serializers that Chainer provides are correctly loaded with the higher (future) versions.
+Links and chains serialized by official serializers that Chainer provides are correctly loaded with the future versions.
```
delta2323: Do we ensure model compatibilities are kept even an **major** version up is done?

docs/source/compatibility.rst
beam2d: I thought there ARE multiple stable versions (e.g. both 1.24.0 and 2.0.0 will be stable after releasing 2.0.0). Isn't it true?

docs/source/compatibility.rst
beam2d: I thought we should do that, otherwise, users that have their own serialized models cannot upgrade to a newer version.

docs/source/compatibility.rst
delta2323: According to the explanation of "stable version" that appeared for the first time in the document ([here](https://github.com/pfnet/chainer/pull/2773/files#diff-337031db06965ccb19ed87653a389a9dR52)), stable version is the revision updates for the latest major versions. So I thought the stable version automatically points the latest version.

docs/source/compatibility.rst
delta2323: Make sense. Thank you.

delta2323: But, to what extent can we guarantee the major version up which could break backward compatibility?

docs/source/compatibility.rst
```diff
@@ -105,97 +91,95 @@ and how users enable or disable the warnings practically.
...
 
 Behaviors of Chainer implementation not stated in the documentation are undefined.
-Undocumented behaviors are not guaranteed to be stable between different minor/revision versions.
+Undocumented behaviors are not guaranteed to be stable between different revision versions.
 
-Minor update may contain changes to undocumented behaviors.
-For example, suppose an API X is added at the minor update.
-In the previous version, attempts to use X cause AttributeError.
-This behavior is not stated in the documentation, so this is undefined.
-Thus, adding the API X in minor version is permissible.
```
delta2323: It seems this example is still valid if we replace "minor update" with "revision update". So I don't think we need to remove it.

docs/source/compatibility.rst
```diff
@@ -105,97 +91,95 @@ and how users enable or disable the warnings practically.
...
 
-On the other hand, warnings may be added at any minor updates for any APIs.
-It means minor updates do not keep backward compatibility of warnings.
+On the other hand, warnings may be added at any revision updates for any APIs.
+It means revision updates do not keep backward compatibility of warnings.
 
 Model Format Compatibility
 --------------------------
 
-Objects serialized by official serializers that Chainer provides are correctly loaded with the higher (future) versions.
```
delta2323: Is the reason why you have changed the term is that you think it is difficult to keep the model compatibilities other than Link and Chains (e.g. Trainer)?

docs/source/compatibility.rst
beam2d: Oh, I see. Hmm, ok, I'll fix the presentation.

docs/source/compatibility.rst
beam2d: API won't be added to the revision update.

docs/source/compatibility.rst
beam2d: Yes

delta2323: I want confirm that we have a consensus about narrowing the support among the developer team.

beam2d: I think we already broke the serialization compatibility of non-model objects.

docs/source/compatibility.rst
delta2323: I see. Thank you.

PR #2782, opened 2017-05-29
Remove chainer_setup_build.py

delta2323:
This PR removes an artifact file `chainer_setup_build.py`, which is no longer used in Chainer v2.


beam2d:
thank you

PR #2783, opened 2017-05-29
Rename option value of reduce option in F.HuberLoss from sum_along_second_axis to sum_each_data

delta2323:
The `reduce` option of `F.HuberLoss.__init__` takes `sum_along_second_axis`. But, what its option name does not reflect actually what it does. Specifically, it aggregate not only the second axis but also the axes later than the second axis.

This PR adds `sum_each_data` to the acceptable value of `reduce` option of `F.HuberLoss.__init__`. This option behaves identically as `sum_along_second_axis`. Also `sum_along_second_axis` is made deprecated.

This change is extracted from #2265.

### DIFF
--- a/chainer/functions/loss/huber_loss.py
+++ b/chainer/functions/loss/huber_loss.py
@@ -1,3 +1,5 @@
+import warnings
+
 import numpy
 
 from chainer import cuda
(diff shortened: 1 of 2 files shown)


delta2323:
This PR is turned out to be wrong. The changes in the original PR made by mitmul-san invalidated the option `sum_along_second_axis` and that was why he made the option deprecated. So the change in this PR should be coupled with the original PR.

PR #2789, opened 2017-05-29
Check h5py version when serializing None

niboshi:
`h5py.Empty` is introduced from `h5py==2.7.0`.


beam2d:
Could you add a test that confirms serializing None raises an error when h5py is old?

niboshi:
OK

niboshi:
Added test.

beam2d:
LGTM

PR #2795, opened 2017-05-30
Update README to give a link to the stable branch

mitmul:
I think we need to add a comment to README in the master branch.


beam2d:
Thank you

PR #2810, opened 2017-06-02
Make vision models copyable

gwtnb:
When one copies a vision model, `self.functions` is not correctly copied. This PR fixes the problem by making `functions` a property.


delta2323:
Could you write unit tests that verifies that the problem is fixed?

t-abe:
I noticed `BuildingBlock` in `resnet.py` has same problem.

delta2323:
Jenkins failed due to the problem cupy/cupy#122. As it is solved, I am re-running GPU tests again. I will merge it after it.

delta2323:
LGTM

PR #2814, opened 2017-06-03
fix chainer.Link docstring

chantera:
I found a wrong example in chainer.Link docstring and fixed arguments order to be valid.


delta2323:
Thank you!

PR #2816, opened 2017-06-05
Use Evaluator instead of TestModeEvaluator in the data-parallel example

beam2d:
We can use Evaluator directly to switch the training mode since v2.


beam2d:
I've also changed the permission of the file so that users can directly run the script (it would be better to make a separate PR, but both changes are small so I just made it in this PR).

delta2323:
LGTM

PR #2821, opened 2017-06-05
Fix a minior bug of resnet.py

t-abe:
`n_layers` argument for `ResNetLayers` was missing.

### DIFF
--- a/chainer/links/model/vision/resnet.py
+++ b/chainer/links/model/vision/resnet.py
@@ -141,7 +141,7 @@ class ResNetLayers(link.Chain):
         # we import CaffeFunction here.
         from chainer.links.caffe.caffe_function import CaffeFunction
         caffemodel = CaffeFunction(path_caffemodel)
-        chainermodel = cls(pretrained_model=None)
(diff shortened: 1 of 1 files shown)


mitmul:
Now it results in a weird error at a Jenkins test job. Investigating it.

mitmul:
Doc test is still failing due to a error, but it must not be related to this PR. So LGTM

mitmul:
I'm sorry, I wrongly merged this PR. The doctest error should have been fixed before merging this.

kiyukuta:
I think we do not need `n_layers` here.

Although base `ResNetLayer.__init__` needs `n_layers`, `ResNetXXXLayers.__init__` does not and  `convert_caffemodel_to_npz ` is called only from specific `ResNetXXLayers` classes.
https://github.com/chainer/chainer/blob/master/chainer/links/model/vision/resnet.py#L699-L706

kiyukuta:
Sorry, never mind, I noticed #3062 will solve this.

mitmul:
@kiyukuta Thank you for pointing out it. I agree with that. Can I revert this PR, @t-abe ?

PR #2824, opened 2017-06-05
Fix get_trainer_with_mock_updater

Hakuyume:
In v1, I have sent a PR to fix this bug (#2472). However, it is restored in v2.


delta2323:
LGTM, I am running the GPU tests.

delta2323:
CI passed.

PR #2839, opened 2017-06-09
Improve download

Hakuyume:
This PR contains two improvements.
- Currently, `get_dataset_directory` ignores `OSError`. However, this behaviour is not good because `OSError` can be raised in other cases.
- `cached_download` contains `makedirs`. It can be replaced with `get_dataset_directory`.


delta2323:
LGTM

delta2323:
I am asking @beam2d, who made this module, if it is OK to change the type of error to raise in `cached_download`?

delta2323:
According to beam2d, as `_dl_cache` is not a directory for caching a dataset, it is not an expected usage of `get_dataset_directory`. Although the functionality of `get_dataset_directory` is identical to what we want here now, we could change its behavior in the future.

Hakuyume:
> Although the functionality of get_dataset_directory is identical to what we want here now, we could change its behavior in the future.

OK, I understood that I should not replace `mkdirs` in `cached_download` with `get_dataset_directory`. How about the former improvement? That is as follows.
> Currently, get_dataset_directory ignores OSError. However, this behaviour is not good because OSError can be raised in other cases.

tests/chainer_tests/dataset_tests/test_download.py
```diff
@@ -121,14 +121,14 @@ def tearDown(self):
     def test_fail_to_make_dir(self):
         with mock.patch('os.makedirs') as f:
             f.side_effect = OSError()
-            with self.assertRaises(RuntimeError):
+            with self.assertRaises(OSError):
```
delta2323: As this PR does not change the behavior of `cached_download`, we need not to change the type of the error.

tests/chainer_tests/dataset_tests/test_download.py
```diff
@@ -121,14 +121,14 @@ def tearDown(self):
...
                 dataset.cached_download('http://example.com')
 
     def test_file_exists(self):
         # Make an empty file which has the same name as the cache directory
         with open(os.path.join(self.temp_dir, '_dl_cache'), 'w'):
             pass
-        with self.assertRaises(RuntimeError):
+        with self.assertRaises(OSError):
```
delta2323: ditto

beam2d:
I wanted to support the case that multiple processes try to make the same directory at the almost same time. Checking the existence of the directory before creating it may not help in this case; see the following example.
```py
# Even if the following condition holds,
if create_directory and not os.path.isdir(path):
    # another process may create the directory here,
    os.makedirs(path)  # in which case it causes an error
```
Such an error is not rare; it is likely to happen when one invokes the same script from multiple processes (e.g. to distribute experiments on different hyperparameters).

Hakuyume:
@beam2d I see. But current code ignores the casesthat a regular file with same name already exists or user have enough permission to make a directory. I think some exception should be raised in those cases.

Hakuyume:
> Although the functionality of get_dataset_directory is identical to what we want here now, we could change its behavior in the future.

I understood this opinion and stopped using `get_dataset_directory` in `cached_download`. However, I think it is better to use the exception that is raised by `os.makedirs` as is because user can know the reason why it failed. So I modified `cached_download` and its test.

Hakuyume:
Is this PR under reviewing?

delta2323:
I am sorry that we have overlooked this PR. @beam2d Do you have any comment on the comment by @Hakuyume ?

niboshi:
I don't catch all the discussion above, but for emulating `exist_ok=True`, checking `e.errno` of `OSError` is the idiom: https://stackoverflow.com/questions/273192/how-can-i-create-a-directory-if-it-does-not-exist

There might be a case where directory is created but still has an error (mode error, for example).

By the way, I agree with @Hakuyume about exception type.

Hakuyume:
@niboshi 
> I don't catch all the discussion above, but for safely emulating exist_ok=True, checking e.errno of OSError is the idiom: https://stackoverflow.com/questions/273192/how-can-i-create-a-directory-if-it-does-not-exist

Thank you for suggestion. But `IOError.errno` does not seem enough.
I tried three situations:
- a directory already existed (this case is ok)
    `IOError.errno` was `17 (= errno.EEXIST)`.
- a directory without write permission already existed (it is better to catch this case)
    `IOError.errno` was `17`.
- a regular file already existed (I want to catch this case)
    `IOError.errno` was `17`.

As shown above, `IOError.errno` is not enough to distinguish these cases.

niboshi:
@Hakuyume 
Thank you. I checked the source code of `os` and it states `EEXIST` method is not safe...
```
    try:
        mkdir(name, mode)
    except OSError:
        # Cannot rely on checking for EEXIST, since the operating system
        # could give priority to other errors like EACCES or EROFS
        if not exist_ok or not path.isdir(name):
            raise
```

>(it is better to catch this case)

It's not clear to me, but by "catch", do you mean "catch and ignore"?
I think in both 2nd and 3rd cases, it's better not to catch (but it's difficult in the 2nd case).

Hakuyume:
> It's not clear to me, but by "catch", do you mean "catch and ignore"?
I think in both 2nd and 3rd cases, it's better not to catch (but it's difficult in the 2nd case).

No, I mean it is better to notify users in the 2nd case (not ignore). We should ignore `OSError` only in the 1st case. As you pointed out, it is difficult to distinguish the 2nd case from the 1st case.

Hakuyume:
Although it cannot distinguish the 2nd case from the 1st case, using `os.path.isdir` is better than current code.

niboshi:
@Hakuyume 
Thank you, it's clear now and I agree with you.

Hakuyume:
@delta2323 Are you waiting for @beam2d 's comment?

delta2323:
Yes, I pinged him.

beam2d:
Sorry for letting you wait. This PR LGTM. It would be helpful for existing users to add a note on changed error types to the docstring.

Hakuyume:
@beam2d Thank you for reviewing. I added a short note.

delta2323:
jenkins, test this please.

delta2323:
test passed

delta2323:
@beam2d Should we backport this PR?

Hakuyume:
@delta2323 I think #3057 is solved by this PR.

delta2323:
Thank you for notifying me. I closed the issue.

PR #2840, opened 2017-06-09
Remove wrong commas from network definitions in ConvNet tutorial

mitmul:
Two models in the convnet tutorial can't work correctly due to the wrong commas at the end of lines inside model definitions.


delta2323:
LGTM

PR #2846, opened 2017-06-12
Add additional special methods in class template

niboshi:
These are mainly used in:
- Datasetes
- Iterators
- Variable and Parameter

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -50,6 +50,7 @@ help:
 
 clean:
 	rm -rf $(BUILDDIR)/*
+	find source -type d -name generated -print0 | xargs -0 rm -rf
 
 html:
 	$(SPHINXBUILD) -b html $(ALLSPHINXOPTS) $(BUILDDIR)/html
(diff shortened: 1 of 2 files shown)


delta2323:
I tried this PR and generated the document in my local environment. But I could not confirm the change. For example, `DictDataset.__len__` is not appeared in the document although it is implemented.

niboshi:
Thank you for testing.
Hmm, below is a screenshot from my build.
Probably you didn't clear `generated/` directories under `sources/`:

```
rm -rf **/generated/
```

screenshot from 2017-06-13 14-25-23

niboshi:
Maybe it should be written in `clean` target in `Makefile`.
`**/generated/` above is an extended syntax of bash, so we will need to use `find` command to do the same.

niboshi:
Added to `make clean` rule. You should be able to use it to build correctly.

delta2323:
OK, it works. Thank you.

delta2323:
LGTM

PR #2847, opened 2017-06-12
Fix a bug of to_gpu in cuda.py

mitmul:
Current to_gpu tries to copy an array on a device to another device by calling `cupy.array(array, copy=True)`, but it causes the error `ValueError: Array device must be same as the current device: array device = 0 while current = 1`. Just calling `array.copy()` inside the context of the device to which it tries to copy the array is fine.

This fixes the test failres of `test_cupy_array2`.


mitmul:
The test failures will be fixed by https://github.com/cupy/cupy/pull/122 instead of this PR. The fundamental problem is in CuPy.

Issue #2693, opened 2017-05-02
Dilated convolution with cuDNN


delta2323:
Now cuDNN v6 supports the dilated convolution, we should use it in `F.DilatedConvolution2DFunction` if possible. Currently this function does use cuDNN if available, but for computing normal convolutions.

delta2323:
This issue has been solved via #2858

PR #2858, opened 2017-06-14
cuDNN v6 dilated convolution

anaruse:
This PR is related to dilated convolution (#2693) and enabling Chainer to use an implementation of dilated convolution in cuDNN v6. Note that you need to use [this branch](https://github.com/anaruse/cupy/tree/dilated_convolution_2d) of Cupy (https://github.com/cupy/cupy/pull/133).


chainer/functions/connection/convolution_2d.py
```diff
@@ -33,7 +33,7 @@ def _pair(x):
 
 class Convolution2DFunction(function.Function):
 
-    def __init__(self, stride=1, pad=0, cover_all=False, **kwargs):
+    def __init__(self, stride=1, pad=0, cover_all=False, dilate=1, **kwargs):
```
ir5: It seems you put `dilate` option in `Convolution2DFunction`, but it is not supported in CPU mode. This is misleading for users. Further, if `Convolution2DFunction` supports dilation feature fully, then `DilatedConvolution2DFunction` will become the subset of it. This means we will have some sort of overlapping codes. I think there are two ways to fix this.

1. Do not modify `Convolution2DFunction` (i.e., do not put `dilate` option here) and modify only the dilated function.
2. Complete to implement `dilate` argument (i.e., implement dilation in CPU mode) and make the current dilated function just calls `Convolution2DFunction` internally.

How do you think about this?

chainer/functions/connection/convolution_2d.py
```diff
@@ -109,13 +110,14 @@ def forward_gpu(self, inputs):
...
-                                      cover_all=self.cover_all)
+                                      cover_all=self.cover_all, d=self.dy)
         assert out_h > 0, 'Height in the output should be positive.'
         out_w = conv.get_conv_outsize(w, kw, self.sx, self.pw,
-                                      cover_all=self.cover_all)
+                                      cover_all=self.cover_all, d=self.dx)
         assert out_w > 0, 'Width in the output should be positive.'
 
         y = cuda.cupy.empty((n, out_c, out_h, out_w), dtype=x.dtype)
+        # print('# conv_2d.py:120, y.shape: {}'.format(y.shape))  # debug
```
ir5: Please remove debug message.

chainer/links/connection/convolution_2d.py
```diff
@@ -101,7 +101,8 @@ class Convolution2D(link.Link):
     """  # NOQA
 
     def __init__(self, in_channels, out_channels, ksize=None, stride=1, pad=0,
-                 nobias=False, initialW=None, initial_bias=None, **kwargs):
+                 dilate=1, nobias=False, initialW=None, initial_bias=None,
```
ir5: Similar problem holds for the link.

anaruse:
Thank you for the feedback.

> It seems you put dilate option in Convolution2DFunction, but it is not supported in CPU mode. This is misleading for users. Further, if Convolution2DFunction supports dilation feature fully, then DilatedConvolution2DFunction will become the subset of it. This means we will have some sort of overlapping codes. I think there are two ways to fix this.
> 
> 1. Do not modify Convolution2DFunction (i.e., do not put dilate option here) and modify only the dilated function.
> 2. Complete to implement dilate argument (i.e., implement dilation in CPU mode) and make the current dilated function just calls Convolution2DFunction internally.

I was not sure which was better when I sent the PR, but now I think the option two would be better mainly because:
* the codes of `Convolution2DFunction` and `DilatedConvolution2DFunction` are alike and there are a lot of duplicated lines among them, except the codes that utilizes cuDNN v5.x or before (which did not support dilated convolution) and calls cuDNN's 1x1 convolutions multiple times for dilated convolution  (#1335). The current implementation of dilated convolution using cuDNN v5.x or before is great, but seems it is not so faster than an implementation with cupy and looks it is not so easy to maintain. Considering that dilated convolution is natively supported by cuDNN from v6 and you can add dilated convolution feature to `Convolution2DFunction` without large amount of code modifications, I think it is good timing to merge `Convolution2DFunction` and `DilatedConvolution2DFunction`.

What would you think?

chainer/functions/connection/convolution_2d.py
anaruse: Noted, will remove it.

chainer/functions/connection/convolution_2d.py
anaruse: (Please look at my comment below)

chainer/links/connection/convolution_2d.py
anaruse: (Please look at my comment below)

ir5:
Thank you for your opinion. Now I discussed with chainer core developers about this since merge is a little big change. They and I consider merging two similar functions is a good contribution, but it is somehow different from the objective in this PR. So, I recommend you to do (1) implementing cuDNN v6 dilated conv and (2) merging the two functions in separate PRs. I guess you may do either one ((1) or (2)) at first. If you feel making two PRs is troublesome, however, doing the two things in one PR is probably admissible.

anaruse:
Thank you for the comment.

I agree your comment that it is not a small change to make two functions together. So, I will focus mainly on implementing dilated convolution feature in `Convolution2DFunction` for CPU and GPU and try to make an amount of code modifications to `DilatedConvolution2DFunction` small in this PR.

ir5:
So, I assume we are working on only (1) in this PR.
I guess we need to check if the result of standard convolution with dilate option is compatible with the existing dilated convolution.
I noticed you add a hidden flag `_use_standard_convolution` in `dilated_convolution_2d.py` which is probably for local test. When I modified this to `True` and ran test for `test_dilated_convolution_2d.py`, however, some test cases failed. My cudnn is a little old (4.0.7), so my environment is probably different with yours. My cupy was up to latest master branch.
I feel the way of testing using `_use_standard_convolution` is a little awkward.
How about adding the tests in dilated convolution to standard convolution so that we can see the test results on CI?

anaruse:
Thank you for the feedback.
As you noticed, the flag `_use_standard_convolution` is for local testing. And I agree with you making that flag adjustable from test scripts like `test_dilated_convolution_2d.py`. So, will fix the codes.

anaruse:
I've fixed `test_dilated_convolution_2d.py` so that it also tests the dilated convolution codes in a class `Convolution2DFunction`.

ir5:
Thank you for adding tests. My comments are:

* I feel it is suitable to move the new tests in `test_convolution_2d.py` to `test_dilated_convolution_2d.py`, because the tests are about `dilate` argument in `Convolution2DFunction`. I guess `testing.parameterize` will reduce overlaps of codes.
* Probably we need to add an explanation of `dilate` argument in docstrings of both function and link. There are many ways to explain, but I guess a simple description is enough. For example, adding an item in "Args:" section such as "dilate (int or pair of ints): Dilation factor of filter applications. For detail, please see :class:`DilatedConvolution2DFunction`." looks enough. Also, it may be worth noting that cuDNN is used in dilated convolution if its version is >= 6 in docstrings. I guess we may write this either in standard convolution or in dilated convolution.
* `_use_standard_convolution` is unused code, so I feel it's better to remove it from this PR.

chainer/functions/connection/dilated_convolution_2d.py
```diff
@@ -61,6 +70,9 @@ def check_type_forward(self, in_types):
             )
 
     def forward_cpu(self, inputs):
+        if getattr(self, '_func', False):
```
ir5: Knowing that we will work on merging tasks (2) right after this PR, this part is unused in this PR as far as I could see. So, to avoid confusion, it might be better to remove this part in this PR. How do you think about this? The same thing holds for `backward_cpu`.

anaruse:
Thank you for the comment.

I've updated the branch based on your comment:
* moving some test codes from `test_dilated_convolution_2d.py` to `test_convolution_2d.py`, 
* adding an explanation of a `dilate` argument of a function class `Convolution2DFunction`, 
* removing some unsed codes in a function class `DilatedConvolution2DFunction`, etc.

Note that I've not added a `dilate` argument to a link class `Convolution2D` yet, so I kept test codes for the link class unchanged. I thinks the addition of the `dilate` argument to the link class should be done at next step.

BTW, no failure has been observed so far when running nosetests with cuDNN 5.1 and 6.0.

chainer/functions/connection/convolution_2d.py
```diff
@@ -388,6 +398,8 @@ def convolution_2d(x, W, b=None, stride=1, pad=0, cover_all=False, **kwargs):
             ``pad=p`` and ``pad=(p, p)`` are equivalent.
         cover_all (bool): If ``True``, all spatial locations are convoluted
             into some output pixels.
+        dilate (int or pair of ints): Dilation factor of filter applications.
+            ``dilate=d`` and ``dilate=(d, d)`` are equivalent.
```
ir5: How about putting a reference to `DilatedConvolution2DFunction` class? I guess dilated convolution is not familiar with beginners.

ir5:
Thank you. Now the code looks good to me. I put one minor comment.

okuta:
@anaruse Cloud you rebase?

anaruse:
I've just updated the branch.

First, this is not a simple conflicts resolution. Please start review from the beginning.

The previous approach was to make cuDNN's dilated convolution available with minimal modification to Chainer. So, only really necessary modification to support cuDNN's dilated convolution was added to `Convolution2DFunction` and the added feature was called from `DilatedConvolution2DFunction` only when cuDNN is selected to use. This approach was practical because `Convolution2DFunction` and `DilatedConvolution2DFunction` were very similar.

However, I found this approach is no longer practical. That is mainly because a lot of modification have been added to `Convolution2DFunction` to support higher order differentiation and they are no longer very similar.

There were two possible options.
1. Take the same approach as before, after adding modification to `DilatedConvolution2DFunction` to support higher order differentiation.
2. Modify `Convolution2DFunction` so that it fully supports dilated convolution for both CPU and GPU.

Given that amount of code change, when the 1st option is taken, to `DilatedConvolution2DFunction` for higher order differentiation is not negligible amount, I think the 1st option is not practical. Thus, I've decided to take the 2nd route.

With this PR, `Convolution2DFuncton` (and also `Deconvolution2DFunction` since they are highly related now) has been modified to fully support dilated convolution and `DilatedConvolution2DFunction` has been deleted. There is still `dilated_convolution_2d` remaining for API compatibility reason.

okuta:
Thanks, I will cehck

chainer/functions/connection/convolution_2d.py
```diff
@@ -27,7 +28,7 @@ def _pair(x):
 
 class Convolution2DFunction(function_node.FunctionNode):
 
-    def __init__(self, stride=1, pad=0, cover_all=False, **kwargs):
+    def __init__(self, stride=1, pad=0, cover_all=False, dilate=1, **kwargs):
```
okuta: Please use `parse_kwargs` .

chainer/functions/connection/convolution_2d.py
```diff
@@ -270,18 +281,19 @@ def backward(self, indexes, grad_outputs):
...
                 x, ggW, stride=(self.sy, self.sx), pad=(self.ph, self.pw),
-                cover_all=self.cover_all)
+                cover_all=self.cover_all, dilate=(self.dy, self.dx))
             ret.append(ggy)
 
         return ret
 
 
-def convolution_2d(x, W, b=None, stride=1, pad=0, cover_all=False, **kwargs):
+def convolution_2d(x, W, b=None, stride=1, pad=0, cover_all=False, dilate=1,
```
okuta: Please use `parse_kwargs` .

chainer/functions/connection/deconvolution_2d.py
```diff
@@ -30,7 +30,7 @@ class Deconvolution2DFunction(function_node.FunctionNode):
 
     cover_all = None
 
-    def __init__(self, stride=1, pad=0, outsize=None, **kwargs):
+    def __init__(self, stride=1, pad=0, outsize=None, dilate=1, **kwargs):
```
okuta: Please use `parse_kwargs` .

chainer/functions/connection/deconvolution_2d.py
```diff
@@ -233,11 +245,14 @@ def _set_cover_all(self, x, W):
...
-            in_h != conv.get_conv_outsize(self.outh, kh, self.sy, self.ph) or
-            in_w != conv.get_conv_outsize(self.outw, kw, self.sx, self.pw))
+            in_h != conv.get_conv_outsize(self.outh, kh, self.sy,
+                                          self.ph, d=self.dy) or
+            in_w != conv.get_conv_outsize(self.outw, kw, self.sx,
+                                          self.pw, d=self.dx))
 
 
-def deconvolution_2d(x, W, b=None, stride=1, pad=0, outsize=None, **kwargs):
+def deconvolution_2d(x, W, b=None, stride=1, pad=0, outsize=None, dilate=1,
```
okuta: Please use `parse_kwargs` .

okuta:
jenkins, test this please

anaruse:
Thanks for your comment.
The branch was modified so that `dilate` argument is passed via `kwargs`, based on your comment.

chainer/functions/connection/convolution_2d.py
```diff
@@ -110,15 +112,20 @@ def forward_gpu(self, inputs):
...
-                                      cover_all=self.cover_all)
+                                      cover_all=self.cover_all, d=self.dx)
         assert out_w > 0, 'Width in the output should be positive.'
 
         y = cuda.cupy.empty((n, out_c, out_h, out_w), dtype=x.dtype)
+        self._use_cudnn = False
         if (not self.cover_all and chainer.should_use_cudnn('>=auto') and
                 x.dtype == W.dtype):
+            self._use_cudnn = True
+        if (self.dy > 1 or self.dx > 1) and _cudnn_version < 6000:
```
okuta: Please merge to L123 statement.

chainer/functions/connection/deconvolution_2d.py
```diff
@@ -41,11 +41,12 @@ def __init__(self, stride=1, pad=0, outsize=None, **kwargs):
             "the gradient w.r.t. x is automatically decided during "
             "backpropagation."
         )
-        argument.assert_kwargs_empty(kwargs)
```
okuta: Do not remove this statement.
Please move to after `parse_kwargs`.

chainer/functions/connection/deconvolution_2d.py
```diff
@@ -342,9 +356,9 @@ def deconvolution_2d(x, W, b=None, stride=1, pad=0, outsize=None, **kwargs):
         "supported anymore. "
         "Use chainer.using_config('cudnn_deterministic', value) "
         "context where value is either `True` or `False`.")
-    argument.assert_kwargs_empty(kwargs)
```
okuta: ditto.

okuta:
jenkins, test this please.

chainer/functions/connection/convolution_2d.py
anaruse: All right, will merge the two if statements.

chainer/functions/connection/deconvolution_2d.py
anaruse: Should `assert_kwargs_empty` be called after `parse_kwargs`? It looks it is checked at `parse_kwargs` whether `kwargs` is empty or not.

chainer/functions/connection/deconvolution_2d.py
anaruse: The same as above.

chainer/functions/connection/deconvolution_2d.py
okuta: `parse_kwargs` uses `pop`.
maybe `assert_kwargs_empty` works correctly.
Cloud you try it?

anaruse: I've just confirmed that unexpected arguments can be detected by `parse_kwargs()`. It looks, if there is `parse_kwargs()`, it is not necessary to call `assert_kwargs_empty()`.  Followings are error message when a unexpected argument (in this case, dummy) is added to `convolution_2d`.

      File "/home/anaruse/.pyenv/versions/3.5.1/lib/python3.5/site-packages/chainer-3.0.0rc1-py3.5.egg/chainer/functions/connection/convolution_2d.py", line 442, in convolution_2d
        dilate, = argument.parse_kwargs(kwargs, ('dilate', 1))
      File "/home/anaruse/.pyenv/versions/3.5.1/lib/python3.5/site-packages/chainer-3.0.0rc1-py3.5.egg/chainer/utils/argument.py", line 12, in parse_kwargs
        raise TypeError('got unexpected keyword argument(s) %s' % args)
    TypeError: got unexpected keyword argument(s) 'dummy'

okuta: OK, I understood.

okuta:
jenkins, test this please.

okuta:
I got error on CUDA v5 environment.
Please fix [this line](https://github.com/chainer/chainer/pull/2858/files#diff-34a15eecaef34d920b685d1f971f6a17R244).

anaruse:
Thank you for pointing it out. it was fixed.

okuta:
jenkins, test this please.

okuta:
jenkins, test this please.

okuta:
Please merge or rebase master, and fix `get_deconv_outsize`.

anaruse:
Thanks for your comment. The conflict was resolved.

okuta:
jenkins, test this please.

okuta:
I got following error. Please fix test of dilated conv.
```
01:13:45 _____ TestDilatedConvolution2DCudnnCall_param_7.test_call_cudnn_backrward ______
01:13:45 
01:13:45 self = <chainer.testing.parameterized.TestDilatedConvolution2DCudnnCall_param_7 testMethod=test_call_cudnn_backrward>
01:13:45 
01:13:45     def test_call_cudnn_backrward(self):
01:13:45         with chainer.using_config('use_cudnn', self.use_cudnn):
01:13:45             y = self.forward()
01:13:45             y.grad = self.gy
01:13:45             name = 'cupy.cudnn.cudnn.convolutionBackwardData_v3'
01:13:45             with mock.patch(name) as func:
01:13:45                 y.backward()
01:13:45 >               self.assertEqual(func.called, self.expect)
01:13:45 E               AssertionError: False != True
```

anaruse:
Could you tell me CUDA version, cuDNN version, etc. when that error occurred?

okuta:
I got error on CUDA7/cuDNN4 and CUDA8/cuDNN51.

anaruse:
> I got following error. Please fix test of dilated conv.

Please run the test again. I think the error is already resolved by the last commit. I've confirmed the error above does not happen with CUDA8 and cuDNN 5.1.

okuta:
jenkins, test this please.

okuta:
LGTM!

PR #2863, opened 2017-06-14
Add F.pad_sequence to the reference

YamaneSasuke:
This PR adds `pad_sequence` function to reference, which seems to be missing after it was implemented.


delta2323:
LGTM

PR #2766, opened 2017-05-24
Improve browsability of API reference

niboshi:
This PR adopts NumPy-like documentation style

PR #2865, opened 2017-06-15
[Backport] Add F.pad_sequence to the reference

delta2323:
Backport of #2863


niboshi:
Please target `v2` branch, and I think you need to modify because we have changed the document style in `master` branch.

delta2323:
I mistakenly made a PR on top of `master`. So I close it and re-submit another one.

delta2323:
I noticed that as this PR is made on top of #2766, we have backport #2766 to merge this PR to v2 (in other words, if we backport this PR, #2766 will be automatically backported). Should we do that?

niboshi:
In my opinion, it's better to simply add `F.pad_sequence` to `v2` branch in the old style.
Backporting #2766 is too cumbersome.

delta2323:
I re-sent a PR (#2884)

PR #2872, opened 2017-06-16
Use zeros instead of empty for ROIPooling2D CPU mode

yuyu2172:
The value can get returned without initialization in some case.
Therefore, we should use `zeros` instead of `empty`.

For example, `sliceh.stop == `sliceh.start`, when height and ymin are very close. 

https://github.com/yuyu2172/chainer/blob/bee0deb0dae30bf0db7a2f630860cb9a4a97116d/chainer/functions/pooling/roi_pooling_2d.py#L94

### DIFF
--- a/chainer/functions/pooling/roi_pooling_2d.py
+++ b/chainer/functions/pooling/roi_pooling_2d.py
@@ -73,9 +73,9 @@ class ROIPooling2D(function.Function):
         bottom_data, bottom_rois = inputs
         channels, height, width = bottom_data.shape[1:]
         n_rois = bottom_rois.shape[0]
-        top_data = numpy.empty((n_rois, channels, self.outh, self.outw),
+        top_data = numpy.zeros((n_rois, channels, self.outh, self.outw),
                                dtype=numpy.float32)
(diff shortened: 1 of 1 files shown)


chainer/functions/pooling/roi_pooling_2d.py
```diff
@@ -73,9 +73,9 @@ def forward_cpu(self, inputs):
         bottom_data, bottom_rois = inputs
         channels, height, width = bottom_data.shape[1:]
         n_rois = bottom_rois.shape[0]
-        top_data = numpy.empty((n_rois, channels, self.outh, self.outw),
+        top_data = numpy.zeros((n_rois, channels, self.outh, self.outw),
                                dtype=numpy.float32)
-        self.argmax_data = numpy.empty(top_data.shape, numpy.int32)
```
niboshi: Write as a comment the reason why `zeros` is needed instead of `empty`.

yuyu2172:
Sorry for being so late.
I updated the code.
Thank you for reviewing.

niboshi:
jenkins, test this please

niboshi:
LGTM!

PR #2877, opened 2017-06-18
Remove INFINITY

okuta:


gwtnb:
What is the purpose of this PR? The `INFINITY` macro is supported in C99 and C++11. Is there any CUDA environment without `INFINITY`?

okuta:
Because NVRTC does not understand `INFINITY`.

takagi:
It makes sense, LGTM!

I suspected that explicit math.h inclusion might be needed,

https://developer.nvidia.com/cuda-math-library

but it just caused "incompatible linkage specification" errors. Compilation with NVCC did not cause these errors even in the same condition with math.h, so NVRTC is supposed to do something different with NVCC.

PR #2886, opened 2017-06-19
[Backport] Use Evaluator instead of TestModeEvaluator in the data-parallel example (#2816)

delta2323:
This PR is the backport of #2816


okuta:
LGTM!

PR #2891, opened 2017-06-20
Improve same_types and remove checking in backward

okuta:
This PR deletes type check at backward because it is redundant.

### DIFF
--- a/chainer/functions/connection/dilated_convolution_2d.py
+++ b/chainer/functions/connection/dilated_convolution_2d.py
@@ -64,7 +64,7 @@ class DilatedConvolution2DFunction(function.Function):
         x, W = inputs[:2]
         b = inputs[2] if len(inputs) == 3 else None
 
-        if not type_check.same_types(*inputs):
+        if not all([isinstance(i, numpy.ndarray) for i in inputs]):
             if b is not None:
                 raise ValueError('numpy and cupy must not be used together\n'
                                  'type(W): {0}, type(x): {1}, type(b): {2}'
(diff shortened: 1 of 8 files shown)


chainer/functions/connection/dilated_convolution_2d.py
```diff
@@ -88,16 +88,6 @@ def forward_gpu(self, inputs):
...
-        if not type_check.same_types(*inputs):
-            if b is not None:
-                raise ValueError('numpy and cupy must not be used together\n'
-                                 'type(W): {0}, type(x): {1}, type(b): {2}'
-                                 .format(type(W), type(x), type(b)))
-            else:
-                raise ValueError('numpy and cupy must not be used together\n'
-                                 'type(W): {0}, type(x): {1}'
-                                 .format(type(W), type(x)))
-
```
rezoo: Why you deleted the above lines?

chainer/utils/type_check.py
```diff
@@ -515,15 +515,10 @@ def expect(*bool_exprs):
...
-    are_numpy_arrays = map(lambda x: issubclass(type(x), numpy.ndarray),
-                           arrays)
-    all_numpy_arrays = all(are_numpy_arrays)
-    if cuda.available:
-        are_cupy_arrays = map(lambda x: issubclass(type(x), cuda.cupy.ndarray),
-                              arrays)
-        return all_numpy_arrays or all(are_cupy_arrays)
-    else:
-        return all_numpy_arrays
+    if all(map(lambda x: isinstance(x, numpy.ndarray), arrays)):
```
rezoo: How about using generator comprehension like that?
```python
if all(isinstance(x, numpy.ndarray) for x in arrays):
```

chainer/utils/type_check.py
```diff
@@ -515,15 +515,10 @@ def expect(*bool_exprs):
...
+        return True
+    return cuda.available and all([isinstance(x, cuda.ndarray)
+                                   for x in arrays])
```
rezoo: Ditto.
```python
return cuda.available and all(isinstance(x, cuda.ndarray) for x in arrays)
```

chainer/utils/type_check.py
okuta: Current code is faster than your suggested code.

rezoo: Is it obvious? I'm just curious to see the difference.

okuta:
I fixed!

rezoo:
LGTM.

PR #2892, opened 2017-06-20
Matmul cleanup

fukatani:
Integrate matmul API by #2426 is incomplete, I clean up `_matmul` and `_batch_matmul`.


chainer/functions/math/matmul.py
```diff
@@ -36,36 +34,20 @@ def _get_ld(a):
...
     if xp is numpy:
-        ret = numpy.empty(a.shape[:2] + b.shape[2:], dtype=a.dtype)
-        for i in six.moves.range(len(a)):
-            ret[i] = numpy.dot(a[i], b[i])
-        return ret
+        if a.ndim <= 2:
+            return numpy.dot(a, b)
+        else:
+            return numpy.einsum('...ij,...jk->...ik', a, b)
     return xp.matmul(a, b)
```
okuta: NumPy1.10.0 has `matmul`. Do we use it?

fukatani: Numpy 1.9.0 doesn't support `matmul`.
And `numpy.einsum` and `numpy.matmul` are almost equivalent performance in numpy.1.11, I don't use `numpy.matmul` now.

I implemented similar code in `depthwise_conv2d`.
https://github.com/chainer/chainer/pull/2067

If you want to use `numpy.matmul` for numpy >= 1.10, I'll replace current implementation.

fukatani: `if xp is numpy and numpy.version.version.split('.')[1] == '9'` is prefer in L.46?

chainer/functions/math/matmul.py
```diff
@@ -36,36 +34,20 @@ def _get_ld(a):
...
 
 def _matmul(a, b, transa=False, transb=False, transout=False):
-    a = array.as_mat(a)
-    b = array.as_mat(b)
     if transout:
-        # (A B)^T = B^T A^T
         transa, transb = not transb, not transa
         a, b = b, a
-    if transa:
-        a = a.T
```
okuta: Which line covers this case ?

chainer/functions/math/matmul.py
fukatani: Sure, I fixed here.

chainer/functions/math/matmul.py
fukatani: if `a.ndim >= 2`, L.40 covers this case.
If `a.ndim == 1`, `a.T == a`

fukatani:
I found this PR should include `basic_math.matmul` modification.
In #2426, we made `Matmul` behavior same as `numpy.matmul`, but `basic_math.MatmulConstVar` and `basic_math.MatmulVarConst` was not changed.

For this purpose, I added some commits.
Sorry for inconvenience.

chainer/functions/math/matmul.py
okuta: OK, I see.

okuta:
Thank you!
LGTM!

PR #2238, opened 2017-02-06
Improve asynchronous host to device copy

okuta:
This PR implement asynchronous copy to `cupy.array`.
The stream option in `cuda.to_gpu` is deprecated, because stream option is difficult.

PR #2907, opened 2017-06-23
fix stream option not working with to_gpu

kmaehashi:
CPU-to-GPU transfer with CUDA stream is not working.
It seems that this error was introduced in #2238.


okuta:
Thanks!
LGTM

PR #2612, opened 2017-04-21
Modify docstring for IDE friendly

ysekky:
In some module, in_channel/in_size parameter accepts None, but only int is described in docstring.

This PR add "or None" to docstring.

PR #2912, opened 2017-06-26
Fix issue of retain_outputs

anaruse:
This PR is to fix an issue of retain_outputs (that is discussed in #2586).

The issue is that an output array might get inaccessible during backward even when it is declared to be retained by `retain_outputs()`. As far as I understand correctly, this issue may happen only when the `variable node` of the output array is a leaf node in a computational graph. So, my proposal to fix/work-around the issue is:
  1. to make a temporary reference from the function to the variable node when an output array is specified to be retained by `retain_outputs()`
  2. to remove the temporary reference when the variable node is used as an input by another function.

I've confirmed that the work-around above actually solve the case mentioned in #2586 while keeping GPU memory usage small.

### DIFF
--- a/chainer/function.py
+++ b/chainer/function.py
@@ -235,6 +235,25 @@ class Function(object):
             if output_indexes_to_retain is not None:
                 for index in output_indexes_to_retain:
                     ret[index].retain_data()
+
+            # Work-around to prevent variable nodes in a computational graph
+            # from being released unexpectedly
+            self._outputs = [None for y in ret]
+            if output_indexes_to_retain is not None:
+                for index in output_indexes_to_retain:
+                    # Make a temporary reference to the variable node
+                    # of the retained output
+                    self._outputs[index] = ret[index].node
+
+            for x in self.inputs:
+                if x.creator is None:
+                    continue
+                for index in range(0, len(x.creator._outputs)):
+                    if (x.creator._outputs[index] is x):
+                        # Remove a temporary reference to my inputs
+                        # from the correspondig creator functions
+                        x.creator._outputs[index] = None
+
             del self._output_indexes_to_retain
 
         if len(ret) == 1:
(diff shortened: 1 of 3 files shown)


beam2d:
I thought that the issue that I have noted in #2586 may also happen when the output node is not a "leaf" (I assume you are meaning the "root" of the backward graph). For example, the following code which is a modification to the code
```py3
z, w = func(x, y)
u = some_function(w)
del w, u
```
will also make w unavailable in the backward of func.

I also think it is dangerous to introduce a circular reference to the computational graph; the current code of this PR looks introducing a memory leak when the user tries to release an output variable of a function that retains the output immediately before applying any other function to that.

beam2d:
I am now suspecting that it is difficult to take the strategy of setting `output_data` just before calling `backward` instead of setting it in `__call__`. The life span of `output_data` should be equal to the life time of the variable node itself.

anaruse:
I agree with you that circular reference in a computational graph is prone to be problematic.

BTW, seems this circular reference is resolved by python GC.
In the case above, if the circular reference is not resolved:
  * The `variable node` of `u` and the `some_function` have reference each other (circular reference), so they are not released when the `variable` of `u` is deleted by user.
  * There is a reference from the `some_function` to the `variable node` of `w`, so the `variable node` of `w` is not released whtn the `variable` of `w` is deleted by user.

anaruse:
> I am now suspecting that it is difficult to take the strategy of setting `output_data` just before calling `backward` instead of setting it in `__call__`.

I've modified the branch based on your comment above. The approach taken now is that 
  1. to set `output_data` in `__call__()` for arrays specified to be retained and
  2. to update `output_data` in `backward()` when corresponding `variable node` is available.

You may wonder why 2. is necessary.  The reason is that the pointers (memory address) of arrays may change between the `__call__()` and the `backward()` considering recompute feature discussed in #2612. If you think it should be done in #2612, I will do so.

PR #2918, opened 2017-06-27
Add Sequential class for easy model definition of a single-stream computational graph

mitmul:
This is inspired by @musyoku's work found here: https://github.com/musyoku/chainer-sequential-chain
It aims at providing another way to define a network in Chainer.

This PR introduces `chainer.Sequential` class which enables users to create a sequential model easily. It focuses only on the network which has a single stream forward pass. This enables to write a three-layer perceptron as appeared in the tutorial in the below manner:

```python
import chainer
import chainer.functions as F
import chainer.links as L
from chainer import Sequential

model = Sequential(
    L.Linear(n_in, n_hidden),
    F.relu,
    L.Linear(n_hidden, n_hidden),
    F.relu,
    L.Linear(n_hidden, n_out)
)

y = model(x)
```

And it also supports built-in list APIs. So users can add `Function`, `Link` and even user-defined function to the model easily like:

```python
from functools import partial
from chainer.links.model.vision.vgg import VGG16Layers

model = Sequential()
model.append(L.Linear(n_out, n_hidden))
model.append(F.relu)
model.append(partial(F.reshape, shape=(1, 3, 224, 224)))
model.append(VGG16Layers())
model.append(lambda x: x['prob'])

y = model(x)
```

Furthermore, supporting built-in list APIs also enables to perform add and mul operations:

```python
model_A = Sequential(L.Linear(10, 10), F.relu)
model_B = Sequential(L.Linear(10, 10), F.sigmoid)

model_C = model_A + model_B

model_D = 3 * model_A   # model_A is used as a builllding block of model_D three times CHain
```

<table style="border: none">
<tr><td>Then, <b>model_C</b> looks like:</td><td>and <b>model_D</b> looks like:</td></tr>
<tr>
<td>

</td>
<td>

</td>
</tr>
</table>

You can see the network structure of a Sequential object just by using `print`:

```python
print(model_C)
```

will show:

```
0       Linear  W(10, 10)       b(10,)
1       relu
2       Linear  W(10, 10)       b(10,)
3       sigmoid
```

## TODO

- [x] Write tests
- [x] Write more documents to methods
- [x] `__radd__` should be able to take any callable objects because `Sequential` can take them
- [x] Use ChainList intead of Chain
- [x] Raise an exception when a Sequentioal which has lambd function is pickled (Use `functools.partial` instead
- [x] When copying a Link, re-instantiate a same Link object to randomize the initial parameters (Not use Link.copy)


mitmul:
Writing tests and adding documents have been basically finished. Please start a review.

muupan:
I think extending ChainList makes code much simpler than extending Chain. Is there any reason why Chain is preferred?

jekbradbury:
Could this behavior be made the default for `ChainList` objects that don't have a user-defined `__call__`? That way you don't need a new class.

mitmul:
Thanks for the comment, @muupan, I'll replace `Chain` with `ChainList` to shrink the amount of codes.

Thank you for you comment, @jekbradbury. Some people want to keep `ChainList` different one from `Sequential`, so I'll make another PR to add `__init__` method to `ChainList`.

ir5:
It seems `Sequential` class is in `link.py`, but `Sequential` is not related with core implementation of `Link` or `ChainList`. So I guess it is better to implement it in other new file such as `chainer/links/model/sequential.py`.

mitmul:
I've been told that the Chainer team has not agreed on the design of this Sequential class, so I marked this as "Don't review" by changing its title. Thank you for your comment, @ir5 , I'll modify the file where this class is defined in. And there is a suggestion to make this feature `experimental` to allow fundamental design change after v3. So I'll also make this experimental later.

mitmul:
Jenkins, test this please

mitmul:
sequential.py is added to separate Sequential class from Chain/Link in file level.

mitmul:
@ir5 Finally we decided to progress this PR to merge. So, could you start reviewing this?

delta2323:
@ir5 Can we assign you as a reviewer? If you cannot spare time to work on this PR. We can take it over.

chainer/sequential.py
```diff
@@ -0,0 +1,383 @@
...
+        """
+        for layer in self._layers:
+            if isinstance(x, tuple):
+                x = layer(*x)
+            else:
+                x = layer(x)
+        return x
+
+    def __reduce__(self):
+        if self._n_lambda > 0:
```
beam2d: Is `_n_lambda` only used here? If so, I feel it's better to count the number of lambda functions on the fly here.

chainer/sequential.py
```diff
@@ -0,0 +1,383 @@
...
+    def __mul__(self, n_repeat):
+        if n_repeat <= 0:
+            return Sequential()
+        ret = self.copy()
+        for _ in range(n_repeat - 1):
+            for layer in self:
+                if isinstance(layer, chainer.Link):
+                    layer = copy.deepcopy(layer)
+                    for param in layer.params(include_uninit=False):
+                        param.initialize(param.shape)
```
beam2d: Is it safe to assume that `param` has an initializer?

mitmul: By the change in [chainer/variable.py L:1203](https://github.com/chainer/chainer/pull/2918/files/f6ba4799086872f636dbabc9e8de1c73ca897795#diff-e131a8dfc2ea641e7e7c63597b9b0d8eR1203), `param` which is a `Parameter` object always has initializer.

chainer/sequential.py
mitmul: Yes, it's used only here. So that's right. I'll change to calculate it here.

mitmul:
Thanks for the comments. I fixed the points.

chainer/sequential.py
beam2d: If a user sets a random array as the initial value, it's detected as a `Constant` initializer and then the parameters will not be sampled again when multiplying the Sequential containing that link. Is it appropriate? One can avoid this kind of thing by enforcing some coding practice (e.g. always use an appropriate initializer (or define one) when you initialize a parameter randomly), but I'm not sure if it's common to users (at least we did not advertise this coding practice).

beam2d: Also, as I wrote in the above comment of the PR thread, this behavior (re-initializing the parameters on multiplying) is not consistent with the behavior of `list`, which does not copy the content. In Python, we usually use list comprehension in such a case to repeat the same but independently initialized values in a list, like `[Linear(100) for _ in range(10)]` instead of `[Linear(100)] * 10`. I understand that the original behavior of `list.__mul__` is confusing (e.g. 10 Linears with _shared_ weights in this example), so it's ok to not reproduce it in `Sequential`, but I'm wondering if it's appropriate to provide an alternative definition of `__mul__`.

beam2d: Do we need to ask other members as well? My opinion may be extreme...

mitmul:
As a consequence of the discussion on chainer.slack.com and offline discussions, I removed `__mul__` operator and added `repeat` method for the same purpose, the new one has three mode: `init`, `copy`, and `deepcopy`. These are different at the way to repeat elements. Please see the docstring of the method. Could you check this again and review? @beam2d @niboshi

niboshi:
`deepcopy` is about implementation, it may be better to express a behavior.

I also think these 3 modes should probably be implemented for copying a (general) link, and `Sequential` would simply use that function for each link.

To implement that, I think we should consider the following matrix.

||uninitialized|initialized|
|-------|-------|-------|
|reinitialize|Copy initializer|Clear data and copy initializer|
|copy parameters|Error?|deepcopy|
|share parameters|?|Copy reference|

mitmul:
@niboshi Thanks for the suggestions. In what cases users want to use such varied copy modes without Sequential? I couldn't come up with the objectives to use all those copy modes without and outside of Sequential. It's very helpful if you would provide examples of those various copy modes.

niboshi:
@mitmul These are some examples I come up with.

* Shared paremeters
Useful to train a shared weight network e.g. Siamese.
That is, first define a sub-network N as a link and copy it to construct f(N(x1, w), N(x2, w)) with shared parameters w.
**uninitialized**: Train from scratch
**initialized**: Train using pretrained N

* Copy parameters
Build an arbitrary large network which includes some repeated structure N.
ex. h( g(N(x1, w1)), N(f(x2), w2))
**initialized**: Initialize N with some pretrained network.

* Re-initialize
**uninitialized**, **initialized**: Given a sub-network N, train a network which includes its structure, but from scratch.

I'm not so sure how those would be implemented in current Chainer. Please correct me if I'm stating something unfit.

chainer/sequential.py
```diff
@@ -0,0 +1,405 @@
...
+            if isinstance(layer, chainer.Link):
+                if layer.__class__.__name__ == type_name:
+                    num += 1
+            else:
+                if layer.__name__ == type_name:
+                    num += 1
+        return num
+
+    def sort(self, key=None, reverse=None):
+        raise NotImplementedError
```
beam2d: `NotImplementedError` in Python means "this method should be overridden by a subclass", so it's not appropriate to use it here. Consider another exception type, or just remove this definition (ditto for `reverse`).

mitmul:
OK, from the comments by @niboshi and @beam2d, I move the `repeat` method under `Link` class and change those names for `mode`.

chainer/sequential.py
mitmul: OK, I remove both.

mitmul: Hmm, but, for example, `Variable`'s `__eq__` raises `NotImplementedError`, so it's intended to be overridden by a subclass?

mitmul: Thanks!

mitmul:
@beam2d @niboshi I moved `repeat` method from `Sequential` to `Link` class and also moved the tests for `repeat` method to `test_link.py`. Seems working now, so please re-start reviewing. Thanks.

niboshi:
I meant those copy modes would be implemented in `Link.copy` and `Link.repeat` would use it internally.
Do you think it makes sense?

mitmul:
@niboshi I think shallow-copied elements are difficult to be restored by serializers, and changing `copy` method of `Link` class will affect a too broad range of Chainer, so I think that's out of scope here. Thank you for clarifying your idea.

niboshi:
@mitmul
Do you mean current Chainer does not support shared-weights networks?
Even if that's the case, we should eventually support it.
And when that happens, `Link.repeat` should not change its behavior.
So I think at least `Link.repeat` should recognize those cases and raise appropriate errors.

In my understanding current behavior of `Link.copy` is shared parameters. If that's correct, we can implement other modes without changing its behavior.

mitmul:
@niboshi No, I don't. Chainer does support shared-weight networks. OK, I understand what you mean. Just moving the part which performs copying `layer` in `repeat` method to `copy` method and setting the default `mode` to `share` are what you are requesting me?

niboshi:
@mitmul Yes, that's what I meant.

mitmul:
Hi @beam2d and @nishino, now I'm fixing this PR and I need your advise.

The `repeat` method has been moved under `Link` class, and now `Link.copy()` takes `mode` option. So, `Sequential` class has `repeat()` method which actually calls `Link.repeat()`. This method repeats `self` `n_repeat` times, then `Sequential.repeat(n_repeat)` returns a `Sequential` object which has multiple `Sequential` objects like: `[Sequential, Sequtneial, ...]`.

But, I think the elements under `Sequential` is better to be extracted when the parent `Sequential` object is repeated. I mean, I think it's better to return

`Sequential(L.Linear(5, 5), F.relu, L.Linear(5, 5), F.relu)`

when `Sequential(L.Linear(5, 5), F.relu).repeat(2)` is called rather than

`Sequential(Sequential(L.Linear(5, 5), F.relu), Sequential(L.Linear(5, 5), F.relu))`.

How do you think?

niboshi:
Sematics of a superclass should not be altered. Otherwise we would not be able to consider `Sequential` as a `Link`.

If such expansion is needed, I'd suggest a different method.

mitmul:
@niboshi Thanks for the reply. How about adding another argument like `expand` to the `Link.repeat()`? And the default for `Sequential` class is `expand=True`, and for the others, the default is `expand=False`. Is it also included in semantics change?

niboshi:
@mitmul That would hinder extensibility of `Link.repeat`.  I mean, if `Sequential.repeat` had an extra argument, we would not be able to add a new argument to `Link.repeat` any more. I think that should be avoided.

beam2d:
The `expand` option only has effects when the link is Sequential, so I think there is even no need of overriding it by Sequential. It might be even ok to set `expand=True` by default in `Link.repeat` (and it's also ok to set it to `False` by default; it's a design choice). I prefer to set it to `False` by default (or even not providing this option) and instead provide `flatten()` method which flattens nested Sequentials.

By the way, I think the word `flatten` is better than `expand` because the latter has a different meaning in Chainer (adding an expansion to Trainer).

mitmul:
Sorry for late reaction. I'll take the last option (add `flatten()` method to `Sequential` class).

mitmul:
Jenkins, test this please

mitmul:
@beam2d @niboshi Please review the updates again. Thanks

chainer/link.py
```diff
@@ -559,6 +585,71 @@ def serialize(self, serializer):
...
+            paramters but its object ID is different from others. If it is
+            ``share``, each block is same to others in terms of not only
+            paramters but also the object IDs because they are shallow-copied,
+            so that when the parameter of one block is changed, all the
+            parameters in the others also change.
+
+        Args:
+            n_repeat (int): Number of times to repeat.
+            mode (str): It should be either ``init``, ``copy``, or ``share``.
+                ``init`` means paramters of each repeated element in the
```
beam2d: s/paramters/parameters/

chainer/link.py
```diff
@@ -559,6 +585,71 @@ def serialize(self, serializer):
...
+                returned :class:`~chainer.Sequential` will be re-initialized,
+                so that all elements have different initial parameters.
+                ``copy`` means that the parameters will not be re-initialized
+                but object itself will be deep-copied, so that all elements
+                have same initial parameters but can be changed independently.
+                ``share`` means all the elements which construct the resulting
```
beam2d: s/construct/consist/?

chainer/link.py
```diff
@@ -559,6 +585,71 @@ def serialize(self, serializer):
...
+                :class:`~chainer.Sequential` object are same object because
+                they are shallow-copied, so that all parameters of elements
+                share each other.
```
beam2d: are shared with each other? (maybe we need English proofreading)

chainer/link.py
```diff
@@ -298,31 +299,56 @@ def register_persistent(self, name):
...
+            ret._params = set(self._params)
+            ret._persistent = set(self._persistent)
+            ret.name = None
+            d = ret.__dict__
+            for name in ret._params:
+                d[name] = copy.copy(d[name])
+                d[name].grad = None
+            return ret
+        elif mode == 'copy' or mode == 'init':
+            ret = copy.deepcopy(self)
```
beam2d: What happens when the mode is `copy` and the link includes uninitialized parameters?

chainer/sequential.py
```diff
@@ -0,0 +1,400 @@
...
+              F.relu,
+              L.Linear(n_hidden, n_out)
+          )
+
+          # Compute the forward pass
+          y = model(x)
+
+        where ``x`` denotes a mini-batch of ``n_in``-dimensional input vectors.
+
+        Furthermore, :class:`~Sequential` supports build-in list APIs, so you
```
beam2d: s/build-in/built-in/

chainer/sequential.py
```diff
@@ -0,0 +1,400 @@
...
+          model.append(lambda x: x['prob'])
+
+          y = model(x)
+
+        The above code example shows how to add some layers to the ``model``
+        using :meth:`~Sequential.append` method and then add a large network
+        (``VGG16Layers``) and finally add a lambda function to extract the
+        ``prob`` output.
+
+        Note that a :class:`~Sequential` link which has at least one ``lmabda``
```
beam2d: s/lmabda/lambda/

chainer/sequential.py
```diff
@@ -0,0 +1,400 @@
...
+          # This is pickable
+          model = Sequential(
+              L.Convolution2D(None, 64, 3, 1, 1),
+              partial(F.max_pooling_2d, ksize=2)
+          )
+
+    Args:
+        layers: The layers which are called in its order. Each component should
+            be a callable object including :class:`~Link` object and
+            :class:`~Function` class.
```
beam2d: I don't think it's good to mention `Function` class here because it is not recommended to reuse the same `Function` object multiple times. It's ok to mention the functions defined under the `chainer.functions` namespace, e.g. `F.relu`.

chainer/sequential.py
```diff
@@ -0,0 +1,400 @@
...
+    def __call__(self, *x):
+        """Forward pass computation.
+
+        This method performs the forward pass computation by giving the input
+        variable ``x`` to the layers registered in the constructor in the same
+        order as the order in which the argments are given to the constructor.
+
+        It should be noted that the input variable is given directly to the
+        first layer and all intermediate outputs generated during the forward
+        pass are also directly feeded to the next layer. Therefore, the number
```
beam2d: s/feeded/fed/

chainer/sequential.py
```diff
@@ -0,0 +1,400 @@
...
+        of output at a layer should be the same as the number of input at the
```
beam2d: s/output/outputs/

beam2d: s/input/inputs/

chainer/sequential.py
```diff
@@ -0,0 +1,400 @@
...
+    def __iter__(self):
+        return iter(self._layers)
+
+    def __reversed__(self):
+        return reversed(self._layers)
+
+    def __contains__(self, item):
+        return item in self._layers
+
+    def __add__(self, other):
```
beam2d: Can we support adding a sequence which is not `Seqeuntial`? (e.g. Python list)

chainer/sequential.py
```diff
@@ -0,0 +1,400 @@
...
+        if isinstance(other, Sequential):
+            ret = Sequential()
+            for layer in self:
+                ret.append(layer)
+            for layer in other:
+                ret.append(layer)
+            return ret
+        elif callable(other):
```
beam2d: I think adding an element directly is inconsistent with the Python list semantics.

chainer/sequential.py
```diff
@@ -0,0 +1,400 @@
...
+        n_lambda = 0
+        for layer in self._layers:
+            if callable(layer) and hasattr(layer, '__name__') \
+                    and layer.__name__ == '<lambda>':
+                n_lambda += 1
+
+        if n_lambda > 0:
+            raise ValueError(
+                'This Sequential object has at least one lambda function as '
+                'its component. Lambda function can\'t be pickled, so please '
```
beam2d: s/can't/cannot/

chainer/sequential.py
```diff
@@ -0,0 +1,400 @@
...
+                'consider to use functools.partial instead of the lambda '
+                'function or use "dill" which is an external package that '
```
beam2d: Add a comma before "which".

chainer/sequential.py
```diff
@@ -0,0 +1,400 @@
...
+
+    def insert(self, i, layer):
+        if not callable(layer):
+            raise ValueError(
+                'All elements of the argment should be callable. But '
+                'given {} is not callable.'.format(layer))
+
+        self._layers.insert(i, layer)
+        if isinstance(layer, link.Link):
+            self.add_link(layer)
```
beam2d: It looks the order of appearances of links in sequential may not correspond to the order of links registered to ChainList. Is it ok? Consider the following code.
```py
seq = Sequential(link1, link2, link3)
seq.insert(0, link0)
```
In this case, the child links are registered like `seq/0 = link1, seq/1 = link2, seq/2 = link3, seq/3 = link0` while the Sequential looks like `[link0, link1, link2, link3]`.

chainer/link.py
```diff
@@ -298,31 +299,56 @@ def register_persistent(self, name):
...
+                ``init`` means parameter variables under the returned link
+                object is re-initialized by calling their
+                :meth:`~chainer.Parameter.initialize` method, so that all the
+                parameters may have different initial values from the original
+                link.
+                ``copy`` means that the link object is deeply copied, so that
+                its parameters are not re-initialized but are also deeply
+                copied. Thus, all parameters have same initial values but can
+                be changed independently.
+                ``share`` means that the link is shallowy copied, so that its
```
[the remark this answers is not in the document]
mitmul: Thanks. Fixed.

chainer/link.py
```diff
@@ -298,31 +299,56 @@ def register_persistent(self, name):
...
-        ret = copy.copy(self)
-        ret._params = set(self._params)
-        ret._persistent = set(self._persistent)
-        ret.name = None
-        d = ret.__dict__
-        for name in ret._params:
-            d[name] = copy.copy(d[name])
-            d[name].grad = None
-        return ret
+        if mode not in ['init', 'copy', 'share']:
```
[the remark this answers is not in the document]
mitmul: Thanks. Fixed.

chainer/link.py
```diff
@@ -298,31 +299,56 @@ def register_persistent(self, name):
...
+            ret = copy.copy(self)
+            ret._params = set(self._params)
+            ret._persistent = set(self._persistent)
+            ret.name = None
+            d = ret.__dict__
+            for name in ret._params:
+                d[name] = copy.copy(d[name])
+                d[name].grad = None
+            return ret
+        elif mode == 'copy' or mode == 'init':
```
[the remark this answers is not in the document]
mitmul: Thanks. Fixed.

chainer/link.py
```diff
@@ -559,6 +585,71 @@ def serialize(self, serializer):
         for name in self._persistent:
             d[name] = serializer(name, d[name])
 
+    def repeat(self, n_repeat, mode='init'):
+        """Repeat myself multiple times to make a :class:`~chainer.Sequential`.
```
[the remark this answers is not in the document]
mitmul: Thanks. Fixed.

chainer/link.py
```diff
@@ -559,6 +585,71 @@ def serialize(self, serializer):
...
+
+        This method returns a :class:`~chainer.Sequential` object which has
+        a same :class:`~chainer.Link` multiple times repeatedly. The ``mode``
+        argument means how to copy myself to repeat.
```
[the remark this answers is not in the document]
mitmul: Thanks. Fixed.

chainer/link.py
```diff
@@ -559,6 +585,71 @@ def serialize(self, serializer):
         for name in self._persistent:
             d[name] = serializer(name, d[name])
 
+    def repeat(self, n_repeat, mode='init'):
+        """Repeat myself multiple times to make a :class:`~chainer.Sequential`.
+
+        This method returns a :class:`~chainer.Sequential` object which has
+        a same :class:`~chainer.Link` multiple times repeatedly. The ``mode``
```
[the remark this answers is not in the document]
mitmul: Thanks. Fixed.

chainer/link.py
```diff
@@ -559,6 +585,71 @@ def serialize(self, serializer):
...
+        argument means how to copy myself to repeat.
+
+        .. admonition:: Example
+
+            You can repeat a same link multiple times to create longer
```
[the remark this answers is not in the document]
mitmul: Thanks. Fixed.

chainer/link.py
```diff
@@ -559,6 +585,71 @@ def serialize(self, serializer):
...
+                    def __call__(self, x):
+                        return F.relu(self.bn(self.conv(x)))
+
+                net = ConvBNReLU().repeat(16, mode='init')
+
+            The ``net`` object contains 16 blocks, each of which is
+            ``ConvBNReLU``. And the ``mode`` was ``init``, so each block
+            is re-initialized with different parameters. If you give
+            ``copy`` to this argument, each block has same values for its
+            paramters but its object ID is different from others. If it is
```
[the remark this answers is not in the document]
mitmul: Thanks. Fixed.

chainer/link.py
mitmul: Thanks. Fixed.

chainer/link.py
mitmul: Thanks. Fixed.

chainer/link.py
mitmul: Thanks. Fixed.

chainer/sequential.py
mitmul: Thanks. Fixed.

chainer/sequential.py
mitmul: Thanks. Fixed.

chainer/sequential.py
mitmul: Thanks. I modified this part to:

```
        layers: The layers which are called in its order. Each component should
            be a callable object including :class:`~Link` object and
            functions defined under the :mod:`chainer.functions`, e.g.,
            :obj:`~chainer.functions.relu`, etc.
```

chainer/sequential.py
mitmul: Thanks. Fixed.

chainer/sequential.py
mitmul: Thanks. Fixed.

chainer/sequential.py
mitmul: Thanks. Fixed.

chainer/sequential.py
mitmul: Thanks. Fixed.

mitmul:
Thanks for the review. I'm thinking about the questions/problems in your comments below now.

- What happens when the mode is copy and the link includes uninitialized parameters?
- Can we support adding a sequence which is not Seqeuntial? (e.g. Python list)
- I think adding an element directly is inconsistent with the Python list semantics.
- It looks the order of appearances of links in sequential may not correspond to the order of links registered to ChainList.

chainer/sequential.py
mitmul: Thanks. I removed this `elif` part.

chainer/sequential.py
mitmul: Hmm, right. Is that so problematic? I don't come up with a concrete situation which is problematic caused by this behavior. So if you can allow this, I would say it's OK.

chainer/sequential.py
mitmul: I think it's possible. Do you mean I should add that feature here?

chainer/link.py
mitmul: The elements of copied and original Sequential become different objects and both keep uninitialized.

mitmul:
@beam2d Thank for the offline discussion! I fixed all requested points and replied to some questions. Please check them again.

chainer/sequential.py
beam2d: I'm concerning serialization. The current spec means two `Sequential` objects may be serialized to different binaries even if they hold the same set of elements with the same parameters and the same order but with different orders on constructing them.

beam2d: For example, `Sequential(link0, link1, link2, link3)` behaves almost same as the above example, but is serialized into incompatible binary.

chainer/sequential.py
beam2d: I thought it may be convenient, but not sure. OK to leave it as is.

mitmul: OK, I'll add it to Sequential by another PR!

chainer/sequential.py
mitmul: Ah, got it. I'll fix it by changing all the indexes when any insertion occurred.

mitmul: Hm, just using `insert` instead of `add_link` resolves this problem?

mitmul: Ah, no. Change the order of links in `_children` list first, and then I need to change all their names.

mitmul: @beam2d Fixed this behavior. The code below seems OK.

```
In [1]: import chainer
In [2]: import chainer.links as L
In [3]: import chainer.functions as F
In [4]: s = chainer.Sequential(L.Linear(1, 2), F.relu, L.Linear(2, 3), F.relu, L.Linear(3, 4))
In [5]: print(s)
0       Linear  W(2, 1) b(2,)
1       relu
2       Linear  W(3, 2) b(3,)
3       relu
4       Linear  W(4, 3) b(4,)
In [6]: [c.name for c in s.children()]
Out[6]: ['0', '1', '2']
In [7]: s.insert(0, L.Linear(1, 1))
In [8]: [c.name for c in s.children()]
Out[8]: ['0', '1', '2', '3']
In [9]: print(s)
0       Linear  W(1, 1) b(1,)
1       Linear  W(2, 1) b(2,)
2       relu
3       Linear  W(3, 2) b(3,)
4       relu
5       Linear  W(4, 3) b(4,)
```

mitmul:
@beam2d I fixed the `insert` method of Sequential. Could you check it again?
https://github.com/chainer/chainer/pull/2918#discussion_r171468881

chainer/sequential.py
```diff
@@ -0,0 +1,400 @@
...
+    def __add__(self, other):
+        if isinstance(other, Sequential):
+            ret = Sequential()
+            for layer in self:
+                ret.append(layer)
+            for layer in other:
+                ret.append(layer)
+            return ret
+        else:
+            raise ValueError('add (+) operator supports only callable '
```
beam2d: Fix the error message (callable which is not Sequential is not allowed anymore)

chainer/sequential.py
```diff
@@ -0,0 +1,400 @@
...
+                             'objects, but {} is not callable'.format(
+                                 str(type(other))))
+
+    def __radd__(self, other):
```
beam2d: Do we still need radd? (callable is not allowed)

chainer/sequential.py
```diff
@@ -0,0 +1,400 @@
...
+        else:
+            raise ValueError('add (+) operator is support only callable '
+                             'objects, but {} is not callable'.format(
+                                 str(type(other))))
+
+    def __iadd__(self, other):
+        if isinstance(other, Sequential):
+            for layer in other:
+                self.append(layer)
+        elif callable(other):
```
beam2d: Remove it for consistency with `__add__`

chainer/sequential.py
```diff
@@ -0,0 +1,400 @@
...
+            del self[i]
+            self.insert(i, layer)
+
+    def __delitem__(self, i):
+        layer = self._layers.pop(i)
+        if isinstance(layer, link.Link):
+            for i, _link in enumerate(self._children):
+                if _link is layer:
+                    del self._children[i]
+                    break
```
beam2d: Shouldn't we rename the children of indexes greater than `i`?

chainer/sequential.py
```diff
@@ -0,0 +1,400 @@
...
+            if type_name == _name:
+                self.remove(_layer)
+
+    def pop(self, i=-1):
+        layer = self._layers[i]
+        del self[i]
+        return layer
+
+    def clear(self):
+        for i, _ in enumerate(self._children):
```
beam2d: It may be good to add a TODO comment to reduce the computational cost (or fix it now if it's simple).

chainer/sequential.py
mitmul: Ah, it's redundant. I removed it. Thanks.

chainer/sequential.py
mitmul: Yeah, thanks for pointing out it. I fixed it.

chainer/sequential.py
mitmul: Right. Modified.

chainer/sequential.py
mitmul: Added a comment.

mitmul:
@beam2d I added renaming part to __delitem__ after the deletion of the specified element. Could you check this PR again?

mitmul:
All the Jenkins tests passed. Could you check this again? @beam2d

chainer/sequential.py
```diff
@@ -0,0 +1,393 @@
...
+                'given {} is not callable.'.format(layer))
+
+        if self._layers[i] is not layer:
+            del self[i]
+            self.insert(i, layer)
+
+    def __delitem__(self, i):
+        layer = self._layers.pop(i)
+        if isinstance(layer, link.Link):
+            for i, _link in enumerate(self._children):
```
beam2d: Is it correct to delete the link whose index is the smallest? Can we assume that the same link object appears only once in the sequential? (I think inserting the same link multiple times is not considered in even ChainList, so we may ignore such situation)

chainer/sequential.py
```diff
@@ -0,0 +1,393 @@
...
+
+    def insert(self, i, layer):
+        if not callable(layer):
+            raise ValueError(
+                'All elements of the argment should be callable. But '
+                'given {} is not callable.'.format(layer))
+
+        self._layers.insert(i, layer)
+        if isinstance(layer, link.Link):
+            self._children.insert(i, layer)
```
beam2d: Is it correct to use the same index between `self._layers` and `self._children`?

chainer/sequential.py
mitmul: Thank you for pointing out this. I fixed this part to use the name of `layer` to determine which element should be removed from `_children` list.

chainer/sequential.py
mitmul: Thank you for pointing out this problem. If the given `layer` is a Link, the index used to insert it into `self._children` should be `j + 1` where `j` is the index of the last `link.Link` object before the `i`-th element in `self._layers`. So I fixed this method to perform this procedure.

mitmul:
@beam2d Thank you for the suggestion.

> How about just using _children by wrapping every non-link callable by a thin link that has no parameters and just calls the original callable?

That is basically possible, but if _children has non-parametric functions, `namedlinks()` from a Sequential object will include such functions besides links. It may not be intuitive for users who want to retrieve actual links which have parameters from a Sequential object. So I think it's better to use `_layers` in addition to `_children`.

mitmul:
@beam2d Hi, could you check this PR again? I confirmed this has passed all Jenkins tests and Travis tests.

beam2d:
LGTM

PR #2919, opened 2017-06-27
[WIP] Fix deadlock in MultiprocessIterator

### LINKED ISSUE #2909: MultiprocessingIterator Deadlock
I am new to chainer and trying to do a custom loop with MultiprocessingIterator. 

Reproduce on Ubuntu `16.04` and Chainer version `2.0`
````
import chainer
import numpy as np

dataset = np.ones((10000, 1000))
loader = chainer.iterators.MultiprocessIterator(
                dataset, 32, n_processes=6,
                shuffle=True, repeat=True)

while loader.epoch < 1:
    x = loader.next()
print('test')
loader.finalize() # hang at worker.join() (randomly, very frequently)
````

### PR DESCRIPTION
**Pull request by niboshi**
Fixes #2909

### DIFF
--- a/chainer/iterators/multiprocess_iterator.py
+++ b/chainer/iterators/multiprocess_iterator.py
@@ -147,8 +147,9 @@ class MultiprocessIterator(iterator.Iterator):
 
     def _init(self):
         finalized = threading.Event()
-        self._index_queue = multiprocessing.Queue()
-        self._data_queue = multiprocessing.Queue()
+        manager = multiprocessing.Manager()
+        self._index_queue = manager.Queue()
+        self._data_queue = manager.Queue()
         self._ordered_data_queue = six.moves.queue.Queue()
         self._unused_mem_queue = six.moves.queue.Queue()
         self._mem_list = []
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by niboshi** [comment]
Hmm, what causes `IOError`...?
```
  File "C:\Miniconda\lib\multiprocessing\queues.py", line 264, in _feed
    send(obj)
IOError: [Errno 232] The pipe is being closed
.....................................................Traceback (most recent call last):
  File "C:\Miniconda\lib\multiprocessing\queues.py", line 264, in _feed
    send(obj)
IOError: [Errno 232] The pipe is being closed
```

File: chainer/iterators/multiprocess_iterator.py [hunk 1]
```diff
@@ -129,8 +147,9 @@ def serialize(self, serializer):
 
     def _init(self):
         finalized = threading.Event()
-        self._index_queue = multiprocessing.Queue()
-        self._data_queue = multiprocessing.Queue()
+        manager = multiprocessing.Manager()
```
okuta: `multiprocessing.Manager()` creates process. And it causes problem.

niboshi:
Hmm... do you have any idea why this causes `IOError: [Errno 232] The pipe is being closed` (Only in Python 2.x)?

niboshi:
See #3076

PR #2923, opened 2017-06-28
Code to fetch the SVHN dataset

rcalland:
SVHN is very similar to MNIST. I added code that will download and create datasets for train and test samples the same way `get_mnist()` does.

This code requires `scipy` in order to read the matlab format that files are officially provided in.

### DIFF
--- a/chainer/datasets/__init__.py
+++ b/chainer/datasets/__init__.py
@@ -2,6 +2,7 @@ from chainer.datasets import cifar  # NOQA
 from chainer.datasets import dict_dataset  # NOQA
 from chainer.datasets import image_dataset  # NOQA
 from chainer.datasets import mnist  # NOQA
+from chainer.datasets import svhn # NOQA
 from chainer.datasets import ptb  # NOQA
 from chainer.datasets import sub_dataset  # NOQA
 from chainer.datasets import transform_dataset  # NOQA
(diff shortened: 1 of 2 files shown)


mattya:
Thank you for the PR!
I have two comments.
- Please add ```get_svhn``` in ```__init__.py``` (https://github.com/rcalland/chainer/blob/b1cb12d726078589bf99b31ab66f553a1ab717a9/chainer/datasets/__init__.py)
- It seems label _i_ corresponds to the number _i+1_ in the image and it is confusing.

rcalland:
The [SVHN website](http://ufldl.stanford.edu/housenumbers/) says:
`10 classes, 1 for each digit. Digit '1' has label 1, '9' has label 9 and '0' has label 10.`
which I guess is less confusing, how about switching to that system?

rcalland:
I just realised I misunderstood slightly the label definitions on the website, and so simply setting digit "0" from label 10 to 0 makes everything sensible!

mattya:
Thank you for the update.
It seems there are some flake8 errors in __init__.py and svhn.py.  Could you fix them?

mattya:
There are still following flake8 errors

>.\chainer\datasets\__init__.py:6:1: H306  imports not in alphabetical order (chainer.datasets.svhn, chainer.datasets.ptb)
>.\chainer\datasets\__init__.py:21:1: H306  imports not in alphabetical order (chainer.datasets.svhn.get_svhn, chainer.datasets.ptb.get_ptb_words)

rcalland:
My version of flake8 didn't show those errors. Fixed.

mattya:
Sorry for late reply, current CI is failing because scipy is not in the default dependency.
Could you add scipy availability check like this? https://github.com/cupy/cupy/blob/d7ad0d34f77de5b111807c0c1998176e9ab7e59a/cupy/sparse/csc.py#L1

mattya:
Thank you for fixing and sorry for late reply.
LGTM!

PR #2925, opened 2017-06-28
[EXPERIMENTAL] Add Cupy memory profiler

sonots:
Fix https://github.com/chainer/chainer/issues/918

This profiler uses function hook and relies on memory usage api of cupy which I added at https://github.com/cupy/cupy/pull/184 to profile behaviors of cupy memory pool.

An example code:

```python
from chainer.function_hooks import CupyMemoryProfileHook

hook = CupyMemoryProfileHook()
with hook:
    trainer.run()
hook.print_summary()
```

Outputs from mnist example become as followings:

```
FunctionName        UsedBytes AcquiredBytes Occurrence
LinearFunction      5.16GB    179.98MB      3900
ReLU                992.77MB  458.97MB      2600
SoftmaxCrossEntropy 5.76MB    5.08MB        1300
Accuracy            350.00KB  351.00KB      700
```

where `UsedBytes` is the memory bytes the function used from cupy memory pool, and `AcquiredBytes` is the actual memory bytes the cupy memory pool acquired from GPU device on the function call, and `Occurrence` is the number of calls.

### DIFF
--- a/chainer/function_hooks/__init__.py
+++ b/chainer/function_hooks/__init__.py
@@ -1,9 +1,11 @@
 from chainer.function_hooks import cuda_profile  # NOQA
+from chainer.function_hooks import cupy_memory_profile  # NOQA
 from chainer.function_hooks import debug_print  # NOQA
 from chainer.function_hooks import timer  # NOQA
 
 
 # import class and function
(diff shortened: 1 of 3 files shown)


sonots:
I prefer my newer version https://github.com/chainer/chainer/pull/2979. Let me close.

PR #2961, opened 2017-07-03
Make Variable unhashable

### LINKED ISSUE #2670: Variable defines __hash__ but does not define __eq__

niboshi:
Fixes #2670 

Reopening https://github.com/chainer/chainer/pull/2727

### DIFF
--- a/chainer/variable.py
+++ b/chainer/variable.py
@@ -1058,10 +1058,8 @@ Actual: {0}'''.format(type(data))
     def __bool__(self):
         raise NotImplementedError()
 
-    def __hash__(self):
-        return super(Variable, self).__hash__()
-
     __array_priority__ = 200
+    __hash__ = None
 
 
 class Parameter(Variable):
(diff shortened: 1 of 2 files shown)


niboshi:
This PR will be affected by #2970. So please wait until it is merged.

delta2323:
OK. Thank you for notifying it.

niboshi:
Rebased to master.

Now we don't have to touch `chainer/computational_graph.py`, because `VariableNode` and `FunctionNode` are now used as graph nodes and they are both hashable.

delta2323:
jenkins, test this please.

niboshi:
(Jenkins has failed, but it's not a fault of this PR.)

delta2323:
I ran the GPU test again.

delta2323:
LGTM

PR #2967, opened 2017-07-04
[WIP] Check cupy.available

niboshi:
(See: https://github.com/cupy/cupy/pull/247)

Check `cupy.available` when setting `chainer.cuda.available`, so that it will be `False` when there is no available device (by `CUDA_VISIBLE_DEVICES=-1`)

https://github.com/cupy/cupy/pull/247 should be merged first.

### DIFF
--- a/chainer/cuda.py
+++ b/chainer/cuda.py
@@ -34,27 +34,26 @@ import chainer
 
 available = False
 cudnn_enabled = False
+_resolution_error = None
 
 try:
     import cupy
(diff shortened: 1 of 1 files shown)


niboshi:
I have a trouble handling memory pool initialization (https://github.com/niboshi/chainer/blob/d69a027ddcf19c71cc386ebd0e3c88aa76d398a8/chainer/cuda.py#L132), which is currently done as module-level code. If we delay availability determination, I have no idea when should they be done.

PR #2977, opened 2017-07-06
Fix the docstring of Upsampling2D

mitmul:
The example shown in the docstring is somehow wrong.
The scope which avoids using cuDNN should be made right before applying `max_pooling_2d` function.
This PR fixes the example code.


mattya:
LGTM!

PR #2979, opened 2017-07-06
Cupy memory profiler with cupy memory hook

sonots:
Fix https://github.com/chainer/chainer/issues/918

This requires cupy memory hook added at https://github.com/cupy/cupy/pull/264

Code example::

    from chainer.function_hooks import CupyMemoryProfileHook
    hook = CupyMemoryProfileHook()
    with hook:
        trainer.run()
    hook.print_report()

Output example::

           FunctionName  UsedBytes  AcquiredBytes  Occurrence
         LinearFunction     5.16GB       179.98MB        3900
                   ReLU   991.82MB       458.97MB        2600
    SoftmaxCrossEntropy     7.71MB         5.08MB        1300
               Accuracy   617.97KB       351.00KB         700

where *FunctionName* is the name of function that calls the hook, and
*UsedBytes* is the memory bytes the function used from cupy memory
pool, and *AcquiredBytes* is the actual memory bytes the cupy memory
pool acquired from GPU device on the function call, and *Occurrence*
is the number of calls.

https://github.com/chainer/chainer/pull/2925 was simpler than this version, but not accurate. Therefore, I prefer this version.

### DIFF
--- a/chainer/function_hooks/cupy_memory_profile.py
+++ b/chainer/function_hooks/cupy_memory_profile.py
@@ -1,0 +1,111 @@
+import sys
+
+import six
+
+from chainer.cuda import memory_pool
+from chainer import function
+
+
+class CupyMemoryProfileHook(function.FunctionHook):
+    """Function hook for measuring memory usage of functions in cupy memory pool.
+
+    Example:
+        Code example::
+
+            from chainer.function_hooks import CupyMemoryProfileHook
+            hook = CupyMemoryProfileHook()
+            with hook:
+                trainer.run()
+            hook.print_report()
+
+        Output example::
+
+            FunctionName        UsedBytes AcquiredBytes Occurrence
+            LinearFunction      5.16GB    179.98MB      3900
+            ReLU                992.77MB  458.97MB      2600
+            SoftmaxCrossEntropy 5.76MB    5.08MB        1300
+            Accuracy            350.00KB  351.00KB      700
+
+        where *FunctionName* is the name of function that calls the hook, and
+        *UsedBytes* is the memory bytes the function used from cupy memory
+        pool, and *AcquiredBytes* is the actual memory bytes the cupy memory
+        pool acquired from GPU device on the function call, and *Occurrence*
+        is the number of calls.
+    Attributes:
+        call_history: List of measurement results. It consists of the function
+            that calls this hook,  the memory bytes the function used from cupy
+            memory pool, and the memory bytes the cupy memory pool acquired
+            from GPU device on the function call.
+    """
+
+    name = 'CupyMemoryProfileHook'
+
+    def __init__(self):
+        self.call_history = []
+
+    def _preprocess(self):
+        self.start_used_bytes = memory_pool.used_bytes()
+        self.start_total_bytes = memory_pool.total_bytes()
+
+    def forward_preprocess(self, function, in_data):
+        self._preprocess()
+
+    def backward_preprocess(self, function, in_data, out_grad):
+        self._preprocess()
+
+    def _postprocess(self, function):
+        used_bytes = memory_pool.used_bytes() - self.start_used_bytes
+        acquired_bytes = memory_pool.total_bytes() - self.start_total_bytes
+        self.call_history.append((function, used_bytes, acquired_bytes))
+
+    def forward_postprocess(self, function, in_data):
+        self._postprocess(function)
+
+    def backward_postprocess(self, function, in_data, out_grad):
+        self._postprocess(function)
+
+    def summary(self):
+        """Returns a summary of memory profiling in functions.
+
+        Returns:
+            A summarized dictionary whose keys are function names and
+            values are dictionaries of ``used_bytes``, ``acquired_bytes``,
+            and ``occurrrence``.
+        """
+        summary = {}
+        for func, used_bytes, acquired_bytes in self.call_history:
+            function_name = func.__class__.__name__
+            if function_name not in summary:
+                summary[function_name] = {'used_bytes': 0,
+                                          'acquired_bytes': 0, 'occurrence': 0}
+            record = summary[function_name]
+            record['used_bytes'] += used_bytes
+            record['acquired_bytes'] += acquired_bytes
+            record['occurrence'] += 1
+        return summary
+
+    def _humanized_size(self, size):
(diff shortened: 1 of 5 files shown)


chainer/function.py
```diff
@@ -622,11 +627,33 @@ def __enter__(self):
...
+        """Callback function invoked when the hook is added.
+
+        Args:
+            function(~chainer.Function): Function object to which
+                the function hook is registered. None if the hook
+                is registered via ``with`` statement.
+        """
+        pass
+
+    def deleted(self, function=None):
```
sonots: I needed this additional hook to successfully clean cupy memory hook registered via both `add_hook` and `__enter__` (with statement).

sonots:
I wonder if the name can be simply `MemoryHook` or not.

chainer/function_hooks/cupy_memory_profile.py
```diff
@@ -0,0 +1,155 @@
...
+                       Accuracy   617.97KB       351.00KB         700
+
+        where *FunctionName* is the name of function that calls the hook, and
+        *UsedBytes* is the memory bytes the function used from cupy memory
+        pool, and *AcquiredBytes* is the actual memory bytes the cupy memory
+        pool acquired from GPU device on the function call, and *Occurrence*
+        is the number of calls.
+    Attributes:
+        call_history: List of measurement results. It consists of the function
+            that calls this hook,  the memory bytes the function used from cupy
```
sonots: thanks, fixed

delta2323:
My team tried to use this feature to estimate the memory usage of some architecture. We notice that we sometimes want the profile more finer resolution. Specifically as some architecture uses same type of layers many times, it would be better if we can get these information for each layer. But it could be difficult because of define-by-run nature of Chainer.

sonots:
Thank you for feedback. I understand the demand, but it looked difficult to do it with current framework.

It would be possible if we create a new hook framework for new FunctionNode by any chances?

delta2323:
I am sorry I could not understand what the nested case is like. Could you give me some examples or test cases for it?

sonots:
it is same thing with > https://github.com/chainer/chainer/pull/3019

delta2323:
OK, then could you write test cases for this usage?

sonots:
yes, sure.

sonots:
Tests are added

sonots:
https://github.com/chainer/chainer/pull/2979#issuecomment-316608822

> Specifically as some architecture uses same type of layers many times, it would be better if we can get these information for each layer.

I talked with tokui-san. We are planning (or hoping) to visualize network graph. To do it we have to distinguish each function on a network, so we may be able to use such information.  But, it is not designed well yet. I leave this issue as a future plan for now.

delta2323:
Thank you for considering the possibility of the extension of this PR. I agree with you in that we should leave the issue and concentrate on finishing the PR with current feature.

chainer/function_hooks/cupy_memory_profile.py
```diff
@@ -0,0 +1,165 @@
+import sys
+
+from chainer import function
+
+from cupy.cuda import memory_hook
```
delta2323: It seems this line raises an error when CuPy is not imported. So, could you check the availability of CuPy and do some fallback if it is not found?

chainer/function_hooks/cupy_memory_profile.py
```diff
@@ -0,0 +1,165 @@
...
+        template = '  '.join('{:>%d}' % w for w in entry_widths)
+        for function_name, used_bytes, acquired_bytes, occurrence in entries:
+            line = template.format(
+                function_name, used_bytes, acquired_bytes, occurrence)
+            file.write(line)
+            file.write('\n')
+        file.flush()
+
+
+class CupyMemoryCumulativeHook(memory_hook.MemoryHook):
```
delta2323: Should we move `CuPyMemoryCumulativeHook` to CuPy?

sonots: I want to place at here because moving to cupy has some disadvantages.

* When requirements of cupy memory profile is changed, we have to modify both cupy and chainer
    1. PR will be seperated
    2. CI will fail until cupy side is merged
    3. Users have to upgrade both chainer and cupy

chainer/function_hooks/cupy_memory_profile.py
sonots: Fixed as https://github.com/chainer/chainer/pull/2979/commits/98c0c33a57ad7a8321e6527304e0afc8041b7c79

sonots:
Rebased to master.

tests/chainer_tests/function_hooks_tests/test_cupy_memory_profile.py
```diff
@@ -0,0 +1,188 @@
+import unittest
+
+import numpy
+import six
+
+import chainer
+from chainer import cuda
+from chainer.cuda import memory_pool
```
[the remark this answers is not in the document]
sonots: Fixed.

delta2323:
Could you put `st:ready-for-review` label when it is ready.

sonots:
In cupy, I can check and change label,

image

But, in chainer, it seems I can not do it although I should have member permission 🤔 

image

sonots:
@delta2323 ready to review, anyway.

delta2323:
OK, I will review again. And I will also check the authentication problem.

delta2323:
jenkins, test this please.

delta2323:
Could you check the jenkins result?

delta2323:
Thank you for checking the result and fixing the code. You substituted `Exp` with `Floor`, which is an instance of `Function` that is callable. But I expect that `Floor` will also become `FunctionNode` to support double backward in the future. I would recommend that it is better to prepare for that time. I mean, `self.f` should be a function that calls `Exp.apply`.

sonots:
Okay, also `Log` is not FunctionNode, yet. I need to change it to one of FunctionNode.

sonots:
fixed, please wait CI passes.

sonots:
waiting jenkins CI passes

delta2323:
jenkins, test this please.

delta2323:
jenkins, test this please.

delta2323:
LGTM

PR #2985, opened 2017-07-10
Deterministic test (Chainer)

niboshi:
This is Chainer counterpart of https://github.com/cupy/cupy/pull/82.

(See also: https://github.com/cupy/cupy/pull/222)

### DIFF
--- a/chainer/testing/random.py
+++ b/chainer/testing/random.py
@@ -110,22 +110,23 @@ def fix_random():
             # Applied to test case class
             klass = impl
 
-            def wrap_setUp(f):
-                def func(self):
-                    _setup_random()
-                    f(self)
-                return func
+            setUp_ = klass.setUp
+            tearDown_ = klass.tearDown
 
-            def wrap_tearDown(f):
-                def func(self):
-                    try:
-                        f(self)
-                    finally:
-                        _teardown_random()
-                return func
+            @functools.wrap(setUp_)
+            def setUp(self):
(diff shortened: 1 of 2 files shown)


chainer/testing/random.py
```diff
@@ -0,0 +1,133 @@
...
+                try:
+                    impl(self, *args, **kw)
+                finally:
+                    _teardown_random()
+            return test_func
+        elif isinstance(impl, type) and issubclass(impl, unittest.TestCase):
+            # Applied to test case class
+            klass = impl
+
+            def wrap_setUp(f):
```
okuta: How about using simple way?
``` python
_setUp = klass.setUp
@functools.wraps(_setUp)
def setUp(self):
    _setup_random()
    _setUp(self)
klass.setUp = setUp
```

chainer/testing/random.py
```diff
@@ -0,0 +1,133 @@
...
+                def func(self):
```
okuta: How about using `@functools.wraps`?

chainer/testing/random.py
```diff
@@ -0,0 +1,133 @@
...
+                    _setup_random()
+                    f(self)
+                return func
+
+            def wrap_tearDown(f):
```
okuta: ditto

niboshi:
Thank you for suggestion! Fixed.

okuta:
LGTM!

PR #2991, opened 2017-07-10
Select the best-resolution timer function

niboshi:
AppVeyor is failing due to insufficient timer resolution (probly).

This PR selects the best-resolution timer in the current environment.

([Python 3 time module documentation](https://docs.python.org/3/library/time.html))

**For Python >=3.3**
* `time.perf_counter` has the highest resolution.

**For Python <3.3**
 * `time.clock`: On Windows, it uses `QueryPerformanceCounter` Win32 API function, which has a good resolution. On Unix, it returns the "processor time", whose definition depends on C `clock` function.
 * `time.time`: Returns the time since the epoch. Not monotonic.

### DIFF
--- a/chainer/training/trainer.py
+++ b/chainer/training/trainer.py
@@ -14,6 +14,16 @@ from chainer.training import extension as extension_module
 from chainer.training import trigger as trigger_module
 
 
+# Select the best-resolution timer function
+try:
+    _get_time = time.perf_counter
+except AttributeError:
+    if os.name == 'nt':
+        _get_time = time.clock
+    else:
+        _get_time = time.time
+
+
 class _ExtensionEntry(object):
 
     def __init__(self, extension, priority, trigger, invoke_before_training):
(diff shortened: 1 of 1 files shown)


chainer/training/trainer.py
```diff
@@ -14,6 +14,16 @@
 from chainer.training import trigger as trigger_module
 
 
+# Select the best-resolution timer function
+try:
+    _get_time = time.perf_counter
```
okuta: How about using `getatter`?

niboshi: Is there any reason to prefer `getattr`?
IMO, `try..catch` way is

* Clean and straight-forward
* EAFP is more pythonic than LBYL
* Faster (speed is not critical in this case, though)

okuta: OK

okuta:
LGTM!

Issue #2984, opened 2017-07-10
Links to GitHub pages are broken


niboshi:
(Rel. #2758)

Links to GitHub pages do not appear in the  Read-the-Docs build (e.g. https://docs.chainer.org/en/stable/reference/generated/chainer.functions.sigmoid.html#chainer.functions.sigmoid).

It works in locally built documentation.

(Thanks @soskek for reporting)

Crissman:
Should be fixed bye #2995. Please test, confirm fixed, and close.

Issue #2822, opened 2017-06-05
The "Edit on GitHub" links in the document are wrong.


kiyukuta:
For example, the link is on the top right of this page.
https://docs.chainer.org/en/stable/index.html

We have to remove "origin" from the URL:
https://github.com/chainer/chainer/blob/origin/v2.0.0/docs/source/index.rst .

kiyukuta:
related #1034 (reoccured)

delta2323:
Thank you for reporting. The original issue rtfd/readthedocs.org#1637 is still open. So I think we should wait to see it.

niboshi:
Do we need this link from the beginning?
I often confuse it as a link to `.py` source file...

kiyukuta:
Thanks for the link of original issue !
I hope it will be fixed soon.

@niboshi 
Yeah, I'm also confused sometime. it will be better to jump `*.py` directly :)

kiyukuta:
Since this is not a Chainer's problem, should I close this issue ?

niboshi:
Maybe we can discuss how to deal with the link here. (remove? replace? leave as is? ... etc)
I'd vote for removal.

kiyukuta:
I also think it is OK to remove the links. 

On the other hand, the links will be help when we found some issue on the docs (e.g. typo).
We can easily find which `rst` file has to be changed.

niboshi:
Reopening the issue, because this link is still showing in our documentation.

Crissman:
Fixed by PR #2999?

PR #2995, opened 2017-07-11
Fix read the docs Fix on Github link

Crissman:
Fix for issue #2984 
Add requirement for Sphinx version 1.3.6 which should fix this issue
https://github.com/rtfd/readthedocs.org/issues/1871

Unable to test locally, since depends on readthedocs.org respecting Sphinx requirement

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -31,7 +31,7 @@ on_rtd = os.environ.get('READTHEDOCS', None) == 'True'
 # -- General configuration ------------------------------------------------
 
 # If your documentation needs a minimal Sphinx version, state it here.
-#needs_sphinx = '1.0'
+needs_sphinx = '1.3.6'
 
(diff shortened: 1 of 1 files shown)


niboshi:
Hmm, I got this error when building with rtd-build:
```
Sphinx version error:
This project needs at least Sphinx v1.3.6 and therefore cannot be built with this version.
```

```
# python -m sphinx --version
Sphinx (sphinx-build) 1.6.3
```

niboshi:
I think the "GitHub links" discussed on https://github.com/rtfd/readthedocs.org/issues/1871 is about "Edit on GitHub" link (which points to source of *documentation*). If that's the case, issue #2822 is about this problem.

#2984 is about links to sources of Chainer classes, methods, etc. (I'm sorry if it was not clear).

niboshi:
I made another PR to fix this problem (#2984).

PR #3016, opened 2017-07-13
Revival batch_matmul for backward compatibility

okuta:
This PR provides compatibility about `batch_matmul`.

`batch_matmul` removed by #2426. But, we can provides it easily.

### DIFF
--- a/chainer/functions/math/matmul.py
+++ b/chainer/functions/math/matmul.py
@@ -1,3 +1,5 @@
+import warnings
+
 import numpy
 
 from chainer import cuda
@@ -153,4 +155,31 @@ def matmul(a, b, transa=False, transb=False):
                [ 2.,  2.]], dtype=float32)
 
     """
+    return MatMul(transa=transa, transb=transb)(a, b)
+
+
+def batch_matmul(a, b, transa=False, transb=False):
+    """Computes the batch matrix multiplications of two sets of arrays.
+
+    Args:
+        a (Variable): The left operand of the batch matrix multiplications.
+            A 2-D array of shape ``(B, N)`` is considered as B
+            :math:`N \\times 1` matrices.
+            A 3-D array of shape ``(B, M, N)`` is considered as B
+            :math:`M \\times N` matrices.
+        b (Variable): The right operand of the batch matrix multiplications.
+            Its array is treated as matrices in the same way as ``a``'s array.
+        transa (bool): If ``True``, transpose each matrix in ``a``.
+        transb (bool): If ``True``, transpose each matrix in ``b``.
+
+    Returns:
+        ~chainer.Variable: The result of the batch matrix multiplications as a
+            3-D array.
+
+    .. deprecated:: v3.0.0
+       batch_matmul is deprecated. Use ``matmul`` instead.
+
+    """
+    warnings.warn('batch_matmul is deprecated. Use matmul instead.',
+                  DeprecationWarning)
     return MatMul(transa=transa, transb=transb)(a, b)
--- a/chainer/functions/math/matmul.py
+++ b/chainer/functions/math/matmul.py
@@ -53,38 +53,11 @@ def _matmul(a, b, transa=False, transb=False, transout=False):
         return numpy.einsum('...ij,...jk->...ik', a, b)
 
 
-def _batch_matmul(a, b, transa=False, transb=False, transout=False):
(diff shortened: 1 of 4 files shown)


muupan:
I think this behaves differently from old `batch_matmul`. The old `batch_matmul` supports 3D x 2D and 2D x 3D but the new `matmul` doesn't.

delta2323:
@okuta have restored old unit tests (rewriting them with `testing.parameterize`). It seems we did not we have unit tests for 2D x 3D and 3D x 2D cases.

muupan:
It seems we didn't have tests for 3D x 2D and 2D x 3D, but batch_matmul actually supported them as documented. You can confirm it in https://github.com/chainer/chainer/issues/1765.

The tests in this PR are for numpy_like_matmul, not for batch_matmul. batch_matmul had tests for batch vector x vector (inner or outer) product: https://github.com/chainer/chainer/blob/895ebb785f9ab89e54acb5bff77f1c0324b4dceb/tests/chainer_tests/functions_tests/math_tests/test_matmul.py#L206-L233

okuta:
Thank you for your information.
I fixed test and implementation.

tests/chainer_tests/functions_tests/math_tests/test_matmul.py
```diff
@@ -126,6 +126,119 @@ def test_matmul_backward_gpu(self):
             cuda.to_gpu(self.gy), atol=1e-2, rtol=1e-2)
 
 
+@testing.parameterize(*testing.product_dict(
+    [
+        # matmul
```
toslunar: maybe `# batched matmul 2d x 2d`

tests/chainer_tests/functions_tests/math_tests/test_matmul.py
```diff
@@ -126,6 +126,119 @@ def test_matmul_backward_gpu(self):
...
+    def _get_forward_answer(self, x1, x2, transa, transb):
+        x1 = x1.reshape(x1.shape[:2] + (-1,))
+        if transa and x1.ndim >= 2:
+            x1 = x1.swapaxes(-1, -2)
+
+        x2 = x2.reshape(x2.shape[:2] + (-1,))
+        if transb and x2.ndim >= 2:
+            x2 = x2.swapaxes(-1, -2)
+
+        if x1.ndim <= 2:
```
toslunar: The condition is always False.

tests/chainer_tests/functions_tests/math_tests/test_matmul.py
```diff
@@ -126,6 +126,119 @@ def test_matmul_backward_gpu(self):
...
+        self.gy = numpy.random.uniform(-1, 1, self.gy_shape).astype(ret_dtype)
+
+        self.op = lambda x, y: F.batch_matmul(
+            x, y, transa=self.transa, transb=self.transb)
+        self.forward_answer = self._get_forward_answer(
+            self.x1, self.x2, self.transa, self.transb)
+
+    def _get_forward_answer(self, x1, x2, transa, transb):
+        x1 = x1.reshape(x1.shape[:2] + (-1,))
+        if transa and x1.ndim >= 2:
```
toslunar: `x1.ndim` is 3, here.

chainer/functions/math/matmul.py
```diff
@@ -154,3 +156,82 @@ def matmul(a, b, transa=False, transb=False):
...
+    def __init__(self, transa=False, transb=False):
+        self.transa = transa
+        self.transb = transb
+
+    def check_type_forward(self, in_types):
+        type_check.expect(in_types.size() == 2)
+        a_type, b_type = in_types
+
+        type_check.expect(
+            a_type.dtype == numpy.float32,
```
toslunar: While `MatMul`'s typecheck is `a_type.dtype.kind == 'f'`, it's also fine to restrict to float32, because `BatchMatMul` will revive for compatibility.  Do you have ideas?

okuta: Let's expand acceptable dtype

okuta: This function is deprecated. Thus, I think that change is unnecessary

okuta:
I fixed.

toslunar:
LGTM

beam2d:
LGTM

PR #2927, opened 2017-06-28
Add print_report() method to Timer hook

sonots:
Added `print_summary()` method to Timer hook so that we can do time profile similarly with the memory profiler of https://github.com/chainer/chainer/pull/2925

An example code:

```python
from chainer.function_hooks import TimerHook

hook = TimerHook()
with hook:
    trainer.run()
hook.print_report()
```

Outputs from mnist example become as followings:

```
FunctionName    ElapsedTime     Occurrence
LinearFunction  1.17sec 3900
ReLU    547.88ms        2600
SoftmaxCrossEntropy     968.39ms        1300
Accuracy        155.53ms        700
``` #

PR #3019, opened 2017-07-14
Make TimerHook reentrant

beam2d:
This PR makes `TimerHook` reentrant, i.e., the forward hook and backward hook can be nested. It is required by #2970 with which the backward of each function is written by Function calls rather than direct array computation.

### DIFF
--- a/chainer/function_hooks/timer.py
+++ b/chainer/function_hooks/timer.py
@@ -41,6 +41,8 @@ class TimerHook(function.FunctionHook):
     def __init__(self):
         self.call_history = []
         self._running_stack = []
+        self._depth = 0
+        self._total_time = 0
 
     def _preprocess(self):
         if self.xp == numpy:
@@ -51,6 +53,7 @@ class TimerHook(function.FunctionHook):
             stop = cuda.Event()
             start.record()
             self._running_stack.append((start, stop))
+        self._depth += 1
 
     def forward_preprocess(self, function, in_data):
         self.xp = cuda.get_array_module(*in_data)
@@ -74,6 +77,11 @@ class TimerHook(function.FunctionHook):
                 start, stop) / 1000
         self.call_history.append((function, elapsed_time))
 
+        assert self._depth > 0
+        self._depth -= 1
+        if self._depth == 0:
+            self._total_time += elapsed_time
+
     def forward_postprocess(self, function, in_data):
         xp = cuda.get_array_module(*in_data)
         assert xp == self.xp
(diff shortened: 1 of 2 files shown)


tests/chainer_tests/function_hooks_tests/test_timer.py
```diff
@@ -98,5 +99,27 @@ def test_backward_cpu(self):
     def test_backward_gpu(self):
         self.check_backward(cuda.to_gpu(self.x), cuda.to_gpu(self.gy))
 
+    def test_reentrant(self):
+        # In/grad data are random; these do not simulate the actually possible
+        # cases.
+        g = functions.Identity()  # any time other than Exp is ok
```
niboshi: Do you mean "any function"?

niboshi:
Check my comment.
LGTM otherwise.

sonots:
Does not this sum elapsed times for nested functions duplicatedly?

I mean that if following scenario is happened:

```
1. FunctionA forward
2.   FunctionB backward
```

The elapsed time for 1 includes the elapsed time for 2.
Thus, the `total_time` sums the elapsed time for 2. twice.

sonots:
One idea is to resolve the issue is to add a new parameter such as `depth` for `call_history` which is incremented in nested calls. Then, `total_time` can be computed by picking histories of `depth` 0.

sonots:
Since I've added `summary` and `print_report` methods at https://github.com/chainer/chainer/pull/2927, they also need to be fixed.

sonots:
I feel the elapsed time for `f(g(x))` should not be added to `call_history` of `f`, but should be added to a `call_history` of a composed function `f・g`, which is a `FunctionNode` now (? I may not well understand https://github.com/chainer/chainer/pull/2970).

tests/chainer_tests/function_hooks_tests/test_timer.py
beam2d: Oh, you are right. I'll fix it.

beam2d:
Thank you for pointing out the duplicated counting of nested calls. Adding `depth` is reasonable for now.

As for `FunctionNode`, it is a replacement of the current `Function` that allows the `backward` method to be differentiable (i.e., the computation made in `backward` is also recorded for further backprop). There would be a case that users want to analyze the performance of inner working of `backward` method, in which case it is not a good idea to skip calling hook functions for FunctionNodes applied inside `backward`.

Anyway, I feel the current hook function system does not fit well to higher order differentiation settings. It is good to think of improving this point in another issue/PR.

beam2d:
Fixed and rebased onto the current master.

chainer/function_hooks/timer.py
```diff
@@ -81,7 +94,8 @@ def backward_postprocess(self, function, in_data, out_grad):
 
     def total_time(self):
         """Returns total elapsed time in seconds."""
-        return sum(t for (_, t) in self.call_history)
+        return self._total_time
+        # return sum(t for (_, t) in self.call_history)
```
niboshi: Please remove this comment.

niboshi:
LGTM!

PR #3032, opened 2017-07-18
Import example modules and util in chainer.training.__init__

### LINKED ISSUE #2981: inconsistency to use example module
I found some inconsistent behavior to import classes.

To use SOME_CLASS in chainer/iterators, the following code works.
```importIterator.py
import chainer
some_class = chainer.iterators.SOME_CLASS() # OK
```
However, to use SOME_CLASS in chainer/training/extensions, the following code does not work.
```importExtension.py
import chainer
some_class = chainer.training.extensions.SOME_CLASS() #NG
```

I feel some inconsistency here, because `chainer/__init__.py` contains both of `import iterator` and `import iterators` but `chaner/training/__init__.py` contains only `import extension` and not `import extensions`.

Is there any reason about this inconsistency?

* Conditions
  - Chainer version
    chainer v2.0.1
  - OS/Platform
    ubuntu 16.04
* Code to reproduce
```importModule.py
import chainer
iteratorExamples = chainer.iterators # OK

import chainer
extensionExamples = chainer.training.extensions #NG
```

* Error messages
```error.py
AttributeError: module 'chainer.training' has no attribute 'extensions'
```

delta2323:
Fix #2981 

This PR adds import of modules that contain example training modules and `chainer.training.util` to `chainer.training.__init__`.

### DIFF
--- a/chainer/training/__init__.py
+++ b/chainer/training/__init__.py
@@ -1,7 +1,11 @@
 from chainer.training import extension  # NOQA
+from chainer.training import extensions # NOQA
 from chainer.training import trainer  # NOQA
 from chainer.training import trigger  # NOQA
+from chainer.training import triggers  # NOQA
 from chainer.training import updater  # NOQA
+from chainer.training import updaters  # NOQA
+from chainer.training import util  # NOQA
 
 
 # import class and function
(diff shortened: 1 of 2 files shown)


niboshi:
Please fix style and import errors.

niboshi:
LGTM!

PR #3043, opened 2017-07-20
Avoid to import matplotlib to set its backend Agg in code

mitmul:
Setting `matplotlib.use('Agg')` inside code might cause errors in some environments other than linux.
Using `MPLBACKEND=Agg` environmental variable is more preferable I think.

### DIFF
--- a/examples/mnist/train_mnist.py
+++ b/examples/mnist/train_mnist.py
@@ -2,12 +2,6 @@
 
 from __future__ import print_function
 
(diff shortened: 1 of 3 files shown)


beam2d:
Note: it looks the CI script for example tests also needs to be updated.

beam2d:
jenkins, test this please

beam2d:
LGTM

PR #3050, opened 2017-07-24
Ignore generated references and tags

niboshi:
Port of https://github.com/cupy/cupy/pull/304 and https://github.com/cupy/cupy/pull/319


delta2323:
LGTM

PR #3056, opened 2017-07-25
Avoid running a hook on uninitialized parameters

### LINKED ISSUE #3041: Optimizer tries to update uninitialized parameters
MWE: https://pastebin.com/DXp4msHx
Running the code will result in the following error:
```
  File "/home/tommi/.pyenv/versions/2.7.12/lib/python2.7/site-packages/chainer/optimizer.py", line 628, in __call__
    g += self.rate * p
TypeError: unsupported operand type(s) for *: 'float' and 'NoneType'
```
The issue is that the parameter `p` above is `None` due to `model.l3` having no initialized parameters yet on the first forward pass.
This happens since when we are using shape placeholders, `model.l3.params()` returns an empty set, which the optimizer tries to update. An easy fix for this is to write the code as `self.l3 = L.Linear(n_units, n_units)` instead, but I think it is confusing for Chainer users that this error occurs when using placeholders. Would it perhaps be possible to modify the optimizer so that it will not update links with uninitialized parameters?

tkerola:
Fixes #3041 
This fixes the issue in the MWE linked in the issue (https://pastebin.com/DXp4msHx).
Will add a test a bit later.

### DIFF
--- a/chainer/optimizer.py
+++ b/chainer/optimizer.py
@@ -185,7 +185,12 @@ class UpdateRule(object):
             return
 
         self.t += 1
+
+        if param.data is None:
+            return
+
         self._prepare(param)
+
         for hook in six.itervalues(self._hooks):
             hook(self, param)
         self.update_core(param)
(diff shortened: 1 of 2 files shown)


niboshi:
As @yuyu2172 told in #3041, how about fixing each hook function, instead of avoid calling them?
There could be hooks which does something with an unitialized parameters.

tkerola:
Sorry for the delay. I will fix each hook function instead, as suggested.

tkerola:
I created a new PR that fixes each hook function instead in #3170, so I am closing this one.

PR #3067, opened 2017-07-28
Wording in contribution guide

niboshi:
*Guideline* is rather a policy than a concrete procedural direction.
I also fixed a broken markup.

TODO: port to CuPy


delta2323:
LGTM

PR #3070, opened 2017-07-28
[WIP] Add retain_all_inputs configuration

mitmul:
Some people want to keep all inputs data of all functions after forward computation.
This PR adds a config value `chainer.config.retain_all_inputs` and `CHAINER_RETAIN_ALL_INPUTS` environment variable.
If it's set to True, Function.retain_inputs does nothing, so the given tuple as an argument is ignored. Then all inputs to any functions are retained even after the forward computation.

- [ ] Write test
- [ ] Add docs more


beam2d:
We can do this just by writing a small function hook (that calls `retain` method of all input variable nodes). It's much cleaner. I close this issue for now; if someone wants this feature provided built-in by Chainer, feel free to open a new issue to request that!

PR #3076, opened 2017-07-31
Improve MultiprocessIterator performance, functionality and stability, using Pool

grafi-tt:
I rewrote the `MultiprocessIterator` using Python's `Pool` functionality. It makes codes simpler and faster, as well as making orphan/zombie process less likely.

Instead of previous "one prefetch invoked at one fetch" implementation, in my implementation background thread invokes prefetch freely, only if the maximum prefetch number allows. I think this strategy is better for utilizing idle CPU/IO time.

Resetting an iterator in the mid of iteration is also supported. Tests are changed accordingly.

The second commit changes per-batch-item shared memory allocation to the bulk allocation. This improved performance a little. 

I took a benchmark of `examples/mnist/train_mnist.py`, with `SerialIterator` changed to `MultiprocessIterator`.  It was done on EC2 g3.4xlarge instance.

```
time python train_mnist.py -g 0
baseline:          537.91s user 170.63s system 263% cpu 4:28.87 total
before bulk alloc: 234.32s user 27.87s  system 215% cpu 2:01.85 total
after bulk alloc:  227.92s user 26.92s  system 215% cpu 1:58.23 total
```

### DIFF
--- a/chainer/iterators/multiprocess_iterator.py
+++ b/chainer/iterators/multiprocess_iterator.py
@@ -344,9 +344,14 @@ class _PrefetchLoop(object):
 # Using `parametarized` funciton (e.g. bound method) with Pool is tricky due to
 # restrictions imposed by Pickle. Picklable types differ across versions.
 # Just using top-level function with globals seems to be safest.
-
 # it doesn't mean thread safety broken or global variables visible;
 # notice that each process uses different address space.
+# To make static linker happy, we first initialize global variables.
+_fetch_dataset = None
+_fetch_mem_size = None
+_fetch_mem_bulk = None
+
+
 def _fetch_setup(dataset, mem_size, mem_bulk):
     global _fetch_dataset, _fetch_mem_size, _fetch_mem_bulk
     _fetch_dataset = dataset
--- a/chainer/iterators/multiprocess_iterator.py
+++ b/chainer/iterators/multiprocess_iterator.py
@@ -255,13 +255,13 @@ class _PrefetchLoop(object):
             if indices is None:  # stop iteration
                 batch = None
             elif pool is None:  # measure mode
-                batch = list(self.dataset[idx] for idx in indices)
+                batch = [self.dataset[idx] for idx in indices]
                 self.mem_size = max(map(_measure, batch))
                 self._allocate_shared_memory()
             else:
                 data_it = pool.imap(
                     _fetch_run, enumerate(indices), chunk_size)
-                batch = list(_unpack(data, self.mem_bulk) for data in data_it)
+                batch = [_unpack(data, self.mem_bulk) for data in data_it]
 
             self.comm.put(batch, self.prefetch_state, reset_count)
             return True
--- a/chainer/iterators/multiprocess_iterator.py
+++ b/chainer/iterators/multiprocess_iterator.py
@@ -97,11 +97,11 @@ class MultiprocessIterator(iterator.Iterator):
             self.dataset, self.batch_size, self.repeat, self.shuffle,
             self.n_processes, self.n_prefetch, self.shared_mem)
 
-        comm = self._comm
-        del self._comm
-        other.__dict__.update(self.__dict__)
-        self._comm = comm
-        other.__class__ = self.__class__
+        other.current_position = self.current_position
+        other.epoch = self.epoch
+        other.is_new_epoch = self.is_new_epoch
+        other._previous_epoch_detail = self._previous_epoch_detail
+        other._order = self._order
 
         other._set_prefetch_state()
         return other
--- a/chainer/iterators/multiprocess_iterator.py
+++ b/chainer/iterators/multiprocess_iterator.py
@@ -2,6 +2,7 @@ from __future__ import division
 from collections import namedtuple
 import multiprocessing
 from multiprocessing import sharedctypes
+import signal
 import threading
 import warnings
 
@@ -11,6 +12,7 @@ import six
 from chainer.dataset import iterator
 
 
+_long_time = 10000000
 _PrefetchState = namedtuple('_PrefetchState', (
     'current_position', 'epoch', 'is_new_epoch',
     'previous_epoch_detail', 'order'))
@@ -63,12 +65,13 @@ class MultiprocessIterator(iterator.Iterator):
         self._comm = _Communicator(self.n_prefetch)
         self.reset()
 
-        # don't need to hold a reference to the thread
         prefetch_loop = _PrefetchLoop(
             self.dataset, self.batch_size, self.repeat, self.shuffle,
             self.n_processes, self.n_prefetch, self.shared_mem, self._comm)
-        thread = threading.Thread(target=prefetch_loop, name='prefetch_loop')
-        thread.start()
+        self._thread = threading.Thread(
+            target=prefetch_loop, name='prefetch_loop')
+        self._thread.setDaemon(True)
+        self._thread.start()
 
     def __next__(self):
         batch, prefetch_state = self._comm.get()
@@ -85,6 +88,7 @@ class MultiprocessIterator(iterator.Iterator):
         if not self._finalized:
             self._finalized = True
             self._comm.terminate()
+            self._thread.join(_long_time)
 
     finalize = __del__
 
@@ -182,7 +186,7 @@ class _Communicator(object):
     def get(self):
         with self._lock:
             if len(self._batch_queue) == 0:
-                self._not_empty_cond.wait()
+                self._not_empty_cond.wait(_long_time)
             batch, prefetch_state = self._batch_queue.pop(0)
             self._not_full_cond.notify()
             return batch, prefetch_state
@@ -338,6 +342,7 @@ _fetch_mem_bulk = None
 
 def _fetch_setup(dataset, mem_size, mem_bulk):
     global _fetch_dataset, _fetch_mem_size, _fetch_mem_bulk
+    signal.signal(signal.SIGINT, signal.SIG_IGN)
     _fetch_dataset = dataset
     _fetch_mem_size = mem_size
     _fetch_mem_bulk = mem_bulk
--- a/chainer/iterators/multiprocess_iterator.py
+++ b/chainer/iterators/multiprocess_iterator.py
@@ -1,5 +1,6 @@
 from __future__ import division
 from collections import namedtuple
+from contextlib import contextmanager
 import multiprocessing
 from multiprocessing import sharedctypes
 import signal
@@ -95,10 +96,11 @@ class MultiprocessIterator(iterator.Iterator):
             self._finalized = True
             if '_comm' in d:
                 self._comm.terminate()
-            if '_thread' in d:
-                t = self._thread
-                while t.is_alive():
-                    t.join(_response_time)
+                if self._comm.join_required:
+                    if '_thread' in d:
+                        t = self._thread
+                        while t.is_alive():
+                            t.join(_response_time)
 
     finalize = __del__
 
@@ -191,6 +193,7 @@ class _Communicator(object):
         self._batch_queue = []
         self._status = _Communicator.STATUS_CONTINUE
         self._reset_count = 0
+        self._join_required = True
 
     @property
     def is_terminated(self):
@@ -223,6 +226,12 @@ class _Communicator(object):
             self._not_full_cond.notify()
             self._reset_count += 1
 
+    # called_from_iterator
+    @property
+    def join_required(self):
+        with self._lock:
+            return self._join_required
+
     # called from thread
     def check(self):
         with self._lock:
@@ -241,6 +250,15 @@ class _Communicator(object):
             if reset_count == self._reset_count:
                 self._batch_queue.append((batch, prefetch_state))
                 self._not_empty_cond.notify()
+
+    # called from thread
+    @contextmanager
+    def no_join(self):
+        with self._lock:
+            self._join_required = False
+        yield
+        with self._lock:
+            self._join_required = True
 
 
 class _PrefetchLoop(object):
@@ -262,7 +280,6 @@ class _PrefetchLoop(object):
         self._interruption_testing = _interruption_testing
 
     def __call__(self):
-
         if self.mem_bulk is None:
             if not self._task():
                 return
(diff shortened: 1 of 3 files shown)


okuta:
This PR fixes #2909 .

niboshi:
@grafi-tt Could you add `TestMultiprocessIteratorConcurrency` test I wrote in #2919? It will check for reported deadlock (Issue #2909).

niboshi:
Oh, the last commit of #2919 includes some debug code. So please use the 2nd-latest commit:

https://github.com/niboshi/chainer/blob/aea575ca60830fb23ce8fdbe0228c2d52ffb39d3/tests/chainer_tests/iterators_tests/test_multiprocess_iterator.py#L378

grafi-tt:
@niboshi Added.

grafi-tt:
Sorry, import order was not sorted. Fixed and force-pushed.

okuta:
Could you merge [this PR](https://github.com/grafi-tt/chainer/pull/1/files) or fixing.

okuta:
jenkins, test this please.

niboshi:
>@niboshi Added.

@grafi-tt Thank you!

grafi-tt:
@okuta Thank you for your fixes. Merged.

grafi-tt:
@okuta The failure on AppVeyer is about maximum number of timeout at `therad.join()` on Windows environment. 
Possible fixes are:

1. use smaller number for timeout
2. set https://docs.python.org/3.6/library/threading.html#threading.TIMEOUT_MAX correctly for Python 3
3. do not join

I don't understand why join operation is required here.

grafi-tt:
I found it's complicated than I thought. I'll think about this problem more, and wrote some fixes and tests (possibly).

chainer/iterators/multiprocess_iterator.py
```diff
@@ -10,6 +12,12 @@
 from chainer.dataset import iterator
 
 
+_long_time = 10000000
```
okuta: Sorry, please use `_long_time = 1000000`. Current time out is too long.

okuta:
If we do not use `join`, we cannot detect resource leak problem.
I want to provide stable framework.

chainer/iterators/multiprocess_iterator.py
```diff
@@ -126,167 +146,285 @@ def serialize(self, serializer):
...
+        indices = self._proceed()
+        if indices is None:  # stop iteration
+            batch = None
+        elif self._pool is None:  # measure mode
+            # To enable terminating the thread on Python 2.7,
+            # we spawn another process.
+            # We don't use Pool here, because creating Pools in short-time
+            # had caused rare deadlock on Python 3.5.2.
+            q = multiprocessing.Queue()
+            p = multiprocessing.Process(target=_measure_run,
```
okuta: Why do you use Process?
How about skipping `join` in measurement step.

chainer/iterators/multiprocess_iterator.py
```diff
@@ -126,167 +146,285 @@ def serialize(self, serializer):
...
+                            break
+            finally:
+                p.terminate()
+                p.join()
+
+            self.mem_size = max(map(_measure, batch))
+            self._allocate_shared_memory()
+        else:
+            future = self._pool.map_async(_fetch_run, enumerate(indices))
+            while not future.ready():
```
okuta: Why do you use `ready` ? How about using `feature.wait(_short_time)`?

okuta:
jenkins, test this please.

chainer/iterators/multiprocess_iterator.py
grafi-tt: I was afraid that memory leak occurs when the iterator is GC'ed if indexing dataset is very very slow. But such a program doesn't functions at all and the leak cleared when the process terminates, so now I think avoiding this leak doesn't worth the complexity of launching another process.

I will add another commit that doesn't launch a process at measurement step.

grafi-tt:
Interruption handling is tested now, whereas writing tests that work on Windows was tricky.
With Python 2.7 on Windows, wait for lock can't be interrupted even if long timeout is set. So I changed the implementation to periodically poll whether a termination request is sent. 1 second interval is compromise between performance and responsivity.

grafi-tt:
@okuta @niboshi Now I think all issues are solved. Could you review changes again, and merge to the main tree if there is no problem?

yuyu2172:
Is this checked to work with `MultiprocessPrallellUpdater`?
I tried your code to train a Convnet on ImageNet, and it always hangs when `MultiprocessParallelUpdater ` is used together.
The same code works fine with `StandardUpdater`.

I use Ubuntu 14.04 CUDA8.0, cuDNN 6.0, and basically my script is same as `examples/imagenet/train_imagenet_data_parallel.py`.

grafi-tt:
@okuta @yuyu2172 I confirmed the issue using the imagenet example, and it is fixed by the last commit. Could you test your code again with this commit?

Fundamental cause is that `multiprocessing.Process` and `multiprocessing.Pool` are *not* thread safe if process starting method is "fork", which is default on unix. In this case, background threads will remain on forked processes.

yuyu2172:
@grafi-tt 
Thank you for fixing. I reran the script, and it worked.

chainer/iterators/multiprocess_iterator.py
```diff
@@ -41,46 +51,81 @@ class MultiprocessIterator(iterator.Iterator):
...
-        batch = self._get()
-        self._invoke_prefetch()  # prefetch for the next iteration
-        return batch
+        else:
+            return batch
 
     next = __next__
 
+    def __del__(self):
+        d = self.__dict__
```
okuta: Cloud you write Pythonic code?
```python
if self._finalized or self._comm is None:
  return
self._comm.terminate()
t = self._thread
if t is None:
  return
while t.is_alive():
  t.join(_response_time)
```

grafi-tt:
@okuta Fixed.

Though there is [an error on Travis CI Python 3.5 Linux build](https://travis-ci.org/chainer/chainer/jobs/271901409), I suppose it is not a problem. This error is a numerical mismatch seemingly that is slightly beyond the threshold. It is unrelated to `MultiprocessIterator`.

okuta:
jenkins, test this please.

tests/chainer_tests/iterators_tests/test_multiprocess_iterator.py
```diff
@@ -374,4 +391,162 @@ def test_iterator_serialize_backward_compat(self):
...
+        """
+        code = code_template.format(dataset=dataset,
+                                    n_processes=n_processes,
+                                    n_prefetch=n_prefetch,
+                                    shared_mem=shared_mem,
+                                    operation=operation)
+        fd, self.code_path = tempfile.mkstemp(suffix='.py')
+        os.write(fd, six.b(code))
+        os.close(fd)
+        sys.path.append(os.path.dirname(self.code_path))
```
okuta: Please use absolute path instead this workaround.

tests/chainer_tests/iterators_tests/test_multiprocess_iterator.py
```diff
@@ -374,4 +391,162 @@ def test_iterator_serialize_backward_compat(self):
...
+                while True:
+                    pass
+            except KeyboardInterrupt:
+                pass
+        else:
+            os.kill(self.p.pid, signal.SIGINT)
+
+    def killall(self):
+        # try waiting the root process
+        for _ in range(10):
```
okuta: Why do not use `self.p.wait(10)`?

tests/chainer_tests/iterators_tests/test_multiprocess_iterator.py
```diff
@@ -374,4 +391,162 @@ def test_iterator_serialize_backward_compat(self):
...
+        dataset = numpy.ones((1000, 1000))
+        it = iterators.MultiprocessIterator(dataset, 10, n_processes=4)
+        for _ in range(10):
+            it.next()
+
+        timeout = 5.
+        t = threading.Thread(target=lambda: it.finalize())
+        t.daemon = True
+        t.start()
+        t.join(timeout)
```
okuta: How about `t.join(5)`?

chainer/iterators/multiprocess_iterator.py
```diff
@@ -41,45 +51,93 @@ class MultiprocessIterator(iterator.Iterator):
...
+    next = __next__
 
-        self.is_new_epoch = False
-        if self._finalized is None:
-            self._init()  # start workers
-            # load for the first iteration
-            for _ in six.moves.range(self.n_prefetch):
-                self._invoke_prefetch()
+    def __del__(self):
+        try:
```
okuta: Why do you use `try`?

chainer/iterators/multiprocess_iterator.py
```diff
@@ -41,45 +51,93 @@ class MultiprocessIterator(iterator.Iterator):
...
+            return
+        if finalized:
+            return
+        self._finalized = True
 
-        batch = self._get()
-        self._invoke_prefetch()  # prefetch for the next iteration
-        return batch
+        try:
+            comm = self._comm
```
okuta: How about assign `None` is initialization to avoid `try`.

chainer/iterators/multiprocess_iterator.py
```diff
@@ -41,45 +51,93 @@ class MultiprocessIterator(iterator.Iterator):
...
+        except AttributeError:
+            return
+        comm.terminate()
 
-    next = __next__
+        try:
```
okuta: How about assign `None` is initialization to avoid `try`.

tests/chainer_tests/iterators_tests/test_multiprocess_iterator.py
```diff
@@ -374,4 +391,162 @@ def test_iterator_serialize_backward_compat(self):
...
+        self.p = subprocess.Popen([sys.executable, self.code_path],
+                                  stdout=stdout, stderr=stderr)
+        if stdout is None:
+            self.child_pids = []
+        else:
+            self.child_pids = list(map(int, self.p.stdout.readline().split()))
+
+    def send_sigint(self):
+        if os.name == 'nt':
+            try:
```
okuta: Why do you use `try`?

tests/chainer_tests/iterators_tests/test_multiprocess_iterator.py
grafi-tt: Though this is not my code, I agree with you.

tests/chainer_tests/iterators_tests/test_multiprocess_iterator.py
```diff
@@ -374,4 +391,162 @@ def test_iterator_serialize_backward_compat(self):
...
+
+class TestMultiprocessIteratorInterruption(unittest.TestCase):
+
+    # unless you're debugging tests, this should be false
+    show_interruption_msg = False
+
+    def setUp(self):
+        self.code_path = None
+        self.sys_path_appended = False
+        if not self.__class__.show_interruption_msg:
```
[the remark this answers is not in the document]
grafi-tt: OK, this test class is not likely to be inherited.

tests/chainer_tests/iterators_tests/test_multiprocess_iterator.py
grafi-tt: `os.kill` also sends `CTRL_C_EVENT` to the test ITSELF (when executed at `cmd.exe`; BTW powershell exhibits different behavior), so we should catch this. I'll add comments to clarify this matter.

tests/chainer_tests/iterators_tests/test_multiprocess_iterator.py
grafi-tt: Thanks a lot for you suggestion. When I was writing the code, I was preoccupied with the idea of the `wait(1)` → `poll()` pattern to ensure interruptability on any platform.

chainer/iterators/multiprocess_iterator.py
grafi-tt: OK, I'll change the code to assign `None` for avoiding `try`.
I'll add comments about that attributes may not be initialized at `__del__`, too.

tests/chainer_tests/iterators_tests/test_multiprocess_iterator.py
```diff
@@ -374,4 +391,162 @@ def test_iterator_serialize_backward_compat(self):
...
+                          n_processes=2,
+                          n_prefetch=1,
+                          shared_mem=shared_mem,
+                          operation='it.next()')
+            time.sleep(1.5)
+            self.send_sigint()
+            self.assertFalse(self.killall())
+
+    def test_interrupt_no_wait_batch(self):
+        for shared_mem in (None, 1000000):
```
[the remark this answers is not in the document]
grafi-tt: sure

tests/chainer_tests/iterators_tests/test_multiprocess_iterator.py
grafi-tt: Whether do you mean?

1. instead of dynamically creating a python script using template string, use statically created python script; or
2. use absolute path for the dynamically created script, not the one created by `tempfile.mkstemp`.

If you mean (1), it seems somehow reasonable, but I don't known where I should create the python script.

I don't agree with (2). Temporal files should be on the standard location. But if the chainer already uses some fixed temporal directory, I won't oppose to use it.

tests/chainer_tests/iterators_tests/test_multiprocess_iterator.py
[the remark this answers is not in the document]
grafi-tt: I was misunderstood. Please ignore the comment.

tests/chainer_tests/iterators_tests/test_multiprocess_iterator.py
okuta: I confused you.
I think that it is possible to run a dynamically generated file using an absolute path.
Maybe we do not modify `sys.path`.

okuta:
Thanks a lot!
jenkins, test this please.

grafi-tt:
@okuta Thank you very much for you detailed review. I checked error messages caused by interruption on AppVeyor (Windows). The result was as expected so I disabled this cluttering messages again.

If there are no other issues, I think you can merge it.

okuta:
OK. I am waiting test.

okuta:
jenkins, test this please.

okuta:
LGTM. Thank you very much!

PR #3081, opened 2017-08-01
Add MultithreadIterator

okuta:
This PR provides multi thread iterator.
- Avoid pickle
- There is a limitation of GIL

### DIFF
--- a/chainer/iterators/multithread_iterator.py
+++ b/chainer/iterators/multithread_iterator.py
@@ -48,6 +48,38 @@ class MultithreadIterator(iterator.Iterator):
 
         self.reset()
 
+    def reset(self):
+        if getattr(self, 'current_position', 0) != 0:
+            raise NotImplementedError(
+                'Reset of MultithreadIterator in the middle of a epoch is '
+                'currently not supported.')
+        if getattr(self, 'epoch', 0) != 0 and self._repeat:
+            raise NotImplementedError(
+                'Reset of repeating MultithreadIterator is currently not '
+                'supported.')
+        if (getattr(self, '_finalized', None) is not None and
+                self._finalized.is_set()):
+            raise NotImplementedError(
+                'Reset of finalized MultithreadIterator is currently not '
+                'supported.')
+
+        self.current_position = 0
+        self.epoch = 0
+        self.is_new_epoch = False
+
+        self._previous_epoch_detail = None
+
+        self._pushed_position = None  # initialized at the first iteration
+
+        if self._shuffle:
+            self._order = numpy.random.permutation(len(self.dataset))
+        else:
+            self._order = None
+
+        if self._finalized is not None:
+            self._next = None
+            self._invoke_prefetch()
+
     def __del__(self):
         self.finalize()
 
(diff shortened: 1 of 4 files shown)


chainer/iterators/multithread_iterator.py
```diff
@@ -0,0 +1,217 @@
...
+        self._pushed_position = i
+
+    def _get(self):
+        n = len(self.dataset)
+        i = self.current_position
+
+        batch = []
+        next = self._next
+        while not next.ready():
+            next.wait(0.5)
```
niboshi: How about simply `next.wait()`?

chainer/iterators/multithread_iterator.py
```diff
@@ -0,0 +1,217 @@
...
+                if not self._repeat:
+                    break
+        self._next = None
+
+        self.current_position = i
+        # Eventually overwrite the (possibly shuffled) order.
+        self._order = self._prefetch_order
+        return batch
+
+    def reset(self):
```
niboshi: I think it's better to bring `reset()` to just below `__init__()`, because it is used for initialization.

chainer/iterators/multithread_iterator.py
```diff
@@ -0,0 +1,217 @@
...
+        if getattr(self, 'current_position', 0) != 0:
+            raise NotImplementedError(
+                'Reset of MultithreadIterator in the middle of a epoch is '
+                'currently not supported.')
+        if getattr(self, 'epoch', 0) != 0 and self._repeat:
+            raise NotImplementedError(
+                'Reset of repeating MultithreadIterator is currently not '
+                'supported.')
+        if getattr(self, '_finalized', None) is not None and \
```
niboshi: Use parentheses instead of `\` for line continuation.

chainer/iterators/multithread_iterator.py
```diff
@@ -0,0 +1,217 @@
...
+            raise NotImplementedError(
+                'Reset of finalized MultithreadIterator is currently not '
+                'supported.')
+
+        self.current_position = 0
+        self.epoch = 0
+        self.is_new_epoch = False
+
+        # use -1 instead of None internally.
+        self._previous_epoch_detail = -1.
```
niboshi: Do we have to hold `-1`? How about holding `None` and converting values just before/after serialization?

chainer/iterators/multithread_iterator.py
okuta: This is common technic to avoid Python interruption problem.

okuta:
I fixed.

niboshi:
Could you elaborate the `ready`/`wait` trick or provide some reference?

niboshi:
Do you mean this issue? https://stackoverflow.com/questions/1408356/keyboard-interrupts-with-pythons-multiprocessing-pool
OK, I got it.

okuta:
Yes

chainer/iterators/multithread_iterator.py
```diff
@@ -0,0 +1,214 @@
...
+        self._prefetch_order = None  # used at the end of each epoch
+
+        self.n_threads = n_threads
+
+        self._finalized = None
+
+        self.reset()
+
+    def reset(self):
+        if getattr(self, 'current_position', 0) != 0:
```
niboshi: If these attributes are initialized in the constructor, `getattr` would not be needed.

niboshi:
LGTM!

PR #3085, opened 2017-08-01
Simplify MultiprocessIterator

okuta:
This PR simplify MultiprocessIterator by using `multiprocessing.pool.Pool`.

### DIFF
--- a/chainer/iterators/multiprocess_iterator.py
+++ b/chainer/iterators/multiprocess_iterator.py
@@ -1,7 +1,7 @@
 from __future__ import division
 import multiprocessing
+from multiprocessing import pool
 from multiprocessing import sharedctypes
-import threading
 import warnings
 
 import numpy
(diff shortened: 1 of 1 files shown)


niboshi:
@okuta This conflicts with #3076. Can you check?

okuta:
This PR fixes #2909 . But we use #3076 .

grafi-tt:
Sorry, it seems I was too late to write my comments.

If I understand correctly, this PR just replaces of hand-written worker management by pool. But my PR is almost rewrites original implementation. So I don't think merging them makes sense. Notable differences are:

* Instated of polling-based queue management, my implementation uses conditional variables.
* In my implementation the prefetch thread runs independently from `__next__` call in the main thread. It may complicate reset implementation, but I think I managed to achieve a simple mechanism using a counter.

I honestly like my own implementation, but this PR can be reasonable conservative choice. If you find any flaws in my codes, I'm happy to fix those.

okuta:
Thank you for your comment.

I am reviewing your PR. Your PR has good features.

PR #3093, opened 2017-08-04
Add warnings about preprocessing for dataset with both grayscale and RGB images.

jinjiren:
This pull request adds a warning about preprocessing the images to the documentation of ImageDataset and LabeledImageDataset.

The problem I had is that there are **a few grayscale images** in the dataset I want to use, while I am not aware of that. Chainer does not automatically convert the images to RGB or provide an option for that. Since previously I have used this dataset with other libraries and it works just fine, I was just confused about the errors `ValueError: all the input array dimensions except for the concatenation axis must match exactly`. 

I understand the decision you made as in #1445, however, I think it will be helpful if such a warning exists in the doc, so that people like me will not spend several hours to detect such an error.

Especially for people coming from other frameworks, such friendly notes are really necessary.

### DIFF
--- a/chainer/datasets/image_dataset.py
+++ b/chainer/datasets/image_dataset.py
@@ -44,6 +44,13 @@ class ImageDataset(dataset_mixin.DatasetMixin):
        formats you want to use (e.g. libpng for PNG images, and libjpeg for JPG
        images).
 
+    .. note::
+       **You are responsible for preprocessing the images before feeding them to
+       a model.** For example, if your dataset contains both RGB and grayscale
+       images, make sure that you convert them to the same format. Otherwise you
+       will get errors because the input dimensions are different for RGB and
+       grayscale images.
+
     Args:
         paths (str or list of strs): If it is a string, it is a path to a text
             file that contains paths to images in distinct lines. If it is a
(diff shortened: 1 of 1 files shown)


niboshi:
Thank you for the pull request!
I think these notes are worth emphasized, considering the trouble you experienced. So how about using `.. warning::` instead of `.. note::`?

jinjiren:
@niboshi Sure

niboshi:
jenkins, test this please

niboshi:
LGTM!

jinjiren:
You are welcome! 

Actually through recent days learning and using chainer in my project, I found that it is indeed a nice tool, however, the problem is that it is hard to find related answers to even some simple but common problems.

I know this is a problem about the community size, and I also appreciate your effort to continue improving this framework. Nevertheless, it would be great if you can put more attention to the doc and providing helpful examples for the existing modules you have. 

For example, I found that Trainer is really convenient after spending time on reading documentations and source code, but at the beginning it just seems too confusing to me.

I think more user-friendly documentation, examples and related resources will make chainer more popular, and hence results in a better and huger community and finally it will form a virtuous cycle.

PR #3095, opened 2017-08-04
[backport] Add warnings about preprocessing for dataset with both grayscale and RGB images.

niboshi:
Backport of #3093


okuta:
LGTM!

PR #3099, opened 2017-08-04
New style linear function

beam2d:
This PR fixes `F.linear` to use the new-style function node interface of #2970.

It depends on #3096 (`check_double_backward`).

### DIFF
--- a/chainer/functions/connection/linear.py
+++ b/chainer/functions/connection/linear.py
@@ -1,5 +1,5 @@
 from chainer import function_node
-from chainer import functions as F
+import chainer.functions
 from chainer.utils import type_check
 
 
@@ -48,12 +48,12 @@ class LinearFunction(function_node.FunctionNode):
         ret = []
         if 0 in indexes:
             gx = linear(gy, W.T)
-            ret.append(F.cast(gx, x.dtype))
+            ret.append(chainer.functions.cast(gx, x.dtype))
         if 1 in indexes:
             gW = linear(gy.T, x.T)
-            ret.append(F.cast(gW, W.dtype))
+            ret.append(chainer.functions.cast(gW, W.dtype))
         if 2 in indexes:
-            gb = F.sum(gy, axis=0)
+            gb = chainer.functions.sum(gy, axis=0)
             ret.append(gb)
 
         return ret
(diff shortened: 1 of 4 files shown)


chainer/functions/connection/linear.py
```diff
@@ -1,14 +1,10 @@
-from chainer import function
+from chainer import function_node
+from chainer.functions.array import cast
+from chainer.functions.math import sum
```
jekbradbury: shadowing a builtin is probably a little risky?

chainer/functions/connection/linear.py
```diff
@@ -43,20 +39,25 @@ def forward(self, inputs):
...
-        gx = gy.dot(W).astype(x.dtype, copy=False).reshape(inputs[0].shape)
-        gW = gy.T.dot(x).astype(W.dtype, copy=False)
-        if len(inputs) == 3:
-            gb = gy.sum(0)
-            return gx, gW, gb
-        else:
-            return gx, gW
+        ret = []
+        if 0 in indexes:
+            gx, = LinearFunction().apply((gy, W.T))
```
jekbradbury: Will this be the standard way to use Chainer `Function`s in the backward of new-style functions? It is also possible to use `F.linear`, right? I believe that's what PyTorch does, in order to make writing the backward pass of new `Functions` more similar to writing new Links.

beam2d: Yes, you can use `F.linear` freely in `backward`. In the above code, I used `LinearFunction().apply()` just for slightly better efficiency (but it might be better to use the plain `F.linear`).

okuta:
Please resolve conflict.

beam2d:
Rebased to the current master.
I also sent another PR to reduce the overhead of `F.cast`: #3191.

beam2d:
Some more changes.
- Start using `functions` directly instead of importing individual modules (core team decided to use this style from this version, not yet explained in the contribution guide though). It removes the confusing hiding of `sum` built-in.
- Use `linear` directly in backward for educational purposes.

okuta:
Please check travis result.

okuta:
LGTM except error.

beam2d:
Fixed. Now it passes the tests except for numerical errors.

okuta:
LGTM!

PR #3100, opened 2017-08-05
Check the ratio of dropout

keisuke-umezawa:
add the assert in ``chainer.functions.Dropout.__init__`` for dropout_ratio.

### DIFF
--- a/chainer/functions/noise/dropout.py
+++ b/chainer/functions/noise/dropout.py
@@ -12,7 +12,7 @@ class Dropout(function.Function):
     """Dropout regularization."""
 
     def __init__(self, dropout_ratio):
-        assert 0.0 <= dropout_ratio <= 1.0
+        assert 0.0 <= dropout_ratio and dropout_ratio <= 1.0
         self.dropout_ratio = dropout_ratio
 
     def check_type_forward(self, in_types):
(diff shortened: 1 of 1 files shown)


chainer/functions/noise/dropout.py
```diff
@@ -12,6 +12,7 @@ class Dropout(function.Function):
     """Dropout regularization."""
 
     def __init__(self, dropout_ratio):
+        assert 0.0 <= dropout_ratio and dropout_ratio <= 1.0
```
niboshi: - It's better to raise `ValueError`, because this is to check for external specification, not internal implementation.
- You can write `0.0 <= dropout_ratio <= 1.0`.

chainer/functions/noise/dropout.py
```diff
@@ -62,7 +63,8 @@ def dropout(x, ratio=.5, **kwargs):
 
     Args:
         x (~chainer.Variable): Input variable.
-        ratio (float): Dropout ratio.
+        ratio (float): Dropout ratio. The ``ratio`` must be
+        ``0.0 <= ratia <= 1.0``, otherwise assertion is failed.
```
niboshi: - Typo: ratia -> ratio
- We don't have to mention assertion.

keisuke-umezawa:
Thank you for the review.
I used the exception and fixed the comment.

I also noticed the function will raise the ZeroDivisionError when the ``ratio`` is 1.0.
So, I tried checking the ``ratio`` is in the range [0, 1).

niboshi:
jenkins, test this please.

niboshi:
LGTM! Thank you!

PR #3101, opened 2017-08-05
Add count_params method to Link

mitmul:
This PR adds `size` property to `Link` class. It enables to expose the size of parameters in a `Link`, `Chain`, `ChainList` and all objects inherited from `Link`. If a `Link` contains uninitialized parameters, a warning message will show up to notify users that the size of them will not be counted in the resulting size.


chainer/link.py
```diff
@@ -558,6 +558,28 @@ def serialize(self, serializer):
         for name in self._persistent:
             d[name] = serializer(name, d[name])
 
+    @property
+    def size(self):
+        """Count the size of parameters.
```
delta2323: I think the "number" of parameter is more appropriate.

delta2323:
It seems that this property counts the number of parameters directly attached to the link of interest only and does not includes parameters of its children. It depends on use cases which specification is suitable. Could you share me what use case do you have in mind?

chainer/link.py
mitmul: I thought "number of parameters" could mean the number of "Parameter" objects in the link. So, I used "size of parameters". How do you think about it?

mitmul:
@delta2323 Thanks for the comment. `Link`'s `namedparams()` surely returns the parameters which was directly attached, but `Chain`'s and `ChainList`'s `namedparams()` will traverse all children, so if `size` is called from `Chain` or `ChainList`, it returns the count of all parameters in the model. For example, If I call this `size` property from the MLP model defined in the official MNIST example, I got:

```
In [1]: from train_mnist import MLP
model = MLP(
In [2]: model = MLP(100, 10)

In [3]: model.size
/Users/shunta/lib/chainer/chainer/link.py:578: UserWarning: /l1/W has not been initialized, so the resulting size will not include the number of parameters in it.
  'not include the number of parameters in it.'.format(name))
/Users/shunta/lib/chainer/chainer/link.py:578: UserWarning: /l2/W has not been initialized, so the resulting size will not include the number of parameters in it.
  'not include the number of parameters in it.'.format(name))
/Users/shunta/lib/chainer/chainer/link.py:578: UserWarning: /l3/W has not been initialized, so the resulting size will not include the number of parameters in it.
  'not include the number of parameters in it.'.format(name))
Out[3]: 210
```

initially because some `Link`s has not been initialized (the above warnings are intended), but after the initial forward pass, I got:

```
In [6]: model(np.zeros((1, 10), dtype=np.float32))
Out[6]: variable([[ 0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.]])

In [7]: model.size
Out[7]: 12210
```

so it works correctly and I think it's useful to know the "number of parameters" for ensuring the correctness of implementation or comparing the model's heaviness with others.

delta2323:
@mitmul Sorry, I wrongly remembered the specification of `namedparams`. I agree with this specification as users typically want to know the number of parameters in a whole architecture.

delta2323:
Could you add unittests that check `Link.param` works even for `Chain` and `ChainList`. Also, could you add unittests that the warning is not issued when parameters are initialized via first forward propagation.

chainer/link.py
delta2323: Then, how about "the total size of parameters"?

niboshi:
I think `size` is too general for this feature. It originates from `ndarray.size`, which represents the size of a tensor. It's straight-forward that the size of a tensor represents the number of elements. `Variable` also has `size` attribute. I think it's ok because a `Variable` is inherently a tensor.

On the other hand, a `Link` can be thought of as a collection of `Parameter`s. In this perspective, `size` should represent the numer of registered `Parameter`s. So I think we need some other name...

Do you come up with any?

mitmul:
@delta2323 OK, I'll add tests for `Chain` and `ChainList`.

@niboshi I understand what you pointed out. How about `count` or `count_params` ?

chainer/link.py
mitmul: Yes, that's better. I'll modify it.

niboshi:
@mitmul 

>@niboshi I understand what you pointed out. How about count or count_params ?
 
Keras uses `count_params`, so it might be ok, but the term "parameter" has the different meanings between Chainer and Keras. It might be still confusing in Chainer. I would prefer more descriptive name such as `count_param_scalars`.

By the way, I would doubt the need of this feature.

You can write `sum(param.size or 0 for param in link.params)` if we change a behavior of `Parameter.size` a bit (to return `None` if it's uninitialized, which is a reasonable change IMO). It's easy and straight-forward for users to write such code.

I think it's better to hear other members' opinions as well.

mitmul:
Thank you for your comment.

Hmm, `count_param_scalars` is too long, so I think `count_params()` or `count` is enough. Well, as for the necessity, this is not intended to fill a lack of functionality of Chainer, so it's surely not essential, but just to improve the usability. I've heard that some users wonder how to know the number of trainable parameters in a model in Chainer sometimes because other frameworks provide the way to know it easily. And in the way by `sum(...)`, it's difficult to know which parameters are not counted in the returned value, but `Link.count` raises warnings for them. So providing `count` or `count_params()` for `Link` is reasonable for this reason I think.

delta2323:
IMO, if it is a routine work to count the number of parameters in a link, it would not be strange that Chainer provides it as a utility function.

delta2323:
It seems to me that `count`, `count_params` could also remind some users of the number of instances of `Parameter` attached to a link in the same way `size` does. So far, I could not come up with some word that is simple enough but definitely points out either of the meanings.

delta2323:
@mitmul @niboshi Do you have some idea how to proceed this issue?

niboshi:
I prefer `count_params` among the names mentioned above. How do you think?

delta2323:
I think it is a good idea to align the semantics of property names if possible to other frameworks. In that sense, `count_params` is OK (It seems PyTorch and Caffe do not have this functionality. We list up parameters inside a model and sum them up explicitly).

delta2323:
@mitmul @niboshi How do you think of it?

mitmul:
I'll change the name to `count_params`

mitmul:
@delta2323 @niboshi I changed the name and added tests for `Chain` and `ChainList`.

chainer/link.py
```diff
@@ -578,6 +578,26 @@ def serialize(self, serializer):
         for name in self._persistent:
             d[name] = serializer(name, d[name])
 
+    def count_params(self):
+        """Count the total size of parameters.
+
+        If the link containts uninitialized parameters, this raises warnings.
```
niboshi: I think we should emphasize that this method counts the number of scalars.
How about something like this?
(Just an example. Please fix as you like)

*Counts the total number of parameters.*

*This method counts the total number of scalar values included in all the `Parameter`s held by this link and its descendants.*

niboshi: this -> this method
warnings -> a warning

chainer/link.py
```diff
@@ -578,6 +578,26 @@ def serialize(self, serializer):
...
+        Returns:
+            The total size of parameters (int)
+
+        """
+
+        size = 0
+        for name, param in self.namedparams():
+            if param.array is None:
+                warnings.warn(
+                    '{} has not been initialized, so the resulting size will '
```
niboshi: `{}` -> `Parameter '{}'` ?
size -> count?

chainer/link.py
mitmul: That's right. Good suggestion

tests/chainer_tests/test_link.py
```diff
@@ -789,6 +801,16 @@ def test_serialize(self):
         mocks['l1'].assert_called_with('x', self.l1.x.data)
         mocks['l2'].assert_called_with('x', self.l2.x.data)
 
+    @unittest.skipUnless(
+        six.PY3, 'Python2.7 has a bug in catch_warnings, so this test is '
+                 'skipped for Python2.7')
```
niboshi: What kind of bug? How about using `testing.assert_warns`?

chainer/link.py
mitmul: Thanks. fixed.

chainer/link.py
mitmul: Thanks. fixed.

tests/chainer_tests/test_link.py
mitmul: I forgot why I used this. I will try assert_warns

mitmul: Hmm, in Python 2.7, catch_warnings can't count the number of warnings correctly. So the `self.assertEqual(len(w), 2)` fails.

niboshi: Ah, I see. Is this check needed? Warning is printed only once anyway.

mitmul: I think we can remove the warning count checking. I removed them.

mitmul:
I think it finished to fix them to follow reviews.

chainer/link.py
```diff
@@ -578,6 +578,31 @@ def serialize(self, serializer):
         for name in self._persistent:
             d[name] = serializer(name, d[name])
 
+    def count_params(self):
+        """Counts the total number of parameters.
+
+        This method counts the total number of scalar values included in all
+        the :class:`~chainer.Parameters` held by this link and its descendants.
```
niboshi: I think this should be:
```
:class:`~chainer.Parameter`\\ s
```

niboshi: I think this should be:
```
:class:`~chainer.Parameter`\\ s
```

niboshi:
@delta2323 LGTM to me except the last comment.

Could you merge if LGTM to you?

niboshi:
@mitmul Please use `testing.assert_warns`.

tests/chainer_tests/test_link.py
```diff
@@ -468,6 +470,12 @@ def test_update_enabled(self):
         self.link.enable_update()
         self.assertTrue(self.link.update_enabled)
 
+    def test_count_params(self):
+        self.assertEqual(self.link.count_params(), 8)
+        with warnings.catch_warnings(record=True) as w:
```
delta2323: Check the number of warnings issued is correct (maybe 1) and the type of the warning is `UserWarning` with `w`. Otherwise, remove `as w` as it is not used.

tests/chainer_tests/test_link.py
```diff
@@ -789,6 +797,12 @@ def test_serialize(self):
         mocks['l1'].assert_called_with('x', self.l1.x.data)
         mocks['l2'].assert_called_with('x', self.l2.x.data)
 
+    def test_count_params(self):
+        self.assertEqual(self.c1.count_params(), 8)
+        with warnings.catch_warnings(record=True) as w:
```
delta2323: ditto

tests/chainer_tests/test_link.py
```diff
@@ -1102,6 +1116,12 @@ def test_serialize(self):
         mocks['0'].assert_called_with('y', l1.y.data)
         mocks['1'].assert_called_with('x', l2.x.data)
 
+    def test_count_params(self):
+        self.assertEqual(self.c1.count_params(), 8)
+        with warnings.catch_warnings(record=True) as w:
```
delta2323: ditto

delta2323:
Strictly speaking, this PR does not check `count_params` does not throw warning when all parameters are initialized. So could you add such test cases?

mitmul:
@delta2323 @niboshi `assert_warns` seems not to return any values when it calls `yield`, so we can't receive any information about the number of warning raised. https://github.com/chainer/chainer/blob/master/chainer/testing/helper.py#L53
I think we need to use `catch_warnings` directly OR stop checking the number of warnings. What should I do?

niboshi:
IMO, you can just use `testing.assert_warns`, and ignore the number of warnings.

mitmul:
Hmm, but I succeeded to count the number of warnings. When two warnings raised, `len(w)` returns 2 if I used `catch_warnings`. But I guess this doesn't work in Python2.

mitmul:
@delta2323 Well, I added the test cases for this:
> Strictly speaking, this PR does not check count_params does not throw warning when all parameters are initialized. So could you add such test cases?

niboshi:
I understand it's possible to count the warnings in Python 3, by using `catch_warnings`.
What I wonder is if it matters.
(No matter how many warnings is raised, the observable outcome to the users is the same.)

tests/chainer_tests/test_link.py
```diff
@@ -471,10 +471,19 @@ def test_update_enabled(self):
...
+        assert len(w) == 2
+        assert w[0].category is UserWarning
+        assert self.link.count_params() == 8
+
+        self.link.u.initialize((2, 3))
+        self.link.v.initialize((2, 3))
+        with warnings.catch_warnings(record=True) as w:
+            warnings.simplefilter('always')
+            self.link.count_params()
+            assert len(w) == 0
```
delta2323: You can check the condition simply with `assert not w` because w is a (empty) list 
(cf. https://docs.python.org/3.6/library/warnings.html#warnings.catch_warnings)

tests/chainer_tests/test_link.py
```diff
@@ -798,11 +807,18 @@ def test_serialize(self):
...
             self.c2.count_params()
+        assert len(w) == 1
+        assert w[0].category is UserWarning
+        assert self.c1.count_params() == 8
 
+        self.c2.l3.x.initialize((3,))
+        with warnings.catch_warnings(record=True) as w:
+            warnings.simplefilter('always')
+            self.c2.count_params()
+        assert len(w) == 0
```
delta2323: ditto

tests/chainer_tests/test_link.py
```diff
@@ -1117,10 +1133,18 @@ def test_serialize(self):
...
             self.c2.count_params()
+        assert len(w) == 1
+        assert w[0].category is UserWarning
+        assert self.c1.count_params() == 8
+
+        self.c2[0][0].y.initialize((2, 3))
+        with warnings.catch_warnings(record=True) as w:
+            warnings.simplefilter('always')
+            self.c2.count_params()
+        assert len(w) == 0
```
delta2323: ditto

tests/chainer_tests/test_link.py
delta2323: But current code is also OK as it is easier to understand.

mitmul:
@delta2323 Thanks for the comment. I fixed the points.

delta2323:
Jenkins, test this please.

delta2323:
@niboshi LGTM. Could you merge this PR if you approve it?

niboshi:
LGTM!

PR #3114, opened 2017-08-07
Fix errors in docstrings

mitmul:
doctest fails for them


mitmul:
@niboshi Thanks for the review! I applied the changes you suggested!

niboshi:
LGTM!

PR #3119, opened 2017-08-07
implement new style function for sigmoid

keisuke-umezawa:
This PR fixes F.sigmoid to use the new-style function node interface of #2970.

I am trying to implement new style function. It seems to pass the unit test.
But there are some problem about this PR.
- Since the #3096 is not merged, the double-grad tests are not written.
- I do not know the right way to dispatch from ``backward`` to ``backward_cpu`` or ``backward_gpu``.


jekbradbury:
Since the transition to new-style functions will involve replacing optimized `backward` implementations that directly call CUDA elementwise kernels with slower combinations of abstracted forward operations, and these forward operations are all fully generic for CPU and GPU, maybe new functions should have one mandatory implementation in terms of forward operations (`backward`) and two optional ones for better performance (`backward_cpu` and `backward_gpu`, same as old-style functions) in the case that the user does not require double differentiation support?

keisuke-umezawa:
@jekbradbury Thank you for the kind reply! Now, I know how to implement the double differentiation . So, I must implement the ``backward`` by using the combinations of forward operations so that the ``grad`` is also generated forward operations.

Then, what should I do is below?
- leave the old functions ``backward_cpu`` and `` backward_gpu``.
- implement the generic ``backward`` by using the forward operations.

keisuke-umezawa:
But what code should dispatch to ``backward``, ``backward_cpu`` or ``backward_gpu``.

jekbradbury:
I just mean that if the backwards execution engine knows whether the user requires double differentiation support, it could use that information to dispatch to either backward, backward_cpu, or backward_gpu. That way the work Chainer developers have done optimizing CUDA implementations for backward_gpu wouldn’t be wasted in the very common case that double differentiation isn’t needed.

beam2d:
I also feel we should keep the current optimized backward implementation. Otherwise, the performance will be deteriorated. I will create an issue for this problem.

keisuke-umezawa:
link to #3147.

keisuke-umezawa:
@beam2d review this please.

chainer/functions/activation/sigmoid.py
```diff
@@ -12,51 +12,88 @@
...
         type_check.expect(in_types[0].dtype.kind == 'f')
 
-    def forward_cpu(self, x):
-        half = x[0].dtype.type(0.5)
-        y = utils.force_array(numpy.tanh(x[0] * half) * half + half)
+    def forward_cpu(self, inputs):
+        x = inputs[0]
+        half = x.dtype.type(0.5)
+        y = utils.force_array(numpy.tanh(x * half) * half + half)
+        self.retain_inputs(())
```
beam2d: This `retain_inputs` is not needed anymore (inputs are not retained by default in FunctionNode).

chainer/functions/activation/sigmoid.py
```diff
@@ -12,51 +12,88 @@
...
         x = inputs[0]
         if chainer.should_use_cudnn('==always') and x.flags.c_contiguous:
             y = cudnn.activation_forward(x, _mode)
+            self.retain_inputs((0,))
+            self._use_cudnn = True
         else:
             y = cuda.elementwise(
                 'T x', 'T y', 'y = tanh(x * 0.5) * 0.5 + 0.5',
                 'sigmoid_fwd')(x)
             self.retain_inputs(())
```
beam2d: ditto

chainer/functions/activation/sigmoid.py
```diff
@@ -12,51 +12,88 @@
...
             gx = cuda.elementwise(
                 'T y, T gy', 'T gx',
                 'gx = gy * y * (1 - y)',
                 'sigmoid_bwd')(y, gy)
         return gx,
 
+    def backward(self, indexes, grad_outputs):
+        y, gy = self.get_retained_inputs()
+        g, = grad_outputs
+        one = y.dtype.type(1)
```
beam2d: These `one` and `two` are not needed in `backward` because scalar-Variable arithmetics resolve the type promotion correctly.

chainer/functions/activation/sigmoid.py
keisuke-umezawa: delete it

chainer/functions/activation/sigmoid.py
keisuke-umezawa: delete it

chainer/functions/activation/sigmoid.py
keisuke-umezawa: just use 1 and 2.

keisuke-umezawa:
@beam2d I fixed them as you said. Please check them.

beam2d:
Jenkins, test this please.

beam2d:
LGTM!

beam2d:
Thank you!

PR #3122, opened 2017-08-08
Remove get_device from examples

naoto0804:
This PR removes `chainer.cuda.get_device` from `./examples/`, which is deprecated in Chainer v2.


niboshi:
jenkins, test this please

niboshi:
LGTM! Thank you very much!

PR #3129, opened 2017-08-09
Deny assigning links in ChainList.init_scope()

### LINKED ISSUE #3124: ChainList doesn't warn self.init_scope()
The following code has a bug.  Chainer doesn't warn the usage of `with self.init_scope()` in the subclass of ChainList.  Could you add a warning message for such mistakes?

* Code to reproduce
```python
import chainer
from chainer import Chain, ChainList
import chainer.functions as F
import chainer.links as L
import numpy as np


class C(ChainList):  # should be 'class C(Chain)'
    def __init__(self):
        super().__init__()
        with self.init_scope():
            self.l1 = L.Linear(5, 5)
            self.l2 = L.Linear(5, 5)

    def __call__(self, x):
        return self.l2(F.relu(self.l1(x)))


c = C()
print(c.l1.W.data)

opt = chainer.optimizers.SGD()
opt.setup(c)
x = np.arange(10).reshape((2, 5)).astype(np.float32)
loss = F.sum(c(x))
opt.update(lambda: loss)
print(c.l1.W.data)
```

output:
```
[[ 0.22224635  0.13709065 -0.0590423   0.31734523  0.76646286]
 [-0.09569775 -0.00810872  0.72896075 -0.50144166 -0.23909038]
 [ 0.24655567 -0.59849507  0.05945947 -0.06633393 -0.05738653]
 [-0.85418522  0.56111503 -0.4280332  -0.19436245 -0.09941436]
 [-0.06522682 -0.43800679  0.7132498   0.49363273 -0.2827867 ]]
[[ 0.22224635  0.13709065 -0.0590423   0.31734523  0.76646286]
 [-0.09569775 -0.00810872  0.72896075 -0.50144166 -0.23909038]
 [ 0.24655567 -0.59849507  0.05945947 -0.06633393 -0.05738653]
 [-0.85418522  0.56111503 -0.4280332  -0.19436245 -0.09941436]
 [-0.06522682 -0.43800679  0.7132498   0.49363273 -0.2827867 ]]
```
No update is taken because the link l1 is not registered.

toslunar:
Resolves #3124

### DIFF
--- a/chainer/link.py
+++ b/chainer/link.py
@@ -861,6 +861,15 @@ class ChainList(Link):
         link.name = str(len(self._children))
         self._children.append(link)
 
+    @contextlib.contextmanager
+    def init_scope(self):
+        with super().init_scope():
+            warnings.warn('''\
+ChainList does not support child link registration by assigning a link within \
+a "with link.init_scope():" block.
+''', UserWarning)
+            yield
+
     def copy(self):
         ret = super(ChainList, self).copy()
         ret._children = list(ret._children)  # copy
--- a/chainer/link.py
+++ b/chainer/link.py
@@ -820,6 +820,13 @@ class ChainList(Link):
         for link in links:
             self.add_link(link)
 
+    def __setattr__(self, name, value):
+        if self.within_init_scope and isinstance(value, Link):
+            raise TypeError(
+                'cannot register a new link'
+                ' within a "with chainlist.init_scope():" block.')
+        super(ChainList, self).__setattr__(name, value)
+
     def __getitem__(self, index):
         """Returns the child at given index.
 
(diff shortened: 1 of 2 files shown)


okuta:
Why do you use warning?
How about using exception?

beam2d:
We have to take care of a case that user registers a parameter to ChainList directly. In such a case, `ChainList.init_scope` still works (because it is equivalent to `Link.init_scope`).

I think it's better to raise a warning (or exception) when one tries to assign a link within `init_scope`, instead of when the user tries to use `init_scope`. It can be done by overriding `__setattr__`.

toslunar:
>I think it's better to raise a warning (or exception) when one tries to assign a link within `init_scope`, instead of when the user tries to use `init_scope`. It can be done by overriding `__setattr__`.

Done.

I changed the warning to a TypeError.  Is it the correct choice among exceptions?

beam2d:
Thanks. Design looks good to me.

okuta:
Thank you.
Please add test and change this PR title.

toslunar:
Done.

okuta:
jenkins, test this please.

okuta:
LGTM!

PR #3130, opened 2017-08-09
Add the arctan2 function

### LINKED ISSUE #3128: Feature request: arctan2 (or atan2)
Chainer has the `arctan` function but doesn't have the arctangent function with two arguments (https://en.wikipedia.org/wiki/Atan2). Tensorflow and PyTorch have it.

- https://www.tensorflow.org/api_docs/python/tf/atan2
- http://pytorch.org/docs/master/torch.html?highlight=atan2#torch.atan2

muupan:
Resolves #3128


gwtnb:
LGTM

PR #3152, opened 2017-08-15
Add alias chainer.FunctionHook

delta2323:
This PR adds `chainer.FunctionHook`, which is an alias of `chainer.function_hook.FunctionHook`. This PR also fixes code so that it does not use `chainer.function.FunctionHook`, which exists for backward compatibility.


okuta:
LGTM!

PR #3153, opened 2017-08-15
Add function_hooks to __init__.py

delta2323:
This PR adds `chainer.function_hooks` to `chainer.__init__.py` to align the convention that modules that concrete objects (like inheritance of `Link`) are imported. As this PR is made on top of #3152 to avoid conflicts, I would like this PR to be merged after the PR.


okuta:
LGTM!

PR #3170, opened 2017-08-17
Avoid running certain hooks on uninitialized params

### LINKED ISSUE #3041: Optimizer tries to update uninitialized parameters

tkerola:
Fixes #3041 
Replaces #3056

### DIFF
--- a/chainer/optimizer.py
+++ b/chainer/optimizer.py
@@ -185,7 +185,8 @@ class UpdateRule(object):
             return
 
         self.t += 1
-        self._prepare(param)
+        if param.data is not None:
+            self._prepare(param)
         for hook in six.itervalues(self._hooks):
             hook(self, param)
         self.update_core(param)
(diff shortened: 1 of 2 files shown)


niboshi:
jenkins, test this please

niboshi:
jenkins, test this please

niboshi:
jenkins, test this please

niboshi:
LGTM!

PR #3178, opened 2017-08-18
Fix doc of clipped_relu

niboshi:
Documentation of `clipped_relu` is broken.
https://docs.chainer.org/en/latest/reference/generated/chainer.functions.clipped_relu.html


okuta:
LGTM!

PR #3179, opened 2017-08-18
Fix doc of leaky_relu

niboshi:
The definition of `leaky_relu` in the current documentation is only valid if `a <= 1`.

https://docs.chainer.org/en/latest/reference/generated/chainer.functions.leaky_relu.html

### DIFF
--- a/chainer/functions/activation/leaky_relu.py
+++ b/chainer/functions/activation/leaky_relu.py
@@ -59,7 +59,12 @@ def leaky_relu(x, slope=0.2):
 
     This function is expressed as
 
-    .. math:: f(x)=\\max(x, ax),
+     .. math::
+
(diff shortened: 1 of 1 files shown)


chainer/functions/activation/leaky_relu.py
```diff
@@ -59,7 +59,12 @@ def leaky_relu(x, slope=0.2):
 
     This function is expressed as
 
-    .. math:: f(x)=\\max(x, ax),
+     .. math::
```
okuta: Why do you use 5 spaces?

niboshi: Sorry, fixed.

okuta:
OK

PR #3180, opened 2017-08-18
Fix Variable.__getitem__ document

niboshi:
Currently "Returns" section of `Variable.__getitem__` documentation is broken.

[`https://docs.chainer.org/en/stable/reference/core/generated/chainer.Variable.html#chainer.Variable.__getitem__`](https://docs.chainer.org/en/stable/reference/core/generated/chainer.Variable.html#chainer.Variable.__getitem__)


delta2323:
LGTM

Issue #3181, opened 2017-08-18
how to use chainer.using_config to stop F.dropout in evaluate/predict process in chainer? 


sharpstill:
F.dropout is only used in train, I confused how to use chainer.using_config in it? How does it works and chainer how to know it is in training or predict?
I already put on
https://stackoverflow.com/questions/45757330/how-to-use-chainer-using-config-to-stop-f-dropout-in-evaluate-predict-process-in
AND  in stackoverflow why answer to little and respond slow?

beam2d:
Sorry for the late response, but we cannot promise to make response during weekends. For the current question, see the answer made in the SO thread.

PR #3184, opened 2017-08-19
fixed document of F.dropout into more explanatory

fiarabbit:
#3181 In the original "chainer.using_config('train', train)", the word "train" is duplicated. So I converted the latter one to boolean.


delta2323:

PR #3187, opened 2017-08-20
Use Trainer in examples/reinforcement_learning/ddpg_pendulum.py

keisuke-umezawa:
Previously, in examples/reinforcement_learning/ddpg_pendulum.py, `Trainer` is not used and
training the model with custom loop.

So, I changed to use Trainer in ddpg_pendulum.py.
- Define Updater
- Define GymIterator
- Use Trainer to learn model parameter.

And I made the file `ddpg_pendulum_custom_loop.py` for the previous implementation with custom loop.

### DIFF
--- a/examples/reinforcement_learning/ddpg_pendulum.py
+++ b/examples/reinforcement_learning/ddpg_pendulum.py
@@ -16,6 +16,7 @@ import numpy as np
 import chainer
 from chainer import functions as F
 from chainer import links as L
+from chainer import training
 from chainer import optimizers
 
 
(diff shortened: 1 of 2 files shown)


ir5:
While it is interesting to see that we can use Trainer class for reinforcement learning, what is the primal benefit for using the Trainer class? As far as I could see, the original version is simpler and easier to understand.

keisuke-umezawa:
@ir5 
Thank you for the review. What I want to do is writing the tutorial for reinforcement learning using this example. As far as I know, I think that we should use the ``Trainer`` class for training the model as the official use case. So, I rewrited the example with ``Trainer`` like the  [examples/ptb/train_ptb.py](https://github.com/chainer/chainer/blob/master/examples/ptb/train_ptb.py) and will try to write the tutorial for the example.

However, as you said, I also recognized that the example with ``Trainer`` is not so good.
Also, I do not really know whether we should use ``Trainer`` for the official example.

By the way, I want to write the tutorial using this example. Is it okay using the example code without ``Trainer``?

muupan:
> As far as I know, I think that we should use the Trainer class for training the model as the official use case.

I don't think such a rule exists. It should be completely ok to write an example or tutorial without Trainer.

ir5:
@muupan: Would you review this PR? I once put myself as an assignee of this PR, but if you are interested in, I would like you to review this one.

At least in my opinion, we shouldn't make the Trainer version as a "main" example because it may give the wrong impression like chainer requires complicated knowledge. I think it is good to put the Trainer version example with a file name like "ddpg_pendulum_trainer.py", though.

> By the way, I want to write the tutorial using this example. Is it okay using the example code without Trainer?

It sounds good. By the way, I think DQN is simpler than DDPG as an example of RL.

keisuke-umezawa:
@muupan @ir5 Thank you for the comment! I just misunderstood the rules because the other examples are written with Trainer.

> > By the way, I want to write the tutorial using this example. Is it okay using the example code without Trainer?
> It sounds good. By the way, I think DQN is simpler than DDPG as an example of RL.

For the first step, I will write the tutorial with dqn example.

muupan:
> For the first step, I will write the tutorial with dqn example.

Good! Then, what about this PR?

keisuke-umezawa:
so, I wii close the pr. If you develop the trainer with which the training loop can be more explicit, let's replace the example!

PR #3163, opened 2017-08-16
New style convolution_2d and deconvolution_2d

beam2d:
This PR implements the new style version of `F.convolution_2d` and `F.deconvolution_2d`. These functions are now twice differentiable (but not three times differentiable until some functions used in backward, e.g. `F.sum`, support double backward).

This PR depends on #3096, and is a part of #3147.

Issue #3147, opened 2017-08-15
Double backward support for v3


beam2d:
This is a todo list of supporting double backward in typical convnets and RNNs. The aim of this issue is to manage the tasks so that no work will duplicate.

The original list is from my colleague that uses double backward in his research (currently using another framework or manual derivation of the first grad). Please add a comment if you want to add another function to the list. Let me know if you want to work on the task (I'll write your name to the list so that it is obvious who is working on what).

- [x] `+`, `*` (#2970)
- [x] `linear` (#3099, @beam2d)
- [x] `convolution_2d` and `deconvolution_2d` (#3163, @beam2d)
- [x] `max_pooling_2d` (#3257, @mitmul)
- [x] `average_pooling_2d` (#3257, @mitmul)
- [x] `upsampling_2d` (#3257, @mitmul)
- [x] `unpooling_2d` (#3257, @mitmul)
- [x] `relu` (#3175, @delta2323)
- [x] `lrelu` (#3177, @niboshi)
- [x] `exp` ( #3199 @aonotas )
- [x] `tanh` (#3200, @unnonouno )
- [x] `softmax` (#3213, @takagi )
- [x] `sigmoid` (#3119, @keisuke-umezawa)
- [x] `lstm` (#3206, @unnonouno)
- [x] `embed_id` (#3183, @unnonouno)
- [x] `mean_squared_error` (#3194, @delta2323)
- [x] `softmax_cross_entropy` (#3296, @delta2323)
- [x] `batch_normalization` (#3275, @beam2d)
- [x] `layer_normalization` (#3219, @niboshi)
- [x] `flatten` (#3190, @okuta)
- [x] `concat` and `split_axis` (#3189, @okuta)
- [x] `reshape`, `expand_dims`, `broadcast_to`, and `sum` (#3188, @okuta)
- [x] `get_item` (#3243, @delta2323)
- [x] `log_softmax` (#3217, @delta2323)
- [x] `rollaxis` (#3306, @delta2323)
- [x] `select_item` (#3308, @delta2323)
- [x] `dropout` (#3356, @bonprosoft)

After we decide on the treatment of existing, optimized backward #3146, I will write a guideline to write new-style functions in v3, and then dispatch tasks to core dev members.

beam2d:
Note that the list is not a promise; we may remove some functions from the list in case of shortage of time.

beam2d:
[...]

PR #3188, opened 2017-08-20
New style reshape, expand_dims, broadcast_to and sum

okuta:
This PR depends on #3096, and is a part of #3147.
This PR improves the behavior of #3099 and #3163.

### DIFF
--- a/chainer/functions/array/broadcast.py
+++ b/chainer/functions/array/broadcast.py
@@ -128,7 +128,7 @@ class BroadcastTo(function_node.FunctionNode):
         shape = self._in_shape
         ndim = len(shape)
         if gx.ndim != ndim:
-            gx = sum.sum(gx, axis=tuple(range(gx.ndim - ndim)))
+            gx = sum.sum(gx, tuple(range(gx.ndim - ndim)), False)
 
         axis = [i for i, sx in enumerate(shape) if sx == 1]
         if len(axis) > 0:
--- a/chainer/functions/array/broadcast.py
+++ b/chainer/functions/array/broadcast.py
@@ -5,7 +5,6 @@ from chainer import cuda
 from chainer import function
 from chainer import function_node
 from chainer.utils import type_check
-from chainer import variable
 
 
 def _backward_one(xp, shape, dtype, g):
@@ -114,8 +113,6 @@ class BroadcastTo(function_node.FunctionNode):
     def forward(self, inputs):
         x, = inputs
         xp = cuda.get_array_module(x)
-        self._in_shape = x.shape
-        self._in_dtype = x.dtype
         if hasattr(xp, 'broadcast_to'):
             return xp.broadcast_to(x, self._shape),
         else:
@@ -126,7 +123,7 @@ class BroadcastTo(function_node.FunctionNode):
 
     def backward(self, indexes, grad_outputs):
         gx, = grad_outputs
-        shape = self._in_shape
+        shape = self.inputs[0].shape
         ndim = len(shape)
         if gx.ndim != ndim:
             gx = chainer.functions.sum(gx, tuple(range(gx.ndim - ndim)), False)
@@ -165,9 +162,6 @@ def broadcast_to(x, shape):
 
     """
     if x.shape == shape:
-        if isinstance(x, variable.Variable):
-            return x
-        else:
-            return variable.Variable(x, requires_grad=False)
+        return chainer.as_variable(x)
     y, = BroadcastTo(shape).apply((x,))
     return y
--- a/chainer/functions/array/broadcast.py
+++ b/chainer/functions/array/broadcast.py
@@ -2,6 +2,8 @@ import six
 
 from chainer import cuda
 from chainer import function
+from chainer import function_node
+from chainer.functions.math import sum
 from chainer.utils import type_check
 
 
@@ -84,7 +86,7 @@ def broadcast(*args):
     return Broadcast()(*args)
 
 
-class BroadcastTo(function.Function):
+class BroadcastTo(function_node.FunctionNode):
 
     """Function that broadcasts an array to a new shape."""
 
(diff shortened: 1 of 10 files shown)


chainer/functions/array/broadcast.py
```diff
@@ -108,11 +111,9 @@ def check_type_forward(self, in_types):
...
 
-    def forward(self, xs):
-        self.retain_inputs(())
-        xp = cuda.get_array_module(*xs)
-        x = xs[0]
-        self._xp = xp
+    def forward(self, inputs):
+        x, = inputs
+        xp = cuda.get_array_module(x)
         self._in_shape = x.shape
```
beam2d: You can fetch input shape/dtype from `self.inputs[0]` (which is a `VariableNode` object) in backward even if the input array is not retained.

chainer/functions/array/broadcast.py
```diff
@@ -123,9 +124,18 @@ def forward(self, xs):
...
 
-    def backward(self, xs, grads):
-        return _backward_one(
-            self._xp, self._in_shape, self._in_dtype, grads[0]),
+    def backward(self, indexes, grad_outputs):
+        gx, = grad_outputs
+        shape = self._in_shape
+        ndim = len(shape)
+        if gx.ndim != ndim:
+            gx = chainer.functions.sum(gx, tuple(range(gx.ndim - ndim)), False)
```
beam2d: I feel it's better to merge these two sums. What do you think?

chainer/functions/array/broadcast.py
```diff
@@ -154,4 +164,10 @@ def broadcast_to(x, shape):
                [0, 1, 2]])
 
     """
-    return BroadcastTo(shape)(x)
+    if x.shape == shape:
+        if isinstance(x, variable.Variable):
```
beam2d: Use `chainer.Variable`.

chainer/functions/array/broadcast.py
```diff
@@ -154,4 +164,10 @@ def broadcast_to(x, shape):
...
+            return x
+        else:
+            return variable.Variable(x, requires_grad=False)
```
beam2d: ditto

beam2d: I feel this code fragment should be separated into a common utility function (like `chainer.to_variable`)...

chainer/functions/array/expand_dims.py
```diff
@@ -18,14 +19,15 @@ def check_type_forward(self, in_types):
...
             type_check.expect(x_type.ndim >= -self.axis - 1)
 
-    def forward(self, x):
-        self.retain_inputs(())
-        self._in_shape = x[0].shape
-        xp = cuda.get_array_module(*x)
-        return xp.expand_dims(x[0], self.axis),
+    def forward(self, inputs):
+        x, = inputs
+        self._in_shape = x.shape
```
beam2d: See the comment in `BroadcastTo`

chainer/functions/array/reshape.py
```diff
@@ -40,13 +39,14 @@ def check_type_forward(self, in_types):
             type_check.expect(
                 type_check.prod(x_type.shape) % size_var == 0)
 
-    def forward(self, x):
-        self.retain_inputs(())
-        self._in_shape = x[0].shape
-        return x[0].reshape(self.shape),
+    def forward(self, inputs):
+        x, = inputs
+        self._in_shape = x.shape
```
beam2d: ditto

chainer/functions/array/reshape.py
```diff
@@ -93,4 +93,10 @@ def reshape(x, shape):
         Actual: 8 != 12
 
     """
-    return Reshape(shape)(x)
+    if x.shape == shape:
+        if isinstance(x, variable.Variable):
```
beam2d: Use `chainer.Variable`

chainer/functions/array/reshape.py
```diff
@@ -93,4 +93,10 @@ def reshape(x, shape):
...
+            return x
+        else:
+            return variable.Variable(x, requires_grad=False)
```
beam2d: ditto

chainer/functions/math/sum.py
```diff
@@ -40,35 +43,27 @@ def check_type_forward(self, in_types):
...
-        if not (len(self._in_shape) == 0 or
-                self.axis is None or self.keepdims):
-            actual_axis = []
-            for axis in self.axis:
-                if axis < 0:
-                    axis += len(self._in_shape)
-                actual_axis.append(axis)
+    def forward(self, inputs):
+        x, = inputs
+        self._in_shape = x.shape
```
beam2d: See the comment to `BroadcastTo`

chainer/functions/array/broadcast.py
beam2d: I made a PR to abstract this code: #3218.

chainer/functions/array/broadcast.py
okuta: I cannot merge two sums, because `keepdims` is different.

chainer/functions/array/broadcast.py
okuta: I fixed

okuta:
Cloud you review again?

chainer/functions/array/broadcast.py
beam2d: I thought the difference in `keepdims` can be caught by expand_dims/squeeze. For example, I think it's simpler and more efficient to first take sum with `keepdims=True` and then squeeze the preceeding dimensions of length 1 (corresponding to the `keepdims` of the first sum in the original code).

beam2d:
I added a comment on the sum in `broadcast_to.backward` as follows (pasted since the original comment is already squeezed as "outdated").

> I thought the difference in keepdims can be caught by expand_dims/squeeze. For example, I think it's simpler and more efficient to first take sum with keepdims=True and then squeeze the preceeding dimensions of length 1 (corresponding to the keepdims of the first sum in the original code).

okuta:
OK, I improved it.

beam2d:
Jenkins, test this please.

beam2d:
Jenkins, test this please

beam2d:
LGTM

PR #3192, opened 2017-08-21
Fix Variable.backward for manually edited requires_grad

beam2d:
I added a quick check to `Variable.backward` to skip backward when no grad is required. Such a case does not happen usually (because the output has `requires_grad=False` when all the inputs have `requires_grad=False`); this PR rather makes it clear when one reads the implementation fo `Variable.backward`.


delta2323:
jenkins, test this please.

niboshi:
jenkins, test this please.

niboshi:
LGTM!

PR #3201, opened 2017-08-21
use test mode in ResNet predict

mottodora:
Inferring from only one sample with train mode BatchNorm does not work.
I think it is necessary to use `with chainer.using_config('train', False)`.

### DIFF
--- a/chainer/links/model/vision/resnet.py
+++ b/chainer/links/model/vision/resnet.py
@@ -265,13 +265,13 @@ class ResNetLayers(link.Chain):
             x = x[:, :, 16:240, 16:240]
         # Use no_backprop_mode to reduce memory consumption
         with function.no_backprop_mode(), chainer.using_config('train', False):
-                x = Variable(self.xp.asarray(x))
-                y = self(x, layers=['prob'])['prob']
-                if oversample:
-                    n = y.data.shape[0] // 10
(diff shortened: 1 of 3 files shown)


chainer/links/model/vision/resnet.py
```diff
@@ -264,13 +265,14 @@ def predict(self, images, oversample=True):
...
         # Use no_backprop_mode to reduce memory consumption
         with function.no_backprop_mode():
-            x = Variable(self.xp.asarray(x))
-            y = self(x, layers=['prob'])['prob']
-            if oversample:
-                n = y.data.shape[0] // 10
-                y_shape = y.data.shape[1:]
-                y = reshape(y, (n, 10) + y_shape)
-                y = sum(y, axis=1) / 10
+            with chainer.using_config('train', False):
```
mitmul: Both train=False and no_backprop_mode should be specified here like this:

```
with chainer.using_config('train', False), function.no_backprop_mode():
```

mitmul:
@mottodora Could you apply the same change to VGG16Layers and GoogLeNet too?

mottodora:
Yes. I've done.

mitmul:
Jenkins, test this please

mitmul:
LGTM

PR #3205, opened 2017-08-21
[backport] Suppot tuple for axis argument in average

knorth55:
backport of #3118 
it doesn't seem to be backported to `chainer-v2`.

### DIFF
--- a/chainer/functions/math/average.py
+++ b/chainer/functions/math/average.py
@@ -1,3 +1,5 @@
+import six
+
 from chainer.functions.array import broadcast
 from chainer.functions.array import reshape
 from chainer.functions.math import sum as sum_mod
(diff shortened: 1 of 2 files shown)


niboshi:
Thank you for PR!
However, this is a new feature.
We add new features only into development branch (`master`).
Please wait until the next major version will be released.

knorth55:
What do you mean by next release?
This feature won't be merged to v2?

niboshi:
No, it won't.
We only merge bug-fixes and enhancements (those that do not change external behavior) to stable versions (currently v2).

knorth55:
There will be no more new feature merged in `v2`?
That's really pity.

I'm satisfied with `v2` and I just want to use this new feature.
You mean I need to switch to `v3` in order to use the small feature.
I haven never used `v3`, but if there is as big difference between `v2` and `v3` as that between `v1` and `v2`, it is troublesome to switch to `v3`.

knorth55:
by the way, i thought this is an enhancement, isn't it?

niboshi:
Thank you for the precious feedback.
We discussed internally about this policy, and decided to loosen it.
From now on, we will backport minor feature addition like this one.

niboshi:
I posted a new backport PR (https://github.com/chainer/chainer/pull/3264), because I could not reopen this one.

knorth55:
Good decision! Thank you!

PR #3213, opened 2017-08-22
Implement new style softmax

takagi:
Rel #3147.

This PR implements the new style softmax layer along with double backward support.

### DIFF
--- a/chainer/functions/activation/softmax.py
+++ b/chainer/functions/activation/softmax.py
@@ -106,7 +106,7 @@ class _SoftmaxGrad(function_node.FunctionNode):
         if 0 in indexes:
             s = chainer.functions.broadcast_to(chainer.functions.sum(
                 y * gy, axis=self.axis, keepdims=True), gy.shape)
-            gy2 = -ggx * s + ga * gy
+            gy2 = ga * gy - ggx * s
             ret.append(gy2)
         if 1 in indexes:
             ggy = ga * y
--- a/chainer/functions/activation/softmax.py
+++ b/chainer/functions/activation/softmax.py
@@ -102,12 +102,14 @@ class _SoftmaxGrad(function_node.FunctionNode):
     def backward(self, indexes, grad_outputs):
         y, gy = self.get_retained_inputs()
         ggx, = grad_outputs
-        gs = sum_.sum(-ggx * y, self.axis, True)
+        gs = sum_.sum(-ggx * y, axis=self.axis, keepdims=True)
         ga = ggx + broadcast.broadcast_to(gs, gy.shape)
         ret = []
         if 0 in indexes:
-            gy = -ggx * gs + ga * gy
-            ret.append(gy)
+            s = broadcast.broadcast_to(
+                sum_.sum(y * gy, axis=self.axis, keepdims=True))
+            gy2 = -ggx * s + ga * gy
+            ret.append(gy2)
         if 1 in indexes:
             ggy = ga * y
             ret.append(ggy)
--- a/chainer/functions/activation/softmax.py
+++ b/chainer/functions/activation/softmax.py
@@ -12,6 +12,14 @@ if cuda.cudnn_enabled:
     _mode = libcudnn.CUDNN_SOFTMAX_MODE_CHANNEL
 
 
+def _get_tensor4d_shape(axis, shape):
+    left_shape = numpy.prod(shape[slice(0, axis)], dtype=numpy.int)
+    center_shape = shape[axis]
+    right_shape = numpy.prod(
+        shape[slice(axis + 1, len(shape))], dtype=numpy.int)
+    return left_shape, center_shape, right_shape, 1
+
+
 class Softmax(function_node.FunctionNode):
 
     """Softmax activation function."""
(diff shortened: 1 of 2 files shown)


takagi:
Thanks to @beam2d , now I can remove WIP.

chainer/functions/activation/softmax.py
```diff
@@ -2,7 +2,9 @@
 
 import chainer
 from chainer import cuda
-from chainer import function
+from chainer import function_node
+from chainer.functions.array import broadcast
```
beam2d: Use `chainer.functions` namespace instead (same for the next line).

chainer/functions/activation/softmax.py
```diff
@@ -50,12 +60,23 @@ def forward(self, x):
             y /= y.sum(axis=self.axis, keepdims=True)
 
         self._x_shape = x[0].shape
```
beam2d: It looks you can remove it.

chainer/functions/activation/softmax.py
```diff
@@ -64,26 +85,35 @@ def backward(self, x, gy):
...
-        left_shape = numpy.prod(shape[slice(0, self.axis)], dtype=numpy.int)
-        center_shape = shape[self.axis]
-        right_shape = numpy.prod(
-            shape[slice(self.axis + 1, len(shape))], dtype=numpy.int)
-        return left_shape, center_shape, right_shape, 1
+    def backward(self, indexes, grad_outputs):
+        y, gy = self.get_retained_inputs()
+        ggx, = grad_outputs
+        gs = sum_.sum(-ggx * y, axis=self.axis, keepdims=True)
+        ga = ggx + broadcast.broadcast_to(gs, gy.shape)
```
beam2d: It's better to subtract `sum(ggx*y)` instead of adding `sum(-ggx*y)`. The number of function applications will be reduced by one.

chainer/functions/activation/softmax.py
```diff
@@ -64,26 +85,35 @@ def backward(self, x, gy):
...
+        ret = []
+        if 0 in indexes:
+            s = broadcast.broadcast_to(
+                sum_.sum(y * gy, axis=self.axis, keepdims=True), gy.shape)
+            gy2 = -ggx * s + ga * gy
```
beam2d: Better to write `ga * gy - ggx * s`.

chainer/functions/activation/softmax.py
takagi: Replaced!

chainer/functions/activation/softmax.py
takagi: Removed!

chainer/functions/activation/softmax.py
takagi: Move the subtract out!

chainer/functions/activation/softmax.py
takagi: Fixed the order!

chainer/functions/activation/softmax.py
```diff
@@ -64,26 +83,35 @@ def backward(self, x, gy):
...
-        left_shape = numpy.prod(shape[slice(0, self.axis)], dtype=numpy.int)
-        center_shape = shape[self.axis]
-        right_shape = numpy.prod(
-            shape[slice(self.axis + 1, len(shape))], dtype=numpy.int)
-        return left_shape, center_shape, right_shape, 1
+    def backward(self, indexes, grad_outputs):
+        y, gy = self.get_retained_inputs()
+        ggx, = grad_outputs
+        gs = -chainer.functions.sum(ggx * y, axis=self.axis, keepdims=True)
+        ga = ggx + chainer.functions.broadcast_to(gs, gy.shape)
```
beam2d: Sorry, I was ambiguous. I though it's better to write it as follows.
```
gs = chainer.functions.sum(...)
ga = ggx - chainer.functions.broadcast_to(gs, ...)
```
so that negation+add (two functions) is replaced by subtract (one function).

tests/chainer_tests/functions_tests/activation_tests/test_softmax.py
```diff
@@ -101,6 +104,33 @@ def test_backward_gpu_non_contiguous(self):
...
 
+    def check_double_backward(self, x_data, gy_data, ggx_data,
+                              use_cudnn='always'):
+        def f(x):
+            return functions.softmax(x, axis=self.axis)
+
+        with chainer.using_config('use_cudnn', use_cudnn):
+            gradient_check.check_double_backward(
+                f, (x_data,), (gy_data,), (ggx_data),
+                **self.check_backward_options)
```
beam2d: Sorry, one another comment: it seems this test is numerically unstable. I recommend to use a slightly looser threshold for double backprop test here.

chainer/functions/activation/softmax.py
takagi: Yes, fixed!

tests/chainer_tests/functions_tests/activation_tests/test_softmax.py
takagi: I fixed to give looser threshold by an order for double backprop test in case that dtype is float16. Is it enough or better to fix for other dtypes too?

beam2d:
LGTM!

PR #3218, opened 2017-08-23
Add chainer.as_variable

beam2d:
This PR adds a convenient function `chainer.to_variable`. It can be used to force a value having a type `Variable`. This function is designed to be used in differentiable functions that may skip computation; in such a case, these functions have to conditionally wrap an array by `Variable`. The `to_variable` can be used in such a case. Note that `requires_grad` is set to `False` because of assuming such usage.

### DIFF
--- a/chainer/__init__.py
+++ b/chainer/__init__.py
@@ -56,6 +56,7 @@ from chainer.serializer import AbstractSerializer  # NOQA
 from chainer.serializer import Deserializer  # NOQA
 from chainer.serializer import Serializer  # NOQA
 from chainer.variable import Parameter  # NOQA
+from chainer.variable import to_variable  # NOQA
 from chainer.variable import Variable  # NOQA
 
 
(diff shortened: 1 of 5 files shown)


chainer/functions/connection/linear.py
```diff
@@ -100,3 +100,4 @@ def linear(x, W, b=None):
         return LinearFunction()(x, W)
     else:
         return LinearFunction()(x, W, b)
+0
```
okuta: Please remove this line

beam2d:
Oops, fixed that part. Also rebased to the master.

niboshi:
How about `as_variable` instead, because it returns `Variable` as-is?

beam2d:
That sounds better to me. One concern is an inconsistency with the existing names like `to_cpu` and `to_gpu`, but I think `as_variable` better explains what it does. I will fix it.

beam2d:
Updated.

okuta:
Please fix import order.

beam2d:
Fixed.

okuta:
LGTM!

PR #3223, opened 2017-08-23
Add Variable.array that is less confusing than Variable.data

### LINKED ISSUE #2908: Make it easier to debug code which mixes up Variable and ndarray
I often encounter a bug caused by mixing up Variable and ndarray. There is one caveat on it; since both have `.data`, an error often occurs far later than the point that we should fix. I want to make it easier to debug such a case.

One idea is to add an alias of `Variable.data`, like `Variable.array`, so that we can avoid such a situation by letting us use the new alias instead of `.data`. If we always use this alias, an error immediately occurs once we try to access the attribute of an ndarray. It is still later than the point we start mixing up Variable and ndarray, though.

beam2d:
This PR adds `Variable.array` attribute, which is equivalent to and less confusing than `Variable.data`. Fix #2908.

### DIFF
--- a/chainer/variable.py
+++ b/chainer/variable.py
@@ -413,12 +413,6 @@ class Variable(object):
         requires_grad (bool): Boolean indicating whether ``grad`` will be set
             in backward calculation.
 
-    Attributes:
-        ~Variable.data: Data array of type either :class:`numpy.ndarray` or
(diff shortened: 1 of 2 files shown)


okuta:
LGTM!

PR #3226, opened 2017-08-23
[backport] fix chainer.Link docstring

### LINKED ISSUE #3225: document bug report: chainer.Link page has incorrect code.
https://docs.chainer.org/en/stable/reference/core/generated/chainer.Link.html?highlight=Link
there is an example:
```
 import chainer
import chainer.functions as F
from chainer import initializers
import numpy as np

class LinearLayer(chainer.Link):

    def __init__(self, n_in, n_out):
        super(LinearLayer, self).__init__()
        with self.init_scope():
            self.W = chainer.Parameter(
                (n_out, n_in), initializers.Normal())
            self.b = chainer.Parameter(
                (n_out,), initializers.Zero())

    def __call__(self, x):
        return F.linear(x, self.W, self.b)
```
but chainer.Parameter constructor function is define as :
class chainer.Parameter(**initializer=None, shape=None, name=None**)
note that the order of argument.
so if we omit the parameter keyword , this is incorrect code.

mitmul:
This PR fixes #3225 by back porting #2814

### DIFF
--- a/chainer/link.py
+++ b/chainer/link.py
@@ -102,9 +102,9 @@ class Link(object):
                   super(LinearLayer, self).__init__()
                   with self.init_scope():
                       self.W = chainer.Parameter(
-                          (n_out, n_in), initializers.Normal())
+                          initializers.Normal(), (n_out, n_in))
                       self.b = chainer.Parameter(
-                          (n_out,), initializers.Zero())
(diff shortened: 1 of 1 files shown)


okuta:
LGTM!

PR #3227, opened 2017-08-23
Fix warning message for cuDNN

### LINKED ISSUE #3209: Fix confusing warning message
https://github.com/chainer/chainer/blob/79d6bf6f4f5c86ba705b8fd377368519bc1fd264/chainer/cuda.py#L84-L86

When cupy is not built with cudnn, one do not have to reinstall chainer but cupy.

mitmul:
Resolves #3209


okuta:
LGTM!

PR #3235, opened 2017-08-24
doc: Fix truncated summary line

### LINKED ISSUE #3234: Documentation bug report: Some of the descriptions in the tables are truncated.
Some of the descriptions in the tables are truncated, e.g. `chainer.functions.batch_l2_norm_squared`, `chainer.functions.normalize` in [Standard Function implementations page](https://docs.chainer.org/en/latest/reference/functions.html).

I made the following observations.
- The docstring for each of them is not truncated and written correctly, e.g. [chainer.functions.batch_l2_norm_squared](https://github.com/chainer/chainer/blob/ef10ba098b7bdfb4af3e1e9ae4f53defe4f1c76b/chainer/functions/math/batch_l2_norm_squared.py#L46) .
- The detail page for each of them is rendered correctly, e.g. [chainer.functions.batch_l2_norm_squared](https://docs.chainer.org/en/latest/reference/generated/chainer.functions.batch_l2_norm_squared.html#chainer.functions.batch_l2_norm_squared).
- (The docstring for each of them has the word `a.k.a`.)

niboshi:
Fixes #3234.

The problem is in autosummary's algorithm on parsing docstring: https://github.com/sphinx-doc/sphinx/blob/1.6.3/sphinx/ext/autosummary/__init__.py#L341
It treats the substring which ends with a period followed by a space (or a EOL) as the "first sentence", which trucate the string at e.g. "a.k.a. SOMETHING".

This PR interferes that algorithm by placing `\` followed by an extra space before the existing space, which translates to an empty string. (I'm not sure if this is the correct use of the sequence. But at least it works)

Additionally I put a code to check if docstrings are compliant with autosummary's algorithm.

TODO: Port to CuPy

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -15,6 +15,7 @@
 import inspect
 import os
 import pkg_resources
+import re
 import six
 import sys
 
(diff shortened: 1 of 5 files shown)


docs/source/conf.py
```diff
@@ -388,6 +389,47 @@ def _get_source_relative_path(source_abs_path):
     return os.path.relpath(source_abs_path, _find_source_root(source_abs_path))
 
 
+def _is_docstring_autosummary_compliant(docstring):
+    doc = docstring.split('\n')
+
+    # Extract until the first blank line if any.
+    try:
+        doc = doc[:doc.index('')]
```
okuta: Is this correct?

niboshi: If `doc` does not have `''`, `index()` will raise `ValueError` (thus `doc` is not modified).
If it does, `index()` returns the index of `''`, so this statement extracts elements up until `''`, if any.

```
>>> a = ['a', 'b', 'c']                                                                                                                                                                                            
>>> a[:a.index('b')]                                                                                                                                                                                               
['a']
```

okuta: OK, I understand.

okuta:
LGTM!

Issue #3251, opened 2017-08-25
[New feature] Feature extraction support for Chainer dataset class


corochann:
This is new spec suggestion.

When I face the situation that I want to extract only the `labels` information from dataset, I feel it difficult with current chainer dataset.

And I experimentally implemented `features` indexer, which supports pandas-like index access to for handling data more easy.

See 
Usage example (jupyter notebook): https://gist.github.com/corochann/bafc0b6209f008e50ebe196b00ba56da

Implementaion: https://github.com/chainer/chainer/pull/3252

This implementation is also benefitial to those who write custom loop training (without trainer module) or use sklearn model. Because most of the time, the model input format is the return value of this features indexer.

I'd like to discuss design if it is reasonable to send PR or not.

corochann:
I separated the PR, and at first submit a PR for only TupleDataset.

https://github.com/chainer/chainer/pull/3343

corochann:
It seems that Chainer team will consider another way to provide this functionality, as discussed in https://github.com/chainer/chainer/pull/3252.

PR #3252, opened 2017-08-25
[WIP] Dataset indexer

corochann:
This is work in progress to support `features` indexer for chainer dataset class.

Refer: 
[[New feature] Feature extraction support for Chainer dataset class](https://github.com/chainer/chainer/issues/3251)

### DIFF
--- a/chainer/dataset/dataset_mixin.py
+++ b/chainer/dataset/dataset_mixin.py
@@ -19,6 +19,7 @@ class DatasetMixin(object):
 
     """
     _features_indexer = None
+    _cache_features = None
 
     def __getitem__(self, index):
         """Returns an example or a sequence of examples.
@@ -101,7 +102,14 @@ class DatasetMixin(object):
     def extract_feature_by_slice(self, slice_index, j):
         """This method may be override to support efficient feature extraction.
         
-        If not override, `extract_feature` is used instead.
+        If not override, `ExtractBySliceNotSupportedError` is raised by default, 
+        and in this case `extract_feature` is used instead.
+
+        Args:
+            slice_index (slice): slice of data index to be extracted
+            j (int): `j`-th feature to be extracted
+
+        Returns: feature
 
         """
         raise ExtractBySliceNotSupportedError
@@ -118,13 +126,31 @@ class DatasetMixin(object):
         Returns: feature
 
         """
-        return self.get_example(i)[j]
+        if self._features_indexer._extract_single_feature:
+            data = self.get_example(i)
+        else:
+            if i not in self._cache_features:
+                print('[DEBUG] caching features...')
+                self._cache_features[i] = self.get_example(i)
+            data = self._cache_features[i]
+        if isinstance(data, tuple):
+            return data[j]
+        elif j == 0:
+            return data
+        else:
+            raise ValueError('[Error] unexpected behavior')
 
     @property
     def features(self):
         if self._features_indexer is None:
             self._features_indexer = DatasetMixinFeaturesIndexer(self)
         return self._features_indexer
+
+    def preprocess_extract_feature(self, item):
+        self._cache_features = {}
+
+    def postprocess_extract_feature(self, item):
+        del self._cache_features
 
 
 class DatasetMixinFeaturesIndexer(BaseFeaturesIndexer):
@@ -148,3 +174,9 @@ class DatasetMixinFeaturesIndexer(BaseFeaturesIndexer):
 
     def extract_feature(self, i, j):
         return self.dataset.extract_feature(i, j)
+
+    def preprocess(self, item):
+        self.dataset.preprocess_extract_feature(item)
+
+    def postprocess(self, item):
+        self.dataset.postprocess_extract_feature(item)
--- a/chainer/dataset/dataset_mixin.py
+++ b/chainer/dataset/dataset_mixin.py
@@ -137,19 +137,14 @@ class DatasetMixinFeaturesIndexer(BaseFeaturesIndexer):
             dataset (DatasetMixin): DatasetMixin instance
         """
         super(DatasetMixinFeaturesIndexer, self).__init__(dataset)
+        self.feature_cache = None
 
     @property
     def features_length(self):
-        return len(self.dataset.features_length)
+        return self.dataset.features_length
 
     def extract_feature_by_slice(self, slice_index, j):
         return self.dataset.extract_feature_by_slice(slice_index, j)
 
     def extract_feature(self, i, j):
         return self.dataset.extract_feature(i, j)
-
-    def preprocess(self):
-        pass
-
-    def postprocess(self):
-        pass
--- a/chainer/dataset/dataset_mixin.py
+++ b/chainer/dataset/dataset_mixin.py
@@ -1,6 +1,9 @@
 import numpy
 import six
 
+from chainer.dataset.indexer import BaseFeaturesIndexer, \
+    ExtractBySliceNotSupportedError
+
 
 class DatasetMixin(object):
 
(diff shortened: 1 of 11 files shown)


chainer/dataset/dataset_mixin.py
```diff
@@ -84,3 +88,68 @@ def get_example(self, i):
...
+        This method may be override to support efficient feature extraction.
+
+        Args:
+            i (int): `i`-th data to be extracted
+            j (int): `j`-th feature to be extracted
+
+        Returns: feature
+
+        """
+        return self.get_example(i)[j]
```
delta2323: Can we assume that each sample of the dataset can accept indexing via integers?

chainer/datasets/dict_dataset.py
```diff
@@ -39,3 +43,50 @@ def __getitem__(self, index):
...
+
+
+class DictDatasetFeaturesIndexer(BaseFeaturesIndexer):
+    """FeaturesIndexer for TupleDataset"""
+
+    def __init__(self, dataset):
+        """
+
+        Args:
+            dataset (TupleDataset): TupleDataset instance
```
delta2323: DictDataset?

delta2323:
Are you thinking of supporting indexers for `TupleDataset` and `DictDataset`?

corochann:
I already implemented the support of `dict_dataset` as well.
https://github.com/chainer/chainer/pull/3252/files#diff-afabfc7ee7723839c8042dc74a3b23b5

As you can see, I'm going to implement for `DatasetMixin`, `ImageDataset`, `LabeledImageDataset` class too.

corochann:
Remaining task is to 
- For DatasetMixin class, I think it is better for default behavior to create a cache to return features.
  (currently, it is calling `get_example` every time in extract_feature)
- write a test

Once I got positive comment for the idea, I will proceed above implementation.

delta2323:
It is a handy feature as I frequently do like `labels = [d[1] for d in dataset]`. So I want to proceed the PR. By the way, is it possible to separate it into smaller ones so that we can review it smoothly?

corochann:
I separated the PR, and at first submit a PR for only TupleDataset.

#3343

Hakuyume:
I'm implementing `PickableDataset` (https://github.com/chainer/chainercv/pull/454) which is another solution for #3251. You can see the usage of it [here](https://github.com/chainer/chainercv/pull/454#issuecomment-336684256). I would like to migrate this dataset from ChainerCV to Chainer because it is not only for CV. With this dataset, we can implement efficient feature selection more easily than Indexer.

corochann:
@Hakuyume 
Thank you for your comment, I don't know similar discussion is going on in chainer cv (while I was making PR from almost 2 month ago...).

1. One important use case for me is that I want to extract label information after splitting the dataset into train and validation, which is `SubDataset` (subclass of `DatasetMixin`), as written in this discussion,
Therefore, even specific dataset class support this feature, I cannot extract labels information efficiently when I split the dataset.
https://github.com/chainer/chainer/pull/3343

2. I guess actually `DictDataset` can do the similar job, which `PickableDataset` want to support.
Please refer this test for demo.
https://github.com/chainer/chainer/pull/3252/files#diff-afabfc7ee7723839c8042dc74a3b23b5R48

3. I think you already noticed it, as you are discussing in the chainercv thread (https://github.com/chainer/chainercv/pull/454). 
One another motivation for this PR is to support efficient data extraction, which means, do not copy numpy array but use view reference when possible.
It can be done by overriding `extract_feature_by_slice` method.
https://github.com/chainer/chainer/pull/3343/files#diff-81b6edfb9c10c745e45d1990196ec9f6R73

One troublesome stuff is that if the user want to optimize their own DatasetMixin class, user need to re-implement their own `extract_feature` method or `extract_feature_by_slice` method.

Hakuyume:
> 1. One important use case for me is that I want to extract label information after splitting the dataset into train and validation, which is SubDataset (subclass of DatasetMixin), as written in this discussion (#3343),
Therefore, even specific dataset class support this feature, I cannot extract labels information efficiently when I split the dataset.

As you pointed out, `SubDataset` erases pickable feature. 
One solution is adding `split` or `sub` method to `PickableDataset` which can make sub dataset with pickable feature. Another solution is modifying `SubDataset` to treat `PickableDataset` in different way.

> 2. I guess actually DictDataset can do the similar job, which PickableDataset want to support.
Please refer this test for demo.

`DictDataset` seems not to provide the feature I want. I want to support efficient access for multiple data.

Assume that `dataset` has three data types, `data_0`, `data_1` and `data_2`. The i-th values of `data_0` and `data_1` can be retrieved from a getter `get_0_1(i)`. And the i-th value of `data_2` can be retrieved from another getter `get_2(i)`.

In the case of `PickableDataset`.
```
sub_dataset = dataset.pick('data_0', 'data_1')
sub_dataset[i]  # get_0_1[0] is called only once
```

In the case of `DictDataset, I guess it will be as follows.
```
data_0, data_1 = dataset[:, ['data_0', 'data_1']]
sub_dataset = DictDataset(data_0=data_0, data_1=data_1)
sub_dataset[i]  # get_0_1[0] is called twice
```

> 3. I think you already noticed it, as you are discussing in the chainercv thread (chainer/chainercv#454).
One another motivation for this PR is to support efficient data extraction, which means, do not copy numpy array but use view reference when possible.
It can be done by overriding extract_feature_by_slice method.
https://github.com/chainer/chainer/pull/3343/files#diff-81b6edfb9c10c745e45d1990196ec9f6R73

Yes. That is a disadvantage of `PickableDataset`.

> One troublesome stuff is that if the user want to optimize their own DatasetMixin class, user need to re-implement their own extract_feature method or extract_feature_by_slice method.

I think that is a huge burden for users. With `PickableDataset`, it becomes more easy.

corochann:
1. Actually, it is not only the case for `SubDataset`, but also when we used `ConcatenatedDataset` and `TransformDataset` etc. So I thought it was easier/user-friendly to introduce unified interface for all chainer dataset, rather than develop specific dataset class. 

2. Sorry, I could not get your point.
If you want to extract `i-th` 'data_0' and 'data_1' from DictDataset
```
data_0, data_1 = dict_dataset.features[i, ['data_0', 'data_1']]
```
is enough.
Also if you want to extract 100-200-th data for example, you can call
```
data_0, data_1 = dict_dataset.features[100:200, ['data_0', 'data_1']]
```
and it returns "view" (not copy) of the dataset.

3. Yeah I agree implementing one extra method for dataset_mixin class is troublesome for users.
But even if you don't override the method, the code works. (but it is just not optimized to return view, but copy can be returned.)
https://github.com/chainer/chainer/pull/3252/files#diff-6ec23c4ee85004cde0ad58ad3fa9877dR117

Hakuyume:
> 2.  Sorry, I could not get your point.
  If you want to extract i-th 'data_0' and 'data_1' from DictDataset
  data_0, data_1 = dict_dataset.features[i, ['data_0', 'data_1']]
   is enough.
  Also if you want to extract 100-200-th data for example, you can call
  data_0, data_1 = dict_dataset.features[100:200, ['data_0', 'data_1']]

I want to get a dataset which contains both `data_0` and `data_1`. Not two datasets.

corochann:
Ah I see.

I don't mind the backend method of how to achieve this feature, and I'm happy if you can proceed this feature support.
I just would like to request that it would be great if this feature can be applicable to extended dataset like `SubDataset`, `ConcatenatedDataset` and `TransformDataset` etc.

mitmul:
@beam2d @Hakuyume @corochann we will consider a unified way to satisfy some related demands (https://github.com/chainer/chainer/pull/3976, https://github.com/chainer/chainer/pull/3343). Let us make this PR pending until we make the decision. (I changed the assignee to @delta2323 because he has already reviewed some part of this PR.)

corochann:
I understand, please make it pending or closed.

corochann:
Seems that Chainer team will consider another way to provide this functionality.

PR #3263, opened 2017-08-29
adding ability to convert mnist to rgb format

rcalland:
this is useful when training networks with multiple datasets (for example SVHN, which is RGB)

### DIFF
--- a/chainer/datasets/mnist.py
+++ b/chainer/datasets/mnist.py
@@ -10,7 +10,7 @@ from chainer.datasets import tuple_dataset
 
 
 def get_mnist(withlabel=True, ndim=1, scale=1., dtype=numpy.float32,
-              label_dtype=numpy.int32):
+              label_dtype=numpy.int32, rgb_format=False):
     """Gets the MNIST dataset.
 
     `MNIST <http://yann.lecun.com/exdb/mnist/>`_ is a set of hand-written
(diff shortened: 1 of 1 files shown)


chainer/datasets/mnist.py
```diff
@@ -46,19 +49,22 @@ def get_mnist(withlabel=True, ndim=1, scale=1., dtype=numpy.float32,
...
-def _preprocess_mnist(raw, withlabel, ndim, scale, image_dtype, label_dtype):
+def _preprocess_mnist(raw, withlabel, ndim, scale, image_dtype, label_dtype,
+                      rgb_format):
     images = raw['x']
     if ndim == 2:
         images = images.reshape(-1, 28, 28)
     elif ndim == 3:
         images = images.reshape(-1, 1, 28, 28)
+        if rgb_format:
+            images = numpy.hstack((images, images, images))
```
rezoo: How about using
```python
images = numpy.broadcast_to(images, (len(images), 3) + images.shape[2:])
```
instead of `hstack`? In my environment:
```
In [48]: %time result = np.broadcast_to(train, (60000, 3, 28, 28))
CPU times: user 140 us, sys: 29 us, total: 169 us
Wall time: 177 us

In [49]: %time result = np.hstack((train, train, train))
CPU times: user 88.2 ms, sys: 158 ms, total: 246 ms
Wall time: 248 ms
```

rezoo:
LGTM.

PR #3265, opened 2017-08-29
doc: Fix typo in Variable.grad

niboshi:
Fix minor typo


delta2323:
LGTM

delta2323:
This PR fixes documents that were added during v3. So it is not trivial to backport this PR.

niboshi:
OK, it doesn't have to be backported, then.

PR #3268, opened 2017-08-29
Fix params option of check_double_backward

beam2d:
This PR fixes the bug of `check_double_backward` with `params` option. In order to fully support it, it adds `params_grad_grad` option. It should be merged before v3 RC, otherwise it will be another breaking change.

### DIFF
--- a/tests/chainer_tests/test_gradient_check.py
+++ b/tests/chainer_tests/test_gradient_check.py
@@ -397,5 +397,23 @@ class TestCheckDoubleBackward(unittest.TestCase):
     def test_multiple_input_output_gpu(self):
         self.check_multiple_input_output(cuda.cupy)
 
+    def check_double_backward_with_params(self, xp):
+        arrays = xp.ones((5, 1), dtype='f')
+        x, gy, ggx, param_a, ggparam = arrays
+
+        param = chainer.Variable(param_a)
+
+        def f(x):
+            return x * param
+
+        gradient_check.check_double_backward(f, x, gy, ggx, param, ggparam)
+
+    def test_double_backward_with_params(self):
+        self.check_double_backward_with_params(numpy)
+
+    @attr.gpu
+    def test_double_backward_with_params(self):
+        self.check_double_backward_with_params(cuda.cupy)
+
 
 testing.run_module(__name__, __file__)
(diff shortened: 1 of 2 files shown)


niboshi:
Please fix the flake8 error.

beam2d:
Fixed.

tests/chainer_tests/test_gradient_check.py
```diff
@@ -397,5 +397,23 @@ def test_multiple_input_output_cpu(self):
...
         self.check_multiple_input_output(cuda.cupy)
 
+    def check_double_backward_with_params(self, xp):
+        arrays = xp.ones((5, 1), dtype='f')
+        x, gy, ggx, param_a, ggparam = arrays
+
+        param = chainer.Variable(param_a)
+
+        def f(x):
+            return x * param
```
niboshi: I think this function is too simple to test 2nd order gradient. How about more complex configuration?

beam2d: Why shouldn't it be simple?

niboshi: 2nd order gradient of this function is zero. If `check_double_backward` had some glitch which incorrectly considers 2nd gradient of any function to be zero, this test case would not detect that glitch.

beam2d: No, the second order gradient of `x * param` is not zero, because it is a bilinear function. Taking gradient w.r.t. `x` after taking gradient w.r.t. `param` results in the gradient `1`. Taking gradient w.r.t. `x` twice results in the gradient `0`. Taking gradient w.r.t. `param` after taking gradient w.r.t. `x` results in the gradient `0`. I think this configuration is enough to check if the gradient check correctly handles the parameter.

beam2d: Note that `1` actually depends on the random inputs `y_grad` and `x_grad_grad` so simply setting gradients to `1` does not pass the test.

niboshi: OK.

niboshi:
jenkins, test this please

niboshi:
LGTM!

PR #3269, opened 2017-08-29
Fix test of to_gpu with Variable argument

niboshi:
Fixed a test of `to_gpu` with `Variable` argument.
Currently, `ValueError: Unsupported dtype object` is raised instead of `TypeError`.


okuta:
LGTM!

PR #3271, opened 2017-08-30
Let optimizer do backprop in no-backprop mode

beam2d:
This PR recovers the performance of `Optimizer.update` by skipping the graph creation during backprop.

### DIFF
--- a/chainer/optimizer.py
+++ b/chainer/optimizer.py
@@ -5,6 +5,7 @@ import warnings
 import numpy
 import six
 
+import chainer
 from chainer import cuda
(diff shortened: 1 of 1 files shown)


beam2d:
I have got a comment that it is too destructive to require existing code that uses `Variable.backward()` directly to be modified in order to keep the current performance (including both speed and memory footprint). I will consider to change the API so that the double backprop capability is made optional and the current code keeps the performance without any change.

beam2d:
I made a PR #3271 which replaces this.

PR #3292, opened 2017-08-31
added get_fashion_mnist helper

kashif:
Helper for the fashion-mnist


hanxiao:
the error comes from 
```
FAIL: test_double_backward_cpu (test_log_softmax.transplant_class.<locals>.C)  parameter: {'shape': (2, 2, 2, 3), 'dtype': <class 'numpy.float16'>}
```
seems not related to this PR

hvy:
jenkins, test this please.

hvy:
Thank you for sending us this PR to use your dataset!

Is it possible to rebase this PR on top of the current master and maybe write a test for it.
For reference, there exists a test for the [CIFAR-10 helper](https://github.com/chainer/chainer/blob/master/tests/chainer_tests/datasets_tests/test_cifar.py).

kashif:
sure will do!

hvy:
jenkins, test this please.

hvy:
@kashif Thank you for the quick response. The call to `_preprocess_mnist ` is however currently missing a positional argument `rgb_format ` and will thus cause an error. 

https://github.com/chainer/chainer/blob/96753116c96e26d33e4d747673114b06b86ebb6b/chainer/datasets/mnist.py#L59-L60

On the other hand, I am not sure if we want to reuse a private function that belongs to a different module. Should we make it public (i.e. removing the leading underscore) or maybe move this code into the MNIST module? @niboshi

kashif:
@hvy ok ready for another review. Regarding the private function I am using 2 of them... so I can change it to public if you want?

niboshi:
Yes, in priciple, you shoudn't use private entities of other modules.
OTOH, it's better not to expose these functions to users, because they're for internal implementation.

How about making a private module such as `chainer.datasets._mnist_helper` and put them into it (without `_` prefix)?

hvy:
@niboshi @kashif Thank you for the feedback. Yes I think that's okay. 

This is not a strong opinion but I would ideally reorganize `chainer/datasets`, as it currently contains both reusable classes and concrete datasets in the same hierarchy. But I guess that's another topic.

hvy:
@kashif Would it be possible to refactor the code according to the feedback above? Sorry for repeatedly coming back to you.

kashif:
sure @hvy, I was a bit confused as to which of the feedback to follow now. Can you kindly write it down in the comment concretely and I will do it 🙇

hvy:
Sorry for the ambiguity. Reorganizing `chainer.datasets` will probably be a future topic. What I meant was this.

> How about making a private module such as chainer.datasets._mnist_helper and put them into it (without _ prefix)?

More concretely, could you please create a module `chainer.datasets._mnist_helper` and move the two functions `_preprocess_mnist ` and `_make_npz ` there, but without the prepending `_` to that module. Then change the current Fashion MNIST helper, as well as the MNIST helper to use those functions instead.

hvy:
jenkins, test this please.

hvy:
@kashif Again, thank you for the changes. I noticed when running the tests locally that they failed because of the mock module. I fixed it in this PR, could you merge this one into your fork/branch? https://github.com/kashif/chainer/pull/1

kashif:
@hvy ah thanks! done

hvy:
@kashif Thank you for the quick response. LGTM (to us).

hvy:
jenkins, test this please.

hvy:
jenkins, test this please

hvy:
FYI, caching logic tests might be unified into a single test instead of running them for each separate dataset, in which case this test will be affected. However, the helper itself should not be.

PR #3105, opened 2017-08-06
Improve docs of softmax_cross_entropy

keisuke-umezawa:
This is PR for improving docs of functions/links. Related issue: #2182

Modified functions/links:

- functions/loss/softmax_cross_entropy.py

PR #3296, opened 2017-09-01
New style softmax_cross_entropy

delta2323:
This PR applies new-style function API to `F.softmax_cross_entropy`, as a part of #3147. This PR depends on double-backpropable `F.rollaxis` (#3306) and `F.select_item` (#3308). So I would like reviewers to check them first.


delta2323:
I finished the first version of new-style `F.softmax_cross_entropy`.
I added `enable_double_backprop` argument to the function that switches to enable/disable double-backpropable implementation. If this is turned off, the function uses existing one. The former could be slower then existing one because of the overhead of creating additional computational graphs, though we have not profile its performance. So the motivation of this option is to enable users to choose these two implementation. By default `enable_double_backprop` is set to `True`.

I slightly changed the behavior of `cache_score` so that it is forcibly disabled when `enable_double_backprop` is `True`. This is because double-backpropable SCE does not cache the intermediate values.

delta2323:
One implementation choice I want to discuss is [here](https://github.com/chainer/chainer/pull/3296/files#diff-7f9fda8b243258115ef7bdac145fec8dR259). To avoid the boundary check of `F.select_item` in the debug mode, this function copy `t`. The other options are to loosen the check of `F.select_item`.

delta2323:
Sorry, several tests are still failing. So I put [WIP] again.

chainer/functions/loss/softmax_cross_entropy.py
```diff
@@ -220,9 +234,54 @@ def backward_gpu(self, inputs, grad_outputs):
...
+        loss = loss / count
+        return chainer.functions.sum(loss)
+    else:
+        return loss
+
+
 def softmax_cross_entropy(
         x, t, normalize=True, cache_score=True, class_weight=None,
-        ignore_label=-1, reduce='mean'):
+        ignore_label=-1, reduce='mean', enable_double_backprop=True):
```
beam2d: It's better to keep the current performance by default, so let's use `False` as the default value. It would be better to add a note to the docstring of `L.Classifier` on how to turn it on (e.g. by passing a lambda function to `lossfun`).

tests/chainer_tests/functions_tests/loss_tests/test_softmax_cross_entropy.py
```diff
@@ -60,10 +65,12 @@ def check_forward(self, x_data, t_data, class_weight, use_cudnn='always'):
...
                 x, t, normalize=self.normalize,
-                cache_score=self.cache_score, class_weight=class_weight)
+                cache_score=self.cache_score, class_weight=class_weight,
+                enable_double_backprop=self.enable_double_backprop)
         self.assertEqual(loss.data.shape, ())
         self.assertEqual(loss.data.dtype, self.dtype)
-        self.assertEqual(hasattr(loss.creator, 'y'), self.cache_score)
+        if not self.enable_double_backprop:
+            self.assertEqual(hasattr(loss.creator, 'y'), self.cache_score)
         loss_value = float(cuda.to_cpu(loss.data))
```
beam2d: How about adding a test that checks if the forward outputs with and without `enable_double_backprop` are almost equal to each other? (or checking both of them against some naive implementation (e.g. using only numpy interface) is also ok).

chainer/functions/loss/softmax_cross_entropy.py
```diff
@@ -220,9 +234,54 @@ def backward_gpu(self, inputs, grad_outputs):
...
+
+    in_use = (t != ignore_label).astype(x.dtype)
+
+    loss = chainer.functions.rollaxis(loss, 1, loss.ndim)
+    loss = chainer.functions.reshape(loss, (-1, loss.shape[-1]))
+
+    # As F.select_item does not allow the elements of t to be negative,
+    # we substitute ignore_label with 0, which is valid.
+    t = t.copy()
+    t[t == ignore_label] = 0
```
beam2d: These two lines can be replaced by `xp.clip(t, 0, loss.shape[1] - 1)`.

delta2323: Strangely, this change failed to pass some of double backprop tests.

tests/chainer_tests/functions_tests/loss_tests/test_softmax_cross_entropy.py
delta2323: Added tests that check consistency of two implementations.

chainer/functions/loss/softmax_cross_entropy.py
delta2323: Done

delta2323:
Fixed the implementation except [the comment](https://github.com/chainer/chainer/pull/3296#discussion_r136747919)

delta2323:
Related to #3105

chainer/functions/loss/softmax_cross_entropy.py
beam2d: That's strange... I will check it.

delta2323: Thank you.

delta2323: Could you check by yourself that the failure occurs in your environment, too.

tests/chainer_tests/functions_tests/loss_tests/test_softmax_cross_entropy.py
```diff
@@ -436,5 +545,109 @@ def test_backward_cpu(self):
...
+                0, 10, (self.x.shape[1],)).astype(self.dtype)
+        else:
+            self.class_weight = None
+
+    def check_consistency(self, xp):
+
+        if self.class_weight is None:
+            class_weight = None
+        else:
+            xp.asarray(self.class_weight)
```
beam2d: It looks you forgot to assign it to `class_weight`

chainer/functions/loss/softmax_cross_entropy.py
beam2d: I modified these lines into `t = t.clip(0, loss.shape[1] - 1)` and it passed all the test cases of `test_softmax_cross_entropy.py`. Could you check it? (FYI: the above `xp.clip` version also passed).

chainer/functions/loss/softmax_cross_entropy.py
```diff
@@ -220,9 +234,55 @@ def backward_gpu(self, inputs, grad_outputs):
...
+    in_use = (t != ignore_label).astype(x.dtype)
+
+    loss = chainer.functions.rollaxis(loss, 1, loss.ndim)
+    loss = chainer.functions.reshape(loss, (-1, loss.shape[-1]))
+
+    # As F.select_item does not allow the elements of t to be negative,
+    # we substitute ignore_label with 0, which is valid.
+    t = t.copy()
+    xp = chainer.cuda.get_array_module((x, t))
+    t = xp.clip(t, 0, loss.shape[1] - 1)
```
beam2d: You don't have to do copy by yourself because `clip` does make a new array. The following is much simpler.
Before:
```
t = t.copy()
xp = chainer.cuda.get_array_module((x, t))
t = xp.clip(t, 0, loss.shape[1] - 1)
```
After
```
t = t.clip(0, loss.shape[1] - 1)
```

chainer/links/model/classifier.py
```diff
@@ -29,6 +29,23 @@ class Classifier(link.Chain):
         compute_accuracy (bool): If ``True``, compute accuracy on the forward
             computation. The default value is ``True``.
 
+    .. note::
+        This link uses :func:`chainer.softmax_cross_entropy` with
+        default arguments as a loss function (specified by ``lossfun``),
+        if users do not explicit change it. In particular, the loss function
```
beam2d: explicit -> explicitly

chainer/links/model/classifier.py
```diff
@@ -29,6 +29,23 @@ class Classifier(link.Chain):
...
+        does not support double backpropagation.
+        If you need second or higher order differentiation, you need to turn
+        it on with ``enable_double_backprop=True``:
+
+    .. admonition:: Example
```
beam2d: Can this example section be included in the above note? (It looks it is currently a separate section)

delta2323: Fixed

chainer/links/model/classifier.py
delta2323: Fixed

beam2d:
Please fix the flake8 error.

beam2d:
@delta2323 Could you rebase it?

beam2d:
Jenkins, test this please

beam2d:
Could you fix the style?

beam2d:
LGTM

PR #3298, opened 2017-09-01
Make double backprop support optional

beam2d:
This PR makes the double backprop support of `Variable.backward()` optional so that existing code does not suffer from increasing memory consumption and possible slow down. It replaces #3271 because this PR automatically stops creating computational graph in `backward` called from `GradientMethod.update()`.

### DIFF
--- a/chainer/gradient_check.py
+++ b/chainer/gradient_check.py
@@ -374,7 +374,7 @@ def check_double_backward(func, x_data, y_grad, x_grad_grad, params=(),
         y = identity.Identity().apply(y)
 
         _set_y_grad(y, gys)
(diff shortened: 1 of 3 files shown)


niboshi:
I think the name `enable_backprop` might be confusing, because `backward()` itself is backpropagation.
I don't have particularly good alternative, though.

Some random ideas:
* enable_further_backprop
* enable_double_backprop
* enable_higher_order_backprop
* enable_grad_backprop

beam2d:
OK, I will use `enable_double_backprop`.

niboshi:
LGTM!

PR #3308, opened 2017-09-02
New style select_item

delta2323:
This PR applies new-style function API to `F.select_item`, as a part of #3147. Double-backpropable version of this function is used in #3296.


beam2d:
LGTM

PR #3315, opened 2017-09-03
Remove unnecessary class variables

himkt:
I removed unused class variables.


delta2323:
Thank you for suggestion, but perhaps these variables can be referenced outside of this class when it is instantiated. Trainer is such an example[https://github.com/chainer/chainer/blob/83d619f92efb6c842eb6fa283c1fa529b43392ab/chainer/training/trainer.py#L222].

himkt:
Oops I'm sorry, I didn't notice it. I close this PR :bow:

delta2323:
Don't worry. Thank you, anyway.

PR #3316, opened 2017-09-03
Speedup Upsampling2D on CPU

mitmul:
This PR improves the speed of Upsampling2D on CPU by avoiding for loops.

| | forward | backward |
|-|---------|-----------|
| before | 0.3773 +- 0.0086 sec | 0.1923 +- 0.0210 sec |
| after | 0.0153 +- 0.0004 sec | 0.0110 +- 0.0003 sec |
| speedup | ~24.7 x | ~17.5 x |

The performance measurement script is here: https://gist.github.com/mitmul/b0041b4afd58eaf9489b423cb0345bbb

PR #3318, opened 2017-09-04
[backport] Speedup Upsampling2D on CPU

mitmul:
This PR backports #3316


niboshi:
It includes changes not existed in the original PR.
Are they needed?
If so, let's backport those changes first (#2978).
Otherwise, it may be inconsistent with other part of the code.

mitmul:
Thanks, I didn't notice that. will fix them

chainer/functions/pooling/upsampling_2d.py
```diff
@@ -1,11 +1,10 @@
+import numpy
```
niboshi: Please revert this, too.

mitmul: Fixed

niboshi:
LGTM!

PR #3324, opened 2017-09-05
[WIP] Move to PyTest instead of nose

mitmul:


tests/chainer_tests/functions_tests/pooling_tests/test_average_pooling_nd.py
```diff
@@ -14,7 +14,7 @@
 from chainer.testing import attr
 from chainer.testing import condition
 from chainer.utils import conv
-import pooling_nd_helper
+from tests.chainer_tests.functions_tests.pooling_tests import pooling_nd_helper
```
kmatsuura: This test passed on Windows with 'from chainer_tests.functions_tests.pooling_tests import pooling_nd_helper'.

tests/chainer_tests/functions_tests/pooling_tests/test_max_pooling_nd.py
```diff
@@ -15,7 +15,7 @@
 from chainer.testing import attr
 from chainer.testing import condition
 from chainer.utils import conv
-import pooling_nd_helper
+from tests.chainer_tests.functions_tests.pooling_tests import pooling_nd_helper
```
kmatsuura: This test passed on Windows with 'from chainer_tests.functions_tests.pooling_tests import pooling_nd_helper'.

tests/chainer_tests/functions_tests/pooling_tests/test_average_pooling_nd.py
mitmul: Thanks, fixed.

tests/chainer_tests/functions_tests/pooling_tests/test_max_pooling_nd.py
mitmul: Thanks, fixed.

niboshi:
It seems `not gpu` or `gpu < 1` does not work?

```python -m pytest -m "not gpu and not slow"```

Error:
```
self = <chainer.testing.parameterized.TestToCPU_param_0 testMethod=test_cupy_array_async2>

    @attr.multi_gpu(2)
    def test_cupy_array_async2(self):
>       x = cuda.to_gpu(self.x, device=1)

self       = <chainer.testing.parameterized.TestToCPU_param_0 testMethod=test_cupy_array_async2>

tests/chainer_tests/test_cuda.py:177: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
chainer/cuda.py:261: in to_gpu
    with _get_device(device):
cupy/cuda/device.pyx:89: in cupy.cuda.device.Device.__enter__ (cupy/cuda/device.cpp:2357)
    ???
cupy/cuda/device.pyx:104: in cupy.cuda.device.Device.use (cupy/cuda/device.cpp:2598)
    ???
cupy/cuda/runtime.pyx:185: in cupy.cuda.runtime.setDevice
    ???
```

**Comment by niboshi** [comment]
It runs successfully with `python -m pytest -m "not gpu and not multi_gpu and not slow"`.

We can't specify the number of available devices, but it's ok for now because currently we don't have tests using more than 2 devices.

When we need such test in the future, I think we can implement custom parser (https://stackoverflow.com/questions/37583078/how-to-specify-custom-marker-with-py-test-command-line-with-sub-set-of-attribute). It's not trivial but doable.

**Comment by okuta** [comment]
Please remove `tests/chainer_tests/functions_tests/pooling_tests/__pycache__/__init__.py`.

**Comment by niboshi** [comment]
How about fixing AppVeyor configuration files as well?

File: .travis.yml [hunk 3]
```diff
@@ -51,15 +51,16 @@ install:
...
   - pip install autopep8
+  - pip install pytest-xdist
 
 script:
   - flake8
   - autopep8 -r . --global-config .pep8 --diff | tee check_autopep8
   - test ! -s check_autopep8
   - cd tests
-  - PYTHONWARNINGS='ignore::FutureWarning,module::DeprecationWarning' nosetests -a '!gpu,!slow' --with-doctest chainer_tests
+  - PYTHONWARNINGS='ignore::FutureWarning,module::DeprecationWarning' pytest -n 2 -m "not gpu and not slow" chainer_tests
```
niboshi: * How about removing `-n 2`? It's saild that multi-process testing is slower than single-process. (if it's removed, pytest-xdist dependency should be removed as well.)
* `-m` specificiation should be `not gpu and not multi_gpu and not slow`.

.travis.yml
```diff
@@ -51,15 +51,16 @@ install:
   - pip install dist/*.tar.gz
   - pip install h5py pillow
   - travis_wait pip install theano
-  - pip install nose hacking mock
+  - pip install pytest hacking mock
```
niboshi: Is it safe to remove nose dependency? I think existing tests still depend on it.

mitmul: OK, I added nose again.

.travis.yml
mitmul: Thanks. Fixed both points.

mitmul:
@okuta @niboshi Thanks for the comments. I fixed the points. Please check it again.

niboshi:
I think we can remove `pytest-xdist` dependency.

mitmul:
@niboshi Right. Removed.

mitmul:
Resolved conflicts

mitmul:
Rebased and fixed the setting for travis (add `and not cudnn`)

PR #3325, opened 2017-09-05
Skip assert_warns check in Python 2

### LINKED ISSUE #3321: assert_warns fails to catch DeprecationWarning on Python 2
`assert_warns` fails to catch `DeprecationWarning` on Python 2 so CI fails.

https://github.com/chainer/chainer/pull/2932

niboshi:
Fixes #3321.

In Python2, `warnings.warn` will not raise warning multiple times from the same stack frame.

### DIFF
--- a/chainer/testing/helper.py
+++ b/chainer/testing/helper.py
@@ -1,5 +1,6 @@
 import contextlib
 import pkg_resources
+import sys
 import unittest
 import warnings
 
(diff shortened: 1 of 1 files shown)


okuta:
jenkins, test this please.

niboshi:
>In Python2, warnings.warn will not raise warning multiple times from the same stack frame.

I'm sorry, it seems this statement is not correct.

okuta:
LGTM!

niboshi:
Note:
We don't understand under what condition this problem happens, but it only happens in Python 2. So we decided to disable it as a temporary workaround.

PR #3327, opened 2017-09-05
Document initializer criteria

### LINKED ISSUE #3287: Document about initializers
Criteria for initializer arguments of links is complicated (https://github.com/chainer/chainer/pull/3259#issuecomment-325562538). We should write some dedicated documentation about that, and let each link documentation point to it. Maybe we can write on [this page](https://docs.chainer.org/en/v2.0.2/reference/initializers.html).

Also we should describe *the default initializer* (which is `LeCunNormal`).

niboshi:
Fixes #3287

Wrote about initializer criteria.
Also updated `L.Linear` documentation as an example.


delta2323:
Could you put `st:ready-for-review` label if it is?

delta2323:
jenkins, test this please.

PR #3335, opened 2017-09-06
Add seealso link in using_config

niboshi:
I think #2862 can be closed with this fix, because configuration page contains an example of `using_config`.


rezoo:
LGTM.

PR #3336, opened 2017-09-06
Overlap data transfer and GPU kernels

anaruse:
This PR enables to transfer next batch of input data to GPUs while various kernels are run on GPUs using current batch of input data, and will help reduce the training time.

In order to use this feature, you should specify a `converter` argument of StandardUpdater or MultiprocessParallelUpdater as below.

    updater = training.StandardUpdater(..., converter=convert.ConcatWithAsyncTransfer(), ...)


chainer/dataset/convert.py
```diff
@@ -139,3 +145,106 @@ def _concat_arrays_with_padding(arrays, padding):
...
     return result
+
+
+class ConcatWithAsyncTransfer(object):
+
+    """Interface to concatenate data and transfer them to GPU asynchronously.
+
+    It enables to transfer next batch of input data to GPU while GPU is
+    running kernels for training using current batch of input data.
+    """
```
beam2d: Write the documentation of the `stream` argument.

chainer/dataset/convert.py
```diff
@@ -139,3 +145,106 @@ def _concat_arrays_with_padding(arrays, padding):
...
+
+    def __init__(self, stream=None):
+        self.stream = stream
+        self.pin_arrays = [[], []]  # numpy arrays with pinned memory
+        self.cp_arrays = [[], []]  # cupy arrays
```
beam2d: If these two attributes are not a part of "public" API, add `_` to the prefix of the names.

chainer/dataset/convert.py
```diff
@@ -139,3 +145,106 @@ def _concat_arrays_with_padding(arrays, padding):
...
+            The type depends on the type of each example in the batch.
+        """
+        if len(batch) == 0:
+            raise ValueError('batch is empty')
+
+        first_elem = batch[0]
+        if (device is None or
+                not isinstance(first_elem, tuple)):
+            # Default concat method is used when device is not set
+            # of input data is not tuple.
```
beam2d: This comment should be moved to the docstring (otherwise just remove the comment because it just explains what one can see in the code).

chainer/dataset/convert.py
```diff
@@ -139,3 +145,106 @@ def _concat_arrays_with_padding(arrays, padding):
...
+
+        pin_arrays = self.pin_arrays.pop(0)
+        cp_arrays = self.cp_arrays.pop(0)
+
+        with cuda.get_device_from_id(device):
+            if self.stream is None:
+                self.stream = cuda.Stream(non_blocking=True)
+            stream = self.stream
+
+            # Concatenates examples in batch into an array and check if
```
beam2d: Concatenates -> Concatenate

chainer/dataset/convert.py
```diff
@@ -139,3 +145,106 @@ def _concat_arrays_with_padding(arrays, padding):
...
+                if (len(cp_arrays) <= i or
+                        np_arrays[i].nbytes != cp_arrays[i].nbytes):
+                    _same_size = False
+
+            if not _same_size:
+                pin_arrays = []
+                cp_arrays = []
+
+            if len(cp_arrays) == 0:
+                # Allocates memory for numpy arrays with pinned memory
```
beam2d: Allocates -> Allocate

chainer/dataset/convert.py
```diff
@@ -139,3 +145,106 @@ def _concat_arrays_with_padding(arrays, padding):
...
+                # and cupy arrays. These arrays are retaind at
+                # self.pin_arrays andself.cp_arrays respectively and
+                # will be reused for subsequent batches as long as
+                # the size are the same.
+
+                # cuda.Stream.null.synchronize()
```
beam2d: Remove unused code.

chainer/dataset/convert.py
```diff
@@ -139,3 +145,106 @@ def _concat_arrays_with_padding(arrays, padding):
...
+                    pin_mem = cupy.cuda.alloc_pinned_memory(np_array.nbytes)
+                    pin_array = numpy.frombuffer(pin_mem,
+                                                 np_array.dtype,
+                                                 np_array.size
+                                                 ).reshape(np_array.shape)
+                    cp_array = cupy.empty_like(np_array)
+
+                    pin_arrays.append(pin_array)
+                    cp_arrays.append(cp_array)
+                # cuda.Stream.null.synchronize()
```
beam2d: ditto (unused code)

chainer/dataset/convert.py
```diff
@@ -139,3 +145,106 @@ def _concat_arrays_with_padding(arrays, padding):
...
+                runtime.deviceSynchronize()
+
+            results = []
+            for i in six.moves.range(len(first_elem)):
+                pin_arrays[i][...] = np_arrays[i]  # copy(CPU): paged -> pinned
+                cp_arrays[i].set(pin_arrays[i], stream)  # copy: CPU to GPU
+                results.append(cp_arrays[i])
+
+            # stream.synchronize()
+            # cuda.Stream.null.synchronize()
```
beam2d: ditto (unused code)

chainer/dataset/convert.py
```diff
@@ -139,3 +145,106 @@ def _concat_arrays_with_padding(arrays, padding):
...
+
+            if len(cp_arrays) == 0:
+                # Allocates memory for numpy arrays with pinned memory
+                # and cupy arrays. These arrays are retaind at
+                # self.pin_arrays andself.cp_arrays respectively and
+                # will be reused for subsequent batches as long as
+                # the size are the same.
+
+                # cuda.Stream.null.synchronize()
+                runtime.deviceSynchronize()
```
beam2d: Write the reason why we need synchronization here (if my understanding is correct, it is needed to ensure that the data transfer for the previous iteration is finished).

chainer/dataset/convert.py
```diff
@@ -139,3 +145,106 @@ def _concat_arrays_with_padding(arrays, padding):
...
+                    pin_array = numpy.frombuffer(pin_mem,
+                                                 np_array.dtype,
+                                                 np_array.size
+                                                 ).reshape(np_array.shape)
+                    cp_array = cupy.empty_like(np_array)
+
+                    pin_arrays.append(pin_array)
+                    cp_arrays.append(cp_array)
+                # cuda.Stream.null.synchronize()
+                runtime.deviceSynchronize()
```
beam2d: Is this synchronization needed?

beam2d:
Do you think this converter can/should be used as the default one?

chainer/dataset/convert.py
anaruse: will add docs about the `stream`

chainer/dataset/convert.py
anaruse: OK, will change in that way.

chainer/dataset/convert.py
anaruse: OK, probably will remote the comment.

chainer/dataset/convert.py
anaruse: Thanks

chainer/dataset/convert.py
anaruse: Thanks

chainer/dataset/convert.py
anaruse: All right, will remote unused code.

chainer/dataset/convert.py
anaruse: OK, will write the reason. BTW, the reason is that, it is necessary to ensure that *ALL* processing including compute and data transfer that have been launched to GPU have been finished.

chainer/dataset/convert.py
anaruse: Probably, no, it is not actually necessary there. will remote it.

chainer/dataset/convert.py
anaruse: Thanks

chainer/dataset/convert.py
anaruse: Thanks

anaruse:
I think that it is OK to use this as default, because this is implemented in safer way based on the assumption that synchronization may not be handled correctly by the caller routine.

anaruse:
The attached shows the effect of this PR. The red circle is where input data is transferred from CPU to GPU. As you can see, the data transfer is being done while compute kernels for the previous iteration are run on GPU.

image

beam2d:
Then, I think we should make it the default converter. I will discuss it with other members as well.

beam2d:
Sorry for the delay; an internal discussion does not converge, so I will merge it after confirming it passes the CI, and will think about the API after that.

beam2d:
Jenkins, test this please.

beam2d:
I now think this implementation includes a little bit too much about CUDA specific logic to be the default implementation (which would be used as the starting point of a customized converter by users). I think the current implementation itself looks OK, but it would be much great if the double-buffered asynchronous transfer logic is extracted as an independent utility, and is just used by the default converter.

anaruse:
Thank you for your feedback. I basically agree with you on the concern that this implementation requires some detailed knowledge about CUDA and it is not so easy for many users to understand this implementation. Give me time to think about how to generalize the asynchronous data transfer feature using double-buffered scheme.

anaruse:
A new class `Conveyer` is added that handles asynchronous data transfer to device using double buffering scheme when possible.  The class `ConcatWithAsyncTransfer` has been rewritten using the class `Conveyer` . I think it got much easier for many users to understand.

chainer/dataset/convert.py
```diff
@@ -139,3 +145,150 @@ def _concat_arrays_with_padding(arrays, padding):
...
+            for key in first_elem:
+                self.conveyer[key].synchronize()
+
+            return result
+
+        else:
+            return to_device(device, _concat_arrays(batch, padding))
+
+
+class Conveyer(object):
```
jekbradbury: Should probably be "Conveyor" (conveyer is a rare alternative spelling)

anaruse: Thank you for pointing it out! Will change to "Conveyor".

chainer/dataset/convert.py
```diff
@@ -139,3 +145,150 @@ def _concat_arrays_with_padding(arrays, padding):
...
+            asynchronously using this stream. If ``None``, asynchronous data
+            transfer is not used.
+    """
+
+    def __init__(self, device=None, stream=None):
+        self.device = device
+        self.stream = stream
+        self.array_set = [[None, None], [None, None]]
+
+    def __call__(self, array):
```
beam2d: I feel the following API is easier to understand.
- Rename `__call__` to `put`, which does not return anything and just launch asynchronous host-to-device copy (followed by allocations if needed).
- Rename `synchronize` to `get` and let it return the target cupy array with synchronization.

chainer/dataset/convert.py
```diff
@@ -139,3 +145,150 @@ def _concat_arrays_with_padding(arrays, padding):
...
+
+            for key in first_elem:
+                self.conveyor(key).synchronize()
+
+            return result
+
+        else:
+            return to_device(device, _concat_arrays(batch, padding))
+
+    def conveyor(self, key):
```
beam2d: How about using `collections.defaultdict` and just directly accessing like `self._conveyor[key]`? You can pass a closure as the constructor: `self._conveyor = collections.defaultdict(lambda: Conveyor(self.device, self.stream))`.

chainer/dataset/convert.py
```diff
@@ -139,3 +145,150 @@ def _concat_arrays_with_padding(arrays, padding):
...
+    running kernels for training using current batch of input data.
+
+    Args:
+        stream (cupy.cuda.Stream): CUDA stream. If ``None``, a stream is
+            automatically created on the first call. Data transfer operation
+            is launched acynchrnously using the stream.
+    """
+
+    def __init__(self, stream=None):
+        self.stream = stream
```
beam2d: I think `stream` and `device` should not be modified by (i.e., exposed to) users, so it's better to name them like `_stream` and `_device`.

chainer/dataset/convert.py
```diff
@@ -3,6 +3,12 @@
 
 from chainer import cuda
 
+try:
+    import cupy
```
beam2d: Use `cuda.cupy` in the code instead of "try-import"ing cupy.

chainer/dataset/convert.py
```diff
@@ -139,3 +145,150 @@ def _concat_arrays_with_padding(arrays, padding):
...
+        return cp_array
+
+    def synchronize(self):
+        # Wait for completion of the data transfer submitted above.
+        # Global synchronizaton is used here for safer reason.
+        # If a caller function is correctly handling the synchronization,
+        # local synchronization (self.stream.synchronize()) may be enough.
+        if (self.device is not None and self.device >= 0 and
+                self.stream is not None):
+            runtime.deviceSynchronize()
```
beam2d: Is it safe to assume that the current device is same as `self.device`? I think it is better to make this method automatically switch the device, or write explicitly that this method assumes that the device is correctly switched by the caller, and fix `ConcatWithAsyncTransfer` to correctly switch the device (the latter might have lower overhead because only two `cudaSetDevice` is needed, i.e., `with get_device_from_id(self.device):`).

anaruse:
Thank you so much for your detailed comment and feedback! I will update the PR based on your comment.

chainer/dataset/convert.py
anaruse: All right, will stop using "try-import" there.

chainer/dataset/convert.py
```diff
@@ -139,3 +145,150 @@ def _concat_arrays_with_padding(arrays, padding):
...
 
     return result
+
+
+class ConcatWithAsyncTransfer(object):
+
+    """Interface to concatenate data and transfer them to GPU asynchronously.
+
+    It enables to transfer next batch of input data to GPU while GPU is
+    running kernels for training using current batch of input data.
```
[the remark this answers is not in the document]
anaruse: added a comment and a sample code.

chainer/dataset/convert.py
anaruse: The name of the variables were changed.

chainer/dataset/convert.py
anaruse: fixed code in that way.

chainer/dataset/convert.py
```diff
@@ -139,3 +145,150 @@ def _concat_arrays_with_padding(arrays, padding):
...
+    def conveyor(self, key):
+        if key not in self._conveyor:
+            self._conveyor[key] = Conveyor(self.device, self.stream)
+        return self._conveyor[key]
+
+
+class Conveyor(object):
+
+    """Interface to handle asynchronous data transfer using double buffering.
+
```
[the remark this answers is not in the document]
anaruse: The usage was added there.

chainer/dataset/convert.py
anaruse: I agree with you. changed API in that way.

chainer/dataset/convert.py
anaruse: I agree with you that it's safer to use `with get_device_from_id(...)`. changed code in that way.

anaruse:
I've just submitted the updated PR. Please review it.

chainer/dataset/convert.py
```diff
@@ -139,3 +140,181 @@ def _concat_arrays_with_padding(arrays, padding):
...
+            else:
+                return to_device(device, _concat_arrays(batch, padding))
+
+
+class Conveyor(object):
+
+    """Interface to handle asynchronous data transfer using double buffering.
+
+    An asynchrous data transfer is initiated by :meth:`put`, and the result,
+    the array transferd to a target device, is obtained by :meth:`get`.
```
beam2d: transferred?

beam2d:
Jenkins, test this please.

chainer/dataset/convert.py
anaruse: Thank you for pointing it out! will fix it.

PR #3340, opened 2017-09-06
doc: Fix GRU stateless/stateful

niboshi:
Current documentation incorrectly describes `L.GRU` as stateless.

### DIFF
--- a/chainer/links/connection/gru.py
+++ b/chainer/links/connection/gru.py
@@ -53,9 +53,10 @@ class StatelessGRU(GRUBase):
     where :math:`\\sigma` is the sigmoid function, and :math:`\\odot` is the
     element-wise product.
 
-    :class:`~chainer.links.StatelessGRU` does not hold the value of
-    hidden vector :math:`h`. So this is *stateless*.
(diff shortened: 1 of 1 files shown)


delta2323:
jenkins, test this please.

chainer/links/connection/gru.py
```diff
@@ -53,7 +53,7 @@ class StatelessGRU(GRUBase):
     where :math:`\\sigma` is the sigmoid function, and :math:`\\odot` is the
     element-wise product.
 
-    :class:`~chainer.links.GRU` does not hold the value of
+    :class:`~chainer.links.StatelessGRU` does not hold the value of
     hidden vector :math:`h`. So this is *stateless*.
```
delta2323: > So this is *stateless*

This sentence is now somewhat odd as users can infer it from the name. So I think something like "As the name indicates, StatelessGRU is *stateless*, meaning that it does not hold ..." is better.

niboshi: Thank you! Fixed.

delta2323:
LGTM

PR #3343, opened 2017-09-07
features indexer for TupleDataset

corochann:
I separated this little big PR: https://github.com/chainer/chainer/pull/3252.
Motivation of the PR is written in issue: https://github.com/chainer/chainer/issues/3251

This is the first PR to support features indexer only for TupleDataset.

### DIFF
--- a/chainer/dataset/indexer.py
+++ b/chainer/dataset/indexer.py
@@ -7,8 +7,8 @@ class ExtractBySliceNotSupportedError(Exception):
 
 
 class BaseIndexer(object):
-    """Base class for Indexer
-    """
+    """Base class for Indexer"""
+
     def __init__(self, *args, **kwargs):
         pass
 
@@ -17,17 +17,19 @@ class BaseIndexer(object):
 
 
 class BaseFeaturesIndexer(BaseIndexer):
+
     """Base class for FeaturesIndexer
-    
-    Let `features` be the instance of `BaseFeaturesIndexer`, then 
+
+    Let `features` be the instance of `BaseFeaturesIndexer`, then
     `features[i, j]` returns `i`-th dataset of `j`-th feature.
-    
+
     `features[ind]` works same with `features[ind, :]`
-    
-    Note that the returned value will be numpy array, even though the 
-    dataset is initilized with otherformat (e.g. list)
-    
+
+    Note that the returned value will be numpy array, even though the
+    dataset is initilized with other format (e.g. list).
+
     """
+
     def __init__(self, dataset, *args, access_feature_by_key=False, **kwargs):
         super(BaseFeaturesIndexer, self).__init__(*args, **kwargs)
         self.dataset = dataset
@@ -37,7 +39,8 @@ class BaseFeaturesIndexer(BaseIndexer):
     def features_keys(self):
         """Returns all the keys of features
 
-        This method must be override when `access_feature_by_key` is `True`. 
+        This method must be override when `access_feature_by_key` is `True`.
+
         """
         if self.access_feature_by_key:
             raise NotImplementedError
@@ -57,20 +60,20 @@ class BaseFeaturesIndexer(BaseIndexer):
         return self.dataset_length, self.features_length
 
     def extract_feature_by_slice(self, slice_index, j):
-        """Extracts `slice_index`-th data's `j`-th feature,
-        where `slice_index` is indices of slice object.
+        """Extracts `slice_index`-th data's `j`-th feature.
 
+        Here, `slice_index` is indices of slice object.
         This method may be override to support efficient feature extraction.
-        If not override, `ExtractBySliceNotSupportedError` is raised by default, 
-        and in this case `extract_feature` is used instead.
+        If not override, `ExtractBySliceNotSupportedError` is raised by
+        default, and in this case `extract_feature` is used instead.
 
         Args:
             slice_index (slice): slice of data index to be extracted
             j (int): `j`-th feature to be extracted
 
         Returns: feature
-
         """
+
         raise ExtractBySliceNotSupportedError
 
     def extract_feature(self, i, j):
@@ -94,9 +97,9 @@ class BaseFeaturesIndexer(BaseIndexer):
             if isinstance(feature_index[0],
                           (bool, numpy.bool, numpy.bool_)):
                 if len(feature_index) != self.features_length:
-                    raise ValueError('Feature index wrong length {} instead of '
-                                     '{}'.format(len(feature_index),
-                                                 self.features_length))
+                    raise ValueError('Feature index wrong length {} instead of'
+                                     ' {}'.format(len(feature_index),
+                                                  self.features_length))
                 feature_index_list = numpy.argwhere(feature_index
                                                     ).ravel()
             else:
@@ -106,7 +109,7 @@ class BaseFeaturesIndexer(BaseIndexer):
             feature_index_list = [feature_index]
         return feature_index_list
 
-    def create_feature_index_list_by_key(self, feature_index):
+    def create_feature_index_list_key(self, feature_index):
         if isinstance(feature_index, slice):
             raise TypeError('Accessing feature by slice is not supported')
         elif isinstance(feature_index, (list, numpy.ndarray)):
@@ -124,7 +127,7 @@ class BaseFeaturesIndexer(BaseIndexer):
     def __getitem__(self, item):
         self.preprocess(item)
         if self.access_feature_by_key:
-            create_feature_index_list_fn = self.create_feature_index_list_by_key
+            create_feature_index_list_fn = self.create_feature_index_list_key
         else:
             create_feature_index_list_fn = self.create_feature_index_list
         if isinstance(item, tuple):
@@ -156,7 +159,7 @@ class BaseFeaturesIndexer(BaseIndexer):
 
     def _extract_feature(self, data_index, j):
         """Format `data_index` and call proper method to extract feature.
-        
+
         Args:
             data_index (int, slice, list or numpy.ndarray):
             j (int or key):
@@ -182,9 +185,9 @@ class BaseFeaturesIndexer(BaseIndexer):
             if isinstance(data_index[0], (bool, numpy.bool, numpy.bool_)):
                 # Access by bool flag list
                 if len(data_index) != self.dataset_length:
-                    raise ValueError('Feature index wrong length {} instead of '
-                                     '{}'.format(len(data_index),
-                                                 self.dataset_length))
+                    raise ValueError('Feature index wrong length {} instead of'
+                                     ' {}'.format(len(data_index),
+                                                  self.dataset_length))
                 data_index = numpy.argwhere(data_index).ravel()
 
             if len(data_index) == 1:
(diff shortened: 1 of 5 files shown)


niboshi:
Just a comment about the name.
I feel "feature indexer" would be more natural.
It's not so common to use nouns as adjective in plural form. (c.f. feature selection).

corochann:
You mean you think it is better to change the class name?
BaseFeaturesIndexer class should be BaseFeatureIndexer or something?

niboshi:
>You mean you think it is better to change the class name?
BaseFeaturesIndexer class should be BaseFeatureIndexer or something?

Yes, I mean it.
Identifiers such as class names should be natural as English words. Of course, that's not without exception (e.g. if the noun were strongly associated with some API element).

I don't feel discomfort at `train.features` because it's natural as English (it's not used as adjective).

corochann:
I'm not sure how to fix travis build failure.

delta2323:
It seems the CI was halted manually. I kicked the test again.

corochann:
Thank you, seems test is ok.
Then I'd like to get comment of this implementation idea & if it is ok I'd like to get review of this code.

Note that this is spec addition, and I believe no backward compatibility break.

niboshi:
I'm sorry for letting you wait.
Some comments:
* You wrote in the motivation that this interface fits well with sklearn. Could you show how fitting it is (e.g. a link to the corresponding sklearn API documentation or sklearn code example)?
* To me, it's somewhat weird that labels are included in "features". Do you think it's common?
* Will this feature benefit datasets other than `TupleDataset`?

corochann:
A1. For example if we want to train sklearn model, we usually call `model.fit(x_train, y_train)`.
      We need to split `x_train` and `y_train` from dataset. But this is not straight forward in current chainer dataset (maybe we need to call `concat_examples` method which is not well known to begiiners).
      With this PR, we can get it like `x_train, y_train = dataset.features[:, :]`
      This is also beneficial when we want to use preprocessing methods like `StandardScaler` to only `y_train`.

A2. This is important point, that the user can understand the meaning/functionality of this method intuitively.
      While I thought "features" may be intuitive, I want to get better suggestion for naming.

A3. I believe yes. I found it useful in following situations.

  3-1. Extract label from DatasetMixin class
          Getting only label information from `LabeledImageDataset` is not obvious.
          If we access `LabeledImageDataset` by index, the unnecessary image file load is executed.
          We can extract labels by `labels = labeled_image_dataset.features[:, 1]` without loading image file.
https://github.com/chainer/chainer/pull/3252/files#diff-6c0e8ee545ff023e31b77d10976ee576R174
          (User need to override your own `extract_feature` method for their own DatasetMixin class, if they want to optimize the performance)

  3-2. Extract label after split the dataset. 
          Let's say we want to split `dataset`, which is `TupleDataset` instance.
          When we split the dataset with chainer's `split_dataset_random` function,
          `train, val = split_dataset_random(dataset, train_data_size, seed)`
          Now, `train` and `val` are `SubDataset` (subclass of `DatasetMixin`) and it is not obvious how to extract labels information of `train`
           With this PR, we can always extract these feature by
           `train_labels = train.features[:, -1]`.

            So the **unified interface can be used for all types of chainer Dataset** to extract any features.

niboshi:
Thank you!
I understand that the point of this feature is to enable indexing along directions other than the example axis.

How do you think about extending the existing `__getitem__` to allow richer indexing (with more than 1 indexer)?
I think this can be done without breaking backward compatibility.

corochann:
Do you mean since current `__getitem__` is expected only 1-dim accessing, so we can extend it for 2-dim accessing for this feature indexing access?

I am not sure which is less confusing to users.
But since the behavior of default `__getitem__` and `features` indexing(which expands each feature in return), I feel adding another name for indexing is better.


There may be a very few possibility, but you may create another indexer (other than feature indexer) in the future as well.
For example, pandas DataFrame has `icol`, `iloc` and `ix` indexer, other than the default `__getitem__`.
I introduced `BaseIndexer` class in newly added `indexer.py` file, by considering this kind of extensibility.

niboshi:
>Do you mean since current `__getitem__` is expected only 1-dim accessing, so we can extend it for 2-dim accessing for this feature indexing access?

I meant not only 2-dim but to allow arbitrary number of indexers.

If the user gives an indexer of a single axis, the behavior would be the same as the current `__getitem__` (thus backward-compatibility would be kept).
When the user gives 2 or more indexers, the behavior would depend on the dataset implementation and possibly actual data.

That would cover the pandas case you mentioned, I guess.

corochann:
Well, there's misunderstanding in naming teminology.
I'm calling `ix`, `iloc`, `loc` in pandas, or `features` of this PR as indexer.

All of the indexer can take 2-dimensional axis indices, but its behavior is different.

Ref: http://ailaby.com/lox_iloc_ix/ (Japanese)

I wrote that it is possible (in the future if necessary) that we can introduce another indexer, which takes 2-dimensional indices but its behavior is different.

niboshi:
@corochann 
Thank you for pointing out.

I'll discussed about the design with other core members.

corochann:
Onegaishimasu

chainer/dataset/indexer.py
```diff
@@ -0,0 +1,191 @@
...
+
+    @property
+    def shape(self):
+        return self.dataset_length(), self.features_length()
+
+    def extract_feature_by_slice(self, slice_index, j):
+        """Extracts `slice_index`-th data's `j`-th feature.
+
+        Here, `slice_index` is indices of slice object.
+        This method may be override to support efficient feature extraction.
```
beam2d: overridden (same for the next line)

tests/chainer_tests/datasets_tests/test_tuple_dataset.py
```diff
@@ -44,5 +44,207 @@ def test_tuple_dataset_overrun(self):
         with self.assertRaises(IndexError):
             td[3]
 
+    def test_tuple_dataset_features(self):
```
beam2d: Split each test into separate methods.

chainer/dataset/indexer.py
```diff
@@ -0,0 +1,191 @@
...
+
+
+class BaseFeatureIndexer(BaseIndexer):
+
+    """Base class for FeatureIndexer
+
+    FeatureIndexer can be accessed by 2-dimensional indices, axis=0 is used for
+    dataset index and axis=1 is used for feature index.
+    For example, let `features` be the instance of `BaseFeatureIndexer`, then
+    `features[i, j]` returns `i`-th dataset of `j`-th feature.
```
beam2d: I do not fully understand the scope of this feature from this docstring. What is the expected behavior of an indexer attached to a dataset in each of the following cases?

Let `d` be a tuple dataset; then
- When `d` has multiple datasets one of which is again a tuple dataset (i.e., `d[i][j]` is a tuple)
- When `d` has a dataset which has variable shape (i.e., `d[0][0]` and `d[1][0]` has a different shape; consider the case of using variable length text input)

corochann:
Thank you for the review.

>  the current PR implements an indexer for existing datasets, but I feel it's better to make a new dataset which has a compatible API as pandas. What do you think?

This question is related to the discussion done in this thread: https://github.com/chainer/chainer/pull/3252

One of my important request is that I would like to use this feature not only for the specific dataset class but also for extended dataset like `SubDataset`, `ConcatenatedDataset` and `TransformDataset` etc.
Because, the dataset class is automatically changed in following situation (which frequently happens in data science workflow).

`SubDataset`: 
1. When the dataset is splitted into train and validation dataset. 
2. When I want to extract small dataset for debugging purpose.

`ConcatenatedDataset`: 
1. When each time-series data (which has depedency, and cannot splitted into train/validation randomly) is my own dataset class, and want to concatenate (combine) each dataset to form one big dataset.

So, I would like to support this feature for `DatasetMixin` class as well. 
I felt it was easy for users that the unified interface can be used for all types of chainer Dataset to extract features.


If it is better to consider the case to make a new dataset (chainercv develop member is considering it https://github.com/chainer/chainercv/pull/454), then is it possible to support similar interface for `DatasetMixin` (`SubDataset`, `ConcatenatedDataset`) as well?

yuyu2172:
+1 for another `Dataset` class that universally solves all problems.
It looks like a better idea than modifying current dataset classes, which are good by itself for being simple.

On top of what have been mentioned, I would like to re-emphasize that the dataset class should be able to efficiently access subset of data (e.g. `a[:, 1]`), which has been discussed in https://github.com/chainer/chainercv/pull/454.
This is a very common scenario especially in computer vision, where you the time it takes to load data is disproportionate across modality (e.g.image takes very long time, but annotations can be retrieved instantly).
Since `DatasetMixIn` need all related data per index, we need an alternative interface that allows users to write a method to load subset of data.
Taking that into account, to implement the above mentioned feature, it is necessary to change interface from `DatasetMixIn` for the new dataset.

corochann:
It seems it is better to get chainer team's direction before proceeding implementation.

I was feeling this is more like "interface class" (in Java terminology) other than class itself.

@yuyu2172 
> Taking that into account, to implement the above mentioned feature, it is necessary to change interface from DatasetMixIn for the new dataset.

Do you mean `SubDataset`/`ConcatenatedDataset` will be a sub class of newly introduced dataset, to support the interface.?

> On top of what have been mentioned, I would like to re-emphasize that the dataset class should be able to efficiently access subset of data (e.g. a[:, 1]), which has been discussed in chainer/chainercv#454.

Currently I implemented two different function, `exract_feature_by_slice(self, slice_index, j)` and `extract_feature(self, i, j)`.
https://github.com/chainer/chainer/pull/3343/files#diff-81b6edfb9c10c745e45d1990196ec9f6R73
https://github.com/chainer/chainer/pull/3343/files#diff-81b6edfb9c10c745e45d1990196ec9f6R76

I think it can be a reference, but I am not sure there is more simpler solution.

corochann:
[Note]

1. `features` indexer is already implemented in Chainer Chemistry.
https://github.com/pfnet-research/chainer-chemistry/blob/master/chainer_chemistry/datasets/numpy_tuple_dataset.py#L52

2. Since I wanted to use this function but I could not wait until merge, I am PERSONALLY using this `features indexer` by myself by overriding all the dataset class externally.

https://github.com/corochann/chainerex/blob/master/chainerex/dataset/indexers/feature_indexer.py#L493
Usage:
https://gist.github.com/corochann/bafc0b6209f008e50ebe196b00ba56da


3. Please decide direction for dataset API.

Even though this PR has opened on last August and there was no progress in this PR in chainer,
It seems there is on-going PR `SlicableDataset`, which aims to support similar functionality
https://github.com/chainer/chainer/pull/3976
If above PR is merged, maybe this PR becomes redundant.

I hope chainer team to decide direction how to make dataset handling easier for users.
Even above PR #3976  is not updated for two months...

niboshi:
@corochann 
I'm sorry for letting you wait.
We will discuss this PR in the next internal meeting.

corochann:
I think it is ok to prioritize PR #3976

corochann:
Seems that Chainer team will consider another way to provide this functionality.
Please also refer
https://github.com/chainer/chainer/pull/3252

PR #3364, opened 2017-09-10
Use correct module path

okuta:
Current code uses `cupy.cuda.cudnn.cudnn`. We should use `cupy.cuda.cuda.cudnn`.

### DIFF
--- a/chainer/__init__.py
+++ b/chainer/__init__.py
@@ -97,7 +97,7 @@ _SHOULD_USE_CUDNN = {
 }
 
 
-_cudnn_version = cuda.cudnn.cudnn.getVersion() if cuda.cudnn_enabled else -1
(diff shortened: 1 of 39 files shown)


delta2323:
I would like to know the motivation behind this PR. Is it related to #3198?

okuta:
No, It is not related.

okuta:
This PR is related cupy/cupy#512.

niboshi:
Why is this change needed?

okuta:
Because `cupy.cuda.cudnn.cudnn` is wrong.

niboshi:
OK.
jenkins, test this please.

okuta:
jenkins, test this please.

niboshi:
LGTM!

PR #3386, opened 2017-09-14
Use cuDNN BN functions even when inputs data type is fp16

anaruse:
The aim of this PR is to use cuDNN's batch-normalization (BN) functions safely when data type of input is fp16.

The cuDNN's BN functions support fp16 data type, but they are not used in Chainer now when data type of input is fp16. As far as I know, that is mainly because data types of some parameters like `gamma` and `beta` in BN layer must be fp32 in cuDNN's BN functions when data type of inputs is fp16 though Chainer assumes data types of inputs and parameters are the same.

Actually I've already sent a PR about this issue before (#2299). The solution of the previous PR is to decide what data type should be used for parameters like `gamma` and `beta` at the `Link` object, and keep the variables with the decided data type. With the previous PR, in case of fp16 input, the parameters data type become fp32 when cuDNN is selected to use, otherwise fp16. However there are mainly two problems:

- It is less flexible. For example, a use case like using cuDNN for forward and backward and using CuPY for backward of backward is not available with the previous solution.
- It is not easy-to-maintain. Both the `Link` and `Function` class are modified so they should be maintained in sync.

Thus, I would like to propose simple solution with this PR as follows.

- To keep the data type of parameters like `gamma` and `beta` the same as the data type of inputs, even when cuDNN is likely to be used. --> No modification to the `Link` class is necessary and it is easier-to-maintain.
- To convert the data type of parameters from fp16 to fp32 or vice versa on demand before or after calling cuDNN's BN functions. --> It is more flexible and the use case like using cuDNN for forward and backward and using CuPY for backward of backward gets available.

Please note that https://github.com/cupy/cupy/pull/492 needs to be used in order to test this PR.

### DIFF
--- a/chainer/functions/normalization/batch_normalization.py
+++ b/chainer/functions/normalization/batch_normalization.py
@@ -79,11 +79,11 @@ class BatchNormalization(function_node.FunctionNode):
             if dtype_param is not dtype:
                 gamma = gamma.astype(dtype_param)
                 beta = beta.astype(dtype_param)
-                _running_mean = self.running_mean.astype(dtype_param)
-                _running_var = self.running_var.astype(dtype_param)
+                running_mean = self.running_mean.astype(dtype_param)
+                running_var = self.running_var.astype(dtype_param)
             else:
-                _running_mean = self.running_mean
-                _running_var = self.running_var
+                running_mean = self.running_mean
+                running_var = self.running_var
 
             oz_dtype = 'd' if x.dtype == 'd' else 'f'
             one = numpy.array(1, dtype=oz_dtype).ctypes
@@ -108,8 +108,8 @@ class BatchNormalization(function_node.FunctionNode):
                 handle, cudnn_mode, one.data, zero.data,
                 x_desc.value, x.data.ptr, x_desc.value,
                 y.data.ptr, derivedBnDesc.value, gamma.data.ptr,
-                beta.data.ptr, factor, _running_mean.data.ptr,
-                _running_var.data.ptr, self.eps,
+                beta.data.ptr, factor, running_mean.data.ptr,
+                running_var.data.ptr, self.eps,
                 self.mean.data.ptr, self.inv_std.data.ptr)
 
             if dtype_param is not dtype:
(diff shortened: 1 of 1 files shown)


chainer/functions/normalization/batch_normalization.py
```diff
@@ -449,6 +472,22 @@ def _zero_if_none(xp, x, shape, dtype):
     return x
 
 
+def _get_dtype_of_tensor_descriptor(desc):
+    cudnn_dtype, _, _, _, _, _, _, _, _ = libcudnn.getTensor4dDescriptor(
```
beam2d: Use `cudnn_dtype, *_ = ...`, which is easier to read.

chainer/functions/normalization/batch_normalization.py
```diff
@@ -75,8 +75,15 @@ def forward(self, inputs):
...
             libcudnn.deriveBNTensorDescriptor(derivedBnDesc.value,
                                               x_desc.value, cudnn_mode)
-            one = numpy.array(1, dtype=dtype).ctypes
-            zero = numpy.array(0, dtype=dtype).ctypes
+            dtype_param = _get_dtype_of_tensor_descriptor(derivedBnDesc)
+            if dtype_param is not dtype:
+                gamma = gamma.astype(dtype_param)
+                beta = beta.astype(dtype_param)
+                self.running_mean = self.running_mean.astype(dtype_param)
+                self.running_var = self.running_var.astype(dtype_param)
```
beam2d: It looks that the original running mean/var arrays are not updated by `batchNormalizationForwardTraining`. Could you add code that writes back the updated values to the original arrays?

anaruse: Ah, I did not notice that. That is a problem. Will fix codes so that the original arrays of running mean/var arrays are updated appropriately.

chainer/functions/normalization/batch_normalization.py
anaruse: Thanks, will update in that way.

anaruse: Seems this way of writing is not OK with python 2.7?

beam2d: Oh, sorry, I didn't notice that. Then it's ok to leave the current code as is.

beam2d:
Jenkins, test this please

beam2d:
LGTM

beam2d:
Thank you!

PR #3388, opened 2017-09-14
Tensor-Core support for convolutions

anaruse:
This PR adds Tensor-Core support for convolutions.

With this PR, Tensor-Core is automatically used in convolution layers when Tensor-Core is available. You can also manually enable/disable use of Tensor-Core by setting global configuration variable `cudnn_use_tensor_core` to True/False like below

    configuration.config.cudnn_use_tensor_core = [True|False]

Please note that https://github.com/cupy/cupy/pull/494 needs to be used in order to test this PR.
Also I would suggest that you apply #3386 first.

### DIFF
--- a/docs/source/reference/core/configuration.rst
+++ b/docs/source/reference/core/configuration.rst
@@ -67,6 +67,14 @@ Note that the default values are set in the global config.
        - If it is ``'never'``, Chainer will never use cuDNN anywhere.
 
    The default value is ``'auto'``.
+``chainer.config.cudnn_use_tensor_core``
+   Flag to configure whether or not to enable Tensor Core operatons in cuDNN.
+
+       - If it is ``True``, Chainer uses cuDNN's Tensor Core operations.
+       - If it is ``False``, Chainer does not use cuDNN's Tensor Core operations.
+       - If it is ``None``, Chainer checks cuDNN version, the data type of input, the compute capability of the GPU used, and configures whether or not to use cuDNN's Tensor Core operations.
+
+   The default value is ``None``.
 
 Users can also define their own configurations.
 There are two ways:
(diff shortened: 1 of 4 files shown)


beam2d:
The code LGTM. Could you write a brief explanation of `chainer.config.cudnn_use_tensor_core` to `docs/source/reference/core/configuration.rst`?

beam2d:
Jenkins, test this please

beam2d:
Jenkins, test this please

beam2d:
CI passed. I will merge after the documentation is made ready.

anaruse:
@beam2d Thank you for your review. 
I've just added the brief explanation of `chainer.config.cudnn_use_tensor_core` to `docs/source/reference/core/configuration.rst`.

docs/source/reference/core/configuration.rst
```diff
@@ -67,6 +67,14 @@ Note that the default values are set in the global config.
        - If it is ``'never'``, Chainer will never use cuDNN anywhere.
 
    The default value is ``'auto'``.
+``chainer.config.cudnn_use_tensor_core``
+   Flag to configure whether or not to enable Tensor Core operatons in cuDNN.
+
+       - If it is ``True``, Chainer uses cuDNN's Tensor Core operations.
```
beam2d: Could you use the values consistent with other options (e.g. `use_cudnn`)? I think it can be automatically mapped like `True -> 'always'`, `False -> 'never'`, and `None -> 'auto'`.

anaruse: Thanks for your comment. I like consistency, so I think your suggestion is better.
I just updated the branch in that way, also the option name was changed to `use_cudnn_tensor_core` considering consistency.

beam2d:
LGTM

beam2d:
Thank you!

PR #3394, opened 2017-09-15
[WIP] Rename `xxx_pooling_nd` to `xxx_pooling`

niboshi:
- [ ] `max_pooling_nd`
- [ ] `average_pooling_nd`
- [ ] `unpooling_nd`

* Note: original functions `xxx_pooling_nd` should be kept for backward compatibility.


delta2323:
Could you let me know the motivation for this PR?

niboshi:
Some of the members discussed in slack that `_pooling` is better than `_pooling_nd`, while existing APIs should not be altered. (https://chainer.slack.com/archives/C5P26TUR1/p1505385954000094)

It's not final decision, though.

delta2323:
Thank you for pointing out the resource. I missed the discussion.

niboshi:
As we have internally discussed this problem again and concluded that we're going to make additional aliases (such as `max_pooling_1d` and `max_pooling_3d`) rather than unifying, I'm closing this PR.

PR #3396, opened 2017-09-15
Serialize trigger in LogReport

Hakuyume:
Some triggers including `IntervalTrigger` have `serialize` method. `LogReport` should call this method in serialization.


rezoo:
OK. LGTM.

PR #3399, opened 2017-09-15
Add a missing device argument to the .to_gpu() call in the to_gpu() method.

drpnd:
This PR adds a missing `device` argument for .to_gpu() calls in the to_gpu() method in `chainer/link.py`.

The following is a case affected by this bug.
```
cnn = MyCNN() # My neural network model
model = chainer.links.Classifier(cnn)
model.to_gpu(0)
```
If I understand correctly, this code is assumed to call `.to_gpu()` method of the `cnn` instance with the `device` argument (`0` in this case) through the instance `model` of chainer.links.Classifier.  However, it actually calls `cnn.to_gpu()` without any arguments.

### DIFF
--- a/chainer/link.py
+++ b/chainer/link.py
@@ -367,7 +367,7 @@ Assign a Parameter object directly to an attribute within a \
         d = self.__dict__
         with cuda._get_device(device):
             for name in self._params:
-                d[name].to_gpu()
+                d[name].to_gpu(device)
             for name in self._persistent:
                 value = d[name]
                 if isinstance(value, numpy.ndarray):


rezoo:
It seems that we do not need to specify the device id because `with cuda._get_device(device)` context implicitly specifies the default argument of `to_gpu()`. In fact, in my environment the following code works well:

```python
import chainer
import chainer.links as L

class MyCNN(chainer.Chain):
    
    def __init__(self):
        super(MyCNN, self).__init__()
        with self.init_scope():
            self.l = L.Linear(100, 10)

    def __call__(self, x):
        return self.l(x)

chainer.cuda.get_device_from_id(0).use()

cnn = MyCNN()
model = chainer.links.Classifier(cnn)
cnn.to_gpu(1)
print(cnn.l.W.data.device)  # <CUDA Device 1>
```

Are there any other reasons to explicitly specify the argument?

rezoo:
I think this PR should be closed since there is no particular objection.
Please feel free to resubmit it if you have any problems.

Issue #2854, opened 2017-06-13
Test failure on test_spatial_transformer_sampler.TestSpatialTransformerSampler


niboshi:
https://travis-ci.org/chainer/chainer/builds/241979452?utm_source=github_status&utm_medium=notification
(Occured in a PR based on `v2` branch. Also confirmed in `master` branch.)

~Looks like this is numeric unstability due to non-deterministic behavior of `cudnnSpatialTfSamplerBackward` function.~ This error has occurred in cpu mode.

```
FAIL: test_backward_cpu (test_spatial_transformer_sampler.TestSpatialTransformerSampler_param_1)  parameter: {'use_cudnn': False}
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/chainer/testing/condition.py", line 74, in wrapper
    fail()
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/chainer/testing/condition.py", line 51, in fail
    instance.fail(msg)
AssertionError: 
Fail: 3, Success: 0
The first error message:
Traceback (most recent call last):
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/chainer/testing/condition.py", line 60, in <lambda>
    lambda: f(ins, *args[1:], **kwargs),
  File "/home/travis/build/chainer/chainer/tests/chainer_tests/functions_tests/array_tests/test_spatial_transformer_sampler.py", line 81, in test_backward_cpu
    self.check_backward(self.x, self.grid, self.grads)
  File "/home/travis/build/chainer/chainer/tests/chainer_tests/functions_tests/array_tests/test_spatial_transformer_sampler.py", line 77, in check_backward
    (x, grid), (grads,), atol=1e-2, rtol=1e-2, eps=1e-5)
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/chainer/gradient_check.py", line 261, in check_backward
    testing.assert_allclose(gx, x.grad, atol=atol, rtol=rtol)
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/chainer/testing/array.py", line 24, in assert_allclose
    x, y, atol=atol, rtol=rtol, verbose=verbose)
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/numpy/testing/utils.py", line 1297, in assert_allclose
    verbose=verbose, header=header)
[...]

PR #3401, opened 2017-09-15
[New feature] Using a class instead of a function in the snapshot extension.

tyohei:
Due to the performance disadvantage of taking snapshots, I had implemented a new snapshot extension.
Which has a feature of:

- Modularity
- Efficiency

At a GPU-installed environment, training the MNIST example (`train_mnist.py`),  the whole training time of 16 epochs, taking snapshot every epoch, took (only experimented once):

- 77 sec with original snapshot
- 50 sec with `ThreadWriter`
- 51 sec with `ProcessWriter`
- 53 sec with `ThreadQueueWriter`
- 48 sec with `ProcessQueueWriter`
- 46 sec with NO snapshots

### Interface
The interface will be almost same as the current snapshot extension.
For example, in the MNIST example ( `train_mnist.py` ), just change like following.

```python
trainer.extend(extensions.Snapshot(), trigger=(frequency, 'epoch'))
```
(For the default arguments, the difference is the uppercase "S" only.)
The `Snapshot` is the extension class (inherits `extension.Extension`) and takes `Writer` object and `Condition` object as arguments.

### Writer
If you'd like to have more efficient snapshotting, using threading or multiprocessing modules, you can do it by changing `Writer` object (code is written in `writer.py`).
```python
from chainer.training.extensions.snapshots import writer as writer_module
writer = writer_module.ProcessWriter()
trainer.extend(extensions.Snapshot(writer=writer), trigger=(frequency, 'epoch'))
```
To change the saving format, such like npz or hdf5, you can do it by passing the function to the writer's keyword argument.
```python
from chainer.training.extensions.snapshots import writer as writer_module
from chainer.training.extensions.snapshots import util
savefun = util.save_npz
writer = writer_module.ProcessWriter(savefun=savefun)
trainer.extend(extensions.Snapshot(writer=writer), trigger=(frequency, 'epoch'))
```

### Condition
Also there is a `Condition` object that determines whether a snapshot should be taken or not. This is similar to the `trigger` attribution, but it is more flexible, and can be easily adapted in ChainerMN.


tYohei:
Due to Cython error https://github.com/cython/cython/issues/1880, I will close this PR and open it again.

niboshi:
Thank you for a great PR, and I'm sorry for letting you wait.
* The whole things should be implemented in `chainer.training.extensions._snapshot` (or perhaps its submodules). Plural form (e.g. `snapshots`) are meant for example instances, such as `chainer.functions.linear`.
* I think this feature should also be added in `_snapshot.snapshot()` and `_snapshot.snapshot_object()`. Backward compatibility can be kept by using `SimpleWriter`.

chainer/training/extensions/snapshots/writer.py
```diff
@@ -0,0 +1,223 @@
...
+
+    def __init__(self, savefun=util.save_npz, task=None):
+        if task is None:
+            self._task = SimpleWriter(savefun)
+        else:
+            self._task = task
+        self._queue = queue.Queue()
+        self._consumer = threading.Thread(target=self.consume,
+                                          args=(self._queue,))
+        self._consumer.start()
```
niboshi: It's not good to access base class'es private attributes (those starting with `_`).
How about making them public interface?
For example, adding
```
def create_queue(self):
    raise NotImplementedError()
```
in the base class and override it in the concrete classes.

And I think the construction logic can be implemented in the base class'es `QueueWriter.__init__()`, i.e.
```
if task is None:
    self._task = self.create_task(savefun)
else:
    self._task = task
self._queue = self.create_queue()
self._consumer = self.create_consumer()
self._consumer.start()
```

chainer/training/extensions/snapshots/writer.py
```diff
@@ -0,0 +1,223 @@
...
+        self._kwds = kwds
+        self._flag = False
+
+    def __call__(self, filename, outdir, target):
+        self._thread = threading.Thread(
+            target=self.save,
+            args=(filename, outdir, target, self._savefun),
+            kwargs=self._kwds)
+        self._thread.start()
+        self._flag = True
```
niboshi: The previous write operation is not guaranteed to be completed, which can lead to unstable behavior if `filename` is not changed between calls. How about joining the previous thread?

chainer/training/extensions/snapshots/writer.py
```diff
@@ -0,0 +1,223 @@
...
+        self._kwds = kwds
+        self._flag = False
+
+    def __call__(self, filename, outdir, target):
+        self._process = multiprocessing.Process(
+            target=self.save,
+            args=(filename, outdir, target, self._savefun),
+            kwargs=self._kwds)
+        self._process.start()
+        self._flag = True
```
niboshi: Ditto. How about joining the previous process?

niboshi:
Also, please write some unit test cases.

tYohei:
Thank you for the reply and some comments to my code.

I will change the code according to your comments, and write some unit tests.

tYohei:
Due to the an error (perhaps: https://github.com/chainer/chainer/issues/2854) I will close and reopen this PR.

tYohei:
Sorry for being late.

I had changed some codes and written some unit tests.
To follow the naming rule in the Chainer, I had moved `chainer/chainer/training/extensions/snapshots/writer.py` to `chainer/chainer/training/writer.py`.
 Also, `condition.py` and `util.py` are removed.

niboshi:
These writers are meant only for snapshots, right?
If so, these classes should be under `_snapshot` or its descendants.
How do you think?

niboshi:
I'm sorry for a delay.
As you pointed out, users should not be forced to use `_snapshot` module.
How about making writer classes as attributes of `Snapshot` class? i.e. users will be access writer classes like `extensions.Snapshot.ThreadWriter()`.

This can be done like this:
```
<_snapshot.py>

class ThreadWriter:
     ...

class Snapshot:
    ThreadWriter = ThreadWriter
    ...
```

niboshi:
@tyohei
Can you afford time on this PR? If you can't, we would like to take it over.
Thank you!

niboshi:
@tyohei 
I made a new PR to take over this one, so I'm closing it now.
We really appreciate your work.
Please feel free to give us any comment or feedback if you have any.
Thanks!

PR #3411, opened 2017-09-17
Simplify clip and clipped_relu

okuta:
`clip` covers `clipped_relu`.

### DIFF
--- a/chainer/functions/activation/clipped_relu.py
+++ b/chainer/functions/activation/clipped_relu.py
@@ -1,11 +1,10 @@
-from chainer.backends import cuda
-from chainer import function_node
-from chainer import utils
-from chainer.utils import type_check
-import numpy
+import warnings
+
+import chainer.functions
+from chainer.functions.math import clip
 
 
-class ClippedReLU(function_node.FunctionNode):
+class ClippedReLU(clip.Clip):
 
     """Clipped Rectifier Unit function.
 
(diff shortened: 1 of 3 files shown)


delta2323:
Should we treat this PR as `to-be-backported`?

okuta:
No. Because `ClippedReLU` is deleted in this PR.

niboshi:
Can we keep `ClippedReLU` as a thin wrapper or something? Otherwise it would break backward compatibility.

okuta:
This change do not break backward compatibility based on Chainer's policy.
And, almost users do not use ClippedReLU directly.

niboshi:
What do you mean?
In general, any identities not starting with `_` are considered "public" in the python convention, whether they're documented or not.

okuta:
I read [this section](https://docs.chainer.org/en/stable/compatibility.html#documentation-error).
But, If this fixing is not user friendly, I propose a different way.

niboshi:
I see.
Let's discuss with other members.

delta2323:
IMO, I agree with keeping `ClippedReLU` if it is not costly and does not make the implementation complicated, as we do not need to dare to break compatibility.

hvy:
@beam2d do you have any comments on this discussion?

I don't have any strong opinion but I see why we'd like to keep the class as it is. The maintenance cost is not that high and [`ClippedReLU`](https://github.com/chainer/chainer/blob/master/chainer/functions/activation/clipped_relu.py) has been reimplemented as a `FunctionNode`.  As for performance, both `clipped_relu` and `clip` are implemented with "optimized" grad computations so I'd image they are quite similar.

okuta:
I updated.

delta2323:
ping @beam2d

beam2d:
Sorry for being late. I don't have any strong opinion on the backward compatibility of `ClippedReLU` class for the next major update, but since this name is imported to `chainer.functions`, I think it's safer to leave it. The implementation can be changed; e.g. `ClippedReLU` can be easily implemented by inheriting `Clip`. We can raise `DeprecationWarning` so that we will eventually remove them.

Btw, I think we should remove all the Function/FunctionNode implementations from `chainer/functions/__init__.py` at some point, and eventually make them look like not supported backward compatibility. (we may also need to rename all of them by prefixing by `_`).

hvy:
@okuta san, do you have any opinions on this? 

I discussed with @beam2d san offline about the details and we said that maybe the class could be added back, throwing a warning in its `__init__` and we make sure to test both the new implementation and the deprecated (temporary) class implementation for the time being. I am not sure though when we should remove the warning/class completely.

okuta:
OK, I will use `Clip`.

okuta:
I fixed.

hvy:
Sorry, could you rebase on the current master?

okuta:
#4307 blocks this PR. So, I cannot rebase or merge this PR.
Is this PR really necessary?
In other words, I think it is necessary to discuss which functions of cuDNN to support.

hvy:
I think it is okay to close this PR if that's okay with the PR author (@okuta) .

As for cuDNN function support, do we not under most circumstances want to support as many as possible?

okuta:
OK, I close this PR.

PR #3413, opened 2017-09-18
[backport] Improve performance of MultiprocessIterator for non tuple/dict datasets

okuta:
This is a backport of #3390


niboshi:
LGTM!

PR #3419, opened 2017-09-18
New style copy function

okuta:
Please see #3147.

### DIFF
--- a/chainer/functions/array/copy.py
+++ b/chainer/functions/array/copy.py
@@ -1,9 +1,9 @@
 from chainer import cuda
-from chainer import function
+from chainer import function_node
 from chainer.utils import type_check
 
 
-class Copy(function.Function):
+class Copy(function_node.FunctionNode):
 
     """Copies the input variable onto the specified device."""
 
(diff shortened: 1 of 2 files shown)


chainer/functions/array/copy.py
```diff
@@ -15,41 +15,17 @@ def check_type_forward(self, in_types):
             in_types.size() == 1
         )
 
-    def forward_cpu(self, x):
-        self.retain_inputs(())
-        self._in_device = cuda.get_device_from_array(x[0])
+    def forward(self, inputs):
+        x, = inputs
+        self._in_device = cuda.get_device_from_array(x).id
+        print(type(x), self._in_device, self.out_device)
```
niboshi: Remove `print`.

chainer/functions/array/copy.py
```diff
@@ -83,4 +59,6 @@ def copy(x, dst):
         -1
 
     """
-    return Copy(dst)(x)
+    assert isinstance(dst, int)
```
niboshi: Shouldn't it accept `Device` as `dst`?

okuta: Maybe, only accept `int`.
https://github.com/chainer/chainer/pull/3419/files#diff-1b7bc2d145846eadbc5d0f8a8fd90c20R22

niboshi: Is there any reason?
IMO, it would be more convenient for user if it accepts `Device` as well.

okuta: This is my mistake.
I added test. and I removed this assert.

okuta:
I fixed.

niboshi:
Please add double-backward tests.

niboshi:
LGTM!

PR #3427, opened 2017-09-19
Stabilize unit tests for F.Ceil and F.Floor

delta2323:
This PR changes the distribution from which inputs are sampled in TestCeil so that tests for the function are likely to pass more. Specifically, we sample less likely from discontinuous points (i.e. integer points).


niboshi:
How about sampling repeatedly until an input with no value near integer point is obtained?
Tests must pass 100% of time.

niboshi:
Also, please consider removing `retry`.

niboshi:
`test_floor` has the same problem. Can you fix it too?

delta2323:
I updated to the tests of `F.Ceil` and applied the same change to `F.Floor`. Also I change the PR title accordingly.

niboshi:
Thanks!

niboshi:
LGTM!

Issue #2860, opened 2017-06-14
Need a data sampler for unevenly distributed labels


apple2373:
We need a way to handle a dataset whose label distribution is highly skewed. For example, when we have 1000 positives and 100 negatives, we want to make sure each batch contains the same number of positives and negatives, oversampling negative examples 10 times more than positives. 

Someone said pytorch has a way to do:
Adding a `sampler` option for iterator as PyTorch can solve the problem I guess.
http://pytorch.org/docs/data.html#torch.utils.data.DataLoader
Source: https://chainer.slack.com/archives/C0LC5A6C9/p1497343348496751
Maybe it’s possible for chainer to have a wrapper for these pytoarch data preprocessing utilities. DataLorder is not involved in gradient computation so it should be easy and take much less time than implementing equivalent functions from scratch.

hvy:
Not directly addressing the same issue, but related. https://github.com/chainer/chainer/pull/3429

hvy:
Bump.

beam2d:
#3429 looks related.

apple2373:
Thank you @mitmul ! Just to clarify I already have my own (dirty coded) data loader so i am not looking for the solution now. But yes, the BalancedSerialIterator is what I was looking for when I originally created the issue. I wonder if it's a best way to keep it only for Chainer Chemistry. I think class imbalance problem appears everywhere if you are working on real world problems... 

PR: #3429  is also related but not exactly what I was looking for because I still need to code sampler by myself. 

Probably the best solution is to provide both sampler option (i.e., PR: #3429) and then also provide some default sampler for class imbalance problem (i.e., a sampler equivalent to BalancedSerialIterator) ?

corochann:
Thank you @apple2373 for the feedback.
I implemented `BalancedSerialIterator`, and I agree with your opinion.

I commented https://github.com/chainer/chainer/pull/3429

> After this PR merged, I think it is nice to introduce BalancedOrderSampler rather than BalancedSerialIterator (used in Chainer Chemistry until now) so that the sampler can be reused among any kinds of Iterators.

hvy:
Fyi, #3429 is now merged. It would be great if we could provide some sort of balanced sampler.

PR #3429, opened 2017-09-19
Add order_sampler option to Iterators

yuyu2172:
# Background
The current interface for iterators can not customize the order of indices.
This becomes problemetic when some kind of constraints need to be satisfied among indices of a batch.
For instance, there is no way to prevent a batch from containing more than two indices with the same value. In the current implementation, this can happen when a batch consists of indices from the end of a order and the beginning of the next order.

There needs to be an extra abstraction for sampling indices.
Initially, I looked into the way PyTorch does this. https://github.com/pytorch/pytorch/blob/master/torch/utils/data/dataloader.py
They seem to get indices returned by an object called `batch_sampler`.
It looked nice, but after working on this direction for a while ([prototype](https://github.com/yuyu2172/chainer/tree/iterator-sampler)), I found that the functionalities of Chainer's iterators and PyTorch's DataLoader are a bit different.

In Chainer, we need to support serialization of the iterator.
This difference is significant because it requires the serialized iterator to recover the state any time, and we need to impose an extra interface requirements for `batch_sampler`.
Namely, we need the `batch_sampler` to pass all information that is necessary to recover its state as well as indices to Iterator.

# What this PR is
I found that it is much more simple to customize a function to generate an order of indices to iterate.
In this approach, we assume the function to satisfy the following interaface:
```python
# current_order: An 1-D array of indices.
# current_position: An integer indicating where the iterator is.
# next_order: An 1-D array of indices.
def order_sampler(current_order, current_position):
     ...
    return next_order
```
I designed the interface so that all the state of an iterator is passed to the function.
Passing `order` back and forth is nice because it requires little change in codebase, and there is no backward-incompatibility.


chainer/iterators/multiprocess_iterator.py
```diff
@@ -48,13 +50,21 @@ class MultiprocessIterator(iterator.Iterator):
         n_prefetch (int): Number of prefetch batches.
         shared_mem (int): The size of using shared memory per data.
             If ``None``, size is adjusted automatically.
+        order_sampler (callable): A callable that generates the order
+            of the indices to sample in the next epoch when a epoch finishes.
+            This function should take two arguements: the current order
+            and the current position of the iterator.
+            This should return the next order, which should have the
+            same length as the dataset.
```
delta2323: Do you think we impose that the returned value should not have duplicated values?

yuyu2172: I do not think it is necessary to prohibit duplicated values.
First of all, the rest of the code works without this constraint.
Second, it becomes necessary to support duplicate values in some application. For instance, in a situation where a user wants to sample some indices in a biased way.

hvy:
@yuyu2172 Sorry for the delayed response. I and @delta2323 if he has time will continue with the review as we discussed this PR within the team today and concluded that this feature should be added. For the record.

hvy:
Since it's been a while, do you mind rebasing?

hvy:
For the record, talked offline with @yuyu2172 to continue working on this PR. 

As suggested in https://github.com/chainer/chainer/issues/2860#issuecomment-380281965 , I think a good approach is to extend the interface as already suggested while also including a single or a minimum set of concrete samplers. Other samplers could be introduced in separate PRs.

yuyu2172:
@hvy 
Can you check this PR?

hvy:
Thanks, will take a look 👍

hvy:
jenkins, test this please.

chainer/iterators/order_samplers.py
```diff
@@ -0,0 +1,46 @@
+import numpy
+
+
+class ShuffleOrderSampler(object):
```
hvy: Let's add class documentation (maybe you were going to later?) describing also that it's called once per epoch etc.

chainer/iterators/order_samplers.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def __init__(self, seed=None):
+        # Use a distinct RandomState in the thread
+        # for deterministic random number generation.
+        # To support 32-bit platform and numpy < 1.11,
+        # the seed is taken in a verbose manner.
+        if seed is None:
+            seed = numpy.asscalar(
+                numpy.random.randint(-(1 << 31), 1 << 31, 1).astype('uint32'))
```
hvy: Can this be simplified to `seed = numpy.random.randint(2**32)` ?

chainer/iterators/order_samplers.py
```diff
@@ -0,0 +1,46 @@
...
+        self._random = numpy.random.RandomState(seed)
+
+    def __call__(self, current_order, current_position):
```
hvy: Do you have an example on how `current_position` and anything but the length of `current_order` is going to be used?

chainer/iterators/order_samplers.py
```diff
@@ -0,0 +1,46 @@
...
+
+        Returns:
+            numpy.ndarray: 1-D array of indices. This is the order in which
+                examples are sampled from a dataset in the next epoch.
+
+        """
+        return self._random.permutation(len(current_order))
+
+
+def no_shuffle_order_sampler(current_order, current_position):
```
hvy: Since this sampler is exposed as a part of the public API, how about using a class instead to be consistent with `ShuffleOrderSampler`?

chainer/iterators/serial_iterator.py
```diff
@@ -30,15 +33,34 @@ class SerialIterator(iterator.Iterator):
...
-    def __init__(self, dataset, batch_size, repeat=True, shuffle=True):
+    def __init__(self, dataset, batch_size,
+                 repeat=True, shuffle=True, order_sampler=None):
         self.dataset = dataset
         self.batch_size = batch_size
         self._repeat = repeat
         self._shuffle = shuffle
 
+        if self._shuffle and order_sampler is not None:
+            warnings.warn('`shuffle` is `True` and a custom '
```
hvy: I think we can be more strict and simply raise an error, since we don't need this for backward compatibility?

tests/chainer_tests/iterators_tests/test_serial_iterator.py
```diff
@@ -205,7 +216,8 @@ def test_iterator_not_repeat_not_even(self):
 
     def test_iterator_shuffle_divisible(self):
         dataset = list(range(10))
-        it = iterators.SerialIterator(dataset, 10)
+        it = iterators.SerialIterator(dataset, 2, shuffle=self.shuffle,
```
hvy: Any particular reason the minibatch size is changed?

tests/chainer_tests/iterators_tests/test_serial_iterator.py
```diff
@@ -308,4 +329,68 @@ def test_iterator_serialize_backward_compat(self):
...
+
+    def test_invalid_order_sampler(self):
+        dataset = [1, 2, 3, 4, 5, 6]
+
+        with self.assertRaises(ValueError):
+            it = iterators.SerialIterator(
+                dataset, 6, shuffle=False,
+                order_sampler=InvalidOrderSampler())
+            it.next()
+            it.next()
```
hvy: I don't think this line is executed/needed since `_order` is first set in the constructor (`reset()`).

docs/source/reference/iterators.rst
```diff
@@ -17,3 +17,18 @@ Chainer provides some iterators that implement typical strategies to create mini
    chainer.iterators.SerialIterator
    chainer.iterators.MultiprocessIterator
    chainer.iterators.MultithreadIterator
+
+
+Order sampler examples
+======================
+
+An Iterator iterate over a dataset according to ``order``, which is 1-D array of the indices of a dataset.
+Order samplers are functions that are called inside of an iterator to generate an order.
```
hvy: Just some suggestion regarding the English, how about `An Iterator iterates over a dataset according to an order represented by an 1-D array of indices. Order samplers are callables that are used by those iterators to generate this array.` ?

tests/chainer_tests/iterators_tests/test_serial_iterator.py
yuyu2172: Oh. This is my mistake.

tests/chainer_tests/iterators_tests/test_serial_iterator.py
yuyu2172: Thank you.

chainer/iterators/serial_iterator.py
```diff
@@ -107,22 +133,25 @@ def serialize(self, serializer):
...
-        else:
-            self._order = None
-
         self.current_position = 0
         self.epoch = 0
         self.is_new_epoch = False
 
         # use -1 instead of None internally.
         self._previous_epoch_detail = -1.
+        self._order = self.order_sampler(numpy.arange(len(self.dataset)), 0)
```
hvy: I'm starting to wonder if we can get rid of the `NoShuffleOrderSampler` class completely if we change this logic to e.g. `self._order = self.order_sampler(...) if self.order_sampler is not None else None`. Would that be possible you think?

docs/source/reference/iterators.rst
```diff
@@ -17,3 +17,18 @@ Chainer provides some iterators that implement typical strategies to create mini
    chainer.iterators.SerialIterator
    chainer.iterators.MultiprocessIterator
    chainer.iterators.MultithreadIterator
+
+
+Order sampler examples
+======================
+
+An Iterator iterates over a dataset according to an order represented by an 1-D array of indices.
```
hvy: Sorry this was my mistake but it should be `a 1-D array`.

chainer/iterators/multiprocess_iterator.py
```diff
@@ -48,6 +50,13 @@ class MultiprocessIterator(iterator.Iterator):
         n_prefetch (int): Number of prefetch batches.
         shared_mem (int): The size of using shared memory per data.
             If ``None``, size is adjusted automatically.
+        order_sampler (callable): A callable that generates the order
+            of the indices to sample in the next epoch when a epoch finishes.
+            This function should take two arguements: the current order
+            and the current position of the iterator.
+            This should return the next order. The size of the order
+            should remain constant.
+            This option can not be used when ``shuffle`` is ``True``.
```
hvy: I think `cannot` is more common.

hvy:
We don't have to provide a full fledge example but how about adding a test with a dummy sampler that makes use of both arguments?

hvy:
Some comments after offline discussions with @beam2d : 
- It looks good in general! 
- One suggestion though (with respect to the default behavior and in terms of backward compatibility): How about changing the default value of `shuffle` to `None`. If it is `None` and no sampler is given, then it behaves as if `shuffle=True`. Then we may raise an error in case `shuffle` is given any non-`None` value along along with a sampler. This probably is more intuitive to the user since no argument it silenced.

chainer/iterators/serial_iterator.py
yuyu2172: Yes. This simplifies the interface of OrderSampler

docs/source/reference/iterators.rst
yuyu2172: Right. Thanks.

chainer/iterators/multiprocess_iterator.py
yuyu2172: Right. Thanks.

yuyu2172:
> One suggestion though (with respect to the default behavior and in terms of backward compatibility): How about changing the default value of shuffle to None. If it is None and no sampler is given, then it behaves as if shuffle=True. Then we may raise an error in case shuffle is given any non-None value along along with a sampler. This probably is more intuitive to the user since no argument it silenced.

Nice.

yuyu2172:
> We don't have to provide a full fledge example but how about adding a test with a dummy sampler that makes use of both arguments?

Done.
I fixed everything suggested.

hvy:
Thanks a lot for the quick fixes! Would you mind fixing the CI failures also?

yuyu2172:
Sorry. Fixed.

corochann:
I just noticed this PR, and I basically vote it.

How about adding `BaseOrderSampler` class to define the interface/method that any OrderSampler class should have, so that it is more easier to implement/maintain other type of OrderSampler?
I guess `def __call__(self, current_order, current_position):` is the interface requirement.


--- I think below discussion can be done by other separated PR as well. ---

I also think OrderSampler class also should have `serialize` method so that it can be serialized/deserialized by iterator.
I implemented `BalancedSerialIterator` which is for imbalanced data, it have internal iterator which must be serialized.
https://github.com/pfnet-research/chainer-chemistry/blob/master/chainer_chemistry/iterators/balanced_serial_iterator.py

After this PR merged, I think it is nice to introduce `BalancedOrderSampler` rather than `BalancedSerialIterator` (used in Chainer Chemistry until now) so that the sampler can be reused among any kinds of Iterators.

yuyu2172:
> I also think OrderSampler class also should have serialize method so that it can be serialized/deserialized by iterator.

Is serializing `order` and  `current_position` not enough for your purpose?

> How about adding BaseOrderSampler class to define the interface/method that any OrderSampler class should have, so that it is more easier to implement/maintain other type of OrderSampler?

@hvy 
What do you think? I think it is ok.

corochann:
> I also think OrderSampler class also should have serialize method so that it can be serialized/deserialized by iterator.

When considering other `OrderSampler` in the future, I think it is nice that Sampler itself can have internal state.

For concrete example, below `BalancedSerialIterator` have internal `IndexIterator` which needs to be serialized. I guess when considering to introduce "`BalancedSampler`", we need this kind of internal state to be serialized.
https://github.com/pfnet-research/chainer-chemistry/blob/master/chainer_chemistry/iterators/balanced_serial_iterator.py#L132-L134

yuyu2172:
> --- I think below discussion can be done by other separated PR as well. ---

I think this discussion should be done in another PR.
There other PRs that depend on this PR, and I think this should be merged quickly.

corochann:
I agree sorry.

hvy:
jenkins, test this please.

hvy:
I see you added the base class which is great I think. It also helps as documentation. The implementation  LGTM but I'll take a final look at the tests.

hvy:
I agree that the serialization can be discussed in a separate issue when it becomes relevant, whether that's now or later. It should be possible to build it on top of this.

docs/source/reference/iterators.rst
```diff
@@ -17,3 +17,18 @@ Chainer provides some iterators that implement typical strategies to create mini
...
+An Iterator iterates over a dataset according to an order represented by a 1-D array of indices.
+Order samplers are callables that are used by those iterators to generate this array.
+
+
+.. autosummary::
+    :toctree generated/
+    :nosignatures
+
+    chainer.iterators.OrderSampler
+    chainer.iterators.ShuffleOrderSampler
```
hvy: It looks like these links are broken. Could you take a look?

chainer/iterators/order_samplers.py
```diff
@@ -0,0 +1,57 @@
...
+    """
+
+    def __init__(self, seed=None):
+        # Use a distinct RandomState in the thread
+        # for deterministic random number generation.
+        # To support 32-bit platform and numpy < 1.11,
+        # the seed is taken in a verbose manner.
+        if seed is None:
+            seed = numpy.random.randint(2**31)
+        self._random = numpy.random.RandomState(seed)
```
hvy: Are you still able to train with shuffle using the global numpy rng? I think this changes the default behavior actually for users who used to fix their seeds and then train with shuffle. How about instead of the seed accept a RandomState, and then use the global state by default?

tests/chainer_tests/iterators_tests/test_serial_iterator.py
```diff
@@ -308,4 +329,93 @@ def test_iterator_serialize_backward_compat(self):
...
+        def order_sampler(order, cur_pos):
+            return numpy.repeat(numpy.arange(3), 2)
+        self.options = {'order_sampler': order_sampler}
+
+    def test_iterator_repeat(self):
+        dataset = [1, 2, 3]
+        it = iterators.SerialIterator(dataset, 2, **self.options)
+        for i in range(3):
+            self.assertEqual(it.epoch, i)
+            self.assertAlmostEqual(it.epoch_detail, i + 0 / 6)
```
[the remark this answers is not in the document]
yuyu2172: I think the current style is better because the relationship with other lines are obvious.

chainer/iterators/order_samplers.py
yuyu2172: I basically copied the existing code.
Am I missing something?

https://github.com/chainer/chainer/blob/master/chainer/iterators/multiprocess_iterator.py#L288

docs/source/reference/iterators.rst
yuyu2172: Oh. My bad.

chainer/iterators/order_samplers.py
yuyu2172: Well, that is only the case for `MultiprocessIterator`.
I will fix the code.

hvy:
Jenkins, test this please.

tests/chainer_tests/iterators_tests/test_serial_iterator.py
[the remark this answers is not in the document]
hvy: I see, then let's leave it as is!

chainer/iterators/order_samplers.py
```diff
@@ -0,0 +1,58 @@
...
+
+    Args:
+        random_state (numpy.random.RandomState): Pseudo-random number
+            generator.
+
+    """
+
+    def __init__(self, random_state=None):
+        if random_state is None:
+            random_state = numpy.random.RandomState(
```
hvy: You don't think we should default to the global random state, i.e. `numpy.random.random.__self__` as motivated here https://github.com/chainer/chainer/pull/3429#discussion_r184279908 ?

hvy: For reference, that's also the suggested behavior in this similar PR https://github.com/chainer/chainer/pull/4187/files#diff-f6c6225c19f136dbb780d0eec66a6963R47

chainer/iterators/serial_iterator.py
```diff
@@ -107,22 +135,29 @@ def serialize(self, serializer):
...
-
         self.current_position = 0
         self.epoch = 0
         self.is_new_epoch = False
 
         # use -1 instead of None internally.
         self._previous_epoch_detail = -1.
+        if self.order_sampler:
+            self._order = self.order_sampler(
+                numpy.arange(len(self.dataset)), 0)
```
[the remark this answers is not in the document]
yuyu2172: Yes. Thank you for clarifying.

chainer/iterators/order_samplers.py
yuyu2172: Thanks.

docs/source/reference/iterators.rst
```diff
@@ -17,3 +17,18 @@ Chainer provides some iterators that implement typical strategies to create mini
...
+
+Order sampler examples
+======================
+
+An Iterator iterates over a dataset according to an order represented by a 1-D array of indices.
+Order samplers are callables that are used by those iterators to generate this array.
+
+
+.. autosummary::
+    :toctree generated/
```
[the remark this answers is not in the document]
yuyu2172: Thanks

hvy:
jenkins, test this please.

hvy:
jenkins, test this please.

hvy:
@yuyu2172 It seems like the docs are failing. Could you please take a look when you have time 🙇

hvy:
jenkins, test this please.

hvy:
jenkins, test this please.

hvy:
jenkins, test this please.

hvy:
jenkins, test this please.

hvy:
LGTM!

PR #3436, opened 2017-09-20
Issue a warning when repeat property of some iterators in Evaluator is True.

delta2323:
This PR modifies `Evaluator` to emit a warning when the `repeat` property of the iterators it has is set to `True` to avoid infinite loops.

When we want to evaluate metrics with a *training* dataset, we sometimes mistakenly reuse the same iterator for training dataset used to train model and feed it to the evaluator. But as the iterator typically iterates over the dataset repeatedly, we are caught in an infinite loop. The objective of this PR is to warn users of such kind of stumbles.

There are other alternatives to achieve this goal:

* Forcibly set the `repeat` property of iterators to `False`.
* Raise errors instead of warnings.

To access the information of iterators, we add a public `repeat` property to `SerialIterator`.


niboshi:
(The point of this post is orthogonal to the choices discussed above)

IMO, it's not a good practice to peek into an undocumented property in ad-hoc manner.
It's possible that some iterators have `repeat` property with completely different meaning.

A better choice would be:
* Check whether it's a `SerialIterator` first, then check `repeat` property. (`if isinstance(iterator, SerialIterator) and iterator.repeat:`)

delta2323:
OK, I will add the check of the type of `iterator`. By the way, `MultiProcessIterator` also has `repeat` argument, so it would be better to add it to possible types.

himkt:
I want to know the current status of this PR 👁

chainer/training/extensions/evaluator.py
```diff
@@ -88,6 +91,22 @@ def __init__(self, iterator, target, converter=convert.concat_examples,
...
+                        getattr(iter, 'repeat', False)):
+                    msg = 'The repeat property of the iterator {} '
+                    'is set to True. Typically, evaluator sweeps '
+                    'over iterators until it raises StopIteration. '
+                    'But this iterator might not raise the exception, '
+                    'which means the evaluation could'
+                    'go into an infinite loop in evaluation. '
+                    'We recommend to check the configuration '
+                    'of iterators'.format(key)
+                    warnings.warn(msg)
```
niboshi: I feel this message is a bit like implementation detail, and relationship between `repeat` property and the behavior is not clearly stated.

How about something like this?

*The repeat property of the iterator {} is set to True. Typically, evaluator sweeps over iterators until it stops iteration, but as its repeat property is True, this iterator might not stop and the evaluator could go into an infinite loop. We recommend to check the configuration of the iterator.*

chainer/training/extensions/evaluator.py
```diff
@@ -148,6 +167,12 @@ def evaluate(self):
         accumulates the reported values to :class:`~chainer.DictSummary` and
         returns a dictionary whose values are means computed by the summary.
 
+        Note that this function assumes that the main iterator raises
+        ``StopIteration`` or some error is raised during the evaluation loop.
+        So, if the iterator does not raise ``StopIteratiron`` and
+        the loop does not raise no exceptions. It could be caught in the
```
niboshi: `...no exceptions. It could be...` -> `...no exceptions, it could be...`

tests/chainer_tests/training_tests/extensions_tests/test_evaluator.py
```diff
@@ -215,4 +217,36 @@ def test_evaluate(self):
...
+        self.context = warnings.catch_warnings(record=True)
+        self.warnings = self.context.__enter__()
+        warnings.filterwarnings(action='always', category=DeprecationWarning)
+
+    def tearDown(self):
+        self.context.__exit__()
+
+    def test_serial_iterator(self):
+        dataset = numpy.ones((4, 6))
+        iterator = iterators.SerialIterator(dataset, 2, repeat=self.repeat)
```
niboshi: `iterators.SerialIterator` -> `self.iterator`?

(Also, I think something like `self.iterator_type` or `self.iterator_class` is better)

tests/chainer_tests/training_tests/extensions_tests/test_evaluator.py
```diff
@@ -215,4 +217,36 @@ def test_evaluate(self):
...
+        with chainer.using_config('debug', self.debug):
+            with warnings.catch_warnings(record=True) as w:
+                extensions.Evaluator(iterator, {})
+
+        if self.debug and self.repeat:
+            expect = 1
+        else:
+            expect = 0
+
+        self.assertEqual(len(w), expect)
```
niboshi: Use `assert` (because we have moved to PyTest)

niboshi:
I agree with @himkt  in that this warning can be raised in non-debug mode, as the performance is not important.

delta2323:
> I agree with @himkt in that this warning can be raised in non-debug mode, as the performance is not important.

Are you thinking of speed in this context? I would say speed is important in some people who use Chainer in non-debug mode. So, I want to profile overhead introduced by this feature and use it in non-debugging mode as well if the overhead is negligible.

chainer/training/extensions/evaluator.py
delta2323: Fixed. I also modified points other than you pointed out. Could you read it again?

delta2323:
@niboshi  I reflected your comment. Could you take a look?

chainer/training/extensions/evaluator.py
delta2323: Fixed.

tests/chainer_tests/training_tests/extensions_tests/test_evaluator.py
delta2323: I used `iterator_class`

tests/chainer_tests/training_tests/extensions_tests/test_evaluator.py
delta2323: Fixed

niboshi:
Thank you for the fixes.

>Are you thinking of speed in this context? I would say speed is important in some people who use Chainer in non-debug mode. So, I want to profile overhead introduced by this feature and use it in non-debugging mode as well if the overhead is negligible.

Yes, I'm talking about speed.
I thought it's not important because `__init__` is called only once, whereas `__call__` is called repeatedly.

delta2323:
Got it. I've removed the condition.

delta2323:
@niboshi  Fixed. Could you review it again?

niboshi:
@delta2323 Thank you for the fix.

Please also remove `debug` configuration in the unit test.

delta2323:
Thanks, I fixed.

niboshi:
@delta2323 
I'm sorry for being late. Could you check test failure?

niboshi:
Hmm, test is still failing.

niboshi:
jenkins, test this please

niboshi:
LGTM!

PR #3437, opened 2017-09-21
[WIP] device.is_cpu and device.is_cuda

niboshi:
This is a proposal (thus WIP).

Currently, to test whether a device is CPU or CUDA, you compare with the hardcoded device ID: e.g. `dev.id == -1` or `int(dev) == -1`. This is not neccessarily intuitive.

This PR offers an uniform method to check the device type.


niboshi:
I think we need more broad refactoring with regard to backends. I'll close this PR for now.

PR #3443, opened 2017-09-21
Delete redundant test settings for get_item

yuyu2172:
This fixes a bug.
The option `dtype` is not used in the test functions.

I don't think it is necessary to check for `int32, uint32, ulonglong, uint64` because these dtypes are not differentiable.
Also, the function `__getitem__` and `scatter_add` are tested with these dtypes in tests for CuPy.


niboshi:
jenkins, test this please

niboshi:
LGTM!

PR #3448, opened 2017-09-23
Add get_variable_or_none to improve backward performance

okuta:
`get_variable` create `Variable` when `_variable` reference is dead.
But, some functions do not need temporary `Variable` object.


beam2d:
I feel the name `get_raw_variable` is confusing a bit. How about an alternative name `get_variable_if_exist` or `get_variable_or_none`?

okuta:
thanks, I will change

okuta:
I fixed.

niboshi:
LGTM!

PR #3468, opened 2017-09-28
inv_std is missing in backward of FixedBatchNormalization if it use cudnn.

zaburo-ch:
If we can use cudnn, the code below raises AttributeError because of missing inv_std.
```
import chainer
import numpy as np
X = np.random.uniform(0, 1, (1, 1))
gamma, beta = np.ones(1), np.zeros(1)
mean, var = np.mean(X, axis=0), np.var(X, axis=0)
X, gamma, beta, mean, var = [chainer.cuda.to_gpu(e, 0) for e in [X, gamma, beta, mean, var]]
X, gamma, beta = [chainer.Variable(e) for e in [X, gamma, beta]]
y = chainer.functions.fixed_batch_normalization(X, gamma, beta, mean, var)
num = chainer.functions.sum(y)
num.backward()
# AttributeError: 'FixedBatchNormalization' object has no attribute 'inv_std'
```

The test of fixed_batch_normalization can not detect the error because `param_shape` used in the test does not make `can_use_cudnn` true in forward computation.

Therefore, I added a shape to the test and fixed it.

### DIFF
--- a/chainer/functions/normalization/batch_normalization.py
+++ b/chainer/functions/normalization/batch_normalization.py
@@ -301,6 +301,12 @@ class FixedBatchNormalization(function_node.FunctionNode):
 
     def backward(self, indexes, grad_outputs):
         x, gamma, mean, var = self.get_retained_inputs()
+
+        if not hasattr(self, 'inv_std'):
+            xp = cuda.get_array_module(x)
+            self.inv_var = xp.reciprocal(var.data + self.eps)
+            self.inv_std = xp.sqrt(self.inv_var, dtype=self.inv_var.dtype)
+
         gy, = grad_outputs
         f = FixedBatchNormalizationGrad(
             self.eps, self.expander, self.axis, self.inv_std, self.inv_var)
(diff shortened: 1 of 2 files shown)


chainer/functions/normalization/batch_normalization.py
```diff
@@ -301,6 +301,12 @@ def forward(self, inputs):
 
     def backward(self, indexes, grad_outputs):
         x, gamma, mean, var = self.get_retained_inputs()
+
+        if not hasattr(self, 'inv_std'):
```
beam2d: Could you move the computation of `inv_var` and `inv_std` to https://github.com/chainer/chainer/pull/3468/files#diff-148edea535bb806260c487ca31869a91R331 and adds `inv_std = None` to https://github.com/chainer/chainer/pull/3468/files#diff-148edea535bb806260c487ca31869a91R234? I feel it's more natural fix (the original code is strange, but such a fix completes the original incomplete implementation).

zaburo-ch:
Thanks for your review. I fixed it.

beam2d:
Jenkins, test this please.

beam2d:
LGTM

PR #3472, opened 2017-09-28
Improve numerical_grad accuracy

okuta:
`numerical_grad` contains subtraction of very close values
So this PR uses float64 to improve accuracy and stability.


niboshi:
* Please allocate `grad` in float64, too.

>```grads = [xp.zeros_like(x) for x in inputs]```

* Write in docstring that gradients will be returned in float64.

okuta:
Thanks!

niboshi:
LGTM!

PR #3479, opened 2017-10-02
[backport] inv_std is missing in backward of FixedBatchNormalization if it use cudnn

beam2d:
This is a backport of #3468.


niboshi:
LGTM!

PR #3486, opened 2017-10-03
Support skipping padded regions in average_pooling_nd

### LINKED ISSUE #3391: `average_pooling_2d` treats padded elements as zeroes
whereas `max_pooling_2d` treats them as non-existent.

Code:
```
import numpy

pad=(0, 1)
ksize=(1, 3)
stride=(1, 3)

x = numpy.array([[[[-2., -1., 0., 1.]]]])
print(x)
y = F.average_pooling_2d(x, ksize, stride=stride, pad=pad)
print(y)
```

Output (`average_pooling_2d`):
```
[[[[-2. -1.  0.  1.]]]]
variable([[[[-1.          0.33333333]]]])   # avg(0, -2, -1) = -1, avg(0, 1, 0) = 0.333
```

In the case of `F.max_pooling_nd`, paddd elements are not involved in max:
```
[[[[-2. -1.  0.  1.]]]]
variable([[[[-1.  1.]]]])  # max(-2, -1) = -1, max(0, -1) = 1
```

For comparison, TensorFlow treats them as non-existent:
```
import tensorflow as tf

with tf.Session():
    x = -tf.constant([[[[-2.], [-1.], [0.], [1.]]]])
    print(x.eval())
    ksize = (1, 1, 3, 1)
    stride = (1, 1, 3, 1)
    y = tf.nn.avg_pool(x, ksize, stride, 'SAME')
    print(y.eval())
```

```
[[[[-2.]
   [-1.]
   [ 0.]
   [ 1.]]]]
[[[[-1.5]
   [ 0.5]]]]   # avg(-2, -1) = -1.5, avg(0, 1) = 0.5
```

Chainer version: v3.0.0rc1, latest master (72773122ce9ffd724ea0a54922c9fc62ea73a915)

niboshi:
Fixes #3391 

Supports skipping padded regions when avraging.


niboshi:
It seems unit test is not sufficient.

chainer/functions/pooling/average_pooling_nd.py
```diff
@@ -105,7 +157,7 @@ def create_pool_desc(self):
             cuda.cudnn.cudnn.CUDNN_POOLING_AVERAGE_COUNT_INCLUDE_PADDING)
```
kmaehashi: (For record) there are two modes available in cuDNN. `CUDNN_POOLING_AVERAGE_COUNT_INCLUDE_PADDING` and `CUDNN_POOLING_AVERAGE_COUNT_EXCLUDE_PADDING`.

kmaehashi:
Refs #4282 (ideep)

kmaehashi:
We need to use `CUDNN_POOLING_AVERAGE_COUNT_EXCLUDE_PADDING` if `pad` is `None`.

kmaehashi:
In the weekly discussion we thought this is high priority. I tentatively set milestone to the next release. Could you restart working on this? @niboshi

niboshi:
@kmaehashi OK.

niboshi:
Updated.
Review #4743 first.

tests/chainer_tests/functions_tests/pooling_tests/test_average_pooling_nd.py
```diff
@@ -52,19 +53,29 @@ def check_forward(self, x_data, use_cudnn='always'):
         pad = self.pad
         x = chainer.Variable(x_data)
         with chainer.using_config('use_cudnn', use_cudnn):
-            y = functions.average_pooling_nd(x, ksize, stride, pad)
+            y = functions.average_pooling_nd(
+                x, ksize, stride, pad, self.pad_value)
         self.assertEqual(y.data.dtype, self.dtype)
         y_data = cuda.to_cpu(y.data)
 
+        def idx_size(idx):
```
kmaehashi: Isn't it better to name it `patch_size`?

chainer/utils/conv.py
```diff
@@ -33,6 +33,23 @@ def get_conv_outsize(size, k, s, p, cover_all=False, d=1):
         return (size + p * 2 - dk) // s + 1
 
 
+def get_conv_slices(
+        size, k, s, p, cover_all=False, d=1, include_pad=True, dtype='l'):
```
kmaehashi: I think it's better to add test case for this.

chainer/functions/pooling/average_pooling_nd.py
```diff
@@ -158,6 +217,11 @@ def average_pooling_nd(x, ksize, stride=None, pad=0):
             window size.
         pad (int or tuple of ints): Spatial padding width for the input array.
             ``pad=p`` and ``pad=(p, p, ..., p)`` are equivalent.
+        pad_value (0 or None):
+            Value to fill the padded region when calculating average.
+            If ``None`` is specified, such region is ignored.
+            The default value is ``0``, therefore the averages are biased
+            towards zero.
```
kmaehashi: How about adding notes similar to this to the docstring of `average_pooling_2d`?
(e.g., `Currently average_pooling_2d treats padded regions as filled by 0 when calculating average.`)

tests/chainer_tests/functions_tests/pooling_tests/test_average_pooling_nd.py
niboshi: `idx` and `patch` are used interchangeably in this context and neither is accurate.
I renamed to `denom`.

chainer/utils/conv.py
niboshi: I think it's ok because it's an undocumented function and tested indirectly.
I made it explicitly private.

chainer/functions/pooling/average_pooling_nd.py
niboshi: Added note.

Issue #3211, opened 2017-08-22
Add ability to setup optimizer on model as argument


Crissman:
Currently, setting up an optimizer requires two lines:

```python
optimizer = chainer.optimizers.Adam()
optimizer.setup(model)
```

To keep parallel structure with the other Classes, I think these steps could be consolidated.

```python
optimizer = chainer.optimizers.Adam(model)
```

This would be an enhancement, and existing functionality would be retained for compatibility.

Good idea?

hvy:
Fixed by #3488.

PR #3488, opened 2017-10-04
Enable optimizer model setup with instantiation

Crissman:
Code to close issue #3211 .

Allow models to be passed as an argument to an optimizer, so
```
optimizer.setup(model)
```
becomes:
```
optimizer = chainer.optimizers.SGD(model)
```

Change added to all `optimizers`. Old two-line method also still valid.

Tested by running new one-line method and old method on SGD, then running new method on all `optimizers` after changing code.

### DIFF
--- a/chainer/optimizer.py
+++ b/chainer/optimizer.py
@@ -478,10 +478,9 @@ class GradientMethod(Optimizer):
 
     """
 
-    def __init__(self, *args):
+    def __init__(self, link):
         super(GradientMethod, self).__init__()
         self.hyperparam = Hyperparameter()
-        link = (args or [None])[0]
         if isinstance(link, link_module.Link):
             self.setup(link)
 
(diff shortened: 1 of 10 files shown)


mitmul:
@Crissman Check the travis-ci's test errors and please fix them.

mitmul:
Jenkins, test this please

mitmul:
LGTM

niboshi:
Note that this code does not work, because the first argument of `SGD` is `lr`.
```
optimizer = chainer.optimizers.SGD(model)
```
I think it's better to make `model` a keyword-only argument.

Crissman:
@niboshi Correct. `model` should be a keyword argument. I've updated my comment above accordingly.

kmaehashi:
Removed `to-be-backported` (as discussed in #3936)

PR #3489, opened 2017-10-04
New style hstack

hvy:
New style hstack function to support double backprop.

I am unsure whether accessing `self.inputs`, i.e. the `VariableNode`s from within the `FunctionNode` is against practices. There is no need to retain the inputs since we only want to know the shapes.

### DIFF
--- a/chainer/functions/array/hstack.py
+++ b/chainer/functions/array/hstack.py
@@ -33,16 +33,17 @@ class Hstack(function_node.FunctionNode):
 
     def backward(self, indexes, grad_outputs):
         gy, = grad_outputs
+        ndim = len(self.inputs[0].shape)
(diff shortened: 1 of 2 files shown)


hvy:
Seems like the CI failed on tests that aren't relevant to this PR.

niboshi:
It seems double-backward test is numerically unstable (for float16).
Can you adjust tolerances?

Tips: you can put `@testing.condition.repeat(10000)` just before `def test_double_backward_cpu()`, for example, to check its stability.

hvy:
Thank, for the comment. I'll all add an option for tests with low precision. And thank you for the tip!

hvy:
jenkins, test this please.

niboshi:
LGTM! Thank you!

PR #3491, opened 2017-10-04
New style where function

kmaehashi:
This PR changes `F.where` function to new style.

### DIFF
--- a/chainer/functions/array/where.py
+++ b/chainer/functions/array/where.py
@@ -1,11 +1,11 @@
 import numpy
 
 from chainer import cuda
-from chainer import function
+from chainer import function_node
 from chainer.utils import type_check
 
 
-class Where(function.Function):
+class Where(function_node.FunctionNode):
 
     """Choose elements depending on condition."""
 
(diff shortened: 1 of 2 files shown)


chainer/functions/array/where.py
```diff
@@ -26,11 +26,13 @@ def forward(self, inputs):
...
-        xp = cuda.get_array_module(inputs[0])
-        condition = inputs[0]
-        gx = xp.where(condition, grads[0], 0)
-        gy = xp.where(condition, 0, grads[0])
+    def backward(self, indexes, grad_outputs):
+        condition = self.inputs[0]
+        xp = cuda.get_array_module(condition.data)
+        g, = grad_outputs
+        zeros = xp.zeros(g.shape, dtype=g.dtype)
+        gx, = Where().apply((condition.data, g, zeros))
```
beam2d: Could you skip the computation of `gx`/`gy` if not needed? Whether or not they are needed can be determined by `indexes`.

chainer/functions/array/where.py
```diff
@@ -26,11 +26,13 @@ def forward(self, inputs):
...
         return xp.where(condition, x, y),
 
-    def backward(self, inputs, grads):
-        xp = cuda.get_array_module(inputs[0])
-        condition = inputs[0]
-        gx = xp.where(condition, grads[0], 0)
-        gy = xp.where(condition, 0, grads[0])
+    def backward(self, indexes, grad_outputs):
+        condition = self.inputs[0]
+        xp = cuda.get_array_module(condition.data)
```
beam2d: You have to call `self.retain_inputs` if you need input data, and use `get_retained_inputs` to obtain the input variable (see the implementation of `linear` for example).

kmaehashi: Fixed.
Some other functions are directly using self.input and they must be fixed too I think?
e.g., https://github.com/kmaehashi/chainer/blob/492c9bb61fd44457c5e81df5c87120807110e7b8/chainer/functions/array/permutate.py#L79

chainer/functions/array/where.py
kmaehashi: Fixed.

beam2d:
Jenkins, test this please

beam2d:
LGTM

PR #3493, opened 2017-10-04
New style scatter_add

hvy:
New style scatter_add function to support double backprop.

### DIFF
--- a/tests/chainer_tests/functions_tests/array_tests/test_scatter_add.py
+++ b/tests/chainer_tests/functions_tests/array_tests/test_scatter_add.py
@@ -36,6 +36,7 @@ class TestScatterAdd(unittest.TestCase):
             -1, 1, self.a_data.shape).astype(numpy.float32)
         self.ggb_data = numpy.random.uniform(
             -1, 1, self.b_data.shape).astype(numpy.float32)
+        self.check_backward_options = {'atol': 5e-4, 'rtol': 1e-4, 'dtype': 'f'}
 
     def check_forward(self, a_data, b_data):
         a = chainer.Variable(a_data)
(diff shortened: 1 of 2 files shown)


niboshi:
Double-backward test is numerically unstable. Please adjust tolerances.

hvy:
Thanks for the update. I changed the tolerances accordingly. 

This test slightly differs from the other similar tests since the underlying CuPy scatter_add does not support operations on float64.

hvy:
jenkins, test this please.

niboshi:
Please fix conflict.

hvy:
@niboshi Fixed.

tests/chainer_tests/functions_tests/array_tests/test_scatter_add.py
```diff
@@ -32,6 +32,12 @@ def setUp(self):
         self.gy_data = numpy.random.uniform(
             -1, 1, self.shape).astype(numpy.float32)
         self.b_data = self.b_data.astype(numpy.float32)
+        self.gga_data = numpy.random.uniform(
+            -1, 1, self.a_data.shape).astype(numpy.float32)
+        self.ggb_data = numpy.random.uniform(
+            -1, 1, self.b_data.shape).astype(numpy.float32)
+        self.check_backward_options = {
```
niboshi: * Shouldn't this be `check_double_backward_options`?
* I think you don't need to explicitly specify `dtype` argument because all inputs are `float32`.

niboshi:
Also, please fix double-backward tolerances. It still causes numerical errors.

tests/chainer_tests/functions_tests/array_tests/test_scatter_add.py
hvy: Removed the redundant dtype and `check_double_backward_options ` since it is only used at one place.

niboshi:
LGTM!

niboshi:
Not backporting because `scatter_add` has not been backported.

PR #3498, opened 2017-10-05
Add a stale bot configuration

beam2d:
This PR adds a configuration of the stale bot: https://github.com/probot/stale. The numbers of `daysUntilStale` and `daysUntilClose` are borrowed from the configuration used at https://github.com/fchollet/keras. We can change the numbers if they are not suited to our case; feel free to propose other numbers.


okuta:
Should we create stale label?

beam2d:
Thanks. Created `stale` label.

PR #3511, opened 2017-10-06
Invoke extensions before training

wkentaro:
## Why?

I sometimes need to know the loss and accuracy before the training (iteration=0), but currently I'm not sure how to do this with the Trainer class.
This PR achieves it.

## Demonstration

It writes loss and accuracy result even if the epoch is 0.

```
cd examples/mnist
python train_mnist.py
```

 

## ~Test~

```
cd tests/chainer_tests/training_tests/triggers_tests
python test_interval_trigger.py
```

### DIFF
--- a/chainer/training/trainer.py
+++ b/chainer/training/trainer.py
@@ -311,7 +311,7 @@ class Trainer(object):
         reporter = self.reporter
         stop_trigger = self.stop_trigger
 
-        # call before training loop
+        # call extensions before training loop
         if self.updater.iteration == 0:
             self.observation = {}
             with reporter.scope(self.observation):
--- a/chainer/training/trainer.py
+++ b/chainer/training/trainer.py
@@ -167,6 +167,10 @@ class Trainer(object):
             self.extend(ext)
 
     @property
+    def is_before_training(self):
+        return self.updater.iteration == 0
+
+    @property
     def elapsed_time(self):
         """Total time used for the training.
 
@@ -312,7 +316,7 @@ class Trainer(object):
         stop_trigger = self.stop_trigger
 
         # call extensions before training loop
-        if self.updater.iteration == 0:
+        if self.is_before_training:
             self.observation = {}
             with reporter.scope(self.observation):
                 for name, entry in extensions:
--- a/chainer/training/trainer.py
+++ b/chainer/training/trainer.py
@@ -25,10 +25,11 @@ except AttributeError:
 
 class _ExtensionEntry(object):
 
-    def __init__(self, extension, priority, trigger):
+    def __init__(self, extension, priority, trigger, call_before_training):
         self.extension = extension
         self.trigger = trigger
         self.priority = priority
+        self.call_before_training = call_before_training
 
 
 class Trainer(object):
@@ -181,7 +182,7 @@ class Trainer(object):
         return _get_time() - self._start_at + self._snapshot_elapsed_time
 
     def extend(self, extension, name=None, trigger=None, priority=None,
-               **kwargs):
+               call_before_training=False, **kwargs):
         """Registers an extension to the trainer.
 
         :class:`Extension` is a callable object which is called after each
@@ -212,6 +213,8 @@ class Trainer(object):
                 iteration by default. If the trigger is not callable, it is
                 passed to :class:`IntervalTrigger` to build an interval
                 trigger.
+            call_before_training (bool): Flag to call extension before
+                training. Default is ``False``.
             priority (int): Invocation priority of the extension. Extensions
                 are invoked in the descending order of priorities in each
                 iteration. If this is ``None``, ``extension.priority`` is used
@@ -253,7 +256,7 @@ class Trainer(object):
 
         extension.name = modified_name
         self._extensions[modified_name] = _ExtensionEntry(
-            extension, priority, trigger)
+            extension, priority, trigger, call_before_training)
 
     def get_extension(self, name):
         """Returns the extension of a given name.
@@ -307,6 +310,14 @@ class Trainer(object):
         update = self.updater.update
         reporter = self.reporter
         stop_trigger = self.stop_trigger
+
+        # call before training loop
+        if self.updater.iteration == 0:
+            self.observation = {}
+            with reporter.scope(self.observation):
+                for name, entry in extensions:
+                    if entry.call_before_training:
+                        entry.extension(self)
 
         # main training loop
         try:
--- a/chainer/training/trainer.py
+++ b/chainer/training/trainer.py
@@ -168,6 +168,18 @@ class Trainer(object):
 
     @property
     def is_before_training(self):
+        """Flag that represents if training has started or not.
+
+        ``True`` represents 'before training' and
+        ``False`` represents 'during/after training'.
+
+        This flag is supposed to be used in :meth:`Extension.__call__`
+        (e.g., :meth:`PlotReport.__call__`) to decide to execute its operation
+        or not. This additional condition is necessary since
+        ``Extension._trigger(trainer)`` is always ``False`` before training
+        and cannot be used.
+
+        """
         return self.updater.iteration == 0
 
     @property
--- a/examples/mnist/train_mnist.py
+++ b/examples/mnist/train_mnist.py
@@ -86,7 +86,8 @@ def main():
     trainer = training.Trainer(updater, (args.epoch, 'epoch'), out=args.out)
 
     # Evaluate the model with the test dataset for each epoch
-    trainer.extend(extensions.Evaluator(test_iter, model, device=device))
+    trainer.extend(extensions.Evaluator(test_iter, model, device=device),
+                   call_before_training=True)
 
     # Dump a computational graph from 'loss' variable at the first iteration
     # The "main" refers to the target link of the "main" optimizer.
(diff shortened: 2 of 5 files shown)


hvy:
Thank you for this PR. I definitely think that there is a demand for this.

Is it the case that you need to call the `__call__` method via `initialize` **in addition to** triggering the extension via `invoke_before_training`? In other words, is it not enough to call `initialize`?

wkentaro:
@hvy 
Do you mean why do I need to change `IntervalTrigger`? It is because current `IntervalTrigger` doesn't fire at `iteration == 0`, so even if we call `__call__` method in `initialize`, that does not trigger the extension in actual. 

For example, `__call__` of `LogReport` is called in `initialize` in the MNIST example, but without `invoke_before_training=True` option to `IntervalTrigger`, below if-block is always skipped because it is the current behavior of `IntervalTrigger`.  
https://github.com/chainer/chainer/blob/1bc30968ddb9972b639f73e0686fe316be0e04b2/chainer/training/extensions/log_report.py#L78

hvy:
Thank you for the explanation. I am not sure exactly which extensions you had in mind but I feel that it is better to actually implement the `initialize` methods for each of those to act appropriately. What do you think?

wkentaro:
I think that is better. But the problem, in that case, is how to give users the option of choosing run `initialize` or not.
Currently, most of all extensions' `initialize` is None and it does not run by default. 

Previously, the extensions had an option of `invoke_before_training` for `__init__`, but it was removed after refactoring (maybe after v2?).
I am not sure whether we can re-add that option.

hvy:
I don't think we should expect the user to set the `initialize` member with `__call__` in any case. It doesn't really speak for a good API. But I see that adding your third argument to the trigger is a flexible solution. @beam2d do you have any inputs on this?

Hakuyume:
Modifying `IntervalTrigger` doesn't seem a good solution. I think each extension that can be called before training should implement `initialize`. If user can enable/disable `initialize`, it should be passed as an argument of the extension.

hvy:
@Hakuyume thanks for the input. Could you explain why you think modifying the `IntervalTrigger` isn't a good solution?

Hakuyume:
`IntervalTrigger` fires when `previous < n * interval <= current`. If we want `IntervalTrigger` fire at the first iterarion, it should be the first time the trigger called. This is not exactly same as `iteration == 0`.

For example, in the case that an `IntervalTrigger(4, 'iteration') is called at 3, 6 ,9, 12... iterations,
it should fire at 3 (0 * 4 <= 3), 6 (4 < 1 * 4 <= 6), 9 (6 < 2 * 4 <= 9), 12 (3 * 4 <=12) .

hvy:
I think we need to reconsider if this is a applies to all extensions, a subset of them, or if firing at 0 should be the expected default behavior even for all of the extensions. If this applies to a subset, I feel https://github.com/chainer/chainer/pull/3511#issuecomment-350159097 is fine.

What do you think @wkentaro @Hakuyume ?

wkentaro:
Thank you for the suggestion @Hakuyume @hvy . I agree with your idea.
I updated my implementation to use `initialize` to achieve trigger extension before training.
Could you review?

wkentaro:
Now the training log looks like below:
```
epoch       iteration   main/loss   validation/main/loss  main/accuracy  validation/main/accuracy  elapsed_time
0           0                       2.27572                              0.1993                    0.801908
1           600         0.194542    0.0923036             0.94195        0.9711                    21.0456
2           1200        0.0733713   0.0769297             0.976917       0.977                     39.2025
3           1800        0.048582    0.071287              0.984517       0.9792                    59.7017
4           2400        0.0361622   0.0717561             0.9883         0.9795                    80.6741
5           3000        0.0296503   0.074036              0.990417       0.9785                    100.863
6           3600        0.0233984   0.083005              0.991867       0.9798                    120.973
7           4200        0.0209341   0.082974              0.993017       0.9785                    141.478
8           4800        0.0178388   0.0849701             0.994133       0.9803                    163.459
9           5400        0.0189349   0.071605              0.993983       0.982                     188.598
10          6000        0.0139158   0.113345              0.99555        0.9753                    210.196
11          6600        0.0150554   0.0873922             0.995133       0.9812                    231.432
12          7200        0.00938825  0.126652              0.997283       0.9743                    253.983
13          7800        0.0132886   0.0926085             0.996183       0.9802                    276.352
14          8400        0.0125203   0.0977689             0.996483       0.9803                    300.099
15          9000        0.0102308   0.100624              0.996983       0.9815                    325.268
```

hvy:
Thank you for the fixes. I'll be taking another look!

examples/mnist/train_mnist.py
```diff
@@ -77,7 +77,9 @@ def main():
     trainer = training.Trainer(updater, (args.epoch, 'epoch'), out=args.out)
 
     # Evaluate the model with the test dataset for each epoch
-    trainer.extend(extensions.Evaluator(test_iter, model, device=args.gpu))
+    evaluator = extensions.Evaluator(test_iter, model, device=args.gpu)
+    evaluator.initialize = evaluator.__call__  # for evaluation at epoch 0
```
hvy: I am not sure about this interface. Can we not add an argument to the constructor instead?

hvy:
Sorry for the late comment but would you mind taking a look? My question is how the user is supposed to control whether the extensions are invoked before training or not.

examples/mnist/train_mnist.py
wkentaro: I think there are 3 options:

1. `invoke_before_training=False/True` which implicitly overwrites `initialize` by `__call__`. This can cause unexpected behavior if user overwrites `Evaluator` and defines a custom `initialize` in it.
2. Passing `initialize=None/Extension.__call__/etc.` to the constructor and it overwrites the `initialize` method of the Extension if it is not `None`. This is fine, but it might be confusing that there is no way to overwrite other methods for example `__call__` and `finalize`. Also, it can't overwrite `initialize` method of extensions created by `extensions.make_extension`, for example `extensions.snapshot()`, so user must manually overwrite it.
3. Current implementation. It explicitly overwrites `initialize`, so it is less confusing. 

Currently, I'm conservative to adopt other implementations because of above problems caused by overwriting method in `__init__`.
What do you think?

wkentaro:
Probably are you waiting for my comments? @hvy 
My comment is here https://github.com/chainer/chainer/pull/3511#discussion_r195986422.

examples/mnist/train_mnist.py
hvy: I think 1. is best but with some slight modifications. What do you think of the following?

- `call_before_training=False/True` passed to both `extend` and the constructors of the extensions. This flag is checked to decide whether to fire or not.
- Do not change the behavior of `initialize` since it's actual designed for a different purpose. Namely to update the trainer on resume.
- Make sure that the "0 iteration" invocations is not triggered on resume (check the iteration count).

hvy:
Sorry for the delay. I wrote some comments and suggestion. Could you take a look at them  https://github.com/chainer/chainer/pull/3511#discussion_r ?

knorth55:
@wkentaro Are you still working on this PR?

wkentaro:
I will work on this in this week.

wkentaro:
Sorry for very late. I was occupied by paper revisions.

I pushed new changes and now it looks much simpler. Could you check it? @hvy

hvy:
Thanks for the changes although you've been busy. I'll try to take a look.

hvy:
Jenkins, test this please.

hvy:
Is it possible to write tests for this feature?

wkentaro:
> Is it possible to write tests for this feature?

I added a test in a74e0ad.

chainer/training/trainer.py
```diff
@@ -308,6 +311,14 @@ def run(self, show_loop_exception_msg=True):
         reporter = self.reporter
         stop_trigger = self.stop_trigger
 
+        # call before training loop
```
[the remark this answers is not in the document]
wkentaro: I fixed it in 5a69154.

wkentaro:
I'm still interested in this feature. Could anyone review this?

toslunar:
Let me put the link to #2639.

wkentaro:
You mean we should pass `call_before_training` argument to each Extension’s `__init__` instead of `Trainer.extend`? Why it is better?

toslunar:
I've just understood reusing `initialize` is not a good idea.  I withdraw my previous comment.

wkentaro:
Thanks. I added it.

chainer/training/trainer.py
```diff
@@ -165,6 +166,10 @@ def __init__(self, updater, stop_trigger=None, out='result',
         for ext in extensions:
             self.extend(ext)
 
+    @property
+    def is_before_training(self):
```
hvy: Could you write a short documentation what this returns and that it can be used by extensions (for e.g. those who'd be interested in writing their own)?

hvy:
Again, sorry for the delay. Could you take a look at my comment above and also resolve the conflicts? The changes otherwise LGTM!

wkentaro:
Thanks for the review. I added the documentation and resolved the conflicts.

hvy:
Jenkins, test this please.

hvy:
LGTM, let me merge this long awaited PR. Thanks and sorry having it running for such a long time @wkentaro @toslunar .

wkentaro:
Thank you for the reviews!

PR #3513, opened 2017-10-07
Support file-like object in npz serializer

niboshi:
This PR support file-like objects in `save_npz`/`load_npz`.

### DIFF
--- a/chainer/serializers/npz.py
+++ b/chainer/serializers/npz.py
@@ -1,5 +1,7 @@
 import numpy
 
+import six
+
 from chainer import cuda
 from chainer import serializer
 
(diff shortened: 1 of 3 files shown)


chainer/serializers/npz.py
```diff
@@ -51,25 +51,29 @@ def __call__(self, key, value):
...
+
     """
     s = DictionarySerializer()
     s.save(obj)
-    with open(filename, 'wb') as f:
-        if compression:
-            numpy.savez_compressed(f, **s.target)
-        else:
-            numpy.savez(f, **s.target)
+    if isinstance(file, str):
```
kmaehashi: In Python 2.x, `unicode` values may be specified as `file`.
I think using `six.string_types` instead of `str` is better here.

niboshi: Fixed!

kmaehashi:
Thanks!  Will mege after AppVeyor pass.

kmaehashi:
LGTM!

PR #3530, opened 2017-10-10
new style abs function

kmaehashi:
This PR implements new style `abs` function.

I also added double backward test case for `neg` (which is already in new style)


okuta:
LGTM!

PR #3539, opened 2017-10-11
new style floor function

kmaehashi:
This PR implements new style `floor` function.

I think that this `backward` implementation is not efficient compared to the old one.
Are there any better way to do `xp.zeros` efficiently with computation graph?

### DIFF
--- a/tests/chainer_tests/functions_tests/math_tests/test_floor.py
+++ b/tests/chainer_tests/functions_tests/math_tests/test_floor.py
@@ -47,6 +47,21 @@ class UnaryFunctionsTestBase(unittest.TestCase):
     def check_backward_gpu(self, op):
         self.check_backward(op, cuda.to_gpu(self.x), cuda.to_gpu(self.gy))
 
+    def check_double_backward(self, op, x_data, y_grad, y_grad_grad):
+        def f(x):
+            x = op(x)
+            return x * x
+        gradient_check.check_double_backward(
+            f, x_data, y_grad, y_grad_grad, dtype=numpy.float64,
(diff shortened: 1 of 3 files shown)


tests/chainer_tests/functions_tests/math_tests/test_floor.py
```diff
@@ -60,6 +75,7 @@ class TestFloor(UnaryFunctionsTestBase):
     def make_data(self):
         x = numpy.random.uniform(-10.0, 10.0, self.shape).astype(self.dtype)
         gy = numpy.random.uniform(-1, 1, self.shape).astype(self.dtype)
+        self.ggy = numpy.random.uniform(-1, 1, self.shape).astype(self.dtype)
```
takagi: Would you let `make_data` to return a triple of `(x, gy, ggy)` instead of assigning `ggy` to the instance variable here?

takagi:
Basically looks good to me. I would like to wait before the discussion about zero grad becomes clear.

tests/chainer_tests/functions_tests/math_tests/test_floor.py
kmaehashi: Thanks, fixed.

takagi:
As mentioned in today's meeting, functions which always have zero gradient may be implemented as plain Python functions without `FunctionNode` subclass.

https://github.com/chainer/chainer/pull/3678/files#diff-f85dbb6d0de3e9f718ab89598f6f0b0a

kmaehashi:
Thank you, I removed FunctionNode class.

takagi:
LGTM!

PR #3543, opened 2017-10-11
fix documention error in Function

kmaehashi:
This PR fixes wrong description of `indexes` arguments for `retain_{input,output}`.


niboshi:
LGTM!

PR #3544, opened 2017-10-11
Loss Scaling

anaruse:
This PR adds a loss scaling feature: http://docs.nvidia.com/deeplearning/sdk/mixed-precision-training/index.html#lossscaling.

You can use the loss scaling feature by just setting the scaling factor using a method `set_loss_scaling_factor()` as below.

    from chainer import training
    
    training.set_loss_scaling_factor(1024.0)
    trainer.run()

When the scaling factor is set and data type of parameters is fp16, a fp32 copy of parameters are created automatically and parameters are updated with fp32.

Please note that only ``StandardUpdater`` is supported for now. After some consensus is obtained that this API is OK, I will make the loss scaling available to other Updaters.

### DIFF
--- a/chainer/optimizer.py
+++ b/chainer/optimizer.py
@@ -8,6 +8,7 @@ import six
 from chainer import cuda
 from chainer import link as link_module
 from chainer import serializer as serializer_module
+from chainer import training
 from chainer import variable
 
 
@@ -526,6 +527,9 @@ class GradientMethod(Optimizer):
         if lossfun is not None:
             use_cleargrads = getattr(self, '_use_cleargrads', True)
             loss = lossfun(*args, **kwds)
+            loss_scaling_factor = training.get_loss_scaling_factor()
+            if loss_scaling_factor is not None:
+                loss *= loss_scaling_factor
             if use_cleargrads:
                 self.target.cleargrads()
             else:
--- a/chainer/training/__init__.py
+++ b/chainer/training/__init__.py
@@ -20,3 +20,38 @@ from chainer.training.trigger import IntervalTrigger  # NOQA
 from chainer.training.updater import ParallelUpdater  # NOQA
 from chainer.training.updater import StandardUpdater  # NOQA
 from chainer.training.updater import Updater  # NOQA
+
+
+from chainer.configuration import config  # NOQA
+from chainer.configuration import global_config  # NOQA
+
+import numpy
+
+
+global_config.update_parameter_in_fp32 = False
+global_config.loss_scaling_factor = None
+
+
+def set_loss_scaling_factor(val):
+    """Sets loss scaling factor."""
+    config.loss_scaling_factor = val
+    config.update_parameter_in_fp32 = True
+
+
+def get_loss_scaling_factor():
+    """Gets loss scaling factor."""
+    return config.loss_scaling_factor
+
+
+def should_update_parameter_in_fp32(dtype):
+    """Determins if the parameter should be updated in fp32.
+
+    Args:
+        dtype (numpy.dtype): data type of the parameter.
+
+    Returns:
+        bool: ``True`` if the parameter should be updated in fp32.
+    """
+    if config.update_parameter_in_fp32 and dtype == numpy.float16:
+        return True
+    return False
--- a/chainer/variable.py
+++ b/chainer/variable.py
@@ -11,6 +11,7 @@ import chainer
 from chainer import cuda
 from chainer import initializers
 from chainer.initializers import constant
+from chainer import training
 from chainer.utils import argument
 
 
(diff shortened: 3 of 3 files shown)


beam2d:
Thank you for the PR.

I think the loss calling should be a feature of backprop computation instead of framework-wide "mode". For example, if the feature can be implemented by adding `loss_scale` option to `Variable.backward` and `chainer.grad`, that's great. We may also want to make `StandardUpdater` support configuring that argument.

IMO, the fp32 update should be treated separately from the loss scaling setting (and thus it's better to split the PR itself as well). I recommend you to implement it as a feature of `UpdateRule` instead of `Parameter`.

anaruse:
I will split this PR to two PRs: one for fp32 update, the other for loss scaling.

anaruse:
I've just submitted a PR for fp32 update as #3708. Please review #3708 first.

anaruse:
The loss scaling feature was re-implemented as a feature of backprop as suggested and the fp32 update feature was removed from this PR.

A `loss_scale` option is added to `Variable.backward` and `chainer.grad`, so that you can use the loss scaling by adding the `loss_scale` option to `Variable.backward` or `chainer.grad` in your code like below.

    loss = lossfunc(y, t)
    loss.backward(loss_scale=1024)

When loss scaling factor is set, the gradient of the loss is scaled up (multiplied by the scaling factor), and the scaling factor is propagated to whole gradients of parameters in a computational graph along the backprop. The gradients of each parameter are scaled down (divided by the scaling factor) just before the parameter is to be updated, so that in principle the same result can be obtained as when loss scaling is not used.

The `loss_scale` option is also added to `StandardUpdater`, so if you use `StandardUpdater`, you can use loss scaling by adding the `loss_scale` option to `StandardUpdater` in your code like below.

    updater = training.StandardUpdater(..., loss_scale=1024)

beam2d:
Sorry for being late. Could you rebase it to the current master?

anaruse:
OK, conflicts to the master branch were resolved. Please check.

chainer/function_node.py
```diff
@@ -875,6 +878,7 @@ def _backprop(outputs, inputs, grad_required, retain_grad, grads):
                 v = node.get_variable_or_none()
                 if v is not None:
                     v.grad_var = g
+                    v.loss_scale = loss_scale
```
beam2d: It's better to use a "private" name, like `v._loss_scale`, and I think it should also be added to the definition of `Variable`. It's enough to add a class attribute `_loss_scale = None`, which is used as the default value.

anaruse: OK, will change the variable name from `loss_scale` to `_loss_scale` and initialize it as you suggested.

anaruse:
The branch was fixed as you suggested. Explanation about `loss_scale` option was also added to related methods.

chainer/optimizer.py
```diff
@@ -581,7 +589,7 @@ def update(self, lossfun=None, *args, **kwds):
                 self.target.cleargrads()
             else:
                 self.target.zerograds()
-            loss.backward()
+            loss.backward(loss_scale=self.loss_scale)
```
beam2d: I guess this line fails if `set_loss_scale` has not been called. `loss_scale` should have the default value (`None`) and be checked.

anaruse: Thank you for pointing it out!  will fix to initialize that `loss_scale`.

anaruse:
Added unit tests for loss scaling. Please check.

tests/chainer_tests/test_function_node.py
```diff
@@ -617,10 +634,16 @@ def forward(self):
         self.y = self.x * self.x
 
     def expected_grad(self):
-        return [2 * self.x * self.gy]
+        ret = [2 * self.x * self.gy]
+        if self.loss_scale is not None:
+            ret = [2 * self.x * self.gy * self.loss_scale]
```
beam2d: How about
```
grad = 2 * self.x * self.gy
if ...:
    grad *= self.loss_scale
return [grad]
```

tests/chainer_tests/test_function_node.py
```diff
@@ -617,10 +634,16 @@ def forward(self):
...
+        return ret
 
     def expected_double_grad(self):
-        return [2 * self.gy]
+        ret = [2 * self.gy]
```
beam2d: ditto

beam2d:
Thank you! LGTM.

beam2d:
Jenkins, test this please.

tests/chainer_tests/test_function_node.py
```diff
@@ -617,10 +634,16 @@ def forward(self):
...
+        grad = 2 * self.x * self.gy
+        if self.loss_scale is not None:
+            grad *= self.loss_scale
+        return [grad]
 
     def expected_double_grad(self):
-        return [2 * self.gy]
+        ggrad = 2 * self.gy
+        if self.loss_scale is not None:
+            ggrad *=  self.loss_scale
```
beam2d: Remove one space after `*=`

anaruse: Ah, sorry, I forgot to run flake8...

beam2d:
Jenkins, test this please

PR #3558, opened 2017-10-13
Define dummy cuda.cupy when cupy is not available

niboshi:
Currently you can't write `xp is cuda.cupy` because `cuda.cupy` is not defined if CuPy is not installed.

This PR defines dummy `cuda.cupy` even when it's not installed.


okuta:
LGTM!

PR #3561, opened 2017-10-14
Always cast all inputs to given dtype in gradient check

hvy:
When passing the `no_grads` argument to the gradient check while at the same time specifying the `dtype` argument, all inputs are cast to the given `dtype` except the ones ignored by `no_grads`.
dtypes of function arguments are thus occasionally mixed, in particular, if a function that is being tested for correct gradients uses a kernel, an error is raises on mismatched dtypes on the kernel call.

This PR fixes this issue by always casting all inputs if `dtype` is specified no matter the value of `no_grads`.

### DIFF
--- a/chainer/gradient_check.py
+++ b/chainer/gradient_check.py
@@ -230,6 +230,9 @@ def check_backward(func, x_data, y_grad, params=(),
     .. seealso::
        :func:`numerical_grad`
     """
+    if dtype is not None and numpy.dtype(dtype).kind != 'f':
+        raise ValueError('`dtype` is allowed only float type')
+
     x_data = _as_tuple(x_data)
     if y_grad is not None:
         y_grad = _as_tuple(y_grad)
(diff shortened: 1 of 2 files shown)


chainer/gradient_check.py
```diff
@@ -269,8 +272,9 @@ def check_backward(func, x_data, y_grad, params=(),
         no_grads = [x.dtype.kind != 'f' for x in xs]
     else:
         if len(no_grads) != len(xs):
-            raise ValueError(
-                'Length of no_grads param and xs should be same.')
+            msg = ('Length of no_grads param and xs should be same.\n'
+                   'Actual: {0} != {1}'.format(len(no_grads), len(xs)))
+            raise ValueError(msg)
```
niboshi: You don't have to assign a temporary variable here.

tests/chainer_tests/test_gradient_check.py
```diff
@@ -358,7 +358,7 @@ def f(x, y):
 
     def test_no_grads_option(self):
         x1 = numpy.array([1], dtype='f')
-        x2 = numpy.array([1], dtype='f')  # grad check for this is skipped
+        x2 = numpy.array([1], dtype='f')
```
niboshi: Why did you remove the comment?

chainer/gradient_check.py
hvy: Thanks, I removed the redundant variable.

tests/chainer_tests/test_gradient_check.py
hvy: Fixed.

niboshi:
Please fix conflict.

hvy:
jenkins, test this please.

hvy:
jenkins, test this please.

niboshi:
LGTM!

PR #3573, opened 2017-10-16
New style pad_sequence

hvy:
New style pad_sequence to support double backprop.

Note that we still need the check `if gy.size == 0:` in order to make it compatible with NumPy 1.9 which changes the shape of the array in case of size 0.


hvy:
Rebased on top of master and adjusted tolerances.

takagi:
LGTM!

PR #3575, opened 2017-10-16
Use a distinct RandomState in the prefetch thread of MultiProcessIterator

grafi-tt:
Numpy shares the internal state of `numpy.random` across all threads. So if we use `numpy.random` from non-main thread, random number generation will be non-deterministic, dependent on context switch timing.

It hurts reproducibility. This PR fix the problem, by creating a distinct `RandomState` for each prefetch loop.

### DIFF
--- a/chainer/iterators/multiprocess_iterator.py
+++ b/chainer/iterators/multiprocess_iterator.py
@@ -280,6 +280,14 @@ class _PrefetchLoop(object):
         self._allocate_shared_memory()
         self._pool = None
 
+        # Use a distinct RandomState in the thread
+        # for deterministic random number generation.
+        # To support 32-bit platform and numpy < 1.11,
+        # the seed is taken in a verbose manner.
+        seed = numpy.asscalar(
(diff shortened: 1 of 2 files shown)


grafi-tt:
@takagi Thank you for your review. Fixed.

takagi:
jenkins test this please.

takagi:
I will merge the change after CI will pass.

takagi:
Would you rebase the master so that the Travis failure passes. The problem is now fixed in the latest master.

grafi-tt:
@takagi Thanks. Now the branch is rebased on the master branch (with my commits squashed to a single commit) and it passed CI.

takagi:
jenkins test this please.

takagi:
Thanks! I have merged.

takagi:
@grafi-tt Would you serialize the RandomState as well so that `MultiProcessIterator` can reproduce the order even when it repeats across serialization.

PR #3578, opened 2017-10-17
New style square function

kmaehashi:
This PR implements new style `square` function.

I also fixed `unary_math_function_test` decorator to support testing FunctionNode.
I think this is transitional; once https://github.com/chainer/chainer/pull/3499 got merged, it is better to rewrite tests with the new function test template.

### DIFF
--- a/chainer/functions/math/square.py
+++ b/chainer/functions/math/square.py
@@ -1,10 +1,10 @@
 from chainer import cuda
-from chainer import function
+from chainer import function_node
 from chainer import utils
 from chainer.utils import type_check
 
 
-class Square(function.Function):
+class Square(function_node.FunctionNode):
 
     @property
     def label(self):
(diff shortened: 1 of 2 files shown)


takagi:
We should follow API compat policy for `unary_math_function_test`, so my previous comment does not suit.

kmaehashi:
Ok, thank you for confirmation!

takagi:
It would be more natural for Chainer to let `unary_math_function_test` take a Python function instead of a `FunctionNode` object as @beam2d mentioned on slack.

https://github.com/chainer/chainer/compare/master...takagi:fix-test-math-function

Here `unary_math_function_test` accept a `Function` object just for backward compatibility and we would be able to deprecate it in the future major bump. (I will update the docstring shortly)

How about this?

kmaehashi:
Thank you for the suggestion, it looks good to me.

takagi:
I posted a PR to fix `unary_math_function_test` to support new style functions as #3603. Would you rebase your change on top of it after it will have been merged into the master? Before that, you can leave this PR.

kmaehashi:
Rebased.

takagi:
LGTM!

PR #3590, opened 2017-10-18
Replace the test framework with PyTest

mitmul:
#3324 has been reverted by #3589 because it included some unrelated changes.
So this PR is for re-applying the changes only related to the purpose which is to use PyTest.


niboshi:
LGTM!

PR #3599, opened 2017-10-19
Fix test_init_docstring

niboshi:
* According to the [documentation](https://docs.python.org/3.6/library/pkgutil.html#pkgutil.walk_packages), `pkgutil.walk_packages` accepts a list as `path` argument, so I fixed accordingly.
* `modname` is not an absolute name. I fixed how modules are imported.


takagi:
LGTM!

PR #3603, opened 2017-10-19
Fix math function testing helper to support new style functions.

takagi:
Rel #3578.

Currently `unary_math_function_unittest` decorator, that is a testing helper for math function layers, doesn't support testing new style functions introduced in Chainer v3. This PR fixes the decorator so that it can be used for the new style functions as well as the old style ones for backward compatibility.


kmaehashi:
LGTM!

PR #3604, opened 2017-10-19
Remove requirements for unit testing

niboshi:
Note: **Even though I put no-compat label, this PR does not break backward-compatibility.** Backward compatibility was broken when testing framework was replaced with PyTest. I put the label simply to signify that.

This PR removes pytest and mock requirements from Chainer normal installation.
That would reduce the possibility of trouble for users resulting from migration to PyTest.

OTOH, some parts of `chainer.testing` package depend on `pytest` and `mock`. Simply removing requirements would result in `ImportError`s when importing `chainer.testing`. To avoid that, I introduced `is_available()` functions to such modules.

This PR should be backported, because PyTest has introduced to stable branch as well.

### DIFF
--- a/chainer/testing/attr.py
+++ b/chainer/testing/attr.py
@@ -1,9 +1,41 @@
-import pytest
+try:
+    import pytest
+    _error = None
+except ImportError as e:
+    _error = e
 
-gpu = pytest.mark.gpu
-cudnn = pytest.mark.cudnn
-slow = pytest.mark.slow
+
+def is_available():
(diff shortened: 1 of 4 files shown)


mitmul:
LGTM. waiting for the Jenkins test finishes

mitmul:
It passed all Jenkins tests

takagi:
No backport is made for this? I have posted #3682.

niboshi:
Thank you!

PR #3615, opened 2017-10-20
new style rdiv function

kmaehashi:
This PR implements new style `rdiv` function.

I also fixed a wrong label of `DivFromConstant`.
The label says `variable / C` but actually it computes `C / variable`.


kmaehashi:
Rebased.

kmaehashi:
Remove unneeded `repeat` decorator and confirmed 10000 repeats.

PR #3617, opened 2017-10-20
new style minimum function

kmaehashi:
This PR implements new style `minimum` function.

### DIFF
--- a/chainer/functions/math/minimum.py
+++ b/chainer/functions/math/minimum.py
@@ -31,21 +31,25 @@ class Minimum(function_node.FunctionNode):
 
     def backward(self, indexes, grad_outputs):
         x1, x2 = self.get_retained_inputs()
-        return MinimumGrad().apply((x1, x2, grad_outputs[0]))
+        return MinimumGrad(x1.data, x2.data).apply((grad_outputs[0],))
 
 
 class MinimumGrad(function_node.FunctionNode):
 
+    def __init__(self, x1, x2):
+        self.x1 = x1
+        self.x2 = x2
+
     def forward_cpu(self, inputs):
-        self.retain_inputs((0, 1))
-        x1, x2, gy = inputs
+        gy, = inputs
+        x1, x2 = self.x1, self.x2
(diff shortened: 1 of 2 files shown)


takagi:
Travis fails on macOS Python 3.4 case at least because of the problem fixed in #3599. Would you rebase the master?

kmaehashi:
Rebased.

chainer/functions/math/minimum.py
```diff
@@ -46,6 +56,13 @@ def backward_gpu(self, inputs, grads):
             'minimum_bwd2')(x1, x2, gy)
         return gx1, gx2
 
+    def backward(self, indexes, grad_outputs):
+        x1, x2 = self.get_retained_inputs()
+        ggx = grad_outputs[0] * 0 + grad_outputs[1] * 0
```
takagi: Name `gx` would be better?

chainer/functions/math/minimum.py
```diff
@@ -46,6 +56,13 @@ def backward_gpu(self, inputs, grads):
...
+        cond = utils.force_array(x1.data <= x2.data)
+        gggy = chainer.functions.where(cond, grad_outputs[0], grad_outputs[1])
```
takagi: Name `ggy` would be better?

chainer/functions/math/minimum.py
takagi: Memory spaces are allocated for each expression result here?

takagi:
Would you rebase the master?

chainer/functions/math/minimum.py
```diff
@@ -18,24 +19,33 @@ def check_type_forward(self, in_types):
         )
 
     def forward_cpu(self, inputs):
```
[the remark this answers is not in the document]
kmaehashi: Fixed.

kmaehashi:
Thanks for review, rebased and fixed accordingly.

takagi:
LGTM!

PR #3623, opened 2017-10-20
remove nose dependency in tests

kmaehashi:
`run_module` function still depends on `nose`.
I changed to use `pytest`.

### DIFF
--- a/chainer/testing/__init__.py
+++ b/chainer/testing/__init__.py
@@ -1,5 +1,3 @@
-import pytest
-
 from chainer.testing import array  # NOQA
 from chainer.testing import helper  # NOQA
 from chainer.testing import parameterized  # NOQA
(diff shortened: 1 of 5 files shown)


chainer/testing/__init__.py
```diff
@@ -1,4 +1,4 @@
-import nose
+import pytest
```
niboshi: This line will cause `ImportError` when importing `chainer.testing` in the environment without `pytest`. (pytest is no longer a requirement for normal chainer installation).
How about importing `pytest` in `run_module` function?

kmaehashi: Fixed as suggested!

kmaehashi:
Removed doctest from the decorator as `pytest` does not support interactive test run.
https://stackoverflow.com/questions/26840034/can-tests-with-pytest-fixtures-be-run-interactively

takagi:
LGTM!

PR #3625, opened 2017-10-20
fix contribution guide for test framework change

kmaehashi:
As test framework is changed from nose to pytest, we need to update contrib guide.

This PR documents the behavior of new decorator proposed in #3624.


okuta:
LGTM!

PR #3627, opened 2017-10-21
Fix document directive typo

okuta:
We cannot compile document.
Please merge as soon as possible.


takagi:
LGTM!

niboshi:
https://github.com/chainer/chainer/pull/3138 is labeled as to-be-backported.
If it is backported, this PR should be backported, too.

takagi:
I checked only v3 branch and missed #3138 not backported yet.

PR #3645, opened 2017-10-23
introduce F.fix and make F.fmod function new style

kmaehashi:
This PR implements new style `fmod` function.

This PR also contains new function `F.fix` because it is needed in fmod backward.
(If it should be another PR, please comment)

### DIFF
--- a/chainer/functions/__init__.py
+++ b/chainer/functions/__init__.py
@@ -190,6 +190,8 @@ from chainer.functions.math.exponential import log2  # NOQA
 from chainer.functions.math.exponential import Log2  # NOQA
 from chainer.functions.math.exponential_m1 import expm1  # NOQA
 from chainer.functions.math.exponential_m1 import Expm1  # NOQA
+from chainer.functions.math.fix import fix  # NOQA
+from chainer.functions.math.fix import Fix  # NOQA
 from chainer.functions.math.floor import floor  # NOQA
 from chainer.functions.math.floor import Floor  # NOQA
 from chainer.functions.math.fmod import fmod  # NOQA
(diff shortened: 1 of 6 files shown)


chainer/functions/math/fix.py
```diff
@@ -0,0 +1,40 @@
+from chainer import cuda
+from chainer import function_node
+from chainer import utils
+from chainer.utils import type_check
+
+
+class Fix(function_node.FunctionNode):
```
okuta: Please remove this class.
Because `backward` of this class always returns `None`.

chainer/functions/math/fmod.py
```diff
@@ -20,16 +21,16 @@ def check_type_forward(self, in_types):
...
         return utils.force_array(m, x.dtype),
 
-    def backward(self, inputs, grad_outputs):
-        xp = cuda.get_array_module(*inputs)
+    def backward(self, indexes, grad_outputs):
+        x, divisor = self.get_retained_inputs()
         gw, = grad_outputs
-        x, divisor = inputs
-        return gw, utils.force_array(xp.fix(x / divisor) * -1 * gw, x.dtype)
+        return gw, chainer.functions.fix(x / divisor) * -1 * gw
```
okuta: How about `return gw, -chainer.functions.fix(x / divisor) * gw`.

chainer/functions/__init__.py
```diff
@@ -188,6 +188,8 @@
 from chainer.functions.math.exponential import Log2  # NOQA
 from chainer.functions.math.exponential_m1 import expm1  # NOQA
 from chainer.functions.math.exponential_m1 import Expm1  # NOQA
+from chainer.functions.math.fix import fix  # NOQA
+from chainer.functions.math.fix import Fix  # NOQA
```
okuta: Please remove this line.

kmaehashi:
Thank you for review, fixed accordingly.

okuta:
LGTM!

PR #3241, opened 2017-08-24
Reorganize tutorial & add new contents

Crissman:
PR for Enhancement #2576.

Intended to give the overall outline update, and links to the appropriate areas code for other works in progress.

PR #3662, opened 2017-10-24
document missing environment variables

kmaehashi:
Added description for undocumented environment variables.


kmaehashi:
I'll remove `CHAINER_CUDNN` from the doc.
As there are two env vars to do similar things (`CHAINER_CUDNN` and `CHAINER_USER_CUDNN`), I think we should consider deprecating `CHAINER_CUDNN`.

kmaehashi:
After reading the code for these environment variables, I rather think that `CHAINER_CUDNN` should not be removed.

`CHAINER_USE_CUDNN` is just a default value for config and may be overridden at runtime if the config is changed in the code.
`CHAINER_CUDNN` forces disabling cuDNN, so it sounds convenient for user developing cuDNN-capable code to confirm that the code will work without cuDNN.
cc/ @niboshi

niboshi:
OK.
Could you revise the description of each variable so that users can understand the difference?

kmaehashi:
OK.

TODO: add `CHAINER_USE_IDEEP`.

kmaehashi:
Fixed.

docs/source/reference/configuration.rst
```diff
@@ -68,7 +68,17 @@ Note that the default values are set in the global config.
...
+``chainer.config.use_ideep``
+   Flag to configure whether or not to use iDeep.
+
+       - If it is ``'always'``, Chainer will try to use iDeep everywhere if possible.
+       - If it is ``'auto'``, Chainer will use iDeep only if it is known that the usage does not degrade the performance.
+       - If it is ``'never'``, Chainer will never use iDeep anywhere.
+
+   The default value is given by ``CHAINER_USE_IDEEP`` environment variable (set to ``'always'``, ``'auto'`` or ``'never'``) if available, otherwise uses ``'never'``.
+
+   Note that in spite of the configuration, optimizers will use iDeep if and only if the chain is converted to iDeep (e.g., ``model.to_intel64()``).
```
niboshi: chain -> link
converted -> converted manually

kmaehashi: Thanks! Fixed.

niboshi:
Please fix jenkins error:
```
environment.rst:: WARNING: document isn't included in any toctree
```

kmaehashi:
Hm, list of environment variables has been merged into configuration page in #3241. I'll rebase to the master.

kmaehashi:
Rebased.

niboshi:
jenkins, test this please

niboshi:
LGTM!

PR #3674, opened 2017-10-25
run OS X test only on master/stable branch to avoid delay

kmaehashi:
As OS X container in Travis is heavy, this PR changes the config to run OS X matrix only on master/stable branch.

See also #3671 for discussion.


niboshi:
LGTM!

PR #3678, opened 2017-10-25
Add sign function

hvy:
Added the sign function with gradient 0 everywhere, mostly because it could be used when reimplementing Functions as FunctionNodes in the backward pass.

### DIFF
--- a/chainer/functions/__init__.py
+++ b/chainer/functions/__init__.py
@@ -227,6 +227,8 @@ from chainer.functions.math.minmax import Min  # NOQA
 from chainer.functions.math.prod import prod  # NOQA
 from chainer.functions.math.prod import Prod  # NOQA
 from chainer.functions.math.scale import scale  # NOQA
+from chainer.functions.math.sign import sign  # NOQA
+from chainer.functions.math.sign import Sign  # NOQA
 from chainer.functions.math.sqrt import rsqrt  # NOQA
 from chainer.functions.math.sqrt import sqrt  # NOQA
 from chainer.functions.math.sqrt import Sqrt  # NOQA
--- a/chainer/functions/__init__.py
+++ b/chainer/functions/__init__.py
@@ -228,7 +228,6 @@ from chainer.functions.math.prod import prod  # NOQA
 from chainer.functions.math.prod import Prod  # NOQA
 from chainer.functions.math.scale import scale  # NOQA
 from chainer.functions.math.sign import sign  # NOQA
-from chainer.functions.math.sign import Sign  # NOQA
 from chainer.functions.math.sqrt import rsqrt  # NOQA
 from chainer.functions.math.sqrt import sqrt  # NOQA
 from chainer.functions.math.sqrt import Sqrt  # NOQA
--- a/chainer/functions/math/sign.py
+++ b/chainer/functions/math/sign.py
@@ -1,36 +1,13 @@
-import numpy
-
+import chainer
 from chainer import cuda
-from chainer import function_node
 from chainer import utils
-from chainer.utils import type_check
-
-
-class Sign(function_node.FunctionNode):
-
-    """Element-wise sign function."""
-
-    @property
-    def label(self):
-        return 'sign'
-
-    def check_type_forward(self, in_types):
-        type_check.expect(in_types.size() == 1)
-
-    def forward_cpu(self, x):
-        return utils.force_array(numpy.sign(x[0])),
-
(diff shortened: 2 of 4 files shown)


chainer/functions/math/sign.py
```diff
@@ -0,0 +1,54 @@
...
+        type_check.expect(in_types.size() == 1)
+
+    def forward_cpu(self, x):
+        return utils.force_array(numpy.sign(x[0])),
+
+    def forward_gpu(self, x):
+        return cuda.cupy.sign(x[0]),
+
+    def backward(self, indexes, grad_outputs):
+        return grad_outputs[0] * 0,
```
okuta: Why don't you use `zeros_like`?

hvy: Thank you for the review. Don't we need to return a Variable object here?

okuta: Sorry, please return `None` in this case.

hvy: By doing so, double backprop won't work when this function is involved in the computational graph. Please correct me if I am wrong.

kmaehashi: I came across the same situation (just return zero) in #3539.  I'd like to know if there is a better way. (maybe ask @beam2d ?)

hvy: Both PyTorch and TensorFlow seem to return 0. This is the code for PyTorch. https://github.com/pytorch/pytorch/blob/3853d5da97ead031d7d77f39fe7ce2de243ce369/tools/autograd/derivatives.yaml#L436-L437

chainer/functions/__init__.py
```diff
@@ -226,6 +226,8 @@
 from chainer.functions.math.prod import prod  # NOQA
 from chainer.functions.math.prod import Prod  # NOQA
 from chainer.functions.math.scale import scale  # NOQA
+from chainer.functions.math.sign import sign  # NOQA
+from chainer.functions.math.sign import Sign  # NOQA
```
niboshi: Please remove `Sign`. We're not going to expose new function classes.

hvy: Fixed.

hvy:
After offline discussion, we've concluded that no class is needed for this function, and that the gradients should be `None` everywhere. I have changed the code accordingly.

niboshi:
If we break the graph, `computational_graph` feature will not work.
Is it really acceptable?

hvy:
Was this also considered? @okuta

hvy:
After offline discussions, it was concluded that the current implementation behaves according to intended design.

hvy:
Rebased on top of master.

tests/chainer_tests/functions_tests/math_tests/test_sign.py
```diff
@@ -0,0 +1,101 @@
...
+    @property
+    def no_grads(self):
+        return (False,)
+
+    def make_data(self):
+        raise NotImplementedError
+
+    def setUp(self):
+        self.eps = 1e-3
+        while True:
```
okuta: Is this necessary?

tests/chainer_tests/functions_tests/math_tests/test_sign.py
```diff
@@ -0,0 +1,101 @@
...
+
+import chainer
+from chainer import cuda
+import chainer.functions as F
+from chainer import gradient_check
+from chainer import testing
+from chainer.testing import attr
+
+
+class UnaryFunctionsTestBase(unittest.TestCase):
```
okuta: How about merging UnaryFunctionsTestBase and TestSign

hvy: Merged the classes.

tests/chainer_tests/functions_tests/math_tests/test_sign.py
hvy: No sorry, this is not necessary for the sign function. Removed!

hvy:
@okuta san, thank you for the review. I fixed the code accordingly.

takagi:
jenkins, test this please.

takagi:
jenkins, test this please.

takagi:
LGTM!

takagi:
I made a PR #3910 to adjust `convolution_nd`'s testing tolerance which makes CI fail.

hvy:
@takagi san, thank you. I'll take a look at the tolerance fix..

PR #3684, opened 2017-10-26
fix coveragerc to measure branch coverage and only target chainer module

kmaehashi:
See https://github.com/chainer/chainer-test/pull/347


okuta:
LGTM!

PR #3685, opened 2017-10-26
Remove PyTest global option configuration

niboshi:
Removes 2 PyTest options globallly set in `setup.cfg`.

- `--maxfail=2`: This should be controlled by the invoker. Also I think CIs should display all errors.
- `--showlocal`: This option dumps local variables. Appearance of the output should be controlled by the invoker.

(Here invoker can be either a CI agent or a developer who invokes pytest manually)


kmaehashi:
:+1: LGTM!

PR #3687, opened 2017-10-26
v3: Fix CuPy intersphinx mapping

niboshi:
I think Chainer stable documentation should link to CuPy stable documentation.


niboshi:
(Related to: https://github.com/cupy/cupy/pull/655)

kmaehashi:
LGTM!

PR #3696, opened 2017-10-26
New style triplet

hvy:
New style triplet loss to support double backprop.

### DIFF
--- a/chainer/functions/loss/triplet.py
+++ b/chainer/functions/loss/triplet.py
@@ -1,11 +1,12 @@
 import numpy
 
+import chainer
 from chainer import cuda
-from chainer import function
+from chainer import function_node
 from chainer.utils import type_check
 
 
-class Triplet(function.Function):
+class Triplet(function_node.FunctionNode):
 
     """Triplet loss function."""
 
@@ -46,27 +47,30 @@ class Triplet(function.Function):
             loss = xp.sum(self.dist_hinge) / N
         else:
             loss = self.dist_hinge
+
+        self.retain_inputs((0, 1, 2))
         return xp.array(loss, dtype=numpy.float32),
 
-    def backward(self, inputs, gy):
-        xp = cuda.get_array_module(*inputs)
+    def backward(self, indexes, grad_outputs):
+        anchor, positive, negative = self.get_retained_inputs()
 
-        anchor, positive, negative = inputs
         N = anchor.shape[0]
-
         x_dim = anchor.shape[1]
+
+        xp = cuda.get_array_module(anchor)
         tmp = xp.repeat(self.dist_hinge[:, None], x_dim, axis=1)
         mask = xp.array(tmp > 0, dtype=numpy.float32)
 
+        gy, = grad_outputs
         if self.reduce == 'mean':
-            g = gy[0] / N
+            g = gy / N
         else:
-            g = gy[0][:, None]
+            g = gy[:, None]
 
-        tmp = 2 * g * mask
-        gx0 = (tmp * (negative - positive)).astype(numpy.float32)
-        gx1 = (tmp * (positive - anchor)).astype(numpy.float32)
-        gx2 = (tmp * (anchor - negative)).astype(numpy.float32)
+        tmp = 2 * chainer.functions.broadcast_to(g, mask.shape) * mask
+        gx0 = tmp * (negative - positive)
+        gx1 = tmp * (positive - anchor)
+        gx2 = tmp * (anchor - negative)
 
         return gx0, gx1, gx2
 
@@ -137,4 +141,4 @@ def triplet(anchor, positive, negative, margin=0.2, reduce='mean'):
         variable(0.4400000274181366)
 
     """
-    return Triplet(margin, reduce)(anchor, positive, negative)
+    return Triplet(margin, reduce).apply((anchor, positive, negative))[0]
--- a/chainer/functions/loss/triplet.py
+++ b/chainer/functions/loss/triplet.py
@@ -68,11 +68,16 @@ class Triplet(function_node.FunctionNode):
             g = gy[:, None]
 
         tmp = 2 * chainer.functions.broadcast_to(g, mask.shape) * mask
-        gx0 = tmp * (negative - positive)
-        gx1 = tmp * (positive - anchor)
-        gx2 = tmp * (anchor - negative)
 
-        return gx0, gx1, gx2
+        ret = []
+        if 0 in indexes:
+            ret.append(tmp * (negative - positive))
+        if 1 in indexes:
+            ret.append(tmp * (positive - anchor))
+        if 2 in indexes:
+            ret.append(tmp * (anchor - negative))
+
+        return ret
 
 
 def triplet(anchor, positive, negative, margin=0.2, reduce='mean'):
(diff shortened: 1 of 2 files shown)


chainer/functions/loss/triplet.py
```diff
@@ -46,29 +47,37 @@ def forward(self, inputs):
...
+    def backward(self, indexes, grad_outputs):
+        anchor, positive, negative = self.get_retained_inputs()
 
-        anchor, positive, negative = inputs
         N = anchor.shape[0]
-
         x_dim = anchor.shape[1]
+
+        xp = cuda.get_array_module(anchor)
         tmp = xp.repeat(self.dist_hinge[:, None], x_dim, axis=1)
```
takagi: `backward` method must be implemented in `chainer.Variable` level with `chainer.functions` instead of NumPy/CuPy level operations so that `Triplet` supports double backward, so we can not use `self.dist_hinge` here but need to compute it again with the retained inputs (or implement as a grad FunctionNode, `TripletGrad`, so that we can implement the gradient in NumPy/CuPy level).

chainer/functions/loss/triplet.py
```diff
@@ -46,29 +47,37 @@ def forward(self, inputs):
...
         mask = xp.array(tmp > 0, dtype=numpy.float32)
```
takagi: We need to get `mask` (and `tmp`) in `chainer.Variable` level with `chainer.functions` instead of NumPy/CuPy level operations in `backward` method.

tests/chainer_tests/functions_tests/loss_tests/test_triplet.py
```diff
@@ -30,6 +30,9 @@ def setUp(self):
         else:
             gy_shape = (self.batchsize,)
```
takagi: Would you remove this unnecessary parentheses?

tests/chainer_tests/functions_tests/loss_tests/test_triplet.py
```diff
@@ -89,6 +95,30 @@ def test_backward_gpu_no_cudnn(self):
...
+        def f(a, p, n):
+            y = functions.triplet(
+                a, p, n, margin=self.margin, reduce=self.reduce)
+            return y * y
+
+        gradient_check.check_double_backward(
+            f, (a_data, p_data, n_data), gy_data,
+            (gga_data, ggp_data, ggn_data), rtol=1e-4, atol=1e-4)
+
+    @condition.retry(10)
```
takagi: So unstable that we need to retry ten times? How about three times or so with proper tolerance?

tests/chainer_tests/functions_tests/loss_tests/test_triplet.py
```diff
@@ -89,6 +95,30 @@ def test_backward_gpu_no_cudnn(self):
...
+    def test_double_backward_cpu(self):
+        self.check_double_backward(
+            self.a, self.p, self.n, self.gy, self.gga, self.ggp, self.ggn)
+
+    @attr.gpu
+    @condition.retry(10)
```
takagi: So unstable that we need to retry ten times? How about three times or so with proper tolerance?

tests/chainer_tests/functions_tests/loss_tests/test_triplet.py
```diff
@@ -89,6 +95,30 @@ def test_backward_gpu_no_cudnn(self):
         self.check_backward(cuda.to_gpu(self.a), cuda.to_gpu(self.p),
                             cuda.to_gpu(self.n), cuda.to_gpu(self.gy))
 
+    def check_double_backward(self, a_data, p_data, n_data, gy_data, gga_data,
+                              ggp_data, ggn_data):
+        def f(a, p, n):
+            y = functions.triplet(
+                a, p, n, margin=self.margin, reduce=self.reduce)
+            return y * y
```
takagi: Is triplet loss linear? I suppose that this squaring is not needed.

chainer/functions/loss/triplet.py
hvy: We only use `dist_hinge` to mask the upstream gradients, i.e. I don't thinks it is necessary to wrap them in `Variables` since there are no gradients being propagated. Please correct me if I am wrong.

chainer/functions/loss/triplet.py
hvy: Please see above comments.

chainer/functions/loss/triplet.py
```diff
@@ -46,29 +47,37 @@ def forward(self, inputs):
...
+        ret = []
+        if 0 in indexes:
+            ret.append(tmp * (negative - positive))
+        if 1 in indexes:
+            ret.append(tmp * (positive - anchor))
+        if 2 in indexes:
+            ret.append(tmp * (anchor - negative))
 
-        return gx0, gx1, gx2
+        return ret
```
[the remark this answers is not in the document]
hvy: Returning a list seems to be a common pattern, e.g. [mse](https://github.com/chainer/chainer/blob/9453976fcb6470fba6d7945d9737a3021e6b0284/chainer/functions/loss/mean_squared_error.py).

tests/chainer_tests/functions_tests/loss_tests/test_triplet.py
hvy: Fixed.

tests/chainer_tests/functions_tests/loss_tests/test_triplet.py
hvy: Thanks, fixed!

tests/chainer_tests/functions_tests/loss_tests/test_triplet.py
hvy: Updated tol and removed retries.

tests/chainer_tests/functions_tests/loss_tests/test_triplet.py
hvy: Updated tol and removed retries.

hvy:
Thank you @takagi san for the review. Please see my comments and fixes.

chainer/functions/loss/triplet.py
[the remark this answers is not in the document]
takagi: The spec says that `FunctionNode.backward` method returns a tuple of Variables ... but, as you say, other `FunctionNode`s, even `Linear`, return a list. Okay, I drop this comment and will fix such `FunctionNode`s if needed.

chainer/functions/loss/triplet.py
takagi: Okay. I got what you intend and it is well considered. It should work, but I still have a bit concern if this style is valid in  `backward` method. Can I confirm it to @beam2d ?

takagi:
Would you fix to pass flake8 on Travis?

hvy:
@takagi san, thank you again for the review and for confirming the usage of the precomputed ndarray.

takagi:
jenkins, test this please.

chainer/functions/loss/triplet.py
takagi: I confirmed that to @beam2d and it is okay as you say that the masking itself makes no effect on propagation. Sorry for bothering you.

takagi:
Would you resolve the conflict? It can be merged after that.

hvy:
@takagi Fixed!

takagi:
LGTM!

PR #3697, opened 2017-10-26
Fix math function test helper to support double backward test of line…

### LINKED ISSUE #3677: unary_math_function_test with non-twice differentiable functions fails
Is it possible to allow the `unary_math_function_test` helper to execute different functions when computing the forward passes for the double backward test?

At the moment, a test will fail if the tested function cannot be differentiated twice. We usually get around this problem by writing a wrapper function that makes it non-linear . Maybe we can achieve the same result this helper that I am not aware of?

There probably exists a better solution, than allowing a different function.

takagi:
Fix #3677.

`unary_math_function_unittest` decorator, which is a testing helper for math functions, now supports testing new style functions including double backward test in #3603. But it cannot perform double backward test for linear functions, which needs some getting around as described below.

https://gist.github.com/beam2d/dd87d97b8d7581f95cd19b1de86c9f4d#testing-the-new-style-function-that-supports-double-backward

This PR introduces `is_linear` argument to the decorator that tells it the target function is linear so that it can perform double backward test for linear functions.

### DIFF
--- a/chainer/testing/unary_math_function_test.py
+++ b/chainer/testing/unary_math_function_test.py
@@ -48,8 +48,15 @@ def _make_data_default(shape, dtype):
     return x, gy, ggx
 
 
+def _nonlinear(func):
+    def aux(x):
+        y = func(x)
+        return y * y
+    return aux
+
+
 def unary_math_function_unittest(func, func_expected=None, label_expected=None,
-                                 make_data=None):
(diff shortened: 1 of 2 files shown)


hvy:
jenkins, test this please.

hvy:
jenkins, test this please.

hvy:
LGTM!

PR #3707, opened 2017-10-27
New style absolute_error

hvy:
New style abolute_error to support double backprop.


hvy:
Rebased on top of master.

hvy:
Removed retries and added this function to the docs (it was missing).

takagi:
Would you rebase on top of the master so that #3837 is included?

takagi:
LGTM, will merge after rebase and CI passes.

PR #3708, opened 2017-10-27
Support parameter update in fp32

anaruse:
This PR adds support of parameter update in fp32.

At present, when the data type of input data is fp16, the data type of parameters (like weights of convolution layers, fully-connected layers, etc.) become fp16 in Chainer. And these paramters are to be updated in fp16 at the end of each training iteration. However, it is known that this entirely fp16 trained models tend to be less accurate than fp32 trained models (see https://arxiv.org/abs/1710.03740 for details).

The purpose of this PR is to match the accuracy to fp32 model even when input data is fp16, and to enable use of the technique "parameter update in fp32" from Chainer. You can use this feature using a method `use_fp32_update` like below.

    optimizer.setup(model)
    optimizer.use_fp32_update()

(*) This PR is a split PR from #3544.


chainer/optimizer.py
```diff
@@ -185,11 +185,31 @@ def update(self, param):
             return
 
         self.t += 1
-        if param.data is not None:
-            self._prepare(param)
-        for hook in six.itervalues(self._hooks):
-            hook(self, param)
-        self.update_core(param)
+
+        use_fp32_update = getattr(self, "_use_fp32_update", False)
```
beam2d: Adding the default values of `_use_fp32_update` and `fp32_param` as class fields (or instance fields initialized in `__init__`) is simpler:
```py
_use_fp32_update = False
_fp32_param = None

def update(self, param):
    ...
    if self._use_fp32_update and param.dtype == numpy.float16:
        if self._fp32_param is None:
            ...
```

chainer/optimizer.py
```diff
@@ -185,11 +185,31 @@ def update(self, param):
...
+        if param.data.dtype == numpy.float16 and use_fp32_update:
+            fp32_param = getattr(self, "_fp32_param", None)
+            if fp32_param is None:
+                self._fp32_param = variable.Variable(
+                    param.data.astype(numpy.float32),
```
beam2d: Use `.array` instead of `.data`.

chainer/optimizer.py
```diff
@@ -486,6 +525,8 @@ def setup(self, link):
         super(GradientMethod, self).setup(link)
         for param in link.params():
             param.update_rule = self.create_update_rule()
+            if getattr(self, "_use_fp32_update", False):
```
beam2d: ditto (set a class or instance field by default)

anaruse:
Thank you for your comments, @beam2d 

I've just updated the PR in the way you suggested. Please check.

takagi:
LGTM.

takagi:
The backport is to be made after #3936 will be merged.

delta2323:
@takagi I wanted to backport #4159, but failed as this PR has not been backported. Is it OK to do so, as #3936 is closed?

PR #3718, opened 2017-10-28
[backport] Use a distinct RandomState in the prefetch thread of MultiProcessIterator

takagi:
The backport of #3575.


okuta:
Is this change acceptable in minor version up?

takagi:
Your pointing out makes sense. MultiProcessIterator takes side-effect from`numpy.random` 's state  and #3575 changes the behavior from the side-effect even in deterministic manner. As you mention, the original PR should be no-compat and we should leave it not backported.

PR #3737, opened 2017-10-30
New style clip

hvy:
New style clip to support double backprop.

### DIFF
--- a/tests/chainer_tests/functions_tests/math_tests/test_clip.py
+++ b/tests/chainer_tests/functions_tests/math_tests/test_clip.py
@@ -8,7 +8,6 @@ from chainer import functions
 from chainer import gradient_check
 from chainer import testing
 from chainer.testing import attr
-from chainer.testing import condition
 
 
 @testing.parameterize(*testing.product({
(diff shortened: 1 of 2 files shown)


niboshi:
Double backward test is unstable. Please rebase and fix it.

hvy:
@niboshi Thanks, fixed!

tests/chainer_tests/functions_tests/math_tests/test_clip.py
```diff
@@ -62,6 +67,24 @@ def test_backward_cpu(self):
...
 
+    def check_double_backward(self, x_data, y_grad, gx_grad):
+        def f(x):
+            y = functions.clip(x, self.x_min, self.x_max)
+            return y * y
+
+        gradient_check.check_double_backward(
+            f, x_data, y_grad, gx_grad, dtype=numpy.float64)
+
+    @condition.retry(5)
```
niboshi: Please do not use `retry`. It can hide serious errors.

hvy: Okay, I'll change the tolerances instead of using retry. More specifically, what types of errors do you mean and does this only apply to double backward tests?

niboshi: No, it doesn't only apply to double backward.
Various kind of errors are possible. One example is that the backward computation could be implemented incorrectly only in limited range of input domain.

hvy:
@niboshi Removed all retries and adjusted the tolerance.

niboshi:
LGTM!

PR #3741, opened 2017-10-30
Allow to_gpu and to_cpu to accept NumPy scalars

### LINKED ISSUE #3740: cuda.to_gpu does not work for numpy scalar types
`cuda.to_gpu` only accepts `numpy.ndarray` or `cupy.ndarray`.
However, as NumPy seems to use `numpy.ndarray` and scalar types (such as `numpy.float32`) interchangeably, it would be preferrable if it accepts scalar types as well.

### Code:
```
import chainer, numpy

a = numpy.array(3, numpy.float32)
print(type(a), a.shape, a.dtype)

b = a + a
print(type(b), b.shape, b.dtype)

x = chainer.cuda.to_gpu(b)
```

### Result:
```
<class 'numpy.ndarray'> () float32
<class 'numpy.float32'> () float32
  File "test-numpy.py", line 9, in <module>
    x = chainer.cuda.to_gpu(b)
  File "/data/work/w/repos/chainer/chainer/cuda.py", line 259, in to_gpu
    '\nActual type: {0}.'.format(type(array)))
TypeError: The array sent to gpu must be numpy.ndarray or cupy.ndarray.
Actual type: <class 'numpy.float32'>.
```

niboshi:
Fixes #3740

### DIFF
--- a/chainer/cuda.py
+++ b/chainer/cuda.py
@@ -253,7 +253,7 @@ def to_gpu(array, device=None, stream=None):
             'Please remove it.', DeprecationWarning)
 
     check_cuda_available()
-    if isinstance(array, (numpy.number, numpy.bool_, numpy.complex)):
+    if isinstance(array, (numpy.number, numpy.bool_)):
         array = numpy.asarray(array)
     if not isinstance(array, (cupy.ndarray, numpy.ndarray)):
         raise TypeError(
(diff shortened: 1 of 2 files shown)


chainer/cuda.py
```diff
@@ -253,9 +253,12 @@ def to_gpu(array, device=None, stream=None):
             'Please remove it.', DeprecationWarning)
 
     check_cuda_available()
+    if isinstance(array, (numpy.number, numpy.bool_, numpy.complex)):
```
okuta: `numpy.number` includes `numpy.complex`.

chainer/cuda.py
```diff
@@ -312,11 +315,14 @@ def to_cpu(array, stream=None):
         check_cuda_available()
         with get_device_from_array(array):
             return array.get(stream)
+    elif isinstance(array, (numpy.number, numpy.bool_, numpy.complex)):
```
okuta: ditto

chainer/cuda.py
niboshi: Fixed!

okuta:
LGTM

PR #3742, opened 2017-10-30
Fix doc reference

### LINKED ISSUE #1587: Replace function with chainer.Function in docstrings
Some arguments of functions are wrongly typed in docstrings. For example, this [`activation`](https://github.com/pfnet/chainer/blob/v1.14.0/chainer/links/connection/mlp_convolution_2d.py#L30) argument's type should be `chainer.Function` but it is `function`. 

This generates wrong links to http://docs.chainer.org/en/stable/reference/function_hooks.html#module-chainer.function in documents.

hvy:
Fixed wrong references in the docs for the arguent type `function`.

This PR partially fixes #1587 .


niboshi:
I think `activation` argument accepts a Python function, doesn't it?
Not `FunctionNode` at least.

hvy:
Thanks for the comment. Yes that is absolutely correct, I'll close this PR.

niboshi:
I think this should be fixed in some way, though.

PR #3744, opened 2017-10-30
Add test cases for various shapes as function input

### LINKED ISSUE #3701: Chainer functions do not work with arrays with shape=()
### Code:
```
import chainer, numpy

class Func(chainer.FunctionNode):
    def forward(self, inputs):
        x, = inputs
        y = x + 2
        assert isinstance(y, numpy.float64)  # !
        return y,

        gy, = grad_outputs
        return gy,

x = numpy.ones((), 'f')
y = Func().apply((x,))

print(y)
```
(Same for `chainer.Function` functions)

### Result:
```
  File "test-func.py", line 15, in <module>
    y = Func().apply((x,))
  File "/data/work/w/repos/chainer/chainer/function_node.py", line 241, in apply
    for y in outputs])
  File "/data/work/w/repos/chainer/chainer/function_node.py", line 241, in <listcomp>
    for y in outputs])
  File "/data/work/w/repos/chainer/chainer/variable.py", line 454, in __init__
    raise TypeError(msg)
TypeError: numpy.ndarray or cuda.ndarray are expected.
Actual: <class 'numpy.float64'>
```

### Cause:
As written in the code, even if `x` is `numpy.ndarray`, `x + 2` results in `numpy.float64` instead of `numpy.ndarray`.

### Version:
current master (97ed424af9579c1a6ac5ded6f3e5c5c29985fe33)

niboshi:
Fixes #3701 
~~Review after #3741~~ (merged)


okuta:
Please rebase this PR.

niboshi:
Rebased.

chainer/function_node.py
```diff
@@ -245,6 +245,12 @@ def apply(self, inputs):
             outputs = self.forward(in_data)
             assert type(outputs) is tuple
 
+        # Convert NumPy scalars to numpy.ndarray
+        outputs = tuple([
```
okuta: I think this convert is redundant.
Currently, a `forward` function must return `ndarray` ,

okuta: How about moving to https://github.com/chainer/chainer/pull/3744/files#diff-3480da470ca3b18f6b879875999e6206L253

niboshi: It that for performance?
Well, indeed I confirmed this PR degrades the performance of `apply` by >5%.

I'll update so that it will generate a nice error with debug mode.

niboshi:
Fixed.

chainer/function_node.py
okuta: Thanks

okuta:
flake8

niboshi:
Fixed.

tests/chainer_tests/test_function_node.py
```diff
@@ -125,12 +140,12 @@ def check_check_type_forward(self):
...
         t1 = ts[0]
-        self.assertEqual(t1.shape, (3,))
-        self.assertEqual(t1.dtype, numpy.float32)
+        assert t1.shape == self.x_shape
+        assert t1.dtype == numpy.float32
 
         t2 = ts[1]
-        self.assertEqual(t2.shape, (3,))
-        self.assertEqual(t2.dtype, numpy.int32)
+        assert t2.shape == self.x_shape
```
okuta: Why do you use `assert`?

okuta: Will we use `assert` in new test?

okuta:
Please check test result. @niboshi

okuta:
jenkins, test this please.

niboshi:
I guess this ndarray type check is already done in master.
I will check if the test is written and will close if it is written.

niboshi:
Changed the PR title accordingly.

okuta:
LGTM!

PR #3753, opened 2017-11-01
Check too small eps in Adam and RMSprop optimizers

niboshi:
The default `eps` of `1e-8` is too small for float16.

This PR lets optimizers raise `ValueError` for such cases, and fixes unit test by explicitly specifying `eps` for float16.

`RuntimeWarning` below will be suppressed with this fix.
```
tests/chainer_tests/optimizers_tests/test_optimizers_by_linear_model.py::TestAdam_param_0::test_linear_model_cpu
  /data/work/w/repos/chainer/chainer/optimizers/adam.py:61: RuntimeWarning: divide by zero encountered in true_divide
    param.data -= self.lr * m / (numpy.sqrt(v) + hp.eps)
  /data/work/w/repos/chainer/chainer/functions/activation/log_softmax.py:19: RuntimeWarning: invalid value encountered in subtract
    y = x - m
```

### DIFF
--- a/chainer/optimizers/adam.py
+++ b/chainer/optimizers/adam.py
@@ -55,7 +55,7 @@ class AdamRule(optimizer.UpdateRule):
             return
         hp = self.hyperparam
         eps = grad.dtype.type(hp.eps)
-        if eps == 0:
+        if hp.eps != 0 and eps == 0:
             raise ValueError(
                 'eps of Adam optimizer is too small for {} ({})'.format(
                     grad.dtype.name, hp.eps))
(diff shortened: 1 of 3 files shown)


okuta:
LGTM!

niboshi:
Ah, I think there could be a possibility that a user specifying `eps=0` explicitly.
I will fix so that it raises `ValueError` only when `hp.eps != 0` and `eps == 0` after casted.

niboshi:
Fixed.

okuta:
LGTM!

PR #3757, opened 2017-11-01
F.tile doctest does not test our function

kmaehashi:
It was just testing numpy.

https://docs.chainer.org/en/stable/reference/generated/chainer.functions.tile.html#chainer.functions.tile


niboshi:
LGTM!

PR #3761, opened 2017-11-01
Make it possible to be fine-tuning with NStepLSTM and NStepBiLSTM

massongit:
(Related to #1816)
I added `initial_model` parameter to `__init__` methods of `NStepLSTM` and `NStepBiLSTM` so that fine-tuning can. A value of this parameter is `NStepLSTM` or `NStepBiLSTM`, and this parameter is used to initialize weights and biases.


niboshi:
Thank you for PR!
In order to copy a link including its weights, you can write as `copy.deepcopy(link)`, by using Python's builtin `copy` module.
Does that achieve this purpose?

massongit:
@niboshi Thank you for your review!
I achieved your purpose.

massongit:
@niboshi Thank you for your review!
I achieved your purpose.

niboshi:
I'm sorry for being unclear. I meant you could use `copy.deepcopy` to duplicate a `NStepLSTM` link itself, including its weights, instead of `initial_model`.
How do you think?

massongit:
@niboshi I understand that you said that we have only to copy `NStepLSTM` link deeply when we do fine-tuning.
I think so, so I would like to close PR.

niboshi:
OK. Thank you for making a suggestion!

PR #3768, opened 2017-11-02
new style matmul functions and operators

kmaehashi:
This PR implements new style `matmul`, `batch_matmul` function and `@` operator.
Confirmed 10000 repeats pass.


mitmul:
LGTM

PR #3769, opened 2017-11-02
fix F.inv test does not test type error as expected

kmaehashi:
The current test code for `F.inv` does not test exception case correctly.

These tests are intended to test shape check logic, but it is catching the TypeError from numpy.

```
>>> numpy.zeros(1, 2, 2)
TypeError: data type not understood
```


niboshi:
LGTM!
To improve these tests further, how about minimizing the code inside `with` statements, by bringing `chainer.Variable(...)` outside them?
The problem would not have happened if original code had followed this practice.

kmaehashi:
Thanks for review, and I agree with it. Fixed.

niboshi:
LGTM!

PR #3770, opened 2017-11-02
fix F.inv to raise exception when input has singular matrices

### LINKED ISSUE #3765: F.inv behaves differently between CPU/GPU
Currently, F.inv ignores error flags (info) from cuBLAS.

https://github.com/chainer/chainer/blob/v3.0.0/chainer/functions/math/inv.py#L56

As a result, it behaves differently between CPU/GPU.

When CPU mode:

```py
chainer.functions.inv(np.zeros(4, dtype=np.float32).reshape((2,2)))
```

raises exception (`numpy.linalg.linalg.LinAlgError: Singular matrix`), while GPU mode:

```py
chainer.functions.inv(cp.zeros(4, dtype=np.float32).reshape((2,2)))
```

no exception occurs and

```py
variable([[ nan,  nan],
          [ nan,  inf]])
```

is returned.

I think it is better to change GPU mode behavior to raise Exception, so that users can notice the error immediately.

kmaehashi:
This fixes #3765.

~~Merge #3769 first.~~ (done)

### DIFF
--- a/chainer/functions/math/inv.py
+++ b/chainer/functions/math/inv.py
@@ -48,13 +48,19 @@ class Inv(function_node.FunctionNode):
 
     def forward_cpu(self, x):
         self.retain_outputs((0,))
-        invx = utils.force_array(numpy.linalg.inv(x[0]))
-        return invx,
+        try:
+            invx = numpy.linalg.inv(x[0])
+        except numpy.linalg.LinAlgError:
+            raise ValueError('Input has singular matrices.')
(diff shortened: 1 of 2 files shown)


okuta:
How about checking when debug_mode is True?
Because this check uses additional computational cost.

kmaehashi:
I agree with it, fixed.

okuta:
LGTM!

PR #3776, opened 2017-11-03
Improve docs of F.forget

niboshi:
* I think the word *internal* is ambiguous. *intermediate* instead would help understanding the behavior more easily.
* *this method* is not appropriate; `F.forget()` is a function.
* etc.


mitmul:
LGTM

PR #3786, opened 2017-11-05
Remove function call to up performance

okuta:


niboshi:
I marked this as `bug` because it includes two bug fixes:
* A bug that outputs the value of `config.use_cudnn` as an error if the value of `config.use_cudnn_tensor_core` is invalid.
* `get_function_check_context` did not correctly handle exception correctly.

niboshi:
LGTM!

PR #3788, opened 2017-11-06
Forget inputs as Variables

### LINKED ISSUE #3449: `F.forget` does not work correctly when inputs are bare ndarray
I found that `F.forget` does not work correctly when inputs to `F.forget` are bare `xp.ndarray` (rather than `chainer.Variable`). When all inputs to `F.forget` are bare ndarray, `backward` of `F.forget` is completely ignored and it is not called, even if trainable parameters are actually involved in the "forgotten" process (i.e., `func` parameter of `F.forget`).

As we usually want to use `F.forget` at the beginning of DNNs, it is natural to give bare ndarrays to `F.forget`. Therefore, I think this is a critical problem.

My minimal reproduction code is as follows. I used Chainer 2.1.0.

```python
import chainer
import chainer.links as L
import chainer.functions as F
import numpy as np

lnk1 = L.Scale(axis=0, W_shape=(1,))
lnk2 = L.Scale(axis=0, W_shape=(1,))

def forward(x, use_forget):
    if use_forget:
        x = F.forget(lambda x_: lnk1(x_), x)
    else:
        x = lnk1(x)

    return lnk2(x)

def print_grad(x, use_forget):
    lnk1.cleargrads()
    y = forward(x, use_forget=use_forget)
    y.backward()
    print(lnk1.W.grad)

in_variable = chainer.Variable(np.array([1], dtype=np.float32))
in_ndarray = np.array([1], dtype=np.float32)

#
# Four test cases
#
print('Chainer version:', chainer.__version__)
print_grad(in_variable, use_forget=False)  # OK
print_grad(in_ndarray, use_forget=False)   # OK
print_grad(in_variable, use_forget=True)   # OK
print_grad(in_ndarray, use_forget=True)    # NG
```

```sh
$ python forget_test.py
Chainer version: 2.1.0
[ 1.]
[ 1.]
[ 1.]
None
```

hvy:
Fixes #3449 .

Basically, it wraps all ndarray inputs to `F.forget` as Variables to ensure that backward is called.

### DIFF
--- a/chainer/functions/util/forget.py
+++ b/chainer/functions/util/forget.py
@@ -118,4 +118,8 @@ def forget(func, *xs):
         the method returns a tuple too.
 
     """
+    # Guarantee that backward is called by wrapping bare ndarrays as
+    # ``Variable`` s with ``requires_grad`` set to ``True``.
+    xs = tuple(x if isinstance(x, variable.Variable) else
+               variable.Variable(x, requires_grad=True) for x in xs)
     return Forget(func)(*xs)
--- a/chainer/functions/util/forget.py
+++ b/chainer/functions/util/forget.py
@@ -101,10 +101,18 @@ def forget(func, *xs):
 
     .. note::
 
-      ``F.forget`` does not support functions which behave differently in
-      multiple calls with the same inputs, such as
-      :meth:`F.dropout() <chainer.functions.dropout>` and
-      :meth:`F.negative_sampling() <chainer.functions.negative_sampling>`.
+        ``F.forget`` does not support functions which behave differently in
+        multiple calls with the same inputs, such as
+        :meth:`F.dropout() <chainer.functions.dropout>` and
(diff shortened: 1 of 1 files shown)


chainer/functions/util/forget.py
```diff
@@ -117,4 +117,8 @@ def forget(func, *xs):
         the method returns a tuple too.
 
     """
+    # Guarantee that backward is called by wrapping bare ndarrays as
+    # ``Variable`` s with ``requires_grad`` set to ``True``.
```
mitmul: How about adding this explanation to the docstring above? I think just adding a note to say it'll be wrapped by Variable if it's ndarray is fine.

hvy:
@mitmul Thank you for the review. I fixed the docs (removed the comments in the code since its duplicate information) and rebased to master.

mitmul:
LGTM

PR #3688, opened 2017-10-26
add tutorial for Reporter and report.

keisuke-umezawa:
related to #3290.

The contents will be:
- What is Reporter?
- Use report in Chain or Link
- Naming rule for the reported values

PR #3795, opened 2017-11-06
Fix the example in the document of Reporter

toslunar:
This PR fixes the error below and removes `:` in `reporter.add_observer('my_observer:', observer)`.

I executed the example in the readthedocs, and found that `/` is inserted between the names of the observation and the key of the reported dict.

```
>>> from chainer import Reporter, report, report_scope
>>>
>>> reporter = Reporter()
>>> observer = object()  # it can be an arbitrary (reference) object
>>> reporter.add_observer('my_observer:', observer)
>>> observation = {}
>>> with reporter.scope(observation):
...     reporter.report({'x': 1}, observer)
...
>>> observation
{'my_observer:/x': 1}
```
with Chainer v3, in Python 3.

### DIFF
--- a/chainer/reporter.py
+++ b/chainer/reporter.py
@@ -35,13 +35,13 @@ class Reporter(object):
        >>>
        >>> reporter = Reporter()
        >>> observer = object()  # it can be an arbitrary (reference) object
-       >>> reporter.add_observer('my_observer:', observer)
+       >>> reporter.add_observer('my_observer', observer)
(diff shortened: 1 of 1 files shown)


mitmul:
Thank you for the PR, LGTM.

mitmul:
Jenkins, test this please.

toslunar:
Related: #3688

mitmul:
LGTM

PR #3803, opened 2017-11-06
[backport] Add Trainer tutorial

niboshi:
Backport of #2620

PR #3810, opened 2017-11-07
fix doctest failure in trainer tutorial

kmaehashi:
This PR fixes doctest failure of trainer tutorial.

`testsetup` is repalced with `testcode` with `:hide:`, as `testsetup` runs before other `testcode` blocks.


niboshi:
LGTM!

niboshi:
Backport after #3803.

PR #3816, opened 2017-11-08
Fix casing of header in README

hvy:
Fixed the casing of the header in the README. 

Unless it has been discussed before, I think this looks better. For reference, see http://blog.apastyle.org/apastyle/2011/06/capitalization-after-colons.html. 

Additionally, if we want to emphasize the fact that this indeed is a _deep learning framework_, I suggest we instead change the header to `Chainer: A Deep Learning Framework`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 <div align="center"><img src="docs/image/chainer_red_h.png" width="400"/></div>
 
-# Chainer: a deep learning framework
+# Chainer: A deep learning framework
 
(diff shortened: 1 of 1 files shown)


kmaehashi:
I think it is better to have one unified branding.

In the Chainer official website, the title says:

> A flexible framework for neural networks

and the page header says:

> A Powerful, Flexible, and Intuitive Framework for Neural Networks

PR #3818, opened 2017-11-08
Remove unnecessary branch in minimum forward.

takagi:
This PR removes an unnecessary branch in `chainer.functions.minimum` forward computation, which might cause slight cost but  it would be enough to ignore.


kmaehashi:
Thanks for the followup! LGTM, I will merge after test pass.

kmaehashi:
LGTM!

PR #3410, opened 2017-09-17
Stabilize upsampling_2d test

okuta:
Current upsampling_2d_test is unstable.

PR #3826, opened 2017-11-08
[backport] Stabilize upsampling_2d test

takagi:
Backport of #3410. It seems to have been left not backported, causing CI of the stable branch unstable.

### DIFF
--- a/tests/chainer_tests/functions_tests/pooling_tests/test_upsampling_2d.py
+++ b/tests/chainer_tests/functions_tests/pooling_tests/test_upsampling_2d.py
@@ -25,12 +25,10 @@ class TestUpsampling2D(unittest.TestCase):
             -1, 1, self.in_shape).astype(self.dtype)
         self.ggx = numpy.random.uniform(
             -1, 1, self.pooled_y.shape).astype(self.dtype)
+        self.check_backward_options = {}
+        self.check_double_backward_options = {'atol': 1e-3, 'rtol': 1e-2}
         if self.dtype == numpy.float16:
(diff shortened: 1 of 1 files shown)


hvy:
jenkins, test this please.

tests/chainer_tests/functions_tests/pooling_tests/test_upsampling_2d.py
```diff
@@ -25,12 +25,10 @@ def setUp(self):
             -1, 1, self.in_shape).astype(self.dtype)
         self.ggx = numpy.random.uniform(
             -1, 1, self.pooled_y.shape).astype(self.dtype)
+        self.check_backward_options = {}
```
hvy: Should we declare `self.check_backward_options` although it seems to remain an empty dict throughout the unit test?

hvy:
@takagi san, although this is a backport, I added a comment. Please ignore it if I'ver misunderstood. FYI, it passed all Jenkins tests.

tests/chainer_tests/functions_tests/pooling_tests/test_upsampling_2d.py
takagi: It can be removed, as you say, but it can be also left because it is not harmful and we could just put tolerance when we want to adjust it in the future. Anyway, it doesn't have very high priority, I think.

hvy: I agree that it's a minor concern.

hvy:
LGTM!

PR #3781, opened 2017-11-03
Use correct module path

okuta:
rel #3364

PR #3829, opened 2017-11-09
[backport] Use correct module path

niboshi:
Backport of #3781


niboshi:
Note that changes to `tests/chainer_tests/functions_tests/connection_tests/test_convolution_2d.py` and `tests/chainer_tests/functions_tests/connection_tests/test_dilated_convolution_2d.py` are omitted, because the code they depend on is not backported. (#2858)

okuta:
LGTM!

PR #3832, opened 2017-11-09
Simplify classification custom loop examples

yuyu2172:
1. I deleted unused arguments
2. I deleted lines that turn input arrays to variables. In the training loop examples, the inputs to networks are arrays. The current style is inconsistent.


examples/mnist/train_mnist_custom_loop.py
```diff
@@ -28,8 +28,6 @@ def main():
                         help='GPU ID (negative value indicates CPU)')
     parser.add_argument('--out', '-o', default='result',
```
toslunar: The argument `--out` doesn't seem to be used.

Crissman:
Merge conflict fixed. LGTM.

okuta:
@yuyu2172 Sorry, late reply. Cloud you resolve conflict?

yuyu2172:
I think this PR is outdated, since the examples now use `out` and `resume`.
Thus, I would like to close this.

okuta:
OK.

PR #3833, opened 2017-11-09
Verbose error in gradient check

hvy:
Made two error messages related to gradient check errors more verbose. 

The messages now include actual/expected values. Simply because when actual values don't match expected values in assertions, we often want to know in what sense they differ (w/o their internal variable name).


niboshi:
LGTM!

PR #3839, opened 2017-11-10
Verbose error message in basic_math

hvy:
More verbose error message that prints out the actual value.

### DIFF
--- a/chainer/functions/math/basic_math.py
+++ b/chainer/functions/math/basic_math.py
@@ -23,7 +23,8 @@ def _convert_value_to_string(value):
         return 'constant array'
     else:
         raise ValueError(
-            'value must be scalar, ndarray, or Variable')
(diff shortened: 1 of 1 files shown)


chainer/functions/math/basic_math.py
```diff
@@ -33,7 +33,8 @@ def _check_constant_type(value):
         return
     else:
         raise ValueError(
-            'value must be scalar, ndarray, or Variable')
+            'Value must be a NumPy scalar, `numpy.ndarray`, `cupy.ndarray`'
```
kmaehashi: I think `scalar` here includes python scalar, not only NumPy scalar.

```
>>> chainer.Variable(np.array(1)) + 1
variable(2)
```

hvy:
@kmaehashi san, thank you for the review. Of course, it is now fixed. I also fixed the spacing and another error message in the same module.

kmaehashi:
LGTM!

PR #3846, opened 2017-11-12
Add the error function (erf)

muupan:
This PR adds the error function (erf).

CPU forward computation requires scipy. Please check if it is the correct way to use scipy in Chainer.


muupan:
I'll try supporting environments without scipy by using math.erf instead.

chainer/functions/__init__.py
```diff
@@ -180,6 +180,8 @@
 from chainer.functions.math.det import batch_det  # NOQA
 from chainer.functions.math.det import BatchDet  # NOQA
 from chainer.functions.math.det import det  # NOQA
+from chainer.functions.math.erf import erf  # NOQA
+from chainer.functions.math.erf import Erf  # NOQA
```
beam2d: We have decided not to expose newly added `Function`/`FunctionNode` classes under `chainer.functions`, so remove this line.

tests/chainer_tests/functions_tests/math_tests/test_erf.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def check_label(self, op, expected):
+        self.assertEqual(op().label, expected)
+
+
+@testing.parameterize(*testing.product({
+    'shape': [(3, 2), ()],
+    'dtype': [numpy.float16, numpy.float32, numpy.float64],
+}))
+class TestErf(UnaryFunctionsTestBase):
```
beam2d: Could you consider using `unary_math_function_unittest` utility to generate test cases? See the usage here: https://docs.chainer.org/en/stable/reference/generated/chainer.testing.unary_math_function_unittest.html#chainer.testing.unary_math_function_unittest

chainer/functions/math/erf.py
```diff
@@ -0,0 +1,66 @@
...
+
+def erf(x):
+    """Elementwise error function.
+
+    .. note::
+       Forward computation in CPU can be slow if
+       `SciPy <https://www.scipy.org/>`_ is not available.
+
+    Args:
+        x (~chainer.Variable): Input variable.
```
beam2d: Add `numpy.ndarray` and `cupy.ndarray` for allowed argument types.

muupan:
Sorry for my late response. I fixed all the issues you mentioned.

beam2d:
Thank you!

PR #3862, opened 2017-11-14
 Refactor unit tests for various backend configuration

niboshi:
This PR depends on ~#3850~.

Currently in unit tests of functions, various data parameters (`x_dtype`, `stride`, `nobias`, etc.) and backend parameters (`use_cudnn`, `autotune`, etc.) are mixed. There are some problems regarding this scheme:
* It's exponentially difficult to add new backends.
* Some tests are duplicated. For example, CPU tests are performed for every combination of `use_cudnn`, `cudnn_deterministic`, `autotune`, etc.

This PR solves this problem by separating data and backend configuration.
To do this,
* `testing.backend.inject_backend_tests` has been introduced. A test writer can apply this decorator to a test case, so that specified test fixtures (e.g. `test_backward`) will be duplicated to as many as backend configuration combinations (specified by the test writer). Each of duplicated fixtures will accept a new argument of type `testing.backend.BackendConfig`. `BackendConfig` has the information of one instance of backend configuration combination. `with` construct can be used against this object to to apply backend configuration (`chainer.use_config` is called accordingly).
* `testing.product` has been extended to support hierarchical structure.
* Applied to `chainer_tests/functions_tests/connection_tests/test_deconvolution_2d.py`.

### DIFF
--- a/chainer/testing/backend.py
+++ b/chainer/testing/backend.py
@@ -18,7 +18,8 @@ class BackendConfig(object):
     ]
 
     def __init__(self, params):
-        assert isinstance(params, dict)
+        if not isinstance(params, dict):
+            raise TypeError('params must be a dict.')
         self._contexts = []
 
         # Default values
@@ -105,9 +106,12 @@ def _wrap_backend_test_method(impl, param, method_name):
 
 
 def inject_backend_tests(method_names, params):
-    assert isinstance(method_names, list)
-    assert isinstance(params, list)
-    assert all(isinstance(_, dict) for _ in params)
+    if not isinstance(method_names, list):
+        raise TypeError('method_names must be a list.')
+    if not isinstance(params, list):
+        raise TypeError('params must be a list of dicts.')
+    if not all(isinstance(d, dict) for d in params):
+        raise TypeError('params must be a list of dicts.')
 
     def wrap(case):
         assert issubclass(case, unittest.TestCase)
--- a/chainer/testing/backend.py
+++ b/chainer/testing/backend.py
@@ -1,0 +1,126 @@
+import functools
+import unittest
+
+import numpy
+
+import chainer
+from chainer import cuda
+from chainer.testing import attr
+
+
+class BackendConfig(object):
+
+    _props = [
+        ('use_cuda', False),
+        ('use_cudnn', 'never'),
+        ('cudnn_deterministic', False),
+        ('autotune', False),
+    ]
+
+    def __init__(self, params):
+        assert isinstance(params, dict)
+        self._contexts = []
+
+        # Default values
+        for k, v in self._props:
+            setattr(self, k, v)
+        # Specified values
+        for k, v in params.items():
+            if not hasattr(self, k):
+                raise ValueError('Parameter {} is not defined'.format(k))
+            setattr(self, k, v)
+
+    @property
+    def xp(self):
+        if self.use_cuda:
+            return cuda.cupy
+        else:
+            return numpy
+
+    def __enter__(self):
+        self._contexts = [
+            chainer.using_config(
+                'use_cudnn', self.use_cudnn),
+            chainer.using_config(
+                'cudnn_deterministic', self.cudnn_deterministic),
+            chainer.using_config(
+                'autotune', self.autotune),
+        ]
+        for c in self._contexts:
+            c.__enter__()
+        return self
+
+    def __exit__(self, typ, value, traceback):
+        for c in reversed(self._contexts):
+            c.__exit__(typ, value, traceback)
+
+    def __repr__(self):
+        lst = []
+        for k, _ in self._props:
+            lst.append('{}={!r}'.format(k, getattr(self, k)))
+        return '<BackendConfig {}>'.format(' '.join(lst))
+
+    def get_func_str(self):
+        """Returns a string that can be used in method name"""
+        lst = []
+        for k, _ in self._props:
+            val = getattr(self, k)
+            if val is True:
+                val = 'true'
+            elif val is False:
+                val = 'false'
+            else:
+                val = str(val)
+            lst.append('{}_{}'.format(k, val))
+        return '__'.join(lst)
+
+    def get_pytest_marks(self):
+        marks = []
+        if self.use_cuda:
+            marks.append(attr.gpu)
+            if self.use_cudnn != 'never':
+                marks.append(attr.cudnn)
+
+        assert all(callable(_) for _ in marks)
+        return marks
+
+
+def _wrap_backend_test_method(impl, param, method_name):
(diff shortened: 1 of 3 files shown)


niboshi:
Rebased.

chainer/testing/backend.py
```diff
@@ -0,0 +1,126 @@
...
+
+    _props = [
+        ('use_cuda', False),
+        ('use_cudnn', 'never'),
+        ('cudnn_deterministic', False),
+        ('autotune', False),
+    ]
+
+    def __init__(self, params):
+        assert isinstance(params, dict)
```
hvy: How about raising an exception instead since the constructor may be called by a user?

chainer/testing/backend.py
```diff
@@ -0,0 +1,126 @@
...
+        return '__'.join(lst)
+
+    def get_pytest_marks(self):
+        marks = []
+        if self.use_cuda:
+            marks.append(attr.gpu)
+            if self.use_cudnn != 'never':
+                marks.append(attr.cudnn)
+
+        assert all(callable(_) for _ in marks)
```
hvy: Do we need this assertion?

chainer/testing/backend.py
```diff
@@ -0,0 +1,126 @@
...
+    for mark in marks:
+        func = mark(func)
+
+    return func, new_method_name
+
+
+def inject_backend_tests(method_names, params):
+    assert isinstance(method_names, list)
+    assert isinstance(params, list)
+    assert all(isinstance(_, dict) for _ in params)
```
hvy: Ditto. How about exceptions instead?

tests/chainer_tests/functions_tests/connection_tests/test_deconvolution_2d.py
```diff
@@ -87,54 +110,67 @@ def setUp(self):
...
+        assert y_actual.data.dtype == self.x_dtype
         testing.assert_allclose(
-            y_cpu.data, y_gpu.data.get(), **self.test_forward_options)
+            y_expected.data, y_actual.data.get(), **self.test_forward_options)
 
     @attr.gpu
-    def test_forward_consistency_im2col(self):
-        self.use_cudnn = 'never'
-        self.test_forward_consistency()
+    def test_forward(self, backend_config):
```
hvy: ~I am not sure about this but is this decorator preserved when the fixture is replaced by the backend injection?~ In any case, I think it can be removed since the `use_cuda` param in the injection ensures it.

hvy:
Thank you for this PR. I added some comments.

We might want to merge #3974 prior to this one, and update the `cuda` import statements to go via the `backends` subpackage.

chainer/testing/parameterized.py
```diff
@@ -85,10 +85,32 @@ def f(klass):
...
+    elif isinstance(parameter, list):
+        # list of lists of dicts
+        if not all(isinstance(_, list) for _ in parameter):
+            raise TypeError('parameter must be list of lists of dicts')
+        if not all(isinstance(_, dict) for l in parameter for _ in l):
+            raise TypeError('parameter must be list of lists of dicts')
+
+        product = list(itertools.product(*parameter))
+        lst = []
+        for dict_lst in product:
```
hvy: How about `for dict_lst in = list(itertools.product(*parameter)):` instead, not instantiating the `product` variable, partially because it shadows the function with the same name?

tests/chainer_tests/functions_tests/connection_tests/test_deconvolution_2d.py
```diff
@@ -87,54 +110,67 @@ def setUp(self):
...
+    def check_backward(self, inputs, grad_outputs, backend_config):
 
-    def check_backward(self, x_data, W_data, b_data, y_grad):
-        xp = cuda.get_array_module(x_data)
+        xp = backend_config.xp
+        if backend_config.use_cuda:
+            inputs = cuda.to_gpu(inputs)
+            grad_outputs = cuda.to_gpu(grad_outputs)
+
+        x_data, W_data, b_data = inputs
```
hvy: This is a small detail but can we keep the variables names consistent with `check_forward()`, i.e. `x, W, b = inputs`? Oh, and I really like the idea of packing the inputs and gradients. A lot less clutter.

tests/chainer_tests/functions_tests/connection_tests/test_deconvolution_2d.py
```diff
@@ -144,49 +180,48 @@ def f(*args):
...
+
+        if backend_config.use_cuda:
+            inputs = cuda.to_gpu(inputs)
+            grad_outputs = cuda.to_gpu(grad_outputs)
+            grad_grad_inputs = cuda.to_gpu(grad_grad_inputs)
 
-    def check_double_backward(self, x_data, W_data, b_data, y_grad,
-                              x_grad_grad, W_grad_grad, b_grad_grad):
-        xp = cuda.get_array_module(x_data)
+        x_data, W_data, b_data = inputs
```
hvy: Ditto.

hvy:
Added a few minor comments.

chainer/testing/backend.py
niboshi: Fixed. this class is not meant to be instantiated by a user, though.

chainer/testing/backend.py
niboshi: Yes, to detect a glitch in the preceding logic earlier.

chainer/testing/backend.py
niboshi: Fixed.

chainer/testing/parameterized.py
niboshi: Fixed.

tests/chainer_tests/functions_tests/connection_tests/test_deconvolution_2d.py
niboshi: Do I need to fix it here? Those code will be refactored in #3499 anyway.

chainer/testing/backend.py
hvy: Thanks! Yes, I assumed that but it wasn't completely clear from the module path/class declaration.

chainer/testing/backend.py
hvy: Okay, then an assertion is appropriate.

tests/chainer_tests/functions_tests/connection_tests/test_deconvolution_2d.py
hvy: I see, thanks for the response.

hvy:
jenkins, test this please.

hvy:
jenkins, test this please.

hvy:
LGTM!

PR #3848, opened 2017-11-13
Refactor (de)convolution_2d

niboshi:
This is just a refactoring, in preparation for integrating iDeep.

* Moved `ndarray` type check to `function_node._check_data_type_forward`.
* Refactored duplicate code as a method (`_calc_out_size`)
* Isolated cuDNN forward calculation as methods.

PR #3865, opened 2017-11-14
Support iDeep

niboshi:
This PR depends on ~#3875~ (thus ~#3848~, ~#3862~, ~#3850~) and ~#3869~.
Also I want ~#3974~ to be merged beforehand.

This PR supports Intel's iDeep (https://github.com/intel/ideep).

* [x] Basic support - #4276
* [x] `F.deconvolution_2d` - #4284
* [x] `F.convolution_2d` - #4284
* [x] `F.relu` -  #4277
* [x] `F.linear` - #4278
* [x] `F.lstm` - #4283
* [x] `F.dropout` - #4288
* [x] `F.split_axis` - #4281
* [x] `F.concat` - #4289
* [x] `F.local_response_normalization` - #4279
* [x] `F.batch_normalization` - #4280
* [x] `F.average_pooling_2d` - #4282
* [x] `F.max_pooling_2d` - #4286
* [x] `optimizers.SGD` - #4285
* [x] `optimizers.MomentumSGD` - #4285


niboshi:
Please merge #3974 before this PR.
I'll move `chainer._ideep` to `chainer.backends.ideep` after it's merged.

chainer/__init__.py
```diff
@@ -92,6 +136,7 @@ def get_function_hooks():
 global_config.use_cudnn = os.environ.get('CHAINER_USE_CUDNN', 'auto')
 global_config.use_cudnn_tensor_core = 'auto'
 global_config.autotune = False
+global_config.use_ideep = os.environ.get('CHAINER_USE_IDEEP', 'auto')
```
kmaehashi: Just for remembrance: need to add this to docs.

chainer/cuda.py
```diff
@@ -255,6 +256,27 @@ def to_gpu(array, device=None, stream=None):
...
+        else:
+            return _array_to_gpu(array, device_, stream)
+
+
+def _array_to_gpu(array, device, stream):
+    if array is None:
+        return None
     if isinstance(array, (numpy.number, numpy.bool_)):
         array = numpy.asarray(array)
     if not isinstance(array, (cupy.ndarray, numpy.ndarray)):
```
kmaehashi: Shouldn't we able to send mdarray to GPU?

chainer/functions/activation/relu.py
```diff
@@ -17,40 +18,70 @@ class ReLU(function_node.FunctionNode):
...
+            return ReLUGradCudnn(x, y).apply((gy,))
+        # Generic implementaion
+        return ReLUGrad2(y).apply((gy,))
 
 
 def _heaviside(x):
-    return (x > 0).astype(x.dtype)
+    if _ideep.is_available() and isinstance(x, _ideep.ideep.mdarray):
+        # ideep.mdarray does not support __gt__ yet
+        return numpy.greater(x, 0).astype(x.dtype)
```
kmaehashi: iDeep document says `'MD-Array' supports python new buffer protocol and operates compatibily with NumPY ND-Array.`, so maybe we can ask iDeep team to support operator overloads?

Currently mdarray is using `tp_getattro` to fallback attribute access to numpy, however, special method lookup (e.g., `__add__` for `+`) does not go through this route. https://docs.python.org/3/c-api/typeobj.html#c.PyTypeObject.tp_getattro
Maybe proxy method or in python layer is needed (by them, or by us?)

chainer/_ideep.py
```diff
@@ -0,0 +1,64 @@
...
+    This function checks ``chainer.config.use_ideep`` and availability
+    of ``ideep`` package.
+
+    Args:
+        level (str): iDeep use level. It must be either ``'==always'`` or
+            ``'>=auto'``. ``'==always'`` indicates that the ``use_ideep``
+            config must be ``'always'`` to use iDeep.
+
+    Returns:
+        bool: ``True`` if the caller should use cuDNN.
```
kmaehashi: s/cuDNN/iDeep/

kmaehashi:
I tried installing iDeep, confirmed ReLU working, and roughly looked through the code.
I'll revisit after merging all dependent PRs.

chainer/__init__.py
niboshi: Yes, maybe in another PR.

chainer/cuda.py
niboshi: Yes, we should be able to handle `mdarray`. Let's revisit after new ideep will be released.

chainer/_ideep.py
niboshi: Thanks, fixed.

arthuryuan1987:
Thanks for your reply (#4163).

Glad to know you have such progress. We all think it is better to co-work with you to integrate iDeep into Chainer. You can access  https://github.com/intel/chainer to reference latest Intel-Chainer.

Firstly, we hope to inform you the information of the latest Intel-Chainer and iDeep. Abstracts in the latest release,
1. Improve iDeep implementation to boost performance.
2. Provide simple iDeep interfaces and eliminate coupling with Chainer framework.
3. Finish integration of iDeep, including integration of all supported functions and some modifications in main frame to introduce iDeep optimization path (`forward_ia`) and `ideep.mdarray`.

And the status,
**Unit test**: 100% pass rate.
**Performance**: Compared with older iDeep, get better performance in inference.
**Convergence**: Get STOA on GoogLeNetV1 and VGG16. Get same convergence result as GPU on ResNet50.

We have reviewed the PR, and have some ideas of the work of supporting iDeep in Chainer. Please let me put them here and take a discussion,
1. We have different understanding in 'ia'/'ideep'. From the PR, 'ideep' is an optimized implementation under 'cpu' (Am I right?). However, in our mind, 'ia' is 'Intel Architecture', defined as a kind of backend, just like `chainer.cuda`. We use iDeep to optimize backend 'ia'. We consider that `forward_ia` (`forward_ideep` in the PR) has same abstract level as `forward_cpu` and `forward_gpu` have. We branch `forward_gpu`, `forward_ia` and `forward_cpu` uniformly in :`func`:`FunctionNode.forward`. You can reference #4163 to have an impression.
2. To implement all operations/functions supported by iDeep. In latest iDeep, We simplify iDeep interface to make stateless and to eliminate coupling with Chainer framework meanwhile. So there will be no `ideep_hint` in `FunctionNode` instance. You can reference Intel-Chainer. We will help to implement operations/functions.
3. From experiences in unit test, convergence test and performance tuning in Intel-Chainer, we need put some additional patches to support `chainer.ia` and `mdarray`. We will provide anything what we knew.

4. We are willing to work on your PR. We plan to provide a version of our idea based on the PR.

niboshi:
@arthuryuan1987 

Thank you for reviewing and raising discussion!

1. In our intention, modules in `chainer.backends` do not necessarily represent hardware backends. It's meant for anything that can be viewed as "backend" from Chainer's point of view. We plan to include higher-level libraries there. As an example, there's an ongoing discussion to introduce `chainer.backends.numexpr` (https://github.com/chainer/chainer/pull/3937#issuecomment-349545258). NumExpr is an even higher-level library than NumPy.

As for `forward_ia`, we prefer to implement a branch in `forward_cpu` as in my PR. There are some reasons:
* There's an assumption in early branching to `forward_ia` that there is (and will be) no difference between functions on how we determine whether we can utilize ideep. Currently it's true, but in the future there may be a function that can utilize ideep only for a particular input shape, for example.
* `all_ready` also assumes the order of input arguments, as it checks the first input for its `ndim`. However, Chainer does not enforce particular order of inputs for functions.
* We are careful about introducing a new (irreversible) interface. Once `forward_ia` is implemented, it would be compatibility-breaking to move back to `forward_cpu`. Conversely, we can move to `forward_ia` any time when we conclude that that's the optimial way.

2. Great! Do you have any plan to update intel/ideep?

3. / 4. Thank you! Please share patches in any form. Post a diff here / Make a PR to my fork branch / post on slack.

arthuryuan1987:
Thanks for your explanation @niboshi.

Agree with your idea of `chainer.backends` and `forward_ia`.

So,
1. We will focus on operations/functions implementation with the latest iDeep interface.
2. We will make a PR to your fork branch to implement it as soon as possible.

Thank you.

arthuryuan1987:
Hi, @niboshi 

Our patch (iDeep integration in ReLU/Conv/Deconv/Linear) for your PR is WIP. It will be delivered by the middle of next week after evaluation. The process includes our internal code review, UT, ST and so on.

Have you got any schedule of the PR (or the feature "Support iDeep") ?

arthuryuan1987:
Hi, niboshi

I have post the PR for you fork branch. (https://github.com/niboshi/chainer/pull/2)

Please take a review.

Thanks.

niboshi:
@arthuryuan1987 
Thank you for PR.

As for the release schedule, we had been thinking of including it in v4.0.0 (expected around March).
In order for that, it should be included in v4.0.0b3 (expected in next week), but it seems that's difficult. We're considering updateing the schedule.

arthuryuan1987:
Thanks for you information.

We will try our best to help to support the feature.

niboshi:
@arthuryuan1987 
We discussed about the module name (under `chainer.backends`). `chainer.backends.ideep` might become inappropriate when in the future we want something which is related IA but is not using iDeep.
We considered `chainer.backends.ia` as you suggested, but that might be confusing, as IA-64 and x86-64 are different things.
We are thinking of `chainer.backends.intel64`.
Do you have any opinion?

arthuryuan1987:
@niboshi 

Agree with you on the naming. We think `chainer.backends.intel64` is better. The naming is more accurate.

niboshi:
@arthuryuan1987 
OK. Thank you for the comment!

chainer/functions/array/transpose_sequence.py
```diff
@@ -26,7 +26,8 @@ def _transpose(xs, length):
 
         outs = tuple([xp.empty((l,) + unit, dtype=dtype) for l in lengths])
         for i, x in enumerate(xs):
-            for p, xi in enumerate(x):
+            # TODO(iDeep): Support iterable
+            for p, xi in enumerate(xp.array(x)):
```
arthuryuan1987: `iDeep` tensor has supported iterable. The workaround could be reverted. Thanks.

arthuryuan1987:
Hi, @niboshi ,

`iDeep` tensor (`mdarray`) could support iterable. The feature is included in latest `iDeep` release.

https://github.com/intel/ideep, branch: `ideep4py` or `hotfix_v1_0_0a`.

niboshi:
Thank you. I reverted the workarounds.

kmaehashi:
All PRs originated from this PR has been merged!

PR #3867, opened 2017-11-14
New style huber_loss

hvy:
New style huber_loss to support double backprop.

### DIFF
--- a/chainer/functions/loss/huber_loss.py
+++ b/chainer/functions/loss/huber_loss.py
@@ -23,8 +23,8 @@ class HuberLoss(function_node.FunctionNode):
         type_check.expect(
             in_types[0].dtype == numpy.float32,
             in_types[1].dtype == numpy.float32,
-            in_types[0].ndim == 2,
-            in_types[1].ndim == 2,
+            in_types[0].ndim >= 2,
+            in_types[1].ndim >= 2,
             in_types[0].shape == in_types[1].shape
         )
 
--- a/chainer/functions/loss/huber_loss.py
+++ b/chainer/functions/loss/huber_loss.py
@@ -1,19 +1,21 @@
 import numpy
 
+import chainer
 from chainer.backends import cuda
-from chainer import function
+from chainer import function_node
 from chainer.utils import type_check
 
 
-class HuberLoss(function.Function):
+class HuberLoss(function_node.FunctionNode):
 
     def __init__(self, delta, reduce='sum_along_second_axis'):
         self.delta = delta
 
         if reduce not in ('sum_along_second_axis', 'no'):
             raise ValueError(
-                "only 'sum_along_second_axis' and 'no' are valid "
+                "Only 'sum_along_second_axis' and 'no' are valid "
                 "for 'reduce', but '%s' is given" % reduce)
+
         self.reduce = reduce
 
     def check_type_forward(self, in_types):
@@ -25,6 +27,7 @@ class HuberLoss(function.Function):
         )
 
     def forward(self, inputs):
+        self.retain_inputs((0, 1))
         xp = cuda.get_array_module(*inputs)
         x0, x1 = inputs
         self.diff = x0 - x1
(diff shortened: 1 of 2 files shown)


niboshi:
Could you separate documentation fix as another PR?
It's ok to include small and trivial fixes, but I think this fix is a bit large.

hvy:
Yes, sure. I'll create a separate PR for the last commit then.

hvy:
Separated into #3950.

niboshi:
Double-backward test fails.
Could you check it?

```
TestHuberLoss_param_0.test_double_backward_cpu

E   AssertionError:                                                                                                                                                                                                
E   Fail: 1, Success: 4266                                                                                                                                                                                         
E                                                                                                                                                                                                                  

E   gradients (numeric):  0.030036967031677603      
E   gradients (backward): 0.10307390057595694       
E                                                   
E                                                   
E   Not equal to tolerance rtol=0.05, atol=0.05     
E                                                   
E   (mismatch 100.0%)                               
E    x: array(0.030036967031677603)                 
E    y: array(0.10307390057595694)                  
E                                                   
E   assert_allclose failed:                         
E     shape: () ()                                  
E     dtype: float64 float64                        
E     i: (0,)                                       
E     x[i]: 0.030036967031677603                    
E     y[i]: 0.10307390057595694                     
E     err[i]: 0.07303693354427934                   
E   x: 0.030036967031677603                         
E   y: 0.10307390057595694      
```

hvy:
The large error is caused by non-differentiable points during the double backward tests (because of `sign`?).  Maybe we can merge this PR after the test template that checks for these points.

hvy:
Rebased!

toslunar:
It's not documented but variables with `ndim>2` were permitted. I'd like to fix *docstring* to allow this, rather than restricting the usage.
```
>>> x
array([[[ 1.,  1.,  1.,  1.],
        [ 1.,  1.,  1.,  1.],
        [ 1.,  1.,  1.,  1.]],

       [[ 1.,  1.,  1.,  1.],
        [ 1.,  1.,  1.,  1.],
        [ 1.,  1.,  1.,  1.]]], dtype=float32)
>>> t
array([[[  0.,   1.,   2.,   3.],
        [  4.,   5.,   6.,   7.],
        [  8.,   9.,  10.,  11.]],

       [[ 12.,  13.,  14.,  15.],
        [ 16.,  17.,  18.,  19.],
        [ 20.,  21.,  22.,  23.]]], dtype=float32)
>>> F.huber_loss(x, t, 10, reduce='no')
variable([[[   0.5,    0. ,    0.5,    2. ],
           [   4.5,    8. ,   12.5,   18. ],
           [  24.5,   32. ,   40.5,   50. ]],

          [[  60. ,   70. ,   80. ,   90. ],
           [ 100. ,  110. ,  120. ,  130. ],
           [ 140. ,  150. ,  160. ,  170. ]]])
```

hvy:
@toslunar Thanks for the feedback. Do you mean that we should allow `ndim>2` only when the `reduce` option is `no`? I suppose we could add another reduce options to reduce all axes except the batch one.

toslunar:
I should have written only "Could you keep the feature, which is not necessarily a bug, in this PR?" I'm not sure if `chainer.functions.huber_loss` *should* support arrays of `ndim>2`, while I'm +1 for supporting them.

See also: #2265

hvy:
@toslunar I see what you mean. I changed the input type check to allow greater dimensions. Does this satisfy your need?

I also agree that arbitrary dimensions should be supported (which is not difficult to implement) but that it's already partially addressed by #2265 .

toslunar:
Let me see, I found `reduce='no'` with `ndim==1` is allowed:
```
>>> F.huber_loss(np.array([0], dtype='f'), np.array([1], dtype='f'), delta=1, reduce='no')
variable([ 0.5])
```

It doesn't work with `ndim==0`, which seems to be fixed by `chainer.utils.force_array`.

hvy:
@toslunar Sorry, I'm not sure I understood you correctly, but the input type check is now the same as before. To further improve the axis support, let's discuss it in #2265 !

chainer/functions/loss/huber_loss.py
```diff
@@ -37,15 +40,20 @@ def forward(self, inputs):
...
+        diff = x0 - x1
+        gy, = grad_outputs
+
+        xp = cuda.get_array_module(self.diff)
         mask = xp.abs(self.diff) <= self.delta
 
-        gx = xp.where(mask, self.diff, self.delta * xp.sign(self.diff))
-        gy_ = gy[0]
+        gx = chainer.functions.where(
+            mask, diff, self.delta * xp.sign(self.diff))
```
toslunar: Do we need `self.diff` for backward compatibility?  If not, `self.diff` can be replaced by `diff.array`.

hvy: It is there simply to avoid having to recompute it in the backward pass (although it simply a subtraction and could be recomputed quite easily). Do you have a good reason why we'd want to change this?

toslunar: Sorry.  I wasn't very familiar with new style functions.
I think `diff = x0 - x1` also computes the value, and then `diff.array` will cost little.  Is it true?

By the way, is there any way to save `x0 - x1` in `Variable` at `forward`?  Then it would not have to retain inputs `x0`, `x1`.  In other words, once `self.diff` becomes `Variable`, this line can be
```
        gx = chainer.functions.where(
            mask, self.diff, self.delta * xp.sign(self.diff))
```

hvy: In terms of performance, I think the difference is pretty small but you are right, I'll fix it!

As for the latter, I also think that there are other use-cases where we'd like to retain intermediate Variables but that is unfortunately not possible as far as I know.

hvy: Removed `self.diff`!

niboshi:
jenkins, test this please

niboshi:
jenkins, test this please

niboshi:
LGTM!
Sorry for being late to merge.

hvy:
Thanks!!

PR #3871, opened 2017-11-14
Avoid zero division error in linear init call

### LINKED ISSUE #3863: L.Linear without in_size raises ZeroDivisionError on first input if the batchsize is 0
This is probably fixed by changing `x.size // x.shape[0]` to `numpy.prod(x.shape[1:])` though it requires numpy.  (Note: The current code is no better than `x[0].size`.)

Motivation: I'd like to initialize a Chain by a 'dry run'.

To reproduce:
```
>>> import chainer
>>> import numpy as np
>>> l=chainer.links.Linear(4)
>>> l(np.ones((2, 3), dtype=np.float32))
variable([[-0.93691719,  1.47195649, -0.66054785,  0.62899256],
          [-0.93691719,  1.47195649, -0.66054785,  0.62899256]])
>>> l(np.ones((0, 3), dtype=np.float32))
variable([], shape=(0, 4))
>>> l=chainer.links.Linear(4)
>>> l(np.ones((0, 3), dtype=np.float32))
  File "/Users/****/lib/python3.6/site-packages/chainer/links/connection/linear.py", line 126, in __call__
    self._initialize_params(x.size // x.shape[0])
ZeroDivisionError: integer division or modulo by zero
>>> chainer.__version__
'3.0.0'
```

hvy:
Fixes #3863, avoiding 0-division when the batch dimension is 0.

### DIFF
--- a/chainer/links/connection/linear.py
+++ b/chainer/links/connection/linear.py
@@ -120,5 +120,8 @@ class Linear(link.Link):
 
         """
         if self.W.data is None:
-            self._initialize_params(x.size // x.shape[0])
+            in_size = 1
+            for dim in x.shape[1:]:
+                in_size *= dim
+            self._initialize_params(in_size)
         return linear.linear(x, self.W, self.b)
--- a/chainer/links/connection/linear.py
+++ b/chainer/links/connection/linear.py
@@ -1,3 +1,6 @@
+import functools
+import operator
+
 from chainer.functions.connection import linear
 from chainer import initializers
 from chainer import link
@@ -120,8 +123,6 @@ class Linear(link.Link):
 
         """
         if self.W.data is None:
-            in_size = 1
-            for dim in x.shape[1:]:
-                in_size *= dim
+            in_size = functools.reduce(operator.mul, x.shape[1:], 1)
             self._initialize_params(in_size)
         return linear.linear(x, self.W, self.b)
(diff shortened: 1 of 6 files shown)


chainer/links/connection/linear.py
```diff
@@ -120,5 +120,8 @@ def __call__(self, x):
 
         """
         if self.W.data is None:
-            self._initialize_params(x.size // x.shape[0])
+            in_size = 1
```
niboshi: How about `functools.reduce(operator.mul, x.shape[1:], 1)`? It's faster and easier to read than manual multiplication.

hvy: Using functools and operator now.

hvy:
@niboshi san, thank you for the comment. I fixed the code accordingly, and also added a test.

toslunar:
Have you checked other `Link`s, e.g. `L.LSTM`?

hvy:
@toslunar san, thank you for the comment. I see that this is an issue for other similar links. In addition to `Linear` and `LSTM`, I noticed that [LayerNormalization](https://github.com/chainer/chainer/blob/963e6937afaab88f5b699890066946ccde476feb/chainer/links/normalization/layer_normalization.py) suffers from this issue. I'll try fixing them too.

hvy:
I've fixed the `Linear`, `LSTM` and `LayerNormalization` links. Unless I've missed some, this PR is ready for review.

niboshi:
Thank you for the fix!

niboshi:
LGTM!

PR #3886, opened 2017-11-16
use stderr for Downloading... message

kmaehashi:
Using stderr instead of stdout for Downloading... message.
stdout is captured by doctest, so it is hard to write doctests using get_mnist etc.


hvy:
LGTM!

PR #3888, opened 2017-11-16
[backport] fix doctest failure in trainer tutorial

niboshi:
Backport of #3810


hvy:
#3886 should fix this issue (as well as other similar ones). Do we still need this PR merged?

niboshi:
As you say, eventually we will remove this fix. But I think it's better to keep it consistent between master and stable branches.

hvy:
I agree. Thanks for the quick response!

kmaehashi:
LGTM!

Issue #3633, opened 2017-10-21
Add Dockerfile for CUDA 8.0


23pointsNorth:
CUDA 7.5 is already outdated. It will be reasonable to move the Dockerfile to cuda8.0?

It will also possibly resolve #2809.

kmaehashi:
Fixed via #3902.

Issue #2809, opened 2017-06-02
example/mnist/train_mnist.py fails when using GPU


DenDen047:
When I executed `chainer/chainer/example/mnist/train_mnist.py` with GPU, the following error occurred.
It seems that it uses both of numpy.ndarray and cupy.ndarray.

I used [the latest official docker container](https://hub.docker.com/r/chainer/chainer/)
GPU : TITAN-X

```sh
root@ce05d3cc6624:~/work/chainer/examples/mnist# ./train_mnist.py --gpu=0
GPU: 0
# unit: 1000
# Minibatch-size: 100
# epoch: 20

/usr/local/lib/python2.7/dist-packages/chainer/training/extensions/plot_report.py:24: UserWarning: matplotlib is not installed on your environment, so nothing will be plotted at this time. Please install matplotlib to plot figures.

  $ pip install matplotlib

  warnings.warn('matplotlib is not installed on your environment, '
Traceback (most recent call last):
  File "./train_mnist.py", line 137, in <module>
    main()
  File "./train_mnist.py", line 133, in main
    trainer.run()
  File "/usr/local/lib/python2.7/dist-packages/chainer/training/trainer.py", line 295, in run
    update()
  File "/usr/local/lib/python2.7/dist-packages/chainer/training/updater.py", line 177, in update
    self.update_core()
  File "/usr/local/lib/python2.7/dist-packages/chainer/training/updater.py", line 189, in update_core
    optimizer.update(loss_func, *in_vars)
  File "/usr/local/lib/python2.7/dist-packages/chainer/optimizer.py", line 411, in update
    loss = lossfun(*args, **kwds)
  File "/usr/local/lib/python2.7/dist-packages/chainer/links/model/classifier.py", line 67, in __call__
    self.y = self.predictor(*x)
  File "./train_mnist.py", line 39, in __call__
    h1 = F.relu(self.l1(x))
  File "/usr/local/lib/python2.7/dist-packages/chainer/links/connection/linear.py", line 93, in __call__
    return linear.linear(x, self.W, self.b)
  File "/usr/local/lib/python2.7/dist-packages/chainer/functions/connection/linear.py", line 107, in linear
    return LinearFunction()(x, W, b)
  File "/usr/local/lib/python2.7/dist-packages/chainer/function.py", line 199, in __call__
    outputs = self.forward(in_data)
  File "/usr/local/lib/python2.7/dist-packages/chainer/functions/connection/linear.py", line 40, in forward
[...]

PR #3898, opened 2017-11-17
Fix Dockerfile to be able to train mnist with GPU

mrteera:
I also have a problem to train mnist with GPU as #2809 #3633.
Using `nvidia/cuda:8.0-cudnn6-devel` instead of `nvidia/cuda:7.5-cudnn5-devel` would fix this problem.

screen shot 2560-11-17 at 5 15 37 pm


kmaehashi:
Thanks for sending PR.
What kind of error do you experience with CUDA 7.5 & Chainer docker image?
Could you please show us the error logs, along with the version of CUDA/Docker/nvidia-docker on host?

mrteera:
I have got the same error as this post: https://github.com/chainer/chainer/issues/2809#issuecomment-305737595
```
cupy.cuda.curand.CURANDError: CURAND_STATUS_LAUNCH_FAILURE
```
Docker version: `17.06.0-ce, build 02c1d87`
nvidia-docker version: `1.0.1-1`
CUDA: `7.5.17`
Chainer: `4.0.0b1`

kmaehashi:
I found that cuRAND library included in CUDA 7.5 does not include PTX code (I could confirm that via `cuobjdump` too):
https://devtalk.nvidia.com/default/topic/941354/all-curand-samples-fail-with-gtx-1080-on-cuda-7-5/

So, some newer GPUs does not support cuRAND on CUDA 7.5.
What GPU are you using? (GTX 1080 or Pascal or later?)

I think it's OK to update Dockerfile to use CUDA 8.

docker/python2/Dockerfile
```diff
@@ -1,9 +1,10 @@
-FROM nvidia/cuda:7.5-cudnn5-devel
+FROM nvidia/cuda:8.0-cudnn6-devel
 
 RUN apt-get update -y && \
     apt-get install -y --no-install-recommends \
     python-dev \
     python-pip && \
     rm -rf /var/lib/apt/lists/* /var/cache/apt/archives/*
 
+RUN pip install -U pip wheel
```
kmaehashi: Is this line needed?

mrteera: It is likely `cupy` and `chainer` required to install `wheel` because I have got this kind of error:
```
error: invalid command 'bdist_wheel'
```

It also give a warning to upgrade pip version. (https://pip.pypa.io/en/stable/news/)
But if you want to keep the original pip from Ubuntu distribution then I can remove pip upgrade and it would be only:

```
RUN RUN pip3 install wheel cupy==4.0.0b1 chainer==4.0.0b1
```

mrteera:
@kmaehashi I have changed from `pip install wheel` to install from `python-wheel` here:
https://github.com/chainer/chainer/pull/3902.

PR #3909, opened 2017-11-18
fix Link tutorial’s linear layer document math

keisuke-nakata:
fix linear layer math document in [Official tutorial's Links section](https://docs.chainer.org/en/stable/tutorial/basic.html#links).

Currently it says "f(x) = Wx + b", but chainer's linear layer actually calculate "f(x) = xW^\top + b".  `Links.Lenear` calls https://github.com/chainer/chainer/blob/master/chainer/functions/connection/linear.py#L78, which self-documents to calculate Y = xW^\\top + b and actually this function [transposes](https://github.com/chainer/chainer/blob/master/chainer/functions/connection/linear.py#L47) the matrix $W$.


niboshi:
jenkins, test this please

niboshi:
jenkins, test this please

niboshi:
LGTM!
Thank you very much!

PR #3926, opened 2017-11-20
fix trainer tutorial evaluate example code

keisuke-nakata:
Example code in https://docs.chainer.org/en/stable/tutorial/trainer.html#evaluate-a-pre-trained-model is broken.

`y = model(x)` raises `IndexError` because `x` is just one data, not a minibatch of length 1. (I have no idea about chaier's interface yet, but maybe it works at some previous version?)

And because the resulting `y` is not `np.array` anymore, we should access `y.data` instead. (or `y.array` ? which is better?)


kmaehashi:
LGTM. Thank you for pointing this out!

Issue #3921, opened 2017-11-20
ADAM method error


pradAI:
Hi,

I am trying set Adam optimization algorithm but receiving the following error, can you please help me how to set parameters?

python3.5/site-packages/chainer-1.22.0-py3.5-linux-x86_64.egg/chainer/optimizers/adam.py", line 51, in lr
return self.alpha * math.sqrt(fix2) / fix1
ZeroDivisionError: float division by zero

Please see the below parameters what i have set.

I am traying execute the following script and it is working when I set to **MomentumSGD** but not **Adam**.
https://github.com/mitmul/ssai-cnn/blob/master/scripts/train.py

Thanks for you support.

    # Optimization settings
    parser.add_argument('--opt', type=str, default='Adam', choices=['MomentumSGD', 'Adam', 'AdaGrad'])
    parser.add_argument('--weight_decay', type=float, default=0.0001)
    parser.add_argument('--alpha', type=float, default=0.001)
    parser.add_argument('--lr', type=float, default=0.0001)
    parser.add_argument('--lr_decay_freq', type=int, default=100)
    parser.add_argument('--lr_decay_ratio', type=float, default=0.1)
    parser.add_argument('--seed', type=int, default=1701)
    args = parser.parse_args()
    return args

niboshi:
Thank you for reporting.
It looks the same issue as #3214. Could you take a look at it (and [the answer at StackOverflow](https://stackoverflow.com/questions/46035824/adam-optimizer-report-error-in-chainer)) and see if it helps?

pradAI:
Thank you for the solution. 

I tried it  but got the same error. 

from chainer.optimizers import Adam
optimizer = Adam(alpha=0.001) 

Thank you

niboshi:
I see.
Could you give the full stack trace?

(Also, I think you edited the first post by mistake. Could you revert it?)

niboshi:
Thank you detailed information.
I have not tested, but it seems the problem is that the script you are using evaluates `optimize.lr` before the first update of the optimizer (https://github.com/mitmul/ssai-cnn/blob/master/scripts/train.py#L312).

Could you remove that line and check if it solves the problem?

pradAI:
Excellent, it is working now. 

Thank you niboshi

niboshi:
You're welcome!

Issue #3214, opened 2017-08-22
Adam optimizer report: ZeroDivisionError: float division by zero


sharpstill:
version: chainer 2.0.2
I use Adam optimizer ,then report error, I found it was caused by this code(fix1==0?):
**in adam.py:**
```
@property
    def lr(self):
        fix1 = 1. - math.pow(self.hyperparam.beta1, self.t)
        fix2 = 1. - math.pow(self.hyperparam.beta2, self.t)
        return self.hyperparam.alpha * math.sqrt(fix2) / fix1
```

error log:
```
Traceback (most recent call last):
  File "AU_rcnn/train.py", line 237, in <module>
    main()
  File "AU_rcnn/train.py", line 233, in main
    trainer.run()
  File "/root/anaconda3/lib/python3.6/site-packages/chainer/training/trainer.py", line 285, in run
    initializer(self)
  File "/root/anaconda3/lib/python3.6/site-packages/chainer/training/extensions/exponential_shift.py", line 48, in initialize
    self._init = getattr(optimizer, self._attr)
  File "/root/anaconda3/lib/python3.6/site-packages/chainer/optimizers/adam.py", line 121, in lr
    return self.hyperparam.alpha * math.sqrt(fix2) / fix1
ZeroDivisionError: float division by zero

```

t-sakai-kure:
Answer?: https://stackoverflow.com/questions/46035824/adam-optimizer-report-error-in-chainer

PR #3931, opened 2017-11-21
Raise appropriate error when `Adam.lr` is evaluated before updating starts

niboshi:
It's often reported that Adam optimizer raises division by zero error (#3214, #3921).
This PR aims to provide more friendly error message for such cases.

### Code
```
import chainer
a = chainer.optimizers.Adam()
a.lr
```

### Error message (This PR)
```
  File "test.py", line 3, in <module>
    a.lr
  File "/data/work/w/repos/chainer/chainer/optimizers/adam.py", line 138, in lr
    return _learning_rate(self.hyperparam, self.t)
  File "/data/work/w/repos/chainer/chainer/optimizers/adam.py", line 19, in _learning_rate
    'Can\'t determine the learning rate of Adam optimizer '
RuntimeError: Can't determine the learning rate of Adam optimizer because the update steps have not been started.
```


niboshi:
I'm not sure if the term "learning rate" is appropriate here, because `alpha` is called "learning rate" in all other major frameworks.

niboshi:
Related: #4056

bkvogel:
As I understand it, the value returned by `lr` corresponds to `alpha_t` in the paper, which can be interpreted as a learning rate or modified `alpha`. Maybe I don't understand why a user would need to access it, but since this seems to be happening, I agree it is good to add a check as you have done. Maybe we could consider removing this property in the future if there is no good reason to access in user code? Until then, I think this is a good solution.

bkvogel:
I think the suggestion in my previous comment can wait to be consider at some point in the future, so LGTM.

PR #3932, opened 2017-11-21
Skip unnecessary array util ops

hvy:
Skip unnecessary array manipulation ops in array utility. 

This is particularly useful if functions are called with Variable arguments.

### DIFF
--- a/chainer/utils/array.py
+++ b/chainer/utils/array.py
@@ -4,10 +4,14 @@ from chainer import cuda
 
 
 def as_vec(x):
+    if x.ndim == 1:
+        return x
     return x.ravel()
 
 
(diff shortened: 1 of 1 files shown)


hvy:
IMHO, this utility is only used in three function implementations and could be removed.

niboshi:
I guess not three, but two.
+1 for removal. Maybe deprecate first?

niboshi:
LGTM!

hvy:
@niboshi Thanks for the comments. In a separate PR #3980, all usages of `utils.array` are removed and each function will raise a deprecation warning when called.

niboshi:
Thank you.

PR #3936, opened 2017-11-21
[backport] Enable optimizer model setup with instantiation

mitmul:
This PR backports #3488


niboshi:
Should we backport this PR?
This is a feature addition, and can change the recommended way of implementing a fundamental part of user's code.

I think it's better to have a trial period.

mitmul:
OK, let's postpone this after the next major version up.

PR #3937, opened 2017-11-21
[WIP] Numexpr speed-ups for activation functions and optimizers

mihirparadkar:
This implements numexpr speed-ups for most activations and the optimizers. The changes are tested and documented, but I don't know what changes are needed to the travis and appveyor configs. I'm also not sure how to test both the numexpr and numpy implementations when numexpr is installed, as I've been testing the numpy versions by uninstalling numexpr (both work).

I'd love some feedback.


Crissman:
@mihirparadkar, both AppVeyor and Travis are failing from flake8 errors. You can see this on your local repo by installing flake8, and then running `flake8` in the top directory of your Chainer repository. This will show all of the flake8 errors. For example, the first two errors are: 
```
./chainer/functions/activation/clipped_relu.py:4:1: H306  imports not in alphabetical order (chainer.numexpr_config.numexpr_enabled, chainer.numexpr_config.numexpr)
./chainer/functions/activation/clipped_relu.py:35:9: F841 local variable 'cap' is assigned to but never used
```

For the first one, change the imports so they're in alphabetical order. For the second, either remove the definition line of `cap` or correct the code below so it is used.

Let me know if that's not clear.

Thanks for your contribution to Chainer!

mihirparadkar:
I'll be able to fix the flake8 errors pretty quickly.

However, I'm primarily concerned about some results I get when benchmarking the train_mnist.py example. I profiled runs with and without numexpr, and it seems like the numexpr slows down the optimizers (activations are still faster, though). I think it's because numexpr is slower than vanilla numpy for small arrays, but I don't know why the activations are affected differently from the optimizers.

I'll ask the numexpr people about it with the minimum reproducible example I have ready. There may be a fix involving precompilation of some sort or setting a minimum parameter size for using numexpr speedup.

Crissman:
Oops. We've gone from a few errors to more, now that the imports have been concatenated to one line. If you'd like help reverting this, feel free to contact me on our [Slack channel](https://bit.ly/chainer-slack).

Crissman:
Mihir, Python 2.7 testing is having an issue with a circular import of `chainer.testing`. You can fix it by making the following changes in `attr.py`:

```
from chainer import numexpr_config
-from chainer import testing
+from chainer.testing.helper import with_requires
```

```
     @wraps(f)
-    @testing.with_requires('numexpr')
+    @with_requires('numexpr')
```

niboshi:
As for `testing/attr.py`,
I think `numexpr_config.numexpr_enabled` does not have to be configured in these decorators, because text fixtures are responsible for such configuration. So I think we don't need `attr.no_numexpr` and `attr.with_numexpr` can be defined just like `attr.cudnn`. (I think `attr.numexpr` is a better name to align with `attr.cudnn`).

niboshi:
We just refactored to introduce `chainer.backends` subpackage.
Could you move `numexpr_config` to `chainer.backends.numexpr`?

mihirparadkar:
Yes, that sounds like a welcome refactor. It'd also help me experiment with another idea of a numba backend. Some preliminary experiments there are even more promising than for the numexpr backend (notebook coming soon)

Crissman:
Mihir,

With the upcoming release of the iDeep, we would like to look at this PR and how it can be integrated into Chainer. Please let us know this week if you're interested to work on this.

Thanks!

mihirparadkar:
@Crissman 

I'll resume work on this PR this week. I also wanted to run the idea of using Numba by you guys; one of the cons of numexpr is that the VM has some overhead so it may not be faster on low-dimensional data or small batch sizes, depending on whether or not Intel VML is installed. From my experiments, numba (http://numba.pydata.org/) has less overhead and can also be used to speed up activation functions.

niboshi:
@mihirparadkar
Thanks!
As they have trade-offs, it would be great to have both implementations so that users can choose suitable one for their application.

mihirparadkar:
I've started implementing the numexpr as a backend, and I'm planning to do the same for numba shortly. However, one difference between numexpr/numba and cupy/iDeep is that numexpr/numba operate directly on numpy arrays, not ideep.mdarray or cupy.ndarray. A lot of the code for cupy/ideep backends rely on knowing the type of the array, but that doesn't work for numba/numexpr. Is there another way to enforce the use of only one backend that's already implemented?

niboshi:
What's in my mind is that a user could configure backends using, say, `USE_NUMEXPR` environment variable and/or  `chainer.config.use_numexpr`, and function impementation would check its value by using `chainer.backends.numexpr.should_use()`, just like `cuda.should_use_cudnn()` and `intel64.should_use_ideep()`. Function implementation would also need to check array types, because probably `ideep.mdarray` does not work with numexpr and numba (I'm not sure about that though).

mihirparadkar:
The environment variable seems like an idiomatic way to go about this, along with the checks for ...should_use_numexpr(). However, if both use_numexpr and use_ideep are true, is it okay to arbitrarily decide which one gets used (in a consistent fashion)?

niboshi:
Currently this is the standard way for users to customize function behavior in Chainer. (If you have any suggestion, that would be also awesome).
As for conflicting configuration, we have no decided strategy yet. Perhaps we can issue error in such case, but it can be another overhead. As for now, I think it's OK to leave it to user's responsibility (and choose arbitrarily).

mihirparadkar:
Closing, further work will show up on #4395.

PR #3940, opened 2017-11-22
Keyword-only link argument in GradientMethod

niboshi:
Improves #3488 by making `link` argument keyword-only.

```
optimizer = chainer.optimizers.SGD(model)  # does not work
```

In #3488, this code does not work and dangerous because `model` is treated as a hyperparameter.
In this PR, user must explicitly specify `link=` keyword.

```
optimizer = chainer.optimizers.SGD(link=model)
```

Also, type check on hyperparameter assignment is implemented, so the former code raises `TypeError`.

I wondered which is better keyword, `link` or `model`, but as `Optimizer.setup()` method has `link` argument, I chose `link`.

### DIFF
--- a/chainer/optimizer.py
+++ b/chainer/optimizer.py
@@ -8,6 +8,7 @@ import six
 from chainer import cuda
 from chainer import link as link_module
 from chainer import serializer as serializer_module
+from chainer.utils import argument
 from chainer import variable
 
 
@@ -53,6 +54,8 @@ class Hyperparameter(object):
 
     """
 
+    _parent = None
+
     def __init__(self, parent=None):
         self._parent = parent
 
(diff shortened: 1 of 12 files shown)


chainer/optimizer.py
```diff
@@ -61,6 +64,19 @@ def __getattr__(self, name):
             raise AttributeError('_parent is not set up yet')
         return getattr(self._parent, name)
 
+    def __setattr__(self, name, value):
+        # If the attribute is not defined as the class attribute of
+        # `Hyperparameter`, it's assumed to be a hyperparameter.
+        if not hasattr(Hyperparameter, name):
+            if not (isinstance(value, (numpy.ndarray, cuda.ndarray))
```
toslunar: I suppose this change makes Hyperparameter less flexible.

niboshi: Do you mean you may want to set a hyperparameter which is not an array nor a scalar?

toslunar: Yes.
I've encountered the case I'd like to pass a string argument to an optimizer, which syncs parameters `'soft'` or `'hard'`.  To be honest, my final code didn't pass the string, and I've not set a string value to a hyperparameter (because the optimizer was not an instance of `GradientMethod`.)

niboshi: So, we need to decide to which extent `Hyperparameter` accepts as a value.

The most conservative option is to accept anything but `Link`, but in this case we would need to add such rules in the future depending on how it is used. I don't think that's a good choice.

I think the minimum is scalars and ndarrays (as in the current code).
Maybe `str` is reasonable, too.

niboshi:
Now that this argument has been removed (See #4141), I'm closing this PR.

Issue #3878, opened 2017-11-15
Unit test fails in debug mode


niboshi:
Current master (67bc10423335b574393a392f14ea98e104f79808)

When unit test is run with environment variable CHAINER_DEBUG=1, test fails.
This is a list of tests failed under `chainer_tests/functions_tests` (all of which are failed with `RuntimeError: NaN is detected on ...`).

It may be expected in some tests, but I doubt it's so in all of them.

```
 TestAveragePooling2DCudnnCall.test_call_cudnn_backward                                                  
 TestAveragePoolingNDCudnnCall.test_call_cudnn_backward                                                  
 TestClassificationSummary.test_forward_cpu         
 TestClassificationSummary.test_forward_gpu         
 TestConvolution2DCudnnCall.test_call_cudnn_backward                                                     
 TestConvolutionNDCudnnCall.test_call_cudnn_backward                                                     
 TestConvolutionNDCudnnCall.test_call_cudnn_forward                                                      
 TestDeconvolution2DCudnnCall.test_call_cudnn_backward                                                   
 TestDeconvolution2DCudnnCall.test_call_cudnn_forward                                                    
 TestDeconvolution2DFunctionCudnnDeterministic.test_called                                               
 TestDeconvolutionNDCudnnCall.test_call_cudnn_backward                                                   
 TestDeconvolutionNDCudnnCall.test_call_cudnn_forward                                                    
 TestDilatedConvolution2DCudnnCall.test_call_cudnn_backward                                              
 TestMaxPooling2DCudnnCall.test_call_cudnn_backward                                                      
 TestMaxPooling2DCudnnCall.test_call_cudnn_forward  
 TestMaxPoolingNDCudnnCall.test_call_cudnn_backward                                                      
 TestMaxPoolingNDCudnnCall.test_call_cudnn_forward  
 TestNStepBiLSTMCudnnCall.test_call_cudnn_forward_training                                               
 TestPadSequence.test_forward_cpu                   
 TestPadSequence.test_forward_gpu                   
```

### Environment
```
$ cat /usr/local/cuda/version.txt 
CUDA Version 9.0.176
```
```
>>> cupy.cuda.cudnn.getVersion()
7003
```

niboshi:
I think the cause of errors for `test_call_cudnn_*` tests are that cuDNN wrapper functions are patched with `mock.patch` in such tests, but without `wraps` argument, thus those calls become no-op.

niboshi:
[...]

PR #3927, opened 2017-11-20
Avoid `NaN` in test of `F.classification_summary`

niboshi:
Fixes a failing case for issue #3878.

PR #3946, opened 2017-11-22
Avoid NaN error in `test_pad_sequence` in debug mode

niboshi:
Fixes a part of #3878.

### DIFF
--- a/tests/chainer_tests/functions_tests/array_tests/test_pad_sequence.py
+++ b/tests/chainer_tests/functions_tests/array_tests/test_pad_sequence.py
@@ -1,13 +1,24 @@
+import contextlib
 import unittest
 
 import numpy
 import six
 
+import chainer
 from chainer import cuda
 from chainer import functions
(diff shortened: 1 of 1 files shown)


hvy:
Are tests including nan rare of unique for this particular test? If not, we should probably define this context manager outside the test so that it can be reused, or if there are other use cases where the debug flag is desirably conditionally overridden. If that's not the case, I think we can merge this PR.

niboshi:
I thought it could be reused, as you pointed out.
The reasons why I didn't expose it are:
* Yes, i think it's rare. As long as I recognize so far, #3927 is the only case it could be reused.
* I don't think it's the best approach to conditionally disable debug mode. When we would improve the tests to eliminate the conditional in the future, we would have to remove this feature. If it were exposed, it would be compatibility breaking.

hvy:
Thanks for the comment. LGTM, and I agree that we don't need to consider this for the time being.

PR #3851, opened 2017-11-13
Move should_use_cudnn and should_use_cudnn_tensor_core to chainer.cuda

niboshi:
Moving these functions to `chainer.cuda` to keep `chainer.__init__.py` uncluttered.

PR #3974, opened 2017-11-27
backends subpackage

niboshi:
This PR introduces `chainer.backends` subpackage to include various backend modules.


niboshi:
I think this PR may be easier to review commit-by-commit, rather than by merged diff.

niboshi:
It's not dependent, but please merge #3850 and #3851 first. Otherwise conflict resolution would be cumbersome.

niboshi:
Rebased.

chainer/__init__.py
```diff
@@ -4,8 +4,8 @@
 import warnings
 
 from chainer import _version
+from chainer import backends  # NOQA
```
kmaehashi: I think it is better to fix following two lines too.

```
from chainer.cuda import should_use_cudnn  # NOQA
from chainer.cuda import should_use_cudnn_tensor_core  # NOQA
```

chainer/__init__.py
```diff
@@ -64,6 +64,7 @@
 
 
 from chainer import _environment_check
+from chainer.backends import cuda
```
kmaehashi: Do we still need this even though `cuda.py` exists?

chainer/__init__.py
niboshi: Thank you. Fixed.

kmaehashi:
LGTM, could you please resolve conflict?

niboshi:
Rebased.

kmaehashi:
Thank you! I'll merge after Travis pass.

kmaehashi:
LGTM!

kmaehashi:
I'll backport this after merging #4019.

PR #3975, opened 2017-11-28
Add ConcatWithAsyncTransfer to the reference manual

beam2d:
Missed from #3336.

### DIFF
--- a/chainer/dataset/__init__.py
+++ b/chainer/dataset/__init__.py
@@ -6,6 +6,7 @@ from chainer.dataset import iterator  # NOQA
 
 # import class and function
 from chainer.dataset.convert import concat_examples  # NOQA
+from chainer.dataset.convert import ConcatWithAsyncTransfer  # NOQA
 from chainer.dataset.convert import to_device  # NOQA
 from chainer.dataset.dataset_mixin import DatasetMixin  # NOQA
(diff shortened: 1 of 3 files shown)


hvy:
@beam2d san, thank you for this PR. At the moment the doc won't build since the module cannot be found. I think we need to import `ConcatWithAsyncTransfer ` under `chainer.dataset.__init__` to expose it, since the class is hidden under `chainer/dataset/convert`.

beam2d:
Thank you for pointing that out. I have also noticed that after seeing the CI failure. Fixed it.

hvy:
Thank you for the quick fix! 

Maybe you've noticed this too but now it seems to fail on the doctest.

```
10:16:48 File "reference/core/generated/chainer.dataset.ConcatWithAsyncTransfer.rst", line ?, in default
10:16:48 Failed example:
10:16:48     ...
10:16:48 Expected nothing
10:16:48 Got:
10:16:48     Ellipsis
10:16:48 
```

beam2d:
Oops, ok, will take a look shortly.

beam2d:
Fixed. Let's see if CI passes...

hvy:
Great, LGTM! I also verified this locally so I'll merge this one.

PR #3976, opened 2017-11-28
SliceableDataset

Hakuyume:
This is a new dataset class that supports partial fetch. With this feature, users can avoid calling heavy functions when they are not needed. For example, in the case of `LabeledImageDataset`, users can calculate the statistics of labels without loading any images.

```python
class MyDataset(sliceable.GetterDataset):
    def __init__(self):
        super().__init__()
        # add getters
        self.add_getter('img', self.get_image)
        self.add_getter('label', self.get_label)

    # usually this function is slow because it loads images
    def get_image(self, i):
        ...
        return img

    # usually fast
    def get_label(self, i):
        ...
        return label

dataset = MyDataset()
dataset[0]  # returns (img, label) using both get_image and get_label

label_dataset = dataset.slice[:, 'label']
label_dataset[0]  # returns label using only get_label (get_image is not called)

# compute statistics of labels
stats = collections.Counter(label_dataset)

# get a subset with label = 0, 1, 2
indices = [i for i, label in enumerate(label_dataset) if label in {0, 1, 2}]
dataset_012 = dataset.slice[indices]
```


yuyu2172:
I think the name of the class implies only subset of the features it supports.
Since it universally handles all functionalities of `SubDataset`, `ConcatenatedDataset`, and `PickableDataset`, how about naming it "UniversalDataset"? :smile: 

By the way, I think there are other possible names.

yuyu2172:
To make connection with https://github.com/chainer/chainer/pull/3343 clear,
I will add an example that does same thing done in the example of #3343 with @Hakuyume 's dataset.
Note that this dataset does not unnecessary load data that are not picked.

```python 
>>> from chainer.datasets import TupleDataset
>>> tuple_dataset = TupleDataset([0, 1, 2], [0, 1, 4])
>>> print(tuple_dataset.features[:, 1])
[0, 1, 4]
>>> print(tuple_dataset.features[1:, 1])
[1, 4]
```

```python
from chainer.datasets import PickableDataset
class MyDataset(PickableDataset):

    def __init__(self):
        super(MyDataset, self).__init__()
        self.keys = ('feat0', 'feat1')
        self.add_getter(self.get_feat0, 'feat0')
        self.add_getter(self.get_feat1, 'feat1')

    def __len__(self):
        return 3

    def get_feat0(self, i):
        return [0, 1, 2][i]
    
    def get_feat1(self, i):
        return [0, 1, 4][i]


dataset = MyDataset()
print(dataset.pick(keys='feat1')[:])
# [0, 1, 4]
print(dataset.sub(1, None).pick(keys='feat1')[:])
# [1, 4]
print(dataset.pick(keys='feat1').sub(1, None)[:])
# [1, 4]
```

yuyu2172:
On `transform`, the current design is not clear about whether it returns a dataset or the dataset calling it gets changed.
Similarly, `concatenate` has this ambiguity.
Also, having `concatenate` as a method is not standard. Usually, it is declared at module level `np.concatenate`.

How about declaring methods `apply_transform` and `concat_datasets`?

Hakuyume:
@yuyu2172 Thank you. I added a new module `chainer.datasets.sliceable` and renamed classes.

`SliceableDataset`: An abstract class that supports `slice`. Since `concatenate` and `transform` become functions (not methods), this class supports only `slice`.
`SliceableGetterDataset`: A class to construct a `SliceableDataset` by getter(s). Users usually use this class.

Hakuyume:
I added sliceable version of `TupleDataset`. With this class,  @yuyu2172 's example can be written as follows.

```python
dataset = TupleDataset(('feat0', [0, 1, 2]), ('feat1', [0, 1, 4]))
print(dataset.slice[:, 'feat1'][:])
# [0, 1, 4]
print(dataset.slice[1:, 'feat1'][:])
# [1, 4]
```

Hakuyume:
An example of `LabeledImageDataset`.

Without `sliceable` datasets,
```python
dataset = LabeledImageDataset('list.txt')

# compute statistics of labels
# images are loaded unnecessarily
stats = collections.Counter(int(dataset[i][1]) for i in range(len(dataset)))

# get a subset with label = 0, 1, 2
include_indices = []
exclude_indices = []
for i in range(len(dataset)):
    # images are loaded unnecessarily
    _, label = dataset[i][1]
    if label in {0, 1, 2}:
        include_indices.append(i)
    else:
        exclude_indices.append(i)
order = include_indices + exclude_indices
dataset_012 = SubDataset(dataset, 0, len(include_indices), order)
```

With `sliceable` datasets,
```python
dataset = SliceableLabeledImageDataset('list.txt')

# compute statistics of labels
# no images are loaded
stats = collections.Counter(map(int, dataset.slice[:, 'label']))

# get a subset with label = 0, 1, 2
# no images are loaded
indices = [i for i, label in enumerate(
    dataset.slice[:, 'label']) if label in {0, 1, 2}]
dataset_012 = dataset.slice[indices]
```

`SliceableLabeldImageDataset` can be implemented using `GetterDataset`.
```python
class SliceableLabeledImageDataset(sliceable.GetterDataset):
    def __init__(self, pairs, root='.'):
        super().__init__()

        with open(pairs) as f:
            self._pairs = [l.split() for l in f]
        self._root = root

        self.add_getter('image', self.get_image)
        self.add_getter('label', self.get_label)

    def __len__(self):
        return len(self._pairs)

    def get_image(self, i):
        path, _ = self._pairs[i]
        return read_image(os.path.join(self._root, path))

    def get_label(self, i):
        _, label = self._pairs[i]
        return np.int32(label)
```

Hakuyume:
@hvy (or other reviewers) Could you check the concept of `sliceable` datasets?  I'm planning to use these `sliceable` datasets in [ChainerCV](https://github.com/chainer/chainercv). I'd like to know whether they can be merged in Chainer because I think these datasets are not specific to vision tasks.

corochann:
I think this class is beneficial to make sub dataset for cross validation easy as well.

For example, when we want to split dataset based on "group" (TimeSeries data etc.) I use [`GroupKFold`](http://scikit-learn.org/stable/modules/generated/sklearn.model_selection.GroupKFold.html#sklearn.model_selection.GroupKFold), which is not supported by chainer's dataset split functionality.

Now I guess I can write as follows:

    from sklearn.model_selection import GroupKFold
    for train_idx, valid_idx in GroupKFold(n_splits=5).split(X, y, groups):
        # make subset dataset class easily by `slice` indexer
        train = dataset.slice[train_idx]
        val = dataset.slice[valid_idx]


If chainer team goes the direction that this PR to merge, I may discard the existing PR https://github.com/chainer/chainer/pull/3343 (even though part of functionality is already supported in [chainer chemistry](https://github.com/pfnet-research/chainer-chemistry/blob/master/chainer_chemistry/datasets/numpy_tuple_dataset.py#L52) ...)


BTW, I have one question & suggestion.
Are you going to introduce the same name dataset class `TupleDataset`, `ConcatenatedDataset` and `TransformDataset` under `sliceable` module?
I guess `sliceable.ConcatenatedDataset` and `sliceable.TransformDataset` `__init__` interfaces are same with existing class, then how about to keep `slicealbe.TupleDataset` `__init__` API to be same as original `TupleDataset` as well? 

Concretely, how about allowing `key` to be `int` instead of `str` and automatically increment key value for `sliceable.TupleDataset`?

So I am suggesting this kind of behavior, 

    from chainer.datasets.sliceable import TupleDataset

    # no need to specify `key` name for dataset, key is automatically assigned to 0, 1, 2...
    dataset = TupleDataset([0, 1, 2], [0, 1, 4])
    print(dataset.slice[:, 0][:])
    # [0, 1, 2]
    dataset_more = TupleDataset(dataset, [0, 1, 8])
    print(dataset_more.slice[:, [1, 2]][:])
    # [(0, 0), (1, 1), (4, 8)]

For the purpose of explicitly naming the key, ` chainer.datasets.sliceable.DictDataset` can be introduced to expect following behavior

    # need to additionally implement this class
    from chainer.datasets.sliceable import DictDataset

    # `key` can be passed as `kwargs` with DictDataset
    x0 = [0, 1, 2]
    x1 = [0, 1, 4]
    dataset = DictDataset(feat0=x0, feat1=x1)
    # this is also same.
    # dataset = DictDataset(**{'feat0': x0, 'feat1': x1})
    print(dataset.slice[:, 'feat0'][:])
    # [0, 1, 2]


I suggested this behavior to follow the existing API so that user can switch to use `sliceable.TupleDataset` from existing `TupleDataset` without addtional re-implementation cost, and explicitly naming might be a burden.
What do you think of this behavior?

Hakuyume:
@corochann Thank you for comments!

> Are you going to introduce the same name dataset class TupleDataset, ConcatenatedDataset and TransformDataset under sliceable module?

Yes. These classes are very similar to existing classes, but somehow different from them. Another solution is using a prefix `Sliceable***`, i.e. `SliceableTupleDataset`, `SliceableConcatenatedDataset`, and `SliceableTransfromDataset`. 

> I guess sliceable.ConcatenatedDataset and sliceable.TransformDataset __init__ interfaces are same with existing class, then how about to keep slicealbe.TupleDataset __init__ API to be same as original TupleDataset as well?

When sliceable datasets are passed to `sliceable.TupleDataset`, the API is similar to `TupleDataset`. Users don't need to specify keys for sliceable datasets because they have their own keys. In this situation, I think the APIs are similar. However, as you pointed out, this name is somehow confusing. Perhaps another name is better. `DictDataset` is an option but the API is no exactly same.
Please note that the API of `sliceable.TransformDataset` is not exactly same as that of `TransfromDataset`. `sliceable.TransformDataset` requires names (`keys`) of the transformed data.

Hakuyume:
@corochann 

> I suggested this behavior to follow the existing API so that user can switch to use sliceable.TupleDataset from existing TupleDataset without addtional re-implementation cost, and explicitly naming key might be a burden.

I agree with you that explicit naming is sometimes a burden. I updated sliceable datasets to accept both name and index. I also implemented auto naming for `TupleDataset`.

Now, we can write as follows.

```python
from chainer.datasets.sliceable import TupleDataset

# omit keys
dataset = TupleDataset([0, 1, 2], [0, 1, 4])
# keys are automatically named
print(dataset.keys)  # ('key_0', 'key_1')
print(dataset.slice[:, 0][:])  # [0, 1, 2]

dataset_more = TupleDataset(dataset, [0, 1, 8])
print(dataset_more.keys)  # ('key_0', 'key_1', 'key_2')
print(dataset_more.slice[:, [1, 2]][:])  # [(0, 0), (1, 1), (4, 8)]

# specify the name of keys
named_dataset = TupleDataset(('feat0', [0, 1, 2]), [0, 1, 4])
# the unspecified key is automatically named
print(named_dataset.keys)  # ('feat0', 'key_1')
# slice takes both key and index (or their mixture)
print(named_dataset.slice[:, ['feat0', 1]][:])  # [(0, 0), (1, 1), (2, 4)]
```

How about this API?

corochann:
Thank you for your very quick action!
I think the above API is ok.
I hope PR review proceeds!!! @Hakuyume @mitmul

Hakuyume:
I updated `TupleDataset` to use `None` instead of `key_***`.
```python
from chainer.datasets.sliceable import TupleDataset

# omit keys
dataset = TupleDataset([0, 1, 2], [0, 1, 4])
print(dataset.keys)  # (None, None)
print(dataset.slice[:, 0][:])  # [0, 1, 2]

dataset_more = TupleDataset(dataset, [0, 1, 8])
print(dataset_more.keys)  # (None, None, None)
print(dataset_more.slice[:, [1, 2]][:])  # [(0, 0), (1, 1), (4, 8)]

# specify the name of a key
named_dataset = TupleDataset(('feat0', [0, 1, 2]), [0, 1, 4])
print(named_dataset.keys)  # ('feat0', None)
# slice takes both key and index (or their mixture)
print(named_dataset.slice[:, ['feat0', 1]][:])  # [(0, 0), (1, 1), (2, 4)]
```

I also updated `TransformDataset` and `GetterDataset` to support the number of keys. Now, users can use all sliceable datasets without explicit naming.

mitmul:
@Crissman Any advice on names of these new datasets? All are enough intuitive?

mitmul:
Jenkins, test this please.

mitmul:
@Hakuyume Please add those dataset classes to the documentation. And please also write a brief explanation (i.e., motivation) of the `sliceable` module and its usage in `docs/source/reference/datasets.rst`. Then, please add example code to the docstrings of those dataset classes. Many of the above discussion can be used in the example code, I suppose.

mitmul:
@Hakuyume Sorry, let us consider to provide these functions in another way. So please ignore my above request temporarily.

mitmul:
It seems that the Chainer team will consider another way to provide this functionality.

PR #3981, opened 2017-11-28
Update documentation for `chainer.backends.cuda`

niboshi:
This PR depends on ~#3974~(merged).
Update documentation to follow the change `chainer.cuda` -> `chainer.backends.cuda`.

(I separated this PR because it should not be backported, whereas #3974 could be.)


niboshi:
Rebased and added a note about the rename.

docs/source/tutorial/function.rst
```diff
@@ -244,7 +244,7 @@ It can be written straight-forward as follows
    Here we used ``cuda.cupy`` instead of directly accessing :mod:`cupy`.
    This is because the ``cupy`` module cannot be imported if the CUDA is not installed.
    In order to keep the implementation valid in non-CUDA environment, we have to defer the access to the ``cupy`` module.
-   Note that the :mod:`chainer.cuda` module can be imported even if the CUDA is not installed.
+   Note that the :mod:`chainer.backends.cuda` module can be imported even if the CUDA is not installed.
    Of course, the module in such environment is almost useless, but if the interpreter does not run through the code accessing CUDA-dedicated functions, the code is still valid.
 
 The CPU and GPU implementations are almost same, except that :mod:`numpy` is replaced by :mod:`cupy` in GPU methods.
```
kmaehashi: Link `cuda.get_array_module` of the line below is dead.

kmaehashi:
LGTM except for comments.

chainer/backends/cuda.py
```diff
@@ -1,20 +1,25 @@
 """Device, context and memory management on CuPy.
 
+.. note::
+   The package ``chainer.cuda`` has been renamed to
+   :mod:`chainer.backends.cuda` as of v4.0.0, but the previous module path
+   ``chainer.cuda`` is also available.
```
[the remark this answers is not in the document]
niboshi: Hmm, from that statement, users would think it's recommended to use `chainer.cuda`.

docs/source/tutorial/function.rst
niboshi: Fixed some markups. Also, I added `from chainer.backends import cuda` in the code because the definition of `cuda` was not clearly stated.

docs/source/upgrade.rst
```diff
@@ -309,7 +309,7 @@ The retained output arrays can be accessed via :attr:`Function.output_data`.
```
kmaehashi: As this is a v1-to-v2 upgrade guide, maybe it is better to add notes like:

> `chianer.backends.cuda` was added in Chainer v3.1.0.  Use `chainer.cuda` for v3.0.0 and earlier. Their functionality is the same.

niboshi: I'm sorry for late reply.
I think we need a separate upgrade guide for v4.
I'm not sure how we should update the upgrade guide for v2.
(v2 user will confuse if this guide is updated.)

kmaehashi: If all the other things on the page are still valid, we can just rename it to "Upgrade Guide from v1 to v2 or later", and add a note I suggested.
Of course it is costly to maintain it forever, so let's set some deprecation date (maybe release of v5?) and remove the entire doc.

> I think we need a separate upgrade guide for v4.

I think big breaking changes like between v1/v2 are not introduced between v2/v3 and v3/v4, so list of breaking changes in one page may be enough I think.
I prefer the one-page style like Cython's ChangeLog: http://cython.readthedocs.io/en/latest/src/changes.html, so that users can read and update their code in chronological order.

chainer/backends/cuda.py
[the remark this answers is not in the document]
kmaehashi: Taking the fact that `cuda.to_gpu` etc. are one of the most frequently used API, no recommendation may confuse users, so I think it is better to clearly state that it will be kept for the foreseeable future.

docs/source/upgrade.rst
kmaehashi: I raised a separate issue for this: #4143.
Let's keep current Upgrade Guide as is (do not change anything in this PR)

niboshi: OK.

chainer/backends/cuda.py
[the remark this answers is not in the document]
kmaehashi: I think my concern was resolved by the description added to the upgrade guide.

kmaehashi:
@niboshi Could you remove the upgrade guide part from this PR? LGTM otherwise!

niboshi:
Changes to upgrade guide removed.

niboshi:
Conflict resolved.

niboshi:
Conflict resolved.

kmaehashi:
Jenkins, test this please.

kmaehashi:
LGTM!

PR #3996, opened 2017-11-29
add warning for batch normalization when batchsize=1 and train=True

soskek:
I have seen some friends made an unnoticed bug about a model with batch normalizations.
For performing model inference, they run the model with a minibatch which has only one example, but they forget to set `chainer.config.train` as `False`. Then, batch normalization calculates the statistics on the lonely minibatch and the output goes wrong. No error occurs.

To prevent such a sad story, this PR adds a warning for the case into batch_normalization.


niboshi:
Thank you for reporting.
What do you mean by going wrong, and what's the cause of the problem?

soskek:
Thank you for reviewing!

Batch normalization calculates vectors of mean and variance of examples and normalize values by the stats.
If only one example in minibatch, mean is the vector of the example itself, and variance is zero vector. As a result, normalization always shifts the example into zero vector (or nan/inf? due to zero 
variance devision).

See this script.
```
import numpy as np
import chainer

bn = chainer.links.BatchNormalization(4)

x3 = np.random.rand(3, 4).astype('f')
print('in\n', x3)
print('out\n', bn(x3).data)

x1 = np.random.rand(1, 4).astype('f')
print('in\n', x1)
print('out\n', bn(x1).data)
```
This result is as follows:
```
in
 [[ 0.61264682  0.59439391  0.00521886  0.79536861]
 [ 0.44258848  0.24370055  0.75638258  0.17760341]
 [ 0.84810323  0.44968283  0.19518055  0.8267982 ]]
out
 [[-0.13106315  1.14704061 -0.98362094  0.65380645]
 [-1.15349758 -1.2889055   1.37162435 -1.41275191]
 [ 1.28456116  0.14186504 -0.38800329  0.75894523]]
in
 [[ 0.68908364  0.72022897  0.47713605  0.77472299]]
out
 [[ 0.  0.  0.  0.]]
```

hvy:
How about explaining this behavior in the warning message, or include it in the docs and refer to the documentation?

soskek:
OK, I'll add and refine the warning.

soskek:
I added descriptions. I am no sure my English is clear, so please suggest anything if you feel.

niboshi:
I wonder what the expected behavior would be in the case where a minibatch happened to include only one sample in normal training.
I think that can happen depending on the dataset size and minibatch size configuration, and in such case this warning will be displayed, right?

soskek:
In training phase (train=True), batch normalization SHOULD NOT be used with minibatchsize=1, if users understand batch normalization algorithm and it breaks training.
So, if this warning appears, the user must have a mistake which maybe one of follows:
1. The user wants to TRAIN the model. But, he does not understand batch normalization, and sets minibatchsize = 1. (CASE OF BAD USER)
2. The user wants to TRAIN the model. He understands batch normalization, but sets minibatchsize = 1 due to a bug or a mistake. (CASE OF TRAINING BUG) 
3. The user wants to TEST the model with minibatchsize = 1. But, he forgets to set train=False due to a bug or a mistake. (CASE OF INFERENCE BUG)

Thus, I think there exists no person who wants to run batch normalization with this setting. In my opinion, we can even output error rather than warning.

niboshi:
Thank you for listing possible cases.
Considering that, I think it's better to raise an error if batchsize = 1 as you said, regardless of the mode.

Also how about moving the check to `F.batch_normalization`?

niboshi:
Thank you for listing possible cases.
Considering that, I think it's better to raise an error if batchsize = 1 as you said, regardless of the mode.

Also how about moving the check to `F.batch_normalization`?

soskek:
OK! I'll follow the agreement.

toslunar:
I would prefer not to raise an error.  It's possible to have batchsize=1 at the last iteration of training.

soskek:
Hmm, I got it. It happens when iterators is not repeated as you say. If the case is typical, I would use warning and modify the message slightly.

niboshi:
If that results in a valid outcome, should we still raise a warning?
It's not possible for users to selectively disable the warning.

soskek:
I am thinking about a perfect solution, but not found. Although I has not yet used the non-repeated iterator (causing batchsize=1), the usage seems to be possible. For not confusing users, no warning looks safe, i.e. closing this PR.
However, I still think users make this behavior unfortunately many times. So, if anyone has idea of good warnings, etc., please tell us or make another PR.

niboshi:
@soskek We discussed internally. Assuming it's relatively rare to specify `repeat=False`, we think we can raise a warning if there's some way to disable the warning.

soskek:
Thanks. I think it's relatively rare. So, should I add a message about the rare case in addition to the current one?

niboshi:
Could you move the warning to `F.batch_normalization`?
Yes, we should mention that could happen accidentally in the last batch.

To provide an interface for disabling the warning, please implement the following:
* Add a file `chainer/warnings.py`
* Derive `chainer.warnings.ChainerWarning` from Python's builtin `UserWarning` class with empty body.
* Derive `chainer.warnings.BatchNormalizationInsufficientBatchWarning` from `chainer.warnings.ChainerWarning`, also with empty body.
* Raise this warning instead of `UserWarning`.

I think the current message is too long.
How about like this?

*A batch with no more than one sample has been given to F.batch_normalization.
F.batch_normalization will always output a zero tensor for such batches.
This could be caused by incorrect configuration in your code (such as running evaluation while chainer.config.train=True),*
*but could also happen in the last batch of training if non-repeating iterator is used.*

soskek:
- Could you move the warning to F.batch_normalization?

OK!

- Add a file chainer/warnings.py
- Derive chainer.warnings.ChainerWarning from Python's builtin UserWarning class with empty body.
- Derive chainer.warnings.BatchNormalizationInsufficientBatchWarning from chainer.warnings.ChainerWarning, also with empty body.
- Raise this warning instead of UserWarning.

One question. These first two parts can be included in this PR, or in another PR followed by this PR because of its genericity?

The message you suggested is pretty nice for me. I'll take it.

niboshi:
@soskek Sorry for being late to merge. I fixed the conflict by myself.
I'll merge after confirming test to pass. :bowing_man:

niboshi:
jenkins, test this please

niboshi:
jenkins, test this please

niboshi:
LGTM!

PR #3551, opened 2017-10-12
Check for non-differentiable inputs

niboshi:
This PR adds an option to `numerical_grad` and `check_backward` to check for non-differentiable inputs.

PR #4003, opened 2017-11-30
Avoid numpy.stack which is supported in NumPy 1.10

niboshi:
numpy.stack is not supported until 1.10.

### DIFF
--- a/chainer/gradient_check.py
+++ b/chainer/gradient_check.py
@@ -190,8 +190,13 @@ def numerical_grad(
                     size = sizes[i_out]
                     cumsize = cumsizes[i_out]
                     shape = shapes[i_out]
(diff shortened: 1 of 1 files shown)


hvy:
LGTM except that `stack` is also used in the seq2seq example https://github.com/chainer/chainer/blob/master/examples/seq2seq/seq2seq.py#L90. The example doesn't seem to document any NumPy particular version. Could we fix this occurrence as well?

hvy:
jenkins, test this please.

hvy:
LGTM except that the docs fail due to upgraded matplotlib version. Merging for now since it is unrelated to this PR.

kmaehashi:
For record: this issue was introduced in #3551 (not backported)

PR #4005, opened 2017-12-01
Add serialize to Summary

### LINKED ISSUE #3978: Resumed reporter reports wrong values
`LogReport` reports wrong values when it was resumed.
Here is a example.

```python
import numpy as np
import chainer


class Model(chainer.Link):

    def __init__(self):
        super().__init__()
        self.i = 0

    def __call__(self, x):
        chainer.reporter.report({'loss': self.i}, self)
        self.i += 1
        return chainer.functions.sum(x)

    def serialize(self, serializer):
        self.i = serializer('i', self.i)


def get_trainer():
    dataset = np.zeros(10)
    iter_ = chainer.iterators.SerialIterator(dataset, 1)

    model = Model()
    optimizer.setup(model)

    updater = chainer.training.StandardUpdater(iter_, optimizer)

    trainer = chainer.training.Trainer(updater, (1, 'epoch'))
    trainer.extend(extensions.LogReport())
    trainer.extend(extensions.PrintReport(['main/loss']))
    trainer.extend(extensions.snapshot(), trigger=(0.5, 'epoch'))

    return trainer


print('Normal')
trainer = get_trainer()
trainer.run()

print('Resumed')
trainer = get_trainer()
chainer.serializers.load_npz('result/snapshot_iter_5', trainer)
trainer.run()
```

This code shows
```
Normal
main/loss 
4.5         
Resumed
main/loss 
7
```
Both of these values should be `4.5` because they are the mean of `0, 1, 2, 3, 4, 5, 6, 7,  8, 9`.

Hakuyume:
Fix #3978


Hakuyume:
Appveyor's failure seems not related to this PR.

tests/chainer_tests/test_reporter.py
```diff
@@ -216,5 +216,20 @@ def test_float(self):
...
+        testing.save_and_load_npz(self.summary, resumed_summary)
+        resumed_summary.add(3.)
+
+        mean = resumed_summary.compute_mean()
+        testing.assert_allclose(mean, 2.)
+
+        mean, std = resumed_summary.make_statistics()
+        testing.assert_allclose(mean, 2.)
+        testing.assert_allclose(std, numpy.sqrt(2. / 3.))
+
```
kmaehashi: Could you please add tests for `DictSummary` class?
(It is acceptable to only write test for the serialization part you added.)

chainer/reporter.py
```diff
@@ -356,3 +363,16 @@ def make_statistics(self):
             stats[name + '.std'] = std
 
         return stats
+
+    def serialize(self, serializer):
+        if isinstance(serializer, serializer_module.Serializer):
+            serializer('_names', json.dumps(list(self._summaries.keys())))
+            for name, summary in six.iteritems(self._summaries):
+                if hasattr(summary, 'serialize'):
```
kmaehashi: Is this line needed?
I think we can always assume that `summary` is a `Summary` instance.

chainer/reporter.py
```diff
@@ -356,3 +363,16 @@ def make_statistics(self):
...
+                    summary.serialize(serializer['_summaries'][name])
+        else:
+            names = json.loads(serializer('_names', ''))
```
kmaehashi: I think we should clear `self._summaries` before starting deserialization.

chainer/reporter.py
```diff
@@ -356,3 +363,16 @@ def make_statistics(self):
             stats[name + '.std'] = std
 
         return stats
+
+    def serialize(self, serializer):
+        if isinstance(serializer, serializer_module.Serializer):
+            serializer('_names', json.dumps(list(self._summaries.keys())))
+            for name, summary in six.iteritems(self._summaries):
+                if hasattr(summary, 'serialize'):
+                    summary.serialize(serializer['_summaries'][name])
```
kmaehashi: Chainer's serializer uses `/` as a separator, so allowing arbitrary keys may cause error in edge cases.

```py
>>> import chainer
>>> s = chainer.DictSummary()
>>> s.add({'/a/b': 10, 'a/b': 20})
>>> chainer.serializers.save_npz('model.npz', s)
>>> s.compute_mean()
{'/a/b': 10.0, 'a/b': 20.0}
>>> chainer.serializers.load_npz('model.npz', s)
>>> s.compute_mean()
{'/a/b': 20.0, 'a/b': 20.0}
```

chainer/reporter.py
Hakuyume: I was not sure that each item of `_summaries` is an instance of `Summary`. I will remove that line.

chainer/reporter.py
Hakuyume: Why? All values are overwritten in deserialization.

chainer/reporter.py
Hakuyume: Thank you for pointing out. I will use a hash (e.g. md5) for the key.

kmaehashi: Hm, I think using index is sufficient.

chainer/reporter.py
kmaehashi: Because users may want to deserialize into existing `DictSummary` instance which has a different set of keys.

chainer/reporter.py
Hakuyume: I see.

chainer/reporter.py
Hakuyume: OK

tests/chainer_tests/test_reporter.py
```diff
@@ -216,5 +217,208 @@ def test_float(self):
...
+        mean = summary.compute_mean()
+        testing.assert_allclose(mean, 2.)
+
+        mean, std = summary.make_statistics()
+        testing.assert_allclose(mean, 2.)
+        testing.assert_allclose(std, numpy.sqrt(2. / 3.))
+
+    def test_serialize_backward_compat(self):
+        with tempfile.NamedTemporaryFile(delete=False) as f:
```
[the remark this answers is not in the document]
Hakuyume: Without `delete=False`, the temporary file is deleted after `numpy.savez` because it closes `f`.

kmaehashi:
Jenkins test this please

tests/chainer_tests/test_reporter.py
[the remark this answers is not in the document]
kmaehashi: I think `numpy.savez` does not close the file.

```
>>> import numpy as np
>>> f = open('test.dat', 'wb')
>>> np.savez(f, np.array(10))
>>> np.savez(f, np.array(10))
```

tests/chainer_tests/test_reporter.py
```diff
@@ -216,5 +217,208 @@ def test_float(self):
...
+            # old version does not save anything
+            numpy.savez(f, dummy=0)
+            chainer.serializers.load_npz(f.name, self.summary)
```
kmaehashi: I noticed that warnigns are not catched.
Please use `with testing.assert_warns(UserWarning):` where warning is expected.
e.g. https://github.com/chainer/chainer/blob/v4.0.0b2/tests/chainer_tests/test_cuda.py#L116-L117

kmaehashi:
Jenkins test this please

tests/chainer_tests/test_reporter.py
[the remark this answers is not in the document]
Hakuyume: https://ci.appveyor.com/project/pfnet/chainer/build/8001/job/8nxy5iuv60n8wla8
AppVeyor fails without `delete=False`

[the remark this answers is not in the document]
Hakuyume: https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile

> Whether the name can be used to open the file a second time, while the named temporary file is still open, varies across platforms (it can be so used on Unix; it cannot on Windows NT or later).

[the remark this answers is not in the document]
kmaehashi: Thanks. I wasn't aware that open temporary files cannot be reopened on Windows.
Could you please revert `delete=True` commit?  Sorry to trouble you.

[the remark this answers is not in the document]
Hakuyume: Sure

kmaehashi:
Jenkins test this please

kmaehashi:
LGTM!

Hakuyume:
@kmaehashi Thank you for reviewing and merging. I think #3978 is a bug and it also exists in v3. Isn't this PR `to-be-backported`?

kmaehashi:
I agree to consider this issue as a bug.  I'll backport this.

PR #4009, opened 2017-12-01
Update VAE examples

rezoo:
The current VAE does not use `MNISTDataset` included in `chainer`, and its coding style is little bit different from examples of MNIST.
This PR adds two VAE (Variational Auto Encoder) examples as described below:

- [x] An example that uses `Trainer` extensions
- [x] An example that uses custom loops and unifies the coding-style to that of MNIST


rezoo:
I finished implementation.

examples/vae/train_vae.py
```diff
@@ -3,154 +3,120 @@
...
-        sum_rec_loss += float(model.rec_loss.data) * len(x.data)
-
-    print('train mean loss={}, mean reconstruction loss={}'
-          .format(sum_loss / N, sum_rec_loss / N))
-
-    # evaluation
-    sum_loss = 0
-    sum_rec_loss = 0
+import matplotlib
+matplotlib.use('Agg')
```
mitmul: In the README.md of this example, how to specify the backend of matplotlib by specifying an environment variable is already shown. So, I think these lines (L:15-16) are not needed.

examples/vae/train_vae.py
```diff
@@ -3,154 +3,120 @@
...
+    print('# dim z: {}'.format(args.dimz))
+    print('# Minibatch-size: {}'.format(args.batchsize))
+    print('# epoch: {}'.format(args.epoch))
+    print('')
+
+    # Prepare VAE model, defined in net.py
+    model = net.VAE(784, args.dimz, 500)
+    if args.gpu >= 0:
+        chainer.cuda.get_device_from_id(args.gpu).use()
+        model.to_gpu()
```
mitmul: This example is using `StandardUpdater` with `device=args.gpu` argument, and the updater automatically transfer the model to the specified GPU in its constructer (https://github.com/chainer/chainer/blob/master/chainer/training/updater.py#L144-L146), so these part (L:47-49) is actually not needed. So how about removing these lines?

examples/vae/train_vae_custom_loop.py
```diff
@@ -0,0 +1,150 @@
...
+import os
+
+import chainer
+from chainer.dataset import convert
+import numpy as np
+
+import net
+
+import matplotlib
+matplotlib.use('Agg')
```
mitmul: Ditto. (these two lines can be removed)

examples/vae/train_vae.py
```diff
@@ -3,154 +3,120 @@
...
+    train_ind = [1, 3, 5, 10, 2, 0, 13, 15, 17]
+    x = chainer.Variable(np.asarray(train[train_ind]))
+    with chainer.no_backprop_mode():
+        x1 = model(x)
+    save_images(x.data, os.path.join(args.out, 'train'))
+    save_images(x1.data, os.path.join(args.out, 'train_reconstructed'))
+
+    test_ind = [3, 2, 1, 18, 4, 8, 11, 17, 61]
+    x = chainer.Variable(np.asarray(test[test_ind]))
     with chainer.no_backprop_mode():
```
mitmul: The network used here doesn't have any layers which switch their behavior by `chainer.config.train`, but users might modify the network and use this training script as it is. So, how about adding `chainer.using_config('train', False)` context along with this? You can specify both by `with chainer.using_config('train', False), chainer.no_backprop_mode()`.

examples/vae/train_vae_custom_loop.py
```diff
@@ -0,0 +1,150 @@
...
+    train_ind = [1, 3, 5, 10, 2, 0, 13, 15, 17]
+    x = chainer.Variable(np.asarray(train[train_ind]))
+    with chainer.no_backprop_mode():
+        x1 = model(x)
+    save_images(x.data, os.path.join(args.out, 'train'))
+    save_images(x1.data, os.path.join(args.out, 'train_reconstructed'))
+
+    test_ind = [3, 2, 1, 18, 4, 8, 11, 17, 61]
+    x = chainer.Variable(np.asarray(test[test_ind]))
+    with chainer.no_backprop_mode():
```
mitmul: Ditto. (how about adding `chainer.using_config('train' False)`?)

rezoo:
Thank you for your comments! I've fixed.

mitmul:
LGTM

Issue #3994, opened 2017-11-29
`variable.debug_print()` when `variable.data` is `None` **bug**


EmbraceLife:
**environment:** 
MacOS Higher Sierra 10.13.1
chainer version 3.0.0
numpy version 1.13.3

**code to run to cause the error**
```python
import chainer
v = chainer.Variable()
v.debug_print()
```
**error message**
`TypeError: unsupported operand type(s) for /: 'NoneType' and 'int'`

**source of error located** 
https://github.com/chainer/chainer/blob/master/chainer/variable.py#L524 
`xp.mean(self.data)` refers to `numpy.mean(v.data)`, referring to `numpy.mean(None)`, which cause the same error message below
`TypeError: unsupported operand type(s) for /: 'NoneType' and 'int'`

**a solution proposed**
```python
if self.data is not None: 
   stats = stats_msg.format(float(xp.mean(self.data)),
                                     float(xp.std(self.data)))
else:
   stats = "None"

return msg.format(summary=self.summary(),
                          grad=grad, shape=self.data.shape if self.data is not None else None,
                          background=type(self.data) if self.data is not None else None,
                          dtype=self.data.dtype if self.data is not None else None, device=device,
                          stats=stats)
```

hvy:
@EmbraceLife, thanks for pointing this out! I'll make a quick fix.

kmaehashi:
This issue will be fixed in the next release (#4018).
Thanks again for reporting!

PR #4018, opened 2017-12-04
Fix debug_print() with empty Variable

hvy:
Fixes an error that is raised when calling `debug_print()` on a `Variable` object without `data` (https://github.com/chainer/chainer/issues/3994).

### DIFF
--- a/chainer/variable.py
+++ b/chainer/variable.py
@@ -499,7 +499,7 @@ Actual: {0}'''.format(type(data))
 
         msg = """{summary}
 - device: {device}
-- backend: {background}
+- backend: {backend}
 - shape: {shape}
 - dtype: {dtype}
 - statistics: {stats}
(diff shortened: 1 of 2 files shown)


chainer/variable.py
```diff
@@ -515,6 +515,15 @@ def debug_print(self):
         with cuda.get_device_from_array(self.data) as dev:
             xp = numpy if int(dev) == -1 else cuda.cupy
 
+            if self.data is None:
+                backend = shape = dtype = stats = None
```
kmaehashi: * Does it make sense to print `device: CPU` when data is None?
* Please add a test case for this.

hvy: - Changed to `device: None`.
- Added test.

hvy:
@kmaehashi san, thanks for the comments. I made changes accordingly (and simplified the code a bit). Could you take another look?

chainer/variable.py
```diff
@@ -499,38 +499,38 @@ def debug_print(self):
...
 - grad: {grad}"""
 
         stats_msg = 'mean={0:.8f}, std={1:.8f}'
 
-        try:
-            device = self.data.device
-        except AttributeError:
-            device = 'CPU'
-
-        with cuda.get_device_from_array(self.data) as dev:
```
kmaehashi: Why is this device context removed?
I think this is needed when the `self.data` is on GPU other than the current GPU.

hvy: Thanks for pointing that out. I was not aware of this requirement. Fixed.

hvy:
Ready for another review.

hvy:
Ready for another review. Fyi, I squad all commits.

chainer/variable.py
```diff
@@ -499,38 +499,42 @@ def debug_print(self):
...
+                stats = stats_msg.format(float(xp.mean(data)),
+                                         float(xp.std(data)))
+            shape = getattr(data, 'shape', None)
+            dtype = getattr(data, 'dtype', None)
+
             if self.grad is None:
                 grad = None
             elif xp.all(self.grad == 0):
-                grad = 0
+                grad = 0.0
```
toslunar: Is this change related to the bugfix?

hvy: The motivation was to unify the type with mean and std, but it does not have to be included in this PR as you point out. If this grad needs to be an int, I can change it back.

hvy: It probably shouldn't be included in this PR anyhow. Reverted the change. Still, please let me know if there is a particular reason to keep it this way.

hvy:
Rebased.

kmaehashi:
LGTM!

PR #4026, opened 2017-12-04
Serialize t of UpdateRule

### LINKED ISSUE #4013: Adam-based training with resumed trainer changes even if randomness is removed
A trainer with Adam has been run until the end and it did snapshot at a middle checkpoint.
After that, if I run a trainer which is resumed with the snapshot, its (future) result is different from the end. Even if random seed is fixed and other explicit randomness is removed, it happens.
Furthermore, I found trainers with SGD do not cause the behavior.

Reproduction code is here.
https://gist.github.com/soskek/3df39e70b116fba848eaa402cb9bbf5a

You can see the behavior as follows:
```
# Result changes if using Adam
python train_mnist.py --adam
python train_mnist.py --adam --resume result/snapshot_iter_4 --out result2
# Not if using SGD
python train_mnist.py
python train_mnist.py --resume result/snapshot_iter_4 --out result2
```

toslunar:
`t` of `UpdateRule` should be serialized since `UpdateRule.update` may depend on `t` (e.g. `AdamRule` does).

As well as `Optimizer.t` is a state, `UpdateRule.t` is a state in the sense of the document of `Trainer`:
> The serialization does not recover everything of the training loop. It only recovers the states which change over the training

The PR breaks the compatibility if an update rule already has`'t'` in `self.state`.

Resolves #4013.

### DIFF
--- a/chainer/optimizer.py
+++ b/chainer/optimizer.py
@@ -274,6 +274,7 @@ class UpdateRule(object):
             serializer (~chainer.AbstractSerializer): Serializer object.
 
         """
+        self.t = serializer('t', self.t)
         if self.state is None:
             if isinstance(serializer, serializer_module.Deserializer):
                 # try to initialize the state to retrieve state entries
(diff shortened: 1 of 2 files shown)


okuta:
jenkins, test this please.

PR #4029, opened 2017-12-04
Check for mixed use of ndarrays

niboshi:
Checks input/output array types of forward operations.

This PR may be unsafe to backport, because existing functions may be mixing cupy and numpy arrays.
Also, some builtin functions will fail to pass this check. They need to be fixed before this PR.

### DIFF
--- a/chainer/function_node.py
+++ b/chainer/function_node.py
@@ -222,8 +222,20 @@ Use apply() method instead.\
         in_data = tuple([x.data for x in input_vars])
         requires_grad = any([x.requires_grad for x in input_vars])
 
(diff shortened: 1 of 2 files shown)


niboshi:
#4030 needs to be merged before this PR.

hvy:
jenkins, test this please.

hvy:
jenkins, test this please.

hvy:
LGTM!

hvy:
Note: Not backported due to the fact that e.g. user-defined functions that are using mixed input/output backends will start to fail.

PR #4030, opened 2017-12-04
Avoid mixing cupy.ndarray and numpy.ndarray in n_step_xxx links

niboshi:
`indices` is always `numpy.ndarray`, leading to mixture of numpy/cupy ndarrays.

This fix can be tested with #4029, but I separated the PRs because #4029 might be unsafe to backport.


hvy:
LGTM. Verified locally using #4029, so will merge this once CI passes.

PR #4032, opened 2017-12-04
added initial version of AMSGrad

kashif:


hvy:
@kashif Thanks again for this interesting PR. Please let us know when you consider it ready for review. 

As a side note, a plot reproducing the experiments in the paper  (comparison with Adam) would be great.

kashif:
@hvy sure thats what I intend to do...

kashif:
thanks to @tkerola for the comparison below:

Loss:

kashif:
@hvy so would be ready for an initial discussion... in keras, and pytorch PR I have added this as an option to Adam optimizer... would it make sense to do the same here? Its a bit more complicated since the if, else will be more messy?

hvy:
For the record, this PR is synced with #4050 since they are closely related (potentially working with the same API).

kashif:
@hvy I have moved the amsgrad as an option to adam rather than an optimizer in itself.

chainer/optimizers/adam.py
```diff
@@ -151,21 +191,24 @@ def __init__(self,
...
+                 weight_decay_rate=_default_hyperparam.weight_decay_rate,
+                 amsgrad=_default_hyperparam.amsgrad):
         super(Adam, self).__init__()
         self.hyperparam.alpha = alpha
         self.hyperparam.beta1 = beta1
         self.hyperparam.beta2 = beta2
         self.hyperparam.eps = eps
         self.hyperparam.eta = eta
         self.hyperparam.weight_decay_rate = weight_decay_rate
+        self.amsgrad = amsgrad
```
tkerola: I think this should be `self.hyperparam.amsgrad = amsgrad`.

chainer/optimizers/adam.py
```diff
@@ -81,8 +92,15 @@ def update_core_cpu(self, param):
...
-                                hp.weight_decay_rate * param.data)
+
+        if hp.amsgrad:
+            vhat = self.state['vhat']
+            vhat = np.maximum(vhat, v)
+            param.data -= hp.eta * (self.lr * m / (np.sqrt(vhat) + hp.eps) +
+                                    hp.weight_decay_rate * param.data)
+        else:
+            param.data -= hp.eta * (self.lr * m / (np.sqrt(v) + hp.eps) +
+                                    hp.weight_decay_rate * param.data)
```
hvy: Do you think it's possible to reduce code duplication by maybe introducing another (local) variable?

chainer/optimizers/adam.py
```diff
@@ -135,13 +169,19 @@ class Adam(optimizer.GradientMethod):
     See: `Fixing Weight Decay Regularization in Adam \
           <https://openreview.net/forum?id=rk6qdGgCZ>`_
 
+    Also added a flag ``amsgrad`` to use the AMSGrad variant of
```
hvy: I think that "Also added " can be skipped since we don't need to any tenses.

hvy:
Sorry for the delayed response. I  added some smaller comments. Do you mind taking another look?

hvy:
I couldn't come up with a straight forward implementation and maybe you have considered this, but do you think it is possible to reduce the kernel code duplications?

kashif:
@hvy yes I didn't want to add if-else in the cuda kernel... but perhaps the newer cuda compilers can use predicated execution to optimize the kernel... its just when i learned cuda it was discouraged to use if-else... have things changed?

hvy:
I am not aware of any optimizations done by the compiler (will double check) but I did some benchmarks with a naively implemented unified kernel with if-else (takes an additional `bool` whether or not amsgrad is used)

| Kernel implementation | amsgrad | Kernel average running time (ms) |
|--|--|--|
| Unified kernels w/ if-else | `false` |  0.10363 |
| Unified kernels w/ if-else | `true` | 0.11507 |
| Separate Kernel w/o if-else | `false` (existing Adam) | 0.096428 |
| Separate Kernel w/o if-else | `true` | 0.10964 |

(Average over 50000 iters on `train_mnist.py`)

What do you think? Maybe it is okay to keep it as is. I am also not sure how to deal with the vhat kernel output that's only required when using amsgrad. In the experiments above, I simply always created the vhat ndarray.

kashif:
@hvy thank you for the benchmarks! 🙇  I also think to keep it as is...

hvy:
Okay, then let's leave it as it is right now.  

Do you mind fixing the flake8 error which seems to be caused by a trailing whitespace and maybe consider https://github.com/chainer/chainer/pull/4032/files#r168401871 ?

hvy:
jenkins, test this please.

chainer/optimizers/adam.py
kashif: yes sure... I can do that let me push it in a few mins

hvy:
LGTM, @tkerola do you think this PR is ready to get merged?

chainer/optimizers/adam.py
```diff
@@ -81,7 +92,13 @@ def update_core_cpu(self, param):
 
         m += (1 - hp.beta1) * (grad - m)
         v += (1 - hp.beta2) * (grad * grad - v)
-        param.data -= hp.eta * (self.lr * m / (np.sqrt(v) + hp.eps) +
+
+        if hp.amsgrad:
+            vhat = self.state['vhat']
+            vhat = np.maximum(vhat, v)
```
toslunar:
```
            np.maximum(vhat, v, out=vhat)
```
to update the state

tkerola:
To reduce the duplicated GPU code, how about using string formatting? `'vhat = {};'.format('max(vhat, v)' if hp.amsgrad else 'v'`. Or would that look cryptic?

kashif:
@tkerola I will try it and see how cryptic it looks...

kashif:
@tkerola it would look something like:

```python
        kernel = '''
            m += one_minus_beta1 * (grad - m);
            v += one_minus_beta2 * (grad * grad - v);
            vhat = {};
            param -= eta * (lr * m / (sqrt(vhat) + eps) +
                            weight_decay_rate * param);
        '''.format('max(vhat, v)' if hp.amsgrad else 'v')
        cuda.elementwise(
            'T grad, T lr, T one_minus_beta1, T one_minus_beta2, T eps, \
                T eta, T weight_decay_rate',
            'T param, T m, T v, T vhat',
            kernel, 'adam')(grad, self.lr, 1 - hp.beta1,
                    1 - hp.beta2, hp.eps,
                    hp.eta, hp.weight_decay_rate,
                    param.data, self.state['m'], self.state['v'],
                    self.state['vhat'] if hp.amsgrad else self.state['v'])
```

Note I would need to send some argument for `vhat` for the case when `hp.amsgrad = False` since its in the argument for both cases... at the moment I am sending it `v` which is doing an unnecessary copy of `vhat = v` in that case... what do you think?

tkerola:
Thanks for the suggestion! To avoid an unnecessary copy, how about writing it perhaps like this?
```python
kernel_params = 'T grad, T lr, T one_minus_beta1, T one_minus_beta2, T eps, \
                 T eta, T weight_decay_rate'
kernel_rets = 'T param, T m, T v'
kernel = '''
    m += one_minus_beta1 * (grad - m);
    v += one_minus_beta2 * (grad * grad - v);
    {}
    param -= eta * (lr * m / (sqrt({}) + eps) +
                    weight_decay_rate * param);
'''.format(*(('vhat = max(vhat, v);', 'vhat') if hp.amsgrad else ('', 'v')))
args = (grad, self.lr, 1 - hp.beta1,
            1 - hp.beta2, hp.eps,
            hp.eta, hp.weight_decay_rate,
            param.data, self.state['m'], self.state['v'])
if hp.amsgrad:
   args += (self.state['vhat'],)
   kernel_rets += ', T vhat'
cuda.elementwise(
    kernel_params,
    kernel_rets,
    kernel, 'adam')(*args)
```

hvy:
@kashif Thanks for the quick response as always. 

I am starting to think that the original separate kernel implementations were preferable, both in terms of readability and performance. I am unsure whether coupling the code tightly to the `vhat` is really maintainable in the long run. Sorry for the inconvenience but would you mind reverting it back to your original implementation? Of course, if you have other opinions, please let us know! @tkerola

tkerola:
I think you might be right. Sorry for the inconvenience.

kashif:
@hvy @tkerola no problem, I will revert it

hvy:
jenkins, test this please.

hvy:
It seems like the doctest build is failing in our CI environment due to some indentation error. Could you try to to resolve it (`make html` from the docs directory)?

kashif:
@hvy done thanks!

hvy:
jenkins, test this please.

hvy:
Thanks for the quick fix. I just restarted the CI build!

hvy:
jenkins, test this please.

hvy:
jenkins, test this please.

hvy:
@kashif thanks for the long-running PR, LGTM!

kashif:
thank you @hvy it was a great learning experience!

PR #4036, opened 2017-12-05
Fix cuda import path

hvy:
Fixed cuda import path to go via the new `chainer.backends` submodule.


hvy:
Backport after https://github.com/chainer/chainer/pull/4035 (backport).

niboshi:
LGTM!

PR #4052, opened 2017-12-06
Linking to Chainer function or Python function for docs

Crissman:
Change documentation for functions that use Chainer functions as arguments to link to the Chainer functions page.

Likewise, change documentation for functions that use Python functions as arguments to link to the Python pages.

To address #1587


hvy:
Thanks for the fixes! I have not yet taken a proper look but it seems Jenkins isn't too happy building the html.
https://ci-chainer.pfidev.jp/job/chainer_pr/2673/TEST=chainer-doc,sakura11=sakura11/console

chainer/gradient_check.py
```diff
@@ -33,8 +33,8 @@ def numerical_grad(
     ``eps``.
 
     Args:
-        f (function): Python function with no arguments that runs forward
-            computation and returns the result.
+        f (`Python function`_): Python function with no arguments\
```
niboshi: How about "`callable`" rather than "Python function"? It's more precise.
Same for other instances.

chainer/function.py
```diff
@@ -94,7 +94,8 @@ class FunctionAdapter(function_node.FunctionNode):
        object changes the reference to a weak one.
 
     Args:
-        function (Function): The function object to wrap.
+        function (:doc:`Chainer function </reference/functions>`):\
```
niboshi: This should link to `chainer.Function` class.

hvy:
It seems like some unrelated code was removed in the last commit.

hvy:
jenkins, test this please.

chainer/function.py
```diff
@@ -94,7 +94,8 @@ class FunctionAdapter(function_node.FunctionNode):
        object changes the reference to a weak one.
 
     Args:
-        function (Function): The function object to wrap.
+        function (~chainer.Function):\
```
niboshi: Avoid line continuation with `\`. Instead, indent the following line correctly. Applied to other instances.

niboshi:
Those links described as "Chainer function" may be confusing, because they do not actually accept `chainer.Function` nor `chainer.FunctionNode`. They accept any Python callable which maps  `chainer.Variable`(s) to `chainer.Variable`(s).

One idea is to make a dedicated section in `reference/functions.rst` for describing such usage, and link to it using such notation as ``` :ref:`Chainer function <function_argument>` ```. You need ```.. _function_argument:``` in the section. That's exactly what's done to describe initializer arguments (ref. docs of `L.Linear`).
How do you think?

toslunar:
In the documents, a "function" is required to satisfy a certain subset of the conditions

1. `backward`-able,
2. parameter-less,

in addition to be a callable.
The first condition I stated is just as in the docstring of `chainer.gradient_check.check_backward`:
```rest
func (callable): A function which gets :class:`~chainer.Variable`\ s
    and returns :class:`~chainer.Variable`\ s. ``func`` must returns
    a tuple of :class:`~chainer.Variable`\ s or one
    :class:`~chainer.Variable`. You can use a
    :class:`~chainer.Function`, :class:`~chainer.FunctionNode` or a
    :class:`~chainer.Link` object or any other function satisfying the
    condition.
```

The second condition is for model components (e.g. `activation`, `lossfun`, etc.)  For example, `L.Classifier` does not register (learnable) parameters from the `lossfun` argument.

It's great to add the information that `chainer.functions.foo_bar_in_lowercase` is a wrapper (of `chainer.functions.FooBar`).  However there's no "function" argument requiring the condition

- being a wrapper function of either a Function object or a FunctionNode object.

It can be a composite of several Function objects.

Crissman:
@toslunar Changed "Function Wrapper" to "wrapper function" to make clear that it could be a composite of several functions, and added the requirement that the wrapper function be "backward-able," i.e. differentiable into the wrapper function explanation section.

Please review!

chainer/links/connection/mlp_convolution_2d.py
```diff
@@ -33,7 +33,7 @@ class MLPConvolution2D(link.ChainList):
         pad (int or pair of ints): Spatial padding width for input arrays at
             the first convolution layer. ``pad=p`` and ``pad=(p, p)`` are
             equivalent.
-        activation (:doc:`Function Wrapper </reference/functions>`): Activation function for internal hidden units.
+        activation (:doc:`function wrapper </reference/functions>`): Activation function for internal hidden units.
```
toslunar: typo (wrapper function)

docs/source/reference/functions.rst
```diff
@@ -1,7 +1,8 @@
-Function Wrappers
+Wrapper Functions
 =================
 
-These functions take an array or a Chainer :class:`Variable` as an argument and return a Chainer :class:`Variable`.
+Wrapper functions are differentiable (backward-able) functions, that take an array or a Chainer :class:`Variable` as an argument and return a Chainer :class:`Variable`, or a tuple of a Chainer :class:`Variable`.
+Wrapper functions may not have learnable parameters when used in models. This is to prevent the functions from changing output due to adjustment of learned parameters during training.
```
toslunar: I meant that a "backward-able" callable should be a `Link` whenever it has parameters (to be trained).  In other words, links that are not added to a target chain of an optimizer will remain unchanged.

The main message would be that we have only to call "wrapper functions" to do differentiable computation, some sentences like
```
Wrapper functions do not have learnable parameters,
so that they do not need to be added to a :class:`Chain` to be trained.
```

chainer/links/connection/mlp_convolution_2d.py
```diff
@@ -33,7 +33,7 @@ class MLPConvolution2D(link.ChainList):
         pad (int or pair of ints): Spatial padding width for input arrays at
             the first convolution layer. ``pad=p`` and ``pad=(p, p)`` are
             equivalent.
-        activation (function): Activation function for internal hidden units.
+        activation (:doc:`function wrapper </reference/functions>`): Activation function for internal hidden units.
```
toslunar: typo (wrapper function)

docs/source/reference/functions.rst
```diff
@@ -1,5 +1,23 @@
-Standard Function implementations
-=================================
+Wrapper Functions
+=================
+
+Wrapper functions are differentiable (backward-able) functions, that take an array or a Chainer :class:`Variable` as an argument and return a Chainer :class:`Variable`, or a tuple of a Chainer :class:`Variable`.
+Wrapper functions may not have learnable parameters when used in models. This is to prevent the functions from changing output due to adjustment of learned parameters during training.
```
toslunar: I meant that a "backward-able" callable should be a `Link` whenever it has parameters (to train).  In other words, callables (including links) that are not added to a target chain of an optimizer will remain unchanged.

The main message would be that we have only to call "wrapper functions" to do differentiable computation, i.e. some sentences like
```
Wrapper functions do not have learnable parameters,
so that they do not need to be added to a :class:`Chain`.
```

Crissman:
Thanks, @toslunar ! Typo and additional comments reflected in docs.

docs/source/reference/functions.rst
```diff
@@ -1,8 +1,8 @@
 Wrapper Functions
 =================
 
-Wrapper functions are differentiable (backward-able) functions, that take an array or a Chainer :class:`Variable` as an argument and return a Chainer :class:`Variable`, or a tuple of a Chainer :class:`Variable`.
-Wrapper functions may not have learnable parameters when used in models. This is to prevent the functions from changing output due to adjustment of learned parameters during training.
+Wrapper functions are backward-able functions (probably differentiable), that take an array or a Chainer :class:`Variable` as an argument and return a Chainer :class:`Variable`, or a tuple of a Chainer :class:`Variable`.
+Wrapper functions should not have learnable parameters when used in models and should not be members of the :class:`Chain`. This is to prevent the functions from changing output due to adjustment of learned parameters during training.
```
toslunar: It might be rare but wrapper functions can be members of a model as below.  I'm in favor of warning that adding a wrapper function to an instance of `Link` as a member has no special effect by Chainer.  No special effect doesn't mean it shouldn't.

An example is `activation` in `chainer/links/connection/mlp_convolution_2d.py`.  Similar example exists in ChainerRL, where a ready-to-use model class has a `self.nonlinearity` member in order to select activation `F.relu`, `F.tanh`, etc.

niboshi:
jenkins, test this please

niboshi:
Also,
* some markups of `Variables` are not correct. ```:class:`Variable` ``` -> ```:class:`~chainer.Variable` ```
* `Args:` and `Returns:` should not be used, as this is not documentation of "a function". Instead they should be described in sentences.

Crissman:
Both comments addressed. Thanks!

niboshi:
jenkins, test this please

docs/source/reference/core/function.rst
```diff
@@ -11,3 +11,13 @@ Function
    chainer.force_backprop_mode
    chainer.no_backprop_mode
    chainer.grad
+
+Wrapper Functions
+=================
+
+Wrapper functions are backward-able functions (probably differentiable), that take an array or a Chainer :class:`~chainer.Variable` as an argument and return a Chainer :class:`~chainer.Variable`, or a tuple of a Chainer :class:`~chainer.Variable`.
```
niboshi: * How about
"Wrapper functions are backward-able (probably differentiable) plain Python functions that take..."
I mean, just saying "functions" is ambiguous.

* Description of inputs/outputs is duplicated with the following paragraphs. Can we simplify it?

docs/source/reference/core/function.rst
```diff
@@ -11,3 +11,13 @@ Function
...
+Wrapper functions should not have learnable parameters when used in models and are usually not members of the :class:`~chainer.Chain`. Even if they do have learnable parameters, these are ignored by Chainer training, to prevent the functions from changing output due to adjustment of learned parameters during training.
```
niboshi: I don't think we have to mention the possibility of wrapper functions having learnable parameters, because Chainer does not assume such usage (correct me if I'm wrong).

How about:
"Wrapper functions do not have learnable parameters and are thus usually not members of Chain." (end of the paragraph)

docs/source/reference/core/function.rst
```diff
@@ -11,3 +11,13 @@ Function
...
+
+Argument inputs are tuples of input :class:`~chainer.Variable`, :class:`~numpy.ndarray` or :class:`~cupy.ndarray` objects. If the input is a :class:`~numpy.ndarray` or a :class:`~cupy.ndarray`, it is automatically wrapped with :class:`~chainer.Variable`.
+
+Wrapper functions returns a :class:`~chainer.Variable` object or a tuple of multiple :class:`~chainer.Variable` objects.
```
niboshi: returns -> return

docs/source/reference/core/function.rst
```diff
@@ -11,3 +11,13 @@ Function
...
    chainer.no_backprop_mode
    chainer.grad
+
+Wrapper Functions
+=================
+
+Wrapper functions are backward-able functions (probably differentiable), that take an array or a Chainer :class:`~chainer.Variable` as an argument and return a Chainer :class:`~chainer.Variable`, or a tuple of a Chainer :class:`~chainer.Variable`.
+Wrapper functions should not have learnable parameters when used in models and are usually not members of the :class:`~chainer.Chain`. Even if they do have learnable parameters, these are ignored by Chainer training, to prevent the functions from changing output due to adjustment of learned parameters during training.
+
+Argument inputs are tuples of input :class:`~chainer.Variable`, :class:`~numpy.ndarray` or :class:`~cupy.ndarray` objects. If the input is a :class:`~numpy.ndarray` or a :class:`~cupy.ndarray`, it is automatically wrapped with :class:`~chainer.Variable`.
```
niboshi: Arrays are not limited `numpy.ndarray` and `cupy.ndarray`, but also can be `ideep.mdarray`.

How about 
"Argument inputs are tuples of Variable or array (e.g. numpy.ndarray and cupy.ndarray) objects. If the input is an array, it is ..."

docs/source/reference/functions.rst
```diff
@@ -1,5 +1,5 @@
-Standard Function implementations
-=================================
+Standard wrapper function implementations
```
niboshi: I think this section title should be left as it was.

docs/source/reference/core/function.rst
Crissman: This wording was worked out with help from @toslunar . I think it should remain as is, to make it clear it can have learnable parameters, but they'll have no effect.

niboshi:
No, it's just an usual numeric error.

niboshi:
LGTM!

PR #3494, opened 2017-10-04
Grouped convolution

anaruse:
The aim of this PR is to enable use of grouped convolution from Chainer.

An argument `group` is added to `Convolution2D` class, so you can easily use grouped convolution by just adding the argument `group` with group count to convolution layers in your network like below.

    conv1 = L.Convolution2D(..., group=8)

This variant of convolution works like a depthwise convolution when group count is equal to number of input/output channels.

You can use this convolution on GPU with cuDNN7, GPU with cupy and CPU with numpy. Please note that https://github.com/cupy/cupy/pull/581 is necessary to use this convolution with cuDNN7. 

AIs
* Add tests

PR #4059, opened 2017-12-08
Fix group argument of {de,}convolution_2d to keyword-only argument.

takagi:
This PR fixes `group` argument of convolution/deconvolution to a keyword-only argument,.

`group` argument is introduced in #3494, but convolution/deconvolution once had `deterministic` argument, which is now removed, at the same argument position. So it would be better to provide `group` argument as a keyword-only argument to avoid the collision.

(This change does not violate the compatibility policy if merged before v4.0.0 official release)

### DIFF
--- a/chainer/functions/connection/convolution_2d.py
+++ b/chainer/functions/connection/convolution_2d.py
@@ -59,7 +59,7 @@ def _get_algorithm_bwd_filter(
 
 class Convolution2DFunction(function_node.FunctionNode):
 
-    def __init__(self, stride=1, pad=0, cover_all=False, group=1, **kwargs):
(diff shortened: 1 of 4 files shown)


bkvogel:
Thank you for the PR! Sorry for the delay in reviewing it. If I understand correctly, you want to make `group` a keyword argument to avoid problems with old code that used `deterministic` failing without user-friendly reasons. I think that seems fine. It also seems that there was no documentation for `group` when it was originally added, and so you added it, which is also good. Is that correct? 
Could you fix the conflicts?

okuta:
@takagi  Cloud you resolve conflicts?

okuta:
I send new PR.

PR #4064, opened 2017-12-08
Dilated conv doesn't work with TensorCore in cuDNN7

anaruse:
This PR is to fix an issue related to combination of dilated convolution and TensorCore in cuDNN7.

As the title indicates, there is no convolution algorithm in cuDNN7 that supports dilated convolution and TensorCore at the same time. With this PR, TensorCore mode is disabled when dilation size > 1.


okuta:
Cloud you add test.

anaruse:
The branch was fixed, so that it can check TensorCore is not enabled when dilation size > 1.

okuta:
OK, I tested.
LGTM!

PR #4069, opened 2017-12-08
Simplify definitions of optimizers

toslunar:
After `UpdateRule` is introduced (#2137) in v2, one has to write a longer boilerplate to add a new optimizer.  The PR aims to reduce the cost.

Please remove the feature that `UpdateRule` accepts *updated* hyperparameters on *init*.  (To be more accurate, the PR doesn't change `UpdateRule` itself but changes the policy to use `UpdateRule`.)

Current:
- An optimizer-writer must write 11 `alpha`s, 11 `beta1`s, ... just for maintaining defaults.
- A user may do something like
```
    def create_update_rule(self):
        return AdamRule(self.hyperparam, eps=10 ** np.random.uniform(-5, -2))
```

Proposal:
- An optimizer-writer must write 5 `alpha`s, 5 `beta1`s, ... just for maintaining defaults.
- A user should write as below, if it's really needed.
```
rule = AdamRule(self.hyperparam)
rule.hyperparam.eps = ...
return rule
```

Once I get feedbacks, I'll add a commit to apply the same change to the other optimizers.


iwiwi:
+1

For example, during the preliminary experiments for my recent paper (1711.04325), I implemented and tried six optimizers in total. At that time, I was surprised to see that the amount of annoying boilerplate that is necessary to create new optimizers was substantially increased, compared to before (probably v1). I think many other researchers implement optimizers of their own frequently. I would appreciate if Chainer team keeps in mind that the flexibility of this aspect is also important for some people.

toslunar:
resolved conflicts: applied the change to AdamW (#4050)

toslunar:
I'm no longer confident that the PR has more pros than cons.  I heard an opinion that reducing boilerplate does not necessarily speed up development.

Let me withdraw the PR, while I suppose it's reasonable to add *new* optimizers in this style.

PR #4073, opened 2017-12-09
Add Local Convolution 2D connection layer

mihirparadkar:
This PR implements locally-connected layers (similar to convolution where each patch learns its own filter).

Thanks for taking a look!


bkvogel:
Thanks you for the PR. I am interested to understand the motivation for having a separate filter learned for each patch. It seems this could require a large amount of storage for the learned parameters. Do you find it more useful than existing convolution2d for some applications?

mihirparadkar:
In many cases, convolution2d is superior due to the reasons you outlined. However, sometimes learning a different filter for each patch is useful when there is some consistent structure to the images. One use case I've seen is that of facial recognition, when the eyes, nose, and mouth are in roughly the same spot for every image, so the filters at the eyes are different than those at the chin, for example. I believe they are also common in audio processing in the 1-dimensional special case.

As for the memory requirements, local convolutions can be more memory intensive than ordinary convolutions, but they may also require fewer output channels, resulting in possible space savings for the outputs. They are also usually less memory-intensive than a fully-connected layer with the same number of input and output neurons.

bkvogel:
Thank you for the response. I think it is interesting. There seem to be flake8 errors, though. Could you fix?

mihirparadkar:
I'm actually confused about what's actually failing in the Travis build. I see these in every log which suggest to me that flake8 and autopep8 are passing.

`The command "flake8" exited with 0.`
`The command "autopep8 -r . --global-config .pep8 --diff | tee check_autopep8" exited with 0.`

It's this command that's failing, but I don't know what `test` does.
`The command "test ! -s check_autopep8" exited with 1.`

bkvogel:
I think maybe if you run `autopep8 --in-place --global-config .pep8 ./tests/chainer_tests/links_tests/connection_tests/test_local_convolution_2d.py` you will see that it should auto-correct that file. You can then run `git diff` to see what it changed. Hopefully that will fix the error, but if not, please let me know.

bkvogel:
jenkins, test this please.

bkvogel:
It seems that our documentation tests failed. Could you add the new function and link to `docs/source/reference/{links,functions}.rst`?

bkvogel:
jenkins, test this please.

bkvogel:
oh, looks like there is still an error:

Document: reference/generated/chainer.functions.local_convolution_2d
11:05:42 --------------------------------------------------------------------
11:05:42 
11:05:42 Warning, treated as error:
11:05:42 WARNING: **********************************************************************
11:05:42 File "reference/generated/chainer.functions.local_convolution_2d.rst", line ?, in default
11:05:42 Failed example:
11:05:42     y.shape
11:05:42 Expected:
11:05:42     (2, 2, 5, 5)
11:05:42 Got:
11:05:42     (2, 5, 5, 2)

bkvogel:
jenkins, test this please.

bkvogel:
It seems the tests have passed. LGTM!

PR #4066, opened 2017-12-08
deepcopy to Link.copy

Crissman:
Fixing Issue #1953

PR #4075, opened 2017-12-10
Fix inconsistency on transferring copied link with uninitialized variables to GPU.

### LINKED ISSUE #4068: Link.to_gpu of Link.copy is inconsistent if there is an uninitialized Variable
For data-parallel training the following code works because `Link.to_gpu` makes in-place copies of parameters on a GPU.
```
    model_0 = L.Classifier(MLP())
    model_0.predictor(np.zeros((1, 784), dtype='f'))
    model_1 = model_0.copy()
    model_0.to_gpu(0)
    model_1.to_gpu(1)
```
The issue is that the second line of the code is required.

cf. The following code works, too.
```
    model_0 = L.Classifier(MLP())
    model_1 = copy.deepcopy(model_0)
    model_0.to_gpu(0)
    model_1.to_gpu(1)
```

To reproduce the issue:
```
import chainer
import chainer.links as L
import cupy
l0 = L.Linear(None, 10)
l1 = l0.copy()
l0.to_gpu(0)
l1.to_gpu(1)
print(l1.W.data)  # =>  None
print(l1.b.data.device)  # =>  <CUDA Device 1>
l0(cupy.random.randn(3, 4).astype('f'))
print(l1.W.data.device)  # =>  <CUDA Device 0>  (!!)
print(l1.b.data.device)  # =>  <CUDA Device 1>
print(l0.W.data is l1.W.data)  # =>  True  (!!)
```

Related: #1953

takagi:
Fix #4068.

This PR fixes inconsistency when a link that has uninitialized variables is `link.copy()`ed, `link.to_gpu()`ed then initialized.

### DIFF
--- a/chainer/variable.py
+++ b/chainer/variable.py
@@ -709,7 +709,7 @@ Actual: {0}'''.format(type(data))
         if self.data is None:
             self._initial_device = (cuda.Device().id
                                     if device is None else device)
-            self._data = [None]
+            self._data = [None]  # Renew placeholder to break sharing
         else:
             self._data = [cuda.to_gpu(self.data, device)]
             if self._grad_var is not None:
--- a/chainer/variable.py
+++ b/chainer/variable.py
@@ -709,6 +709,7 @@ Actual: {0}'''.format(type(data))
         if self.data is None:
             self._initial_device = (cuda.Device().id
                                     if device is None else device)
+            self._data = [None]
         else:
             self._data = [cuda.to_gpu(self.data, device)]
             if self._grad_var is not None:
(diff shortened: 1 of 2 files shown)


delta2323:
Could you add an unit test that checks the problem reported in #4066 is also resolved?

takagi:
I added another unit test that checks links copied to different GPUs.

takagi:
@beam2d Would you review this? Another PR #4066 is dependent on this to merge.

chainer/variable.py
```diff
@@ -709,6 +709,7 @@ def to_gpu(self, device=None):
         if self.data is None:
             self._initial_device = (cuda.Device().id
                                     if device is None else device)
+            self._data = [None]
```
beam2d: Could you add a short comment on this line, like `# renew the placeholder`? I could not understand this line at first sight.

takagi: Okay, and I will add a thin object that represents initialized/uninitialized data, for which I use list, as another PR.

takagi: Added a comment!

beam2d:
LGTM

PR #4078, opened 2017-12-11
Fix test failure with cuDNN v6

kmaehashi:
It seems that cuDNN v6 does not support backward computation of dilated convolution when deterministic algorithm (`CUDNN_CONVOLUTION_BWD_DATA_ALGO_1`) is selected.


kmaehashi:
Jenkins test this please

takagi:
LGTM!

PR #3351, opened 2017-09-07
Add early stopping

himkt:
This PR is related to https://github.com/chainer/chainer/pull/3314

I add the feature Early-Stopping by implementing EarlyStoppingTrigger.
It can see the example of using early stopping in this [example](https://github.com/chainer/chainer/pull/3350/files#diff-741ac22e5ba54630e03d10325cc8c89e)

- [x] check guideline
- [x] add test
- [x] add document

PR #4079, opened 2017-12-11
Modify detect training length

himkt:
Related to #3351 
I have tried to change the way to detect `training_length`.

### DIFF
--- a/chainer/training/extensions/progress_bar.py
+++ b/chainer/training/extensions/progress_bar.py
@@ -6,7 +6,6 @@ import time
 
 from chainer.training import extension
 from chainer.training.extensions import util
-from chainer.training import triggers
 
 
 class ProgressBar(extension.Extension):
--- a/chainer/training/extensions/progress_bar.py
+++ b/chainer/training/extensions/progress_bar.py
@@ -43,10 +43,7 @@ class ProgressBar(extension.Extension):
         # initialize some attributes at the first call
         if training_length is None:
             t = trainer.stop_trigger
-            if not hasattr(t, '_get_training_length'):
-                raise TypeError(
-                    'cannot retrieve the training length from %s' % type(t))
-            training_length = t._get_training_length()
+            training_length = t.get_training_length()
 
         stat_template = self._status_template
         if stat_template is None:
--- a/chainer/training/extensions/progress_bar.py
+++ b/chainer/training/extensions/progress_bar.py
@@ -44,7 +44,7 @@ class ProgressBar(extension.Extension):
         # initialize some attributes at the first call
         if training_length is None:
             t = trainer.stop_trigger
-            if not isinstance(t, triggers.IntervalTrigger):
+            if not (hasattr(t, 'period') and hasattr(t, 'unit')):
                 raise TypeError(
                     'cannot retrieve the training length from %s' % type(t))
             training_length = self._training_length = t.period, t.unit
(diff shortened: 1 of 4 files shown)


himkt:
I have changed the way to detect `training_length` in InternalTrigger and EarlyStoppingTrigger.
BTW, I found that there are errors in testing `test_manual_schedule_trigger.py` and `test_interval_trigger.py`, which occurs `master` branch.

himkt:
I removed the WIP prefix because all checks were passed.
But in my developing environment, errors occurred in testing test_manual_schedule_trigger.py and test_interval_trigger.py, which also occurs master branch like below.
`AttributeError: module 'chainer.testing' has no attribute 'condition'`

What do I mistake or forget something which I have to do?

himkt:
The test failures will be fixed in #4103 .

himkt:
After #4103 was merged, I confirmed that tests for trigger and extension were passed.

chainer/training/extensions/progress_bar.py
```diff
@@ -44,10 +43,10 @@ def __call__(self, trainer):
         # initialize some attributes at the first call
         if training_length is None:
             t = trainer.stop_trigger
-            if not isinstance(t, triggers.IntervalTrigger):
+            if not hasattr(t, '_get_training_length'):
```
beam2d: This method is used as an interface of trigger, so I think it's better not to prefix it by `_`. How about using `get_training_length`?

chainer/training/triggers/early_stopping_trigger.py
```diff
@@ -45,6 +45,7 @@ def __init__(self, check_trigger=(1, 'epoch'), monitor='main/loss',
         self.monitor = monitor
         self.verbose = verbose
         self.already_warning = False
+        self.max_trigger = max_trigger  # TODO(This is confusing)
```
beam2d: How about making `max_trigger` argument accepting any trigger that supports `get_training_length`? Then, we do not need to add `self.max_trigger` and then can make `get_training_length` just delegates to `self._max_trigger.get_training_length()`.

chainer/training/extensions/progress_bar.py
beam2d: I also feel it's more natural to let duck-typing check the code here. I mean just removing this if-then-raise block. What do you think? It will show the trigger type and the cause of the error (`'FooTrigger' object has no attribute 'get_training_length'`), which are more informative for users.

himkt: > This method is used as an interface of trigger
I understand! I remove the prefix.

himkt: > I also feel it's more natural to let duck-typing check the code here.

Thanks for the comment! I agree with your suggestion.
It would be better to show informations

chainer/training/triggers/early_stopping_trigger.py
himkt: Thanks for the great suggestion! I remove `self.max_trigger` and use `self._max_trigger.get_training_length`.

himkt:
Thanks for your kindly comments.
I update some implementation based on your suggestion.

beam2d:
Jenkins, test this please

beam2d:
LGTM

beam2d:
Thank you!

PR #4085, opened 2017-12-12
avoid using np.stack which is introduced in NumPy 1.10

### LINKED ISSUE #4081: Grouped convolution/deconvolution does not work in CPU mode with NumPy 1.9
We should not use `xp.stack` which is introduced in NumPy 1.10.

kmaehashi:
Fixes #4081
Refs #4003


kmaehashi:
For record, this issue was introduced in #3494 (no need to backport)

niboshi:
LGTM!

PR #4021, opened 2017-12-04
[backport] backends subpackage

kmaehashi:
Backport of #3974

PR #4091, opened 2017-12-13
fix test_init_docstring to use importlib to find package

kmaehashi:
This PR removes the (indirect) use of deprecated `imp` module.
https://docs.python.org/3/library/imp.html

I confirmed that this PR resolves the issue in https://github.com/chainer/chainer/pull/4021. (surprisingly)


hvy:
LGTM!

PR #4093, opened 2017-12-13
restructure linear's backward to reduce transpose times and make it m…

jzhoulon:
…ore friendly to do optimization(also aligned with convolution, relu's implementation)
unit test result: ALL PASSED
Currently, linear backward is just use forward's code under transpose, if separate them with each other, then more optimization can be done, like other layers(convolution, relu)...

### DIFF
--- a/chainer/functions/connection/linear.py
+++ b/chainer/functions/connection/linear.py
@@ -57,15 +57,90 @@ class LinearFunction(function_node.FunctionNode):
 
         ret = []
         if 0 in indexes:
-            gx = linear(gy, W.T)
+            gx, = LinearGradData(self).apply((gy, W))
             ret.append(chainer.functions.cast(gx, x.dtype))
         if 1 in indexes:
-            gW = linear(gy.T, x.T)
+            gW, = LinearGradWeight(self).apply((x, gy))
             ret.append(chainer.functions.cast(gW, W.dtype))
         if 2 in indexes:
             gb = chainer.functions.sum(gy, axis=0)
             ret.append(gb)
 
+        return ret
+
+
+class LinearGradData(function_node.FunctionNode):
+
+    def __init__(self, linear):
+        return
+
+    def forward(self, inputs):
+        self.retain_inputs((0, 1))
+        gy, W = inputs
+
+        if not type_check.same_types(*inputs):
+            raise ValueError('numpy and cupy must not be used together\n'
+                             'type(gy): {0}, type(W): {1}'
+                             .format(type(gy), type(W)))
+        if (isinstance(gy, numpy.ndarray) and
+                not (gy.flags.c_contiguous or gy.flags.f_contiguous) and
+                1 in gy.shape):
+            gy = numpy.ascontiguousarray(gy)
+
+        gx = gy.dot(W).astype(gy.dtype, copy=False)
+        self.retain_inputs((0, 1))
+        return gx,
+
+    def backward(self, indexes, grad_outputs):
+        gy, W = self.get_retained_inputs()
+        ggx, = grad_outputs
+
+        ret = []
+        if 0 in indexes:
+            ggy = linear(ggx, W)
+            ret.append(chainer.functions.cast(ggy, gy.dtype))
+        if 1 in indexes:
+            gw, = LinearGradWeight(self).apply((ggx, gy))
+            ret.append(chainer.functions.cast(gw, W.dtype))
+        if 2 in indexes:
+            ggb = chainer.functions.sum(ggx, axis=0)
+            ret.append(ggb)
+        return ret
+
+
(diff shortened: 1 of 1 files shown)


kmaehashi:
Thank you for the pull-request.
Just to confirm, is this PR intends to make iDeep integration go smooth?

jzhoulon:
yes, it is to make ideep more comfortable(do more optimization). Thanks very much

chainer/functions/connection/linear.py
```diff
@@ -69,6 +69,81 @@ def backward(self, indexes, grad_outputs):
...
+class LinearGradData(function_node.FunctionNode):
+
+    def __init__(self, linear):
+        return
+
+    def forward(self, inputs):
+        self.retain_inputs((0, 1))
+        gy, W = inputs
+
+        if not type_check.same_types(*inputs):
```
kmaehashi: We now implemented this check in FunctionNode-level (see https://github.com/chainer/chainer/pull/4029), so you don't have to check this.
(same for `LinearGradWeight`)

chainer/functions/connection/linear.py
```diff
@@ -57,10 +57,10 @@ def backward(self, indexes, grad_outputs):
 
         ret = []
         if 0 in indexes:
-            gx = linear(gy, W.T)
+            gx, = LinearGradData(self).apply((gy, W))
```
kmaehashi: I think it is better to align the argument order with `LinearGradWeight`.
(i.e., `apply((W, gy))` instead of `apply((gy, W))`)

chainer/functions/connection/linear.py
```diff
@@ -69,6 +69,81 @@ def backward(self, indexes, grad_outputs):
         return ret
 
 
+class LinearGradData(function_node.FunctionNode):
+
+    def __init__(self, linear):
```
kmaehashi: Please remove the constructor if it is unused.

chainer/functions/connection/linear.py
```diff
@@ -69,6 +69,81 @@ def backward(self, indexes, grad_outputs):
...
+            raise ValueError('numpy and cupy must not be used together\n'
+                             'type(gy): {0}, type(W): {1}'
+                             .format(type(gy), type(W)))
+        if (isinstance(gy, numpy.ndarray) and
+                not (gy.flags.c_contiguous or gy.flags.f_contiguous) and
+                1 in gy.shape):
+            gy = numpy.ascontiguousarray(gy)
+
+        gx = gy.dot(W).astype(gy.dtype, copy=False)
+        self.retain_inputs((0, 1))
```
kmaehashi: Inputs are already retained above.
(same for `LinearGradWeight`)

chainer/functions/connection/linear.py
```diff
@@ -69,6 +69,81 @@ def backward(self, indexes, grad_outputs):
...
+        ggx, = grad_outputs
+
+        ret = []
+        if 0 in indexes:
+            ggy = linear(ggx, W)
+            ret.append(chainer.functions.cast(ggy, gy.dtype))
+        if 1 in indexes:
+            gw, = LinearGradWeight(self).apply((ggx, gy))
+            ret.append(chainer.functions.cast(gw, W.dtype))
+        if 2 in indexes:
```
kmaehashi: Is this block by mistake? (there are only 2 inputs for LinearGradData)

kmaehashi:
Jenkins test this please

chainer/functions/connection/linear.py
jzhoulon: ok, I will modify it.

chainer/functions/connection/linear.py
jzhoulon: I will remove it, thanks

chainer/functions/connection/linear.py
jzhoulon: I will remove it, thanks

chainer/functions/connection/linear.py
jzhoulon: I will remove it, thanks

chainer/functions/connection/linear.py
jzhoulon: yes, gb is not need in  grad data double grad, I will remove it, thanks

kmaehashi:
Jenkins test this please

jzhoulon:
Hi, kmaehashi, is there any log shows why Jenkins build failed or how can I reproduce it? In my local test, it seems pre-ci are all passed..Thanks very much

kmaehashi:
Jenkins test this please

kmaehashi:
@jzhoulon Hi, thanks for the update.  Maybe this error was caused by failure of our Jenkins server. I'll add comment if further action from you is necessary.

kmaehashi:
LGTM!

PR #4099, opened 2017-12-14
v3: fix cuda import to use chainer.backends

kmaehashi:
https://github.com/chainer/chainer/pull/4021#issuecomment-351321748


hvy:
LGTM!

Issue #2745, opened 2017-05-17
Unify convolution/pooling functions for all spatial dimensionalities


beam2d:
I sometimes saw users that do not know the existence of `ConvolutionND` and `*_pooling_nd`. In particular, they say "there is no Convolution1D" (or 3D), although it is supported by the ND version. There are two approaches to resolve this issue.

1. Make explicit versions for typical dimensionalities (Convolution1D, Convolution3D, ...)
2. Unify them and use it everywhere (Convolution, max_pooling, ...)

I feel the second approach is much simpler. As for the Functions, we only have to make an alias of `*_nd` without the suffix `_nd`, e.g. `convolution`, `max_pooling`, etc. As for the Links, we have to think about how to let it know the dimensionality of the parameters. Thanks to the lazy initialization of the parameters, it can be automatically inferred in most cases, so that we can make `Convolution` that can be constructed by a code like `Convolution(out_size, kernel_size)`. We can also explicitly specify the spatial dimensionality by `ndim` option.

Another interesting aspect of this approach is that, once we fix the existing code to use this unified version, the definition of a convnet designed for a specific spatial dimensionality (e.g., 2D) is directly reusable for other spatial dimensionalities (e.g., 1D and 3D), because there is no information about the dimensionality appearing in the model definitions.

mitmul:
I vote for the 2nd option, to make a unified interface.

delta2323:
I like the second option.

beam2d:
Is anyone interested in working on it?

rezoo:
I'm interested in, and I'm now unifying `ConvolutionND` to support double backpropagation like `Convolution2D`.

Issue #3986, opened 2017-11-29
Need to adjust tolerance


kmaehashi:
Following tests are occasionally failing, maybe because the tolerance is too sensitive.

- [ ] tests.chainer_tests.functions_tests.loss_tests.test_contrastive /TestContrastive_param_21/test_forward_cpu/
- [ ] tests.chainer_tests.functions_tests.loss_tests.test_contrastive /TestContrastive_param_23/test_backward_gpu_no_cudnn/
- [ ] tests.chainer_tests.functions_tests.pooling_tests.test_max_pooling_2d /TestMaxPooling2D_param_1/test_double_backward_gpu_no_cudnn/
- [ ] tests.chainer_tests.links_tests.connection_tests.test_deconvolution_2d /TestDeconvolution2DParameterShapePlaceholder_param_10/test_backward_gpu/
- [ ] tests.chainer_tests.test_gradient_check.TestCheckDoubleBackward.test_double_backward_with_params_gpu
- [ ] tests.chainer_tests.functions_tests.pooling_tests.test_unpooling_nd.TestUnpoolingND_param_3.test_double_backward_cpu
- [ ] tests.chainer_tests.functions_tests.pooling_tests.test_average_pooling_nd.TestAveragePoolingND_param_1.test_double_backward_gpu_no_cudnn

Note: this PR is different from #3766 in the sense that the error (difference) is not "too large".

niboshi:
I think `batch_normalization` and `batch_renormalization` have some problem other than numerical tolerance. #3766

kmaehashi:
Thank you, I removed them from the list.

hvy:
I just noticed a quite large error for `TestAverage_param_26/test_backward_cpu/` https://github.com/chainer/chainer/issues/3766#issuecomment-347757626

hvy:
@kmaehashi We can remove `.../TestAverage_param_26/test_backward_cpu/` from this list as well since it is appended to #3766!

kmaehashi:
I added:
-  tests.chainer_tests.functions_tests.pooling_tests.test_max_pooling_2d/TestMaxPooling2D_param_1/test_double_backward_gpu_no_cudnn/
-  tests.chainer_tests.links_tests.connection_tests.test_deconvolution_2d/TestDeconvolution2DParameterShapePlaceholder_param_10/test_backward_gpu/

kmaehashi:
Added:
[...]

PR #4110, opened 2017-12-15
New style ConvolutionND/DeconvolutionND

rezoo:
Related to #2745. This challenging PR aims to update ConvolutionND/DeconvolutionND functions to support double back propagations.
(I finally want to integrate `Convolution2D` and `ConvolutionND`, and derive `Convolution2D` as a special case of `ConvolutionND`, however, in the first stage this PR aims to rewrite `ConvolutionND` with reference to `Convolution2D`, and support double back propagations.)

### DIFF
--- a/chainer/functions/connection/convolution_nd.py
+++ b/chainer/functions/connection/convolution_nd.py
@@ -5,7 +5,8 @@ from six import moves
 import chainer
 from chainer.backends import cuda
 from chainer import configuration
-from chainer import function
+from chainer import function_node
+import chainer.functions
 from chainer.functions.connection import convolution_2d
 from chainer.utils import conv
 from chainer.utils import conv_nd
@@ -23,7 +24,7 @@ if cuda.cudnn_enabled:
         libcudnn.CUDNN_CONVOLUTION_BWD_DATA_SPECIFY_WORKSPACE_LIMIT
 
 
-class ConvolutionND(function.Function):
+class ConvolutionND(function_node.FunctionNode):
 
     def __init__(self, ndim, stride=1, pad=0, cover_all=False):
         self.ndim = ndim
@@ -66,15 +67,15 @@ class ConvolutionND(function.Function):
 
         # Make patch array.
         if xp is numpy:
-            self.col = conv_nd.im2col_nd_cpu(
+            col = conv_nd.im2col_nd_cpu(
                 x, ksize, stride, pad, cover_all=self.cover_all)
         else:
-            self.col = conv_nd.im2col_nd_gpu(
+            col = conv_nd.im2col_nd_gpu(
                 x, ksize, stride, pad, cover_all=self.cover_all)
 
         # Compute correlation.
         axes = tuple(moves.range(1, ndim + 2))  # (1, 2, ..., N+1)
-        y = xp.tensordot(self.col, W, (axes, axes)).astype(x.dtype, copy=False)
+        y = xp.tensordot(col, W, (axes, axes)).astype(x.dtype, copy=False)
 
         # Apply bias if given.
         if b is not None:
@@ -152,6 +153,7 @@ class ConvolutionND(function.Function):
         return y,
 
     def forward(self, inputs):
+        self.retain_inputs((0, 1))  # retain only x and W
         x, W = inputs[:2]
         b = inputs[2] if len(inputs) == 3 else None
 
@@ -173,17 +175,64 @@ class ConvolutionND(function.Function):
         else:
             return self._forward_cudnn(x, W, b)
 
-    def _backward_xp(self, x, W, b, gy, xp):
-        dims = x.shape[2:]     # (n, c_I, d_1, d_2, ..., d_N)
-        stride = self.stride
-        pad = self.pad
-        ndim = self.ndim
+    def backward(self, indexes, grad_outputs):
+        x, W = self.get_retained_inputs()
+        gy, = grad_outputs
 
+        ret = []
+        if 0 in indexes:
+            x_shape = x.shape[2:]
+            gx = chainer.functions.deconvolution_nd(
+                gy, W, stride=self.stride, pad=self.pad, outsize=x_shape)
+            ret.append(gx)
+        if 1 in indexes:
+            gW, = ConvolutionNDGradW(self).apply((x, gy))
+            ret.append(gW)
+        if 2 in indexes:
+            axis = (0,) + tuple(moves.range(2, gy.ndim))
+            gb = chainer.functions.sum(gy, axis=axis)
+            ret.append(gb)
+
+        return ret
+
+
+class ConvolutionNDGradW(function_node.FunctionNode):
+
+    def __init__(self, convnd):
+        W_node = convnd.inputs[1]
+        self.ndim = convnd.ndim
+        self.ksize = W_node.shape[2:]
+        self.stride = convnd.stride
+        self.pad = convnd.pad
+        self.cover_all = convnd.cover_all
+        self.W_dtype = W_node.dtype
+
+    def _use_cudnn(self, x, gy):
+        return (
+            chainer.should_use_cudnn('>=auto')
+            and not self.cover_all
+            and x.dtype == self.W_dtype
+            and gy.dtype == self.W_dtype
+            and self.ndim > 1)
+
+    def forward(self, inputs):
+        self.retain_inputs((0, 1))
+        x, gy = inputs
+
+        xp = cuda.get_array_module(*inputs)
+        if xp is numpy:
+            return self._forward_xp(x, gy, numpy)
+        elif not self._use_cudnn(x, gy):
+            return self._forward_xp(x, gy, cuda.cupy)
+        else:
+            return self._forward_cudnn(x, gy)
+
+    def _forward_xp(self, x, gy, xp):
         # Compute filter weight gradient.
         # (n, _, out_1, out_2, ..., out_N)
-        out_axes = (0,) + tuple(moves.range(2, ndim + 2))
+        out_axes = (0,) + tuple(moves.range(2, self.ndim + 2))
         # (n, _, _, ..., _, out_1, out_2, ..., out_N)
-        col_axes = (0,) + tuple(moves.range(ndim + 2, ndim * 2 + 2))
+        col_axes = (0,) + tuple(moves.range(self.ndim + 2, self.ndim * 2 + 2))
 
         # NumPy raises an error when the array is not contiguous.
         # See: https://github.com/chainer/chainer/issues/2744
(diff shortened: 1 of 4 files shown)


rezoo:
I finished implementation. After merging this PR I would like to unify Convolution2D and ConvolutionND.
(It includes implementation of dilated convolution and grouped convolution in ConvolutionND)

rezoo:
@hvy Hi, is there any progress?

okuta:
I will check this.

hvy:
Sorry for the late response and thank you @okuta san for the comments.

chainer/functions/connection/convolution_nd.py
```diff
@@ -5,7 +5,8 @@
 import chainer
 from chainer.backends import cuda
 from chainer import configuration
-from chainer import function
+from chainer import function_node
+import chainer.functions
```
hvy: We can probably skip this import since `chainer` is already imported.

chainer/functions/connection/convolution_nd.py
```diff
@@ -173,17 +175,64 @@ def forward(self, inputs):
...
+            axis = (0,) + tuple(moves.range(2, gy.ndim))
+            gb = chainer.functions.sum(gy, axis=axis)
+            ret.append(gb)
+
+        return ret
+
+
+class ConvolutionNDGradW(function_node.FunctionNode):
+
+    def __init__(self, convnd):
```
hvy: How about passing each argument separately? I think its interface is better aligned with `ConvolutionND.__init__` that way.

tests/chainer_tests/functions_tests/connection_tests/test_convolution_nd.py
```diff
@@ -187,6 +195,91 @@ def test_backward_gpu_im2col_nobias(self):
...
+                                 cover_all=self.cover_all)
+            return y * y  # make the function nonlinear
+
+        with chainer.using_config('use_cudnn', use_cudnn):
+            with chainer.using_config('autotune', self.autotune):
+                gradient_check.check_double_backward(
+                    f, args, y_grad, grad_grads,
+                    dtype='d', atol=5e-3, rtol=5e-2)
+
+    @condition.retry(3)
```
hvy: Could you make the tests pass without `retry` by adjusting the tolerances? In general, I think we want to avoid relying on `retry` since it may hide bugs or numerical instabilities.

tests/chainer_tests/functions_tests/connection_tests/test_convolution_nd.py
```diff
@@ -187,6 +195,91 @@ def test_backward_gpu_im2col_nobias(self):
...
+    def test_double_backward_cpu(self):
+        self.check_double_backward(self.x, self.W, self.b, self.gy,
+                                   self.ggx, self.ggW, self.ggb,
+                                   use_cudnn='always')
+
+    @condition.retry(3)
```
hvy: Ditto for other occurrences for `retry`.

tests/chainer_tests/functions_tests/connection_tests/test_convolution_nd.py
```diff
@@ -187,6 +195,91 @@ def test_backward_gpu_im2col_nobias(self):
...
+
+                ggb = xp.empty((len(b_data) * 2,), dtype=self.b.dtype)
+                ggb[::2] = b_grad_grad
+                b_grad_grad = ggb[::2]
+                self.assertFalse(b_grad_grad.flags.c_contiguous)
+
+        args = (x_data, W_data)
+        grad_grads = (x_grad_grad, W_grad_grad)
+        if b_data is not None:
+            args = args + (b_data,)
```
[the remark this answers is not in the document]
rezoo: Fixed.

tests/chainer_tests/functions_tests/connection_tests/test_convolution_nd.py
```diff
@@ -187,6 +195,91 @@ def test_backward_gpu_im2col_nobias(self):
...
+            grad_grads = grad_grads + (b_grad_grad,)
```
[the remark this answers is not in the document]
rezoo: Fixed.

tests/chainer_tests/functions_tests/connection_tests/test_deconvolution_nd.py
```diff
@@ -192,6 +200,80 @@ def test_backward_gpu(self):
...
+
+                ggb = xp.empty((len(b_data) * 2,), dtype=b_grad_grad.dtype)
+                ggb[::2] = b_grad_grad
+                b_grad_grad = ggb[::2]
+                assert not b_grad_grad.flags.c_contiguous
+
+        args = (x_data, W_data)
+        grad_grads = (x_grad_grad, W_grad_grad)
+        if b_data is not None:
+            args = args + (b_data,)
```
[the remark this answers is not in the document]
rezoo: Fixed.

tests/chainer_tests/functions_tests/connection_tests/test_deconvolution_nd.py
```diff
@@ -192,6 +200,80 @@ def test_backward_gpu(self):
...
+            grad_grads = grad_grads + (b_grad_grad,)
```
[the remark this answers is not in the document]
rezoo: Fixed.

chainer/functions/connection/convolution_nd.py
rezoo: Fixed.

chainer/functions/connection/convolution_nd.py
rezoo: This is because I simply followed the design of API in `Convolution2D`. I think either is fine because it is an internal design problem that does not need to be released (it is necessary to adopt to either, though).

tests/chainer_tests/functions_tests/connection_tests/test_convolution_nd.py
rezoo: Is there any discussions on introduction of `condition.retry` decorator in GitHub? I think it is useful and important in the sense that is avoids the problem that the test fails stochastically.

tests/chainer_tests/functions_tests/connection_tests/test_convolution_nd.py
rezoo: See above.

rezoo:
Thank you, @okuta and @hvy! I've fixed several problems and commented.

tests/chainer_tests/functions_tests/connection_tests/test_convolution_nd.py
hvy: Not as I am aware of unfortunately. Do you know how frequently this test fails for a reasonable atol, rtol? It would make sense if a function contains non-differentiable points but else, we shouldn't need it (slightly off topic but it's addressed in this PR https://github.com/chainer/chainer/pull/3499)? But again, I see it's aligned with the `Convolution2D` tests, so it's fine I think given low/small margin failures.

chainer/functions/connection/convolution_nd.py
hvy: Ah, I see, then I agree we should leave it this way. It's not really labelled as a non-public API so I personally think we should think it through but it should of course be consistent.

chainer/functions/connection/convolution_nd.py
hvy: Thanks a lot!

hvy:
Thank you for the quick responses. I added a comment regarding the retry decorator!

tests/chainer_tests/functions_tests/connection_tests/test_convolution_nd.py
rezoo: In my environment it fails with the probability of 10%. Of course it can be solved by carefully adjusting the parameters of atol and rtol, however, I think relaxation of such parameters is essentially equivalent to the use of `condition.retry`. At least the tests of `Convolution2D` should also be changed for consistency if we want to remove `condition.retry` decorators.

rezoo:
I added a comment.

tests/chainer_tests/functions_tests/connection_tests/test_convolution_nd.py
hvy: They're not really equivalent in the sense that a function containing non-differentiable points may be detected by adjusting the tolerances but are hidden with retries. 

But I see that e.g. Convolution2D tests have the same problem (also see https://github.com/chainer/chainer/issues/3986). I am currently running the Convolution2D tests with 1000 repetitions per fixture. So far ~80% done and they've all passed. I discussed with @niboshi about the failure rate and with this in mind the rate seems a bit high. Is there anything we can do about the current failure rate?

Sorry for the really long reply!!

rezoo: > They're not really equivalent in the sense that a function containing non-differentiable points may be detected by adjusting the tolerances but are hidden with retries.
> Is there anything we can do about the current failure rate?
? Sorry, but I'm not sure why you regard this problem as important. Do you want to detect a non-differentiable point that `Convolution2D` has?
Honestly speaking, I'm not sure why it is so important, but I'm OK to fix this issue if you consider it is so important, but I would be glad if you would tell me the best parameters of `rtol` and `atol`.

hvy:
jenkins, test this please.

hvy:
For the record, we agreed to use the `retry` decorator, partially for consistency with Convolution2D. 

However, the more general issue remain for how to write backward tests, e.g. what tolerances and the number of retries to accept, but this will be addressed in a later discussion.

niboshi:
I'm sorry, I didn't follow the discussion.
I think it's worthwhile to mention that `retry` has a problem that it can hide implementation glitches that can happen depending on the inputs. Such problem was occurred in the past (I forgot where it was).
It's ok if it will be removed eventually.

hvy:
@niboshi Thanks for the input. Basically, we should come up with more specific guidelines for how these functions should be tested. Once that's done, we can go through all existing functions including this one. I'll try merging this PR tomorrow if that is okay with you.

hvy:
@rezoo Sorry for the late response. Putting the previous discussions about tolerances aside, I tried both functions locally and I indeed think ConvolutionND is fine to merge since it is very stable.

However, I noticed that DeconvolutionND fails often (even without repeats) and the errors are quite large (too large to be fixed by tolerance adjustments). Could you please confirm this?

rezoo:
Ok. Can you tell me the detail of the condition for help?

rezoo:
I changed the condition of DeconvolutionND with that of ConvolutionND, and confirmed that there does not exist any errors. Could you confirm?

hvy:
Thanks for the update. I'll check them!

hvy:
jenkins, test this please.

hvy:
Again, thank you for the fixed. I verified it on CI and also locally. LGTM!

PR #4087, opened 2017-12-12
avoid using xp.stack in examples

kmaehashi:
Fixes https://github.com/chainer/chainer/pull/4003#issuecomment-348389484

PR #4117, opened 2017-12-18
fix seq2seq example not working with GPU

kmaehashi:
As @lboloni pointed out, it does not work in GPU mode.

https://github.com/chainer/chainer/pull/4087#commitcomment-26328157


hvy:
LGTM! Sorry, this should not have gotten away in the first place.

PR #4121, opened 2017-12-19
Change the size of out1 in inc4c and inc4d 

takaaki82:
I defined the size of out1 in inc4c and inc4d in the same way as CNTK.


mitmul:
Jenkins, test this please

mitmul:
@takaaki82 Thank you for sending this PR! I confirmed that the change introduced by this PR makes sense because it corrects the number of channels of the current code to match it with the original definition of Inception-V2 in Caffe: https://github.com/pertusa/InceptionBN-21K-for-Caffe/blob/master/deploy.prototxt
So, LGTM!

PR #4133, opened 2017-12-21
Fix document: chainer.backends.cuda

toslunar:
`chainer.cuda` is available for backward compatibility.
Use `chainer.backends.cuda` (#3974) in documents.


kmaehashi:
Thanks for the PR.
The documentation discussion related to backend refactoring is ongoing in https://github.com/chainer/chainer/pull/3981.

PR #4135, opened 2017-12-21
Update VAE example

mitmul:
This PR updates VAE example to reflect the addition of the updaters module.


kmaehashi:
This has been fixed in #4208.

PR #4136, opened 2017-12-22
Accumulate the gradient to a tuple for lazy add operation

LuoYuanke:
The accumulated add is to fuse the add operations to improve the performance of backward. Take the below piece of graph for example. The x variable is input for 3 functions, so the gx should be accumulated on the 3 functions backward. Currently implementation add the current gx to the new gx on each functions backward. However on googlenet/googlenet v2, there are many forked functions for one input, so the fuse for accumulated add is meaningful. 
                                     __________X___________
                                     |               |                |
                                 Func1      Func2     Func3 
This patch is platform independent. It should both benefit for GPU and CPU devices. The basic idea for patch accumulate the gx to a tuple instead of doing add operation when the new gx need to be added to current gx. When the gx is inputted as gy for the next layer, the actual fused add should be done on the tuple of gx. We can call the approach as lazy add. Since there are 2 backward functions in the framework, the lazy add is applied both in _backward_main() and _back_prop(). We implemented a function node (AccumulateAdd) to calculate the sum of a tuple of array. Fuse add optimization can be done in this function node, and we will submit that patch in another PR.


niboshi:
Thank you for a PR!
Do you have any benchmarking?
I tested on convnet-benchmarks (https://github.com/soumith/convnet-benchmarks), but yet to see an improvement on googlenet benchmark.

Also, I'm curious how this PR helps ideep integration.

LuoYuanke:
We got 4% performance improvement for googlenet on skx8180. However in this PR, it is just general approach to accumulate the element for sum in a tuple. In this PR we can NOT see any performance improvement because we have not implemented optimized sum operation. We will implement it in next PR. 
The optimized sum is to call "ideepy.acc_add(xs)" to the sum. The "ideepy.acc_add(xs)" call mkldnn sum interface to do sum which is optimized primitive.
We have sum primitive in mkldnn library which accept several arrays and sum them together. On CPU side, fuse several sum into one need less instruction and is cache friendly. Take "o = a + b + c" as example, "o = a + b; o = o + c" need 
 4 load instructions for a, b, o, c and 2 store instruction for o. Fuse "o = a + b + c" only need 3 load instruction for a, b, c and one store instruction for o.

The final forward code of acc_sum will looks like as below.

     def forward(self, xs):
         self.len = len(xs)
         use_ideep = (chainer.ia.check_ideep_enabled()) and \
             (xs[0].ndim == 2 or xs[0].ndim == 4)
         if use_ideep:
             for x in xs:
                 if not isinstance(x, chainer.ia.mdarray):
                     use_ideep = False
                     break

         if use_ideep:
             # optimized path
             y = chainer.ia.acc_add(xs)
         else:
             # general path
             y = xs[0] + xs[1]
             for x in xs[2:]:
                 y += x
         return utils.force_array(y),

niboshi:
Thank you, I understand the background.

We have some concerns and suggestions as the following:

1. This PR may increase the intermediate memory consumption (although I have not observed such degradation in googlenet test). So this feature may be better implemented in an optional feature.

* How about introducing a `chainer.config` entry and disable it by default?
* This flag is to be checked in `FunctionNode.apply()` and kept in `FunctionNode` as an attribute. In backward operations, lazy-accumulation will be done only if this flag is on.
* Also how about make it experimental? (by using `utils.experimental()`)

2. We allow `FunctionNode.backward_accumulate()` to be overridden by user. In such case, if its `grad_inputs` includes lazily-accumulated gradients, the overridden function can break. In order to avoid such situation, how about the following?

* Introduce an availability flag as a `FunctionNode`'s attribute, which is off by default.
* Before calling `FunctionNode.backward_accumulate()`, if (1) this flag is off, AND (2) `backward_accumulate` is overridden, ensure gradients are accumulated. (this flag is meant to be set to on by user who implemented overridden `backward_accumulate` which supports lazily-accumulated gradients).
* You can test if a method is overridden by `FunctionNode.__dict__['backward_accumulate'] is not type(self).__dict__['backward_accumulate']`.

Please ask me if you have any questions.

LuoYuanke:
Thanks for the good suggestion. We will revise our patch.

LuoYuanke:
Hi Niboshi

I add the "lazy_grad_sum" in configuration. Is it equal to check the config in backward instead of check in an attribute in FunctionNode? The gradient sum on occurs in backward, so check it in backward not to affect inference code.

For override of FunctionNode.backward_accumulate(), lazy grad sum doesn't expected the return value of FunctionNode.backward_accumulate() be a tuple. It can be an array. So when FunctionNode.backward_accumulate() is overridden, it doesn't break the lazy grad sum, but make lazy grad sum not work as well as expected.

Thanks

niboshi:
@LuoYuanke 

The point of storing the configuration in `FunctionNode` at forward computation is that, users do not have control of the configuration variable at the moment of function's backward() call. They can only configure different values for different functions at forward computation.

As for `FunctionNode.backward_accumulate()`, the problem is that overridden `backward_accumulate()` expects tuple of `Variable`s, not tuple of tuples. So we need to accumulate the gradients *before* calling overridden `backward_accumulate()`.

LuoYuanke:
@niboshi 

Ok. That's good. Let me check if I understand well. So user can call in their code function node by function node as below?
`with configuration.using_config('lazy_grad_sum', True):
    func_node.apply(...).`

As for `FunctionNode.backward_accumulate()`, in the patch it do gradient sum before calling `backward_accumulate()`. So the out_grad is tuple of variable no matter `backward_accumulate()
` is overridden or not.
             if func.lazy_grad_sum:
                 for y in outputs:
                     grad = get_grad(y)
                     if isinstance(grad, tuple):
                         grad = chainer.functions.accumulate_add(grad)
                         set_grad(y, grad)
             out_grad = tuple([get_grad(y) for y in outputs])
             ...
             gxs = func.backward_accumulate(
                 target_input_indexes, out_grad, in_grad)           
             ...

niboshi:
>So user can call in their code function node by function node as below?

Yes, exactly.

>As for `FunctionNode.backward_accumulate()`, in the patch it do gradient sum before calling `backward_accumulate()`. So the out_grad is tuple of variable no matter `backward_accumulate()
` is overridden or not.

OK, thank you.
By the way, if we don't check if `backward_accumulate()` is overridden, users won't be able to turn `lazy_grad_sum` on if they are using any functions which has the function overridden. It may be OK as the first version of this feature, though.

LuoYuanke:
Hi Niboshi,

Do you have any plan to adopt this feature?

Thanks

chainer/functions/math/acc_add.py
```diff
@@ -0,0 +1,74 @@
...
+    elif g_input is not None:
+        sum_gx += g_input,
+
+    if len(sum_gx) == 0:
+        sum_gx = None,
+
+    return sum_gx
+
+
+class AccumulateAdd(function_node.FunctionNode):
```
niboshi: OTOH, this function can be used generally (not limited gradient accumulation).
How about extending `F.basic_math.Add`?

chainer/function_node.py
```diff
@@ -308,6 +309,8 @@ def apply(self, inputs):
                     retained_data.append(outputs[index])
                 self._retained_output_data = tuple(retained_data)
 
+            self.lazy_grad_sum = configuration.config.lazy_grad_sum
+
```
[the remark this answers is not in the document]
LuoYuanke: Have added it in the new commitment.

chainer/functions/math/acc_add.py
```diff
@@ -0,0 +1,74 @@
+from chainer import function_node
+from chainer import utils
+from chainer.utils import type_check
+
+
+def concat_variable(gx, g_input):
```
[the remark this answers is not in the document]
LuoYuanke: Added chainer/_backprop_utils.py and moved concat_variable to this file.

chainer/functions/math/acc_add.py
LuoYuanke: Moved `AccumulateAdd()` to baisc_math.py.

niboshi:
Thank you for the fix.
For `AccumulateAdd` and `accumulate_add`, instead of introducing new functions, I think it can be implemented in `Add` (to make it accept more than 2 arguments).
Could you fix as such?

LuoYuanke:
Thanks. I have revised the code for the suggestions. Could you help to review?

niboshi:
Thank you.
Could you consider this comment?

>For `AccumulateAdd` and `accumulate_add`, instead of introducing new functions, I think it can be implemented in `Add` (to make it accept more than 2 arguments).
Could you fix as such?

LuoYuanke:
Ok. I understand now, that's elegant. :) Will update patch soon.

LuoYuanke:
Merged `accumulate_add` to `add`, and rename `AccumulateAdd` to `MultiAdd`.

LuoYuanke:
@niboshi 
I refactor the `accumulate_add`. Could you help to review?

chainer/variable.py
```diff
@@ -929,6 +930,15 @@ def get_grad(node):
                 return grads[node]
             return node.grad_var
 
+        def set_grad(node, value):
+            if node is None:
+                return None
```
niboshi: `None` doesn't have to be returned.

chainer/functions/math/basic_math.py
```diff
@@ -169,16 +169,49 @@ def backward(self, indexes, gy):
...
+        y = xs[0] + xs[1]
+        for x in xs[2:]:
+            y += x
+
+        return utils.force_array(y),
+
+    def backward(self, indexes, gy):
+        gys = ()
+        for i in range(self.len):
+            gys += gy[0],
```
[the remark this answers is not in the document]
LuoYuanke: Yes. It's better. :)

chainer/function_node.py
```diff
@@ -308,6 +311,10 @@ def apply(self, inputs):
                     retained_data.append(outputs[index])
                 self._retained_output_data = tuple(retained_data)
 
+            self.lazy_grad_sum = configuration.config.lazy_grad_sum
+            if self.lazy_grad_sum is True:
```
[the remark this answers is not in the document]
LuoYuanke: You mean when `lazy_grad_sum` is disabled, we also print experimental log?

chainer/functions/math/basic_math.py
```diff
@@ -169,16 +169,49 @@ def backward(self, indexes, gy):
...
     """Element-wise addition.
 
     Returns:
         ~chainer.Variable: Output variable.
     """
-    if isinstance(rhs, variable.Variable):
-        return Add().apply((self, rhs))[0]
-    _check_constant_type(rhs)
-    return AddConstant(rhs).apply((self,))[0]
+    if (len(xs) == 2):
```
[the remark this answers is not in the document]
LuoYuanke: Yes. :)

chainer/functions/math/basic_math.py
```diff
@@ -169,16 +169,49 @@ def backward(self, indexes, gy):
...
+        self = xs[0]
+        rhs = xs[1]
```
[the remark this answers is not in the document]
LuoYuanke: Yes.

chainer/variable.py
LuoYuanke: Yes. :)

chainer/function_node.py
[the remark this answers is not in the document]
niboshi: Ah, I meant you can write `if self.lazy_grad_sum:`.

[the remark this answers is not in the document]
LuoYuanke: Ok. I understand. :)

niboshi:
jenkins, test this please

chainer/function_node.py
```diff
@@ -865,7 +890,11 @@ def _backprop(outputs, inputs, grad_required, retain_grad, grads):
                 # Accumulate the duplicated gradients here
                 cur_gx = grads.get(node, None)
                 if cur_gx is not None:
-                    g = g + cur_gx
+                    if x.creator is None:
```
niboshi: Should we enclose this change with `if func.lazy_grad_sum:`?

LuoYuanke: Yes. It reminder me that the cur_gx may be tuple, because the current function's sibling may enable lazy_grad_sum. I'd like to support `variable + tuple of variable` in `+` operator. Do you agree?

LuoYuanke: Yes. It also reminder me that the cur_gx may be tuple, because the function's sibling may enable lazy_grad_sum. Also fixed such cases.

LuoYuanke:
@niboshi 
I modify the code as you suggest. Could you help to review?

tests/chainer_tests/functions_tests/math_tests/test_basic_math.py
```diff
@@ -216,6 +216,58 @@ def test_rpow_double_backward_gpu(self):
         self.double_backward_gpu(lambda x, y: y.__rpow__(x))
 
 
+@testing.parameterize(*testing.product({
+    'shape': [(3, 2), ()],
+    'dtype': [numpy.float16, numpy.float32, numpy.float64],
+}))
+class TestBinaryOpTuple(unittest.TestCase):
```
niboshi: I think `TestBinaryOpTuple` is not appropriate because what is tested is multiple-operand arithmetics.
How about `TestMultipleAdd` or something like that?

tests/chainer_tests/functions_tests/math_tests/test_basic_math.py
```diff
@@ -216,6 +216,58 @@ def test_rpow_double_backward_gpu(self):
...
+        x2 = chainer.Variable(x2_data)
+        x3 = chainer.Variable(x3_data)
+        y = func(x1, x2, x3)
+        options = {}
+        if self.dtype == numpy.float16:
+            options = {'atol': 1e-4, 'rtol': 1e-3}
+        testing.assert_allclose(
+            op(op(self.x1, self.x2), self.x3), y.data, **options)
+
+    def forward_cpu(self, func, op):
```
niboshi: Indirection with this method is needed, because only adding operation can be tested.

tests/chainer_tests/functions_tests/math_tests/test_basic_math.py
```diff
@@ -216,6 +216,58 @@ def test_rpow_double_backward_gpu(self):
...
+        self.forward_cpu(func, op)
+
+    def check_backward(self, op, x1_data, x2_data, x3_data, y_grad):
+        options = {}
+        if self.dtype == numpy.float16:
+            options = {'atol': 5e-3, 'rtol': 5e-2}
+        gradient_check.check_backward(op, (x1_data, x2_data, x3_data), y_grad,
+                                      dtype=numpy.float64, **options)
+
+    def backward_cpu(self, op):
```
niboshi: Ditto. (Indirection with this method is needed)

tests/chainer_tests/functions_tests/math_tests/test_basic_math.py
```diff
@@ -216,6 +216,58 @@ def test_rpow_double_backward_gpu(self):
...
+        self.check_backward(op, self.x1, self.x2, self.x3, self.gy)
+
+    def test_add_backward_cpu(self):
+        op = chainer.functions.add
+        self.backward_cpu(op)
+
+
```
niboshi: Please add double-backward test as well.

tests/chainer_tests/functions_tests/math_tests/test_basic_math.py
```diff
@@ -216,6 +216,58 @@ def test_rpow_double_backward_gpu(self):
...
+    def check_forward(self, func, op, x1_data, x2_data, x3_data):
+        x1 = chainer.Variable(x1_data)
+        x2 = chainer.Variable(x2_data)
+        x3 = chainer.Variable(x3_data)
+        y = func(x1, x2, x3)
+        options = {}
+        if self.dtype == numpy.float16:
+            options = {'atol': 1e-4, 'rtol': 1e-3}
+        testing.assert_allclose(
+            op(op(self.x1, self.x2), self.x3), y.data, **options)
```
niboshi: You can simply write `self.x1 + self.x2 + self.x3` here.

tests/chainer_tests/functions_tests/math_tests/test_basic_math.py
LuoYuanke: Yes. I agree.

tests/chainer_tests/functions_tests/math_tests/test_basic_math.py
LuoYuanke: Ok.

tests/chainer_tests/functions_tests/math_tests/test_basic_math.py
LuoYuanke: I'm not quite understand `Indirection`. But I will update the test code to replace `op` with `self.x1 + self.x2 + self.x3`.

tests/chainer_tests/functions_tests/math_tests/test_basic_math.py
LuoYuanke: Ok.

LuoYuanke:
Refactor test code.

chainer/_backprop_utils.py
```diff
@@ -0,0 +1,36 @@
...
+    elif gx is not None:
+        sum_gx = gx,
+
+    if isinstance(g_input, tuple):
+        sum_gx += g_input
+    elif g_input is not None:
+        sum_gx += g_input,
+
+    if len(sum_gx) == 0:
+        sum_gx = None,
```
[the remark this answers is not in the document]
LuoYuanke: Fixed. And return variable when only one of the inputs in variable.

LuoYuanke:
@toslunar 
Sorry for response later. Thanks for the comments. I clean the code when `lazy_grad_sum` is turned off. The test case is under developing.

LuoYuanke:
Hi toslunar/niboshi

I notice there is some change in master. Very glad to see ideep has been integrated to master chainer. However there is a conflict in my PR and I don't know the process to solve it. The test case for lazy_grad_sum is pushed to github. If it is possible, could you integrated the patch of the PR to the master?

LuoYuanke:
@toslunar 

Test case is added in test_variable.py. Could you help to review if there is any other issues. Thanks.

chainer/__init__.py
```diff
@@ -147,7 +147,9 @@ def is_arrays_compatible(arrays):
 global_config.use_cudnn = os.environ.get('CHAINER_USE_CUDNN', 'auto')
 global_config.use_cudnn_tensor_core = 'auto'
 global_config.autotune = False
-global_config.use_ideep = os.environ.get('CHAINER_USE_IDEEP', 'never')
+global_config.use_ideep = os.environ.get('CHAINER_USE_IDEEP', 'auto')
```
niboshi: Please revert to `'never'`.

tests/chainer_tests/test_variable.py
```diff
@@ -1683,4 +1683,110 @@ def test_intel64_to_cpu(self):
...
+        f2.backward = mock.MagicMock(return_value=(gy10, gy11, gy12))
+        self.y = y
+        self.f2 = f2
+        self.gy10 = gy10
+        self.gy11 = gy11
+        self.gy12 = gy12
+        self.gx = gy00 + gy01 + gy02
+
+    def tearDown(self):
+        self.x = None
```
niboshi: Do we need these cleanup?

niboshi: jenkins, test  this please

niboshi:
jenkins, test this please

tests/chainer_tests/test_variable.py
LuoYuanke: Should not need. I'll remove tearDown().

chainer/__init__.py
LuoYuanke: Sorry for the mistake and thank you for the review. I'll revert it.

LuoYuanke:
@niboshi 
I have updated the patch according to your comments. :)

niboshi:
Thank you!
Finally, could you add `chainer.functions.add` on top of function list documentation? (`docs/source/reference/functions.rst`)
We would need a new section (like "Arithmetic functions"?).

LuoYuanke:
Ok. By the way, do we need move `chainer.functions.absolute` from `Matchematical functions` to `Arithmetic functions`?

niboshi:
Thank you for pointing out.
We may consider reordering of the functions later, but for now could you leave them as they are?

LuoYuanke:
Ok, I will only add `chainer.functions.add` to `Arithmetic functions`.

niboshi:
Thanks!
jenkins, test this please

LuoYuanke:
Hi niboshi,
All checks have passed. Is there some process we should follow before merge the PR?

niboshi:
LGTM!
Thank you very much and I'm sorry for taking time to merge.

LuoYuanke:
Thanks!

PR #4137, opened 2017-12-22
Finalize iterator in evaluator (in trainer)

kuenishi:
This addresses an unreported deadlock bug found during ChainerMN testing.

`MultiProcessIterator` has a bug, when its finalization code runs during program finishing garbage collection it may cause a deadlock in one process, main thread running `__del__()` waits for `self._thread.join()` and never returns. Thus after all main Python path finished the process just hangs with [weird stacktrace](https://gist.github.com/kuenishi/029684b3e6ae3f4bc71510fd47e5cd8e#file-backtrace).

I reproduced with a [simple patching to an evaluate method](https://gist.github.com/kuenishi/029684b3e6ae3f4bc71510fd47e5cd8e#file-diff) which is close to [ChainerMN's new Evaluator usage](https://github.com/chainer/chainermn/pull/160), at current master. As of ChainerMN, this happens with most popular Chainer versions, such as 3.1, 3.2 and 4.0.0b2. Thus I request this patch to be backported to 3.2.

Simplest way to work around this issue is to call iterator's `finalize()` method directly from user code, after all training finished and _before the program exits_, or just use other iterators.

I also suggest removing `__del__()` method from Evaluator and add `__enter__()` and `__exit__()` methods instead, to use with `with` clause.

Note: it looks like [this line](https://github.com/chainer/chainer/blob/master/chainer/iterators/multiprocess_iterator.py#L321) creates cycle reference and may cause re-entrance on `MultiProcessIterator.__del__()` .

### DIFF
--- a/chainer/training/extensions/evaluator.py
+++ b/chainer/training/extensions/evaluator.py
@@ -193,3 +193,8 @@ class Evaluator(extension.Extension):
             summary.add(observation)
 
         return summary.compute_mean()
+
+    def finalize(self):
+        for _, iterator in self._iterators.items():
+            if hasattr(iterator, 'finalize'):
+                iterator.finalize()
--- a/chainer/training/extensions/evaluator.py
+++ b/chainer/training/extensions/evaluator.py
@@ -195,6 +195,12 @@ class Evaluator(extension.Extension):
         return summary.compute_mean()
 
     def finalize(self):
-        for _, iterator in self._iterators.items():
-            if hasattr(iterator, 'finalize'):
-                iterator.finalize()
+        """Finalizes the updater object.
+
+        This method calls the `finalize` method of each iterator that
+        this updater has.
+        It is called at the end of training loops.
+
+        """
+        for iterator in six.itervalues(self._iterators):
+            iterator.finalize()
--- a/chainer/training/extensions/evaluator.py
+++ b/chainer/training/extensions/evaluator.py
@@ -195,11 +195,10 @@ class Evaluator(extension.Extension):
         return summary.compute_mean()
 
     def finalize(self):
-        """Finalizes the updater object.
+        """Finalizes the iterator objects.
 
-        This method calls the `finalize` method of each iterator that
-        this updater has.
-        It is called at the end of training loops.
+        This method calls the `finalize` method of each iterator.
+        They are called at the end of training loops.
 
         """
         for iterator in six.itervalues(self._iterators):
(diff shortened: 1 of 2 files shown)


chainer/training/extensions/evaluator.py
```diff
@@ -193,3 +193,8 @@ def evaluate(self):
             summary.add(observation)
 
         return summary.compute_mean()
+
+    def finalize(self):
```
kmaehashi: Could you please copy the equivalent code (and pydoc) from Updater?
In the imagenet example code, `finalize` of `train_iter` is called from there.

https://github.com/chainer/chainer/blob/v4.0.0b2/chainer/training/updater.py#L169-L178

(Note: iterator base class `chainer.dataset.Iterator` defines `finalize` method, so `hasattr` is not needed)

kuenishi: Thank you for the review. Pushed 74c144f .

chainer/training/extensions/evaluator.py
```diff
@@ -193,3 +193,14 @@ def evaluate(self):
...
+        """Finalizes the updater object.
```
kmaehashi: Ah, this is not updater, but evaluator.

chainer/training/extensions/evaluator.py
```diff
@@ -193,3 +193,14 @@ def evaluate(self):
...
+
+        This method calls the `finalize` method of each iterator that
+        this updater has.
```
kmaehashi: ditto

kuenishi:
I'm very sorry! Pushed another fix.

chainer/training/extensions/evaluator.py
```diff
@@ -195,11 +195,10 @@ def evaluate(self):
         return summary.compute_mean()
 
     def finalize(self):
-        """Finalizes the updater object.
+        """Finalizes the iterator objects.
```
kmaehashi: Hm, as this is a finalization method of evaluator, how about saying `Finalizes the evaluator object.`?
User-defined class that inherits `Evaluator` may want to do other finalization things for evaluator by overriding this method.  In that case the docs looks odd.
This rule fits with the fact that pydoc of Updater's `finalize` methods says `Finalizes the updater object.`

kuenishi: mmm... looks like I've been mad today.

kuenishi:
I newly created https://github.com/chainer/chainer/pull/4137 to prevent merging this ugly history of commits.

PR #4139, opened 2017-12-22
[WIP] Introduce sticky option to Link.to_gpu to support transfer between GPU devices

### LINKED ISSUE #4129: Troubles with moving models into devices
Dear Chainer community. 
I faced with some strange error, I try to do data parallel computation, as  it in the [manual](https://docs.chainer.org/en/stable/tutorial/gpu.html#data-parallel-computation-on-multiple-gpus-without-trainer)

I ran it on four `Tesla V100`:

```
models['model_0'] = MyModel()
models['model_0'].to_gpu(0)

for i in range(1,n_of_gpu):
    models["model_{}".format(i)] = copy.deepcopy(models['model_0'])
    models["model_{}".format(i)].to_gpu(i)
    print("model: {} sent to gpu: {}".format(i,i))
print(models["model_1"].embed.W.data.device)
print(models["model_2"].embed.W.data.device)
print(chainer.__version__)
print(cupy.__version__)
```
The output is:
```
model: 1 sent to gpu: 1
model: 2 sent to gpu: 2
model: 3 sent to gpu: 3
<CUDA Device 0>
<CUDA Device 0>
<CUDA Device 0>
3.2.0
2.2.0

```
So, models don't move into devices `1`  `2` and `3`. What's wrong?

kmaehashi:
Fixes #4129.

TODOs:
- [ ] Support `sticky` option in `Variable.to_gpu`
- [ ] Fix all Links that overrides `to_gpu`

e.g., https://github.com/chainer/chainer/blob/v5.0.0b2/chainer/links/connection/peephole.py#L73-L78

### DIFF
--- a/chainer/link.py
+++ b/chainer/link.py
@@ -358,9 +358,9 @@ Assign a Parameter object directly to an attribute within a \
         Args:
             device: Target device specifier. If omitted, the current device is
                 used.
-            sticky (bool): If ``True``, the link will be transferred to the
+            sticky (bool): If ``False``, the link will be transferred to the
                 specified GPU device even if the Link is already on another
-                GPU. If ``False``, ``to_gpu`` does nothing if the Link is
+                GPU. If ``True``, ``to_gpu`` does nothing if the Link is
                 already on GPU. If omitted, ``True`` is used as the default
                 in Chainer v4. Note that the default is planned to be changed
                 to ``False`` in the future release (possibly in Chainer v5).
@@ -389,11 +389,12 @@ if the Link is already on GPU, `to_gpu` does nothing.
 
 In Chainer v4, `sticky` option has been introduced to `Link.to_gpu` to \
 control this behavior.
-You can specify `sticky=True` option to `to_gpu` to perform inter-GPU transfer.
-If you don't want to perform inter-GPU transfer, explicitly specify
-`sticky=False` so that you can disable this warning.
+You can specify `sticky=False` option to `to_gpu` to perform inter-GPU \
+transfer.
+If you don't want to perform inter-GPU transfer, explicitly specify \
+`sticky=True` so that you can disable this warning.
 
-The default behavior is planned to be changed to "non-sticky" in the future \
+The default behavior is planned to be changed to `sticky=False` in the future \
 release (possibly in Chainer v5).
 '''.format(dst=device_id, src=self._device_id), FutureWarning)
                     return
--- a/chainer/link.py
+++ b/chainer/link.py
@@ -780,12 +780,11 @@ Assign a Link object directly to an attribute within a \
             d[name].to_cpu()
         return self
 
-    def to_gpu(self, device=None):
-        with cuda._get_device(device):
-            super(Chain, self).to_gpu()
-            d = self.__dict__
-            for name in self._children:
-                d[name].to_gpu()
+    def to_gpu(self, device=None, sticky=None):
+        super(Chain, self).to_gpu(device, sticky)
+        d = self.__dict__
+        for name in self._children:
+            d[name].to_gpu(device, sticky)
         return self
 
     def to_intel64(self):
@@ -946,11 +945,10 @@ class ChainList(Link):
             link.to_cpu()
         return self
 
-    def to_gpu(self, device=None):
-        with cuda._get_device(device):
-            super(ChainList, self).to_gpu()
-            for link in self._children:
-                link.to_gpu()
+    def to_gpu(self, device=None, sticky=None):
+        super(ChainList, self).to_gpu(device, sticky)
+        for link in self._children:
+            link.to_gpu(device, sticky)
         return self
 
     def to_intel64(self):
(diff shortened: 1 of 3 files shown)


hvy:
jenkins, test this please.

hvy:
It seems like the (MNIST) examples fail. Could you please take a look?

kmaehashi:
Jenkins, test this please.

kmaehashi:
Jenkins, test this please.

kmaehashi:
MNIST model-parallel example fails becausae it depends on the current behavior.

In this example, the model is [manually transferred to both GPU-0 and GPU-1](https://github.com/chainer/chainer/blob/v4.0.0b3/examples/mnist/train_mnist_model_parallel.py#L15-L29).
Then the [constructor](https://github.com/chainer/chainer/blob/v4.0.0b3/examples/mnist/train_mnist_model_parallel.py#L88-L89) of `StandardUpdater` [automatically calls `to_gpu` of the target link of the optimizer](https://github.com/chainer/chainer/blob/v4.0.0b3/chainer/training/updaters/standard_updater.py#L65-L67) (even though our model is already on GPU-0 and GPU-1).
Formerly, if the target link is already on the GPU, `to_gpu` is ignored.
However, this PR changed the behavior so that `to_gpu` performs GPU-to-GPU transfer, which causes the failure.
Maybe we need some discussion.

Crissman:
@beam2d Please check.

kmaehashi:
For record: `StandardUpdater` started to call `to_gpu` in v1.20.0 to make behavior consistent with `ParallelUpdater` in https://github.com/chainer/chainer/pull/2024

beam2d:
I really like this change, but we should be very careful on this kind of compatibility breaks (esp. when an official example is affected). We can fix `StandardUpdater` and the model-parallel example, but we cannot directly fix user code based on the example and the updater (e.g. copy-paste coded).

I suggest to start from `FutureWarning` in v4 and eventually change it in v5. The warning should be raised when `to_gpu` is called on an array on another gpu. `StandardUpdater` and the model-parallel example should be fixed as well so that it avoids the warning case (i.e., they should be fixed to be compatible with both v4 and b5 versions).

kmaehashi:
I come up with 3 ideas: (argument names are tentative)

* **Idea 1: Redesign Link and its child classes to explicitly allow each parameters to be on different device**: add `skip_if_already_on_gpu=True` argument to `{Link,Chain,ChainList}.to_gpu`, and avoid inter-GPU transfer if the flag is True and it is already on GPU. The default value for the flag will become `False` in v5. StandardUpdater calls `to_gpu(skip_if_already_on_gpu=True)` to support model-parallel Links.
* **Idea 2: Fix StandardUpdater to support model-parallel setup** : add `model_parallel=False` argument to `StandardUpdater.__init__`, and avoid calling `optimizer.target.to_gpu()` if it is True. Fix example code to use it.
* **Idea 3: Consider model-parallel as out of scope of `StandardUpdater` and introduce new updater**: introduce `ModelParallelUpdater` that is quite similar to `StandardUpdater` but does not call `optimizer.target.to_gpu()`. Fix example code to use it.

From compatibility perspective:

* Idea 1:
  * does not break compatibility of example code
  * breaks compatibility of code relying on current `Link.to_gpu` behavior (in v5)

* Idea 2 & 3 both:
  * breaks compatibility of example code (in v4)
  * breaks compatibility of code relying on current `Link.to_gpu` behavior (in v5)

I feel Idea 1 is straight forward.

beam2d:
Option 1 looks good. One caveat is that user-defined Link that overrides to_gpu will be broken. I think it's tolerable because we can print an error message when the argument is not implemented.

kmaehashi:
Or add `def _to_gpu(self, device, skip_if_already_on_gpu): ...` and call it from `to_gpu`?
I feel it is not good to let users copy `skip_if_already_on_gpu=True` to every `to_gpu` overrides to migrate to Chainer v4, because it will be change to `skip_if_already_on_gpu=False` in Chainer v5.

kmaehashi:
I've introduced `sticky` option to `Link.to_gpu`. (`sticky` == `skip_if_already_on_gpu` in Idea 1.)
If this is acceptable, I'll fix all Links overriding `to_gpu` under `chianer/links`.

chainer/link.py
```diff
@@ -358,24 +358,59 @@ def to_gpu(self, device=None):
         Args:
             device: Target device specifier. If omitted, the current device is
                 used.
+            sticky (bool): If ``True``, the link will be transferred to the
```
beam2d: Looks the description for `True` is actually for `False` (and vice versa for the other one)?

kmaehashi: Sorry, you're right. Fixed.

kmaehashi:
I removed `to-be-backported` as fix in this PR became `no-compat` after discussion.

toslunar:
The status of the PR seems asking for approvals.  Could you remove "[WIP]" from the title?

I'm for the PR.

kmaehashi:
Sorry for not updating the status.
I added TODOs in the first comment.

PR #4141, opened 2017-12-22
Use method chain at Optimizer.setup 

toslunar:
#3488 allowed
```
optimizer = chainer.optimizers.SGD(link=model)
```
but it costs writing a new optimizer.

Instead I propose method-chain
```
optimizer = chainer.optimizers.SGD().setup(model)
```

This PR removes the feature (added to v4.0.0b2).
Nonetheless this PR is "more compatible to v3" solution to merge the two lines on an optimizer than #3488 is.

### DIFF
--- a/chainer/optimizer.py
+++ b/chainer/optimizer.py
@@ -383,6 +383,7 @@ class Optimizer(object):
         self.t = 0
         self.epoch = 0
         self._hooks = collections.OrderedDict()
+        return self
 
     def update(self, lossfun=None, *args, **kwds):
         """Updates the parameters.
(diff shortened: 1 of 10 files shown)


mitmul:
@toslunar Thank you for sending this PR. I understood the point. Even if what #3488 introduced turned to be a keywaord-only argument, users may assume all optimizers can take `link` option in the constructors, so that it actually forces all the new optimizers be able to take `link` argument. That's the pain for you, right? If so, I agree to remove the feature and introduces your solution instead. How do you think @Crissman ? Is it also acceptable for you?

Crissman:
Looks good to me. Thanks for the better approach.

kmaehashi:
Note: This is `no-compat` with v4.0.0b2, but has a compatibility with v4.0.0b1 or earlier. (so it does not break our compatibility rule)

Issue #3221, opened 2017-08-23
Defer initialization of size of BatchNormalization and infer it at first iteration.


delta2323:
The input dimension of several links (e.g. `L.Linear`, `L.Convolution`) do not have to be fed and is inferred by the minibatch size of the first iteration. Theoretically we should do the similar thing with `L.BatchNormalization`.

tkanmae:
Hi, I started working on this issue. To implement this lazy inference,`BatchNormalization` needs to know an axis or axes of channel dimension(s) in an input tensor, instead of a size or a shape of input channels. [My work-in-progress implementation](https://github.com/tkanmae/chainer/tree/infer-input-size-in-batchnorm) has `BatchNormalization` with a constructor like

``` python
def __init__(self, axis=1, decay=0.9, eps=2e-5, ...):
``` 
where `axis` is an int or a tuple of ints that specifies an axis or axes of channel dimension(s) in an input tensor. The default value of `1` assumes that a mini-batch size is at the first dimension, channels is at the second dimension and the rest is the spatial dimensions of a sample.

But this will break compatibility with existing code. 

Is there any better approach to implement this feature? What do you think?

delta2323:
Thank you for your interest! I would suggest that the improved implementation should accommodate the original usages. If users assume the original API use this API, the function wrongly interpret the first argument as `axis`. 

Similar problem occurred when we introduce input size inference of `L.Linear`. So its implementation could be helpful.

One idea I come up with is to apply two-step change.
1. Add `None` to the types that `size` accept and add `axis` argument. If size is `None`, `axis` is used to infer the channel dimension(s).
2. Allow users to remove the first argument. If the first argument is removed, users have to specify non-default options as keyword arguments (See [the document of `L.Linear`](https://github.com/chainer/chainer/blob/master/chainer/links/connection/linear.py#L74)).

tkanmae:
Thanks for the feedback. I will continue working on it.

delta2323:
Hi. Can you spare time to continue to work on this? If it is difficult, we may take over your work. thank you.

tkanmae:
Sorry, I've forgot this, but I'm still willing to work on it.  Can you give me a few days to see how far I can go?  If you have a coming deadline and have little time, I'm happy to hand this over to you.

delta2323:
Thank you for your reply. We do not have any special deadline. I would happy if you work on it.

delta2323:
Hi, how are you?

[...]

PR #4149, opened 2017-12-26
Infer input size in batchnorm

tkanmae:
This addresses #3221. If `size`  is None, BatchNormalization infers the channel size from the input array during the first call of `__call__()`. The axis or axes of channels will be determined by `axis` parameters, which default value is 1. To maintain the backward compatibility, the `axis` comes the last of the arguments.

### DIFF
--- a/chainer/links/normalization/batch_normalization.py
+++ b/chainer/links/normalization/batch_normalization.py
@@ -82,11 +82,6 @@ class BatchNormalization(link.Link):
         self._axis = (axis,) if isinstance(axis, int) else tuple(axis)
         self._dtype = dtype
 
-        self.avg_mean = numpy.zeros(self._size, dtype=self._dtype)
-        self.register_persistent('avg_mean')
-        self.avg_var = numpy.zeros(self._size, dtype=self._dtype)
-        self.register_persistent('avg_var')
-
         with self.init_scope():
             if use_gamma:
                 if initial_gamma is None:
@@ -102,12 +97,11 @@ class BatchNormalization(link.Link):
                 beta_initializer.dtype = self._dtype
                 self.beta = variable.Parameter(beta_initializer)
 
-        if self._size is not None:
-            self._initialize_params(self._size)
-
     def _initialize_params(self, shape):
         self.avg_mean = numpy.zeros(shape, dtype=self._dtype)
+        self.register_persistent('avg_mean')
         self.avg_var = numpy.zeros(shape, dtype=self._dtype)
+        self.register_persistent('avg_var')
         if hasattr(self, 'gamma'):
             self.gamma.initialize(shape)
         if hasattr(self, 'beta'):
@@ -137,7 +131,7 @@ class BatchNormalization(link.Link):
                 statistics.
 
         """
-        if not self.avg_mean.shape:
+        if self.gamma.data is None:
             if self._size is not None:
                 self._initialize_params(self._size)
             else:
--- a/chainer/links/normalization/batch_normalization.py
+++ b/chainer/links/normalization/batch_normalization.py
@@ -1,3 +1,5 @@
+import collections
+
 import numpy
 
 from chainer.backends import cuda
@@ -33,10 +35,12 @@ class BatchNormalization(link.Link):
     fine-tuning mode.
 
     Args:
-        size (int, tuples of ints or None): Size (or shape) of channel
-            dimensions. If ``None``, the size will be determined from an axis
-            or axes of channel dimensions given by ``axis`` during the first
-            forward pass.
+        size (int, tuple of ints, None or tuple of Nones): Size (or shape) of
+            channel dimensions. If ``None``, the size will be determined from
+            that of the second dimension of the input batch during the first
+            forward pass. If a tuple of ``None``s, the size will be determined
+            from that of the second and the following dimension(s) of the input
+            batch.
         decay (float): Decay rate of moving average. It is used on training.
         eps (float): Epsilon value for numerical stability.
         dtype (numpy.dtype): Type to use in computing.
@@ -44,10 +48,6 @@ class BatchNormalization(link.Link):
             unit(1) which makes no effect.
         use_beta (bool): If ``True``, use shifting parameter. Otherwise, use
             unit(0) which makes no effect.
-        axis (int or tuple of ints): Axis or axes of channel dimension(s). When
-             ``size``` is ``None``, this value is used to determine parameter
-             sizes during the first forward pass. Otherwise, this value is
-             ignored.
 
     See: `Batch Normalization: Accelerating Deep Network Training by Reducing\
           Internal Covariate Shift <https://arxiv.org/abs/1502.03167>`_
@@ -68,9 +68,9 @@ class BatchNormalization(link.Link):
 
     """
 
-    def __init__(self, size=None, decay=0.9, eps=2e-5,
+    def __init__(self, size, decay=0.9, eps=2e-5,
                  dtype=numpy.float32, use_gamma=True, use_beta=True,
-                 initial_gamma=None, initial_beta=None, axis=1):
+                 initial_gamma=None, initial_beta=None):
         super(BatchNormalization, self).__init__()
 
         self.N = 0
@@ -78,8 +78,10 @@ class BatchNormalization(link.Link):
         self.decay = decay
         self.eps = eps
 
-        self._size = size
-        self._axis = (axis,) if isinstance(axis, int) else tuple(axis)
+        if isinstance(size, collections.Iterable):
+            self._size = size
+        else:
+            self._size = (size,)
         self._dtype = dtype
 
         with self.init_scope():
@@ -97,7 +99,7 @@ class BatchNormalization(link.Link):
                 beta_initializer.dtype = self._dtype
                 self.beta = variable.Parameter(beta_initializer)
 
-        if self._size is not None:
+        if not self._size or any(s is not None for s in self._size):
             self._initialize_params(self._size)
 
     def _initialize_params(self, shape):
@@ -135,8 +137,7 @@ class BatchNormalization(link.Link):
 
         """
         if not hasattr(self, 'avg_mean'):
-            shape = tuple(x.shape[i] for i in self._axis)
-            self._initialize_params(shape)
+            self._initialize_params(x.shape[1:len(self._size) + 1])
 
         argument.check_unexpected_kwargs(
             kwargs, test='test argument is not supported anymore. '
--- a/chainer/links/normalization/batch_normalization.py
+++ b/chainer/links/normalization/batch_normalization.py
@@ -33,8 +33,10 @@ class BatchNormalization(link.Link):
     fine-tuning mode.
 
     Args:
-        size (int or tuple of ints): Size (or shape) of channel
-            dimensions.
+        size (int, tuples of ints or None): Size (or shape) of channel
+            dimensions. If ``None``, the size will be determined from an axis
+            or axes of channel dimensions given by ``axis`` during the first
+            forward pass.
         decay (float): Decay rate of moving average. It is used on training.
         eps (float): Epsilon value for numerical stability.
         dtype (numpy.dtype): Type to use in computing.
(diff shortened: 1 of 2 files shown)


tkanmae:
I must have accidentally deleted `self.` in test code. Just fixed it.

chainer/links/normalization/batch_normalization.py
```diff
@@ -33,15 +33,20 @@ class BatchNormalization(link.Link):
...
         decay (float): Decay rate of moving average. It is used on training.
         eps (float): Epsilon value for numerical stability.
         dtype (numpy.dtype): Type to use in computing.
         use_gamma (bool): If ``True``, use scaling parameter. Otherwise, use
             unit(1) which makes no effect.
         use_beta (bool): If ``True``, use shifting parameter. Otherwise, use
             unit(0) which makes no effect.
+        axis (int or tuple of ints): Axis or axes of channel dimension(s). When
+             ``size``` is ``None``, this value is used to determine parameter
+             sizes during the first forward pass.
```
delta2323: It does not say how this value is used when `size` is not `None`. So, how about adding "Otherwise, this value is not used"?

chainer/links/normalization/batch_normalization.py
```diff
@@ -62,32 +67,41 @@ class BatchNormalization(link.Link):
...
+                beta_initializer = initializers._get_initializer(initial_beta)
+                beta_initializer.dtype = self._dtype
+                self.beta = variable.Parameter(beta_initializer)
+
+    def _initialize_params(self, shape):
+        self.avg_mean = numpy.zeros(shape, dtype=self._dtype)
+        self.register_persistent('avg_mean')
+        self.avg_var = numpy.zeros(shape, dtype=self._dtype)
+        self.register_persistent('avg_var')
+        self.gamma.initialize(shape)
```
delta2323: If `use_gamma` is `False`, the instance does not have `gamma` attribute. So, please check its existence (e.g. with `hasattr`).

chainer/links/normalization/batch_normalization.py
```diff
@@ -62,32 +67,41 @@ class BatchNormalization(link.Link):
...
+        self.beta.initialize(shape)
```
delta2323: ditto (check existence of `self.beta`)

chainer/links/normalization/batch_normalization.py
```diff
@@ -62,32 +67,41 @@ class BatchNormalization(link.Link):
...
     """
 
-    def __init__(self, size, decay=0.9, eps=2e-5, dtype=numpy.float32,
-                 use_gamma=True, use_beta=True,
-                 initial_gamma=None, initial_beta=None):
+    def __init__(self, size=None, decay=0.9, eps=2e-5,
+                 dtype=numpy.float32, use_gamma=True, use_beta=True,
+                 initial_gamma=None, initial_beta=None, axis=1):
         super(BatchNormalization, self).__init__()
-        self.avg_mean = numpy.zeros(size, dtype=dtype)
```
delta2323: It changes the behavior of the documented API `self.avg_mean` slightly because after this change, we cannot get this attribute immediately after instantiation when `size` is not `None`. So, please call `self._initialize_param` in `__init__` when  `size` is not `None`.

chainer/links/normalization/batch_normalization.py
```diff
@@ -62,32 +67,41 @@ class BatchNormalization(link.Link):
...
-        self.register_persistent('avg_mean')
-        self.avg_var = numpy.zeros(size, dtype=dtype)
```
delta2323: Same as `avg_mean`.

delta2323:
Please test the cases in which `use_gamma` and/or `use_beta` is `False` when `size=None`,

delta2323:
Please test the case in which we can have get `self.avg_mean`, `self.avg_var` attributes even before the first forward propagation and `size` argument of `__init__` is `None`.

chainer/links/normalization/batch_normalization.py
tkanmae: The current implementation in master branch does not set `gamma` and `beta` if `use_gamma` and `user_beta` is `False`, respectively. I just tried the following code

```python
>> bn = BatchNormalization(16, use_gamma=False, use_beta=False)
>> hasattr(bn, 'beta')
False
>> hasattr(bn, 'gamma')
False
```

Am I missing something here?

tkanmae: My bad, I got what you mean. You want me to check if an instance has the attributes in `_initialize_params()`. I will fix what you pointed out as soon as possible. Thanks!

tkanmae:
@delta2323 I have worked on the issues you pointed out.

Now, the constructor initializes `avg_mean` and `avg_var` using `np.zeros(size)` regardless of whether `size` is `None`. This makes `avg_mean` and `avg_var `become `np.array(0.0, dtype=float32)` if `size` is `None`. What do you think?

chainer/links/normalization/batch_normalization.py
delta2323: Yes, that was what I wanted to say. Sorry for confusing you.

delta2323:
> Please test the case in which we can get self.avg_mean, self.avg_var attributes even before the first forward propagation when size argument of __init__ is None.

Sorry, I was mistaken in my previous comment. I wanted to check when `__init__` is *not* `None`. That is consistent with [this comment]()

tkanmae:
@delta2323 I should have caught it. I reverted relevant commits, had `_initialize_param()` get called when `size` is not `None`, and add test cases as you requested.

chainer/links/normalization/batch_normalization.py
```diff
@@ -33,15 +33,21 @@ class BatchNormalization(link.Link):
...
+            forward pass.
         decay (float): Decay rate of moving average. It is used on training.
         eps (float): Epsilon value for numerical stability.
         dtype (numpy.dtype): Type to use in computing.
         use_gamma (bool): If ``True``, use scaling parameter. Otherwise, use
             unit(1) which makes no effect.
         use_beta (bool): If ``True``, use shifting parameter. Otherwise, use
             unit(0) which makes no effect.
+        axis (int or tuple of ints): Axis or axes of channel dimension(s). When
+             ``size``` is ``None``, this value is used to determine parameter
```
delta2323: Please describe how we infer `size` using `axis` explicitly.

delta2323:
Please add unit tests in which `axis` is different from the default value `1`.

delta2323:
I noticed that current code does not work when `axis` is a list and its values are non-contiguous.

Code to reproduce
```
from chainer import links
import numpy

link = links.BatchNormalization(None, axis=(0, 2))
x = numpy.random.randn(2, 3, 4).astype('f')
link(x)
link(x)
```

Log
```
%python test.py 
  File "test.py", line 8, in <module>
    link(x)
  File "/Users/oonokenta/dev/chainer2/chainer/links/normalization/batch_normalization.py", line 169, in __call__
    running_var=self.avg_var, decay=decay)
  File "/Users/oonokenta/dev/chainer2/chainer/functions/normalization/batch_normalization.py", line 586, in batch_normalization
    (x, gamma, beta))[0]
  File "/Users/oonokenta/dev/chainer2/chainer/function_node.py", line 242, in apply
  File "/Users/oonokenta/dev/chainer2/chainer/function_node.py", line 325, in _check_data_type_forward
  File "/Users/oonokenta/dev/chainer2/chainer/functions/normalization/batch_normalization.py", line 46, in check_type_forward
    gamma_type.shape == beta_type.shape,
  File "/Users/oonokenta/dev/chainer2/chainer/utils/type_check.py", line 524, in expect
    expr.expect()
  File "/Users/oonokenta/dev/chainer2/chainer/utils/type_check.py", line 482, in expect
Invalid operation is performed in: BatchNormalization (Forward)

Expect: in_types[0].shape[1:3] == in_types[1].shape
Actual: (3, 4) != (2, 4)
```

tkanmae:
@delta2323 I want to make sure what expected behavior is in those two cases that you pointed out.  According to the following doc of `functions.batchnormalization()`,

> It takes the input variable ``x`` and two parameter variables ``gamma`` and
    ``beta``. The parameter variables must both have the same dimensionality,
    which is referred to as the channel shape. This channel shape corresponds
    to the dimensions in the input which are not averaged over. Since the
    first dimension of the input corresponds to the batch size, the second
    dimension of `x` will correspond to the first dimension of the channel
    shape, the third dimension of `x` will correspond to the second channel
    dimension (if it exists) and so on. Therefore, the dimensionality of the
    input must be at least one plus the number of channel dimensions. The
    total effective "batch size" will then be considered to be the product of
    all dimensions in `x` except for the channel dimensions.

>    As an example, if the input is four dimensional and the parameter
    variables are one dimensional, then it is assumed that the first
    dimension of the input is the batch size, the second dimension is the
    channel size, and the remaining two dimensions are considered
    to be spatial dimensions that will be averaged over along with the
    batch size in the batch normalization computations. That is,
    the total batch size will be considered to be the product of all
    input dimensions except the second dimension.

the desired value of `axis`  is 1 or a sequence of `ints` that should be starting from 1 to `n-1`, where `n` is the dimension of an input. So, the cases you pointed out (`axis=2` and `axis=(0, 2)`) should be invalid, and  `BatchNormalization` should raises ValueError.

Is this expected behavior?

toslunar:
> the desired value of axis is 1 or a sequence of ints that should be starting from 1 to n-1, where n is the dimension of an input. 

`axis` must be one of `()`, `(1,)`, `(1, 2)`, `(1, 2, 3)`, ..., or `1` in the current implementation.  It doesn't need to end with ndim - 1.

Fixing `(fixed_)batch_normalization()` is an option.  Another option is raise an error as tkanmae wrote.  Yet another option is to have different interface.  I came up with two interfaces:
- `ndim_channels` to specify `axis = tuple(range(1, 1 + ndim_channels))`,
- `size=(None, None)` or `size=(-1, -1)` to specify `axis=(1, 2)`

tkanmae:
I'm with the second option of @toslunar's 

> - size=(None, None) or size=(-1, -1) to specify axis=(1, 2)

Given the constraint in the current implementation, the users can not select arbitrary axes with `axis` argument. This is not what the user may expect from the name of the argument. In NumPy, you can select arbitrary axes with `axis` argument. 

I think that specifying `size` with `None` or a tuple of `None`s is more appropriate and is in line with the current API.

delta2323:
@tkanmae I agree with @toslunar's design. Just for confirmation, are you thinking to remove `axis` option when we adopt this API?

tkanmae:
@delta2323 Yes, I am. I will remove `axis` argument, and enable `size` to have `None` or a tuple of `None`s to indicate that it will infer the channel size(s) from an input array.

delta2323:
OK. I agree with it. Could you try to implement it?

tkanmae:
You bet.

tkanmae:
@delta2323 I worked on new API as @toslunar suggested. How do you think?

delta2323:
API design is basically OK to me. We need to consider when size if a mixture of `None`'s and integers (e.g. `(None, 2, None, 1)`).

delta2323:
@tkanmae #4246 is working to extend BN to enable us to specify batch axes. As the PR might affect to this PR, could you take a look?

delta2323:
Sorry not #4246 but #4266

tkanmae:
@delta2323 I took a look at the PR. I agree with both of you in that these PRs can be done separately, though some degree of coordination may be necessary from my side. As @toslunar suggested, I think that it make sense to just adding `size = None` in `BatchNormalization` once that PR is accepted.

tkanmae:
@delta2323 I think it is better to resume working on this after #4266 is settled. How do you think?

delta2323:
> I think it is better to resume working on this after #4266 is settled. How do you think?

@tkanmae Sorry for late response. I agree with you. Let's discuss in the PR.

delta2323:
I reconsider the design of automatic input size inference and would like to adopt #4266 as its basic design. If we would do so, a part of your implementation in this PR would not be used. I'm really sorry for our decision and would be happy if you would join the discussion in #4266.

tkanmae:
@delta2323 Thanks, no problem!

delta2323:
@tkanmae Thank you very much. I really appreciate effort you made.

delta2323:
@tkanmae @toslunar Now #4266 has been merged, can we close this PR? I think it would be better to create new PRs if we want to add features that are not covered by #4266.

toslunar:
It seems easy to merge `chainer/master`: https://github.com/toslunar/chainer/commits/infer-input-size-in-batchnorm

toslunar:
Note: `axis` in early conversation is called `key_axis` in #4266.  It is the complement of the aggregate axes.

kmaehashi:
This PR has been taken over to #4673. As the review in #4673 is ongoing, I think it's OK to close this.
@toslunar Could you close this PR?

Issue #4143, opened 2017-12-25
Create a list of breaking changes in one page (upgrade guide from v2 to v3 and v4)


kmaehashi:
From the discussion here: https://github.com/chainer/chainer/pull/3981#discussion_r155132277
I'd like to prepare docs that users can read and update their code in chronological order.

kmaehashi:
Fixed via #4151

PR #4151, opened 2017-12-26
Add upgrade guide for v3 and v4

### LINKED ISSUE #3592: Write upgrade guide from v2 to v3
We have released [v3.0.0](https://github.com/chainer/chainer/releases/tag/v3.0.0), but we don't have the upgrade guide yet.
It's better to have one even if compatibility is mostly kept.

kmaehashi:
This fixes #3592 and #4143.

I renamed existing upgrade guide (v1 to v2) as `upgrade_v2.rst`, as it is huge.
Then new upgrade guide (v2 to v3, v3 to v4) is placed as `upgrade.rst`, with a link to `upgrade_v2.rst`.
So, I recommend to review per-commit rather than full diff.

When backporting, please remove the v4 part.


niboshi:
This PR solves #3592.

docs/source/upgrade.rst
```diff
@@ -1,819 +1,124 @@
...
+   v2 = chainer.Variable(cupy.arange(10, dtype=cupy.float32))  # GPU
 
-.. admonition:: Example
+   # The line below raises an exception, because arguments are on different device.
+   F.maximum(v1, v2)
 
-   Consider the following model definition and the code to call it in test mode written for Chainer v1.
+Prior to v4, the above code raises an exception like ``ValueError: object __array__ method not producing an array``, which was difficult to understand.
+In v4, the error message would become ``ValueError: numpy and cupy arrays are mixed in the forward input (Maximum)``.
+This kind of error usually occur by mistake (for example, not performing ``to_gpu`` for some variables).
```
niboshi: `occurs`?

docs/source/upgrade.rst
```diff
@@ -1,819 +1,124 @@
...
+.. attention::
 
-      # Chainer v1
-      import chainer.functions as F
+   As the argument validation is strictened, call of new-style functions (inheriting :class:`FunctionNode` class) intentionally mixing NumPy/CuPy arrays in arguments will not work in Chainer v4.
+   Please transfer all arrays to the same device before calling functions.
 
-      class MyModel(chainer.Link):
-          ...
+Note that old-style function classes (inheriting :class:`Function`) are not affected.
```
niboshi: This sentence is not true. `FunctionAdapter` raises the error.

docs/source/upgrade.rst
```diff
@@ -1,819 +1,124 @@
...
 
-          def __call__(self, x, train=True):
-              return f(F.dropout(x, train=train))
+Reproducibility Improvement with MultiProcessIterator
```
niboshi: Do we need this topic in upgrade guide?

docs/source/upgrade.rst
```diff
@@ -1,819 +1,124 @@
...
+-----------------------------------------------------
 
-      m = MyModel(...)
-      y = m(x, train=False)
+Prior to v4, NumPy random states are shared between threads when using :class:`chainer.iterators.MultiprocessIterator`.
+This release fixed this issue by using a separate random state for a sub thread to improve reproducibility of experiments.
+See the discussion in `#3575 <https://github.com/chainer/chainer/pull/3575>`_ for more details.
 
-   In Chainer v2, it should be updated into the following code:
+Update of Docker Images
```
niboshi: ditto. (Do we need this topic in upgrade guide?)

docs/source/upgrade.rst
```diff
@@ -1,819 +1,124 @@
...
+-----------------------------------
 
-      m = MyModel(...)
-      with chainer.using_config('train', False):
-          y = m(x)
+This release introduces new-style functions (classes inheriting from :class:`FunctionNode`) that support double backward (gradient of gradient).
+See the `Release Note for v3.0.0 <https://github.com/chainer/chainer/releases/tag/v3.0.0>`_ for the usage of this feature.
 
-.. _upgrade-configurations:
+Many of :doc:`reference/functions` are already migrated to new-style, although some of functions are still old-style (classes inheriting from :class:`Function`).
```
niboshi: Do we need this paragraph?
Instead, we should state that writing new functions with `FunctionNode` is encouraged.

docs/source/upgrade.rst
```diff
@@ -1,819 +1,124 @@
...
+   Please use the function APIs documented in :doc:`reference/functions`.
 
-These configurations can be modified in two ways.
+Changed Behavior of matmul Function
+-----------------------------------
 
-- Simply subsituting a new value to an entry, like ``chainer.config.train = False``.
-- Using the ``chainer.using_config`` context manager.
-  It can be used with the ``with`` statement of Python as follows::
+The behavior of :func:`chainer.functions.matmul` has been changed to behave like ``matmul`` function of NumPy.
```
niboshi: ``` ``matmul`` ``` -> Link to `numpy.matmul` documentation

docs/source/upgrade.rst
```diff
@@ -1,819 +1,124 @@
...
---------
+``use_cudnn`` argument has been removed from :func:`chainer.functions.spatial_transformer_grid` and :func:`chainer.functions.spatial_transformer_sampler`.
+See the discussion in `#2955 <https://github.com/chainer/chainer/pull/2955>`_ for more details.
 
-.. _upgrade-volatile-removed:
+.. attention::
 
-Volatile flag is removed
-~~~~~~~~~~~~~~~~~~~~~~~~
+   Codes using ``use_cudnn`` argument of :func:`chainer.functions.spatial_transformer_grid` and :func:`chainer.functions.spatial_transformer_sampler` require modification to work with Chainer v3.
```
niboshi: codes -> the existing code?

docs/source/upgrade.rst
```diff
@@ -1,819 +1,124 @@
...
-  It can be used with the ``with`` statement of Python as follows::
+The behavior of :func:`chainer.functions.matmul` has been changed to behave like ``matmul`` function of NumPy.
+See the discussion in `#2426 <https://github.com/chainer/chainer/pull/2426>`_ for more details.
 
-      with chainer.using_config('train', False):
-          do something  # this code runs with chainer.config.train == False
+.. attention::
 
-  It recovers the original configuration after quitting the ``with`` block.
+   Codes using :func:`chainer.functions.matmul` may require modification to work with Chainer v3.
```
niboshi: ditto. (codes -> the existing code?)

niboshi:
I think we should mention optimizer setup (#3488, #3940), too.

niboshi:
Let's keep from mentioning optimizer setup until the discussioin in  #4141 will be settled.

docs/source/upgrade.rst
kmaehashi: I wrote it here because the issue had `no-compat` label, but I now think it is not needed to be on the upgrade guide, because we don't guarantee that the reproducibility of computation results between any two version.

docs/source/upgrade.rst
kmaehashi: I think this is needed, because CUDA 7.5 users can no longer run the official docker image.

docs/source/upgrade.rst
kmaehashi: In this paragraph I mean "currently not all functions supports double backprop". I'll make it clear.

> we should state that writing new functions with FunctionNode is encouraged.

I agree with this.

docs/source/upgrade.rst
niboshi: OK. More precisely, in general we don't guarrantee the reproducibility within the same version, if my understanding is correct. (#3575 seems to be an exception but it's not documented anyway).

docs/source/upgrade.rst
kmaehashi: I mean, users with older NVIDIA driver version cannot run the Docker image, although we still continue to support CUDA 7.5.
https://github.com/NVIDIA/nvidia-docker/wiki/CUDA#requirements

kmaehashi:
Thanks for review! I fixed accordingly and rebased.

> Let's keep from mentioning optimizer setup until the discussioin in #4141 will be settled.

OK.

niboshi:
Now #4141 has been merged.
I think it's ok whether or not to mention it, as the difference (between v3 method and meregd one) is not so large.

kmaehashi:
#4141 is just a syntax sugar so I feel it is not needed to be in this guide.

(BTW #4141 lacks pydoc fixes ("Returns:") so we should fix that)

niboshi:
jenkins, test this please

niboshi:
LGTM!

PR #4154, opened 2017-12-27
Remove empty retain

hvy:
Remove unneeded empty retain (it is by default empty in `FunctionNode`s).


okuta:
LGTM!

kmaehashi:
I think this is to-be-backported.

okuta:
OK, I will

PR #4156, opened 2017-12-27
Write comment about direction normalization in gradient_check.check_backward

niboshi:
The purpose of normalization of the direction vector is not easily guessed from the code.
I described the purpose and assumption.


hvy:
LGTM!

PR #4159, opened 2017-12-28
Add support for post-update hook registration to optimizers

tkerola:
Currently, `Optimizer` hooks are only called before the parameter update. This PR adds support
for registering `Optimizer` and `UpdateRule` hooks that are called after the parameter update.
This is useful for implementing methods such as Mean Teacher (https://arxiv.org/abs/1703.01780) that do additional parameter updates after the main update rule. Without this PR, the user currently has to reimplement a set of methods `MeanTeacherAdam`, `MeanTeacherSGDMomentum`, etc. in order to try out different optimizers with the same training scheme, which is cumbersome and not scalable. Post-update hooks solve this problem by allowing the user to define one single hook that can then be used with all current optimizers.

Also, with the post-update `UpdateRule` hooks, if would be possible to easily implement the modified weight decay proposed for AdamW (https://openreview.net/forum?id=rk6qdGgCZ) by combining this hook with a regular (pre-update) `UpdateRule` hook in a scalable manner for all supported optimizers.


tests/chainer_tests/optimizers_tests/test_optimizers.py
```diff
@@ -45,4 +47,71 @@ def test_hyperparams(self):
...
+
+class SimpleChain(chainer.Chain):
+
+    def __init__(self):
+        super(SimpleChain, self).__init__()
+        with self.init_scope():
+            self.w = chainer.Parameter(42, (), 'w')
+
+    def __call__(self, x):
+        return (x - self.w)**2
```
delta2323: Add spaces between `**` operator.

delta2323:
jenkins, test this please.

beam2d:
I feel the method names `add_hook` and `add_post_update_hook` looks inconsistent. How about adding an optional flag to `add_hook` so that users can use `add_hook` for both purposes, like `add_hook(hook, name, timing='post')`? It is also good to let each hook have its own default flag so that most users do not have to specify the timing by themselves.

tkerola:
That's a good idea, to avoid unnecessary boilerplate code. I will update the PR as you suggested.

tkerola:
I added a `timing` attribute used for selecting the timing of the hook. For backwards compatibility, the timing will default to `pre` if not specified.

chainer/optimizer.py
```diff
@@ -141,30 +143,42 @@ def state(self):
...
+        The hook function is called before or after any updates (see the timing
+        attribute).
 
         Args:
             hook (callable): Hook function to be added. It takes two
                 arguments: the update rule object and the parameter variable.
             name (str): Name of the hook function. The name attribute of the
                 hook function is used by default.
+            timing (str): Specifies when the hook is called. If 'auto', the
+                timimg property of the hook will decide the timing.
```
delta2323: Please write that `auto` is treated as `pre` if `timing` property is not available.

chainer/optimizer.py
```diff
@@ -554,9 +599,11 @@ def reallocate_cleared_grads(self):
                     xp = cuda.get_array_module(param.data)
                     param.grad = xp.zeros_like(param.data)
 
-    def call_hooks(self):
+    def call_hooks(self, hooks=None):
```
delta2323: I noticed that by adding the `hooks` option, users can invoke hooks that are not registered. I think it does not break documented behaviors of this modules but would like to know if it is an intended behavior.

delta2323:
@beam2d Do you think we should promote `call_for_each_params` and `timing` properties in existing hooks to documented ones?

tests/chainer_tests/optimizers_tests/test_optimizers.py
```diff
@@ -45,4 +47,72 @@ def test_hyperparams(self):
...
+
+    def get_hyperparam(self, name):
+        return getattr(self.target.w.update_rule.hyperparam, name)
+
+    def test_hooks(self):
+        w_pre = np.copy(self.target.w.data)
+        h_pre = WeightSaveHook()
+        h_post = WeightSaveHook()
+        self.create()
+        self.optimizer.add_hook(h_pre)
```
delta2323: Current tests do not check the behavior of `timing=pre` case.
I would suggest to put `timing='pre'` explicitly here and check `timing=auto` case as an another test.

beam2d:
@delta2323 Yes, they should be documented.

delta2323:
@beam2d OK. Thank you.
@tkerola Could you add `timing` to the document of hook functions?

chainer/optimizer.py
tkerola: Good catch. I did not intend this behavior. I think it's best to avoid this.

tests/chainer_tests/optimizers_tests/test_optimizers.py
tkerola: Ok, will do.

chainer/optimizer.py
tkerola: Ok, sure.

tkerola:
Ok, I updated the PR according to your comments.

tkerola:
Hmm, I am not sure why TravisCI tests fail in `chainer_tests/test_variable.py`. That should be unrelated to this PR.

tkerola:
Hmm, this may be due to NumPy 1.14? https://github.com/numpy/numpy/blob/master/doc/release/1.14.0-notes.rst#many-changes-to-array-printing-disableable-with-the-new-legacy-printing-mode

kmaehashi:
I fixed the issue in #4172 and restarted the travis build for this PR.

delta2323:
@kmaehashi Thank you for your comment on the Travis problem.

chainer/optimizer.py
delta2323: One idea is to feed `str` that takes either `pre` or `post` and get hooks inside of the method, it loses flexibility, though.

chainer/optimizer.py
```diff
@@ -675,10 +730,14 @@ class WeightDecay(object):
 
     Attributes:
         rate (float): Coefficient for the weight decay.
+        timing (string): Specifies when this hook should be called by the
+                         Optimizer/UpdateRule. Valid values are 'pre'
+                         (before any updates) and 'post' (after any updates).
```
delta2323: Could you write that `timing` is added from v4.0.0? You can use `versionadded` markup (See usage [here](http://www.sphinx-doc.org/en/stable/markup/para.html#directive-versionadded))

chainer/optimizer.py
```diff
@@ -707,10 +766,14 @@ class Lasso(object):
...
```
delta2323: ditto (`versionadded::`)

chainer/optimizer.py
```diff
@@ -741,9 +804,13 @@ class GradientClipping(object):
 
     Attributes:
         threshold (float): L2 norm threshold of gradient norm.
+        timing (string): Specifies when this hook should be called by the
+                         Optimizer/UpdateRule. Valid values are 'pre'
+                         (before any updates) and 'post' (after any updates).
```
delta2323: ditto (`versionadded::`)

chainer/optimizer.py
```diff
@@ -786,9 +853,16 @@ class GradientNoise(object):
         noise_func (function): Noise generating function which by default
             is given by `Adding Gradient Noise Improves Learning for Very Deep\
             Networks <https://arxiv.org/pdf/1511.06807>`_.
+
+    Attributes:
+        timing (string): Specifies when this hook should be called by the
+                         Optimizer/UpdateRule. Valid values are 'pre'
+                         (before any updates) and 'post' (after any updates).
```
delta2323: ditto (`versionadded::`)

chainer/optimizer.py
```diff
@@ -823,10 +897,14 @@ class GradientHardClipping(object):
     Attributes:
         lower_bound (float): The lower bound of the gradient value.
         upper_bound (float): The upper bound of the gradient value.
+        timing (string): Specifies when this hook should be called by the
+                         Optimizer/UpdateRule. Valid values are 'pre'
+                         (before any updates) and 'post' (after any updates).
```
delta2323: ditto (`versionadded::`)

chainer/optimizer.py
tkerola: Ok, I added it. Please see if I understood the syntax correctly.

tkerola:
I added the version tag to the documentation. Did I understand the format correctly?

delta2323:
> I added the version tag to the documentation. Did I understand the format correctly?

I confirmed it works.

delta2323:
Do you think some users want to switch a invocation timing of a hook from (or vice versa) during training process? To realize it with current implementation, they need to delete the hook from the pre-hook list and add new one to the post-hook list. One drawback is that they cannot take over states of the hook when the timing is changed. I would like your opinions.

tkerola:
I am not aware of any such use-case where the user would want to switch the timing during training. If you are, please let me know. If such functionality is needed, we could implement it by looping over the `_hooks` dict twice (before and after the update) and checking the value of the `timing` attribute. But that would cause unnecessary overhead if such a use-case is not needed.
With the current implementation, in addition to the workaround you mentioned, another way of switching the timing would be to create two hooks (one for pre and another for post), that could share a common state and an `enabled` variable, that allows the user to switch between them.
So, my opinion is not to enable smooth switching of timing during training, unless there is an established use-case for it already, since it will cause additional overhead. What do you think?

chainer/optimizer.py
tkerola: Fixed.

chainer/optimizer.py
tkerola: Fixed.

chainer/optimizer.py
tkerola: Fixed.

chainer/optimizer.py
tkerola: Fixed.

delta2323:
@tkerola Thank you for your comment. I agree with you. So, we will consider it when we find such a use case.

delta2323:
jenkins, test this please.

delta2323:
The following warning occurred. Could you check it?

```
22:51:18 /home/user/.local/lib/python3.5/site-packages/chainer-4.0.0b2-py3.5.egg/chainer/optimizer.py:docstring of chainer.optimizer.GradientClipping:None: WARNING: more than one target found for cross-reference 'timing': chainer.optimizer.GradientNoise.timing, chainer.optimizer.WeightDecay.timing, chainer.optimizer.GradientClipping.timing, chainer.optimizer.Lasso.timing
```

tkerola:
Hmm, I cannot seem to reproduce that warning. Could you please tell me the environment you used when getting that warning? (Or point me to the AppVeyor/TravisCI build that caused that error if you found it there?)

tkerola:
I tried to make the annotation in the docstring more specific. Does this solve the issue?

delta2323:
Thank you. I'll rerun the test.

delta2323:
Jenkins, test this please.

delta2323:
LGTM

delta2323:
If I understand correctly, we do not have to make a backport for this PR as the final release of Chainer v3 has been finished and and Chainer v4 already has this feature.

Issue #3098, opened 2017-08-04
CaffeFunction doesn't load global average pooling correctly


belltailjp:
In case a caffemodel fed to CaffeFunction contains global average pooling layers, these layers have `global_pooling: true` parameter but no window size values are stored in the caffemodel file.
However, current CaffeFunction implementation (specifically, definition of `_stup_pooling` in `caffe_function.py`) doesn't take care of this value (more specifically `param.global_pooling` should be checked within the code), and eventually generated AveragePooling2D object has 0 in both `kw` and `kh`. As the result forward pass cannot be run.

Here is the minimum reproduction code and the actual error I got.
(test caffemodel `DenseNet_161.caffemodel` can be downloaded from https://github.com/shicai/DenseNet-Caffe )

* Chainer: 2.0.2
* Cupy: 1.0.2
* Python: 3.6.1 (using pyenv, anaconda3-4.4.0)
* OS: Ubuntu 16.04 x64
* CUDA: 8.0
* cuDNN: 5.1

```
import cupy
import chainer
from chainer.links import caffe

net = caffe.CaffeFunction("/path/to/DenseNet_161.caffemodel")
net.to_gpu(0)
x = cupy.random.random((1, 3, 224, 224)).astype(cupy.float32)
net(inputs={'data': x}, outputs=['fc6'])
```

Full error message:

```
% CHAINER_CUDNN=0 python reproduction.py
  File "reproduction.py", line 8, in <module>
    net(inputs={'data': x}, outputs=['fc6'])
  File "/home/foobar/.pyenv/versions/anaconda3-4.4.0/lib/python3.6/site-packages/chainer/links/caffe/caffe_function.py", line 208, in __call__
    output_vars = func(*input_vars)
  File "/home/foobar/.pyenv/versions/anaconda3-4.4.0/lib/python3.6/site-packages/chainer/links/caffe/caffe_function.py", line 558, in __call__
    return self.func(x, *self.args, **self.kwargs)
  File "/home/foobar/.pyenv/versions/anaconda3-4.4.0/lib/python3.6/site-packages/chainer/functions/pooling/average_pooling_2d.py", line 144, in average_pooling_2d
    return AveragePooling2D(ksize, stride, pad, False)(x)
  File "/home/foobar/.pyenv/versions/anaconda3-4.4.0/lib/python3.6/site-packages/chainer/function.py", line 200, in __call__
    outputs = self.forward(in_data)
[...]

PR #4161, opened 2017-12-28
Support Caffe global pooling functions

hvy:
Fixes https://github.com/chainer/chainer/issues/3098

Wrapped the Chainer pooling functions so that the kernel sizes can be computed dynamically to cover the whole input. 

Caffe reference https://github.com/BVLC/caffe/blob/master/src/caffe/proto/caffe.proto#L921

### DIFF
--- a/chainer/links/caffe/caffe_function.py
+++ b/chainer/links/caffe/caffe_function.py
@@ -328,7 +328,14 @@ class CaffeFunction(link.Chain):
         else:
             raise RuntimeError('Stochastic pooling is not supported')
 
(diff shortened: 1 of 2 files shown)


hvy:
Thank you for the comment, I added a test as you said!

mitmul:
LGTM

kmaehashi:
Is this PR closed without merging?

hvy:
I think it is a mistake so I'll reopen it.

mitmul:
Ooops, sorry! I'm gonna merge this.

hvy:
@mitmul Maybe it should be backported as well?

PR #4167, opened 2018-01-09
update pydocs for #4141

kmaehashi:
Update pydoc for change introduced in #4141 (optimizer setup syntax sugar)

### DIFF
--- a/chainer/optimizer.py
+++ b/chainer/optimizer.py
@@ -376,6 +376,14 @@ class Optimizer(object):
         Args:
             link (~chainer.Link): Target link object.
 
+        Returns:
+            The optimizer instance.
+
+        .. note::
+           As of v4.0.0, this function returns the optimizer instance itself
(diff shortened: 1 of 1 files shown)


toslunar:
Thanks!

chainer/optimizer.py
```diff
@@ -376,6 +376,14 @@ def setup(self, link):
         Args:
             link (~chainer.Link): Target link object.
 
+        Returns:
+            The optimizer instance.
+
+        .. note::
+           As of v4.0.0, this function returns the optimizer instance itself
+           so that you can instantiate and setup the optimizer in one line,
+           e.g., ``optimizer = SomeOptimizer().setup(model)``.
```
toslunar: `.setup(link)` or `.setup(link=model)` would be more informative to me.

kmaehashi: Thanks, fixed as suggested.

PR #4172, opened 2018-01-10
add unit test for NumPy 1.14 array textual representation

kmaehashi:


tests/chainer_tests/test_variable.py
```diff
@@ -1323,22 +1323,7 @@ def test_backward_gpu(self):
...
-     'repr': 'variable([[ 0.,  1.],\n          [ 2.,  3.]])',
-     'str': 'variable([[ 0.  1.]\n          [ 2.  3.]])'},
-    {'x_shape': (2, 2,), 'dtype': np.float64,
-     'repr': 'variable([[ 0.,  1.],\n          [ 2.,  3.]])',
-     'str': 'variable([[ 0.  1.]\n          [ 2.  3.]])'},
-    {'x_shape': (3,),  'dtype': np.float32,
-     'repr': 'variable([ 0.,  1.,  2.])', 'str': 'variable([ 0.  1.  2.])'},
-)
-class TestUnnamedVariableToString(unittest.TestCase):
+class _TestUnnamedVariableToStringBase(object):
```
niboshi: I think the prefix `_` is not necessary because this is a test code.

tests/chainer_tests/test_variable.py
```diff
@@ -1433,6 +1453,38 @@ def test_str_gpu(self):
...
+    {'x_shape': (2, 2,), 'dtype': np.float32,
+     'repr': 'variable x([[0., 1.],\n            [2., 3.]])',
+     'str': 'variable x([[0. 1.]\n            [2. 3.]])'},
+    {'x_shape': (), 'dtype': np.float32,
+     'repr': 'variable x(0.)', 'str': 'variable x(0.)'},
+)
+@testing.with_requires('numpy>=1.14')
+class TestNamedVariableToStringModern(
+        _TestNamedVariableToStringBase, unittest.TestCase):
+    # Textual representation of arrays in NumPy 1.13 or earlier.
```
niboshi: `NumPy 1.14 or later`?

tests/chainer_tests/test_variable.py
kmaehashi: Actually I first did that, but pytest discovers classes starting with `Test` even if it does not inherit `unittest.TestCase`.
https://pytest.readthedocs.io/en/reorganize-docs/new-docs/user/naming_conventions.html

tests/chainer_tests/test_variable.py
kmaehashi: Fixed!

tests/chainer_tests/test_variable.py
niboshi: Oh, then, how about not using `Test` as prefix?

kmaehashi: OK, I fixed to `UnnamedVariableToStringTestBase`.
I found some tests using `*TestBase` class name.
https://github.com/chainer/chainer/blob/master/tests/chainer_tests/test_function_node.py#L502
https://github.com/chainer/chainer/blob/master/tests/chainer_tests/optimizers_tests/test_optimizers_by_linear_model.py#L82

niboshi:
jenkins, test this please

niboshi:
LGTM!

PR #4187, opened 2018-01-16
allow SerialIterator to use arbitrary random number generator

levelfour:
In some cases of distributed machine learning, it is needed to synchronize batch order of the iterator among multiple processes.
In order to achieve it, private random number generators are needed to avoid being interfered by other random number generation processes (e.g. dropout).
This PR alters `SerialIterator` to enable handling private RNG.
If no RNG is specified, it uses the numpy default RNG, which is compatible to the previous behavior.

For detail, please refer to chainer/chainermn#185.

### DIFF
--- a/chainer/iterators/serial_iterator.py
+++ b/chainer/iterators/serial_iterator.py
@@ -30,14 +30,23 @@ class SerialIterator(iterator.Iterator):
         shuffle (bool): If ``True``, the order of examples is shuffled at the
             beginning of each epoch. Otherwise, examples are extracted in the
             order of indexes.
+        rng: Random number generator (RNG) used for shuffling. If ``None``,
+            numpy default RNG is used.
 
     """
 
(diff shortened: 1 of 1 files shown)


toslunar:
The test failure (`TestL2Normalization_param_3.test_backward_cpu` in AppVeyor) seems irrelevant to this PR.

mitmul:
Is this related? https://github.com/chainer/chainer/pull/3429

Crissman:
For testing purposes, can you provide an example of a private RNG?

levelfour:
Thank you for the reply. Then I will reimplement after that the interface is fixed.

hvy:
Fyi, https://github.com/chainer/chainer/pull/3429 has now been merged. Maybe you can use the new `ShuffleOrderSampler` taking a rng as constructor argument?

hvy:
Talked offline with @levelfour and we agreed that we can close this PR. Please raise an issue if there are any issues using the `ShuffleOrderSampler`.

PR #3972, opened 2017-11-27
Show warning when Tensor-Core ready algo is not selected

anaruse:
This PR is related to https://github.com/chainer/chainer/issues/3922.

This PR makes it safer to select convolution algorithm of cuDNN for Tensor Core. Also, it shows warning messages like below when Tensor Core ready algorithms are not selected probably because amount of work-sapce for cuDNN is not large enough for the algorithms.

    /*/chainer/functions/connection/convolution_2d.py:206: RuntimeWarning: An algorithm(0) w/o Tensor Core support is selected. Tensor Core is not used for convolution with following configs: (n, c, h, w, kh, kw) = (32, 96, 27, 27, 5, 5). It might be because cuDNN workspace size is not large enough for Tensor Core (current size:16777216, required size:29600043).
    /*/chainer/functions/connection/deconvolution_2d.py:241: RuntimeWarning: An algorithm(0) w/o Tensor Core support is selected. Tensor Core not is used for convolution with following configs: (n, c, h, w, kh, kw) = (32, 96, 27, 27, 5, 5). It might be because cuDNN workspace size is not large enough for Tensor Core (current size:16777216, required size:22137259).
    /*/chainer/functions/connection/convolution_2d.py:360: RuntimeWarning: An algorithm(3) w/o Tensor Core support is selected. Tensor Core is not used for convolution with following configs: (n, c, h, w, kh, kw) = (32, 96, 27, 27, 5, 5). It might be because cuDNN workspace size is not large enough for Tensor Core (current size:16777216, required size:18883103).

PR #4189, opened 2018-01-16
Check cuDNN convolution algorithm

anaruse:
This PR adds methods to check whether Tensor Core enabled algorithm is selected or not when Tensor Core is available. If not, it shows warning messages like below, so that users can see which layer does not use Tensor Core.

    /*/chainer/functions/connection/deconvolution_2d.py:292: RuntimeWarning: Tensor Core mode is set but the selected convolution backward data algorithm(0) is not Tensor Core enabled algo. (xN,xC,xH,xW)=(128,512,28,28), (kH,kW)=(1,1), (yN,yC,yH,yW)=(128,1024,14,14). 

Please note that this PR is related to https://github.com/cupy/cupy/pull/890, and you should apply it first. Also this is related to https://github.com/chainer/chainer/pull/3972 that was closed.

### DIFF
--- a/chainer/functions/connection/convolution_2d.py
+++ b/chainer/functions/connection/convolution_2d.py
@@ -259,7 +259,9 @@ class Convolution2DFunction(function_node.FunctionNode):
                 conv_desc.value, y_desc.value, _fwd_pref, workspace_size)
 
         if use_tensor_core:
-            algo = self._tensor_core_adjust_algo()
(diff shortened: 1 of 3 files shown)


okuta:
I merged cupy/cupy#715. Cloud you fix this PR?

anaruse:
I close this PR, because now this fix can be done solely in cupy/cupy#890.

PR #4190, opened 2018-01-17
Fix backward of NormalizeL2

toslunar:
The backward computation of `NormalizeL2` was wrong (for non-negligible `eps`).


okuta:
LGTM!

PR #4191, opened 2018-01-17
Fix backward of BatchRenormalizationFunction and add tests

toslunar:
- It seems ` + self.eps` is missing in a line in `backward` of `BatchRenormalizationFunction`.
- At `Test(Fixed)BatchNormalization` and `Test(Fixed)BatchRenormalization`, take `eps` into account and add tests with larger `eps`.

I left as it is elsewhere but `@condition.retry(3)` might become redundant.


okuta:
Please resolve conflicts

okuta:
LGTM!

PR #4192, opened 2018-01-17
Fix document about links in chainer.functions

toslunar:
Links are not in `chainer.functions`.  (From v2?)

### DIFF
--- a/docs/source/reference/functions.rst
+++ b/docs/source/reference/functions.rst
@@ -10,7 +10,7 @@ functions, which users should use.
 .. note::
    As of v1.5, the concept of parameterized functions are gone, and they are
    replaced by corresponding :class:`~chainer.Link` implementations. They are
-   put in the :mod:`~chainer.links` namespace.
+   found in the :mod:`~chainer.links` namespace.
 
 ..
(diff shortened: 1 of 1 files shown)


niboshi:
I think this is not valid as a note as of v1.5 (because they were there at that moment).

How about somethink like this?
"... are replaced by corresponding Link implementations. They are put in the chainer.links namespace."

docs/source/reference/functions.rst
```diff
@@ -10,9 +10,7 @@ functions, which users should use.
 .. note::
    As of v1.5, the concept of parameterized functions are gone, and they are
    replaced by corresponding :class:`~chainer.Link` implementations. They are
-   still put in the :mod:`~chainer.functions` namespace for backward
-   compatibility, though it is strongly recommended to use them via the
-   :mod:`chainer.links` package.
+   put in the :mod:`~chainer.links` namespace.
```
Crissman: found in the :mod:`~chainer.links` namespace.

okuta:
@Crissman Can you merge this?

Crissman:
LGTM. Merging!

PR #4199, opened 2018-01-18
[RFC] Require "scalar" variables to be 0-dim

toslunar:
The current definition in `Variable.backward` of a "scalar" variable is an array containing 1 element.  This may cause unexpected behavior that only occurs when `batch_size=1`.  The PR proposes to fix the definition to being a 0-dimensional array.


mitmul:
We discussed this PR offline, and we agreed it's OK to merge. But we have a release tomorrow, so please give us more time until merge for ensuring it doesn't cause significant problems.

mitmul:
Jenkins, test this please

mitmul:
All tests passed. LGTM

PR #4223, opened 2018-01-24
Add Dockerfiles for intel/chainer

mitmul:


niboshi:
https://github.com/chainer/chainer/pull/4223/files#diff-7875ccd19cfb45549a123015e987ea60R59

We need to fix this after iDeep integration will be finished in Chainer.

mitmul:
@niboshi Now should I progress this PR?

niboshi:
@mitmul Yes, could you fix it?

mitmul:
@niboshi Yes, I fiexd. Now I'm checking those Dockerfilies can correctly work.

mitmul:
@niboshi How should I test these Dockerfile?

niboshi:
Hmm, perhaps to run test of some function (that supports ideep) and check if `_forward_ideep` is actually called?
Btw, don't we have to install ideep4py?

kmaehashi:
I think we just need a line of glog installation and `pip install ideep4py`.
I also think we can suspend this work as iDeep update that eliminates glog dependency and numpy version issue should be coming in near future.

niboshi:
Let's wait for the next release of iDeep.

niboshi:
@mitmul 
iDeep 1.0.3 has been released.
Can you update the installation script?

niboshi:
FYI: dockerfiles in intel/ideep
https://github.com/intel/ideep/tree/ideep4py/docker

kmaehashi:
I discussed with iDeep team in Chainer slack.

* Let's remove `centos` version of Dockerfile as it is quite similar to `ubuntu:16.04` one.
* They gave us the following simple version of Dockerfile. Let's use it.

```
FROM ubuntu:16.04
RUN apt-get update -y && \
   python python-pip python-dev
RUN pip install --upgrade pip setuptools
RUN pip install --default-timeout=10000 ideep4py chainer
```

I guess `--default-timeout=10000` is not needed, as download from PyPI to DockerHub infrastructure should be fast enough. I think we can remove this for now and append it later if needed (i.e. build failed on DockerHub).

niboshi:
>I guess `--default-timeout=10000` is not needed, as download from PyPI to DockerHub infrastructure should be fast enough.

Does this apply to user environments?
I think these dockerfiles are also intended for end users.

kmaehashi:
`--default-timeout` (equivalent to `--timeout`) is a timeout for the HTTP connection establishment , not timeout for whole download process. If we need that, we should append that to everywhere.

I think it is generally considered as internet connection issue (or pypi is down) if connection to pypi server failed for 15 seconds (the default value).

kmaehashi:
@mitmul Shall I takeover this? I'd like to include this to the next RC.

mitmul:
@kmaehashi Yes, please

kmaehashi:
Moved to #4476

Issue #3754, opened 2017-11-01
Serialize random states in iterators for better reproducibility


takagi:
Iterator implementations generate new order with `numpy.random` when they come across the end of a dataset and repeat over it.

However they currently do not serialize the random state but the new order after they repeat cannot be reproduced across resume/suspend.

Managing and serializing the random state in iterators will give us better reproducibility on that situation.

See #3575 .

grafi-tt:
@takagi Sorry for late reply! I agree with you, this would be useful feature. 

I have been reluctant to work on the feature because of the amount of work. It isn't very large work, but at least it should be implemented on `SerialIterator` as well as `MultiprocessIterator` in backward compatible manner with test codes, and some document update is very desirable.

Now I'm feeling I can spare time to the issue. Hopefully I'll send a PR until the mid January.

grafi-tt:
Serialization order sampler required. Related: #5042

grafi-tt:
bump

grafi-tt:
bump

PR #4230, opened 2018-01-24
Have RandomState instances unique to each batch index, can be used by e.g. random augmentation

grafi-tt:
I've made iterators store `numpy.random.RandomState` instances those are bound to batch indices. They can be retrieved by `chainer.iteartors.get_random_state()`.

The implementation is trivial for `SerialIterator`. For `MultithreadIterator`, a set of the backup states (implemented as tuples) is required to support `reset` and `serialize` correctly. For `MultiprocessIterator`, shared memory is necessary.

This feature is very useful when you use `MultiprocessIterator`. To perform random augmentation correctly with this iterator class, you need to set the random seed at the fetch of first batch, as the global random state remains the same on every forked processes.

Making the random seed deterministic is rather hard, because for each part of a batch to be fetched, assignment of a process that does fetch is completely non-deterministic. You may mitigate the situation by reseeding the random state at every fetch, but it hurts the long periodicity of Mersenne Twistter.

## overhead
If `chainer.iteartors.get_random_state()` is not called, there is almost no overhead. If called, overhead for `RandomState.get_state()` and/or `RandomState.set_state()` are incurred; but a benchmark suggests it's still negligible.

I performed the benchmark bellow on my Linux desktop machine, with i5-3550 (4 cores) processor. It just iterates over MNIST dataset with random flip augmentation. The size of batches is very small compared to most real workload, so the overhead of iteration is rate-limiting. Even on such a extreme setting, the measured performance impact is less than 10%.

```python
import time

from chainer.datasets import get_mnist, TransformDataset
from chainer.iterators import (get_random_state,
                               SerialIterator,
                               MultiprocessIterator,
                               MultithreadIterator)
import numpy


def random_flip_local(img):
    random = get_random_state()
    if random.uniform() >= 0.5:
        img = numpy.fliplr(img)
    return img


def random_flip_global(img):
    random = numpy.random
    if random.uniform() >= 0.5:
        img = numpy.fliplr(img)
    return img


N_PROC = 4
COUNT = 10000


def main():
    for it_cls in (SerialIterator, MultithreadIterator, MultiprocessIterator):
        for aug_meth in (random_flip_local, random_flip_global):
            dataset, _ = get_mnist(withlabel=False, ndim=2)
            dataset = TransformDataset(dataset, aug_meth)
            options = {}
            t1 = time.perf_counter()
            if it_cls == MultithreadIterator:
                options['n_threads'] = N_PROC
            if it_cls == MultiprocessIterator:
                options['n_processes'] = N_PROC
            it = it_cls(dataset, N_PROC, **options)
            for _ in range(COUNT):
                it.next()
            t2 = time.perf_counter()
            print("{:20} {:18} {:6.2f}us".format(it_cls.__name__, aug_meth.__name__,
                                            (t2 - t1) / COUNT * 1000 * 1000))


if __name__ == '__main__':
    main()
```

The result is:

```
SerialIterator       random_flip_local   52.33us
SerialIterator       random_flip_global  50.77us
MultithreadIterator  random_flip_local  315.76us
MultithreadIterator  random_flip_global 284.20us
MultiprocessIterator random_flip_local  527.36us
MultiprocessIterator random_flip_global 504.12us
```

As I've thought this issue is more important, so I made it prior to #3754.

### DIFF
--- a/chainer/iterators/__init__.py
+++ b/chainer/iterators/__init__.py
@@ -7,3 +7,5 @@ from chainer.iterators import serial_iterator  # NOQA
 from chainer.iterators.multiprocess_iterator import MultiprocessIterator  # NOQA
 from chainer.iterators.multithread_iterator import MultithreadIterator  # NOQA
 from chainer.iterators.serial_iterator import SerialIterator  # NOQA
+
+from chainer.iterators.random_state import get_random_state  # NOQA
--- a/chainer/iterators/multiprocess_iterator.py
+++ b/chainer/iterators/multiprocess_iterator.py
@@ -14,7 +14,6 @@ from chainer.dataset import iterator
 
 
 _response_time = 1.
-_short_time = 0.001
 _PrefetchState = namedtuple('_PrefetchState', (
     'current_position', 'epoch', 'is_new_epoch',
     'previous_epoch_detail', 'order'))
--- a/chainer/iterators/multiprocess_iterator.py
+++ b/chainer/iterators/multiprocess_iterator.py
@@ -1,5 +1,6 @@
 from __future__ import division
 from collections import namedtuple
+import ctypes
 import multiprocessing
 from multiprocessing import sharedctypes
 import signal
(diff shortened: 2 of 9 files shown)


delta2323:
Thank you for sending the PR! I think this PR consists of several parts.
- Add `iterators.get_random_state`
- Use `get_random_state` in `SerialIterator`
- Use `get_random_state` in `MultiProcessIterator`

Could you consider to split the PR into these parts and send them as PRs respectively?

Crissman:
@grafi-tt Is it possible to split this PR as mentioned by @delta2323?

Thanks!

delta2323:
> 1. implement iterators.get_random_state that can be called during iteration by user, and iterators.random_state.set_random_state that should be called by iterator implementation,

Looking at this sentence solely, we can work `get_random_state` and `set_random_state` in parallel. Isn't it possible?

delta2323:
One idea would to make a branch that implements 1 and 5 and a PR from the branch first. Then, create branches for 2, 3, and 4 respectively from the branch and send PRs (one for each branch). We'll review the PR for 1 and 5 first. It is also OK to separate 1 and 5, too.

delta2323:
Thank you! I'll take a look.

grafi-tt:
#4448

PR #4238, opened 2018-01-26
Fix unused argument(--out) of seq2seq example

okayu9:
Seq2seq example does not use the argument of "--out".
I fixed it to work the argument correctly.


niboshi:
jenkins, test this please

niboshi:
LGTM!
Thank you!

PR #4243, opened 2018-01-26
Skip unnecessary backward computation in matmul

anaruse:
This PR is to reduce unnecessary computation in `MatMul`.

When you specify one of inputs to matmul as `required_grad=False` like below, there is no need to compute the gradient of that input, but it is computed with current master. This PR reduces that unnecessary computation.

    a = Variable(a)
    b = Variable(b, required_grad=False)
    c = F.matmul(a, b)

I also noticed that `MatMulGrad` is not necessary if you add small fix to `MatMul`. Thus there is no `MatMulGrad` in this branch.

BTW, I found this issue when I was looking at Chainer chemistry.

### DIFF
--- a/chainer/functions/math/matmul.py
+++ b/chainer/functions/math/matmul.py
@@ -113,7 +113,7 @@ class MatMul(function_node.FunctionNode):
         else:
             y = _matmul(a, b, self.transa, self.transb, self.transc)
         if self.dtype is not None and y.dtype != self.dtype:
-            y = y.astype(self.dtype, copy=False)
+            y = y.astype(self.dtype)
         return utils.force_array(y),
 
     def backward(self, indexes, grad_outputs):
--- a/chainer/functions/math/matmul.py
+++ b/chainer/functions/math/matmul.py
@@ -70,9 +70,10 @@ def _get_check_index(trans, right, row_idx=0, col_idx=1):
 
 class MatMul(function_node.FunctionNode):
 
-    def __init__(self, transa=False, transb=False):
+    def __init__(self, transa=False, transb=False, transc=False):
         self.transa = transa
         self.transb = transb
+        self.transc = transc
 
     def check_type_forward(self, in_types):
         type_check.expect(in_types.size() == 2)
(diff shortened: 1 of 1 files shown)


chainer/functions/math/matmul.py
```diff
@@ -81,84 +82,54 @@ def check_type_forward(self, in_types):
...
+        gy, = grad_outputs
 
-    def __init__(self, transa=False, transb=False):
-        self.transa = transa
-        self.transb = transb
+        ga = None
+        if 0 in indexes:
+            ga, = MatMul(self.transc, not self.transb,
+                         self.transa).apply((gy, b))
+            ga.data = ga.data.astype(a.dtype)
```
okuta: Pelase use `chainer.functions.cast`.

chainer/functions/math/matmul.py
```diff
@@ -81,84 +82,54 @@ def check_type_forward(self, in_types):
...
-    def forward(self, inputs):
-        self.retain_inputs((0, 1, 2))
-        a, b, gy = inputs
-        a_shape = a.shape
-        b_shape = b.shape
+        gb = None
+        if 1 in indexes:
+            gb, = MatMul(not self.transa, self.transc,
+                         self.transb).apply((a, gy))
+            gb.data = gb.data.astype(b.dtype)
```
okuta: Pelase use `chainer.functions.cast`.

chainer/functions/math/matmul.py
anaruse: Well, I think we don't need to use a function for data type conversion here.

If we use `chainer.functions.cast`, inverse data type conversion is to be done by backward() of cast, when you double-backprop it. However, I think `MatMul` handles that data conversion in it, so I don't think we need to use `chainer.function.cast` here.

What would you think on this?

chainer/functions/math/matmul.py
anaruse: (The same above)

chainer/functions/math/matmul.py
anaruse: Assuming there is a network that computes `c` = MatMul(`a`, `b`) in it and dtype of `a`, `b` and `c` are fp16, fp32 and fp32, respectively (I'm not sure whether this combination of data type is practical or not...). When you backprop it, `ga` will be computed by something like `ga` = MatMul(`gc`, `b`). A data type of `ga` becomes fp32, because data type of `gc` and `b` are fp32. However data type of `a` and `ga` must be the same. That's why data type conversion for `ga` (from fp32 to fp16) is necessary here.

Here is the main issue. If you use a function (`chainer.functions.cast`) for this data type conversion, what would happen when you double-backprop it?  `ggc` will be computed by something like `ggc` = MatMul(`gga`, `b`). I think the MatMul for double-backprop should handle the same combination of inputs in terms of data type as the MatMul for forward, say `c` = MatMul(`a`, `b`). However, data type of `gga` gets fp32 thanks to backward() of `chainer.functins.cast`. If we don't use a function here, `gga` becomes fp16.  I think is is better in consistency point of view.

What would you think?

okuta: OK, how about introduce `dtype` option to `MatMul` function?
And, `astype` has a `copy` option. Please set `copy=False`.

anaruse: That's good idea! will fix in that way.

chainer/functions/math/matmul.py
```diff
@@ -111,6 +112,8 @@ def forward(self, x):
             y = a * b
         else:
             y = _matmul(a, b, self.transa, self.transb, self.transc)
+        if self.dtype is not None and y.dtype != self.dtype:
+            y = y.astype(self.dtype, copy=False)
```
okuta: This case do not need `copy=False` option.

anaruse: Ah..., thank you for pointing it out! will remote the option.

okuta:
jenkins, test this please.

okuta:
LGTM!

PR #4246, opened 2018-01-26
Fix CUDNN_STATUS_EXECUTION_FAILED when NStepBiLSTM is using CuPy

jxpxxzj:
# Fix CUDNN_STATUS_EXECUTION_FAILED
## Issue description
Model definition:
```
# ...
self.lstm = L.NStepBiLSTM(n_layers=3, in_size=96, out_size=128, dropout=0.3)
# ...
```

In [n_step_rnn.py](https://github.com/chainer/chainer/blob/5d4673bcc3589ee07d0185571629675581740a31/chainer/links/connection/n_step_rnn.py#L15) 
Line 15:
```
    return numpy.argsort([-len(x.data) for x in lst]).astype('i')
```
When using CuPy, the type of `x.data` is `cupy.cuda.MemoryPointer`, which doesn't have `__len__()`
However, the problem will not appear when using CPU only, because the type of `x.data` is `numpy.ndarray`, which could be passed to `len()` by invoking `__len__()`.

## Fixing

While the type of `x` is `chainer.Variable`, and has `__len__()`, so it should be changed to `len(x)` to cover both CPU and GPU.

### DIFF
--- a/chainer/links/connection/n_step_rnn.py
+++ b/chainer/links/connection/n_step_rnn.py
@@ -12,7 +12,7 @@ from chainer import variable
 
 
 def argsort_list_descent(lst):
-    return numpy.argsort([-len(x.data) for x in lst]).astype('i')
+    return numpy.argsort([-len(x) for x in lst]).astype('i')
 
 
 def permutate_list(lst, indices, inv):


delta2323:
Thank you for reporting the issue and sending the PR! It seems to me that the problem is not only in `argsort_list_descent` but also in how the function is used in `NStepLSTM`, as if `argsort_list_descent` is designed to be CPU/GPU agnostic, `x.data` should be `cupy.ndarray` when using GPU if it is `numpy.ndarray` when CPU. So, it is better to look at `NStemLSTM`

delta2323:
Is it possible for you to give me a complete minimal reproducible code?

jxpxxzj:
Sorry for lately reply. I am busy in fininshing the other parts of my project, and the source code was edited because I didn't use any version control system.

I detected the exception through the traceback it provided, then I attached a debugger and fixed it by editing the code as the pull request provided. 
However, the problem disappeared after I refactoried my code and updated the chainer I installed. In my early test, I set the` batch_size` to 1, which maybe the source of the exception if only pass a single sequence to the link. 

You can leave this pull request or just close it, I would reopen or create a issue If I find a proper way to that situation. Anyway this change seems doesn't break my training process, and the `len(x.data)` should have the same semantic as `chainer.Variable.__len__` so I kept it on my own computer.

delta2323:
Thank you for providing us with your situation. 

> In my early test, I set thebatch_size to 1, which maybe the source of the exception if only pass a single sequence to the link.

It seems to me to worth investigating. I'll take a look.

delta2323:
I tried a minibatch with size 1 with the following code, but could not reproduce your error:

```
import chainer
import numpy as np
import cupy as cp

gpu = True  # False for CPU                                                                                        
xp = cp if gpu else np

xs = xp.random.uniform(-1, 1, (1, 10, 96)).astype(np.float32)
xs = [chainer.Variable(x) for x in xs]

l = L.NStepBiLSTM(n_layers=3, in_size=96, out_size=128, dropout=0.3)
if gpu:
    l.to_gpu()

l(None, None, xs)
```

delta2323:
@jxpxxzj  I close this issue and will open when I can reproduce the error. You can also open the issue when you find out the reason. Thank you for reporting the issue!

delta2323:
@jxpxxzj  Sorry, before closing the issue. I would like to confirm if my understanding is correct.

jxpxxzj:
Yes, you just close it since I can't reproduce it.

PR #4247, opened 2018-01-27
Update examples and tests for chainer.backends.cuda

vilyaair:
Update examples and tests in accordance with #3974 and #3981


vilyaair:
Sorry, I should use autopep8 to correct my code.
Fix some docstrings that is not updated in #3981

PR #4249, opened 2018-01-29
Use class instead of list to represent variable data

takagi:
Currently a list is used to represent `Variable`'s data with its initialized/uninitialized state.

The change was introduced in #2072 and a list was chosen there because a list is one of the most primitive data structure in Python, but it is a bit confusing and hard for readers to understand because it is not typical use of a list.

In this PR, I replace the list with a class to make it easier to understand for readers.


okuta:
Did you check performance?

takagi:
No.

takagi:
I close this PR.

PR #4253, opened 2018-01-30
Tensordot support

anaruse:
This PR is to support tensordot as Chainer's function. You can use this function like numpy tensordot as below. Off course, it is backprop-able.

    import chainer
    import numpy
    a = chainer.Variable(numpy.random.rand(4,3,2))
    b = chainer.Variable(numpy.random.rand(3,2,5))
    c = chainer.functions.tensordot(a, b, axes=((1,2), (0,1)))

### DIFF
--- a/chainer/functions/math/tensordot.py
+++ b/chainer/functions/math/tensordot.py
@@ -1,0 +1,128 @@
+import collections
+import numpy
+import six
+
+from chainer.backends import cuda
+from chainer import function_node
+import chainer.functions
+from chainer import utils
+from chainer.utils import type_check
+
+import inspect
+import os
+
+
+def _loc():
+    frame = inspect.currentframe().f_back
+    ret = '%s, %s, %s' % (os.path.basename(frame.f_code.co_filename),
+                          frame.f_code.co_name, frame.f_lineno)
+    return ret
+
+
+def _tensordot(a, b, a_axes, b_axes, c_axes):
+    
+    xp = cuda.get_array_module(a)
+    axes = (a_axes[1], b_axes[0])
+    y = xp.tensordot(a, b, axes)
+
+    trans = [None for i in range(y.ndim)]
+    
+    table_a = [1 if i in a_axes[0] else 0 for i in range(a.ndim)]
+    table_a = numpy.cumsum(table_a) - 1
+    print('# {}, table_a:{}'.format(_loc(), table_a))
+    for i, dst in enumerate(c_axes[0]):
+        j = a_axes[0][i]
+        j = table_a[j]
+        trans[dst] = j
+
+    table_b = [1 if i in b_axes[1] else 0 for i in range(b.ndim)]
+    table_b = numpy.cumsum(table_b) - 1
+    print('# {}, table_b:{}'.format(_loc(), table_b))
+    for i, dst in enumerate(c_axes[1]):
+        j = b_axes[1][i]
+        j = table_b[j]
+        trans[dst] = j
+        trans[dst] += len(a_axes[0])
+
+    print('# {}, trans:{}'.format(_loc(), trans))
+        
+    do_transpose = False
+    for i in range(y.ndim):
+        if trans[i] != i:
+            do_transpose = True
+    if do_transpose:
+        print('# {}, y.shape(befor):{}'.format(_loc(), y.shape))
+        y = xp.transpose(y, trans)
+        print('# {}, y.shape(after):{}'.format(_loc(), y.shape))
+    
+    return y
+
+
+class TensorDot(function_node.FunctionNode):
+
+    def __init__(self, axes=2, a_axes=None, b_axes=None, c_axes=None):
+        self.axes = axes
+        self.a_axes = a_axes
+        self.b_axes = b_axes
+        self.c_axes = c_axes
+
+    def forward(self, inputs):
+        self.retain_inputs((0, 1))
+        a, b = inputs
+
+        if self.a_axes is None or self.b_axes is None:
+            a_axes = [None, None]
+            b_axes = [None, None]
+            axes = self.axes
+            if isinstance(axes, collections.Sequence):
+                if len(axes) != 2:
+                    raise ValueError('Axes must consist of two arrays.')
+                a_axes[1], b_axes[0] = axes
+                if numpy.isscalar(a_axes[1]):
+                    a_axes[1] = a_axes[1],
+                if numpy.isscalar(b_axes[0]):
+                    b_axes[0] = b_axes[0],
+            else:
+                a_axes[1] = list(six.moves.range(a.ndim - axes, a.ndim))
+                b_axes[0] = list(six.moves.range(axes))
+            a_axes[0] = [i for i in six.moves.range(a.ndim) if i not in a_axes[1]]
+            b_axes[1] = [i for i in six.moves.range(b.ndim) if i not in b_axes[0]]
+            self.a_axes = a_axes
+            self.b_axes = b_axes
+
+        if self.c_axes is None:
+            c_axes = [None, None]
+            c_axes[0] = list(six.moves.range(len(a_axes[0])))
+            c_axes[1] = list(six.moves.range(len(a_axes[0]), len(a_axes[0]) + len(b_axes[1])))
+            self.c_axes = c_axes
+        
+        print('# {}, axes:{}'.format(_loc(), self.axes))
+        print('# {}. a.shape:{}'.format(_loc(), a.shape))
+        print('# {}, a_axes: {}'.format(_loc(), self.a_axes))
+        print('# {}. b.shape:{}'.format(_loc(), b.shape))
+        print('# {}, b_axes: {}'.format(_loc(), self.b_axes))
+        print('# {}, c_axes: {}'.format(_loc(), self.c_axes))
+
+        y = _tensordot(a, b, self.a_axes, self.b_axes, self.c_axes)
+        return utils.force_array(y),
+
+    def backward(self, indexes, grad_outputs):
+        a, b = self.get_retained_inputs()
+        gy, = grad_outputs
+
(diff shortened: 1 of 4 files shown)


hvy:
Sorry for my late response, and thank you for this PR! It looks like tests are already in-place but were there uncovered cases that you were aware of? Also, it'd be great if you could add documentations for the implemented FunctionNode as discussed earlier and I think we can proceed.

anaruse:
@hvy: Thank you for your response.

I've just added docstrings to `tensordot` (I borrowed most sentences from cupy.tensordot).
Regarding tests, there are tests already and I'd like you to check if the coverage is enough or not.
Anyway, please review it!

hvy:
@anaruse Thank you for the clarification. I'll check accordingly!

hvy:
jenkins, test this please.

hvy:
Could you add reference to the docs (https://github.com/anaruse/chainer/blob/tensordot_support/docs/source/reference/functions.rst) so that this function is listed among the other ones?

anaruse:
Thank you for your comment. I've added the reference to the docs.

hvy:
jenkins, test this please.

chainer/functions/math/tensordot.py
```diff
@@ -0,0 +1,171 @@
...
+        if self.c_axes is None:
+            c_axes = [[], []]  # 0:row axes, 1:col axes
+            c_row_ndim = len(self.a_axes[0])
+            c_col_ndim = len(self.b_axes[1])
+            c_axes[0] = six.moves.range(c_row_ndim)
+            c_axes[1] = six.moves.range(c_row_ndim, c_row_ndim + c_col_ndim)
+            self.c_axes = c_axes
+
+        c = _tensordot(a, b, self.a_axes, self.b_axes, self.c_axes)
+        if self.dtype is not None and c.dtype != self.dtype:
```
hvy: It is possible to skip this conditioning and let `utils.force_array(c, self.dtype)` take care of it?

chainer/functions/math/tensordot.py
```diff
@@ -0,0 +1,171 @@
...
+        self.c_axes = c_axes
+        self.dtype = dtype
+
+        if isinstance(axes, collections.Sequence):
+            if len(axes) != 2:
+                raise ValueError('axes must consist of two arrays.')
+        elif isinstance(axes, int):
+            pass
+        else:
+            raise TypeError('axes must be int or tuple of int')
```
hvy: What do you think of making the error message slightly more verbose?

tests/chainer_tests/functions_tests/math_tests/test_tensordot.py
```diff
@@ -0,0 +1,141 @@
...
+        self.gga = numpy.random.uniform(
+            .5, 1, self.a_shape).astype(self.a_dtype)
+        self.ggb = numpy.random.uniform(
+            .5, 1, self.b_shape).astype(self.b_dtype)
+
+        self.op = lambda a, b: F.tensordot(a, b, axes=self.axes)
+        self.forward_answer = self._get_forward_answer(self.a, self.b,
+                                                       self.axes)
+
+    def _get_forward_answer(self, a, b, axes):
```
hvy: I see that this method is aligned to other tests but I don't think it's needed for tensor dot (since the variable name is descriptive and there are no branches).

hvy:
As for tests, how about covering the following cases.

- Shapes with ndim 0, 1, and 2 (with axes that are valid for corresponding shapes. Maybe ndim 0 doesn't have to be supported?)
- 'axes:' 0
- default axes (maybe this isn't needed)
- Invalid axes, e.g. negative value or larger tuples

chainer/functions/math/tensordot.py
```diff
@@ -0,0 +1,171 @@
...
+
+    def check_type_forward(self, in_types):
+        type_check.expect(in_types.size() == 2)
+        a_type, b_type = in_types
+
+        type_check.expect(
+            a_type.dtype.kind == 'f',
+            b_type.dtype.kind == 'f',
+            a_type.ndim >= 1,
+            b_type.ndim >= 1,
```
hvy: Is ndim of 0 too trivial? Since numpy/cupy support it, I think I'd be nice if Chainer could too.

hvy:
It LGTM except some comments. Could you please take a look?

chainer/functions/math/tensordot.py
anaruse: Thank you for your suggestion. I didn't know that `utils.force_array` has `dtype` option.

chainer/functions/math/tensordot.py
anaruse: Agree. will check if it works well when ndim is zero.

anaruse: I'm wondering how to handle an input with ndim=0, mainly because `chainer.Variable` does not accepts a scalar as its input. If you make it ndarray with shape of (1), then it becomes ndim=1.
Well, do we really need to consider the case of ndim=0?

anaruse:
Thank you for your detailed feedback, @hvy!

I've fixed the branch based on your suggestion and added several tests for wider coverage (except ndim=0 case).
Please review again!

chainer/functions/math/tensordot.py
hvy: Ah, I think you can create a non-scalar ndarray with ndim=0 using e.g. 

```
x_data = numpy.random.uniform(-1, 1, ()))
x = chainer.Variable(x)
```

I am not sure how important this feature is but I noticed that e.g. matmul does support this case. Does it require a lot of changes to the code? https://github.com/chainer/chainer/blob/master/chainer/functions/math/matmul.py#L83

hvy:
Thanks for the quick response. Could you please take another look at my comment?

chainer/functions/math/tensordot.py
anaruse: Thank you!  I didn't know how to make a ndarray with ndim=0. I don't think it requires a lot of changes to support it. will fix soon.

anaruse:
Thank you for your feedback. I've just fixed the branch. Now it support a ndarray with ndim=0.

anaruse:
I've just updated the branch so that it skips processing for transposing output tensor when it is obviously not necessary.

chainer/functions/math/tensordot.py
```diff
@@ -0,0 +1,175 @@
...
+    if a.ndim < a_col_ndim or b.ndim < b_row_ndim:
+        raise ValueError('dimension of input tensors must be '
+                         'greater equal to dot-axes count ({})'
+                         .format(a_col_ndim))
+    for a_axis, b_axis in zip(a_axes[1], b_axes[0]):
+        if a.shape[a_axis] != b.shape[b_axis]:
+            raise ValueError('shape mismatch')
+
+    if a.ndim == 0 or b.ndim == 0:
+        y = a * b
```
hvy: Is this for performance reasons or does `xp.tensordot` in this case not work?

tests/chainer_tests/functions_tests/math_tests/test_tensordot.py
```diff
@@ -0,0 +1,183 @@
...
+
+        {'a_shape': (3, 2), 'b_shape': (2, 4), 'axes': 1, 'gc_shape': (3, 4)},  # NOQA
+        {'a_shape': (3, 2), 'b_shape': (2, 4), 'axes': (1, 0), 'gc_shape': (3, 4)},  # NOQA
+        {'a_shape': (3, 2), 'b_shape': (4, 2), 'axes': (1, 1), 'gc_shape': (3, 4)},  # NOQA
+        {'a_shape': (2, 3), 'b_shape': (4, 2), 'axes': (0, 1), 'gc_shape': (3, 4)},  # NOQA
+        {'a_shape': (2, 3), 'b_shape': (2, 4), 'axes': (0, 0), 'gc_shape': (3, 4)},  # NOQA
+
+        {'a_shape': (2), 'b_shape': (3), 'axes': 0, 'gc_shape': (2, 3)},  # NOQA
+        {'a_shape': (2, 3), 'b_shape': (), 'axes': 0, 'gc_shape': (2, 3)},  # NOQA
+        {'a_shape': (), 'b_shape': (2, 3), 'axes': 0, 'gc_shape': (2, 3)},  # NOQA
```
hvy: How about testing `{'a_shape': (), 'b_shape': (),...},` too.

chainer/functions/math/tensordot.py
```diff
@@ -0,0 +1,175 @@
...
+import six
+
+from chainer.backends import cuda
+from chainer import function_node
+from chainer import utils
+from chainer.utils import type_check
+
+
+def _tensordot(a, b, a_axes, b_axes, c_axes=None):
+
```
hvy: Quite a nitpick but I think we can remove this empty line!

hvy:
Sorry for my delayed response. I added some small comments. Do you mind taking another look?

chainer/functions/math/tensordot.py
anaruse: I thought `xp.tensordot` did not work when ndim=0, but seems it was because `list` was used for `axes` option at that time. It works OK if tuple is used. will use `xp.tensordot` even when ndim=0.

tests/chainer_tests/functions_tests/math_tests/test_tensordot.py
anaruse: OK, will add that test too.

chainer/functions/math/tensordot.py
anaruse: All right, will remote that empty line :)

anaruse:
Thank you for your feedback. I've just updated the branch based on your feedback. Please review again!

hvy:
jenkins, test this please.

hvy:
LGTM! Thank you again for this PR and all the quick fixes!

PR #4260, opened 2018-01-30
use variable.array instead of variable.data in reporter.py

crcrpar:
This PR is trivial, but from v3, using variable.data is discouraged as noted [here](https://github.com/chainer/chainer/blob/master/chainer/variable.py#L624).

### DIFF
--- a/chainer/reporter.py
+++ b/chainer/reporter.py
@@ -198,7 +198,7 @@ def report(values, observer=None):
                   # This chain just computes the mean absolute and squared
                   # errors between the prediction and y.
                   pred = self.predictor(x)
-                  abs_error = F.sum(F.abs(pred - y)) / len(x.array)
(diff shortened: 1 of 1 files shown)


chainer/reporter.py
```diff
@@ -198,7 +198,7 @@ def __call__(self, x, y):
                   # This chain just computes the mean absolute and squared
                   # errors between the prediction and y.
                   pred = self.predictor(x)
-                  abs_error = F.sum(F.abs(pred - y)) / len(x.data)
+                  abs_error = F.sum(F.abs(pred - y)) / len(x.array)
```
okuta: Please use `len(x)` .

crcrpar: OK, thank you.

okuta:
jenkins, test this please.

okuta:
Thanks!
LGTM

PR #4261, opened 2018-01-31
Add cudnn mode selection in softmax

anaruse:
This PR aims to improve performance of softmax in non-convolution cases.

cuDNN has two modes for softmax, CUDNN_SOFTMAX_MODE_CHANNEL and CUDNN_SOFTMAX_MODE_INSTANCE (please see [here](https://docs.nvidia.com/deeplearning/sdk/cudnn-developer-guide/index.html#cudnnSoftmaxMode_t)). For now, CUDNN_SOFTMAX_MODE_CHANNEL is always used in Chainer regardless of shape of input tensor and target axis. It is basically good for CNN, but it could be slow depending of shape of tensor. Actually, it is very slow when (N,C,H,W) = (many, some, 1, 1).

This PR adds feature to select cudnn mode based on shape of input tensor. With this PR,  CUDNN_SOFTMAX_MODE_INSTANCE is selected only when (H,W) = (1,1) and CUDNN_SOFTMAX_MODE_CHANNEL is used as default.

### DIFF
--- a/chainer/functions/activation/softmax.py
+++ b/chainer/functions/activation/softmax.py
@@ -10,7 +10,8 @@ if cuda.cudnn_enabled:
     cudnn = cuda.cudnn
     libcudnn = cuda.cuda.cudnn
     _algorithm = libcudnn.CUDNN_SOFTMAX_ACCURATE
-    _mode = libcudnn.CUDNN_SOFTMAX_MODE_CHANNEL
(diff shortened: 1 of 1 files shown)


niboshi:
jenkins, test this please

niboshi:
LGTM!

PR #4262, opened 2018-01-31
Add Swish activation function.

mizuno-gsinet:
This PR implements [Swish activation function](https://arxiv.org/abs/1710.05941) and the corresponding Link with trainable parameter `beta`.

I wrote this for self-education. Any comments or suggestions would be appreciated.

### DIFF
--- a/chainer/functions/__init__.py
+++ b/chainer/functions/__init__.py
@@ -31,6 +31,8 @@ from chainer.functions.activation.softmax import softmax  # NOQA
 from chainer.functions.activation.softmax import Softmax  # NOQA
 from chainer.functions.activation.softplus import softplus  # NOQA
 from chainer.functions.activation.softplus import Softplus  # NOQA
+from chainer.functions.activation.swish import swish  # NOQA
+from chainer.functions.activation.swish import Swish  # NOQA
 from chainer.functions.activation.tanh import tanh  # NOQA
 from chainer.functions.activation.tanh import Tanh  # NOQA
 from chainer.functions.activation.tree_lstm import tree_lstm  # NOQA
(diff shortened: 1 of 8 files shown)


bkvogel:
Thank you for the PR! The paper seems to claim this is slightly better than existing activation functions. Have you tried using it in any networks? How did it perform compared to an existing activation function such as ReLU?

mizuno-gsinet:
To be honest, I have only tested it with a toy regression problem. In the experiment, Swish obviously outperformed ReLU, Tanh, etc. But I am not sure if it works with real-world problems.

bkvogel:
Thank you for answering. I don't think it is necessary to try it on real world problems but if you have time, you might find it interesting to experiment with running the Chainer MNIST and CIFAR examples using Swish instead of ReLU to see how it compares.

mizuno-gsinet:
Thank you for your review! As you suggested, I added `None` option for `beta_shape` and updated docstring. I think the paper also suggests training beta per unit.

To compare the performance, I added modified versions of the CIFAR example, in which all ReLUs are replaced by Swish ([`add-swish-bench`](https://github.com/mizuno-gsinet/chainer/tree/add-swish-bench)). However, my machine is so powerless that I could only run them for a few epochs. The results are as follows.

```bash
# ReLU, took ~70s per epoch.
python3 train_cifar.py --gpu 0 --dataset cifar10
# Swish(None), took ~75s per epoch.
python3 train_cifar_swish.py --gpu 0 --dataset cifar10
# Swish(()), took ~80s per epoch.
python3 train_cifar_swish0.py --gpu 0 --dataset cifar10
```

So far, I dont't think it's unreasonably slow compared to ReLU...

bkvogel:
jenkins, test this please.

bkvogel:
Thank you for making the changes and trying it on CIFAR. LGTM!

PR #4266, opened 2018-02-01
Add axis option to batch normalization

anaruse:
This PR aims to make an implementation of batch normalization in Chainer bit more general-purpose and make it available for graph convolution network.

A typical shape of activation tensors in graph convolution is (B, V, F), where B, V and F are number of mini-batch, number of vertexes and number of features/channels respectively (AFAIK). If you want to apply batch-norm to it, normalization should be performed over B+V dimensions. This can be thought as considering V dimension as kind of second mini-batch dimension. Current batch-norm implementation in Chainer assumes that mini-batch dimensionality is one and always in first axis. Thus you need to call `F.reshape` before and after batch-norm like below.

    bnorm = L.BatchNormalization(F)

    h = F.reshape(h, (B*V, F))
    h = bnorm(h)
    h = F.reshape(h, (B, V, F))

However, this will require additional time, mainly because some processing for maintaining computational graph is necessary for `F.reshape`. To mitigate this issue, I would propose to add `axis` option to batch normalization. With this option, you can specify over which dimensions/axes normalization is to be performed. The above case can be described as below with this PR.

    bnorm = L.BatchNormalization(F, axis=(0,1))

    h = bnorm(h)

My motivation behind this PR is to use batch-norm for graph convolution network, but I suppose this is beneficial for other use cases. Any feedback would be appreciated.


delta2323:
Thank you for sending the PR. We are discussing to add a channel-size inference feature to Batch Normalization in #4149, so that we can specify `None` or some thing like `(1, None, 2)` to `size`. As both PRs changes the API of BN, could you take a look at the PR, although I think these two PRs can be done independently?

anaruse:
Thank you for letting me know #4149.

I looked through #4149. I also think the two PRs are basically independent. However I have one concern that you don't need to consider the usage like `size = (None, None)` with #4149 anymore, once this PR is accepted. That is because you can infer a shape of parameters like beta and gamma from a shape of input tensor and the option `axis`.

toslunar:
With this PR, it becomes more reasonable choice to add only `size=None` in #4149.  Nonetheless `size=(None, None)` allows one to assert `len(self.key_axis) == 2`.

chainer/functions/normalization/batch_normalization.py
```diff
@@ -31,20 +32,34 @@ def __init__(self, eps=2e-5, mean=None, var=None, decay=0.9):
...
             gamma_type.shape == beta_type.shape,
         )
+        if self.axis is None:
+            type_check.expect(
+                x_type.ndim >= gamma_type.ndim + 1,
+                x_type.shape[1:1 + M] == gamma_type.shape,
+            )
+        else:
+            type_check.expect(
+                x_type.ndim > len(self.axis),
```
toslunar: Could you `type_check` the shape?
I'd add a method like `_compute_key_axis` that will be called from either `check_type_forward` or `forward`.

anaruse: Thank you for your suggestion!  I think adding a method like `_compute_key_axis` is good idea.

anaruse:
@toslunar: Thank you for your comment and suggestion. The branch was revised in the way you suggested. Please review again.

chainer/functions/normalization/batch_normalization.py
```diff
@@ -53,14 +83,21 @@ def forward(self, inputs):
         if self.running_mean is None:
             self.running_mean = xp.zeros_like(gamma)
             self.running_var = xp.zeros_like(gamma)
-        self.mode = _BNMode(x, gamma)
+
+        self.axis = _compute_axis(x.ndim, gamma.ndim, self.axis)
+        self.key_axis = _compute_key_axis(x.ndim, gamma.ndim, self.axis)
+        for i, j in enumerate(self.key_axis):
+            if gamma.shape[i] != x.shape[j]:
```
delta2323: It seems to me that this condition is already checked in `check_type_forward`.

chainer/functions/normalization/batch_normalization.py
```diff
@@ -420,25 +486,26 @@ def backward(self, inputs, grad_outputs):
...
         self.cudnn_dtype_ok = self.is_for_conv2d or (x.dtype != numpy.float16)
 
     def get_cudnn_mode(self):
         assert self.cudnn_dim_ok
-        if self.is_for_conv2d:
-            return libcudnn.CUDNN_BATCHNORM_SPATIAL
-        return libcudnn.CUDNN_BATCHNORM_PER_ACTIVATION
+        # if self.is_for_conv2d:
+        #     return libcudnn.CUDNN_BATCHNORM_SPATIAL
+        # return libcudnn.CUDNN_BATCHNORM_PER_ACTIVATION
```
delta2323: Could you remove code that is no longer used instead of leaving it as a comment?

chainer/functions/normalization/batch_normalization.py
```diff
@@ -567,6 +635,14 @@ def batch_normalization(x, gamma, beta, **kwargs):
             be ``None``.
         decay (float): Decay rate of moving average. It is used during
             training.
+        axis (int or tuple of int): Axis over which normalization is
+            performed. When axis is None, it is determined from input
```
delta2323: ``None`` (use double quotes).  The same is true of other programming expressions.

delta2323:
I noticed that this change breaks the existing behavior of `F.batch_normalization` whose document says channel axes must be contiguous, as pointed as in [the comment](https://github.com/chainer/chainer/pull/4149#issuecomment-357587350) in #4149 . That is why we do not specify which axes are channels with tuple in the PR #4149, as we do in this PR.

Anyway, I would suggest we make a consensus on specification of `F.BatchNormalization` and `L.BatchNormalization` first. As most people relevant seems agree on the former except the no-compatibility problem I mentioned in the previous section, I would like to discuss on the latter.

delta2323:
> With this PR, it becomes more reasonable choice to add only size=None in #4149.

I want to confirm my understanding of [@toslunar 's comment](https://github.com/chainer/chainer/pull/4266#issuecomment-363324270). Are you thinking to add `axis` option to `L.BatchNormalization` as well to specify which axes we treat as batch dimentions?

anaruse:
Thank you for your detailed review, @delta2323. As you suggested, I will wait to reflect you comment to the branch until we get aligned on compatibility concern.

BTW, I don't think this is severe compatibility breakage because there is no behavior change as long as `axis` option is not used. In other words, existing models will work as before.

Crissman:
@delta2323 Please confirm compatibility ok, since existing implementations will work.

delta2323:
Sorry for late reply. I reconsidered carefully and understood that this PR does not break compatibility and [this comment of mine](https://github.com/chainer/chainer/pull/4266#issuecomment-364628931) was wrong.

anaruse:
Thank you for your comments @delta2323.  I've fixed several issues that you pointed out and added tests for axis option.

delta2323:
Thank you! I'll take a look

delta2323:
@toslunar Could you take a look at and double-check this PR?

chainer/functions/normalization/batch_normalization.py
```diff
@@ -70,17 +103,21 @@ def forward(self, inputs):
...
         # TODO(niboshi): Refactor calculation of expander and axis into a
         # function and call it just before they are used.
 
         # expander inserts singleton dimensions to gamma and beta so that they
         # can be broadcasted with x.
-        head_ndim = gamma.ndim + 1
-        expander = (None, Ellipsis) + (None,) * (x.ndim - head_ndim)
+        expander = [None for _ in range(x.ndim)]
+        for i, j in enumerate(self.key_axis):
+            expander[j] = slice(gamma.shape[i])
```
toslunar: `slice(None)` works.

chainer/functions/normalization/batch_normalization.py
```diff
@@ -279,7 +318,9 @@ def forward(self, inputs):
             gy = cuda.cupy.ascontiguousarray(gy)
             dtype = x.dtype
             handle = cudnn.get_handle()
-            x_desc = cudnn.create_tensor_descriptor(_as4darray(x))
+            x_desc = cudnn.create_tensor_descriptor(
+                _as4darray(x, self.key_axis))
+            cudnn_mode = self.mode.get_cudnn_mode()
```
toslunar: `cudnn_mode` is available 8 lines above.  Did you intend to move the line here?

tests/chainer_tests/functions_tests/normalization_tests/test_batch_normalization.py
```diff
@@ -164,6 +164,157 @@ def test_double_backward(self, backend_config):
             backend_config)
 
 
+@testing.parameterize(*testing.product_dict(
```
toslunar: Could you merge these tests into `TestBatchNormalization`?
How about `if self.axis is None`, after concatenating the parameter sets and adding `'axis': None` to the older sets.

chainer/functions/normalization/batch_normalization.py
anaruse: Thank you. I didn't know that it works.

chainer/functions/normalization/batch_normalization.py
anaruse: I will remove the one at line 315. Probably, this was back when I merged the main branch to resolve lots of conflicts. Yes, I intentionally moved it here, so that cudnn_mode is obtained at the same place as the others.

tests/chainer_tests/functions_tests/normalization_tests/test_batch_normalization.py
anaruse: Do you mean I should support `axis` option for ideep? I have no idea how to do that and I'm not appropriate person to do that.

anaruse:
Thanks for your feedback, @toslunar .  I've updated the branch based on your feedback except tests. Could you check it again?

tests/chainer_tests/functions_tests/normalization_tests/test_batch_normalization.py
toslunar: No.  My concern was just that most of the lines of `TestBatchNormalizationAxis` are copied from `TestBatchNormalization`.  I'm willing to work on this issue as another PR.

toslunar:
Jenkins, test this please.

PR #3402, opened 2017-09-15
Fix crash bug when a child is reporting a value when using MultiprocessParallelUpdater

tkerola:
Fixes #3389 
Without the suggested change, Chainer will crash with error message `KeyError: 'Given observer is not registered to the reporter.'` when reporting values from a child that resides on a different GPU.

Being able to report values from a child Chain is useful for monitoring certain values of the network.

PR #4270, opened 2018-02-02
Fix multiprocess bug in multiprocess parallel updater

okuta:
I found a multiprocessing bug at #3402. This bug appears on Python 3 environment.
I propose to use `forkserver` mode.


okuta:
I found another way. Please see #3402. So, I close this PR.

PR #4276, opened 2018-02-06
iDeep (basic part)

niboshi:
Separated basic part of iDeep integration from #3865


chainer/__init__.py
```diff
@@ -89,6 +93,45 @@ def get_function_hooks():
...
 
 
+def _load_array_types():
+    global _array_types
+    global _cpu_array_types
+    if _array_types is None:
+        _array_types = [numpy.ndarray]
+        _cpu_array_types = [numpy.ndarray]
+
+        if cuda.available:
```
kmaehashi: `backends.cuda` instead of `cuda`?

chainer/backends/cuda.py
```diff
@@ -238,7 +238,7 @@ def to_gpu(array, device=None, stream=None):
     """Copies the given CPU array to the specified device.
 
     Args:
-        array (numpy.ndarray, cupy.ndarray, None, list or tuple):
+        array (array, None, list or tuple):
```
hvy: Doctest seems to be failing because of an ambiguous reference `WARNING: more than one target found for cross-reference 'array': chainer.Parameter.array, chainer.Variable.array`.

chainer/link.py
```diff
@@ -376,6 +378,20 @@ def to_gpu(self, device=None):
         self._cpu = False
         return self
 
+    def to_intel64(self):
```
kmaehashi: Do we have to provide `to_intel64` method, even though the use of iDeep is not determined by array data type?
I feel it is better for users not to be aware of difference of array types, as ideep.mdarray proxies methods to numpy array (and it now supports operator overload).
(The only concern is that ideep.mdarray does not support `__str__`/`__repr__` so that it shows `<Swig Object of type 'std::vector< mdarray >::value_type *' at 0x7f2011708650>` instead of np.array2string output when used interactively. Maybe we can ask Intel guys to implement this...)

chainer/variable.py
```diff
@@ -18,16 +19,16 @@ def _check_grad_type(func, x, gx):
     if x.data is None or gx is None:
         # ``x.data is None`` implies that the data array is not retained
         return
-    if not isinstance(gx, type(x.data)):
-        msg = ('Type of data and grad mismatch\n%s != %s' %
+    if not chainer.is_arrays_compatible((gx, x.data)):
+        msg = ('Type of data and grad mismatch\ngrad:%s != data:%s' %
```
kmaehashi: I prefer to have space after `:`.

chainer/backends/intel64.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+_ideep_version = None
+_error = None
+
+try:
+    import ideep4py as ideep  # NOQA
+    _ideep_version = 0
+except ImportError as e:
+    _error = e
```
kmaehashi: I think it better to make `mdarray` always available (as in [cuda](https://github.com/niboshi/chainer/blob/d5cd05e596c80943f4259940682a5eab593ec055/chainer/backends/cuda.py#L53-L58)), so that we don't have to check `is_ideep_available` everywhere just for type testing?

kmaehashi:
Maybe add `ideep4py` to `extra_requires` in `setup.py`?
(BTW I recently found that multiple requirements can be specified at once, e.g., `pip install -e '.[stylecheck,docs]'`)

chainer/__init__.py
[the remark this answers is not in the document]
niboshi: Fixed.

chainer/backends/cuda.py
niboshi: I couldn't replicate the error due to other non-related failures in my environment.
I hope it's fixed.

chainer/backends/intel64.py
niboshi: Fixed.

chainer/variable.py
niboshi: I agree. Fixed.

chainer/link.py
niboshi: It is needed in optimzers (See #4285).
When updating a parameter, iDeep will be used only when its underlying array is `mdarray`.
It's not automatically converted.

chainer/backends/intel64.py
```diff
@@ -0,0 +1,102 @@
...
+    """Determines if we should use iDeep.
+
+    This function checks ``chainer.config.use_ideep`` and availability
+    of ``ideep4py`` package.
+
+    Args:
+        level (str): iDeep use level. It must be either ``'==always'`` or
+            ``'>=auto'``. ``'==always'`` indicates that the ``use_ideep``
+            config must be ``'always'`` to use iDeep.
+
```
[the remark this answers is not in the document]
niboshi: I think it's introduced in analogy with `should_use_cudnn`, but we have not yet decided the specification of this argument. Also iDeep does not provide a way to retrieve version number.
(I don't think iDeep will have the versioning convention simlar to cuDNN, in which version numbers are represented by integers.)

Maybe we should remove this argument?

niboshi:
>Maybe add `ideep4py` to `extra_requires` in `setup.py`?

Do we need this?
In my understanding, currently ideep4py is not required from any configurations defined in setup.py.

chainer/__init__.py
```diff
@@ -89,6 +93,45 @@ def get_function_hooks():
...
+    _load_array_types()
+    return _array_types
+
+
+def get_cpu_array_types():
+    _load_array_types()
+    return _cpu_array_types
+
+
+def is_arrays_compatible(arrays):
```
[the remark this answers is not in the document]
niboshi: I neither have a strong opinion, but do *weakly* prefer explicit tuple as the current PR.
Perhaps it's better to hear other people's opinions.

chainer/link.py
kmaehashi: Are there any problem if `should_use_ideep` is used in `update_core_cpu` of Optimizers?
(Maybe I am missing something...)

niboshi: The point is `update_core_cpu` updates the array *in-place*. Thus the array has to be converted to `mdarray` prior to updates.

chainer/backends/intel64.py
```diff
@@ -0,0 +1,106 @@
...
+    """
+    if _ideep_version is None:
+        return False
+
+    inputs = [x.data if isinstance(x, chainer.variable.Variable)
+              else x for x in inputs]
+
+    return (ideep.check_ndim(inputs, supported_ndim)
+            and (isinstance(inputs[0], ideep.mdarray)
```
hvy: Is it enough to only check the first input?

chainer/backends/intel64.py
```diff
@@ -0,0 +1,106 @@
...
+
+    use_ideep = config.use_ideep
+    if use_ideep not in flags:
+        raise ValueError('invalid use_ideep configuration: %s '
+                         '(must be either of "always", "auto", or "never")' %
+                         repr(use_ideep))
+    return flags[use_ideep]
+
+
+def inputs_all_ready(inputs, supported_ndim=(2, 4)):
```
hvy: I see that this interface is aligned with `ideep.check_ndim` but are the supported ndims something that the caller of this function should be aware of?

chainer/variable.py
```diff
@@ -730,6 +739,31 @@ def to_gpu(self, device=None):
             if node._data is not None:
                 node.retain_data()
 
+    def to_intel64(self):
+        """Copies the data and gradient arrays to intel64 specific mdarray
```
hvy: A nitpick but should we add a punctation to this sentence?

chainer/backends/intel64.py
niboshi: Yes, some functions only support 4-dim arrays. The caller should explicitly specify it in such cases.

chainer/variable.py
hvy: Yes exactly, a small detail.

chainer/backends/intel64.py
hvy: Okay I see, thanks for the explanation.

niboshi: Fixed. Thanks!

chainer/variable.py
niboshi: Fixed. Thanks!

chainer/backends/intel64.py
niboshi: Fixed.

chainer/backends/intel64.py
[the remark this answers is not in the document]
niboshi: I removed the argument.

chainer/__init__.py
```diff
@@ -89,6 +93,45 @@ def get_function_hooks():
     return ret
 
 
+def _load_array_types():
+    global _array_types
+    global _cpu_array_types
+    if _array_types is None:
+        _array_types = [numpy.ndarray]
+        _cpu_array_types = [numpy.ndarray]
```
hvy: Since we are modifying globals, I feel that the above two variables should be temporary variables and that `_array_types` which is used in the condition should be set to a non-`None` value at the very end. Is it unnecessary you think?

tests/chainer_tests/test_variable.py
```diff
@@ -1606,4 +1608,92 @@ def test_loss_scale_gpu(self):
...
+    def setUp(self):
+        self.x_data = np.random.uniform(-1, 1, self.shape).astype(self.dtype)
+
+    def _check_variable_shape_and_dtype(self, var):
+        assert var.data.shape == self.shape
+        assert var.data.dtype == self.dtype
+        assert var.shape == self.shape
+        assert var.dtype == self.dtype
+
+    @skip_if_no_ideep()
```
kmaehashi: Is this annotation needed? (considering the discussion in https://github.com/cupy/cupy/pull/909)

chainer/link.py
niboshi: Yes, I think that's correct.

chainer/__init__.py
niboshi: Ah, I see.
I thought synchronization is not needed because it's protected by GIL, but the call to `is_ideep_available()` may incur a IO wait (although it's not the case in the current implementation).

tests/chainer_tests/test_variable.py
niboshi: Fixed. Thanks!

kmaehashi:
LGTM!

tests/chainer_tests/test_link.py
```diff
@@ -3,9 +3,11 @@
 
 import mock
 import numpy
+import pytest
```
hvy: Seems like Travis doesn't like this import.

tests/chainer_tests/test_variable.py
```diff
@@ -7,10 +7,12 @@
 
 import mock
 import numpy as np
+import pytest
```
hvy: Ditto.

chainer/__init__.py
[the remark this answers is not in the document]
hvy: I personally don't that they need to be consistent. `get_array_module` may very well accept a single array while `is_arrays_compatible` is meant to compare multiple instances.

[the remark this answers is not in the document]
niboshi: Ah, that's a good point. Actually `is_arrays_comparible` can take a single array, but conceptually, your point may make sense. (The reason why I want to make it a tuple is more pragmatic, though: to allow future extension).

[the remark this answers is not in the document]
hvy: I see. Then I feel we can leave it as is. My understanding is also that this function is mainly for internal usage (as it's not documented) in which case we could reconsider the interface at a later time.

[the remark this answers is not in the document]
kmaehashi: OK. My initial motivation was that I want to avoid confusion between usage of two similar functions (which accepts tuples and which accepts va-args) while coding, even for core developers, but I think @hvy's comment makes sense. Let's leave as it is.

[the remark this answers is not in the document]
hvy: (NumPy is indeed confusing in that sense...)

niboshi:
Is it OK to merge?

hvy:
jenkins, test this please.

chainer/__init__.py
```diff
@@ -89,6 +93,50 @@ def get_function_hooks():
...
+    return _array_types
+
+
+def get_cpu_array_types():
+    _load_array_types()
+    return _cpu_array_types
+
+
+def is_arrays_compatible(arrays):
+    arrays = [_ for _ in arrays if _ is not None]
```
beam2d: minor comment: do not use `_` in this way. `_` is basically used for variable names which are syntactically needed but never used. Here the variable is used to test if it's not None and also used to fill the list, so it should be explicitly named.

chainer/__init__.py
```diff
@@ -89,6 +93,50 @@ def get_function_hooks():
...
+    if len(arrays) == 0:
+        return True
+    if type(arrays[0]) is backends.cuda.ndarray:
+        types = backends.cuda.ndarray
+    else:
+        types = get_cpu_array_types()
+    return all(isinstance(_, types) for _ in arrays)
```
beam2d: ditto: `_` is used as an argument of `isinstance`.

chainer/backends/cuda.py
```diff
@@ -281,12 +281,16 @@ def _array_to_gpu(array, device, stream):
     assert device is DummyDevice or isinstance(device, Device)
     if array is None:
         return None
+
     if isinstance(array, (numpy.number, numpy.bool_)):
         array = numpy.asarray(array)
+    elif isinstance(array, intel64.mdarray):
+        # ideep.mdarray to numpy.ndarray
```
beam2d: I feel this comment does not explain anything but what is written in the code.

hvy:
jenkins, test this please.

chainer/__init__.py
niboshi: Fixed.

chainer/__init__.py
niboshi: Fixed.

chainer/backends/cuda.py
niboshi: Fixed.

hvy:
jenkins, test this please.

hvy:
jenkins, test this please.

kmaehashi:
Jenkins, test this please.

hvy:
jenkins, test this please.

hvy:
jenkins, test this please.

hvy:
LGTM!

PR #4278, opened 2018-02-06
iDeep: F.linear

niboshi:
iDeep support in F.linear (Separated from #3865. Based on #4276.)

### DIFF
--- a/chainer/functions/connection/linear.py
+++ b/chainer/functions/connection/linear.py
@@ -33,7 +33,6 @@ class LinearFunction(function_node.FunctionNode):
     def forward(self, inputs):
         if (intel64.should_use_ideep('>=auto')
                 and intel64.inputs_all_ready(inputs)):
-
             # iDeep implementation
             self._use_ideep = True
             return self._forward_ideep(inputs)
--- a/chainer/functions/connection/linear.py
+++ b/chainer/functions/connection/linear.py
@@ -8,7 +8,7 @@ from chainer.utils import type_check
 
 class LinearFunction(function_node.FunctionNode):
 
-    _use_ideep = False
+    _use_ideep_config = None
 
     def check_type_forward(self, in_types):
         n_in = in_types.size()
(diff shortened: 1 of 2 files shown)


chainer/functions/connection/linear.py
```diff
@@ -45,26 +53,114 @@ def forward(self, inputs):
...
+
     def backward(self, indexes, grad_outputs):
-        x, W = self.get_retained_inputs()
+        ret = []
         gy, = grad_outputs
 
+        x, W = self.get_retained_inputs()
+
+        if self._use_ideep:
+
```
beam2d: Remove this blank line

chainer/functions/connection/linear.py
```diff
@@ -45,26 +53,114 @@ def forward(self, inputs):
...
+                gW, = LinearGradWeight().apply((x, gy))
+                ret.append(chainer.functions.cast(gW, W.dtype))
+            if 2 in indexes:
+                gb = chainer.functions.sum(gy, axis=0)
+                ret.append(gb)
+
+        return ret
+
+
+class LinearGradDIdeep(function_node.FunctionNode):
```
beam2d: Why are these backward function node classes separate from the original ones?

niboshi: That would require passing `use_ideep` flag to each grad class and make explicit `if-else` branches.
Rather than that, isn't it cleaner to separate the classes?

chainer/functions/connection/linear.py
niboshi: Fixed.

chainer/functions/connection/linear.py
beam2d: Hmm. I'm concerning the maintenance cost; the implementation of the ideep version looks very similar to the original one (esp. the backward implementations are almost same). Can't we at least unify the backward implementations?

niboshi: OK, will do.

niboshi:
Sorry, previous backward implementation was not up to date to the latest intel-chainer implementation.
This version uses existing grad classes, just as you suggested.

Some notes:
* `use_ideep` config values are saved in forward pass and manually set in backward and double-backward pass.
* `gW` is correctly casted to `W.dtype` (instead of `gy.dtype`).

beam2d:
Jenkins, test this please.

beam2d:
LGTM

PR #4282, opened 2018-02-06
iDeep: F.average_pooling_2d

niboshi:
iDeep support in F.average_pooling_2d (Separated from #3865. Based on #4276.)

### DIFF
--- a/chainer/functions/pooling/average_pooling_2d.py
+++ b/chainer/functions/pooling/average_pooling_2d.py
@@ -37,15 +37,15 @@ class AveragePooling2D(pooling_2d.Pooling2D):
         y_w = conv.get_conv_outsize(
             w, self.kw, self.sx, self.pw, self.cover_all)
         assert y_w > 0, 'Width in the output should be positive.'
-        self.pd = self.sy * (y_h - 1) + self.kh - h - self.ph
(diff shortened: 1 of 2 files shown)


chainer/functions/pooling/average_pooling_2d.py
```diff
@@ -21,6 +26,30 @@ def forward_cpu(self, x):
...
+
+        n, c, h, w = x[0].shape
+        y_h = conv.get_conv_outsize(
+            h, self.kh, self.sy, self.ph, self.cover_all)
+        assert y_h > 0, 'Height in the output should be positive.'
+        y_w = conv.get_conv_outsize(
+            w, self.kw, self.sx, self.pw, self.cover_all)
+        assert y_w > 0, 'Width in the output should be positive.'
+        self.pd = self.sy * (y_h - 1) + self.kh - h - self.ph
+        self.pr = self.sx * (y_w - 1) + self.kw - w - self.pw
```
kmaehashi: Is it intentional to keep `pd` and `pr` as an instance attribute?

niboshi: I don't think so. Fixed.

kmaehashi:
LGTM!

Issue #4235, opened 2018-01-25
Add chainer.get_version_info()


mitmul:
Sometimes I want related package versions like this:

```
Chainer: 3.0.0b1
CUDA Version: 8.0
cuDNN Version: 5.11
CuPy: 2.0.0b1
```

This is actually the output of this code:

```
import chainer
import cupy

print('Chainer:', chainer.__version__)
print('CUDA:', chainer.cuda.cuda.runtime.runtimeGetVersion() / 1000.)
print('cuDNN:', chainer.cuda.cudnn.cudnn.getVersion() / 1000.)
print('CuPy:', cupy.__version__)
```

But I think it's better to provide more easy way for user to get those information, e.g., `chainer.get_version_info()` etc.

kmaehashi:
:+1: Related to https://github.com/cupy/cupy/issues/756

niboshi:
Some considerations.
* `version_info` might give an impression to be a tuple (c.f. `sys.version_info`)
* In Chainer, they're runtime configuration (unlike NumPy's `show_config()` and CuPy). We should be careful not to give an impression that they're build-time configuration.

mitmul:
It may be useful for user support. Just asking "call `chainer.get_version_info()` before posting questions to the forum" may help us to smoothly start the support.

niboshi:
I agree.
How about `get_runtime_info()`? (just an idea)

niboshi:
I briefly implemented it: #4298
Please leave any comments.

niboshi:
Fixed via #4559.

PR #4298, opened 2018-02-07
Runtime info

niboshi:
Fixes: #4235
Useful for troubleshooting.

```
Chainer: 4.0.0b3
NumPy: 1.13.0
CuPy: 4.0.0b3
CUDA driver: 9010
CUDA runtime: 9000
cuDNN: None
cuDNN enabled: False
NCCL: None
```

### DIFF
--- a/chainer/__init__.py
+++ b/chainer/__init__.py
@@ -29,6 +29,8 @@ from chainer import variable  # NOQA
 
 # import class and function
 # These functions from backends.cuda are kept for backward compatibility
+from chainer._runtime_info import get_runtime_info  # NOQA
+from chainer._runtime_info import print_runtime_info  # NOQA
 from chainer.backends.cuda import should_use_cudnn  # NOQA
 from chainer.backends.cuda import should_use_cudnn_tensor_core  # NOQA
 from chainer.configuration import config  # NOQA
--- a/chainer/_runtime_info.py
+++ b/chainer/_runtime_info.py
@@ -1,0 +1,41 @@
+import sys
+
+import numpy
+import six
+
+import chainer
+from chainer.backends import cuda
+
+
+class _RuntimeInfo(object):
+
+    chainer_version = None
+    numpy_version = None
+    cupy_info = None
+
+    def __init__(self):
+        self.chainer_version = chainer.__version__
+        self.numpy_version = numpy.__version__
+        self.cupy_info = cuda.get_runtime_info()
+
+    def __str__(self):
+        s = six.StringIO()
+        s.write('''Chainer: {}\n'''.format(self.chainer_version))
+        s.write('''NumPy: {}\n'''.format(self.numpy_version))
+        s.write(str(self.cupy_info))
+        return s.getvalue()
+
+
+def get_runtime_info(as_text=False):
+    ri = _RuntimeInfo()
+    if as_text:
+        return str(ri)
+    else:
+        return ri
+
+
+def print_runtime_info(out=None):
+    if out is None:
+        out = sys.stdout
+    text = get_runtime_info(as_text=True)
+    out.write(text)
--- a/chainer/_runtime_info.py
+++ b/chainer/_runtime_info.py
@@ -11,18 +11,18 @@ class _RuntimeInfo(object):
 
     chainer_version = None
     numpy_version = None
-    cupy_info = None
+    cuda_info = None
 
     def __init__(self):
         self.chainer_version = chainer.__version__
         self.numpy_version = numpy.__version__
-        self.cupy_info = cuda.get_runtime_info()
+        self.cuda_info = cuda.get_runtime_info()
 
     def __str__(self):
         s = six.StringIO()
         s.write('''Chainer: {}\n'''.format(self.chainer_version))
         s.write('''NumPy: {}\n'''.format(self.numpy_version))
-        s.write(str(self.cupy_info))
+        s.write(str(self.cuda_info))
         return s.getvalue()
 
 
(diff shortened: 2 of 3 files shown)


niboshi:
I'll write tests and documentation if interface is OK.

niboshi:
I removed `get_runtime_info(as_text=)` argument, as it's fairly intuitive to write `str(chainer.get_runtime_info())` or `print(chainer.get_runtime_info())`.

chainer/__init__.py
```diff
@@ -29,6 +29,8 @@
 
 # import class and function
 # These functions from backends.cuda are kept for backward compatibility
+from chainer._runtime_info import get_runtime_info  # NOQA
```
kmaehashi: I think these lines should not be placed after the comment above.

chainer/_runtime_info.py
```diff
@@ -0,0 +1,36 @@
...
+    cuda_info = None
+
+    def __init__(self):
+        self.chainer_version = chainer.__version__
+        self.numpy_version = numpy.__version__
+        self.cuda_info = cuda.get_runtime_info()
+
+    def __str__(self):
+        s = six.StringIO()
+        s.write('''Chainer: {}\n'''.format(self.chainer_version))
```
kmaehashi: Are there any reasons to use `'''` instead of  `'`?

chainer/backends/cuda.py
```diff
@@ -67,9 +67,48 @@ class ndarray(object):
...
+            try:
+                self.nccl_version = cuda.nccl.get_version()
+            except AttributeError:
+                pass
+
+    def __str__(self):
+        s = six.StringIO()
+        s.write('''CuPy: {}\n'''.format(self.cupy_version))
+        if self.cupy_version is not None:
+            s.write('''CUDA runtime: {}\n'''.format(self.cuda_runtime_version))
```
kmaehashi: Personally I prefer printing the driver version first, then runtime version, considering the software stack.

chainer/_runtime_info.py
```diff
@@ -0,0 +1,36 @@
+import sys
+
+import numpy
+import six
+
+import chainer
+from chainer.backends import cuda
+
+
+class _RuntimeInfo(object):
```
kmaehashi: I'm not sure when to prefix `_` to class names in Chainer. Any thoughts?

niboshi: I wanted to signify that this class is implementation detail and not supporsed to instantiated by user.
(I think there's no established convention in the world when or when not to add this prefix, but I simply thought it won't harm anyone)

chainer/_runtime_info.py
niboshi: Because these are strings meant to be presented to user, which means they can reasonably contain `'`.

chainer/__init__.py
niboshi: It should work because the comment explicitly states what the subjects it's talking about.
If the comment is to be moved to just in front of the subjects, I think these imports should be separated from other imports.

chainer/backends/cuda.py
niboshi: Fixed!

chainer/__init__.py
kmaehashi: OK.  I think it is better to move legacy things to https://github.com/niboshi/chainer/blob/d5cf9389835e2bf83857633b53daa5dd271459f5/chainer/__init__.py#L69 but it's not related to this PR.

niboshi: OK, I'll leave it as is for this PR.

kmaehashi:
As build-time version of CUDA/cuDNN is also available via https://github.com/cupy/cupy/pull/937, could you add these information as well?

kmaehashi:
Ah, it may be out of scope as this inteface is for `Runtime Info`...
(Maybe it's better to print that info in `cupy.show_config` instead)

niboshi:
I thought that kind of build-time information can also be included.
(Do you think `RuntimeInfo` is inappropriate?)

kmaehashi:
I think it's nice to display every information in one place, but I'm wondering how this feature should be differentiated with `cupy.show_config()` (if we implement it).
Let every CUDA things be in `cupy.show_config()` and call it from `chainer.print_runtime_info()`?

niboshi:
Perhaps `_CudaRuntimeInfo` can retrieve the whole text from `cupy.show_config()` and print it as is.
I think there's still a meaning for `_CudaRuntimeInfo`, as even if `cupy` is not available it may print some information (such as environment variables and system information).

kmaehashi:
> Perhaps _CudaRuntimeInfo can retrieve the whole text from cupy.show_config() and print it as is.

I think that is good. `xp.show_config()` directly prints to stdout, so maybe another hidden API is needed.

> I think there's still a meaning for _CudaRuntimeInfo, as even if cupy is not available it may print some information (such as environment variables and system information).

I agree with that.

kmaehashi:
@niboshi Do you have plans to work on CuPy side implementation? If not, I can takeover this task.
As CuPy is getting more focus from community (and raising number of support issues), I'd like to define interface to diagnose issues in CuPy v4.

What I have in my mind is:

* `cupyx.get_runtime_info()`: return runtime information object, with `format` method.
* `cupy.show_config()`: print `cupyx.get_runtime_info().format()` to stdout.

niboshi:
@kmaehashi OK, please go head.
Do you mean you want to take over this PR, too?

kmaehashi:
I've moved `_CudaRuntimeInfo` to CuPy (https://github.com/cupy/cupy/pull/1067) with some modification including addition of CUDA Path and build-time version.

> Do you mean you want to take over this PR, too?

Yes, I think it's better.

niboshi:
@kmaehashi 
OK. Then I'll review.

kmaehashi:
I took over the task in #4559.

PR #4300, opened 2018-02-07
Avoid keeping reference to function nodes in CupyMemoryProfileHook

niboshi:
`CupyMemoryProfileHook` keeps `FunctionNode` references in the call history, thus the memory held by these functions are never released within the hook scope.

That can unnecessarily cause OOM error if multiple loops are measured at once.

```
hook = function_hooks.CupyMemoryProfileHook()
with hook:
    train_loop1()
    train_loop2()
    train_loop3()  # memory consumption is accumulated
hook.print_report()
```
(Note: the numbers in the profiling result is not affected by accumulation)

This fix is to keep function names in the call history instead of function objects.
This is backward-incompatible, because the content of call history (which is documented) is altered.


beam2d:
LGTM

PR #4301, opened 2018-02-07
Unchain function nodes in backward

niboshi:
Currently, all the function nodes in the computational graph is kept alive until whole backward computation is finished. Essentially, function nodes whose gradients are already propagated to all the input nodes are no longer needed and can be freed. If such function node holds large arrays as its attributes (usually in order to save common calculation between forward() and backward()) it unnecessarily occupies the memory.

This PR is to free such function nodes.

There are two types of references to such function nodes:
1. References from output variables' `creator` property.
2. `seen_set` in the backward graph traversal.

For 1., `creator` nodes of each variables are unchained as traversing backward.
For 2., weak references is used in `seen_set`.

## Code:
This is somewhat artificial example, but can demonstrate the effect.
```
import chainer
import cupy


class FooFunc(chainer.FunctionNode):
    def __init__(self, name):
        super().__init__()
        self.name = name

    def __del__(self):
        self.bar = None
        print("{:10d} DEL: {}".format(cuda.memory_pool.used_bytes(), self.name))

    def forward(self, inputs):
        # Dummy array which can be used in backward()
        self.bar = cupy.ones((100, 100, 100, 10), cupy.float32)
        x, = inputs
        y = x + 1
        print("{:10d} FWD: {}".format(cuda.memory_pool.used_bytes(), self.name))
        return y,

        gy, = grad_outputs
        print("{:10d} BWD: {}".format(cuda.memory_pool.used_bytes(), self.name))
        return gy,


def func(x):
    hs = x,
    hs = FooFunc('a').apply(hs)
    hs = FooFunc('b').apply(hs)
    hs = FooFunc('c').apply(hs)
    hs = FooFunc('d').apply(hs)
    hs = FooFunc('e').apply(hs)
    h, = hs
    return h


def main():
    x = chainer.Variable(cupy.ones((2,)))
    y = func(x)
    y.grad = cupy.ones_like(y.data)
    y.backward()
    print(y, y.grad, x.grad)


main()
```

## Result (master):
```
  40001024 FWD: a
  80001536 FWD: b
 120001536 FWD: c
 160001536 FWD: d
 200001536 FWD: e
 200001536 BWD: e
 200001536 BWD: d
 200001536 BWD: c
 200001536 BWD: b
 200001536 BWD: a
variable([ 6.  6.]) [ 1.  1.] [ 1.  1.]
 160000512 DEL: e
 120000512 DEL: d
  80000512 DEL: c
  40000512 DEL: b
       512 DEL: a
```

## Result (This PR):
```
  40001024 FWD: a
  80001536 FWD: b
 120001536 FWD: c
 160001536 FWD: d
 200001536 FWD: e
 200001536 BWD: e
 160001536 DEL: e
 160001536 BWD: d
 120001536 DEL: d
 120001536 BWD: c
  80001536 DEL: c
  80001536 BWD: b
  40001536 DEL: b
  40001536 BWD: a
      1536 DEL: a
variable([ 6.  6.]) [ 1.  1.] [ 1.  1.]

```

### DIFF
--- a/chainer/variable.py
+++ b/chainer/variable.py
@@ -911,6 +911,7 @@ Actual: {0}'''.format(type(data))
                 if e.status != 38:  # cudaErrorNoDevice
                     raise
 
+        enable_double_backprop = chainer.config.enable_backprop
         is_debug = chainer.is_debug()
 
         cand_funcs = []
@@ -929,12 +930,23 @@ Actual: {0}'''.format(type(data))
         grads[self._node] = self._grad_var
 
         def add_cand(cand):
-            if cand not in seen_set:
+            # weakref is used for `seen_set` so that unchained function nodes
+            # can be freed in the backward process.
+            # If a weakref corresponding to `cand` is already in the set,
+            # it's guaranteed to be alive, because `cand` is alive.
+            # If a weakref in the set has been invalidated, its equality with
+            # `cand` is guaranteed to be negative.
+            # See the documentation of the weakref module.
+            cand_ref = weakref.ref(cand)
+            if cand_ref not in seen_set:
                 # Negate since heapq is min-heap
                 heapq.heappush(cand_funcs, (-cand.rank, len(seen_set), cand))
-                seen_set.add(cand)
+                seen_set.add(cand_ref)
 
         add_cand(self.creator_node)
+
+        if not enable_double_backprop:
+            self.unchain()
 
         def get_grad(node):
             if node is None:
@@ -1051,6 +1063,8 @@ Actual: {0}'''.format(type(data))
 
                 if x.creator_node is not None:
                     add_cand(x.creator_node)
+                    if not enable_double_backprop:
+                        x.unchain()
 
             del gxs  # to reduce memory usage
             if initial_device is not None:
--- a/chainer/variable.py
+++ b/chainer/variable.py
@@ -911,7 +911,10 @@ Actual: {0}'''.format(type(data))
                 if e.status != 38:  # cudaErrorNoDevice
                     raise
 
-        enable_double_backprop = chainer.config.enable_backprop
+        unchain_creator_funcs = not (
+            chainer.config.enable_backprop
+            or chainer.config.keep_graph_on_report)
+
         is_debug = chainer.is_debug()
 
         cand_funcs = []
@@ -945,7 +948,7 @@ Actual: {0}'''.format(type(data))
 
         add_cand(self.creator_node)
 
-        if not enable_double_backprop:
+        if unchain_creator_funcs:
             self.unchain()
 
         def get_grad(node):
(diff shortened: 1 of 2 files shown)


niboshi:
It breaks backward compatibility because repeated call of `backward()` will be ignored.
```
x = chainer.Variable(numpy.ones((1,), numpy.float32))
y = x * x + x
y.backward(); print(x.grad)
y.backward(); print(x.grad)
y.backward(); print(x.grad)
```
master:
```
[ 3.]
[ 6.]
[ 9.]
```
This PR:
```
[ 3.]
[ 3.]
[ 3.]
```
So it may be better to make it configurable, but I wonder if this master's behavior is documented (at least it's not tested, because the unit test is successfull (except irrelevant doctest code)).

beam2d:
Maybe it's not documented, but we should be careful to change such a fundamental behavior. I think it is well known (at least among old users) that multiple calls of backward accumulate the gradients, so there may exist user code depending on this behavior.

hvy:
Should this PR be further discussed (maybe raise an issue to enhance the docs) or closed for the time being? I also think that user can benefit from this feature but agree with @beam2d .

niboshi:
I think this PR needs not to be closed.

Backward compatibility can be kept by making the behavior configurable.
But I think it's better to change the default behavior to unchaining (thus no-compat), but issue an error if user tries to call backward() twice without disabling this behavior.
That would benefit most of users, and still avoid silently breaking existing user code.

hvy:
Then let's leave it as is and continue with the discussion. I'd personally like to see this feature but maybe we should ask the others too.

toslunar:
How about moving the core of `Variable.backward()` to `FunctionNode.backward()`, which would allow users to write the following?
```
func = y.creator_node
del y
func.backward()
```

toslunar:
I confirmed this approach works:
```
    cont = y.backward(execute=False)
    del y
    cont()
```
results
```
$ python test4301.py
  40001024 FWD: a
  80001536 FWD: b
 120001536 FWD: c
 160001536 FWD: d
 200001536 FWD: e
variable([1. 1.]) variable([32. 32.])
 200001024 BWD: e
 160001536 DEL: e
 160001536 BWD: d
 120002048 DEL: d
 120001536 BWD: c
  80002048 DEL: c
  80001536 BWD: b
  40002048 DEL: b
  40001536 BWD: a
      2048 DEL: a
variable([32. 32.]) variable([1. 1.])
[variable([32., 32.])]
```
(code: https://github.com/chainer/chainer/compare/master...toslunar:c534a0964e9e2e812eba2ab3301b1862651cd4ad)

However it required some refactoring like `grads.pop`.  It's welcome to take over the code.  If no one does, I'll work on this issue after #4772 is merged.

niboshi:
After an internal discussion, we concluded to close the PR for now.

* It doesn't improve total memory consupmtion; it's only beneficial when large memory is consumed during backward computation. Currently we haven't come up with such case.
* It may not be worth introducing another configurable parameter that can potentially break the exising code.
* We may revisit the PR when we recognize a case where this fix is beneficial.

niboshi:
Please feel free to reopen ths PR if anyone wants this feature.

PR #4306, opened 2018-02-08
fix to pass doctest around optimizer.setup

kmaehashi:
Fix issue introduced in #4290.

`optimizer.setup(model)` now returns itself, so it will print sth like `<chainer.optimizers.sgd.SGD object at 0x7f45831ea940>` when interactive.


hvy:
LGTM!

PR #4044, opened 2017-12-06
Add test for inconsistent backend input outputs

hvy:
Add tests for `FunctionNode.apply()` input, output backend checks (exceptions are raised when ndarray backends are mixed https://github.com/chainer/chainer/pull/4029).

Note that this PR should not be backported since #4029 isn't.

PR #4312, opened 2018-02-09
Improve error messages of array type check.

mizuno-gsinet:
This is a proposal of improved error messages in `FunctionNode.apply()`. It also changes the type of exception.

The new message should be more understandable, for example, when a user forgot to use `utils.force_array()` in their own FunctionNode.

Original and new error messages look like these:

Original

```
ValueError: numpy and cupy arrays are mixed in the forward output (MyFunc).
<class 'cupy.core.core.ndarray'>, <class 'float'>
```

This PR

```
TypeError: forward outputs must be `ndarray`s of the same array module (numpy or cupy) (MyFunc).
Actual: <class 'cupy.core.core.ndarray'>, <class 'float'>
```

### DIFF
--- a/chainer/function_node.py
+++ b/chainer/function_node.py
@@ -227,10 +227,10 @@ Use apply() method instead.\
 
         # Check for input array types
         if not chainer.is_arrays_compatible(in_data):
-            raise ValueError(
+            raise TypeError(
                 'incompatible array types are mixed in the forward input '
                 '({}).\n'
-                '{}'.format(
+                'Actual: {}'.format(
                     self.label,
                     ', '.join(str(type(x)) for x in in_data)))
 
@@ -264,10 +264,10 @@ Use apply() method instead.\
                 'Actual: {}'.format(self.label, type(outputs)))
 
         if not chainer.is_arrays_compatible(outputs):
-            raise ValueError(
+            raise TypeError(
                 'incompatible array types are mixed in the forward output '
                 '({}).\n'
-                '{}'.format(
+                'Actual: {}'.format(
                     self.label,
                     ', '.join(str(type(x)) for x in outputs)))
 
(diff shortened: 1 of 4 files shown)


kmaehashi:
I think this fix is reasonable. How do you think @niboshi?
When merged we should fix the doc too. https://docs.chainer.org/en/latest/upgrade.html#prohibition-of-mixed-use-of-arrays-on-different-devices-in-function-arguments

mizuno-gsinet:
I rebased and adopted the new message from #4346. It now looks like this:

```
TypeError: incompatible array types are mixed in the forward output (MyFunc).
Actual: <class 'cupy.core.core.ndarray'>, <class 'float'>
```

Does this still have a point?

kmaehashi:
I think message like `Types of all arrays in the forward input must be the same` is better, but would like to ask opinion of @niboshi san.
I'm also positive for changing the exception type.

kmaehashi:
Jenkins, test this please.

niboshi:
@kmaehashi 
Sorry, I overlooked the mention.
I agree with change of exception type.

kmaehashi:
Ok, I'll merge this after test pass.

kmaehashi:
Jenkins, test this please.

kmaehashi:
@mizuno-gsinet Could you please fix unit tests?
https://github.com/chainer/chainer/blob/v4.0.0b4/tests/chainer_tests/links_tests/connection_tests/test_embed_id.py#L56-L67

mizuno-gsinet:
I fixed them.

kmaehashi:
Jenkins, test this please.

kmaehashi:
Jenkins, test this please.

kmaehashi:
Sorry, some more tests for this case has been added since then (via #4044), and test is failing.
https://github.com/chainer/chainer/blob/e5fe298e7d69f9ba1c773b5b1ab01b74dfd32e08/tests/chainer_tests/test_function_node.py#L325-L357
Could you please fix them also?

mizuno-gsinet:
Sure. I rebased and fixed them.

kmaehashi:
Thank you!
Jenkins, test this please.

kmaehashi:
LGTM!

PR #4317, opened 2018-02-10
Use version in external URL

vilyaair:
Some external URLs in the document directly refer to the master branch, and they should be pointed to the corresponding tag.

I utilize sphinx.ext.extlinks to achieve this goal. However, I wonder if it is appropriate to use names 'blob' and 'tree', and actually github automatically redirects to correct URL whichever original URL is blob or tree.

### DIFF
--- a/docs/source/tutorial/function.rst
+++ b/docs/source/tutorial/function.rst
@@ -608,4 +608,4 @@ offers a utility function :func:`chainer.gradient_check.check_backward` that fol
    unittest.TextTestRunner().run(suite)
 
 
-You can find many examples of function tests under ``tests/chainer_tests/function_tests`` directory.
+You can find many examples of function tests under :tree:`tests/chainer_tests/function_tests` directory.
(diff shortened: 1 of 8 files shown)


niboshi:
jenkins, test this please

docs/source/tutorial/function.rst
```diff
@@ -608,4 +608,4 @@ offers a utility function :func:`chainer.gradient_check.check_backward` that fol
    unittest.TextTestRunner().run(suite)
 
 
-You can find many examples of function tests under ``tests/chainer_tests/function_tests`` directory.
+You can find many examples of function tests under :tree:`tests/chainer_tests/function_tests` directory.
```
niboshi: `function_tests` -> `functions_tests`

vilyaair: I appreciate it. I should scrutinize my code more carefully.

niboshi:
jenkins, test this please

niboshi:
LGTM!
Thank you very much!

PR #4323, opened 2018-02-13
Avoid keeping reference to function nodes in TimerHook

kmaehashi:
It's better to apply the same fix made in #4300 (CupyMemoryProfileHook).


niboshi:
LGTM!

PR #4326, opened 2018-02-13
Use ResNeXt in the ImageNet example

vilyaair:
It's a pity that ResNeXt has been implemented but not used in the ImageNet example.


beam2d:
Good catch!

PR #4339, opened 2018-02-15
remove input type consistency checks

kmaehashi:
These codes are invalid as iDeep is now introduced.
Note that type consistency is now checked in FunctionNode level (#4029), so we can safely remove these checks.

Type check code in `F.linear` will be removed in #4278.


niboshi:
LGTM!

niboshi:
Not to be backported because #4029 has been merged only in master branch.

PR #4345, opened 2018-02-15
Default statistics to get array modules

### LINKED ISSUE #3027: extensions.ParameterStatistics cannot work with default parameter in GPU mode
`extensions.ParameterStatistics` raise error in GPU mode at `numpy.count_nonzero` or `numpy.percentile` at https://github.com/chainer/chainer/blob/master/chainer/training/extensions/parameter_statistics.py#L106

The type of `link.param` is cupy.ndarray but these functions cannot take this type.
Other functions (like numpy.mean) use the instance method (like hoge.mean).

hvy:
Fixes #3027

### DIFF
--- a/chainer/training/extensions/parameter_statistics.py
+++ b/chainer/training/extensions/parameter_statistics.py
@@ -121,9 +121,8 @@ class ParameterStatistics(extension.Extension):
                             attr_name=attr_name,
                             function_name=function_name
                         )
-                        if (hasattr(value, '__iter__')
-                            or (isinstance(value, cuda.ndarray)
-                                and value.size > 1)):
+                        if (isinstance(value, (numpy.ndarray, cuda.ndarray))
+                                and value.size > 1):
                             # Append integer indices to the keys if the
                             # statistic function return multiple values
                             statistics.update({'{}/{}'.format(key, i): v for
--- a/chainer/training/extensions/parameter_statistics.py
+++ b/chainer/training/extensions/parameter_statistics.py
@@ -41,7 +41,8 @@ class ParameterStatistics(extension.Extension):
             the values.
         skip_nan_params (bool): If ``True``, statistics are not computed for
             parameters including NaNs and a single NaN value is immediately
-            reported instead.
+            reported instead. Otherwise, this extension will simply try to
+            compute the statistics without performing any checks for NaNs.
     """
     default_name = 'parameter_statistics'
     priority = extension.PRIORITY_WRITER
--- a/chainer/training/extensions/parameter_statistics.py
+++ b/chainer/training/extensions/parameter_statistics.py
@@ -1,6 +1,7 @@
 import numpy
 import six
 
+from chainer.backends import cuda
 from chainer import reporter
 from chainer.training import extension
 from chainer.training import trigger as trigger_module
@@ -42,19 +43,21 @@ class ParameterStatistics(extension.Extension):
     default_name = 'parameter_statistics'
     priority = extension.PRIORITY_WRITER
 
+    # avoid computations involving NaNs, e.g. when initial gradients are NaNs
+    check_nan_params = False
+
     # prefix ends with a '/' and param_name is preceded by a '/'
     report_key_template = ('{prefix}{link_name}{param_name}/{attr_name}/'
                            '{function_name}')
 
     default_statistics = {
-        'mean': numpy.mean,
-        'std': numpy.std,
-        'min': numpy.min,
-        'max': numpy.max,
-        'zeros': lambda x: numpy.count_nonzero(x == 0),
-        'percentile': lambda x: numpy.percentile(x, (0.13, 2.28, 15.87,
-                                                     50, 84.13, 97.72,
-                                                     99.87))
+        'mean': lambda x: cuda.get_array_module(x).mean(x),
+        'std': lambda x: cuda.get_array_module(x).std(x),
+        'min': lambda x: cuda.get_array_module(x).min(x),
+        'max': lambda x: cuda.get_array_module(x).max(x),
+        'zeros': lambda x: cuda.get_array_module(x).count_nonzero(x == 0),
+        'percentile': lambda x: cuda.get_array_module(x).percentile(
+            x, (0.13, 2.28, 15.87, 50, 84.13, 97.72, 99.87))
     }
 
     def __init__(self, links, statistics=default_statistics,
@@ -65,7 +68,7 @@ class ParameterStatistics(extension.Extension):
             links = links,
         self._links = links
 
-        self._statistics = statistics
+        self._statistics = statistics or {}
 
         attrs = []
         if report_params:
@@ -103,7 +106,12 @@ class ParameterStatistics(extension.Extension):
                         # since the statistics function should make no
                         # assumption about the axes
                         params = getattr(param, attr_name).ravel()
-                        value = function(params)
+                        if self.check_nan_params and \
+                                cuda.get_array_module(params).isnan(params) \
+                                    .any():
+                            value = numpy.nan
+                        else:
+                            value = function(params)
                         key = self.report_key_template.format(
                             prefix=self._prefix + '/' if self._prefix else '',
                             link_name=link_name,
@@ -111,7 +119,10 @@ class ParameterStatistics(extension.Extension):
                             attr_name=attr_name,
                             function_name=function_name
                         )
-                        if hasattr(value, '__iter__'):
+                        if hasattr(value, '__iter__') or \
+                                (isinstance(value,
+                                            (numpy.ndarray, cuda.ndarray))
+                                 and value.size > 1):
                             # Append integer indices to the keys if the
                             # statistic function return multiple values
                             statistics.update({'{}/{}'.format(key, i): v for
--- a/chainer/training/extensions/parameter_statistics.py
+++ b/chainer/training/extensions/parameter_statistics.py
@@ -106,9 +106,9 @@ class ParameterStatistics(extension.Extension):
                         # since the statistics function should make no
                         # assumption about the axes
                         params = getattr(param, attr_name).ravel()
-                        if self.check_nan_params and \
-                                cuda.get_array_module(params).isnan(params) \
-                                    .any():
+                        if (self.check_nan_params
+                            and (cuda.get_array_module(params).isnan(params)
+                                 .any())):
                             value = numpy.nan
                         else:
                             value = function(params)
@@ -119,10 +119,9 @@ class ParameterStatistics(extension.Extension):
                             attr_name=attr_name,
                             function_name=function_name
                         )
-                        if hasattr(value, '__iter__') or \
-                                (isinstance(value,
-                                            (numpy.ndarray, cuda.ndarray))
-                                 and value.size > 1):
+                        if (hasattr(value, '__iter__')
+                            or (isinstance(value, cuda.ndarray)
+                                and value.size > 1)):
                             # Append integer indices to the keys if the
                             # statistic function return multiple values
                             statistics.update({'{}/{}'.format(key, i): v for
--- a/chainer/training/extensions/parameter_statistics.py
+++ b/chainer/training/extensions/parameter_statistics.py
@@ -68,7 +68,7 @@ class ParameterStatistics(extension.Extension):
             links = links,
         self._links = links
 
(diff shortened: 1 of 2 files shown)


chainer/training/extensions/parameter_statistics.py
```diff
@@ -42,19 +43,21 @@ class ParameterStatistics(extension.Extension):
     default_name = 'parameter_statistics'
     priority = extension.PRIORITY_WRITER
 
+    # avoid computations involving NaNs, e.g. when initial gradients are NaNs
+    check_nan_params = False
```
niboshi: How about making it `__init__` argument?
Also write documentation instead of comment.

chainer/training/extensions/parameter_statistics.py
```diff
@@ -103,15 +106,23 @@ def __call__(self, trainer):
                         # since the statistics function should make no
                         # assumption about the axes
                         params = getattr(param, attr_name).ravel()
-                        value = function(params)
+                        if self.check_nan_params and \
+                                cuda.get_array_module(params).isnan(params) \
+                                    .any():
```
niboshi: Avoid backslashes for line continuation. Use parentheses instead.
Also, place `and` at the beginning of the next line.
(See PEP-8).

chainer/training/extensions/parameter_statistics.py
```diff
@@ -103,15 +106,23 @@ def __call__(self, trainer):
...
                             link_name=link_name,
                             param_name=param_name,
                             attr_name=attr_name,
                             function_name=function_name
                         )
-                        if hasattr(value, '__iter__'):
+                        if hasattr(value, '__iter__') or \
+                                (isinstance(value,
+                                            (numpy.ndarray, cuda.ndarray))
+                                 and value.size > 1):
```
niboshi: Ditto.

tests/chainer_tests/training_tests/extensions_tests/test_parameter_statistics.py
```diff
@@ -7,64 +7,101 @@
...
+        return extension
 
 
 @testing.parameterize(
-    {'statistics': {'min': numpy.min}, 'expect': 4},
-    {'statistics': {'mean': numpy.mean, 'std': numpy.std}, 'expect': 8},
-    {'statistics': extensions.ParameterStatistics.default_statistics,
-     'expect': 48},
-    {'statistics': {}, 'expect': 0}
+    {'links': [chainer.links.Linear(3, 2)], 'statistics': {'min': lambda x: cuda.get_array_module(x).min(x)}, 'report_params': True, 'report_grads': True, 'prefix': None, 'expect': 4},  # NOQA
```
niboshi: Break lines instead of NOQA.

chainer/training/extensions/parameter_statistics.py
```diff
@@ -103,15 +106,23 @@ def __call__(self, trainer):
...
                         key = self.report_key_template.format(
                             prefix=self._prefix + '/' if self._prefix else '',
                             link_name=link_name,
                             param_name=param_name,
                             attr_name=attr_name,
                             function_name=function_name
                         )
-                        if hasattr(value, '__iter__'):
+                        if hasattr(value, '__iter__') or \
+                                (isinstance(value,
```
okuta: Why do you add this checking?

chainer/training/extensions/parameter_statistics.py
```diff
@@ -65,7 +68,7 @@ def __init__(self, links, statistics=default_statistics,
             links = links,
         self._links = links
 
-        self._statistics = statistics
+        self._statistics = statistics or {}
```
okuta: Please use `is None` when you check `None`. See also [PEP8](https://www.python.org/dev/peps/pep-0008/#programming-recommendations).

chainer/training/extensions/parameter_statistics.py
hvy: Added it since cupy ndarrays don't have the `__iter__` attribute. Simplified it a bit though since the check is not needed for numpy.

chainer/training/extensions/parameter_statistics.py
hvy: Thanks, fixed!

chainer/training/extensions/parameter_statistics.py
hvy: Thank you, fixed!

chainer/training/extensions/parameter_statistics.py
hvy: Fixed!

tests/chainer_tests/training_tests/extensions_tests/test_parameter_statistics.py
hvy: Fixed.

chainer/training/extensions/parameter_statistics.py
hvy: Okay, fixed!

hvy:
@okuta @niboshi thank you for the quick review. I fixed most of the issues, could you please take another look?

chainer/training/extensions/parameter_statistics.py
```diff
@@ -38,6 +39,9 @@ class ParameterStatistics(extension.Extension):
         prefix (str): Optional prefix to prepend to the report keys.
         trigger: Trigger that decides when to aggregate the results and report
             the values.
+        skip_nan_params (bool): If ``True``, statistics are not computed for
+            parameters including NaNs and a single NaN value is immediately
+            reported instead.
```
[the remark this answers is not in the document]
hvy: Added more descriptive text.

chainer/training/extensions/parameter_statistics.py
```diff
@@ -103,15 +107,22 @@ def __call__(self, trainer):
...
                             prefix=self._prefix + '/' if self._prefix else '',
                             link_name=link_name,
                             param_name=param_name,
                             attr_name=attr_name,
                             function_name=function_name
                         )
-                        if hasattr(value, '__iter__'):
+                        if (hasattr(value, '__iter__')
+                            or (isinstance(value, cuda.ndarray)
+                                and value.size > 1)):
```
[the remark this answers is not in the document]
hvy: I simplified the check to only check if the output is an ndarray with size > 1.

hvy:
Thank you for the feedback. I made some changes accordingly.

chainer/training/extensions/parameter_statistics.py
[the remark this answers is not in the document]
niboshi: Ah, I mean, please state that `False` is assumed by default.

chainer/training/extensions/parameter_statistics.py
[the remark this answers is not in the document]
niboshi: How about using `chainer.get_array_types()`?

chainer/training/extensions/parameter_statistics.py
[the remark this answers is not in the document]
hvy: I have purposely left it out (and similarly for other arguments) since it is duplicate information with the function definition(?). Although I agree it sometimes helps in terms of readability. Do you still think we should add it?

chainer/training/extensions/parameter_statistics.py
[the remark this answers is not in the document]
hvy: Ah, that is perfect. Thank you!

chainer/training/extensions/parameter_statistics.py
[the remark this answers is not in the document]
niboshi: Ah, I see.
You're right, and it seems in many other places default value is not necessarily explicitly stated.
Thank you for pointing out.

tests/chainer_tests/training_tests/extensions_tests/test_parameter_statistics.py
```diff
@@ -3,68 +3,166 @@
...
         pattern = r'^(.+/){2,}(data|grad)/.+[^/]$'
         for name in six.iterkeys(self.trainer.observation):
+            if self.report_params and self.report_grads:
+                assert any(attr in name for attr in ['data', 'grad'])
+            elif self.report_params:
+                assert 'data' in name
+            elif self.report_grads:
+                assert 'grad' in name
+            if self.prefix is not None:
+                name.startswith(self.prefix)
```
niboshi: Did you forget `assert`?

chainer/training/extensions/parameter_statistics.py
```diff
@@ -47,25 +53,24 @@ class ParameterStatistics(extension.Extension):
...
                  report_params=True, report_grads=True, prefix=None,
-                 trigger=(1, 'epoch')):
+                 trigger=(1, 'epoch'), skip_nan_params=False):
 
         if not isinstance(links, (list, tuple)):
             links = links,
         self._links = links
 
-        self._statistics = statistics
+        self._statistics = statistics if statistics is not None else {}
```
niboshi: It's better to split:
```
if statistics is None:
    statistics = {}
```

tests/chainer_tests/training_tests/extensions_tests/test_parameter_statistics.py
```diff
@@ -3,68 +3,166 @@
...
     def test_report_key_pattern(self):
-        extension = extensions.ParameterStatistics(self.links)
-        self.trainer.extend(extension)
+        self.trainer.extend(self.create_extension())
         self.trainer.run()
 
         pattern = r'^(.+/){2,}(data|grad)/.+[^/]$'
         for name in six.iterkeys(self.trainer.observation):
+            if self.report_params and self.report_grads:
+                assert any(attr in name for attr in ['data', 'grad'])
```
niboshi: Doesn't this check covered with regex check?

tests/chainer_tests/training_tests/extensions_tests/test_parameter_statistics.py
```diff
@@ -3,68 +3,166 @@
...
+            elif self.report_params:
+                assert 'data' in name
```
niboshi: How about checking with a capture of regex?

tests/chainer_tests/training_tests/extensions_tests/test_parameter_statistics.py
```diff
@@ -3,68 +3,166 @@
...
+            elif self.report_grads:
+                assert 'grad' in name
```
niboshi: Ditto.

chainer/training/extensions/parameter_statistics.py
hvy: Fixed!

tests/chainer_tests/training_tests/extensions_tests/test_parameter_statistics.py
hvy: Removed the redundant check.

tests/chainer_tests/training_tests/extensions_tests/test_parameter_statistics.py
hvy: Thanks, fixed.

tests/chainer_tests/training_tests/extensions_tests/test_parameter_statistics.py
hvy: Fixed!

tests/chainer_tests/training_tests/extensions_tests/test_parameter_statistics.py
hvy: Oh, thanks for pointing out. Fixed.

hvy:
Thanks for the comments. They should be fixed, could you take another look?

tests/chainer_tests/training_tests/extensions_tests/test_parameter_statistics.py
```diff
@@ -143,15 +143,17 @@ def test_report_key_pattern(self):
 
         pattern = r'^(.+/){2,}(data|grad)/.+[^/]$'
         for name in six.iterkeys(self.trainer.observation):
+            if self.prefix is not None:
+                assert name.startswith(self.prefix)
+
+            match = re.search(pattern, name)
```
niboshi: Should use `re.match`.

hvy: Fixed.

niboshi:
LGTM!

PR #4346, opened 2018-02-15
Fix error messages for incompatible arrays

niboshi:
Now that `ideep.mdarray` is introduced, current error message is not preciese.


okuta:
LGTM!

PR #4348, opened 2018-02-15
Cast to int type in split_axis

okuta:
Fix bug when we use Python3 with iDeep.


niboshi:
In which case it causes the problem?
Can you add a test case?

okuta:
Already default test is broken in my environment.

okuta:
I added test.

niboshi:
LGTM!

PR #4349, opened 2018-02-15
Fix BatchMatMulGrad.backward

toslunar:
If fewer indices are needed in `BatchMatMulGrad.backward`, it may cause `UnboundLocalError: local variable 'ggab' referenced before assignment`.  The bug was introduced in #3768.

Please review this after #4331 is merged.


okuta:
Please rebase

toslunar:
Rebased.

I'll make a new PR (that corresponds to the first commit) to fix this issue in v3.

okuta:
LGTM!

PR #4350, opened 2018-02-16
Prefer data type objects over character codes in the NumPy

vilyaair:
According to [NumPy reference](https://docs.scipy.org/doc/numpy/reference/arrays.scalars.html#built-in-scalar-types), use of the character codes is discouraged. Accordingly, I change the character codes in favor of data type object in the _documents_ and the _examples_. Nevertheless, the character codes is used so extensively in the internal codes that the benefit of modifying character codes is trivial compared to heavy workloads and potential bugs.

Take type check as an example:
https://github.com/chainer/chainer/blob/7af2e141c7f58d54af77af9ea45b5274b18ebd16/docs/source/tutorial/type_check.rst#L113

This snippet of code could be changed to `utils.type_check.expect(numpy.issubdtype(x_type, numpy.floating))`. Likewise, to check whether the type is integer, we can use `numpy.issubdtype(x_type, numpy.integer)`. However, these modifications, I believe, do not justify their advantages.


okuta:
Thanks.

jenkins, test this please.

okuta:
LGTM!

toslunar:
https://docs.scipy.org/doc/numpy/reference/arrays.scalars.html is the reference of numpy *scalars*.  The reference of dtype (https://docs.scipy.org/doc/numpy/reference/arrays.dtypes.html) says:
> Note that the scalar types are not dtype objects, even though they can be used in place of one whenever a data type specification is needed in NumPy.

I'm not sure if numpy discourages the use of character codes in array creation, while I agree that numpy discourages creation of scalar type using character code as
```
np.dtype('f').type(1.0)  # => np.float32(1.0)
```

vilyaair:
I am positive that NumPy prefers data type objects over character codes with respect to array types.
The [NumPy User Guide](https://docs.scipy.org/doc/numpy/user/basics.types.html) suggests:

> Array types can also be referred to by character codes, mostly to retain backward compatibility with older packages such as Numeric. Some documentation may still refer to these, for example:
```
>>> np.array([1, 2, 3], dtype='f')
array([ 1.,  2.,  3.], dtype=float32)
```
>We recommend using dtype objects instead.

toslunar:
Thank you, @vilyaair.  You are right and I'm sorry.

vilyaair:
@toslunar  It is unnecessary for you to feel sorry. Conversely, I am very pleased that you question this PR to make it better, and make me more insightful toward NumPy data types.

PR #4352, opened 2018-02-16
Detect backward of old-style functions in chainer.grad

### LINKED ISSUE #4114: No error or warning with enable_double_backprop=True and chainer.Function
There is no errors or warnings when `Variable.backward` or `chainer.grad` is called with `enable_double_backprop=True` for a computational graph that contains `chainer.Function`. Backprop from the obtained gradients results in wrong gradients.

I think Chainer should warn or raise an error for such cases.

toslunar:
Fix the issue that `chainer.grad` ignores `Variable._old_style_grad_generator`.

Close #4114.

### DIFF
--- a/chainer/function_node.py
+++ b/chainer/function_node.py
@@ -744,6 +744,10 @@ def grad(outputs, inputs, grad_outputs=None, grad_inputs=None, set_grad=False,
             continue
         visited_funcs.add(func)
         for x in func.inputs:
+            # Raise error here if x is created by Function.backward.
+            # In such case, we don't know exact inputs of the creator.
+            x._check_old_style_gradient()
+
             if not x.requires_grad:
                 continue
             forward_graph[x].append(func)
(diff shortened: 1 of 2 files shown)


toslunar:
The PR is not a pure bugfix, but the change to safe side.

It's possible to ignore an old-style function whose backward doesn't depend on any of xs in `chianer.grad(ys, xs)`.  In other words, some double-backprop having been allowed can be correct even if Chainer doesn't know the correctness.

It's hard, however, to know which variables `Function.backward` really depends on.  (I suppose Chainer cannot *prove* anyway the correctness for the borderline usage above.)
- To keep the memory-efficiency, it seems to me that the backward should not keep references of `inputs` and `grad_outputs`.
- `Function` may cache any intermediate value other than `retain_inputs` and `retain_outputs`, while such caching is discouraged.

mitmul:
Jenkins, test this please.

mitmul:
LGTM

PR #4366, opened 2018-02-19
Make install command more concise

vilyaair:
I hope that this change will not engender installation failure on various platforms.


vilyaair:
I hope that last commit  efad3a4 will not ruin continuous integration services.
Moreover, from my limited knowledge, `python setup.py install` and such, are [legacy methods](https://docs.python.org/3.6/install/index.html) of installing Python modules.

niboshi:
Thank you for suggestion!
We should indeed encourage `pip` for users.

On the other hand, we should still support old method (`python setup.py install`).
We are installing Chainer in various way in CI for that reason (`.travis.yml` and `appveyor.ml`).

So could you revert those two files?

docs/source/install.rst
```diff
@@ -73,17 +73,13 @@ Install Chainer from source
...
-  $ cd chainer-x.x.x
-  $ python setup.py install
+  $ pip install chainer-x.x.x.tar.gz
 
 You can also install the development version of Chainer from a cloned Git repository::
 
-  $ git clone https://github.com/chainer/chainer.git
-  $ cd chainer
-  $ python setup.py install
+  $ pip install git+https://github.com/chainer/chainer.git
```
niboshi: For git installation, users may want to modify Chainer code.
How about keeping cloning part? i.e.
```
$ git clone https://github.com/chainer/chainer.git
$ cd chainer
$ pip install .
```

docker/python2/Dockerfile
```diff
@@ -8,4 +8,4 @@ RUN apt-get update -y && \
     python-setuptools && \
     rm -rf /var/lib/apt/lists/* /var/cache/apt/archives/*
 
-RUN pip install cupy==4.0.0b4 chainer==4.0.0b4
+RUN pip install chainer[cuda]==4.0.0b4
```
kmaehashi: We're going to use wheel in Dockerfile of Chainer/CuPy, so could you please keep as it is for now?
https://github.com/cupy/cupy/issues/967

vilyaair: Definitely, sorry for any inconvenience.

vilyaair:
With respect to img package, [imp package has been deprecated since Python 3.4](https://docs.python.org/3/library/imp.html), and what's worse, `imp.load_source` is undocumented. Nevertheless, my known solution is ugly and tricky: https://stackoverflow.com/a/67692

niboshi:
@vilyaair 
I think it's ok to leave it for now.
We've never heard of any problem regarding this so far.

jenkins, test this please.

niboshi:
LGTM!

PR #4368, opened 2018-02-19
Fix test of multiprocess_parallel_updater

niboshi:
`test_multiprocess_parallel_updater` does not specify appropriate test marks.

### DIFF
--- a/tests/chainer_tests/training_tests/updaters_tests/test_multiprocess_parallel_updater.py
+++ b/tests/chainer_tests/training_tests/updaters_tests/test_multiprocess_parallel_updater.py
@@ -290,4 +290,5 @@ if __name__ == '__main__':
     def test_multi_device(self):
         self.check_update_uses_raw_array(2)
 
+
 testing.run_module(__name__, __file__)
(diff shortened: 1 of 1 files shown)


okuta:
Thank you for your fixing.
I will check this.

kmaehashi:
I think it is better to place the code string as a separate file, so that flake8 can be applied. (as long as there're no side effects)

niboshi:
I agree, but we need a uniform convention to place such files.
e.g.
* What's the directory name / file name convention?
* What if .PNG files are required instead of .py files?

I think it's better to postpone to another issue.

kmaehashi:
We already have `*.png` in https://github.com/chainer/chainer/tree/master/tests/chainer_tests/datasets_tests/image_dataset, but I think it's OK to do that in separate PR anyway.

niboshi:
Oh, I see...
Then we can follow that in some way.
Anyway I'd like to seperate that in another PR, because it's essentially different topic (and the change is not trivial).

okuta:
jenkins, test this please.

okuta:
LGTM. I will merge soon.

PR #4376, opened 2018-02-20
Revert "Skip multiprocess_parallel_updater test for timeout"

niboshi:
Reverts chainer/chainer#4370

This workaround is no longer needed, as https://github.com/cupy/cupy/pull/954 is reverted (in https://github.com/cupy/cupy/pull/964).


kmaehashi:
LGTM! 
I'll merge this (after AppVeyor pass) so that fixes in #4368 can be confirmed.

PR #4378, opened 2018-02-21
Optimize imports for PyCharm code completion

vilyaair:
PyCharm provides confusing code completion due to redundant imports. If you care about the sequence of code completion,  it is advised to [clean PyCharm system cache](https://www.jetbrains.com/help/pycharm/cleaning-system-cache.html). For example,

Original code completion:
original

Code completion after modifications:
modified

This PR, I believe, will make code completion less confusing and less prone to errors for programmers.

Furthermore, I will appreciate it if anyone would tell me the disadvantages of these changes. I guess that it will not generate all cached compiled byte-code files when importing modules.


kmaehashi:
I confirmed that the result of the following are the same (except for orders), with/without this PR.

```py
import sys
import chainer
for m in sys.modules.keys():
    print(m)
```

As for the latter, I think it has no negative effects. But it's better if such code (relying on implicit import) can be detected automatically...

vilyaair:
@kmaehashi Python indeed loads these modules, but code completion of PyCharm recognizes only modules mentioned in the corresponding `__init__.py`. On the contrary, the code completion of iPython will display all loaded modules.

vilyaair:
At present, most problems concerning PyCharm code completion have been addressed. However, there are a few remaining issues:

- Will this PR cause negative effects?

- Should some modules be added in the corresponding `__init__.py`? e.g. [`chainer/utils/argument.py`](https://github.com/chainer/chainer/blob/master/chainer/utils/argument.py)

- Related third party modules make PyCharm unable to find reference. e.g. https://github.com/chainer/chainer/blob/648928409187612e2c649bc293db0f6257cfabea/chainer/gradient_check.py#L271
`import six` -> `import six.moves` could resolve this problem. Is it deserved to make such modifications?

- [`chainer/backends/cuda_fusion.py`](https://github.com/chainer/chainer/blob/master/chainer/backends/cuda_fusion.py) requires mechanism to prevent the exception due to import CuPy, which is similar to [`chainer/backends/cuda.py`](https://github.com/chainer/chainer/blob/master/chainer/backends/cuda.py). It may be unnecessary because of this snippet of codes: https://github.com/chainer/chainer/blob/648928409187612e2c649bc293db0f6257cfabea/chainer/backends/cuda.py#L52

vilyaair:
I have encountered another issue with or without this PR. The following snapshot illustrates this issue.
import
PyCharm offers inaccurate code completion. `configuration`, `cuda`, and etc. should not be displayed in this list. The resolution is to modify from `from ... import ...` to `import ...` in all Python file excluding `__init__.py`. Is it merited to address this problem?

chainer/backends/cuda.py
```diff
@@ -36,8 +36,8 @@
 import six
 
 import chainer
-from chainer.backends import intel64
-from chainer.configuration import config
+import chainer.backends.intel64
+import chainer.configuration
```
niboshi: It's too verbose to import every module in this way. Also, it incurs extra symbol look-up cost in runtime.
Could you revert those changes?

vilyaair:
It is not a very time-consuming procedure. I use regular expression to get pretty good results that is high recall and not very high precision. However, there is a merge in those commits, and hence, reverting those commits is challenging. Therefore, I attempt to rewrite this PR.

niboshi:
Ah, it's not mandatory that commit history is clean.
You can use any way to revert the changes.
Thanks!

niboshi:
Hi, what's the situation on this PR? :)
I didn't see any misbehavior, so could you continue to remove newly added aliases?

vilyaair:
I believe that I have remove all aliases which should be deleted.
If some aliases like in the [`chainer/functions/array/__init__.py`](https://github.com/chainer/chainer/blob/4d6e80e30aeb3963e69f56da7957602c06296bb2/chainer/functions/array/__init__.py) are eliminated, PyCharm cannot find certain references, such as `get_item` in the following snippet of codes:
https://github.com/chainer/chainer/blob/4d6e80e30aeb3963e69f56da7957602c06296bb2/chainer/__init__.py#L202

niboshi:
Ah, as I said above, we should not add new aliases just for PyCharm.
We would have to maintain those aliases which is costly and prone to errors.

vilyaair:
OK, I appreciate your advice.

niboshi:
@vilyaair 
Yesterday we squashed a commit in the master branch, thus the history of the branch has been slightly altered.
This was very unusual operatioin in order to eliminate an accidental addition of a huge file.

Unfortunately this PR is affected by that operation.
We are very sorry, but could you rebase this PR onto the current master?

vilyaair:
I could not understand that my push would cause such situation. My rebase result is at my local branch [redundant-imports-rebase](https://github.com/vilyaair/chainer/tree/redundant-imports-rebase). However, while it was pushed, PyCharm prompted me to merge or rebase anyway. If rebase anyway was selected, push would fail. Therefore, I chose to merge, and the result became the current situation.

vilyaair:
Luckily, force push might address this issue, despite I always fear that force push could ruin commit history.

niboshi:
jenkins, test this please

niboshi:
jenkins, test this please

niboshi:
jenkins, test this please

niboshi:
LGTM!
Thank you very much!

PR #4380, opened 2018-02-21
Fix numpy dtypes bug in cuda.to_gpu and cuda.copy

tkerola:
I found a bug in `cuda.to_gpu` and `cuda.copy` that would not change the device when the device identifier is a numpy dtype. Users might use numpy to generate the device ids, and would therefore be subject to this subtle bug.

## MWE
Illustrating the problem for `to_gpu`, but the same thing happens for `copy`.
```
from chainer import cuda
import numpy as np

gpus = [0, 1, 2, 3]
batch_size = 2
eval_gpus = np.repeat(gpus, batch_size)

xs = []
for gpu in eval_gpus:
    y = np.ones(1)
    x = cuda.to_gpu(y, device=gpu)
    xs.append(x)

for x in xs:
    print(x.device)
```

## Output

Without this PR:
```
<CUDA Device 0>
<CUDA Device 0>
<CUDA Device 0>
<CUDA Device 0>
<CUDA Device 0>
<CUDA Device 0>
<CUDA Device 0>
<CUDA Device 0>
```

With this PR (expected):
```
<CUDA Device 0>
<CUDA Device 0>
<CUDA Device 1>
<CUDA Device 1>
<CUDA Device 2>
<CUDA Device 2>
<CUDA Device 3>
<CUDA Device 3>
```

### DIFF
--- a/chainer/backends/cuda.py
+++ b/chainer/backends/cuda.py
@@ -218,8 +218,7 @@ def get_device(*args):
 
 def _get_device(*args):
     for arg in args:
-        if isinstance(arg, _integer_types) and not numpy.issubdtype(
-                type(arg), numpy.bool_):
+        if type(arg) is not bool and isinstance(arg, _integer_types):
             check_cuda_available()
             return Device(arg)
         if isinstance(arg, ndarray):
--- a/chainer/backends/cuda.py
+++ b/chainer/backends/cuda.py
@@ -138,14 +138,13 @@ if available:
     pinned_memory_pool = cupy.get_default_pinned_memory_pool()
 
 
+_integer_types = six.integer_types + (numpy.integer,)
 if six.PY2:
     try:
         from future.types.newint import newint as _newint
-        _integer_types = six.integer_types + (numpy.integer,) + (_newint,)
+        _integer_types += (_newint,)
     except ImportError:
-        _integer_types = six.integer_types + (numpy.integer,)
-else:
-    _integer_types = six.integer_types + (numpy.integer,)
+        pass
 
 
 # ------------------------------------------------------------------------------
@@ -219,7 +218,8 @@ def get_device(*args):
 
 def _get_device(*args):
     for arg in args:
-        if isinstance(arg, _integer_types):
+        if isinstance(arg, _integer_types) and not numpy.issubdtype(
+                type(arg), numpy.bool_):
             check_cuda_available()
             return Device(arg)
         if isinstance(arg, ndarray):
--- a/chainer/backends/cuda.py
+++ b/chainer/backends/cuda.py
@@ -141,11 +141,11 @@ if available:
 if six.PY2:
     try:
         from future.types.newint import newint as _newint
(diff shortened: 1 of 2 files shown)


chainer/backends/cuda.py
```diff
@@ -141,11 +141,11 @@ def __ne__(self, other):
...
     try:
         from future.types.newint import newint as _newint
-        _integer_types = six.integer_types + (_newint,)
+        _integer_types = six.integer_types + (numpy.integer,) + (_newint,)
     except ImportError:
-        _integer_types = six.integer_types
+        _integer_types = six.integer_types + (numpy.integer,)
 else:
-    _integer_types = six.integer_types
+    _integer_types = six.integer_types + (numpy.integer,)
```
niboshi: How about factoring out common part?: 
```
_integer_types = six.integer_types + (numpy.integer,)
if six.PY2:
     _integer_types +=  (_newint,)
```

chainer/backends/cuda.py
```diff
@@ -219,7 +219,7 @@ def get_device(*args):
 
 def _get_device(*args):
     for arg in args:
-        if type(arg) in _integer_types:
+        if isinstance(arg, _integer_types):
```
niboshi: In this way `bool` is evaluated as `True`.
How about excluding `bool` explicitly?

chainer/backends/cuda.py
tkerola: That's a good idea, thanks.

chainer/backends/cuda.py
tkerola: Good catch, done.

chainer/backends/cuda.py
```diff
@@ -219,7 +218,8 @@ def get_device(*args):
 
 def _get_device(*args):
     for arg in args:
-        if type(arg) in _integer_types:
+        if isinstance(arg, _integer_types) and not numpy.issubdtype(
+                type(arg), numpy.bool_):
```
niboshi: Ah, I meant `bool`, not `numpy.bool_`.
(`isinstance(numpy.bool_, numpy.integer)` and `isinstance(numpy.bool_, six.integer_types)`) are both `False`, but `isinstance(bool, six.integer_types)` is `True`).

So the correct condition would be:
`if type(arg) is not bool and isinstance(arg, _integer_types):`

tkerola: Isn't my condition also correct? It works for both `numpy.bool`, `numpy.bool_` and `bool`.

niboshi: I see. So the outcome would be the same, but it does duplicate check and slower (`numpy.bool_` never satisfies the first condition). Also note that `numpy.bool` and `bool` are identical.

tkerola: Aha, bool and np.bool are identical. Then your solution is cleaner. Thanks! I will update it.

niboshi:
Thank you for the fix.
jenkins, test this please

niboshi:
LGTM!

PR #4041, opened 2017-12-05
Add shift function

tkerola:
This PR implements the shift function, proposed in the following paper: https://arxiv.org/abs/1711.08141

It is a useful alternative to spatial convolutions for reducing the number of model parameters.

PR #4383, opened 2018-02-22
[backport] add F.shift to docs

niboshi:
Backport of #4379


kmaehashi:
Thanks, but `F.shift` was introduced in https://github.com/chainer/chainer/pull/4041 and was not in backported to v3 branch.

PR #4395, opened 2018-02-22
Added numexpr backend. Supersedes #3937

mihirparadkar:
Linked: #3937 
I opened a new pull request because I wanted to merge a different branch.


chainer/backends/numexpr.py
```diff
@@ -0,0 +1,103 @@
...
+
+import numpy
+
+_numexpr_version = None
+_error = None
+
+try:
+    import numexpr  # NOQA
+    _numexpr_version = 0
+    numexpr_dtypes = {
```
niboshi: Prefix this variable with underscore (`_`).

mihirparadkar:
It's WIP in the same sense as the IDeep backend, in that it doesn't support most of the activations and optimizers. However, as you said, it does work for the places it is implemented. I'll fix those minor things shortly.

niboshi:
"WIP" in the PR title means "this PR is not ready for merge". In that sense, iDeep integration is no longer WIP (because it's already merged).

For this PR, it's OK to merge with minimal support and work on more implementation after that.

mihirparadkar:
In that case, "witness the firepower of this fully armed and operational numexpr backend" :)

Issue #4377, opened 2018-02-20
Sparse matmul


anaruse:
I submitted a PR to Chainer-chemistry (https://github.com/pfnet-research/chainer-chemistry/pull/90) on sparse matmul for graph convolution. And now I'm considering how to generalize it for Chainer, because the implementation of sparse matmul in the PR above only supports a use case of "dense C = sparse A * dense B" and it is not general purpose. In addition, computation of gradient of sparse matrix is not supported.

Now, I think following use cases and corresponding APIs are necessary.

1. dense C = sparse A * dense B: `sparse_matmul_sp_dn(a_data, a_row, a_col, a_shape, b)`
2. dense C = dense A * sparse B: `sparse_matmul_dn_sp(a, b_data, b_row, b_col, b_shape)`
3. sparse C = dense A * dense B: `sparse_matmul_dn_dn(a, b, c_row, c_col, c_shape)`

Regarding sparse matrix format, I think that COO format with kind of SoA layout in which data array, row index array and colum index array are kept separately as `xp.ndarray` or `chainer.Variable` is practical choice. I considered a use of `scipy.sparse.coo_matrix`, but I think it is not available here because `chainer.Variable` accepts only `xp.ndarray` and does not accept arbitrary class (including `scipy.sparse.coo_matrix`) as its data. I also considered a use of numpy's structured array but seems it is not supported by cupy.

Regarding use cases, the case 1. and 2. are interchangeable by transpose, but I think it is better to provide two APIs for ease of use. You probably wonder why the case 3. is necessary. This is mainly for computation of gradient of sparse matrix. I'm not sure this should be exposed to users, so a private function might be sufficient for this.

What would you think on this? Any feedback would be appreciated.

anaruse:
Or is it better that it exposes one function `sparse_matmul` for three use cases and let users make sparse matrix object before calling it?

1. dense C = sparse A * dense B:

       sp_a = mk_sparse_matrix(a_data, a_row, a_col, a_shape)
       c = sparse_matmul(sp_a, b)

2. dense C = dense A * sparse B:

       sp_b = mk_sparse_matrix(b_data, b_row, b_col, b_shape)
       c = sparse_matmul(a, sp_b)

3. sparse C = dense A * dense B:

       sp_c = mk_sparse_matrix(None, c_row, c_col, c_shape)
       sp_c = sparse_matmul(a, b, sp_c)

kmaehashi:
Implemented via #4397.

PR #4397, opened 2018-02-23
Sparse matmul support

anaruse:
This PR aims to support sparse matmul in Chainer (this is related to https://github.com/chainer/chainer/issues/4377 and https://github.com/pfnet-research/chainer-chemistry/pull/90).

I implemented a function named `sparse_matmul` which computes matrix multiplication of sparse and dense matrix. The usage of this function is as follows (assuming `a` and `b` are matrix or 3D tensor).

    sp_a = F.sparse_dense2coo(a)
    c = F.sparse_matmul(sp_a, b)

You can also use this function for batched sparse matrix multiplication (actually this is my main focus) like `matmul`. It supports backward and double-backward, so that you can compute gradients of sparse and dense matrix and gradients of gradients as well.

Please note that CPU version is not implemented now because I don't have good idea on efficient CPU implementation using numpy or scipy for batched sparse matrix multiplication.

### DIFF
--- a/chainer/functions/__init__.py
+++ b/chainer/functions/__init__.py
@@ -235,6 +235,9 @@ from chainer.functions.math.prod import prod  # NOQA
 from chainer.functions.math.prod import Prod  # NOQA
 from chainer.functions.math.scale import scale  # NOQA
 from chainer.functions.math.sign import sign  # NOQA
+from chainer.functions.math.sparse_matmul import sparse_coo_matrix  # NOQA
+from chainer.functions.math.sparse_matmul import sparse_dense2coo  # NOQA
+from chainer.functions.math.sparse_matmul import sparse_matmul  # NOQA
 from chainer.functions.math.sqrt import rsqrt  # NOQA
 from chainer.functions.math.sqrt import sqrt  # NOQA
 from chainer.functions.math.sqrt import Sqrt  # NOQA
--- a/chainer/functions/math/sparse_matmul.py
+++ b/chainer/functions/math/sparse_matmul.py
@@ -24,10 +24,12 @@ class sparse_coo_matrix(object):
             return x
         if self.data.ndim == 2:
             nb = self.data.shape[0]
-            x = xp.zeros((nb, self.shape[0], self.shape[1]), dtype=self.data.dtype)
+            x = xp.zeros((nb, self.shape[0], self.shape[1]),
+                         dtype=self.data.dtype)
             for i in range(nb):
                 nnz = len(xp.where(self.data[i] != 0)[0])
-                x[i, self.row[i, :nnz], self.col[i, :nnz]] = self.data[i, 0:nnz]
+                x[i, self.row[i, :nnz], self.col[i, :nnz]] = self.data[i,
+                                                                       0:nnz]
             return x
 
 
@@ -66,7 +68,7 @@ def sparse_dense2coo(x, ldnz=None, use_variable=False):
     else:
         raise ValueError('')
 
-    
+
 def _sparse_matmul(sp_data, sp_row, sp_col, sp_shape, dn,
                    transa, transb, transc):
     xp = cuda.get_array_module(dn)
@@ -90,20 +92,21 @@ def _sparse_matmul(sp_data, sp_row, sp_col, sp_shape, dn,
     C_dtype = B.dtype
     if B.dtype == xp.float16:
         C_dtype = xp.float32
-        
+
     _m, _ = A_shape
     if dn.ndim == 2:
         nb = 1
         _k, _n = B.shape
         nnz, = A_data.shape
-        C =  xp.zeros((_m, _n), dtype=C_dtype)
+        C = xp.zeros((_m, _n), dtype=C_dtype)
     else:
         nb, _k, _n = B.shape
         _, nnz = A_data.shape
-        C =  xp.zeros((nb, _m, _n), dtype=C_dtype)
+        C = xp.zeros((nb, _m, _n), dtype=C_dtype)
 
     size = nb * nnz * _n
-    _cupy_sparse_matmul()(nb, _m, _n, _k, nnz, A_data, A_row, A_col, B, C, size=size)
+    _cupy_sparse_matmul()(nb, _m, _n, _k, nnz, A_data, A_row, A_col, B, C,
+                          size=size)
 
     if B.dtype == xp.float16:
         C = C.astype(B.dtype)
@@ -146,14 +149,16 @@ def _cupy_sparse_matmul():
 class SparseMatMul(function_node.FunctionNode):
 
     def __init__(self, sp_row, sp_col, sp_shape,
-                 transa=False, transb=False, transc=False):
+                 transa=False, transb=False, transc=False,
+                 dtype=None):
         if sp_row.ndim != sp_col.ndim:
             raise ValueError('ndim of sp_row and sp_col must be the same.')
         if sp_row.ndim != 1 and sp_row.ndim != 2:
             raise ValueError('ndim of sp_row and sp_col must be one or two.')
         for i in range(sp_row.ndim):
             if sp_row.shape[i] != sp_col.shape[i]:
-                raise ValueError('shape of sp_row and sp_col must be the same.')
+                _msg = 'shape of sp_row and sp_col must be the same.'
+                raise ValueError(_msg)
         if len(sp_shape) != 2:
             raise ValueError('len(sp_shape) must be two.')
         self.sp_row = sp_row
@@ -162,6 +167,7 @@ class SparseMatMul(function_node.FunctionNode):
         self.transa = transa
         self.transb = transb
         self.transc = transc
+        self.dtype = dtype
 
     def check_type_forward(self, in_types):
         type_check.expect(in_types.size() == 2)
@@ -177,15 +183,15 @@ class SparseMatMul(function_node.FunctionNode):
             type_check.expect(
                 sp_type.shape[0] == self.sp_row.shape[0],
                 dn_type.shape[0] == self.sp_row.shape[0],
-            )            
+            )
 
     def forward(self, inputs):
         self.retain_inputs((0, 1))
         sp, dn = inputs
         c = _sparse_matmul(sp, self.sp_row, self.sp_col, self.sp_shape, dn,
                            self.transa, self.transb, self.transc)
-        return utils.force_array(c),
-        
+        return utils.force_array(c, self.dtype),
+
     def backward(self, indexes, grad_outputs):
         sp, dn = self.get_retained_inputs()
         g_c, = grad_outputs
@@ -195,15 +201,17 @@ class SparseMatMul(function_node.FunctionNode):
             g_sp = SparseMatMulGradSP(self.sp_row, self.sp_col, self.sp_shape,
                                       self.transc,
                                       not self.transb,
-                                      self.transa).apply((g_c, dn))[0]
-    
+                                      self.transa,
+                                      dtype=sp.dtype).apply((g_c, dn))[0]
+
         g_dn = None
         if 1 in indexes:
             g_dn = SparseMatMul(self.sp_row, self.sp_col, self.sp_shape,
                                 not self.transa,
                                 self.transc,
-                                self.transb).apply((sp, g_c))[0]
-                
+                                self.transb,
+                                dtype=dn.dtype).apply((sp, g_c))[0]
+
         return g_sp, g_dn
 
 
@@ -212,7 +220,7 @@ def _sparse_matmul_gradsp(a, b, c_row, c_col, c_shape,
     xp = cuda.get_array_module(a)
     if xp is numpy:
         raise RuntimeError('_sparse_matmul_gradsp is available only on GPU')
-    
+
     if transa:
         A = a.swapaxes(-1, -2)
     else:
@@ -240,12 +248,13 @@ def _sparse_matmul_gradsp(a, b, c_row, c_col, c_shape,
         _, nnz = C_row.shape
         C_data = xp.zeros((nb, nnz), dtype=A.dtype)
 
-    size = nb * nnz;
-    _cupy_sparse_matmul_gradsp()(nb, _m, _n, _k, nnz, A, B, C_data, C_row, C_col, size=size)
+    size = nb * nnz
+    _cupy_sparse_matmul_gradsp()(nb, _m, _n, _k, nnz, A, B,
+                                 C_data, C_row, C_col, size=size)
 
     return C_data
 
-    
+
 def _cupy_sparse_matmul_gradsp():
     return cuda.cupy.ElementwiseKernel(
         'int32 nb, int32 _m, int32 _n, int32 _k, int32 nnz, \
@@ -285,14 +294,16 @@ def _cupy_sparse_matmul_gradsp():
 class SparseMatMulGradSP(function_node.FunctionNode):
 
     def __init__(self, sp_row, sp_col, sp_shape,
-                 transa=False, transb=False, transc=False):
+                 transa=False, transb=False, transc=False,
+                 dtype=None):
         if sp_row.ndim != sp_col.ndim:
             raise ValueError('ndim of sp_row and sp_col must be the same.')
         if sp_row.ndim != 1 and sp_row.ndim != 2:
             raise ValueError('ndim of sp_row and sp_col must be one or two.')
         for i in range(sp_row.ndim):
             if sp_row.shape[i] != sp_col.shape[i]:
-                raise ValueError('shape of sp_row and sp_col must be the same.')
+                _msg = 'shape of sp_row and sp_col must be the same.'
+                raise ValueError(_msg)
         if len(sp_shape) != 2:
             raise ValueError('len(sp_shape) must be two.')
         self.sp_row = sp_row
@@ -301,7 +312,8 @@ class SparseMatMulGradSP(function_node.FunctionNode):
         self.transa = transa
         self.transb = transb
         self.transc = transc
-        
+        self.dtype = dtype
+
     def check_type_forward(self, in_types):
         type_check.expect(in_types.size() == 2)
         a_type, b_type = in_types
@@ -317,15 +329,16 @@ class SparseMatMulGradSP(function_node.FunctionNode):
                 a_type.shape[0] == self.sp_row.shape[0],
                 b_type.shape[0] == self.sp_row.shape[0],
             )
-                
+
     def forward(self, inputs):
         self.retain_inputs((0, 1))
         a, b = inputs
-        c = _sparse_matmul_gradsp(a, b, self.sp_row, self.sp_col, self.sp_shape,
+        c = _sparse_matmul_gradsp(a, b,
+                                  self.sp_row, self.sp_col, self.sp_shape,
                                   self.transa, self.transb, self.transc)
-        return utils.force_array(c),
+        return utils.force_array(c, self.dtype),
 
-        
+
 def sparse_matmul(a, b, transa=False, transb=False):
     """Computes the batched multiplication of sparse and dense matrix.
 
@@ -341,16 +354,16 @@ def sparse_matmul(a, b, transa=False, transb=False):
         transb (bool): If ``True``, each matrix in ``b`` will be transposed.
 
     Returns:
-        _chainer.Variable or sparse_coo_matrix: Result of matrix multiplication.
+        _chainer.Variable or sparse_coo_matrix: Result of mat-mul.
     """
     if (isinstance(a, sparse_coo_matrix) and
-          isinstance(b, (chainer.Variable, numpy.ndarray, cuda.ndarray))):
+            isinstance(b, (chainer.Variable, numpy.ndarray, cuda.ndarray))):
         return SparseMatMul(a.row, a.col, a.shape,
                             transa=transa,
                             transb=transb,
                             transc=False).apply((a.data, b))[0]
     elif (isinstance(a, (chainer.Variable, numpy.ndarray, cuda.ndarray)) and
-        isinstance(b, sparse_coo_matrix)):
+          isinstance(b, sparse_coo_matrix)):
         return SparseMatMul(b.row, b.col, b.shape,
                             transa=not transb,
                             transb=not transa,
--- a/chainer/functions/math/sparse_matmul.py
+++ b/chainer/functions/math/sparse_matmul.py
@@ -1,0 +1,360 @@
+import chainer
+from chainer.backends import cuda
+from chainer import function_node
+from chainer import utils
+from chainer.utils import type_check
+import numpy
+
+
+class sparse_coo_matrix(object):
+
+    def __init__(self, data, row, col, shape, use_variable=False):
+        self.data = data
+        if use_variable:
+            self.data = chainer.Variable(self.data)
+        self.row = row
+        self.col = col
+        self.shape = shape  # (row, col)
+
+    def to_dense(self):
+        xp = cuda.get_array_module(self.data)
+        if self.data.ndim == 1:
+            x = xp.zeros(self.shape, dtype=self.data.dtype)
+            x[self.row, self.col] = self.data
+            return x
+        if self.data.ndim == 2:
+            nb = self.data.shape[0]
+            x = xp.zeros((nb, self.shape[0], self.shape[1]), dtype=self.data.dtype)
+            for i in range(nb):
+                nnz = len(xp.where(self.data[i] != 0)[0])
+                x[i, self.row[i, :nnz], self.col[i, :nnz]] = self.data[i, 0:nnz]
+            return x
+
+
+def sparse_dense2coo(x, ldnz=None, use_variable=False):
+    xp = cuda.get_array_module(x)
+    if x.ndim == 2:
+        _row, _col = xp.where(x != 0)
+        nnz = len(_row)
+        if ldnz is None or ldnz < nnz:
+            ldnz = nnz
+        data = xp.zeros((ldnz), dtype=x.dtype)
+        row = xp.full((ldnz), -1, dtype=xp.int32)
+        col = xp.full((ldnz), -1, dtype=xp.int32)
+        data[0:nnz] = x[_row, _col]
+        row[0:nnz] = xp.array(_row).astype(xp.int32)
+        col[0:nnz] = xp.array(_col).astype(xp.int32)
+        shape = x.shape
+        return sparse_coo_matrix(data, row, col, shape, use_variable)
+    elif x.ndim == 3:
+        # first axis is considered as batch axis
+        nb = x.shape[0]
+        if ldnz is None:
+            ldnz = 0
+        for i in range(nb):
+            ldnz = max(ldnz, len(xp.where(x[i] != 0)[0]))
+        data = xp.empty((nb, ldnz), dtype=x.dtype)
+        row = xp.empty((nb, ldnz), dtype=xp.int32)
+        col = xp.empty((nb, ldnz), dtype=xp.int32)
+        for i in range(nb):
+            coo = sparse_dense2coo(x[i], ldnz)
+            data[i] = coo.data
+            row[i] = coo.row
+            col[i] = coo.col
+        shape = x.shape[1:]
+        return sparse_coo_matrix(data, row, col, shape, use_variable)
+    else:
+        raise ValueError('')
+
+    
+def _sparse_matmul(sp_data, sp_row, sp_col, sp_shape, dn,
+                   transa, transb, transc):
+    xp = cuda.get_array_module(dn)
+    if xp is numpy:
+        raise RuntimeError('_sparse_matmul is available only on GPU')
+
+    A_data = sp_data
+    if transa:
+        A_row = sp_col
+        A_col = sp_row
+        A_shape = [sp_shape[1], sp_shape[0]]
+    else:
+        A_row = sp_row
+        A_col = sp_col
+        A_shape = sp_shape
+    if transb:
+        B = dn.swapaxes(-1, -2)
+    else:
+        B = dn
+
+    C_dtype = B.dtype
+    if B.dtype == xp.float16:
+        C_dtype = xp.float32
+        
+    _m, _ = A_shape
+    if dn.ndim == 2:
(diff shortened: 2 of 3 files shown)


anaruse:
I've updated the branch, so that you can use `sparse_matmul` on CPU.

hvy:
jenkins, test this please.

chainer/functions/__init__.py
```diff
@@ -235,6 +235,9 @@
 from chainer.functions.math.prod import Prod  # NOQA
 from chainer.functions.math.scale import scale  # NOQA
 from chainer.functions.math.sign import sign  # NOQA
+from chainer.functions.math.sparse_matmul import sparse_coo_matrix  # NOQA
```
hvy: Should this class be exposed through the `chainer.functions` module?

chainer/functions/math/sparse_matmul.py
```diff
@@ -0,0 +1,486 @@
+import chainer
+from chainer.backends import cuda
+from chainer import function_node
+from chainer import utils
+from chainer.utils import type_check
+import numpy
+import warnings
```
hvy: Could you reorder the imports? Here's an example https://github.com/chainer/chainer/blob/master/chainer/utils/array.py ?

chainer/functions/math/sparse_matmul.py
```diff
@@ -0,0 +1,486 @@
...
+
+try:
+    from scipy import sparse
+except ImportError:
+    warnings.warn("SciPy seems not available on your system. A CPU"
```
hvy: Could you cache the failure and raise a warning when `scipy.sparse` is first called instead? E.g. like https://github.com/chainer/chainer/blob/master/chainer/datasets/svhn.py

chainer/functions/math/sparse_matmul.py
```diff
@@ -0,0 +1,486 @@
...
+                  " implementation of sparse_matmul uses SciPy, so that you"
+                  " cannot use sparse_matmul on CPU.")
+
+
+class sparse_coo_matrix(object):
```
hvy: Could you rename this class using CamelCase?

hvy:
It seems like `atomicAdd` is called with a double (which is not supported). Could you double check?

hvy:
I added some comments but I now see we'd like to discuss the design.

We probably want to generalize it as much as possible for Chainer, @corochann do you see any concerns or have any ideas how to go from here? I'd be happy to get any updates on the discussion from https://github.com/pfnet-research/chainer-chemistry/pull/90 .

chainer/functions/__init__.py
anaruse: Perhaps it will be used from users though it is not used even from the tests. I remote it from `__init__.py` now and will add it again when it is certainly necessary.

chainer/functions/math/sparse_matmul.py
anaruse: will fix in that way!

chainer/functions/math/sparse_matmul.py
anaruse: will use CamelCase for the class name.

anaruse:
Regarding `atomicAdd` with double precision, it is available on Pascal and Volta GPUs, but not available on Maxwell and older GPUs. I forgot this. will fix code, so that `atomicCAS` is used when GPU is Maxwell or older one.

anaruse:
Thank you for your comments, @hvy.  I've fixed the branch based on your suggestions.  Please review again.

anaruse:
I've just submitted a PR to CuPy on support of double precision `atomicAdd` on Maxwell or older GPUs. It is highly related to CUDA hence I think it should be done in CuPy.

hvy:
Thank you for the fixes and the double precision support in CuPy. I'm sorry for the delayed response and will take another look today and tomorrow.

hvy:
I just discussed with @beam2d how to expose the conversion function `sparse_dense2coo` and the wrapper class `SparseCooMatrix`. Would it be possible to expose them under `chainer.utils` instead of where they currently are under `chainer.functions`, as these definitions are not functions themselves but **used** by them.

chainer/functions/math/sparse_matmul.py
```diff
@@ -0,0 +1,512 @@
...
+try:
+    from scipy import sparse
+    _scipy_available = True
+except ImportError:
+    _scipy_available = False
+
+
+class SparseCooMatrix(object):
+
+    def __init__(self, data, row, col, shape, use_variable=False):
```
hvy: When is `use_variable=True` needed? In my understanding, gradients are not propagated to the sparse matrix.

chainer/functions/math/sparse_matmul.py
```diff
@@ -0,0 +1,512 @@
...
+        self.data = data
+        if use_variable:
+            self.data = chainer.Variable(self.data)
+        self.row = row
+        self.col = col
+        self.shape = shape  # (row, col)
```
hvy: Can we afford any data validation here, e.g. since `to_dense` depends on `ndim`.

chainer/functions/math/sparse_matmul.py
```diff
@@ -0,0 +1,512 @@
...
+
+    def to_dense(self):
```
hvy: How about adding tests for this method?

chainer/functions/math/sparse_matmul.py
```diff
@@ -0,0 +1,512 @@
...
+        if self.data.ndim == 1:
+            x = xp.zeros(self.shape, dtype=self.data.dtype)
+            x[self.row, self.col] = self.data
+            return x
+        if self.data.ndim == 2:
+            nb = self.data.shape[0]
+            x = xp.zeros((nb, self.shape[0], self.shape[1]),
+                         dtype=self.data.dtype)
+            for i in range(nb):
+                nnz = len(xp.where(self.data[i] != 0)[0])
```
hvy: How about using `xp.count_nonzero(self.data[i])`

chainer/functions/math/sparse_matmul.py
```diff
@@ -0,0 +1,512 @@
...
+def _sparse_matmul(sp_data, sp_row, sp_col, sp_shape, dn,
+                   transa, transb, transc, dtype=None):
+    if dtype is None:
+        dtype = numpy.result_type(sp_data.dtype, dn.dtype)
+
+    A_data = sp_data
+    if transa:
+        A_row = sp_col
+        A_col = sp_row
+        A_shape = [sp_shape[1], sp_shape[0]]
```
hvy: For convention, how about using a tuple instead of a list.

chainer/functions/math/sparse_matmul.py
```diff
@@ -0,0 +1,512 @@
...
+
+
+def _sparse_matmul_cpu(A_data, A_row, A_col, A_shape, B, dtype):
+    # A_shape: (_m, _k)
+    # B.shape: ((nb,) _k, _n)
+    # A_data/row/col.shape: ((nb,) ldnz)
+    if not _scipy_available:
+        msg = "SciPy seems not available on your system. A CPU" \
+              " implementation of sparse_matmul uses SciPy, so that you" \
+              " cannot use sparse_matmul on CPU."
```
hvy: Some minor grammatical comments but could you change to `SciPy seems to be unavailable on you system. ...` and `...SciPy, so you cannot use sparse_matmul on the CPU`. Sorry if nitpicking.

chainer/functions/math/sparse_matmul.py
anaruse: It is used when grads of sparse matrix are necessary as you thought.  I've heard from a few researchers that there are applications in which grads of sparse matrix are needed.

hvy: To what extent do you think we should support it (in this PR)? I talked with @beam2d and concluded that we could wait with this feature, because we need to think though the interface and maybe also include differentiable conversions (`FunctionNode` implementations) between dense and sparse representations?

chainer/functions/math/sparse_matmul.py
anaruse: All right, I will add some validation code here for `shape` among others.

chainer/functions/math/sparse_matmul.py
anaruse: Thank you!  That is better.

chainer/functions/math/sparse_matmul.py
anaruse: All right, I will add tests for `to_dense`.

chainer/functions/math/sparse_matmul.py
anaruse: Dose that mean that you would not rush to conclude on how to handle gradients of sparse matrix? I understand it. So, should I delete codes that are related to this feature like `SparseMatMulGrandSP` among others from this PR?

anaruse:
Thank you very much for your detailed comments, @hvy. For the moment, I pushed a commit that reflects most comments from you.

hvy:
Thank you for the very quick updates, I'll take a look at the changes.

chainer/functions/math/sparse_matmul.py
hvy: No exactly, that's what we thought. But I am not entirely sure how gradients of `A` in `A(sparse) * B (dense) = C` should be defined. In theory, they become dense and thus `A` after applying an update is no longer a sparse matrix, but this probably depends on the use case. Do you have other sources?

hvy: For the record, MXNet supports [sparse gradient](http://pytorch.org/docs/master/sparse.html) it seems. PyTorch have a sparse module that's experimental and not that well documented. I'm not sure about TensorFlow but there are discussions with varying conclusions. TVM mentions nothing about sparse representations.

anaruse:
cupy/cupy#1071 was merged, so I remove the work-around for double precision `atomicAdd` on Maxwell generation GPUs from this PR.

hvy:
For the record after offline discussions, 
- we do not have to make dense -> coo conversion differentiable for now and putting those features under the utils module is fine.
- we should provide sparse gradients (`data` in the sparse matrix).

chainer/functions/math/sparse_matmul.py
```diff
@@ -0,0 +1,429 @@
...
+        c = _sparse_matmul(sp, self.sp_row, self.sp_col, self.sp_shape, dn,
+                           self.transa, self.transb, self.transc, self.dtype)
+        return utils.force_array(c, self.dtype),
+
+    def backward(self, indexes, grad_outputs):
+        sp, dn = self.get_retained_inputs()
+        g_c, = grad_outputs
+
+        g_sp = None
+        if 0 in indexes:
```
hvy: We usually use follow this idiom when checking the indices. Could you consider using it? https://github.com/chainer/chainer/blob/c2ad311fb043d7fb37acdd97f473651637df595c/chainer/functions/math/trigonometric.py#L49-L54

chainer/functions/math/sparse_matmul.py
```diff
@@ -0,0 +1,429 @@
...
+                                  self.sp_row, self.sp_col, self.sp_shape,
+                                  self.transa, self.transb, self.transc,
+                                  self.dtype)
+        return utils.force_array(c),
+
+    def backward(self, indexes, grad_outputs):
+        a, b = self.get_retained_inputs()
+        g_sp, = grad_outputs
+
+        g_a = None
```
hvy: Ditto, regarding the `ret = []` idiom.

chainer/functions/math/sparse_matmul.py
```diff
@@ -0,0 +1,429 @@
...
+        g_b = None
+        if 0 in indexes:
+            g_b = SparseMatMul(self.sp_row, self.sp_col, self.sp_shape,
+                               not self.transc, self.transa, not self.transb,
+                               dtype=b.dtype).apply((g_sp, a))[0]
+
+        return g_a, g_b
+
+
+def sparse_matmul(a, b, transa=False, transb=False):
```
hvy: I wonder if we should rename this function `batch_sparse_matmul` for consistency with https://docs.chainer.org/en/stable/reference/generated/chainer.functions.batch_matmul.html, or `batch_sparse_dense_matmul` to be more explicit, what do you think?

chainer/utils/sparse.py
```diff
@@ -0,0 +1,73 @@
+import chainer
+from chainer.backends import cuda
+
+
+class SparseCooMatrix(object):
```
hvy: Can it be shortened to `CooMatrix`, more similar to scipy's `coo_matrix`?

chainer/utils/sparse.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def __init__(self, data, row, col, shape, use_variable=False):
```
hvy: What do you think of always wrapping the data in a `Variable` (instead of `use_variable`) and let the user control the gradient with an additional `require_grad` setter/getter property. That would simplify the interface a bit and make it more similar to the `Variable`. The overhead isn't significant and maybe more future proof?

chainer/utils/sparse.py
```diff
@@ -0,0 +1,73 @@
...
+            x = xp.zeros((nb, self.shape[0], self.shape[1]),
+                         dtype=self.data.dtype)
+            for i in range(nb):
+                nnz = xp.count_nonzero(self.data[i])
+                x[i, self.row[i, :nnz],
+                  self.col[i, :nnz]] = self.data[i, :nnz]
+            return x
+
+
+def sparse_dense2coo(x, ldnz=None, use_variable=False):
```
hvy: How about renaming it to `to_coo` to again mimic scipy? Again here too, could we skip `use_variable` ?

tests/chainer_tests/functions_tests/math_tests/test_sparse_matmul.py
```diff
@@ -0,0 +1,293 @@
...
+        sp_a = utils.sparse_dense2coo(a_data)
+        sp_gga = utils.sparse_dense2coo(a_grad_grad)
+        func = F.math.sparse_matmul.SparseMatMul(
+            sp_a.row, sp_a.col, sp_a.shape,
+            transa=self.transa, transb=self.transb, transc=False)
+
+        def op(a, b):
+            return func.apply((a, b))[0]
+        gradient_check.check_double_backward(
+            op, (sp_a.data, b_data), c_grad, (sp_gga.data, b_grad_grad),
```
hvy: Can you test the function interface without using the `FunctionNode` class + `apply`?

tests/chainer_tests/utils_tests/test_sparse.py
```diff
@@ -0,0 +1,34 @@
...
+import numpy
+
+from chainer import testing
+from chainer import utils
+
+
+def _setup_tensor(_min, _max, shape, dtype, threshold=None):
+    y = numpy.random.uniform(_min, _max, shape).astype(dtype)
+    if threshold is not None:
+        y[numpy.where(y < threshold)] = 0
```
hvy: I think you can simplify it to `y[y < threshold] = 0`.

hvy:
@anaruse Thanks for your fixes. I added some comments again, could you take another look?

chainer/functions/math/sparse_matmul.py
corochann: `batch_matmul` is dperecated as written in the document.

```
Deprecated since version v3.0.0:
batch_matmul is deprecated. Use matmul instead.
```

Original numpy `matmul` already acts like "batch_matmul" when input is 3-dim. Therefore, I think current terminology is already consistent with `matmul` API.

chainer/utils/sparse.py
corochann: > How about renaming it to to_coo to again mimic scipy?

I agree, it's more simple.

> Again here too, could we skip use_variable ?

In that case, do we need additional require_grad option?

chainer/utils/sparse.py
corochann: +1 for

> let the user control the gradient with an additional require_grad setter/getter property.

In that case, do we need additional `require_grad` option?

corochann:
FYI: sparse matmul is merged in experimental branch in Chainer Chemistry now.  
https://github.com/pfnet-research/chainer-chemistry/pull/90

But I think the implementation is refactored and this chainer implementation is more clear.
I would like to consider integrating it to chainer chemistry after this PR is merged.

chainer/functions/math/sparse_matmul.py
hvy: Ah, thanks a lot for commenting on that. Then I think we should definitely keep it as is.

chainer/utils/sparse.py
hvy: That would even make it more consistent to the Variable. I think it's an good idea.

chainer/functions/math/sparse_matmul.py
```diff
@@ -0,0 +1,429 @@
...
+
+try:
+    from scipy import sparse
+    _scipy_available = True
+except ImportError:
+    _scipy_available = False
+
+
+def _sparse_matmul(sp_data, sp_row, sp_col, sp_shape, dn,
+                   transa, transb, transc, dtype=None):
```
corochann: Just a comment:
Should it be `_sparse_matmul_coo` or something by considering that the other type of sparse_matmul (matmul with CSR matrix etc.) will be introduced in the future?

Same applies for all methods/class except `sparse_matmul`. 
I think it is ok to keep `sparse_matmul`, since it is checking the class type to see which function to be used.

chainer/functions/math/sparse_matmul.py
anaruse: OK, I will edit code to conform that style.

chainer/functions/math/sparse_matmul.py
anaruse: Thanks, @corochann , for your comment. I also think `sparse_matmul` is better as a name of this function.

chainer/utils/sparse.py
anaruse: I agree with you. `SparseCooMatrix` is a bit redundant, actually. I will change it to `CooMatrix`.

tests/chainer_tests/utils_tests/test_sparse.py
anaruse: Thank you! That's better.

tests/chainer_tests/functions_tests/math_tests/test_sparse_matmul.py
anaruse: The reason why I used `apply` here is that one of inputs to `sparse_matmul` is not `Variable` and is `CooMatrix`.  The check routines like `check_double_backward` in `gradient_check.py` gives tuple of `Variable` to `func`. So, `sparse_matmul` can not be used as `func` for those check routines, AFAIK.

Is there any better way to work-around this without using `apply`?

anaruse:
Thank you for your feedback, @hvy and @corochann. I've just updated the branch. Please review again.

chainer/functions/math/sparse_matmul.py
hvy: I agree we can be more specific here, although we can probably come back and fix these later considering them implementation details. @anaruse Could you please revisit the names of these functions?

chainer/utils/sparse.py
```diff
@@ -0,0 +1,72 @@
+import chainer
+from chainer.backends import cuda
+
+
+class CooMatrix(object):
```
hvy: Could you add documentations to this class, `to_dense` and `to_coo`?

tests/chainer_tests/utils_tests/test_sparse.py
```diff
@@ -0,0 +1,34 @@
...
+}))
+class TestCooMatrix(unittest.TestCase):
+
+    def test_to_dense(self):
+        if self.nbatch > 0:
+            x_shape = (self.nbatch, self.shape[0], self.shape[1])
+        else:
+            x_shape = self.shape
+        x0 = _setup_tensor(.5, 1, x_shape, self.dtype, .75)
+        sp_x = utils.to_coo(x0)
```
hvy: How about checking some basic properties of `sp_x` such as shapes of its attributes?

chainer/functions/math/sparse_matmul.py
```diff
@@ -0,0 +1,425 @@
...
+            continue;
+        }
+        assert(i_n < _n);
+        C val_C = 0.0;
+        for (int i_k = 0; i_k < _k; i_k++) {
+            int i_A = i_k + _k * (i_m + _m * i_b);
+            int i_B = i_n + _n * (i_k + _k * i_b);
+            A val_A = _A[i_A];
+            B val_B = _B[i_B];
+            val_C += (C)(val_A * val_B);
```
hvy: I don't think it's an issue here but how about preferring `static_cast` in the elementwise kernels for safety?

chainer/functions/math/sparse_matmul.py
```diff
@@ -0,0 +1,425 @@
...
+    def __init__(self, sp_row, sp_col, sp_shape,
+                 transa=False, transb=False, transc=False,
+                 dtype=None):
+        if sp_row.ndim != sp_col.ndim:
+            raise ValueError('ndim of sp_row and sp_col must be the same.')
+        if sp_row.ndim != 1 and sp_row.ndim != 2:
+            raise ValueError('ndim of sp_row and sp_col must be one or two.')
+        for i in range(sp_row.ndim):
+            if sp_row.shape[i] != sp_col.shape[i]:
+                _msg = 'shape of sp_row and sp_col must be the same.'
```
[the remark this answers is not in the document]
anaruse: I got it. will remove the preceding underscore.

tests/chainer_tests/utils_tests/test_sparse.py
anaruse: OK, will add some checking code for `sp_x`

chainer/functions/math/sparse_matmul.py
anaruse: OK, will use `static_cast`.

chainer/utils/sparse.py
anaruse: All right, will add documentations to `CooMatrix`, `to_dense` and `to_coo`.

chainer/functions/math/sparse_matmul.py
anaruse: Thanks for your comments, @corochann and @hvy.
So how about renaming them as follows (while keeping `sparse_matmul`)?
 * _sparse_matmul* --> _coo_matmul*
 * SparseMatMul* --> CooMatMul*

tests/chainer_tests/utils_tests/test_sparse.py
hvy: Can we use Python asserts for the checks, since these are caught and handled by pytest?

hvy:
jenkins, test this please.

tests/chainer_tests/utils_tests/test_sparse.py
anaruse: I got it. will use assert for the checks there.

hvy:
CI seems to fail due to missing `numpy.matmul` for older versions. Could you maybe consider adding a workaround similar to this? https://github.com/chainer/chainer/blob/master/chainer/functions/math/matmul.py#L48

chainer/functions/math/sparse_matmul.py
hvy: Should we be specific about both the LHS and RHS, e,g. `CooDenseMatMul`?

anaruse:
Could you give me a pointer on how to build the docs locally?

chainer/functions/math/sparse_matmul.py
anaruse: AFAIK, matrix multiplication of sparse vs sparse is not so common. So, I don't think we should be specific like that.

hvy:
I'm afraid there is not documentation on it at the moment but you can `cd` in to `docs` and run `make html` . You should install Chainer using `pip install -e .` in the same repository beforehand.

tests/chainer_tests/utils_tests/test_sparse.py
hvy: Looks great, thanks a lot!

tests/chainer_tests/functions_tests/math_tests/test_sparse_matmul.py
```diff
@@ -0,0 +1,295 @@
...
+        if self.nbatch > 0:
+            shape = [self.nbatch, shape[0], shape[1]]
+        return shape
+
+    def _matmul(self, a, b):
+        if self.transa:
+            a = a.swapaxes(-1, -2)
+        if self.transb:
+            b = b.swapaxes(-1, -2)
+        return numpy.matmul(a, b)
```
hvy: Oh, I think we need to check if numpy has matmul here too.

anaruse:
I've just fixed the issues. Could you run tests again?

hvy:
Jenkins, test this please.

hvy:
Thanks, I started the build just now!

chainer/functions/math/sparse_matmul.py
hvy: I see, it might be clear from the arguments too. Could you change the names to your above suggestions?

chainer/utils/sparse.py
```diff
@@ -0,0 +1,102 @@
...
+def to_coo(x, ldnz=None, requires_grad=False):
+    """Returns a sparse matrix in COO format of a dense matrix.
+
+    Args:
+        x (numpy.ndarray or cupy.ndarray): Input dense matrix. The ndim of
+            ``x`` must be two or three. If ndim is two, it is treated as
+            a single matrix. If three, it is treated as batched matrices.
+        ldnz (int): Size of arrays for data, row index and column index to be
+            created. The Actual size becomes max(nnz, ldnz) where nnz is number
+            of non-zero elmeents in a input dense matrix.
```
hvy: Oh, a small typo at `elemeents`.

hvy: Just a small question, is "ldnz" common terminology, would you mind explaining what is stands for?

chainer/functions/math/sparse_matmul.py
```diff
@@ -0,0 +1,432 @@
...
+        '''
+        int i_n = (i % _n);
+        int i_A = (i / _n);
+        int i_b = (i / _n) / nnz;
+        if (i_b >= nb) {
+            continue;
+        }
+        int i_k = A_col[i_A];
+        if (i_k < 0) {
+            continue;
```
hvy: When do we encounter these negative indices?

hvy:
jenkins, test this please.

chainer/utils/sparse.py
anaruse: Thanks, will fix the typo.

Re: "ldnz", it stands for "Leading Dimension of array for Non-Zero elements" and shows the size of array which holds matrix entries, row indexes or column indexes. I follow naming convention of matrix libraries like blas, lapck, etc. in which, for example, scalar variable `ldA` indicates the size of array `A`.

chainer/functions/math/sparse_matmul.py
anaruse: You may see the negative indexes when it is batched sparse matrices and number of non-zero elements in each sparse matrix differ.

hvy:
jenkins, test this please.

hvy:
@anaruse Not sure if this one https://github.com/chainer/chainer/pull/4397/files/5eeff1763f274deefcccbe43966d4ce2c48e7f52#r187516691 got away unnoticed. Would you mind changing the names as you previously suggested?

chainer/functions/math/sparse_matmul.py
anaruse: Oh, I'm sorry, I missed your comment above. So, I will change the names in that way.

chainer/utils/sparse.py
hvy: Thanks you for your explanation! I see it is used for indexing in general. It seems though that it is not tested. Should we add tests since it's a public interface?

anaruse:
I've just renamed the names of some functions and classes as we discussed. Sorry to bother you again, but could you check again?

chainer/utils/sparse.py
anaruse: Thank you so much! I found a bug in `to_dense` of `CooMatrix` thanks to the tests for `ldnz`.

chainer/functions/math/sparse_matmul.py
```diff
@@ -0,0 +1,429 @@
...
+        2. C (dense) = A (dense) * B (sparse)
+
+    Args:
+        a (Variable or CooMatrix): The left operand of mat-mul.
+        b (Variable or CooMatrix): The right operand of mat-mul.
+        transa (bool): If ``True``, each matrix in ``a`` will be transposed.
+        transb (bool): If ``True``, each matrix in ``b`` will be transposed.
+
+    Returns:
+        _chainer.Variable: Result of batched mat-mul.
```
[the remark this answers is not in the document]
anaruse: Strictly speaking, I think yes.  So, should `CooMatrix` be `~chainer.utils.CooMatrix` as well?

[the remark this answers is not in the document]
hvy: They actually should (links were broken). Thanks for that catch, could you fix it?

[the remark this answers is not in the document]
anaruse: Yes, I will.

anaruse:
I'm sorry to bother you over and over, but could you check again?

hvy:
No worries, I am rather sorry for the exact opposite. I'll rerun the CI to check your fixes!

hvy:
jenkins, test this please.

hvy:
jenkins, test this please.

hvy:
The CI passes and it looks good except a minor detail that the `CooMatrix` link for the argument seems broken. Maybe you have to write using `(:class: ... or ... :class: ...)` ?

anaruse:
I dig into the link issue above, and seems we need to add/edit .rst file for `CooMatrix`, etc. to solve the issue. I will push the update after some local verification.

anaruse:
Could you check again?  Probably, the update above will fix the link issue.

hvy:
Ah I see, I'll try rebuilding the docs.

hvy:
jenkins, test this please.

PR #4402, opened 2018-02-23
add Codecov.io configuration

kmaehashi:
This PR introduces Codecov.io to measure test coverage for diff introduced in each Pull-Request (rather than overall coverage).
The coverage will be uploaded via Jenkins (chainer-test).

Codecov.io covers most of features in Coveralls, but it does not visualize branch coverage.
So I continue to keep Coveralls configuration as is.

The meaning of the configuration is:

* Disable commenting on PRs. Just leave commit status on PR (like Travis).
* Do not measure overall coverage (it is currently measured by Coveralls).
* Measure coverage for diffs introduced in PRs.
  For now, target (threshold to decide `✔︎` or `X`) is set to `0%` to avoid unnecessary confusion of contributors.
  We can reconsider this rate afterwards, if needed.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,11 @@
+comment: false
+
+coverage:
+  status:
+    # Disable coverage measurement for overall codebase.
+    project: off
+    # Enable coverage measurement for diff introduced in the pull-request,
+    # but do not mark "X" on commit status for now.
+    patch:
+      default:
+        target: '0%'


niboshi:
Is it still not possible to see the coverage report from this PR?

kmaehashi:
As coverage will be sent from Jenkins, https://github.com/chainer/chainer-test/pull/399 PR needs to be merged to see the report.
It is better to merge this PR first, as merging https://github.com/chainer/chainer-test/pull/399 will trigger Codecov with default configuration (e.g., commenting on PRs, etc.)

niboshi:
LGTM!

PR #4406, opened 2018-02-26
use CuPy wheels in Dockerfiles

kmaehashi:


docker/python3/Dockerfile
```diff
@@ -4,8 +4,8 @@ RUN apt-get update -y && \
     apt-get install -y --no-install-recommends \
     python3-dev \
     python3-pip \
-    python3-setuptools \
-    python3-wheel && \
+    python3-wheel \
+    python3-setuptools && \
```
kmaehashi: This change is just an decorative fix to make it consistent with python2 Dockerfile.

mitmul:
LGTM. I confirmed I can use CuPy and Chainer in the container created based on the new Docker images.

mitmul:
All tests passed.

PR #4408, opened 2018-02-26
Fix Variable tests

mitmul:
Variable tests failed due to warnings caused by Variables with scalar arrays.


kmaehashi:
LGTM!

kmaehashi:
Note: this issue was introduced in https://github.com/chainer/chainer/pull/4199

PR #4413, opened 2018-02-26
Add test to ensure no mutable default arguments

niboshi:
Related to: #4359

Added test to ensure mutable default arguments are not used.

(After this PR, I'm considering merging `test_init_docstring.py` into this file)


kmaehashi:
:+1: We have to fix these first.

```
E           AssertionError: Some functions have mutable values as default arguments:
E             
E             chainer.computational_graph.build_computational_graph: arg='variable_style' type=<type 'dict'>
E             chainer.computational_graph.build_computational_graph: arg='function_style' type=<type 'dict'>
E             chainer.computational_graph.__init__: arg='variable_style' type=<type 'dict'>
E             chainer.computational_graph.__init__: arg='function_style' type=<type 'dict'>
E             chainer.links.model.vision.vgg.__call__: arg='layers' type=<type 'list'>
E             chainer.links.model.vision.vgg.extract: arg='layers' type=<type 'list'>
E             chainer.links.model.vision.googlenet.__call__: arg='layers' type=<type 'list'>
E             chainer.links.model.vision.googlenet.extract: arg='layers' type=<type 'list'>
E             chainer.links.model.vision.googlenet.__call__: arg='layers' type=<type 'list'>
E             chainer.links.model.vision.googlenet.extract: arg='layers' type=<type 'list'>
E             chainer.links.model.vision.resnet.__call__: arg='layers' type=<type 'list'>
E             chainer.links.model.vision.resnet.extract: arg='layers' type=<type 'list'>
E             chainer.links.model.vision.vgg.__call__: arg='layers' type=<type 'list'>
E             chainer.links.model.vision.vgg.extract: arg='layers' type=<type 'list'>
E             chainer.training.extensions.parameter_statistics.__init__: arg='statistics' type=<type 'dict'>
E             chainer.training.extensions.parameter_statistics.__init__: arg='statistics' type=<type 'dict'>
```

vilyaair:
It is possibly my fault. At that time, I have observed these mutable default arguments. However, if I modify these arguments, I have to change corresponding documents. Consequently, the default values of arguments in the codes are not consistent to these in the documents, e.g.
https://github.com/chainer/chainer/blob/c4fe0bba3b68184fcd59bac95bdd7d4054aac572/chainer/links/model/vision/googlenet.py#L186-L207
changes to
```
    def __call__(self, x, layers=None, **kwargs):
        """__call__(self, x, layers=None)

        Computes all the feature maps specified by ``layers``.

        .. warning::

           ``train`` argument is not supported anymore since v2.
           Instead, use ``chainer.using_config('train', train)``.
           See :func:`chainer.using_config`.

        Args:
            x (~chainer.Variable): Input variable. It should be prepared by
                ``prepare`` function.
            layers (list of str): The list of layer names you want to extract.
                The default is ``['prob']``.

        Returns:
            Dictionary of ~chainer.Variable: A directory in which
            the key contains the layer name and the value contains
            the corresponding feature map variable.

        """

        if layers is None:
            layers = ['prob']
```

I considered that such changes caused the default arguments not well manifested in the document. Therefore, I then decided to not touch them.
If you think that above codes should also be modified, would you mind my rectifying them?

niboshi:
I don't think documentation has to be modified.

```
"""__call__(self, x, layers=['prob']) 
```

This line can be kept so that the effective default value is left visible in the docs.

vilyaair:
I have made this patch, and hope that you are willing to accept it.
[Avoid_mutable_default_arguments.zip](https://github.com/chainer/chainer/files/1765343/Avoid_mutable_default_arguments.zip)

niboshi:
It seems good. Could you make a PR?
Thanks!

vilyaair:
This test would not examine another mutable default argument issue, e.g.
```
l = [0]
def f(a, L=None):
    if L is None:
        L = l
    L.append(a)
    return L

print(f(1))
print(f(2))
print(f(3))
```
This will print
```
[0, 1]
[0, 1, 2]
[0, 1, 2, 3]
```

niboshi:
The problem seems not of the default argument, but of the logic.
It explicitly modifies the global object.
I think we cannot do anything automatically for such code.

kmaehashi:
Blocked by #4822

kmaehashi:
As #4822, which blocks this PR, introduces API changes, this PR cannot be backporerd.

toslunar:
#4822 is merged.

kmaehashi:
pfnCI, test this please.

kmaehashi:
Jenkins failures are not related to this PR. Retriggered Travis and waiting for results.

kmaehashi:
LGTM!

PR #4414, opened 2018-02-27
Prefer chainer.backends.cuda in recent added codes

vilyaair:
Some recent added codes still use deprecated `chainer.cuda`, and accordingly, I modify them to conform to latest recommendation `chainer.backends.cuda`.


kmaehashi:
Good catch! Jenkins, test this please.

kmaehashi:
LGTM!

PR #4416, opened 2018-02-27
add backward test for old-styled sclar variable

kmaehashi:
Add tests for features deprecated in #4199.


niboshi:
How about using pytest-style asserts?
(`assert`, `pytest.raises`)

niboshi:
Ah, I see. It's aligned with other asserts in the same file.

niboshi:
LGTM!

PR #4419, opened 2018-02-28
Avoid mutable default arguments

vilyaair:
Related to: #4413


kmaehashi:
Jenkins, test this please.

hvy:
Sorry, the `ParameterStatistics` was just updated, could you resolve the conflict? 

Please note that this changes its behavior when passing `None` to its constructor. I think that is fine but we should rewrite it's test (e.g. pass `{}` instead of `None`) and maybe label it as no compat?

vilyaair:
I suggest to use [`MappingProxyType`](https://docs.python.org/3/library/types.html#types.MappingProxyType), so the following codes 
https://github.com/chainer/chainer/blob/5d7033e170c7ebbb85a74f416a4d78d120570133/chainer/training/extensions/parameter_statistics.py#L65
could be rewrote as `def __init__(self, links, statistics=MappingProxyType(default_statistics), `.
Do you think that this method is acceptable?

kmaehashi:
It seems `MappingProxyType` was added in Python 3.3? We still need to support Python 2.7.

vilyaair:
Actually, there is another a little more complicated and wierd solution: `tuple(six.viewitems(default_statistics))`. Though such change will not affect code logic, this is not dict at all.

hvy:
Thanks for the additional solutions. I personally think that it is okay to change the behavior (i.e. `None` by default that is then mapped to the default dict) since it doesn't seem like this extension is that widely used with non-default arguments.

chainer/computational_graph.py
```diff
@@ -85,12 +85,16 @@ class ComputationalGraph(object):
 
     """
 
-    def __init__(self, nodes, edges, variable_style=_var_style,
-                 function_style=_func_style, rankdir='TB',
+    def __init__(self, nodes, edges, variable_style=None,
```
kmaehashi: I think we need to add documentations about default values of these two parameters.
Otherwise users cannot tell what can be specified (currently it can be guessed from the default value in the documentation).
https://docs.chainer.org/en/stable/reference/generated/chainer.computational_graph.ComputationalGraph.html

In addition, this breaks backward compatibility of this behavior.
https://github.com/chainer/chainer/pull/4419/files#diff-54541860b0345091e1a50777c35c730bR77

chainer/computational_graph.py
```diff
@@ -191,8 +195,8 @@ def _skip_variable(nodes, edges):
 
 
 def build_computational_graph(
-        outputs, remove_split=True, variable_style=_var_style,
-        function_style=_func_style, rankdir='TB', remove_variable=False,
+        outputs, remove_split=True, variable_style=None,
```
kmaehashi: ditto

vilyaair:
@kmaehashi 
I really appreciate your close examination.

I feel that it is probably more sensible to use MappingProxyType for Python3 and do not compel to avoid mutable default arguments for Python2.
Moreover, I find that [`chainer.training.extensions.dump_graph`](https://docs.chainer.org/en/stable/reference/generated/chainer.training.extensions.dump_graph.html) also has similar problem.

chainer/computational_graph.py
```diff
@@ -85,13 +85,13 @@ class ComputationalGraph(object):
 
     """
 
-    def __init__(self, nodes, edges, variable_style=_var_style,
-                 function_style=_func_style, rankdir='TB',
-                 remove_variable=False, show_name=True):
+    def __init__(self, nodes, edges, rankdir='TB', remove_variable=False,
```
kmaehashi: This change is unacceptable in the sense that it also breaks compatibility by changing positional argument into keyword-only argument.

> In addition, this breaks backward compatibility of this behavior.
> https://github.com/chainer/chainer/pull/4419/files#diff-54541860b0345091e1a50777c35c730bR77

I think it is acceptable to change this behavior, i.e., use `variable_style=None` and use `_var_style` if `variable_style is None`, because
* This feature is not widely-used with non-optional arguments (as hvy noted), and
* I guess no one requires Chainer v1.23.0-style output any more.

Could you please revert your additional commits and update the documentation?

vilyaair: How about the following modification?
```
    def __init__(self, nodes, edges, variable_style=_var_style,
                 function_style=_func_style, rankdir='TB',
                 remove_variable=False, show_name=True):
```
->
```
    def __init__(self, nodes, edges, *args, **kwargs):
        self.nodes = nodes
        self.edges = edges
        args_list = argument.parse_kwargs(kwargs,
                                          ('variable_style', _var_style),
                                          ('function_style', _func_style),
                                          ('rankdir', 'TB'),
                                          ('remove_variable', False),
                                          ('show_name', True),
                                          )
        (self.variable_style,
         self.function_style,
         self.rankdir,
         self.remove_variable,
         self.show_name,
         ) = args + args_list[len(args):]

```

vilyaair: Additionally, it also causes unexpected result to use `variable_style=None` and to use `_var_style` if `variable_style is None`, the purpose of which is just to pass the tests.
The following codes illustrate this phenomenon:
```
l = [0]
def f(a, L=None):
    if L is None:
        L = l
    L.append(a)
    return L

print(f(1))
print(f(2))
print(f(3))
```
This will print
```
[0, 1]
[0, 1, 2]
[0, 1, 2, 3]
```

kmaehashi: As far as I understand, the motivation of prohibiting use of mutable default arguments is to avoid future bug in case implementers / reviewers overlook the fact that default values of arguments will not be initialized every time the function gets called.

> How about the following modification?

According to your idea, functions using `parse_args` cannot accept mutable arguments from users as it calls `deepcopy`. This behavior cannot be predicted from its name `parse_args`, and can be a hot bed of bugs which is worse than the "mutable default argument" issue.

> Additionally, it also causes unexpected result to use variable_style=None and to use _var_style if variable_style is None, the purpose of which is just to pass the tests.

I think you are talking about different problem (i.e., modification of global variable affects the following execution), and I believe such mistakes can be found during the review.

vilyaair: What if `parse_args` function returns plain `args + values` rather than `deepcopy`?

kmaehashi: I'm still negative about introducing `parse_args` because of two reasons:

* We should separately maintain the method signature for Sphinx documentation of such methods. This kind of error is more likely to happen (e.g., #4539) than the error like modifying mutable default arguments.
* Unlike `parse_kwargs`, `parse_args` cannot be eliminated forever. (`parse_kwargs` can be replaced with keyword-only argument syntax once we stop supporting Python 2.7 (although not planned yet).)

vilyaair: I am with you completely.  Admittedly, `parse_args` function is a bad way to avoid mutable default arguments.

kmaehashi:
Thanks for the update!
Jenkins, test this please.

kmaehashi:
LGTM!

PR #4432, opened 2018-03-06
chmod 644 *.py

toslunar:


niboshi:
Maybe we should remove those shebangs?

niboshi:
Or maybe we should keep executable bits for example scripts.

kmaehashi:
+1 to keep it as is for examples, unless there is a strong motivation to do so.
At least one user invoking the script directly. http://pongsuke.hatenadiary.jp/entry/2017/03/14/132907

toslunar:
I reverted the changes in examples.

niboshi:
LGTM!

PR #4434, opened 2018-03-07
Typecheck split_at

corochann:
When we use `split_dataset` or `split_dataset_random`, it won't raise error when we set `float` value to the second argument of `split_at`.

It is likely happen when we want to split dataset according to the ratio of train/validation dataset.
e.g.,
```
train_ratio = 0.8
# here actually we must change type to int, but did not notice, because no error is raised!
train, val = split_dataset_random(dataset, len(dataset) * ratio)
```

This PR is to check type of `split_at` to be `int`. After this PR, above code will raise `TypeError`.

### DIFF
--- a/chainer/datasets/sub_dataset.py
+++ b/chainer/datasets/sub_dataset.py
@@ -96,6 +96,9 @@ def split_dataset(dataset, split_at, order=None):
 
     """
     n_examples = len(dataset)
+    if not isinstance(split_at, int):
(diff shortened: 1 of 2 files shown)


niboshi:
Please use `six.integer_types` and `numpy.integer` instead of `int`.

corochann:
thanks for comment, updated.

niboshi:
jenkins, test this please

niboshi:
LGTM!

PR #4435, opened 2018-03-07
Remove future.types.newint

niboshi:
Now that we use `isinstance` to check if a variable is integer, `newint` is no longer needed.

Related: #1629, #4380


bkvogel:
LGTM!

PR #4436, opened 2018-03-07
reset Link._device_id to None in to_intel64

kmaehashi:
`Link._device_id` does not change when transferred from GPU to intel64.

### DIFF
--- a/tests/chainer_tests/test_link.py
+++ b/tests/chainer_tests/test_link.py
@@ -1142,6 +1142,7 @@ class TestIntel64(unittest.TestCase):
     def test_cpu_to_intel64(self):
         link = self.link
         link.to_intel64()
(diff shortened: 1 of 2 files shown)


tests/chainer_tests/test_link.py
```diff
@@ -1142,6 +1142,7 @@ def _assert_arrays_equal(self, array, expected_array):
     def test_cpu_to_intel64(self):
         link = self.link
         link.to_intel64()
+        self.assertEqual(link._device_id, None)
```
niboshi: Use `assert`, and also `is` instead of `==` to compare with `None`.
(Ditto for others)

kmaehashi: Thanks, fixed.

niboshi:
LGTM!

PR #4437, opened 2018-03-07
fix Evaluator.device docs

kmaehashi:
`training` -> `validation`

### DIFF
--- a/chainer/training/extensions/evaluator.py
+++ b/chainer/training/extensions/evaluator.py
@@ -53,7 +53,7 @@ class Evaluator(extension.Extension):
             just a link object, the link is registered by the name ``'main'``.
         converter: Converter function to build input arrays.
             :func:`~chainer.dataset.concat_examples` is used by default.
-        device: Device to which the training data is sent. Negative value
+        device: Device to which the validation data is sent. Negative value
             indicates the host memory (CPU).
         eval_hook: Function to prepare for each evaluation process. It is
             called at the beginning of the evaluation. The evaluator extension
@@ -63,7 +63,7 @@ class Evaluator(extension.Extension):
 
     Attributes:
         converter: Converter function.
-        device: Device to which the training data is sent.
+        device: Device to which the validation data is sent.
         eval_hook: Function to prepare for each evaluation process.
         eval_func: Evaluation function called at each iteration.
 


niboshi:
Hmm, is this feature intended only for validation sets?
Maybe we should simply state a "dataset", like this?

"Evaluator: Trainer extension to evaluate models on a dataset."

kmaehashi:
I agree that Evaluator may also be used for test datasets by directly invoking it (instead of `trainer.extend`). However, if we expect such use-case, I think it is better to define a separate utility class rather than providing it as trainer extensions (like `IteratorProgressBar` proposed in https://github.com/chainer/chainer/pull/3013). I think having such utility helps users writing evaluation code. How about doing this discussion in separate issue?

niboshi:
OK, let's keep this topic open for discussion.
For this PR,  I'll merge with current change.

kmaehashi:
OK.

----

I found the above-mentioned use case (directly invoking Evaluator) in example. https://github.com/chainer/chainer/blob/v4.0.0b4/examples/ptb/train_ptb.py#L254-L259

Besides that, it is also possible to use Evaluator to run evaluation against arbitrary iterators (e.g., training dataset) as you pointed out.

niboshi:
Besides that, we are using the name `'test_iter'` as the argument to `Evaluator`, where seemingly it's intended to be used as validation sets. This is not great for education.

https://docs.chainer.org/en/stable/tutorial/trainer.html#add-extensions-to-the-trainer-object
https://docs.chainer.org/en/stable/tutorial/gpu.html?run-neural-networks-on-a-single-gpu#run-neural-networks-on-a-single-gpu

kmaehashi:
I'll fix the tutorial in separate PR.

kmaehashi:
Hm, it seems `test_iter` stands for "iterator for test dataset", and test dataset is used for validation.

```
test_iter = iterators.SerialIterator(test, batchsize, False, False)
```

I think it is more natural to split validation dataset from training dataset, but I also think it is acceptable to just use test dataset in this kind of examples. But maybe it is better to add some comments to these code.

niboshi:
You're referring to the trainer tutorial? https://docs.chainer.org/en/stable/tutorial/trainer.html

> but I also think it is acceptable to just use test dataset in this kind of examples.

No, we should never confuse test and validation set, even with comments.
Especially in tutorial, because this is maybe one of the most common beginner's pitfalls of machine learning.
We should at least remove/rewrite "validation" labels in plots.

niboshi:
LGTM!

kmaehashi:
I found a similar discussion in Keras issue.
https://github.com/keras-team/keras/issues/1753

PR #4448, opened 2018-03-08
prepare API to get/set RandomState instances unique to each batch index

grafi-tt:
Part of #4230.

`iterators.get_random_state` (to be called by user) and `iterators.random_state.set_random_state` (to be called by iterator implementation) are implemented, and `get_random_state` is documented.

Currently iterators set dummy state by `set_random_state` which raises `NotImplementedError` immediately when `get_random_state` is called.

### DIFF
--- a/chainer/iterators/multiprocess_iterator.py
+++ b/chainer/iterators/multiprocess_iterator.py
@@ -14,7 +14,6 @@ from chainer.dataset import iterator
 
 
 _response_time = 1.
-_short_time = 0.001
 _PrefetchState = namedtuple('_PrefetchState', (
     'current_position', 'epoch', 'is_new_epoch',
     'previous_epoch_detail', 'order'))
--- a/chainer/iterators/multiprocess_iterator.py
+++ b/chainer/iterators/multiprocess_iterator.py
@@ -278,11 +278,7 @@ class _PrefetchLoop(object):
 
         # Use a distinct RandomState in the thread
         # for deterministic random number generation.
-        # To support 32-bit platform and numpy < 1.11,
-        # the seed is taken in a verbose manner.
-        seed = numpy.asscalar(
-            numpy.random.randint(-(1 << 31), 1 << 31, 1).astype('uint32'))
-        self._random = numpy.random.RandomState(seed)
+        self._random = random_state.derive_random_state()
 
         self._interruption_testing = _interruption_testing
 
--- a/chainer/iterators/multiprocess_iterator.py
+++ b/chainer/iterators/multiprocess_iterator.py
@@ -11,6 +11,7 @@ import numpy
 import six
 
 from chainer.dataset import iterator
+from chainer.iterators import random_state
 
 
 _response_time = 1.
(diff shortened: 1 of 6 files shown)


delta2323:
Sorry for not my responding so long. Thank you for making a smaller PR. I'll take a look.

grafi-tt:
@delta2323 Many thanks, and congratulation for 4.0.0 RC release!

chainer/iterators/multiprocess_iterator.py
```diff
@@ -94,7 +94,10 @@ def __next__(self):
             self._previous_epoch_detail, self._order) = prefetch_state
         if batch is None:
             raise StopIteration
-        else:
+
+        def dummy_state():
+            raise NotImplementedError
+        with random_state.set_random_state(dummy_state):
```
delta2323: It seems to me this context manager doesn't take any effect on the return `batch`. If so, are you thinking to modify the code in the PRs following this one?

chainer/iterators/random_state.py
```diff
@@ -0,0 +1,89 @@
...
+
+import numpy
+
+
+_random_state = threading.local()
+_random_state.value = None
+
+
+def get_random_state():
+    """Get a :class:`~numpy.random.RandomState` during iteration.
```
delta2323: I think "during iteration" is not needed as users can call this function any time, although the most expected use case is to call it during iterations.

chainer/iterators/random_state.py
```diff
@@ -0,0 +1,89 @@
...
+
+    The instance obtained is unique to each index on a batch.  You can
```
delta2323: From the same reason as the previous one, I would suggest to describe it as one of expected use cases.

chainer/iterators/random_state.py
delta2323: It would be better to note that returned instance is a thread-local variable.

chainer/iterators/random_state.py
```diff
@@ -0,0 +1,89 @@
...
+    during prefetch are silently discarded.
+
+    Returns:
+        The :class:`numpy.random.RandomState` instance bound to the batch
+        index, if a :class:`~chainer.dataset.Iterator` instance is fetching
+        data; otherwise ``None``.
+
+    """
+    state = _random_state.value
+    if not isinstance(state, numpy.random.RandomState):
```
delta2323: What do you assume as a type of `state` other than `RandomState`?

chainer/iterators/random_state.py
```diff
@@ -0,0 +1,89 @@
...
+def create_random_states(num):
+    """Derive new PRNG states from the global one.
+
+    Args:
+        num (int): The number of the states to be created.
+
+    Returns:
+        A list of :class:`numpy.random.RandomState` instances.
+    """
+    return [derive_random_state() for _ in range(num)]
```
delta2323: I think `six.moves.range` is better because `num` can be arbitrarily large.

delta2323:
I added several comments. Could you check that?

delta2323:
@grafi-tt Hi, can you spare time to work on this PR? But I heard from other members that we may not need this PR any more. Do you have some opinion?

grafi-tt:
@delta2323 Sorry, I don't have enough time to work on the PRs. I'm fine with closing them. 

It seems `MultiprocessIterator` isn't of high priority at all, because `MultithreadIterator` works well on typical workloads, as numpy and recent version of Pillow faithfully release GIL.

Just out of curiously, what other members say? (If you can't reveal, it's OK.)

PR #4450, opened 2018-03-08
grouped convolution for intel64 backed environment

tkng:
The aim of this PR is to enable use of grouped convolution when using intel64 backend.

Since basic implementation for grouped convolution was done before intel64 backend was merged, current convolution implementation just fallbacks to naive im2col implementation if `groups` option was specified.

In this PR, I improved to call ideep4py convolution function even when `groups` keyword is specified.

Simple benchmark result shows drastical performance improvement, ideep4py is roughly 20 times faster than naive implementation. In addition, memory consumption is also improved.

```
## benchmarker:         release 4.0.1 (for python)
## python version:      3.6.3
## python compiler:     GCC 7.2.0
## python platform:     Linux-4.13.0-32-generic-x86_64-with-Ubuntu-17.10-artful
## python executable:   /usr/bin/python3
## cpu model:           Intel(R) Core(TM) i5-7260U CPU @ 2.20GHz  # 2200.000 MHz
## parameters:          loop=10, cycle=1, extra=0

##                        real    (total    = user    + sys)
default                20.2652   20.4700   11.8500    8.6200
ideep                   0.9327    1.8500    1.7000    0.1500

## Ranking                real
ideep                   0.9327  (100.0) ********************
default                20.2652  (  4.6) *

## Matrix                 real    [01]    [02]
[01] ideep              0.9327   100.0  2172.8
[02] default           20.2652     4.6   100.0
```

Here is the benchmark script.
```
import numpy
import chainer
import chainer.functions as F
from benchmarker import Benchmarker

x = numpy.random.uniform(-1, 1, (10, 64, 221, 221)).astype(numpy.float32)
w = numpy.random.uniform(-1, 1, (64, 8, 3, 3)).astype(numpy.float32)
x = chainer.Variable(x)

with Benchmarker(10, width=20) as bench:
    @bench('default')
    def default(bm):
        for i in bm:
            with chainer.using_config('use_ideep', 'never'):
                y = F.convolution_2d(x, w, None, 1, 1, groups=8)
    @bench('ideep')
    def insertion(bm):
        for i in bm:
            with chainer.using_config('use_ideep', 'always'):
                y = F.convolution_2d(x, w, None, 1, 1, groups=8)
```


chainer/backends/intel64.py
```diff
@@ -103,12 +103,15 @@ def inputs_all_ready(inputs, supported_ndim=(2, 4)):
         bool: ``True`` if all conditions meet.
 
     """
+
+    def check_array_type(a):
```
niboshi: Currently this function is used only from this module.
Could you make it private by prefixing with underscore (`_`)?

Also, `check_*` is not appropriate for a function that returns a logical value.
How about something like `_is_array_types_valid`, `_is_array_types_supported`, etc.?

chainer/functions/connection/convolution_2d.py
```diff
@@ -235,22 +230,36 @@ def _forward_grouped_convolution(self, x, W, b):
 
         xp = cuda.get_array_module(x)
 
-        _x = x.reshape(N, G, iCg, iH, iW)
+        _x = x.reshape((N, G, iCg, iH, iW))
```
niboshi: Are these changes on `reshape` arguments essential?
If not, could you undo them?

chainer/functions/connection/convolution_2d.py
```diff
@@ -235,22 +230,36 @@ def _forward_grouped_convolution(self, x, W, b):
...
 
-        y = xp.concatenate(_ys, axis=1)  # (N, oC, oH, oW)
+            if self._use_ideep:
+                _ys.push_back(_y)
+            else:
+                _ys.append(_y)
+
+        # (N, oC, oH, oW)
+        if self._use_ideep:
+            y = intel64.ideep._ideep4py.concat.Forward(_ys, 1)
```
niboshi: Instead of branching every time an element is pushed to `_ys`, how about constructing `_ys` as a list and converting to a `mdarrayVector` here?

chainer/functions/connection/convolution_2d.py
tkng: The former form`x.reshape(N, G, iCg, iH, iW)` is not allowed when the type of `x` is `mdarray`. Therefore we need this change here for now. Maybe we can improve the implementation of ideep4py, though.

niboshi: I see. Thank you for clarification!

tkng:
Thanks for your advice, I completed improvements. Please confirm.

Luckily, the result of unit tests were all green at this time.

tkng:
How about the status of this PR? I'm thinking that the code itself is ready to merge.

niboshi:
I'm sorry for the delay.
The code looks good. Could you please resolve the conflict?
I'll merge after that.

kmaehashi:
@tkng Ping?

niboshi:
I resolved the conflict, as it seems trivial.

@kmaehashi Could you check it?

kmaehashi:
LGTM!

kmaehashi:
Jenkins, test this please.

kmaehashi:
LGTM!

kmaehashi:
@niboshi Do you think this is to be backported? (I think it's ok to backport this)

niboshi:
@kmaehashi 
Hmm, I don't have a strong opinion about that, but I think it's beneficial for v4 users to broaden iDeep support in v4.

kmaehashi:
OK, then I'll backport this PR.

PR #4456, opened 2018-03-12
Expose F.theano_function

kmaehashi:


hvy:
How about adding documentations to this function?

kmaehashi:
I found it does not make much sense as generally users should use this function via `L.TheanoFunction`

PR #4457, opened 2018-03-12
Fix to disable using iDeep in optimizers if to_intel64 is not called

kmaehashi:
Currently iDeep cannot be disabled when `MomentumSGD` is used.
This makes performance worse due to the copy overhead.

### DIFF
--- a/chainer/optimizers/momentum_sgd.py
+++ b/chainer/optimizers/momentum_sgd.py
@@ -37,10 +37,10 @@ class MomentumSGDRule(optimizer.UpdateRule):
             self.state['v'] = xp.zeros_like(param.data)
 
         # For iDeep
-        if isinstance(param.data, intel64.mdarray):
-            if intel64.inputs_all_ready((self.state['v'],)):
-                self.state['v'] = intel64.ideep.array(
-                    self.state['v'], itype=intel64.ideep.wgt_array)
+        if (isinstance(param.data, intel64.mdarray) and
+                intel64.inputs_all_ready((self.state['v'],))):
+            self.state['v'] = intel64.ideep.array(
+                self.state['v'], itype=intel64.ideep.wgt_array)
 
     def update_core_cpu(self, param):
         grad = param.grad
--- a/chainer/optimizers/momentum_sgd.py
+++ b/chainer/optimizers/momentum_sgd.py
@@ -37,9 +37,10 @@ class MomentumSGDRule(optimizer.UpdateRule):
             self.state['v'] = xp.zeros_like(param.data)
 
         # For iDeep
-        if intel64.inputs_all_ready((self.state['v'],)):
-            self.state['v'] = intel64.ideep.array(
-                self.state['v'], itype=intel64.ideep.wgt_array)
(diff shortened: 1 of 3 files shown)


chainer/optimizers/momentum_sgd.py
```diff
@@ -37,9 +37,10 @@ def init_state(self, param):
             self.state['v'] = xp.zeros_like(param.data)
 
         # For iDeep
-        if intel64.inputs_all_ready((self.state['v'],)):
-            self.state['v'] = intel64.ideep.array(
-                self.state['v'], itype=intel64.ideep.wgt_array)
+        if isinstance(param.data, intel64.mdarray):
+            if intel64.inputs_all_ready((self.state['v'],)):
```
niboshi: How about using `and` operator?

chainer/variable.py
```diff
@@ -1273,18 +1274,21 @@ def __reduce__(self):
...
         if self.data is None:
             if device is None:
                 device = cuda.Device().id
+            self._initial_backend = 'cuda'
             self._initial_device = device
 
     def to_intel64(self):
         super(Parameter, self).to_intel64()
         if self.data is None:
+            self._initial_backend = 'intel64'
```
niboshi: I think strings are in general not optimal for this kind of use, unless it's interned.
Maybe it does not matter because `to_xxx` methods are inherently not meant to be fast.

Do you have any thoughts?

chainer/variable.py
```diff
@@ -1320,6 +1324,10 @@ def initialize(self, shape):
         self.data = data
         self.grad = grad
 
+        # Convert the array for iDeep.
```
niboshi: https://github.com/chainer/chainer/pull/4457/files#diff-e131a8dfc2ea641e7e7c63597b9b0d8eR1316
Don't we have to check `_initial_backend` here?
I understand it works without checking, but `_initial_device` is now subsidiary concept of `_initlal_backend`, so I think whenever we use `_initial_device`, we should check `_initial_backend` first.

chainer/optimizers/momentum_sgd.py
kmaehashi: Fixed.

chainer/variable.py
kmaehashi: I interned these string constants.
When more backends are added, maybe it's better to define them in each backend like `chainer.backneds.cuda.ID`.

chainer/variable.py
kmaehashi: Fixed to use `_initial_backend`.

chainer/variable.py
niboshi: What I meant was to use `sys.intern` and `==` for comparison (maybe overkill).
In current implementation, there's no reason to use strings.
How about either using unique objects (i.e. `object()`) or going back to simple string comparison with `==`?

kmaehashi: Ah, I see. I reverted back to the previous code anyway.

niboshi:
LGTM!

PR #4458, opened 2018-03-12
add benchmark framework for Chainer

kmaehashi:
I implemented benchmark framework for Chainer using [ASV](https://github.com/airspeed-velocity/asv).

Currently the following benchmarks are implemented:

* convnet-benchmark
* Overall performance of MNIST example
* Functions: forward / backward of `F.{convolution_2d, sqrt, rsqrt}`

I'd like to improve benchmark coverage later.

### DIFF
--- a/benchmarks/.gitignore
+++ b/benchmarks/.gitignore
@@ -1,0 +1,5 @@
+html/
+results/
+env/
+chainer/
+cupy/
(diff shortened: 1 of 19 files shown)


kmaehashi:
I discussed with @okuta and decided to prepare a different repos for benchmark code.

PR #4461, opened 2018-03-13
Save snapshot in the OS default permission

belltailjp:
Currently Chainer saves every snapshots in `rw-------` permission, since it uses tempfile inside.
I think this behavior is a bit different from what users usually expect, because for example `chainer.serializers.save_npz` writes output in the OS default permission, which is typically `rw-r--r--`. File creation through Python (or OS/shell) will also follow that rule by default.

This PR creates temporary directory and write a target file into the tmp dir then move to target directory, instead of using tempfile. By this approach, snapshot will automatically have OS default permission,

#### ToDo
* ~Tests~ Done
* ~Check compatibility with older Python, other OS, ...~ (CI will do?)

### DIFF
--- a/chainer/training/extensions/_snapshot.py
+++ b/chainer/training/extensions/_snapshot.py
@@ -87,7 +87,5 @@ def _snapshot_object(trainer, target, filename, savefun):
     try:
         savefun(tmppath, target)
         shutil.move(tmppath, os.path.join(trainer.out, fn))
-    except Exception as e:
-        raise e
     finally:
         shutil.rmtree(tmpdir)
--- a/chainer/training/extensions/_snapshot.py
+++ b/chainer/training/extensions/_snapshot.py
@@ -82,13 +82,12 @@ def _snapshot_object(trainer, target, filename, savefun):
     fn = filename.format(trainer)
     prefix = 'tmp' + fn
 
-    tmpdir = tempfile.TemporaryDirectory(dir=trainer.out)
-    tmppath = os.path.join(tmpdir.name, fn)
-    print(tmpdir.name)
+    tmpdir = tempfile.mkdtemp(prefix=prefix, dir=trainer.out)
+    tmppath = os.path.join(tmpdir, fn)
     try:
         savefun(tmppath, target)
         shutil.move(tmppath, os.path.join(trainer.out, fn))
     except Exception as e:
         raise e
     finally:
-        tmpdir.cleanup()
+        shutil.rmtree(tmpdir)
--- a/chainer/training/extensions/_snapshot.py
+++ b/chainer/training/extensions/_snapshot.py
@@ -68,9 +68,7 @@ def snapshot(savefun=npz.save_npz,
             output file path and the trainer object.
         filename (str): Name of the file into which the trainer is serialized.
             It can be a format string, where the trainer object is passed to
-            the :meth:`str.format` method. Permission of the target file will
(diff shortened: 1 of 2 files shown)


niboshi:
How about creating a temporary directory with `tempfile.mkdtemp`, and creating a temporary file inside it?
That way, permission is set to OS-default, and it seems a cleaner way to me.
(We have to delete the directory manually, though)

belltailjp:
Thanks for the suggestion, I think that's nicer. I don't need to do `os.umask` twice, which is quite dirty.

By the way, if this is considered as a bug, I think I don't have to mention about it in the document, so I will revert document update.

niboshi:
Yes, I think we don't need the mention, as it's the expected behavior for most users.

belltailjp:
Thanks for your comment and suggestion!
I have completed what I think I need to do.

chainer/training/extensions/_snapshot.py
```diff
@@ -81,12 +81,14 @@ def snapshot(trainer):
 def _snapshot_object(trainer, target, filename, savefun):
     fn = filename.format(trainer)
     prefix = 'tmp' + fn
-    fd, tmppath = tempfile.mkstemp(prefix=prefix, dir=trainer.out)
+
+    tmpdir = tempfile.TemporaryDirectory(dir=trainer.out)
```
niboshi: Please use `prefix` here.

chainer/training/extensions/_snapshot.py
```diff
@@ -81,12 +81,14 @@ def snapshot(trainer):
...
+    tmppath = os.path.join(tmpdir.name, fn)
+    print(tmpdir.name)
```
niboshi: Please remove this line.

tests/chainer_tests/training_tests/extensions_tests/test_snapshot.py
```diff
@@ -1,6 +1,7 @@
 import unittest
 
 import mock
+import os
```
niboshi: Move this import to just above `import unittest`. (because `os` is a standard module.)

tests/chainer_tests/training_tests/extensions_tests/test_snapshot.py
```diff
@@ -21,4 +22,29 @@ def test_trigger(self):
...
+        self.assertTrue(os.path.exists('myfile.dat'))
+
+    def test_clean_up_tempdir(self):
+        snapshot = extensions.snapshot_object(self.trainer, 'myfile.dat')
+        snapshot(self.trainer)
+
+        left_tmps = [fn for fn in os.listdir('.') if fn.startswith('tmp')]
+        self.assertEqual(len(left_tmps), 0)
+
+    def tearDown(self):
```
niboshi: Please move `tearDown` to below `setUp`.

chainer/training/extensions/_snapshot.py
kmaehashi: `TemporaryDirectory` seems unavailable in Python 2.7.

niboshi: Good catch. Please use `tempfile.mkdtemp` instead.

belltailjp:
Thanks for your review! Updated the code.

niboshi:
jenkins, test this please

niboshi:
LGTM!

belltailjp:
Thanks so much!!!!!!

PR #4466, opened 2018-03-14
Fix to use CUDNN_BN_MIN_EPSILON

### LINKED ISSUE #1960: Use CUDNN_BN_MIN_EPSILON
`cudnn.h` defines `CUDNN_BN_MIN_EPSILON`.
Remove magic number.
https://github.com/pfnet/chainer/blob/ef14cb9c3c59ca0bf8eb253f79357769e2c290c1/chainer/functions/normalization/batch_normalization.py#L44

Rel #1955.

kmaehashi:
Fixes #1960 using value exposed via https://github.com/cupy/cupy/pull/1011.
Also fixed wrong comments.

### DIFF
--- a/chainer/functions/normalization/batch_normalization.py
+++ b/chainer/functions/normalization/batch_normalization.py
@@ -24,8 +24,8 @@ class BatchNormalization(function_node.FunctionNode):
         self.running_mean = mean
         self.running_var = var
 
-        # Note: cuDNN requires that eps be greater than or equal to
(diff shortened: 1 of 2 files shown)


niboshi:
How about using that constant for generating the error message, instead of hard-coding the actual value?
I also think the comment should not mention the value, because it may be outdated if the value of the constant is changed.

kmaehashi:
Thanks, fixed as suggested.

niboshi:
LGTM!

PR #110, opened 2015-06-30
Model zoo (fix #3)

beam2d:
This PR fixes #3 by adding functions.caffe.CaffeFunction, which loads a pretrained Caffe model and emulates its subnetwork. The usage is written in the reference documentation, though I also show it here. This is an example of loading BVLC reference CaffeNet and computes `fc8` blob from `data` blob.

```python
# Load the model
func = CaffeFunction('path/to/bvlc_reference_caffenet.caffemodel')

# Minibatch of size 10
x_data = numpy.ndarray((10, 3, 227, 227), dtype=numpy.float32)
...  # (Fill the minibatch here)

# Forward the pretrained net
x = Variable(x_data)
y, = func(inputs={'data': x}, outputs=['fc8'])
```

The inputs dict describes how to initialize the blobs by variables. The outputs list indicates which blobs to be output as variables. Also user can extracts parameterized layers from CaffeFunction.fs, which is a FunctionSet object containing Linear or Convolution2D objects.

PR #4470, opened 2018-03-15
Avoid use of `from __future__ print_function`

niboshi:
`from __future__ print_function` can always be replaced with `fileobj.write`.
For Chainer code, it's better to be consistent.

For examples, `from __future__ print_function` may be confusing for beginners.


kmaehashi:
Could you please confirm code style failures?

niboshi:
Fixed.

kmaehashi:
There are still some errors.
```
./examples/modelzoo/evaluate_caffe_net.py:58:59: E901 SyntaxError: invalid syntax
./examples/ptb/train_ptb_custom_loop.py:126:80: E501 line too long (80 > 79 characters)
```

niboshi:
Thanks.

>./examples/modelzoo/evaluate_caffe_net.py:58:59: E901 SyntaxError: invalid syntax

For this error, I think it's better to remove `file=sys.stderr` than using `sys.stderr.write` for consistency with other examples.
How do you think?

kmaehashi:
I agree. It seems there are no strong meaning to use stderr.
https://github.com/chainer/chainer/pull/110/commits/a73b91cb07fda21ac8c4e02cce5d3e30b487b5bd

niboshi:
Thanks, fixed as such.

chainer/function_hooks/debug_print.py
```diff
@@ -50,13 +49,14 @@ class PrintHook(function_hook.FunctionHook):
     name = 'PrintHook'
 
     def __init__(self, sep='', end='\n', file=sys.stdout, flush=True):
+        # TODO(niboshi): deprecate sep
```
kmaehashi: Let's remove this line as https://github.com/chainer/chainer/pull/4471 got merged.

examples/modelzoo/evaluate_caffe_net.py
```diff
@@ -124,8 +122,7 @@ def forward(x, t):
         del x, y, loss, accuracy
 
         count += args.batchsize
-        print('{} / {}'.format(count, len(dataset)), end='\r', file=sys.stderr)
-        sys.stderr.flush()
+        print('{} / {}'.format(count, len(dataset)), end='\r')
```
kmaehashi: `end='\r'` also needs to be removed.

niboshi: Ah, it seems we should replace thsese `print`s with `sys.stdout.write`.

niboshi:
Fixed.

kmaehashi:
LGTM!

PR #4471, opened 2018-03-15
Deprecate `sep` argument of `PrintHook`

niboshi:
It is never used, becaue there's only one argument to `print`.

### DIFF
--- a/chainer/function_hooks/debug_print.py
+++ b/chainer/function_hooks/debug_print.py
@@ -22,7 +22,7 @@ class PrintHook(function_hook.FunctionHook):
     ``backward`` methods without inserting print functions into
     Chainer's library code.
 
-    Attributes:
+    Args:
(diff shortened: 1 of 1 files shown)


chainer/function_hooks/debug_print.py
```diff
@@ -22,7 +23,7 @@ class PrintHook(function_hook.FunctionHook):
     Chainer's library code.
 
     Attributes:
-        sep: Separator of print function.
+        sep: *(deprecated since v4.0.0)* Ignored.
```
kmaehashi: This is a list of attributes, not arguments, so is it better to keep `self.sep = sep` as is?

niboshi: Oh, I suspect it was intended to be argument documentation...

niboshi: Fixed.

kmaehashi:
LGTM!

PR #4473, opened 2018-03-15
Add M-SVAG optimizer

tkerola:
This PR adds the M-SVAG optimizer (Momentum Stochastic Variance-Adapted Gradient): https://arxiv.org/abs/1705.07774

I have found it to generalize better than Adam in my experiments.
Please let me know if you think it's worth including it in Chainer.


chainer/optimizers/msvag.py
```diff
@@ -0,0 +1,178 @@
...
+
+               T factor;
+               if (m == 0 && v == 0)
+                   factor = 0.0;
+               else
+                   factor = min(1.0, max(0.0, mt2 / (mt2 + rho * s)));
+
+               param -= eta * (lr * mt * factor +
+                               weight_decay_rate * param);''',
+            'adam')(grad, hp.lr, 1 - hp.beta,
```
toslunar: Could you rename the kernel: `'adam'` -> `'msvag'` or `'m_svag'`?

chainer/optimizers/msvag.py
```diff
@@ -0,0 +1,178 @@
...
+    def update_core_cpu(self, param):
+        grad = param.grad
+        if grad is None:
+            return
+        hp = self.hyperparam
+        m, v = self.state['m'], self.state['v']
+
+        rho = (((1.0 - hp.beta) ** 2) * (1.0 - self.beta_power ** 2) /
+               (((1.0 - self.beta_power) ** 2) * (1.0 - hp.beta ** 2)))
+        rho = min(rho, 0.99)
```
toslunar: The authors' implementation uses `0.9999`.  It seems to me that this line virtually forces `beta >= 0.01 / 2`.

tkerola: Good catch! Will change it.

toslunar:
Jenkins, test this please.

PR #4476, opened 2018-03-16
add Dockerfiles for Chainer with iDeep

kmaehashi:
Took over task from #4223.


kmaehashi:
I'll configure Docker Cloud to build images for these Dockerfiles.

PR #4479, opened 2018-03-16
Fix stride in the `ResNeXt50` example

akitotakeki:
The original implementation of ResNeXt uses stride 2 in the 3x3 convolution, instead of the first 1x1 convolution when using the bottleneck architecture.
https://github.com/facebookresearch/ResNeXt


vilyaair:
Good observation!

Stride 2 on the 1x1 convolution is used only for the original MSRA ResNet, while stride 2 on the 3x3 convolution is used for facebook Torch models, as is evident from codes of Detectron:
https://github.com/facebookresearch/Detectron/blob/a026d7753512d7f8c1b92235c89e02c38f7b1cea/lib/modeling/ResNet.py#L227-L229
https://github.com/facebookresearch/Detectron/blob/a026d7753512d7f8c1b92235c89e02c38f7b1cea/lib/core/config.py#L871-873

Likewise, I am wondering which version of ResNet we ought to utilize in the chainer, and from my perspective, I prefer stride 2 on the 3x3 convolution.

bkvogel:
Thank you for the PR! I think it looks good. I understand the current Chainer ResNet follows the original MSRA ResNet, right? vilyaair, If you would like it to also support the "Facebook-style" ResNet with stride 2 on the 3x3 conv, one idea would be to add an optional keyword argument to the `__init__()` of `Resent50` so that the user can configure the mode to either "MSRA" or "Facebook", with "MSRA" being default so as not to break backward compatibility. What do you think? I think it is also fine to do this as a separate PR if you prefer me to merge this one first. Any thoughts?

vilyaair:
@bkvogel  I believe it is sensible choice to fix ResNeXt50 example. I do not mind either methods, and personally prefer to implement it in this PR with a view to possible less review workload.

bkvogel:
Thank you for your comments. @akitotakeki, what do you think?

akitotakeki:
@bkvogel  I also do not mind fixing the ResNeXt50 example in either way.

vilyaair:
@akitotakeki 
I believe that your commit  8ec7f4a is pretty good, and accordingly prepare to close my PR. I am wondering why you reverted this commit?

akitotakeki:
I thought it would be better to revert PR to comply with the title of this PR, but I'll follow your opinion.

akitotakeki:
Could you review this?

bkvogel:
Sorry, I did not notice the latest changes. I think it looks good.
Jenkins, test this please.

bkvogel:
It seems that the train_imagenet example is broken by this change:

10:23:36   File "examples/imagenet/train_imagenet.py", line 166, in <module>
10:23:36     main()
10:23:36   File "examples/imagenet/train_imagenet.py", line 77, in main
10:23:36     'resnext50': resnet50.ResNeXt50,
10:23:36 AttributeError: 'module' object has no attribute 'ResNeXt50'

I think maybe you need to update the `import resent50` line to point to the new location of ResNeXt50? Can you make the change? Thanks!

akitotakeki:
I added the `import resnext` line.

bkvogel:
Thanks. I'll merge once the tests pass.
Jenkins, test this please.

bkvogel:
LGTM!

PR #4486, opened 2018-03-19
Fix sectioning of "Datasets" reference

beam2d:
I found that the toc of reference manual has incorrect section levels. This PR fixes it.

Before:
image

After:
image


okuta:
LGTM!

PR #4497, opened 2018-03-19
Implement Variable.xp

### LINKED ISSUE #4113: Feature request: Variable.xp to get the array module of its data array
I often need to call `chainer.cuda.get_array_module` for `Variable` objects. Like `Link.xp`, `Variable.xp` would be a useful property.

kmaehashi:
Fixes #4113.

One thing to discuss is that what to return when the variable is not initialized (`Variable(None)`).
Considering that the following facts, I decided to return NumPy in such case:

* `xp` for empty Link returns NumPy (i.e., `chainer.Link().xp == numpy`)
* Users can safely replace `chainer.cuda.get_array_module(v)` with `v.xp` without considering any side-effects

### DIFF
--- a/chainer/variable.py
+++ b/chainer/variable.py
@@ -488,6 +488,16 @@ Actual: {0}'''.format(type(data))
         return variable_str(self)
 
     @property
+    def xp(self):
+        """Array module for this variable.
+
+        Depending on which of CPU/GPU this variable is on, this property
+        returns :mod:`numpy` or :mod:`cupy`.
+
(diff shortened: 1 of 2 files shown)


okuta:
It is good feature. I will merge this.
Cloud you add test?

kmaehashi:
Thanks, I added tests.
I added `to-be-backported` as I think this is OK to be backported to v4.

okuta:
LGTM!

PR #4499, opened 2018-03-20
Call FunctionHooks in chainer.grad

### LINKED ISSUE #4116: `chainer.grad` does not call `backward_{pre,post}process` of function hooks
The following code can be used for reproduction. It does not print anything on `grad`.

```py
import chainer
import numpy

class MyHook(chainer.FunctionHook):
    def __init__(self, name):
        self.name = name

    def backward_preprocess(self, f, in_d, out_g):
        print(self.name, ':', 'preproc', f.label)

    def backward_postprocess(self, f, in_d, out_g):
        print(self.name, ':', 'postproc', f.label)

x = chainer.Variable(numpy.array([2.]))
y = x * x
with MyHook('backward'):
    y.backward()
with MyHook('grad'):
    chainer.grad([y], [x])
```

which shows

```
backward : preproc _ * _
backward : postproc _ * _
```

I confirmed it with Chainer 3.2.0 and 4.0.0b2.

hvy:
Fixes #4116 .

### DIFF
--- a/chainer/function_node.py
+++ b/chainer/function_node.py
@@ -891,7 +891,27 @@ def _backprop(outputs, inputs, grad_required, retain_grad, grads, loss_scale):
         gys = tuple([gy if not isinstance(gy, tuple) else
                      chainer.functions.add(*gy)
                      for gy in gys])
+
+        # Call pre-backward hooks
+        hooks = chainer.get_function_hooks()
+        if func._n_local_function_hooks != 0:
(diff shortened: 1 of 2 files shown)


kmaehashi:
Jenkins, test this please.

kmaehashi:
LGTM!

kmaehashi:
Added `to-be-backported` as this is a bug.

PR #4503, opened 2018-03-20
Travis test against v4 branch

kmaehashi:
As `v4` branch is now stable version, we should change the branch name used in Travis.


okuta:
LGTM!

PR #4505, opened 2018-03-20
Remove eps from batch normalization statistics

toslunar:
Fix that `eps` is doubly added in test mode.  This PR changes the behavior of `F.batch_normalization` if it is called without cudnn.  (Therefore it also changes `L.BatchNormalization`.)


okuta:
LGTM!

PR #4506, opened 2018-03-20
Fix doctest

okuta:


kmaehashi:
How about marking the doctest as SKIP (considering this behavior is tested in the unit test)?

```py
>>> dataset.concat_examples(w)  # doctest: +SKIP
```

Other solutions proposed in the document seems inappropriate for example codes. https://docs.python.org/3/library/doctest.html#warnings

okuta:
I fixed.

hvy:
jenkins, test this please.

kmaehashi:
LGTM!

PR #4510, opened 2018-03-21
Add chainer.config.dtype and use it in initializers and dataset loaders.

beam2d:
This PR aims at making it easy to use fp16. By setting an environment variable `CHAINER_DTYPE=float16`, you can let weight initializers and datasets generate fp16 arrays by default instead of fp32. It makes each model definition independent from which dtype to use.

I confirmed that MNIST example runs without modifying other parts (except the eps parameter of Adam, whose default value is too small for fp16). This change does not alter the behavior of existing code unless `CHAINER_DTYPE` is set.

### DIFF
--- a/chainer/__init__.py
+++ b/chainer/__init__.py
@@ -142,7 +142,12 @@ global_config.autotune = False
 global_config.use_ideep = os.environ.get('CHAINER_USE_IDEEP', 'never')
 global_config.lazy_grad_sum = bool(int(
     os.environ.get('CHAINER_LAZY_GRAD_SUM', '0')))
-global_config.dtype = numpy.dtype(os.environ.get('CHAINER_DTYPE', 'float32'))
+
+_chainer_dtype = os.environ.get('CHAINER_DTYPE', 'float32')
+if _chainer_dtype not in ('float16', 'float32', 'float64'):
+    raise TypeError('incorrect dtype name in CHAINER_DTYPE: "{}". '
+                    'Only float16/32/64 are allowed.'.format(_chainer_dtype))
+global_config.dtype = numpy.dtype(_chainer_dtype)
 
 
 def is_debug():
--- a/chainer/__init__.py
+++ b/chainer/__init__.py
@@ -142,6 +142,7 @@ global_config.autotune = False
 global_config.use_ideep = os.environ.get('CHAINER_USE_IDEEP', 'never')
 global_config.lazy_grad_sum = bool(int(
     os.environ.get('CHAINER_LAZY_GRAD_SUM', '0')))
+global_config.dtype = numpy.dtype(os.environ.get('CHAINER_DTYPE', 'float32'))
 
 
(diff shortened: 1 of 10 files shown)


chainer/__init__.py
```diff
@@ -150,6 +150,7 @@ def is_arrays_compatible(arrays):
 global_config.use_ideep = os.environ.get('CHAINER_USE_IDEEP', 'never')
 global_config.lazy_grad_sum = bool(int(
     os.environ.get('CHAINER_LAZY_GRAD_SUM', '0')))
+global_config.dtype = numpy.dtype(os.environ.get('CHAINER_DTYPE', 'float32'))
```
kmaehashi: It's better to show a nice error message if user specified wrong value.

chainer/datasets/cifar.py
```diff
@@ -130,20 +134,21 @@ def load(archive, file_name):
...
-def _preprocess_cifar(images, labels, withlabel, ndim, scale):
+def _preprocess_cifar(images, labels, withlabel, ndim, scale, dtype):
     if ndim == 1:
         images = images.reshape(-1, 3072)
     elif ndim == 3:
         images = images.reshape(-1, 3, 32, 32)
     else:
         raise ValueError('invalid ndim for CIFAR dataset')
-    images = images.astype(numpy.float32)
+    dtype = chainer._get_dtype(dtype)
```
kmaehashi: Need `import chainer` and `_get_dtype` -> `get_dtype`

kmaehashi:
Besides these fixes, we need to fix code that uses fp32 by default. e.g., https://github.com/chainer/chainer/blob/v4.0.0rc1/chainer/links/normalization/batch_normalization.py#L65

chainer/__init__.py
beam2d: Added a check for the value of CHAINER_DTYPE.

chainer/datasets/cifar.py
beam2d: fixed

beam2d:
I will check other parts of Chainer (including BN you wrote) that should be fixed.

beam2d:
I checked functions and links that should be fixed to support fp16 mode, and found that there are so many 
things to fix. I feel it's better to fix them in separate PRs to make review quicker. I will make an issue to manage them.

kmaehashi:
@beam2d How do you want to proceed with this PR? I think it's OK to merge this first, then fix other functions and links (#4582).

beam2d:
I think it's good to first merge this PR and then proceed to #4582.

kmaehashi:
Jenkins, test this please.

beam2d:
I found that adding `.. currentmodule:: chainer` resolves the problem and still keeps `initializers` appear at the module index. I fixed it and rebased to the current master.

kmaehashi:
LGTM!

PR #4495, opened 2018-03-19
document arithmetic special functions

kmaehashi:
This fixes #3821.
Related to #4480.

PR #4516, opened 2018-03-22
improve docs for supported operators in Variable

kmaehashi:
After the offline discussion with @hvy regarding #4495, I thought it is better to explicitly state what operators are supported for Variables.

### DIFF
--- a/chainer/variable.py
+++ b/chainer/variable.py
@@ -422,8 +422,10 @@ class Variable(object):
     :class:`~chainer.variable.VariableNode` object of
     a computational graph. If the variable is constructed by the user, the node
     is *root* and does not hold any parent. If the variable is constructed by a
-    :class:`~chainer.FunctionNode` object, the node holds a reference to its
-    parent called :attr:`creator_node`.
+    :class:`~chainer.FunctionNode` object (i.e., by calling functions under
+    ``chainer.functions`` or user-defined functions), or by using operators
+    (see the list below), the node holds a reference to its parent called
+    :attr:`creator_node`.
     This reference is used in backpropagation to backtrack the graph.
 
     Users can disable (resp. enable) this chaining behavior by calling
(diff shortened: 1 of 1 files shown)


chainer/variable.py
```diff
@@ -432,6 +434,23 @@ class Variable(object):
...
+        * Subtraction: ``a - b`` (:meth:`__sub__`, :meth:`__rsub__`)
+        * Multiplication: ``a * b`` (:meth:`__mul__`, :meth:`__rmul__`)
+        * Division: ``a / b`` (:meth:`__div__`, :meth:`__rdiv__`, \
+                               :meth:`__truediv__`, :meth:`__rtruediv__`)
+        * Floor Division: ``a // b`` (:meth:`__floordiv__`, \
+                                      :meth:`__rfloordiv__`)
+        * Exponentiation: ``a ** b`` (:meth:`__pow__`, :meth:`__rpow__`)
+        * Matirx Multiplication: ``a @ b`` (:meth:`__matmul__`, \
+                                            :meth:`__rmatmul__`)
+        * Negation (Arithmetic): ``- a`` (:meth:`__neg__`)
```
hvy: How about `-a`, removing the whitespace?

hvy:
Do you think `abs` should be added to the list as well?

chainer/variable.py
kmaehashi: I followed expressions used in https://docs.python.org/3/library/operator.html#mapping-operators-to-functions.

kmaehashi:
`abs` is not an operator, so I think just documenting `__abs__` is enough.
(I feel it is better to encourage use of `F.abs` rather than `abs` for consistency.)

hvy:
I see, thanks for the clarification. Then let's leave it as is.

hvy:
jenkins, test this please.

hvy:
jenkins, test this please.

hvy:
jenkins, test this please.

hvy:
LGTM!

Issue #4513, opened 2018-03-22
Support for discriminative loss function


dBeker:
As suggested in contribution guide, I open an issue for a new feature to be implemented and to discuss the design.  
  
I have implemented "discriminative loss function" described in https://arxiv.org/pdf/1708.02551.pdf. This loss can improve the segmentation quality along with some post-processing clustering steps.   

I would like to create a PR following the design discussion.

delta2323:
bump

kmaehashi:
Implemented in #5313

PR #4525, opened 2018-03-26
Add discriminative loss

dBeker:
This is the PR for issue #4513


hknerdgn:
You should not call it "discriminative loss" since there may be many different "discriminative" losses and it is very general to call it that. The naming in the paper is unfortunate I think. You may call it discriminative-margin-based-clustering loss  or something like that.

dBeker:
Suc a naming makes sense. I will rename it, thanks.

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,248 @@
+import cupy as cp
+import numpy as np
+
+import chainer.functions as F
+
+
+class DiscriminativeMarginBasedClusteringLoss:
```
delta2323: Flake8 complains old-style class declaration. Use new-style one (`class Foo(object)`).

delta2323:
Please register the function to the document [here](https://github.com/chainer/chainer/blob/master/docs/source/reference/functions.rst#loss-functions).

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,248 @@
...
+    of trained network.
+    For cross-entropy based approaches, if the pixel value is wrong,
+    the loss value will be same independent from wrong pixel's location.
+    However, for segmentation, even though network gives wrong pixel output, 
+    it is desirable to have it
+    as close as possible to the original position.
+    By applying discriminative loss function, groups of segmentation instances 
+    can be moved together.
+
+    This loss function calculates three different parameters:
```
delta2323: We follow [Google style guide](http://sphinxcontrib-napoleon.readthedocs.io/en/latest/index.html#google-vs-numpy) in writing documents by using Napoleon, which is an extension of Sphinx that allows us to write documents following the guideline. So, could you write in that way?

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,248 @@
...
+        if norm != 1:
+            raise Exception(
+                'Currently discriminative loss only supports L1 Norm')
+
+        if norm == 2:
+            self.norm = self.l2_norm
+        else:
+            self.norm = self.l1_norm
+
+        return
```
delta2323: We do not have to write `return` to have `__init__` return `None`.

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,248 @@
...
+        self.beta = beta
+        self.gamma = gamma
+        self.max_n_clusters = max_n_clusters
+
+        # L1 or L2 norm is allowed only
+        if norm != 1:
+            raise Exception(
+                'Currently discriminative loss only supports L1 Norm')
+
+        if norm == 2:
```
delta2323: I think this condition is never held because `norm` should be `1` by the previous `if` statement. Do you want to do some thing like this?

```python
if norm == 1:
    self.norm = self.l1_norm
elif norm == 2:
    self.norm = self.l2_norm
else:
    raise Exception(...)
```

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,248 @@
...
+        return
+
+    def l1_norm(self, x, axis):
+        return F.sum(F.absolute(x), axis=axis)
+
+    def l2_norm(self, x, axis):
+        return F.sqrt(F.sum(x ** 2, axis=axis))
+
+    def variance_term(self, pred, gt, means, delta_v, gt_idx):
+        """
```
delta2323: Write 1-line summary of this method in the first line (otherwise, it violates flake8).

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,248 @@
...
+        :param x:
+        x[0] = segmentation prediction output
+        x[1] = segmentation ground truth
+        x[2] = number of objects in ground truth
+        x[3] = indexes of non-zero ground truths
+        :return:
+        Loss value
+        """
+
+        assert len(x) == 4
```
delta2323: If we do not use variable-length argument, we do not have to check the number of arguments with assertion.

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,248 @@
...
+        assert len(x[0].shape) == 4
```
delta2323: if we assume `x[0]` is NumPy or CuPy ndarray, `x[0].ndim == 4` is equivalent.

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,248 @@
...
+        assert len(x[1].shape) == 4
```
delta2323: ditto (use `x[1].ndim == 4` if it is a ndarray).

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,248 @@
+import cupy as cp
```
delta2323: We would like to write Chainer functions in CPU/GPU-agnostic way. When users run code in CPU, `cupy` is not needed (and is typically not unavailable). One typical way to check in code is to use [`get_array_module`](https://docs.chainer.org/en/v4/reference/util/generated/chainer.backends.cuda.get_array_module.html). See its usage e.g. [here](https://github.com/chainer/chainer/blob/3ba7d4d0ddcc0202225161f75e6d6fed07ffa25e/chainer/functions/math/ceil.py).

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,248 @@
...
+
+        _var = self.norm((p - m), 2)
+        _var = F.maximum(cp.asarray(0.0, np.float32),
+                         _var - dv) ** 2  # Suppress inlier distance
+        _var = _var * g[:, :, 0, :]
+
+        var_term = 0.0
+        for i in range(bs):
+            var_term += F.sum(_var[i, gt_idx[i]]) / F.sum(g[i, gt_idx[i], 0])
+        var_term = var_term / cp.asarray(bs, dtype=np.float32)
```
delta2323: I think `ver_term /= bs` works without converting `bs` into ndarray.

dBeker:
@delta2323 Thanks for your time for checking the branch. I have pushed the required changes.

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,345 @@
...
+        _var = _var * g[:, :, 0, :]
+
+        var_term = 0.0
+        for i in range(bs):
+            var_term += F.sum(_var[i, gt_idx[i]]) / F.sum(g[i, gt_idx[i], 0])
+        var_term /= bs
+
+        return var_term
+
+    def distance_term(self, means, delta_d, n_objects):
```
delta2323: same as `variance_term`

delta2323:
Please add unit tests for this function class. Tests for other Chainer functions (e.g. [here](https://github.com/chainer/chainer/blob/master/tests/chainer_tests/functions_tests/activation_tests/test_softplus.py)) can be helpful to understand how to do that.

delta2323:
Sorry for making you wait. I'll check the update.

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,350 @@
...
+        self.gamma = gamma
+        self.max_n_clusters = max_n_clusters
+
+        # L1 or L2 norm is allowed only
+        if norm == 1:
+            self.norm = self._l1_norm
+        elif norm == 2:
+            self.norm = self._l2_norm
+        else:
+            raise Exception("Norm can only be 1 or 2")
```
delta2323: Could you use `ValueError`[https://docs.python.org/3/library/exceptions.html#ValueError], which is used when a function get wrong arguments?

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,350 @@
...
+        # Calculate losses
+        l_var = self._variance_term(prediction, labels, c_means, self.delta_v,
+                                    gt_idx)
+        l_dist = self._distance_term(c_means, self.delta_d, n_objects)
+        l_reg = self._regularization_term(c_means, n_objects)
+
+        return self.alpha * l_var + self.beta * l_dist + self.gamma * l_reg
+
+    def __call__(self, *args):
+        return self.apply(*args)
```
delta2323: Could you let me know why you prepare `apply` method? If the reason is to align with implementation of `FunctionNode`, I think we do not have to do that because this class is not a subclass of `FunctionNode`. Just providing `__call__` method would be enough.

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,350 @@
...
+            (Push force)
+        - Regularization loss:
+            Small regularization loss to penalize weights against overfit
+
+    Args:
+        x (tuple) : Contains several inputs
+                - x[0] = segmentation prediction output
+                - x[1] = segmentation ground truth
+                - x[2] = number of objects in ground truth
+                - x[3] = indexes of non-zero ground truths
```
delta2323: might be helpful if we provide shapes of each element.

delta2323:
Could you add `from chainer.functions.loss.discriminative_loss.py import discriminative_margin_based_clustering_loss  # NOQA` to `chainer/functions/__init__.py` to enable the shortcut `chainer.functions.discriminative_margin_based_clustering_loss` ?

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,350 @@
+from chainer.backends import cuda
+import chainer.functions as F
```
delta2323: To avoid cyclic dependency in executing `import chainer`, do not use alias names such as `chainer.functions.expand_dims`. Could you use full path such as `chainer.functions.array.expand_dims.expand_dims`? Implementation of VAE losses [code](https://github.com/chainer/chainer/blob/master/chainer/functions/loss/vae.py) might be helpful.

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,350 @@
...
+        """Function to preprocess inputs
+
+        Args:
+            prediction (nd-array): Prediction output
+            labels (nd-array): Ground truth output
+        Returns:
+            nd-array : Prediction output
+            nd-array : Ground truth output
+        """
+        # Reshape layers to prepare for processing
```
delta2323: I think it would be better to write this comment to the document of this function.

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,350 @@
...
+
+            p_sum = F.sum(p_item, axis=2)
+            g_sum = F.cast(F.sum(g_item, axis=2), p_sum.dtype)
+            _mean_sample = p_sum / g_sum
+
+            n_fill_objects = max_n_objects - n_objects[i]
+
+            if n_fill_objects != 0:
+                _fill_sample = module.zeros(
+                    (module.asnumpy(n_fill_objects), n_filters),
```
delta2323: Do we have to convert `n_fill_objects` to NumPy array? I guess `module.zeros(n_fill_objects n_filters)` should work.

dBeker: Unfortunately, it does not work and we get `TypeError: 'cupy.core.core.ndarray' object cannot be interpreted as an integer` error. That's why I added this conversion. That's why I added this conversion.

dBeker:
@delta2323 I have completed the requested changes. I would be glad if you can check. Thanks

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,378 @@
...
+from chainer.functions.array.cast import cast
+from chainer.functions.array.concat import concat
+from chainer.functions.array.expand_dims import expand_dims
+from chainer.functions.array.reshape import reshape
+from chainer.functions.array.stack import stack
+from chainer.functions.array.transpose import transpose
+from chainer.functions.math.average import average
+from chainer.functions.math.basic_math import absolute
+from chainer.functions.math.maximum import maximum
+from chainer.functions.math.sum import sum
```
delta2323: Python has `sum` as a built-in function. So I think it is better not to use the name and to hide the built-in function.

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,378 @@
...
+        delta_v (float): Minimum distance to start penalizing variance
+        delta_d (float): Maximum distance to stop penalizing distance
+        max_n_clusters (int): Maximum possible number of clusters.
+        norm (int): Norm to calculate pixels and cluster center distances
+        alpha (float): Weight for variance loss      (alpha * variance_loss)
+        beta (float): Weight for distance loss       (beta * distance_loss)
+        gamma (float): Weight for regularization loss(gamma * regularizer_loss)
+
+    Returns:
+        float: (alpha * variance_loss)+
```
delta2323: I think the returned value is not a float but a Chainer variable holding a NumPy or CuPy array whose shape is `()`.

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,378 @@
...
+            Loss to penalize distances between pixels which are belonging
+            to same instance. (Pull force)
+        - Distance loss:
+            Loss to penalize distances between the centers of instances.
+            (Push force)
+        - Regularization loss:
+            Small regularization loss to penalize weights against overfit
+
+    Args:
+        x (tuple) : Contains several inputs
```
delta2323: Could you explicitly write that each elements of `x` is either `numpy.ndarray`, `cupy.ndarray` or `Variable` holding them (e.g. [F.ReLU](https://github.com/chainer/chainer/blob/master/chainer/functions/activation/relu.py#L152))

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,378 @@
...
+        elif norm == 2:
+            self.norm = self._l2_norm
+        else:
+            raise ValueError("Norm can only be 1 or 2")
+
+    def _l1_norm(self, x, axis):
+        """Function to calculate L1 Norm by given axes
+
+        Args:
+            x (nd-array): Input matrix to calculate norm
```
delta2323: This function takes either `numpy.ndarray` or `cupy.ndarray` or Chainer variable holding them as `x`. So could you replace `nd-array` with 

```
:class:`~chainer.Variable\` or :class:\`numpy.ndarray\` or :class:`cupy.ndarray`
```

as we do in other Chainer functions (e.g. [ReLU](https://github.com/chainer/chainer/blob/master/chainer/functions/activation/relu.py#L152))

delta2323: Same is true for other documents using `nd-array`.

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,378 @@
...
+        """Function to calculate variance term
+
+        Args:
+            pred (nd-array): Prediction output
+            gt (nd-array): Ground truth output
+            means (nd-array): Instance means
+            delta_v (float): Coefficient to decide 'pull force' power
+            gt_idx (tuple / nd-array): Indexes of ground truth instances
+        Returns:
+            float : variance loss
```
delta2323: ditto (the returned value is Chainer variable)

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,378 @@
...
+    def _distance_term(self, means, delta_d, n_objects):
+        """Function to calculate distance term
+
+        Args:
+            means (nd-array): Instance means
+            delta_d (float): Coefficient to decide 'push force' power
+            n_objects (nd-array): Instance count in current input
+
+        Returns:
+            float : distance loss
```
delta2323: ditto (the returned value is Chainer variable)

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,378 @@
...
+            - Add weights to all and return loss value
+
+        Args:
+            x (tuple) : Contains several inputs
+                - x[0] = segmentation prediction output
+                - x[1] = segmentation ground truth
+                - x[2] = number of objects in ground truth
+                - x[3] = indexes of non-zero ground truths
+        Returns:
+            float : Loss value
```
delta2323: ditto (the returned value is Chainer variable)

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,378 @@
...
+        max_n_clusters (int): Maximum possible number of clusters.
+        norm (int): Norm to calculate pixels and cluster center distances
+        alpha (float): Weight for variance loss      (alpha * variance_loss)
+        beta (float): Weight for distance loss       (beta * distance_loss)
+        gamma (float): Weight for regularization loss(gamma * regularizer_loss)
+
+    Returns:
+        float: (alpha * variance_loss)+
+               (beta * distance_loss) +
+               (gamma * regularizer_loss)
```
delta2323: ditto (the returned value is Chainer variable)

tests/chainer_tests/functions_tests/loss_tests/test_discriminative_margin_based_loss.py
```diff
@@ -0,0 +1,96 @@
...
+import numpy
+
+import chainer
+from chainer.backends import cuda
+from chainer import functions
+from chainer import testing
+from chainer.testing import attr
+
+
+@testing.parameterize(*(testing.product({
```
delta2323: I think we do not have to use `testing.product`. 

```
@testing.parameterize([
   {'delta_v': 0.5, 'delta_d': 5, ....},
   {'delta_v': 3, 'delta_d': 10, ....}
])
```
is enough.

tests/chainer_tests/functions_tests/loss_tests/test_discriminative_margin_based_loss.py
```diff
@@ -0,0 +1,96 @@
...
+        out = functions.discriminative_margin_based_clustering_loss(
+            x_data, self.delta_v, self.delta_d,
+            self.max_n_clusters, self.norm, self.alpha,
+            self.beta, self.gamma)
+        return out
+
+    def check_forward_cpu(self, x_data, t_data):
+        t = chainer.Variable(t_data)
+        out = self.get_result(x_data)
+        assert(out.data == t.data)
```
delta2323: I think it is better to admit some numerical errors by using e.g. [`numpy.testing.assert_almost_equal`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.testing.assert_almost_equal.html)

tests/chainer_tests/functions_tests/loss_tests/test_discriminative_margin_based_loss.py
```diff
@@ -0,0 +1,96 @@
...
+
+    def check_forward_gpu(self, x_data, t_data):
+        t = chainer.Variable(t_data)
+        out = self.get_result(x_data)
+        out.to_cpu()
+        t.to_cpu()
+        assert (out.data == t.data)
```
delta2323: ditto (consider numerical errors)

tests/chainer_tests/functions_tests/loss_tests/test_discriminative_margin_based_loss.py
```diff
@@ -0,0 +1,96 @@
...
+    @attr.gpu
+    def test_forward_gpu_cpu(self):
+        inputs_gpu = [cuda.to_gpu(self.input), cuda.to_gpu(self.gt),
+                      self.gt_obj, self.gt_obj_idx]
+        inputs_cpu = [cuda.to_cpu(self.input), cuda.to_cpu(self.gt),
+                      self.gt_obj, self.gt_obj_idx]
+        cpu_res = self.get_result(inputs_cpu)
+        gpu_res = self.get_result(inputs_gpu)
+        gpu_res.to_cpu()
+        assert(cpu_res.data == gpu_res.data)
```
delta2323: ditto (consider numerical errors)

chainer/functions/loss/discriminative_loss.py
delta2323: I found sphinx-build raises a syntax error at this line. Could you investigate and fix it?

```
sphinx-build -b html -d build/doctrees source build/html

...(omit)...

Exception occurred:
  File "/Users/oonokenta/dev/chainer/docs/source/_docstring_check.py", line 46, in error
    raise InvalidDocstringError(msg, self, iline)
_docstring_check.InvalidDocstringError: Invalid indentation of `Returns` section

on chainer.functions.discriminative_margin_based_clustering_loss
at line 55: "                 (beta * distance_loss) +"

docstring:
00 Discriminative margin based clustering loss function
01 
02 This is the implementation of following paper:
03 https://arxiv.org/pdf/1802.05591.pdf
04 
05 In segmentation, one of the biggest problem is having noise at the output
06 of a trained network.
07 For cross-entropy based approaches, if the pixel value is wrong,
08 the loss value will be same independent from wrong pixel's location.
09 However, for segmentation, even though network gives wrong pixel output,
10 it is desirable to have it
11 as close as possible to the original position.
12 By applying discriminative loss function, groups of segmentation instances
13 can be moved together.
14 
15 This loss function calculates three different parameters:
16     - Variance Loss:
17         Loss to penalize distances between pixels which are belonging
18         to same instance. (Pull force)
19     - Distance loss:
20         Loss to penalize distances between the centers of instances.
21         (Push force)
22     - Regularization loss:
23         Small regularization loss to penalize weights against overfit
24 
25 :param x: Contains several inputs
26           - x[0] = segmentation prediction output     (n, i, w, h)
27           - x[1] = segmentation ground truth          (n, i, w, h)
28           - x[2] = number of objects in ground truth  (n,)
29           - x[3] = indexes of non-zero ground truths  (n, variable)
30 :type x: tuple
31 :param where,:
32                - n is batch size
33                - i is total instance count
34                - w is width of the image
35                - h is height of the image
36                - variable is variable length depending on list size
37 :param delta_v: Minimum distance to start penalizing variance
38 :type delta_v: float
39 :param delta_d: Maximum distance to stop penalizing distance
40 :type delta_d: float
41 :param max_n_clusters: Maximum possible number of clusters.
42 :type max_n_clusters: int
43 :param norm: Norm to calculate pixels and cluster center distances
44 :type norm: int
45 :param alpha: Weight for variance loss      (alpha * variance_loss)
46 :type alpha: float
47 :param beta: Weight for distance loss       (beta * distance_loss)
48 :type beta: float
49 :param gamma: Weight for regularization loss(gamma * regularizer_loss)
50 :type gamma: float
51 
52 :returns:
53 
54           (alpha * variance_loss)+
55                  (beta * distance_loss) +
56                  (gamma * regularizer_loss)
57 :rtype: float
58
```

delta2323:
Could you add tests for backward? [`gradient_check.check_backward`](https://docs.chainer.org/en/stable/reference/generated/chainer.gradient_check.check_backward.html) and [`gradient_check.check_double_backward`](https://github.com/chainer/chainer/blob/v4.1.0/chainer/gradient_check.py#L575) may be useful. See e.g. [here](https://github.com/chainer/chainer/blob/master/tests/chainer_tests/functions_tests/loss_tests/test_squared_error.py#L54) for their usage.

dBeker:
I wonder if we need to add any backward test. As this loss does not introduce any backward layer but it is using built-in functions, as soon as we test each of those layers individually, this function's backward computation should also be tested.

delta2323:
I wanted to confirm this function is entirely composed of differentiable functions, that is, if a computational graph is correctly created. In that sense, it might be OK to check only the gradient associated to input variables are modified by backpropagating from the final loss.

dBeker:
@delta2323 I have fixed the issues that you mentioned. I would be glad if you can check it. Thanks.

delta2323:
Thank you for fixing the code. I'll take a look tomorrow.

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,467 @@
...
+            self.norm = self._l2_norm
+        else:
+            raise ValueError("Norm can only be 1 or 2")
+
+    def _l1_norm(self, x, axis):
+        """Function to calculate L1 Norm by given axes
+
+        Args:
+            x (:class:`~chainer.Variable\` or \
+            :class:\`numpy.ndarray\` or \
```
delta2323: Are back slashes of backquotes intentional?

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,467 @@
...
+    of a trained network.
+    For cross-entropy based approaches, if the pixel value is wrong,
+    the loss value will be same independent from wrong pixel's location.
+    However, for segmentation, even though network gives wrong pixel output,
+    it is desirable to have it
+    as close as possible to the original position.
+    By applying discriminative loss function, groups of segmentation instances
+    can be moved together.
+
+    This loss function calculates three different parameters:
```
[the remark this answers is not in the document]
delta2323: Could you remove the empty line of l.415?

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,467 @@
...
+                       :class:`cupy.ndarray`) :
+                         segmentation ground truth          (n, i, w, h)
+                - x[2](:class:`~chainer.Variable` or \
+                       :class:`numpy.ndarray` or \
+                       :class:`cupy.ndarray`) :
+                         number of objects in ground truth  (n,)
+                - x[3](:class:`~chainer.Variable` or \
+                       :class:`numpy.ndarray` or \
+                       :class:`cupy.ndarray`) :
+                         indexes of non-zero ground truths  (n, variable)
```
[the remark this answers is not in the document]
delta2323: Could you add an empty line between l.427 and l.428?

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,467 @@
...
+        where,
+            - n is batch size
+            - i is total instance count
+            - w is width of the image
+            - h is height of the image
+            - variable is variable length depending on list size
```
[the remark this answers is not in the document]
delta2323: Could you change as follows?

```diff
-        where,
-            - n is batch size
-            - i is total instance count
-            - w is width of the image
-            - h is height of the image
-            - variable is variable length depending on list size
+            where,
+
+                - n is batch size
+                - i is total instance count
+                - w is width of the image
+                - h is height of the image
+                - variable is variable length depending on list size
```

chainer/functions/loss/discriminative_loss.py
dBeker: Yes, they are intentional. Do you suggest to remove?

dBeker:
@delta2323 I have pushed the changes

chainer/functions/loss/discriminative_loss.py
delta2323: I'm just curious why you did so because you didn't put back slashes in other places.

dBeker: If it is a multi-line explanation of the type, I appended backslashes for easier readability. If it is a single line explanation, I did not add. As the only multi-line type explanation is :class:`~chainer.Variable\` or \
:class:\`numpy.ndarray\` or \
:class:`cupy.ndarray` 
I used backslashes only for those.

delta2323: What I meant was you added back slashes to backquotes of `numpy.ndarray` (i.e. `` \`numpy.ndarray\` ``) while you do not those of `cupy.ndarray` (i.e. `` `cupy.ndarray` ``) .

delta2323:
`make html` cause a warning, which is treated as an error by configuration. I  pass the command with the following change. Could you try it?

```
%(pr-4525)[U]git diff
diff --git a/chainer/functions/loss/discriminative_loss.py b/chainer/functions/loss/discriminative_loss.py
index 9613e1d0a..faf1765ea 100644
--- a/chainer/functions/loss/discriminative_loss.py
+++ b/chainer/functions/loss/discriminative_loss.py
@@ -413,6 +413,7 @@ def discriminative_margin_based_clustering_loss(
     By applying discriminative loss function, groups of segmentation instances
     can be moved together.
     This loss function calculates three different parameters:
+
         - Variance Loss:
             Loss to penalize distances between pixels which are belonging
             to same instance. (Pull force)
@@ -424,7 +425,6 @@ def discriminative_margin_based_clustering_loss(
 
     Args:
         x (tuple) : Contains several inputs
-
                 - x[0](:class:`~chainer.Variable` or \
                        :class:`numpy.ndarray` or \
                        :class:`cupy.ndarray`) :
```

chainer/functions/loss/discriminative_loss.py
dBeker: Oh, now got it. That was a mistake. I cleaned them. Thanks for catching.

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,467 @@
...
+        l_reg = self._regularization_term(c_means, gt_idx)
+
+        chainer.config.type_check = buffer
+        return self.alpha * l_var + self.beta * l_dist + self.gamma * l_reg
+
+
+def discriminative_margin_based_clustering_loss(
+        x, delta_v, delta_d, max_n_clusters,
+        norm=1, alpha=1.0, beta=1.0, gamma=0.001):
+    """Discriminative margin based clustering loss function
```
tkerola: `margin based` -> `margin-based`

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,467 @@
...
+
+    This is the implementation of following paper:
```
tkerola: `the following paper`

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,467 @@
...
+    https://arxiv.org/pdf/1802.05591.pdf
+
+    In segmentation, one of the biggest problem is having noise at the output
+    of a trained network.
+    For cross-entropy based approaches, if the pixel value is wrong,
+    the loss value will be same independent from wrong pixel's location.
```
tkerola: `the wrong pixel's`

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,467 @@
...
+    However, for segmentation, even though network gives wrong pixel output,
```
tkerola: `the network gives`

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,467 @@
...
+    it is desirable to have it
+    as close as possible to the original position.
+    By applying discriminative loss function, groups of segmentation instances
+    can be moved together.
+    This loss function calculates three different parameters:
```
tkerola: `the following three`

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,467 @@
...
+        - Variance Loss:
+            Loss to penalize distances between pixels which are belonging
+            to same instance. (Pull force)
+        - Distance loss:
+            Loss to penalize distances between the centers of instances.
+            (Push force)
+        - Regularization loss:
+            Small regularization loss to penalize weights against overfit
```
tkerola: `overfitting.` (with a dot at the end)

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,467 @@
...
+        norm (int): Norm to calculate pixels and cluster center distances
+        alpha (float): Weight for variance loss      (alpha * variance_loss)
+        beta (float): Weight for distance loss       (beta * distance_loss)
+        gamma (float): Weight for regularization loss(gamma * regularizer_loss)
+
+    Returns:
+        (:class:`~chainer.Variable` or \
+        :class:`numpy.ndarray` or \
+        :class:`cupy.ndarray`):
+            (alpha * variance_loss)+
```
tkerola: Please add a space after the parenthesis.

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,467 @@
...
+        l_shape = labels.shape
+
+        prediction = reshape(prediction, shape=(
+            p_shape[0], p_shape[1], p_shape[2] * p_shape[3]))
+        labels = reshape(labels, shape=(
+            l_shape[0], l_shape[1], l_shape[2] * l_shape[3]))
+
+        return prediction, labels
+
+    def __call__(self, args):
```
tkerola: How about explicitly accepting 4 input instead of a tuple? The function would become more self-documenting and the function signature would become easier to understand.

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,467 @@
...
+from chainer.functions.math.basic_math import absolute
+from chainer.functions.math.maximum import maximum
+from chainer.functions.math.sum import sum as c_sum
+import warnings
+
+
+class DiscriminativeMarginBasedClusteringLoss(object):
+    """Discriminative margin based clustering loss function
+
+    This is the implementation of following paper:
```
tkerola: `the following paper`

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,467 @@
...
+from chainer.functions.array.transpose import transpose
+from chainer.functions.math.average import average
+from chainer.functions.math.basic_math import absolute
+from chainer.functions.math.maximum import maximum
+from chainer.functions.math.sum import sum as c_sum
+import warnings
+
+
+class DiscriminativeMarginBasedClusteringLoss(object):
+    """Discriminative margin based clustering loss function
```
tkerola: `margin-based`

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,467 @@
...
+        # L1 or L2 norm is allowed only
+        if norm == 1:
+            self.norm = self._l1_norm
+        elif norm == 2:
+            self.norm = self._l2_norm
+        else:
+            raise ValueError("Norm can only be 1 or 2")
+
+    def _l1_norm(self, x, axis):
+        """Function to calculate L1 Norm by given axes
```
tkerola: I think this docstring can be removed, since the function name is self-documenting.

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,467 @@
...
+
+        Returns:
+            :class:`~chainer.Variable` or \
+            :class:`numpy.ndarray` or \
+            :class:`cupy.ndarray` : Norm applied to given axes
+        """
+        return c_sum(absolute(x), axis=axis)
+
+    def _l2_norm(self, x, axis):
+        """Function to calculate L2 Norm by given axes
```
tkerola: ditto

chainer/functions/loss/discriminative_loss.py
tkerola: Also, please link to the arxiv landing page instead of the PDF.

chainer/functions/loss/discriminative_loss.py
tkerola: It would also make the Chainer API more consistent, since then this loss function would look more like `F.softmax_cross_entropy`, which is more intuitive.

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,467 @@
...
+        m = expand_dims(means, axis=3)
+
+        p = broadcast_to(p, (bs, n_instances, n_filters, n_loc))
+        g = broadcast_to(g, (bs, n_instances, n_filters, n_loc))
+        m = broadcast_to(m, (bs, n_instances, n_filters, n_loc))
+
+        m = cast(m, p.dtype)
+        g = cast(g, p.dtype)
+
+        module = cuda.get_array_module(p)
```
tkerola: Please name `module` as `xp` instead, since this is the convention in other Chainer code, and makes the function easier to read.

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,467 @@
...
+        Returns:
+            :class:`~chainer.Variable` or \
+            :class:`numpy.ndarray` or \
+            :class:`cupy.ndarray` : variance loss
+        """
+
+        bs, n_filters, n_loc = pred.shape
+        n_instances = gt.shape[1]
+
+        # Prepare each item with same size by broadcasting missing axes
```
tkerola: This comment is redundant and can be removed.

dBeker:
@tkerola I have pushed the changes that you suggest.

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,434 @@
...
+    In segmentation, one of the biggest problem is to have noise at the output
+    of a trained network.
+    For cross-entropy based approaches, if the pixel value is wrong,
+    the loss value will be same independent from the wrong pixel's location.
+    Even though the network gives wrong pixel output, it is desirable
+    to have it as close as possible to the original position.
+    By applying discriminative loss function, groups of segmentation instances
+    can be moved together.
+    This loss function calculates the following three parameters:
+        - Variance Loss:
```
beam2d: Insert a blank line before listing and remove the indentation of each item (these indentations cause doc build error). It should look like:

```
    This loss function calculates the following three parameters:

    - Variance loss:
      Loss to penalize...
    - Distance loss:
      Loss to penalize...
    - Regularization loss:
      Small regularization...

    Args:
        ...
```

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,434 @@
...
```
beam2d: ditto

dBeker:
@delta2323 @tkerola All the requested changes are done and html build succeeded. Pending for your final approval to be merged

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,437 @@
...
+        Small regularization loss to penalize weights against overfitting.
+
+    Args:
+        delta_v (float): Minimum distance to start penalizing variance
+        delta_d (float): Maximum distance to stop penalizing distance
+        max_n_clusters (int): Maximum possible number of clusters.
+        norm (int): Norm to calculate pixels and cluster center distances
+        alpha (float): Weight for variance loss      (alpha * variance_loss)
+        beta (float): Weight for distance loss       (beta * distance_loss)
+        gamma (float): Weight for regularization loss(gamma * regularizer_loss)
```
tkerola: Please add a space before the parenthesis.

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,437 @@
...
+    This is the implementation of the following paper:
+    https://arxiv.org/abs/1802.05591
+
+    In segmentation, one of the biggest problem is to have noise at the output
+    of a trained network.
+    For cross-entropy based approaches, if the pixel value is wrong,
+    the loss value will be same independent from the wrong pixel's location.
+    Even though the network gives wrong pixel output, it is desirable
+    to have it as close as possible to the original position.
+    By applying discriminative loss function, groups of segmentation instances
```
tkerola: `a discriminative`

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,437 @@
...
+    can be moved together.
+    This loss function calculates the following three parameters:
+
+    - Variance Loss:
+        Loss to penalize distances between pixels which are belonging
+        to same instance. (Pull force)
```
tkerola: `the same instance`

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,437 @@
...
+            n_objects (nd-array): Instance count in current input
+
+        Returns:
+            float : distance loss
+        """
+
+        bs, n_instances, n_filters = means.shape
+        m = cast(means, means.dtype)
+
+        module = cuda.get_array_module(means)
```
tkerola: Please use `xp` instead of `module` for consistency.

chainer/functions/loss/discriminative_loss.py
tkerola: Please address the above comment.

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,437 @@
...
+
+    def _regularization_term(self, means, gt_idx):
+        """Function to calculate regularization term
+
+        Args:
+            means (nd-array): Instance means
+            gt_idx (tuple / nd-array): Indexes of ground truth instances
+
+        Returns:
+            float : distance loss
```
tkerola: Please indicate that the return value is a `chainer.Variable`, not a float.

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,437 @@
...
+            n_objects (int): Instance number in current input
+            max_n_objects (int): Maximum possible instance number
+            gt_idx (tuple / nd-array):
+                Indexes of ground truth instances
+
+        Returns:
+            tuple :
+            (:class:`~chainer.Variable` or \
+            :class:`numpy.ndarray` or \
+            :class:`cupy.ndarray`)  distance loss
```
tkerola: It seems like the return value is the mean, not the distance loss? Does the documentation need to be corrected?

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,437 @@
...
+        """Function to calculate cluster means
+
+        Args:
+            pred (:class:`~chainer.Variable` or \
+                  :class:`numpy.ndarray` or \
+                  :class:`cupy.ndarray`) : Prediction output
+            gt (:class:`~chainer.Variable` or \
+                :class:`numpy.ndarray` or \
+                :class:`cupy.ndarray`) : Ground truth output
+            n_objects (int): Instance number in current input
```
tkerola: It seems like `n_objects` is an array, not an `int`. Please update the documentation to correctly reflect the current dtypes.

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,437 @@
...
+            max_n_objects (int): Maximum possible instance number
+            gt_idx (tuple / nd-array):
```
tkerola: Please make sure that the documentation is consistent. You use `or` previously, but here you use `/`.

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,437 @@
...
+            p_shape[0], p_shape[1], p_shape[2] * p_shape[3]))
+        labels = reshape(labels, shape=(
+            l_shape[0], l_shape[1], l_shape[2] * l_shape[3]))
+
+        return prediction, labels
+
+    def __call__(self, prediction, labels, n_objects, gt_idx):
+        """Applies discriminative margin based clustering loss
+
+        Steps are:
```
tkerola: `The steps are`

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,437 @@
...
+        gamma (float): Weight for regularization loss(gamma * regularizer_loss)
+
+    Returns:
+        (:class:`~chainer.Variable` or \
+        :class:`numpy.ndarray` or \
+        :class:`cupy.ndarray`):
+            (alpha * variance_loss) +
+            (beta * distance_loss) +
+            (gamma * regularizer_loss)
+    """
```
tkerola: Please update this docstring to be identical to the one used for `DiscriminativeMarginBasedClusteringLoss`, since they should be the same.

dBeker: I am quite confused with this one. The wrapper contains initialization + result parameter which is obtained by calling __call__ function of `DiscriminativeMarginBasedClusteringLoss` class. However, class documentation only contains the initialization information. Hence, I am planning to remove output value from `DiscriminativeMarginBasedClusteringLoss` class. What do you think?

tkerola: Oh, sorry, this is my mistake. You're right that the docstrings should not be identical. What I meant was that both docstrings contain some identical paragraphs, so I meant that the typos that were fixed in the other docstring should also be fixed in this one.

chainer/functions/loss/discriminative_loss.py
delta2323: @dBeker Could you reflect @tkerola 's comment?

dBeker:
@delta2323 Now fixed, my bad missed that one.

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,442 @@
...
+            (:class:`~chainer.Variable` or \
+            :class:`numpy.ndarray` or \
+            :class:`cupy.ndarray`):
+                (alpha * variance_loss) +
+                (beta * distance_loss) +
+                (gamma * regularizer_loss)
+        """
+
+        buffer = chainer.config.type_check
+        chainer.config.type_check = False
```
delta2323: Sorry, I missed this code. Can't we change code while preventing the loss from disabling type checking?

dBeker:
@delta2323 I fixed the type_check parameter

dBeker:
@delta2323 I would be glad if you can check the last commits and merge if everything is ok

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,439 @@
...
+
+import warnings
+
+
+class DiscriminativeMarginBasedClusteringLoss(object):
+
+    """Discriminative margin-based clustering loss function
+
+    This is the implementation of the following paper:
+    https://arxiv.org/abs/1802.05591
```
delta2323: Is the citation correct? The paper in #4513 is https://arxiv.org/pdf/1708.02551.pdf

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,439 @@
...
+
+
+def discriminative_margin_based_clustering_loss(
+        prediction, labels, n_objects, gt_idx,
+        delta_v, delta_d, max_n_clusters,
+        norm=1, alpha=1.0, beta=1.0, gamma=0.001):
+    """Discriminative margin-based clustering loss function
+
+    This is the implementation of the following paper:
+    https://arxiv.org/abs/1802.05591
```
delta2323: ditto

chainer/functions/loss/discriminative_loss.py
dBeker: Fixed.

chainer/functions/loss/discriminative_loss.py
dBeker: Fixed

dBeker:
@delta2323 This branch is ready to be merged, FYI

kmaehashi:
@delta2323 ping?

delta2323:
Sorry for late response. I'll review the PR.

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,219 @@
...
+        l_dist = 0.0
+        l_reg = 0.0
+        count = 0
+        means = []
+        xp = cuda.get_array_module(embeddings)
+
+        zeros = xp.asarray(0.0)
+
+        # Calculate cluster means
+        for c in range(self.max_embedding_dim):
```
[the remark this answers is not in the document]
delta2323: With height and width.

```
x = np.arange(48).reshape(3, 4, 2, 2)
t = np.random.randint(0, 5, (3, 2, 2))

xx = np.repeat(x[None], 5, 0)
mask = np.repeat(np.arange(5)[..., None, None], x.shape[1], 2)[..., None, None] == t
means = (xx * mask).sum(axis=(1, 3, 4))
```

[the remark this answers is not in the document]
delta2323: We can parallelize computation without explicit for-loop. But at the cost of it, the memory consumption will be multiplied by the number of cluster C (for mean, variance, and regularization) or by C^2 (distance). I'm wondering if there is more memory-efficient way.

chainer/functions/loss/discriminative_loss.py
```diff
@@ -401,39 +185,35 @@ def discriminative_margin_based_clustering_loss(
...
-        (:class:`~chainer.Variable` or \
-        :class:`numpy.ndarray` or \
-        :class:`cupy.ndarray`):
-            (alpha * variance_loss) +
-            (beta * distance_loss) +
-            (gamma * regularizer_loss)
+        Variance Loss(:class:`~chainer.Variable` or \
+                      :class:`numpy.ndarray` or \
+                      :class:`cupy.ndarray`):
+                      variance loss multiplied by alpha
```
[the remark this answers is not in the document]
dBeker: That's true. Actually, all the output parameters can get two different values: 
- Chainer Variable
- Float

Float occurs when GT does not contain any annotation (all zero). Even though it is not an expected behavior, this code can result float output. I will change accordingly

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,219 @@
...
+        count = 0
+        means = []
+        xp = cuda.get_array_module(embeddings)
+
+        zeros = xp.asarray(0.0)
+
+        # Calculate cluster means
+        for c in range(self.max_embedding_dim):
+            # Create mask for instance
+            mask = xp.expand_dims(labels == c + 1, 1)
```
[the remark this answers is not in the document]
dBeker: That was requested by @t-abe . Actually, -even though very unlikely - there may be more cluster than the embedding dimensions.

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,219 @@
...
+        norm (int): Norm to calculate pixels and cluster center distances
+        alpha (float): Weight for variance loss    (alpha * variance_loss)
+        beta (float): Weight for distance loss     (beta * distance_loss)
+        gamma (float): Weight for regularizer loss (gamma * regularizer_loss)
+
+    Returns:
+        Variance Loss(:class:`~chainer.Variable` or \
+        :class:`numpy.ndarray` or \
+        :class:`cupy.ndarray`):
+            variance loss multiplied by alpha
```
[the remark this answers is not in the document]
dBeker: Ditto

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,195 @@
...
+                                 emb.shape[4]))
+        ms = []
+        for c in range(self.max_embedding_dim):
+            # Create mask for instance
+            mask = xp.expand_dims(labels == c + 1, 1)
+            ms.append(mask)
+        ms = xp.stack(ms, 0)
+        mns = c_sum(emb * ms, axis=(3, 4))
+        mns = mns / xp.maximum(xp.sum(ms, (2, 3, 4))[:, :, None], 1)
+        mns_exp = mns[:, :, :, None, None]
```
delta2323: Just for confirmation. Cluster means are functions of `embeddings`, but we don't have to propagate error via means, do we?

chainer/functions/loss/discriminative_loss.py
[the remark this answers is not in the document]
delta2323: Could you let me confirm my understanding? If I understand correctly, `max_embedding_dim` works as C in [the original paper](https://arxiv.org/pdf/1708.02551.pdf). But from the name of the variable, it should be dim(x_i) <= `max_embedding_dim` should be held where x_i is the variable used in e.g. eq. (1) of the paper. So, do you mean dim(x_i) <= C holds typically?

dBeker:
@delta2323 Sorry for the inconvenience, I had some wrong commits due to unsync fork. 
Please follow up with new and clean PR at https://github.com/chainer/chainer/pull/5313

PR #4528, opened 2018-03-26
Fix temporary file permission issue in LogReport

niboshi:
Fixed the same issue as #4461.
This kind of use of temporary directory is very common, so I made a context manager as `utils.tempdir()`, and applied to all such instances.

### DIFF
--- a/chainer/utils/__init__.py
+++ b/chainer/utils/__init__.py
@@ -1,3 +1,7 @@
+import contextlib
+import shutil
+import tempfile
+
 import numpy
 
 from chainer.utils import walker_alias  # NOQA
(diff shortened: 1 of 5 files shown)


okuta:
LGTM!

chainer/utils/__init__.py
```diff
@@ -34,3 +38,13 @@ def force_type(dtype, value):
...
+
+
+@contextlib.contextmanager
+def tempdir(**kwargs):
+    # A context manager that defines a lifetime of a temporary directory.
+    temp_dir = tempfile.mkdtemp(**kwargs)
+    try:
+        yield temp_dir
+    finally:
+        shutil.rmtree(temp_dir, ignore_errors=True)
```
kmaehashi: Is it safe to ignore errors? (c.f. `tempfile.TemporaryDirectory` raise error when failure of cleanup: https://github.com/python/cpython/blob/3.6/Lib/tempfile.py#L777-L811)

niboshi: I agree.
I made it optional and `False` by default.

kmaehashi:
LGTM!

@okuta Test passed. Could you merge?

PR #4532, opened 2018-03-27
use --no-cache-dir in Dockerfile

kmaehashi:
Chainer part of the same fix as in https://github.com/cupy/cupy/pull/1060.


niboshi:
LGTM!

PR #4534, opened 2018-03-27
[WIP] Fix `load_npz` to refuse loading pickled objects by default

### LINKED ISSUE #4530: Safer loading
Hi!
I was just studying chainer's serialization/deserialization. 

Very good job, but to make numpy loading safe one minor change needed:
```python
from chainer.serializers import NpzDeserializer
def load_npz(file, obj, path='', strict=True):
    with numpy.load(file, allow_pickle=False) as f:
        d = NpzDeserializer(f, path=path, strict=strict)
        d.load(obj)
```
Please notice `allow_pickle=False`. By default, `numpy.load` allows using pickle, which can execute any code (pickle is very unsafe, that's why can't be used to share models, you probably know it).

I've checked that in my case this fix doesn't break already saved models, so probably it is a good solution.

kmaehashi:
This fixes #4530.


chainer/serializers/npz.py
```diff
@@ -154,8 +155,10 @@ def __call__(self, key, value):
         return value
 
 
-def load_npz(file, obj, path='', strict=True):
-    """Loads an object from the file in NPZ format.
+def load_npz(file, obj, path='', strict=True, **kwargs):
+    """load_npz(file, obj, path='', strict=True, allow_pickle=False)
```
niboshi: Put `*, ` before `allow_picke`.

tests/chainer_tests/serializers_tests/test_npz.py
```diff
@@ -445,6 +445,30 @@ def test_load_without_path(self):
...
+
+        # Should success if `allow_pickle=True`.
+        file.seek(0)
+        l.obj[0] = 'overwritten'
+        npz.load_npz(file, l, allow_pickle=True)
+        assert l.obj[0] == 'data'
+
+        # Should raise an error if `allow_pickle=False` (default).
+        file.seek(0)
+        self.assertRaises(
```
niboshi: How about `pytest.raises`?

chainer/serializers/npz.py
kmaehashi: Fixed.

tests/chainer_tests/serializers_tests/test_npz.py
kmaehashi: I fixed to use `with self.assertRaises()`.
Are there merits of `pytest.raises` over `self.assertRaises`?

niboshi: It's about consistency, as it's pytest's standard way and we're moving towards pytest.

kmaehashi: I thought mixing two kinds of notation in the same codebase is rather inconsistent and confusing for users/contributors, considering this is an open source project.

I'm also worrying that by strongly depending on one test framework makes future migration a tough work (considering migration from nose to pytest was pretty easy because our test code does not strongly depend on nose).

niboshi: I understand the points, and actually I'm not confident about that.
Maybe we should make a clear policy.

There are some conterarguments about those points:
* It's inevitable to mix different conventions if we're moving one convention to another.
* We can say using `assert` also depends on pytest.

niboshi: As for `pytest.raises` specifically, there's one advantage over `self.assertRaises`: no need of `self`. If the checking logic is factored out from the `TestCase` class (there may be many reasons to do so), we need to pass around `self` instance, if we stick with the old convention.

kmaehashi: I come up with the simple policy:

* Keep avoiding `import pytest` in our test code base (as we did in nose)
* Permit use of simple `assert`

Thoughts behind the policy above is that I want to minimize the requirements of test framework when we have to choose another framework to migrate.

* Why avoid `import pytest`?
    * Once we allow `import pytest`, we allow every feature in pytest to be used in future test code.
    * This rule is simple enough for contributors.
* Why simple `assert` can be a exception:
    * It is shorter than `self.assertEqual` and also improves readability. (= cost-effective)
    * Even with simple assert, we can say that the test code is still compatible with standard `unittest` (although without nice dump of arguments)

c.f., In NumPy, it seems that they are paying much effort to abstracting out test framework by their own and gradually moving from nose to pytest.

* https://github.com/numpy/numpy/issues/9352#issue-240027620
* https://github.com/numpy/numpy/tree/master/numpy/testing has pytest_tools and nose_tools, and `assert_raises` is defined for each framework.
  * https://github.com/numpy/numpy/blob/master/numpy/testing/pytest_tools/utils.py#L1190
  * https://github.com/numpy/numpy/blob/master/numpy/testing/nose_tools/utils.py#L1208

niboshi: I made an issue for this topic: #4557.
Let's keep `self.assertRaises` in this PR.

kmaehashi: Thanks!

niboshi:
Hmm, it seems there's no `numpy.load(allow_picke=)` argument in NumPy 1.9.

kmaehashi:
Ok, I'll fix the code so that it will use `allow_pickle` only after NumPy 1.10.
(Maybe we need to think about deprecation of old NumPy also.)

kmaehashi:
We discussed some ideas:

1. Drop NumPy 1.9 support in v5 and do not backport this.
    * Pros: No dirty workarounds.  Also reduces cost in many other ways (e.g. testing).
    * Cons: NumPy 1.9 users are affected.
2. Use `allow_pickle=None` as default, which defaults to True (more specifically, not specifying allow_pickle) if NumPy 1.9 and False if NumPy 1.10 or later.
   If `allow_pickle=False` is explicitly specified for NumPy 1.9, raise an error.
    * Pros: Can support NumPy 1.9.
    * Cons: Models saved by NumPy 1.10 may not be loaded by NumPy 1.9.
3. Make the default behavior controllable via environment variable.
    * Pros: Can support NumPy 1.9 and still compatible between different NumPy versions.
    * Cons: Another configuration item is added.

kmaehashi:
If we go with idea 1 or 2, NumPy 1.10+ users may not be able to customize pre-defined models defined in Chainer (e.g., calling `add_persistent` on [chainer.links.model.vision.vgg](https://github.com/chainer/chainer/blob/v4.0.0rc1/chainer/links/model/vision/vgg.py)). However, users can workaround that issue by specifying `pretrained_model=None` to constructor of such link and manually calling `npz.load_npz(allow_pickle=True)` later. So this change is `no-compat` (and cannot be backported to v4) but I think is still acceptable.

I'd like to proceed with idea 2. Do you think it is OK? @niboshi

niboshi:
I'm sorry for delay.
I don't understand the relation between `add_persistent` and `pretrained_model`.
Could you clarify that?

kmaehashi:
@niboshi `load_npz` will be called without `allow_pickle` from built-in Links under `chainer/links/model/vision`.
What we concerned at that time was that if pre-trained model file contains user-defined persistent object, the model cannot be specified to `allow_pickle`.
https://docs.chainer.org/en/stable/reference/generated/chainer.links.VGG16Layers.html?highlight=VGG16Layers

niboshi:
Is this PR still WIP?

PR #4551, opened 2018-03-30
Improve chainer.utils.argument exception message

vilyaair:
I want to make exception message more similar to that in Python.

### DIFF
--- a/chainer/utils/argument.py
+++ b/chainer/utils/argument.py
@@ -1,3 +1,6 @@
+import inspect
+
+
 def check_unexpected_kwargs(kwargs, **unexpected):
     for key, message in unexpected.items():
         if key in kwargs:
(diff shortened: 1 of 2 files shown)


niboshi:
Could you write a test to check for error message?
You can use `six.assertRaisesRegex`.

chainer/utils/argument.py
```diff
@@ -8,8 +8,12 @@ def parse_kwargs(kwargs, *name_and_values):
     values = [kwargs.pop(name, default_value)
               for name, default_value in name_and_values]
     if kwargs:
-        args = ', '.join(["'%s'" % arg for arg in kwargs.keys()])
-        raise TypeError('got unexpected keyword argument(s) %s' % args)
+        import inspect
```
niboshi: Please move `import` to the top of the file.

vilyaair: I consider that this exception is raised so rarely that importing `inspect` module at top of the file might incur additional performance overhead, because [`import inspect`](https://github.com/chainer/chainer/search?utf8=%E2%9C%93&q=%22import+inspect%22&type=) is nearly zero used in the chainer core codes, in contrast with [`parse_kwargs`](https://github.com/chainer/chainer/search?utf8=%E2%9C%93&q=parse_kwargs&type=) used more frequently.

niboshi:
jenkins, test this please

PR #4559, opened 2018-04-02
Implement interface to print runtime information

kmaehashi:
Implements interface to print runtime information. Took over #4298.
Requires https://github.com/cupy/cupy/pull/1067.

~~TODO: add test.~~ (done)

### DIFF
--- a/chainer/_runtime_info.py
+++ b/chainer/_runtime_info.py
@@ -1,0 +1,41 @@
+import sys
+
+import numpy
+import six
+
+import chainer
+from chainer.backends import cuda
+
+
+class _RuntimeInfo(object):
+
+    chainer_version = None
+    numpy_version = None
+    cupy_info = None
+
+    def __init__(self):
+        self.chainer_version = chainer.__version__
+        self.numpy_version = numpy.__version__
+        self.cupy_info = cuda.get_runtime_info()
+
+    def __str__(self):
+        s = six.StringIO()
+        s.write('''Chainer: {}\n'''.format(self.chainer_version))
+        s.write('''NumPy: {}\n'''.format(self.numpy_version))
+        s.write(str(self.cupy_info))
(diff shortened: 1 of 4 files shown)


kmaehashi:
Added tests.

### Output examples

When CuPy is unavailable:

```
>>> import chainer
Chainer: 4.0.0rc1
NumPy: 1.14.2
CuPy: Not Available
```

When CuPy is available:

```
>>> import chainer
Chainer: 4.0.0rc1
NumPy: 1.14.2
CuPy:
  CuPy Version          : 4.0.0rc1
  CUDA Build Version    : 9000
  CUDA Driver Version   : 9000
  CUDA Runtime Version  : 9000
  cuDNN Build Version   : 7100
  cuDNN Version         : 7100
  NCCL Build Version    : 2102
```

chainer/_runtime_info.py
```diff
@@ -0,0 +1,45 @@
...
+
+def get_runtime_info():
+    return _RuntimeInfo()
+
+
+def print_runtime_info(out=None):
+    if out is None:
+        out = sys.stdout
+    out.write(str(get_runtime_info()))
+    out.flush()
```
niboshi: `flush` method may not exist.
It's better to either avoid it or test its existence first.

niboshi: There's no official specification of file-like objects.
You can define an object with only `write` method and call it a file-like.

kmaehashi: OK, I fixed to first check existence of the flush method.
(It seems other features (`CupyMemoryProfileHook`, `ProgressBar`, etc...) also assumes `flush`.)

niboshi: I see. It's better to fix them as well.

niboshi:
LGTM!

Issue #4546, opened 2018-03-30
Add extra support for different learning rate policy.


JirenJin:
To reproduce results of other papers, usually we would like to follow exactly the same implementation details reported in their paper (or code, if available). However, the support for the learning rate policy in Chainer is not enough, compared with other frameworks, e.g., Caffe has many learning rate policies:
```
//    - fixed: always return base_lr.  
//    - step: return base_lr * gamma ^ (floor(iter / step))  
//    - exp: return base_lr * gamma ^ iter  
//    - inv: return base_lr * (1 + gamma * iter) ^ (- power)  
//    - multistep: similar to step but it allows non uniform steps defined by  
//      stepvalue  
//    - poly: the effective learning rate follows a polynomial decay, to be  
//      zero by the max_iter. return base_lr (1 - iter/max_iter) ^ (power)  
//    - sigmoid: the effective learning rate follows a sigmod decay  
//      return base_lr ( 1/(1 + exp(-gamma * (iter - stepsize))))  
```

Since Chainer now only supports `ExponentialShift` and `LinearShift`, it is not so convenient to tune the learning rate policy (without writing custom `Trainer` hooks or updaters). IMO, providing common learning rate policy is deserved to be included in the official implementation. 

I would like to submit PRs for implementing some of the above policies if possible.

kmaehashi:
Thank you for the feature request!
We welcome PRs to extend learning rate policies.

kmaehashi:
Refs #4667

kmaehashi:
Refs #4894

kmaehashi:
- [X] fixed
- [ ] step (#4894)
- [X] exp
- [X] inv (#4565)
- [ ] multistep
- [X] poly (#4693)
- [ ] sigmoid

niboshi:
Now that we have several `*Shift` extensions, I think it's a good time to refactor them.

PR #4565, opened 2018-04-03
Add `InverseShift` extension

jinjiren:
This PR aims to add support for InverseShift extension, which is compatible to the `inv` learning rate policy in Caffe. 

As mentioned in #4546, having support for various learning rate schedule policies will make reproducing results of existing papers easier.

I'm planning to add more `Shift` extensions, however, before that I would like to see whether this one is proper for merge.


bkvogel:
Thank you for the PR! Please fix flake8 errors.

jinjiren:
@bkvogel I fixed the flake8 errors (checked locally). Sorry that I forgot to use the `flake8` installed by `hacking` package, instead I had used the common `flake8` (which gives no errors to the previous code...).

bkvogel:
I think it looks good. My only comment is that there are not any tests. Although it is not required, if you can think of any ways to test the expected behavior, please consider adding tests. For an example, see the existing extensions tests such as https://github.com/chainer/chainer/blob/master/tests/chainer_tests/training_tests/extensions_tests/test_exponential_shift.py

kmaehashi:
👍Please add it to the docs also.
https://github.com/chainer/chainer/blob/master/docs/source/reference/training.rst#optimizer-behavior-control

bkvogel:
@jinjiren could you also please fix the conflict in `__init__.py`? thanks!

jinjiren:
Added unittests and docs.

bkvogel:
Jenkins, test this please.

bkvogel:
Jenkins, test this please.

bkvogel:
LGTM. The tests passed except one that looks unrelated to your PR.

PR #4566, opened 2018-04-03
[RFC] Add error handler interface to trainer extensions

kuenishi:
This is rather a request-for-comment to my suggestion to add error handling interface to trainer.

More than 90% use cases of trainer extensions just `finalize()` would be sufficient for cleaning up extensions. But for some use cases like snapshots, or [checkpoints](http://chainermn.readthedocs.io/en/stable/reference/#chainermn.create_multi_node_checkpointer) it would be nice if the trainer extensions have error handling interface to preserve the *latest* trainer image, or models. See changes in this PR on snapshot extension, which has additional `snapshot_on_error` flag for actual implementation.

Moreover, my fault tolerance prototyping work that is going underwater requires this feature to handle GPU failures by leaving the computing cluster in a graceful manner rather than just the process quitting with some signals.

Note: This pull request does not have enough documentation, nor any test codes. Not worth testing or merging for now. If the discussion goes well I'd like to put my time more.

### DIFF
--- a/chainer/training/extension.py
+++ b/chainer/training/extension.py
@@ -98,6 +98,20 @@ class Extension(object):
         """
         pass
 
+    def on_error(self, exp, exc_info):
+        """Handle the error during training before finalize.
+
+        This method is called when an exception is thrown during the
+        training loop. An extension that needs different error handling from finalize,
+        can override this method to handle errors.
+
+        Args:
+             exp (Exception): arbitrary exception thrown during update loop.
+             exc_info: exception info e.g. stack trace
+
+        """
+        pass
+
     def serialize(self, serializer):
         """Serializes the extension state.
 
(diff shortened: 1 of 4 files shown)


kmaehashi:
I was just thinking of the similar thing while reviewing #4545 today :)
There may be another use case for this method: ChainerUI currently just become silent when the training script dies, but it's nice if friendly message and stacktrace can be displayed on the Web UI.

As for the falut-tolerance viewpoint, I'm not sure if exceptions are thrown as we expect in such case.

nitpick: maybe we want to ignore exceptions raised during processing on_error handler?

kuenishi:
Thanks for positive feedback! I'll start working on this.

> As for the falut-tolerance viewpoint, I'm not sure if exceptions are really thrown as we expect in such case.
That depends, but I hear most cases of GPU errors are returned as  `CUDA_UNKNOWN_ERROR` or such exception; a few may lead to just hanging. I want to start from majority cases, followed by corners.

> nitpick: maybe we want to ignore exceptions raised during processing on_error handler?
Good point. All exceptions thrown by error handlers must be ignored.

kuenishi:
Opened #4630 as a complete work aiming for merge.

PR #4577, opened 2018-04-06
Fix to_intel64 to check if it is suitable for iDeep

### LINKED ISSUE #4572: Creating ``ideep.array`` without checking ndim and dtype in ``chianer.Variable.to_intel64``.
When running the below command, it outputs error.
It seems that creating ``ideep.array`` without checking ndim and dtype in ``chianer.Variable.to_intel64``.

```
$ python snipet.py 
terminate called after throwing an instance of 'mkldnn::error'
  what():  std::exception
Aborted (core dumped)
```

snipet.py
```
import os                                                                         
os.environ["CHAINER_USE_IDEEP"] = "auto"

import numpy as np
import chainer
from chainer import cuda, Function, gradient_check, report, training, utils, Variable
from chainer import datasets, iterators, optimizers, serializers
from chainer import Link, Chain, ChainList
import chainer.functions as F
import chainer.links as L
from chainer.datasets import mnist

class Conv1D(chainer.Chain):
    def __init__(self, in_channels, out_channels, ksize=3, stride=1, pad=1):
        super(Conv1D, self).__init__()
        with self.init_scope():
            self.conv=L.ConvolutionND(1, in_channels, out_channels, ksize, stride=stride, pad=pad)

    def __call__(self, x):
        return F.relu(self.conv(x))

model = Conv1D(3, 32)

gpu_id = -1  # Set to -1 if you use CPU
if gpu_id >= 0:
    model.to_gpu(gpu_id)
if os.environ["CHAINER_USE_IDEEP"] == "auto":
    model.to_intel64()

print('reached end')
```

kmaehashi:
Fixes #4572.

### DIFF
--- a/chainer/variable.py
+++ b/chainer/variable.py
@@ -776,16 +776,16 @@ Actual: {0}'''.format(type(data))
         intel64.check_ideep_available()
         data = self.data
         if data is not None:
-            if isinstance(data, numpy.ndarray):
-                # numpy.ndarray to ideep
-                self._data = [
-                    intel64.ideep.array(
-                        data, itype=intel64.ideep.wgt_array)]
-            elif isinstance(data, cuda.ndarray):
-                # cupy.ndarray to ideep
-                self._data = [
-                    intel64.ideep.array(
-                        data.get(), itype=intel64.ideep.wgt_array)]
+            if isinstance(data, cuda.ndarray):
+                # cupy.ndarray to numpy.ndarray
+                data = data.get()
+            if (isinstance(data, numpy.ndarray) and
+                    intel64.inputs_all_ready((data,))):
+                # numpy.ndarray to ideep.mdarray
+                data = intel64.ideep.array(
+                    data, itype=intel64.ideep.wgt_array)
+            self._data = [data]
+
         if self._grad_var is not None:
             self._grad_var.to_intel64()
             # ensure that the node tracks the device migration
--- a/tests/chainer_tests/test_link.py
+++ b/tests/chainer_tests/test_link.py
@@ -1284,5 +1284,14 @@ class TestIntel64(unittest.TestCase):
         # Persistent scalar
         assert link.ps == self.ps_scalar
 
+    @attr.ideep
+    def test_intel64_unsupported(self):
+        # Test for persistents that cannot be transferred to iDeep.
+        with self.link.init_scope():
+            self.link.no_ideep = numpy.ones((2, 2, 2), numpy.float32)
+            self.link.register_persistent('no_ideep')
+        self.link.to_intel64()
+        assert isinstance(self.link.no_ideep, numpy.ndarray)
+
 
 testing.run_module(__name__, __file__)
--- a/tests/chainer_tests/test_link.py
+++ b/tests/chainer_tests/test_link.py
@@ -1284,11 +1284,19 @@ class TestIntel64(unittest.TestCase):
         # Persistent scalar
         assert link.ps == self.ps_scalar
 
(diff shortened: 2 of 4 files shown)


kmaehashi:
When calling `Variable.to_intel64()` and `Link.to_intel64()`, we should validate that the array being transferred is supported by iDeep.
Issue can be reproduced with code in #4572 (for Variable) and the following code (for Link).

```py
import chainer
import numpy as np
l = chainer.Link()
l.add_persistent('x', np.zeros((3,3,3), dtype=np.float32))  # not supported by ideep
l.to_intel64()  # abort (core dumped)
```

chainer/variable.py
```diff
@@ -776,16 +776,16 @@ def to_intel64(self):
...
-                        data.get(), itype=intel64.ideep.wgt_array)]
+            if isinstance(data, cuda.ndarray):
+                # cupy.ndarray to numpy.ndarray
+                data = data.get()
+            if (isinstance(data, numpy.ndarray) and
+                    intel64.inputs_all_ready((data,))):
+                # numpy.ndarray to ideep.mdarray
+                data = intel64.ideep.array(
+                    data, itype=intel64.ideep.wgt_array)
+            self._data = [data]
```
niboshi: Hmm, I think it should be `self.data = data`, because `VariableNode._data` should be updated as well. Maybe another issue, though.

tests/chainer_tests/test_link.py
```diff
@@ -1284,5 +1284,14 @@ def test_intel64_to_cpu(self):
         # Persistent scalar
         assert link.ps == self.ps_scalar
 
+    @attr.ideep
```
niboshi: `@attr.ideep` is not needed (it's put on the class.)

tests/chainer_tests/test_link.py
```diff
@@ -1284,5 +1284,14 @@ def test_intel64_to_cpu(self):
...
+    def test_intel64_unsupported(self):
```
niboshi: How about adding a test from `cupy.ndarray` to ideep as well?

tests/chainer_tests/test_variable.py
```diff
@@ -1773,6 +1772,28 @@ def test_intel64_to_cpu(self):
...
+}))
+@attr.ideep
+class TestIntel64Unsupported(unittest.TestCase):
+
+    """Tests for arrays that should not be converted to iDeep array."""
+
+    def setUp(self):
+        self.x_data = np.random.uniform(-1, 1, self.shape).astype(self.dtype)
+
+    def test_to_intel64(self):
```
niboshi: How about adding a test from `cupy.ndarray` to ideep as well?

chainer/variable.py
kmaehashi: I guess the indentation level of this block is wrong...?

* https://github.com/chainer/chainer/blob/v4.0.0rc1/chainer/variable.py#L762-L765
* https://github.com/intel/chainer/blob/8fe5698d6e8e34675fcbf75ce721107d7c27c706/chainer/variable.py#L737-L740

I'll fix this in separate PR.

kmaehashi:
Thanks for the review! Fixed accordingly.

chainer/variable.py
niboshi: Oh, I see...

niboshi:
Please fix flake8 error.

kmaehashi:
Fixed.

niboshi:
(memo: travis failure is not relevant)

niboshi:
LGTM!

PR #4578, opened 2018-04-06
fix document in EarlyStoppingTrigger

mori97:
Fixed a mistake in the docs of EarlyStoppingTrigger.
The type of max_trigger should be 'trigger object' but it was int.


kmaehashi:
LGTM!

PR #4587, opened 2018-04-09
Support None for the in_channels argument in ConvolutionND/DeconvolutionND

### LINKED ISSUE #2937: ConvolutionND can't  in_channels = None
`L.Convolution2D` can `input_channels = None` or omit input_channels.
But, `L.ConvolutionND` can't. 
What about making this feature be able to use in `L.ConvolutionND`.

mitmul:
This PR resolves #2937


hvy:
@mitmul It hit me that `DeconvolutionND` should be addressed too but maybe you were on it already!

mitmul:
Right, thanks for noting. I'll work on that too.

mitmul:
Jenkins!! Test this please

hvy:
Do you think we should update the docs?

hvy:
Maybe this can be addressed in a separate PR but I think we move the hard coded test fixtures to parametrized values, and then add parameters for e.g. testing convolution ND with ndim >= 3, other kernel sizes, strides, in/out channels, etc.

tests/chainer_tests/links_tests/connection_tests/test_deconvolution_nd.py
```diff
@@ -20,12 +20,14 @@
     'dtype': [numpy.float32],
     'use_cudnn': ['always', 'auto', 'never'],
     'used_outsize': ['case1', 'case2', 'None'],
+    'in_channels_none': [True, False],
```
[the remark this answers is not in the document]
mitmul: I'll check the reason and will comment here.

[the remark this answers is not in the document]
mitmul: @hvy Hm, I couldn't remember why it should be as is. So I fixed it.

mitmul:
@hvy Could you take a look into it again? I just followed your suggestion.

hvy:
Thanks for the quick fix. I'm thinking if we shouldn't support omitting the `in_channels` as in `Convolution2D` altogether, not just setting it to `None`. I think this would require us to set `ksize=None` by default and then basically copy this code https://github.com/chainer/chainer/blob/master/chainer/links/connection/convolution_2d.py#L126-L127 . Do you see any concerns doing this? (I'm wondering why the default `ksize` value for ND was different from 2D).

mitmul:
@hvy Thank you for the suggestion. FOr Convolution/DeconvolutionND, it takes `ndim` argument as the first argument to the constructor, so that I think there's no way to omit only the second argument `in_channels`. We can give `None` to it, but can't completely omit it because it creates ambiguity, e.g., `(ndim, out_channels)` can't be discriminated from `(ndim, in_channels)` which should raise an error for insufficient arguments. So, I think it's safe not to provide the omitting option for `in_channel` argument in ND operators. Does it make sense?

mitmul:
Jenkins, test this please

hvy:
Shouldn't `(ndim, in_channels)` and `(ndim, out_channels)` both raise errors anyway, meaning we don't have to discriminate between those cases? 

We require at least three arguments, and if three are given, interpret them as `(ndim, out_channels, ksize)` and in case of four `(ndim, in_channels, out_channels, ksize)` . Please correct me if I'm wrong.

mitmul:
Ah, you are right. Sorry I was confused. Thanks for clarifying it!

mitmul:
@hvy Updated the code and add examples to the docstrings. Please check it again :)

chainer/links/connection/convolution_nd.py
```diff
@@ -57,11 +57,15 @@ class ConvolutionND(link.Link):
 
     """
 
-    def __init__(self, ndim, in_channels, out_channels, ksize, stride=1, pad=0,
-                 nobias=False, initialW=None, initial_bias=None,
+    def __init__(self, ndim, in_channels, out_channels, ksize=None, stride=1,
+                 pad=0, obias=False, initialW=None, initial_bias=None,
```
hvy: Ops, `obias` a careless?

mitmul: @hvy Thanks, I fixed this point in the latest commit already. Please check it again :)

mitmul:
There was a flake8 error, but fixed.

mitmul:
Jenkins, test this please.

chainer/links/connection/convolution_nd.py
```diff
@@ -82,7 +133,11 @@ def __init__(self, ndim, in_channels, out_channels, ksize, stride=1, pad=0,
                 initial_bias = initializers._get_initializer(initial_bias)
                 self.b = variable.Parameter(initial_bias, out_channels)
 
-    def __call__(self, x):
+    def _initialize_params(self, in_channels):
+        W_shape = (self.out_channels, in_channels) + self.ksize
```
hvy: Do you have to take `self.groups` into account, as it's done in `Convolution2D`?

chainer/links/connection/deconvolution_nd.py
```diff
@@ -79,7 +131,13 @@ def __init__(self, ndim, in_channels, out_channels, ksize, stride=1, pad=0,
                 initial_bias = initializers._get_initializer(initial_bias)
                 self.b = variable.Parameter(initial_bias, out_channels)
 
-    def __call__(self, x):
+    def _initialize_params(self, in_channels):
+        W_shape = (in_channels, self.out_channels) + self.ksize
```
hvy: Ditto regarding groups!

hvy:
Sorry for the delayed comments. Do we have to take the `groups` arguments into account when computing the size of the `W`? It seems we didn't before but was the intended behavior?

mitmul:
@hvy Sorry for my late reply. I fixed the initialization part to consider the `groups` arguments. Could you check the code again?

hvy:
Rather sorry for my delay. Thanks for the changes, I'll take a look!

hvy:
Sorry again, the `groups` argument was maybe not necessary in this PR. Or rather, now I'm feeling we're responsible for adding tests for that argument too. @mitmul what do you think?

mitmul:
Jenkins, test this please

chainer/links/connection/convolution_nd.py
```diff
@@ -55,24 +57,73 @@ class ConvolutionND(link.Link):
...
+
+    def __init__(self, ndim, in_channels, out_channels, ksize=None, stride=1,
+                 pad=0, nobias=False, initialW=None, initial_bias=None,
                  cover_all=False, dilate=1, groups=1):
         super(ConvolutionND, self).__init__()
 
-        ksize = conv_nd.as_tuple(ksize, ndim)
+        if ksize is None:
+            ndim, out_channels, ksize, in_channels = \
+                ndim, in_channels, out_channels, None
```
hvy: Is the `ndim` reassignment redundant?

chainer/links/connection/deconvolution_nd.py
```diff
@@ -53,14 +55,62 @@ class DeconvolutionND(link.Link):
...
+
+    def __init__(self, ndim, in_channels, out_channels, ksize=None, stride=1,
+                 pad=0, nobias=False, outsize=None, initialW=None,
+                 initial_bias=None, dilate=1, groups=1):
         super(DeconvolutionND, self).__init__()
 
-        ksize = conv_nd.as_tuple(ksize, ndim)
+        if ksize is None:
+            ndim, out_channels, ksize, in_channels = \
+                ndim, in_channels, out_channels, None
```
hvy: Ditto regarding reassignment.

hvy:
LGTM except minor comments.

chainer/links/connection/convolution_nd.py
mitmul: Ooops, right

mitmul:
@hvy Thanks, I fixed the points.
Jenkins, test this please

mitmul:
jenkins test this please

hvy:
LGTM!

PR #4592, opened 2018-04-09
Fix to_intel64 not updating VariableNode.data

kmaehashi:
Fixes the issue found in https://github.com/chainer/chainer/pull/4577#discussion_r180058824


niboshi:
LGTM!

PR #4599, opened 2018-04-10
Improve docstrings about axis of Caffe functions

toslunar:
I found a user, who does not care about Caffe, was confused in using `L.Bias`. I'd like to tell in the document that the meaning of the argument `axis` there is different from numpy's.


delta2323:
LGTM

PR #4601, opened 2018-04-10
[backport] Add Sequential class for easy model definition of a single-stream computational graph

beam2d:
Backport of #2918


mitmul:
LGTM. We need to backport #4605 too.

PR #4604, opened 2018-04-10
improve docs of get_device_from_array

### LINKED ISSUE #4595: Inconsistent behavior between `cuda.get_array_module` and `cuda.get_device_from_array`
`cuda.get_array_module` will return correct module even if the input is a chainer `Variable`, however, `cuda.get_device_from_array` will return `DummyDeviceType` if the input is `Variable`.

Though we can still simply use `variable.data` to get the correct results, the ambiguous behavior makes user confused when debugging (at least for me). Thus I think it is better to modify the `cuda.get_device_from_array` to have the similar behavior to `cuda.get_array_module`, i.e., automatically detect the `Variable` as input, and use `variable.data` to get the expected result.

Or at least, we should mention this difference between the two functions.

kmaehashi:
Fix #4595.

### DIFF
--- a/chainer/backends/cuda.py
+++ b/chainer/backends/cuda.py
@@ -174,7 +174,7 @@ def get_device_from_array(*arrays):
         Especially note that, unlike :func:`get_array_module`, this method
         does not recognize :class:`~chainer.Variable` objects.
         If you need to get device from the :class:`~chainer.Variable` instance
-        ``v``, you need to use ``get_device_from_array(v.data)``.
+        ``v``, you need to use ``get_device_from_array(v.array)``.
 
     Args:
         arrays (:class:`cupy.ndarray` or list of :class:`cupy.ndarray`):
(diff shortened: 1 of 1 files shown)


chainer/backends/cuda.py
```diff
@@ -168,8 +168,16 @@ def get_device_from_array(*arrays):
 
     The device on which the given CuPy array reside is returned.
 
+    .. note::
+
+        This method only recognizes :class:`cupy.ndarray`\\ s in arguments.
+        Especially note that, unlike :func:`get_array_module`, this method
+        does not recognize :class:`~chainer.Variable` objects.
+        If you need to get device from the :class:`~chainer.Variable` instance
+        ``v``, you need to use ``get_device_from_array(v.data)``.
```
mitmul: Accessing the array object via `data` attribute is discouraged. (Please see: [Variable.data](https://docs.chainer.org/en/stable/reference/core/generated/chainer.Variable.html?highlight=Variable#chainer.Variable.data))
So, I think it's better to use `v.array` here.

kmaehashi: Thank you, fixed.

PR #4606, opened 2018-04-10
Fix installation order of hacking

toslunar:
It seems pycodestyle==2.4.0 breaks something.  To keep hacking-related packages from being overwritten by pycodestyle, let's install hacking later.


niboshi:
LGTM!

Issue #2903, opened 2017-06-22
A training loop got stuck in a certain condition with multi-processing updater and opencv. 


apple2373:
A training loop got stuck in a certain condition with multi-processing updater and opencv. The issue does not appear when I use Pillow or serial updater. 

* Conditions
  - Chainer version: 2.0
  - CuPy version: 1.0.0.1
  - OS/Platform Ubuntu 14.04.5 (for PFN ppl sakura server 1)
  - CUDA/cuDNN version: V8.0.44, I don't know the way to check CUDNN version....
* Code to reproduce
https://github.com/apple2373/chainer-train-stuck
Note that this requires other libraries such as  chainercv, open cv, etc.... 
then 
`python train.py --gpu 0 --mode 0`
* Error messages, stack traces, or logs
No message when stuck but the I can get the following message when I abort it. 

stsutsui@sakura1:/mnt/sakura201/stsutsui/chainer-train-stuck$ python train.py --gpu 0 --mode 0
epoch       iteration   elapsed_time  main/loss   main/accuracy
0           1           7.3559        1.07867     0.498882       
^CProcess Process-8:..........................................]  1.00%
Process Process-7:###########.................................] 35.29%
Traceback (most recent call last):rations
Traceback (most recent call last):me to finish: 0:00:00.
  File "/mnt/sakura201/stsutsui/anadonda2/lib/python2.7/multiprocessing/process.py", line 258, in _bootstrap
  File "/mnt/sakura201/stsutsui/anadonda2/lib/python2.7/multiprocessing/process.py", line 258, in _bootstrap
    self.run()
  File "/mnt/sakura201/stsutsui/anadonda2/lib/python2.7/multiprocessing/process.py", line 114, in run
    self._target(*self._args, **self._kwargs)
  File "/mnt/sakura201/stsutsui/anadonda2/lib/python2.7/site-packages/chainer/iterators/multiprocess_iterator.py", line 386, in _worker
    self.run()
  File "/mnt/sakura201/stsutsui/anadonda2/lib/python2.7/multiprocessing/process.py", line 114, in run
    self._target(*self._args, **self._kwargs)
  File "/mnt/sakura201/stsutsui/anadonda2/lib/python2.7/site-packages/chainer/iterators/multiprocess_iterator.py", line 386, in _worker
    cnt, mem_index, index = in_queue.get()
    cnt, mem_index, index = in_queue.get()
[...]

PR #4607, opened 2018-04-11
Stalled dataset detection in MultiprocessIterator

niboshi:
This fix should be able to detect stalled detaset (e.g. the case in #2903) and redirect users to appropriate documentation.


niboshi:
Ready for review.

chainer/iterators/multiprocess_iterator.py
```diff
@@ -62,20 +87,20 @@ class MultiprocessIterator(iterator.Iterator):
...
     _interruption_testing = False  # for testing
     _finalized = False
+    _prefetch_loop = None
     _comm = None
-    _thread = None
 
     def __init__(self, dataset, batch_size, repeat=True, shuffle=None,
                  n_processes=None, n_prefetch=1, shared_mem=None,
-                 order_sampler=None):
+                 order_sampler=None, dataset_timeout=30.0):
```
okuta: I think this change is no compatible. How about you?

niboshi: In which case do you think it's no-compat?

okuta: The current implementation does not cause a timeout.
Is a 30 second timeout fine?

niboshi: It depends on how user dataset is implemented.
For example, it could read a file with stale nfs handle.

I think 30 seconds is reasonable.
Usually dataset realization does not take a second. 30 seconds is a reasonable time for human to run watch-for-error-and-fix loop.

kmaehashi: Some datasets like `ZippedImageDataset` requires initialization at first access; opening the whole ZIP file may take seconds. How about this idea?
* By default, show warning in the parent process (instead of raising an exception) when stall is detected.
* Introduce a bool flag to raise an exception instead of warning.

niboshi: I think that sounds good.

niboshi: I fixed as such.
I didn't introduce the flag, because users can turn the warning into an error by using `warnings.filterwarnings`.

okuta: OK, Thanks

okuta:
Please check appveyor result.

okuta:
Hmm, AppVeryor test was failed.

niboshi:
Fixed. PTAL

okuta:
jenkins, test this please.

okuta:
LGTM!

PR #4608, opened 2018-04-11
Make updaters call `new_epoch` automatically

### LINKED ISSUE #3044: StandardUpdater should call Optimizer.new_epoch appropriately
I noticed that StandardUpdater doesn't call Optimizer.new_epoch.
So if you use an optimizer with StandardUpdater and Trainer, optimizer never knows what epoch is going on.

Fixing the issue is simple: just add
```python
if self.is_new_epoch:
    optimizer.new_epoch()
```
[here](https://github.com/chainer/chainer/blob/master/chainer/training/updater.py#L200).

Questions are

1. Is this behavior is intended?
2. If so, is it for compatibility?
3. If not intended, how do you think about changing this behavior, which breaks compatibility?

Thanks,

beam2d:
Fixed #3044. This PR lets the built-in updaters automatically call `Optimizer.new_epoch()` when `is_new_epoch` is true for the main iterator.

**This change is incompatible with v4.** To aid the incompatibility, I added `use_auto_new_epoch` flag to `Optimizer`, which is set by the updaters. When this flag is set, `new_epoch()` called outside the updater raises an error so that users can notice duplicated calls of `new_epoch()`. It means that existing code that uses any updater and `new_epoch` simultaneously will raise an error. This error can be fixed by 1) not calling `new_epoch` outside the updater, or 2) passing `auto_new_epoch=False` to the updater.

### DIFF
--- a/chainer/optimizer.py
+++ b/chainer/optimizer.py
@@ -371,9 +371,10 @@ class Optimizer(object):
             :meth:`update` method.
         ~Optimizer.epoch: Current epoch. It is incremented by the
             :meth:`new_epoch` method.
-        use_auto_new_epoch: Boolean flag to indicate if :meth:`new_epoch` will
-            be called by the updater. Updater should set this flag to ``True``
-            if it automatically calls :meth:`new_epoch`.
+        ~Optimizer.use_auto_new_epoch: Boolean flag to indicate if
+            :meth:`new_epoch` will be called by the updater. Updater should
+            set this flag to ``True`` if it automatically calls
+            :meth:`new_epoch`.
 
     """
 
--- a/chainer/optimizer.py
+++ b/chainer/optimizer.py
@@ -463,7 +463,8 @@ class Optimizer(object):
             if not self.use_auto_new_epoch:
                 raise RuntimeError(
                     'invalid new_epoch call with auto=True.\n'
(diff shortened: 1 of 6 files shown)


niboshi:
Please fix the style error.

niboshi:
Please check test failure.

niboshi:
>chainer/optimizer.py:docstring of chainer.Optimizer:None: WARNING: more than one target found for cross-reference 'use_auto_new_epoch': chainer.optimizers.MomentumSGD.use_auto_new_epoch, chainer.optimizers.SGD.use_auto_new_epoch, chainer.optimizers.AdaGrad.use_auto_new_epoch, chainer.optimizers.RMSprop.use_auto_new_epoch, chainer.GradientMethod.use_auto_new_epoch, chainer.optimizers.SMORMS3.use_auto_new_epoch, chainer.Optimizer.use_auto_new_epoch, chainer.optimizers.NesterovAG.use_auto_new_epoch, chainer.optimizers.Adam.use_auto_new_epoch, chainer.optimizers.AdaDelta.use_auto_new_epoch, chainer.optimizers.RMSpropGraves.use_auto_new_epoch

It seems you need to specify class names in the docstrings.

beam2d:
Fixed the docstring.

niboshi:
jenkins, test this please

niboshi:
LGTM!

PR #4614, opened 2018-04-12
simplify rsqrt

kmaehashi:


niboshi:
What is the point of this change?
In my benchmark, they're equally as fast.

(In NumPy, `x ** -1.5` is faster than `x ** 3`.)

```
power_float         :   412.896 us   +/-65.611 us    415.540 us   +/-79.013 us
power_int           :   562.633 us   +/-93.729 us    555.432 us   +/-127.811 us
```

okuta:
Retaining output is more memory efficient. So, I accept this PR.

okuta:
LGTM!

PR #4616, opened 2018-04-12
[backport] fix seq2seq example ignoring limitation for target length

mitmul:
Backport of #4611


mitmul:
Should be targetted for v4.

PR #4618, opened 2018-04-12
[backport] Save snapshot in the OS default permission

niboshi:
Backport of #4461


niboshi:
We've merged a backport PR #4524 already but it was incorrectly based on v3 branch.
We need to revert it and make another merge into v4 instead.

kmaehashi:
LGTM!

PR #4619, opened 2018-04-12
Revert "[backport] Save snapshot in the OS default permission"

niboshi:
Reverts chainer/chainer#4524

This backport PR was originally intended for v4 but was incorrectly based on v3.


kmaehashi:
LGTM!
(CI failures are ignored as this is caused by hacking order issue fixed in #4606 which is not backported)

PR #4620, opened 2018-04-13
Revert "[backport] Implement caffe export"

mitmul:
Reverts chainer/chainer#4567

It should have been backported to v4 branch.


kmaehashi:
LGTM!
(CI failures are ignored as this is caused by hacking order issue fixed in #4606 which is not backported)

PR #4622, opened 2018-04-13
rewrite installation guide to align with CuPy's installation guide

kmaehashi:
I removed `protobuf 3.0+` from optional dependencies, as it is automatically installed as a dependency of Chainer.
(Moreover, prptobuf wheels hosted on PyPI includes C++ implementations; there's no need of manually installing it separately)

I also added pointer to FAQ section of CuPy.

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -6,61 +6,64 @@ Installation
 Recommended Environments
 ------------------------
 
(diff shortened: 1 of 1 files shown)


kmaehashi:
Thank you! Could you backport this to v4?

PR #4630, opened 2018-04-16
Add error handler interface to trainer extensions

kuenishi:
This commit also adds `snapshot_on_error` option to snapshot extension as a simple example of error handling in extensions, which I suggested at #4566. Use cases are:

* use cases like snapshots, or [checkpoints](http://chainermn.readthedocs.io/en/stable/reference/#chainermn.create_multi_node_checkpointer) it would be nice if the trainer extensions have error handling interface to preserve the *latest* trainer image, or models. See changes in this PR on snapshot extension, which has additional `snapshot_on_error` flag for actual implementation.
* ChainerUI currently just become silent when the training script dies, but it's nice if friendly message and stacktrace can be displayed on the Web UI.
* To handle some cases node failure or GPU failure in parallel training by ChainerMN.

### DIFF
--- a/chainer/training/extensions/_snapshot.py
+++ b/chainer/training/extensions/_snapshot.py
@@ -6,7 +6,7 @@ from chainer.training import extension
 from chainer import utils
 
 
-def snapshot_object(target, filename, savefun=npz.save_npz):
+def snapshot_object(target, filename, savefun=npz.save_npz, **kwargs):
     """Returns a trainer extension to take snapshots of a given object.
 
     This extension serializes the given object and saves it to the output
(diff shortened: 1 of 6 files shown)


chainer/training/trainer.py
```diff
@@ -315,6 +315,19 @@ def run(self, show_loop_exception_msg=True):
...
+                    try:
+                        # It is guaranteed all handlers are called,
+                        # but exceptions thrown by those handlers are
+                        # just printed and ignored, as well as its
+                        # return values.
+                        handler(self, e, sys.exc_info()[2])
+                    except Exception as he:
+                        f.write('Exception in error handler: {}\n'.format(he))
+                        traceback.print_tb(sys.exc_info()[2])
+                        f.write('Traceback (most recent call last):\n')
```
kmaehashi: These two lines should be reversed?

kmaehashi:
BTW, do you want this feature to be backported to v4.x?

kmaehashi:
Also, please confirm the CI error (it seems it behaves differently on Python 2.7).

kuenishi:
Thanks for the review and sorry for leaving for months. I'll try fixing 2.7 failure soon.

> BTW, do you want this feature to be backported to v4.x?

No, I don't.

kuenishi:
Now I believe I fixed failures in Python 2 and updated to the comments. If the test passes it'll be ready for review again.

examples/mnist/train_mnist.py
```diff
@@ -85,7 +85,8 @@ def main():
 
     # Take a snapshot for each specified epoch
     frequency = args.epoch if args.frequency == -1 else max(1, args.frequency)
-    trainer.extend(extensions.snapshot(), trigger=(frequency, 'epoch'))
+    trainer.extend(extensions.snapshot(snapshot_on_error=True),
```
kmaehashi: I think it's better to avoid modifying the example, as this file is often used as a boilerplate and not all users want this behavior. (when an error occurs **during** the 10th iteration, a snapshot named `snapshot_iter_10` will be produced, but users cannot distinguish if it with the snapshot that **completed** the 10th iteration.)

chainer/training/extensions/_snapshot.py
```diff
@@ -69,11 +87,28 @@ def snapshot(savefun=npz.save_npz,
...
             It can be a format string, where the trainer object is passed to
             the :meth:`str.format` method.
+        snapshot_on_error (bool): Whether to take a snapshot in case trainer
+            loop has been failed.
 
     """
-    @extension.make_extension(trigger=(1, 'epoch'), priority=-100)
+    snapshot_on_error = utils.argument.parse_kwargs(
+        kwargs, ('snapshot_on_error', False),
+        deterministic="deterministic argument is not supported anymore. "
```
kmaehashi: ditto

kuenishi:
I fixed on your review comments and now the test has passed agian too.

kmaehashi:
LGTM!
Jenkins, test this please.

kmaehashi:
LGTM!

kuenishi:
Thanks!

PR #4631, opened 2018-04-16
fix heading level of Caffe docs

kmaehashi:
Currently both "Caffe Reference Model Support" and "Caffe Model Export Support" appearing in the references on the left.


okuta:
LGTM!

PR #4638, opened 2018-04-17
Add Group Normalization

lyakaap:
It is an implementation of GroupNormalization.
https://arxiv.org/pdf/1803.08494.pdf 

GN can outperform in tasks such as time-series, GANs, and object detection.
This implementation uses batch normalization internally because we can exploit their optimized implementation with less effort.

### DIFF
--- a/chainer/functions/normalization/group_normalization.py
+++ b/chainer/functions/normalization/group_normalization.py
@@ -1,0 +1,70 @@
+import six
+import warnings
+
+from chainer.backends import cuda
+from chainer.functions.array import broadcast
+from chainer.functions.array import reshape
+from chainer.functions.normalization import batch_normalization
+
+
+def group_normalization(x, groups, gamma, beta, eps=1e-5):
+    """Group normalization.
+
+    This function implements a "group normalization"
+    which divides the channels into groups and computes within each group
+    the mean and variance, then normalize by these statistics,
+    scales and shifts them.
+
+
+    Args:
+        x (~chainer.Variable): Batch tensors.
+            First dimension of this value must be the size of minibatch and
+            second dimension must be the number of channels.
+            Moreover, this value must have one or more following dimensions,
+            such as height and width.
+        groups (int):
+            The number of channel groups.
+            This value must be a divisor of the number of channels.
+        gamma (~chainer.Variable): Scaling parameter.
+        beta (~chainer.Variable): Shifting parameter.
+        eps (float): Epsilon value for numerical stability of normalization.
+
+
+    Returns:
+        ~chainer.Variable: The output variable which has the same shape
+        as :math:`x`.
+
+    See: `Group Normalization <https://arxiv.org/abs/1803.08494>`_
+    """
+    if x.ndim <= 2:
+        raise ValueError('Input dimension must be bigger than 2.')
+
+    xp = cuda.get_array_module(x)
+
+    batch_size, channels = x.shape[:2]
+    original_shape = x.shape
+
+    if channels % groups != 0:
+        raise ValueError(
+            'Argument \'groups\' must be a divisor of the number of channel.')
+
+    # By doing this reshaping, calling batch_normalization function is
+    # equivalent to Group Normalization.
+    x = reshape.reshape(x, (1, batch_size * groups, -1))
+    dummy_gamma = (xp.ones(batch_size * groups).astype(xp.float32))
+    dummy_beta = (xp.zeros(batch_size * groups).astype(xp.float32))
+
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore")
+        x = batch_normalization.batch_normalization(
+            x, dummy_gamma, dummy_beta, eps=eps)
+
+    x = reshape.reshape(x, original_shape)
+
+    target_shape = [1, channels] + [1 for _ in six.moves.xrange(x.ndim - 2)]
+    gamma_broadcast = broadcast.broadcast_to(
+        reshape.reshape(gamma, target_shape), x.shape)
+    beta_broadcast = broadcast.broadcast_to(
+        reshape.reshape(beta, target_shape), x.shape)
+
+    return x * gamma_broadcast + beta_broadcast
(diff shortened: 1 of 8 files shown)


crcrpar:
I wonder why this implementation assumes inputs' dimension is greater than 2 because group normalization will work if the input is 2D.
Though the original paper assumes group normalization is applied to images, some implementations e.g. pytorch accepts 2D input.

Ignore this if wrong and/or ignorable.

crcrpar:
I appreciate your response.

> Or you mean, GN should also accept 2D[batch size, channel size] input?

No, I was just wondering the reason and understood well now.
Thank you.

lyakaap:
No problem. And thanks for pointing.
I'm not sure whether it should accept 1d input or not, I changed error message clearer.

tkerola:
Hi, I saw that `groups` was changed to `n_groups` in this PR, and would like to comment on consistency in the Chainer API. In `Convolution2DFunction`, `groups` is already used to indicate the number of groups used in grouped convolution. As the meaning of groups in this PR is related (each group is processed separately), I would like to hear you opinion on whether both functions should have the same keyword naming in order to make the API consistent? https://github.com/chainer/chainer/blob/master/chainer/functions/connection/convolution_2d.py#L39
If so, perhaps the we should revert to use `groups` instead of `n_groups` in this PR. Thanks.

lyakaap:
I didn't know that and I also think it would be good to fit the naming consistency.
Also in tensorflow, `groups` is used in GN arg name. https://github.com/tensorflow/tensorflow/blob/d0f5bc17560fc97bcc7de9164aa3b237a8d5221d/tensorflow/contrib/layers/python/layers/normalization.py#L167
I will update.

niboshi:
jenkins, test this please

niboshi:
(Ignore AppVeyor failure)

I observe two failures in our internal test. Could you check them?
One is in multi-gpu test:
```
E   AssertionError: Parameterized test failed.
E   
E   Base test method: GroupNormalizationTest.test_forward_multi_gpu
E   Test parameters:
E     dtype: <type 'numpy.float32'>
E     shape: (5, 4, 15)
E     groups: 1
E   
E   AssertionError: 
E   Fail: 3, Success: 0
E   
E   The first error message:
E   Traceback (most recent call last):
E     File "chainer/testing/condition.py", line 60, in <lambda>
E       lambda: f(ins, *args[1:], **kwargs),
E     File "/work/chainer/tests/chainer_tests/links_tests/normalization_tests/test_group_normalization.py", line 66, in test_forward_multi_gpu
E       self.check_forward(x)
E     File "/work/chainer/tests/chainer_tests/links_tests/normalization_tests/test_group_normalization.py", line 32, in check_forward
E       y = self.link(x_data)
E     File "chainer/links/normalization/group_normalization.py", line 89, in __call__
E       x, self.groups, self.gamma, self.beta, self.eps)
E     File "chainer/functions/normalization/group_normalization.py", line 67, in group_normalization
E       x, dummy_gamma, dummy_beta, eps=eps)
E     File "chainer/functions/normalization/batch_normalization.py", line 798, in batch_normalization
E       axis).apply((x, gamma, beta))[0]
E     File "chainer/function_node.py", line 258, in apply
E       outputs = self.forward(in_data)
E     File "chainer/functions/normalization/batch_normalization.py", line 250, in forward
E       self.inv_std[expander], gamma, beta)
E     File "chainer/functions/normalization/batch_normalization.py", line 694, in _apply_bn_fwd
E       )(x, mean, inv_std, gamma, beta)
E     File "cupy/core/elementwise.pxi", line 580, in cupy.core.core.ElementwiseKernel.__call__
E     File "cupy/core/elementwise.pxi", line 148, in cupy.core.core._preprocess_args
E   ValueError: Array device must be same as the current device: array device = 0 while current = 1
```

Another is documentation. Please add the link to documentation, too.
```
Warning, treated as error:
WARNING: Undocumented entities found.

chainer.links.GroupNormalization
```

lyakaap:
About multi-gpu forward issue:
In multi-gpu test, I followed the test code of batch_normalization,
https://github.com/chainer/chainer/blob/a137f39b2a511ef5ff40a9a363e13a0df651c221/tests/chainer_tests/links_tests/normalization_tests/test_batch_normalization.py#L123
I guess device inconsistency comes from this part, but I would like to know why this part became like this.

niboshi:
Chainer functions and links should perform computation on the device on which input arrays are allocated, no matter what the current device is.

It seems the problem is that `F.group_normalization` does not set the current device to `x.device`.

https://github.com/chainer/chainer/pull/4638/files#diff-9ff199fca61b38b330297137894e651dR61

Here `dummy_gamma` and `dummy_beta` are allocated on the device 0 whereas `x` is on the device 1, causing `F.batch_normalization` to fail.

niboshi:
jenkins, test this please

lyakaap:
Thanks for the explanation. I pushed the modification.

chainer/functions/normalization/group_normalization.py
```diff
@@ -0,0 +1,78 @@
...
+    if channels % groups != 0:
+        raise ValueError('Argument: \'groups\' must be a divisor '
+                         'of the number of channel.')
+
+    # By doing this reshaping, calling batch_normalization function becomes
+    # equivalent to Group Normalization.
+    # And redundant dimension is added in order to utilize ideep64/cuDNN.
+    x = reshape.reshape(x, (1, batch_size * groups, -1, 1))
+
+    with cuda.get_device_from_array(x):
```
niboshi: `cuda.get_device_from_array` only accept `cupy.ndarray`s.
Please fix it as `with cuda.get_device_from_array(x):`.
(I confirmed the test to pass with this change).

niboshi:
Thanks for a quick fix!
Jenkins, test this please

niboshi:
Jenkins, test this please

kmaehashi:
Jenkins, test this please.

kmaehashi:
Jenkins, test this please.

niboshi:
Jenkins, test this please.

niboshi:
LGTM!
Thank you very much!

Issue #4016, opened 2017-12-02
Proposal chainer.einsum


fukatani:
I want to discuss `chainer.einsum` function.

I think `chainer.einsum` can be implemented as follows.
1. Implemented `chainer.diagonal` and `chainer.moveaxis` (as function).
2. Copy `cupy.einsum`.
3. Replace cupy module to chainer.functions.
ex.
`cupy.diagonal` to chainer.diagonal, `cupy.matmul` to `chainer.matmul`.

But I am not confident that sufficient performance will come out..

PR #4644, opened 2018-04-17
Add `F.einsum`

toslunar:
See also #4016.

This PR implements einsum as a single `FunctionNode`. This approach is different from https://github.com/chainer/chainer/issues/4016#issue-278690083.

If three or more operands are given, as discussed `Prod.backward` somewhere, I believe that the backward would be faster if we separated FunctionNode into steps of `numpy.einsum_path` (c.f. `optimize` option of np.einsum).
It's possible that sum operations in a single operand (e.g. `b` in `einsum('ab,ac->c', x, y)`) should be preprocessed alone for the same reason.


kmaehashi:
We should be aware that there is an inconsistency between syntax of `einsum` subscripts between C parser and Python parser (which is contained in this PR).

toslunar:
On `einsum` subscripts, the permissive syntax is natural to me and the parser in this PR is based on the "newer" parser.  (I further generalized it to a kind of "diag einsum" syntax.)
The issue numpy/numpy#10926 was avoided at the commit 6e176507b6dbb14ab7b959c3bfb383e1757eb406 in this PR, probably.

toslunar:
With the patch cupy/cupy#1181, `python -m pytest chainer_tests/functions_tests/math_tests/test_einsum.py` passes (including gpu tests).

toslunar:
The PR requires fast `cupy.einsum`.

kmaehashi:
@toslunar Could you tell me what PR blocks this PR?

kmaehashi:
Thank you for clarification!

> This PR needs work. numpy.einsum (with default optimize=False before the next version 1.15) with 3 or more operands is also slow. To support faster einsum in CPU, this PR should be totally rewritten as I commented:

Do you mean you don't need this work if `optimize=True` can be used? I'm not sure how much is that work, but I think it is also acceptable to just use `optimize=True` for NumPy 1.12+ and show warning when older NumPy (that does not support optimize) is used.

toslunar:
cupy/cupy#1218 is merged

okuta:
jenkins, test this please.

chainer/functions/math/einsum.py
```diff
@@ -0,0 +1,303 @@
...
+        for s in sorted(set(tmp_subscripts)):
+            if s == '@' or tmp_subscripts.count(s) == 1:
+                output_subscript += s
+
+    # Make sure number operands is equivalent to the number of terms
+    if len(input_subscripts.split(',')) != len(operands):
+        raise ValueError("Number of einsum subscripts must be equal to the "
+                         "number of operands.")
+
+    return (input_subscripts, output_subscript, operands)
```
okuta: Please remove parenthesis.

tests/chainer_tests/functions_tests/math_tests/test_einsum.py
```diff
@@ -0,0 +1,317 @@
...
+        inputs = _tuple_to_gpu(self.inputs)
+        if self.dtype == numpy.float16:
+            self.check_forward(inputs, atol=1e-3, rtol=1e-3)
+        else:
+            self.check_forward(inputs)
+
+    def check_backward(self, inputs_data, output_grad, atol, rtol):
+        gradient_check.check_backward(
+            self.op, inputs_data, output_grad, atol=atol, rtol=rtol,
+            dtype=numpy.float32)
```
okuta: Why don't use `float64`?

tests/chainer_tests/functions_tests/math_tests/test_einsum.py
```diff
@@ -0,0 +1,317 @@
...
+
+    def check_double_backward(
+            self, inputs_data, y_grad, inputs_grad_grad,
+            atol, rtol):
+        def nonlinear(*inputs):
+            y = self.op(*inputs)
+            return y * y
+        gradient_check.check_double_backward(
+            nonlinear, inputs_data, y_grad, inputs_grad_grad,
+            atol=atol, rtol=rtol, dtype=numpy.float32)
```
okuta: ditto.

chainer/functions/math/einsum.py
```diff
@@ -0,0 +1,303 @@
...
+            if xp is not cuda.cupy:  # no setflags in cupy
+                diag_y.setflags(write=True)
+            diag_y[...] = _einsum(
+                xp, dtype, self.in_subs, ''.join(direct_sub), *inputs
+            )[expander]
+        return y,
+
+
+def einsum(*operands):
+    input_subscripts, output_subscript, ioperands = \
```
okuta: Please add document

tests/chainer_tests/functions_tests/math_tests/test_einsum.py
```diff
@@ -0,0 +1,317 @@
...
+        self.check_backward(
+            _tuple_to_gpu(self.inputs),
+            cuda.to_gpu(self.g), atol=1e-2, rtol=5e-2)
+
+    def check_double_backward(
+            self, inputs_data, y_grad, inputs_grad_grad,
+            atol, rtol):
+        def nonlinear(*inputs):
+            y = self.op(*inputs)
+            return y * y
```
toslunar: #4978

toslunar:
Jenkins, test this please.

toslunar:
The Travis failure (`TestL2Normalization_param_18.test_backward_cpu`) is irrelevant to the PR.

okuta:
LGTM. Please resolve conflicts.

okuta:
jenkins, test this please.

okuta:
LGTM!

PR #4646, opened 2018-04-17
[WIP] Add option to allow None as zero in gradient_check

toslunar:


toslunar:
It's rather a bugfix, while simplifying tests is an enhancement.  I separated the bugfix part as #4654.

toslunar:
Moved to #4978

PR #4649, opened 2018-04-17
Fix Chain.repeat

mori97:
There was a error when I tried to use Chain.repeat.

```
class ConvBNReLU(chainer.Chain):
    def __init__(self):
        super(ConvBNReLU, self).__init__()
        with self.init_scope():
            self.conv = L.Convolution2D(
                None, 64, 3, 1, 1, nobias=True)
            self.bn = L.BatchNormalization(64)
    
    def __call__(self, x):
        return F.relu(self.bn(self.conv(x)))

net = ConvBNReLU().repeat(16, mode='init')
```

`TypeError: copy() takes 1 positional argument but 2 were given`

This is the example in the document of Link.repeat.
It fails because Chain.copy does not have "mode" parameter.

I fixed this bug and added some tests for Chain.copy and Chain.repeat.

### DIFF
--- a/chainer/link.py
+++ b/chainer/link.py
@@ -847,13 +847,13 @@ Assign a Link object directly to an attribute within a \
         with self.init_scope():
             setattr(self, name, link)
 
-    def copy(self):
+    def copy(self, mode='share'):
         ret = super(Chain, self).copy()
         ret._children = set(ret._children)
         d = ret.__dict__
         for name in ret._children:
             # copy child links recursively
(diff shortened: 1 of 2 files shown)


mitmul:
Jenkins, test this please.

mitmul:
Thank you for the PR. I think it also happens in `ChainList`, so I'll make another PR with the same fix for `ChainList` after I merge this PR.

mitmul:
LGTM

PR #4654, opened 2018-04-18
Fix no_grads option of check_backward

toslunar:
- "their gradients are ignored" in the docstring was ambiguous.
- `check_backward` is supposed to compare backward with the numerical one. Do not check computational graph.

### DIFF
--- a/chainer/gradient_check.py
+++ b/chainer/gradient_check.py
@@ -355,8 +355,14 @@ def check_backward(
     If :math:`r` is chosen from uniform distribution, we can conclude with
     high probability that the gradient of :math:`f` itself is correct.
 
-    If input objects (``x1_data`` or/and ``x2_data`` in this example) represent
-    integer variables, their gradients are ignored.
+    If the function is non-differentiable with respect to some input objects,
+    we can check its backprop to such objects by ``no_grads`` argument.
+    ``gradient_check`` computes numerical backward to inputs that correspond to
+    ``False`` in ``no_grads``. It also asserts the backprop lefts gradient
+    ``None`` for inputs that correspond to ``True`` in ``no_grads``.
+    The default of ``no_grads`` argument is the tuple of truth values whether
+    input objects (``x1_data`` or/and ``x2_data`` in this example) represent
+    integer variables.
 
     You can simplify a test when ``MyFunc`` gets only one argument::
 
@@ -479,10 +485,6 @@ def check_backward(
             if x.grad is not None:
                 raise RuntimeError(
                     'gradient of int variable must be None')
-        else:
-            if x.grad is None:
-                raise RuntimeError(
-                    'gradients of some arguments are not calculated')
 
     if len(xs) - no_grads.count(True) + len(params) == 0:
         # When there is no float variables, we need not to check gradient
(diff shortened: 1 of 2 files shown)


chainer/gradient_check.py
```diff
@@ -479,10 +485,6 @@ def check_backward(
             if x.grad is not None:
```
hvy: We can merge the above two conditions since the `else` case is removed.

chainer/gradient_check.py
```diff
@@ -355,8 +355,14 @@ def check_backward(
     If :math:`r` is chosen from uniform distribution, we can conclude with
     high probability that the gradient of :math:`f` itself is correct.
 
-    If input objects (``x1_data`` or/and ``x2_data`` in this example) represent
-    integer variables, their gradients are ignored.
+    If the function is non-differentiable with respect to some input objects,
+    we can check its backprop to such objects by ``no_grads`` argument.
+    ``gradient_check`` computes numerical backward to inputs that correspond to
+    ``False`` in ``no_grads``. It also asserts the backprop lefts gradient
```
hvy: Could you rephrase this as `... asserts that the backprop leaves gradients as None for inputs ...` ?

hvy:
Is the documentation for `no_grads` argument still correct after this change? Maybe we should change it accordingly too.

hvy:
jenkins, test this please.

toslunar:
This proposal changes a part of undocumented behavior
- from: `check_backward` asserts that `backward` "skips" gradient of a variable *if and only if* it's a `no_grad` variable.
- to: `check_backward` asserts ... *if* ...

In other words, from checking algorithm to checking analysis.

[no-compat] The PR makes more `check_backward` tests pass.

[needs-discussion] There's a choice between
- with the original behavior, a user can detect (inefficient) zero grads that should be replaced with `None`; or
- with this change, a user will not mind the computational graph here.

I found the *only-if* part of the check confused some contributors (#4656). It also makes `check_double_backward` require non-affine forward.

niboshi:
>The PR makes more `check_backward` tests pass.

Could you elaborate what kind of test would pass with this change?
If a backward function is supposed to return a gradient but actually returned `None`, does this change make the test pass?

toslunar:
>>The PR makes more `check_backward` tests pass.
>
>Could you elaborate what kind of test would pass with this change?
>If a backward function is supposed to return a gradient but actually returned `None`, does this change make the test pass?

- This PR does not make this test pass in general.
- If a backward function is supposed to return a *zero* gradient but actually returned `None`, the PR make the test pass.
- If a backward function is supposed to return `None` but actually returned a zero gradient, the PR does not make the test pass, (because `no_grads` is set).
  - However users will have less motivation to distinguish `None` from zero after the PR.
- If a user forget to set `no_grads` to a non-differentiable float input, the result changes to `AssertionError` (on the comparison of directional derivatives) from `RuntimeError('gradients of some arguments are not calculated')`.  I think neither of the messages is user-friendly for this non-differentiable case.

niboshi:
Thank you for clarification.
This change looks OK to me.

hvy:
jenkins, test this please.

hvy:
I also think this PR is okay additionally considering the fact the it used to be poorly documented anyway.

hvy:
jenkins, test this please.

hvy:
LGTM!

PR #4673, opened 2018-04-23
Infer input size in batchnorm using aggregate axes

### LINKED ISSUE #3221: Defer initialization of size of BatchNormalization and infer it at first iteration.

toslunar:
This PR takes over #4149.  Make `size` optional in `L.BatchNormalization` and infer it using `axis` (aggregate axes), which is added in #4266.

Close #3221.


toslunar:
To the reviewer: it's ok to me to split the example document as a new PR.

chainer/links/normalization/batch_normalization.py
```diff
@@ -68,35 +69,114 @@ class BatchNormalization(link.Link):
...
+            >>> bn.avg_mean.shape
+            (3,)
+
+            >>> bn = chainer.links.BatchNormalization(axis=0)
+            >>> y = bn(x)
+            >>> bn.avg_mean.shape
+            (3, 32, 32)
+
+            You can omit ``size`` if ``axis`` is given.
+
```
delta2323: Could you write examples for the case in which `size` is `None`?

delta2323:
Could you add unit tests to test `size=None` case.

chainer/links/normalization/batch_normalization.py
```diff
@@ -68,35 +69,132 @@ class BatchNormalization(link.Link):
...
+            >>> bn.avg_mean.shape
+            (3,)
+            >>> y = bn(x)
+            >>> y.shape
+            (10, 3, 32, 32)
+            >>> np.testing.assert_allclose(
+            ...     y.array.mean(axis=(0, 2, 3)), 0, atol=1e-6)
+            >>> np.testing.assert_allclose(
+            ...     y.array.var(axis=(0, 2, 3)),
+            ...     x.var(axis=(0, 2, 3)) / (x.var(axis=(0, 2, 3)) + 2e-5),
```
delta2323: I think `np.ones(3, dtype=np.float32)` is simpler.

delta2323: I noticed that BN has `beta` and `gamma`, so, the output `y` is whitened (mean=0, var=1) only when `beta` and `gamma` are initial values or they are removed by setting  `use_gamma` and `use_beta` options to `True` (the example comes under the first case). Therefore, users might misunderstand and think `y` is always whitened if we put these assertions. (sorry, I should have noticed that earlier).

delta2323: Alternative ideas are:
1. Set `use_gamma` and `use_beta` to `False`.
2. Write as `y` is whitened because `beta` and `gamma` are initial values.
In both case, I would suggest to write that `y` is not whitened in general.

toslunar:
Fixed

okuta:
LGTM!

okuta:
@delta2323 Do you have any comments?

chainer/links/normalization/batch_normalization.py
```diff
@@ -68,35 +69,139 @@ class BatchNormalization(link.Link):
...
+            By default, channel axis is (or starts from) the 1st axis of the
+            input shape.
+
+        2. Give the aggregate axes:
+
+            from Chainer v5
+
+            With ``axis`` option, similarly to NumPy, you may specify the
+            aggregate axes, which are treated as the "batch" axes for the
+            batch statistics. You can omit ``size`` if ``axis`` is given.
```
delta2323: It might be helpful to write explicitly that creation of `avg_mean` and `avg_var` is deferred until first forward propagation.

delta2323:
LGTM, waiting for the CI.

delta2323:
jenkins, test this please.

delta2323:
I'm working on it.

niboshi:
When the backport of this PR is merged, the backport of #5017 should also be merged in the same release.

delta2323:
Sorry for late I made a backport PR #5050.

delta2323:
We decided not to backport this PR, as discussed in #5050.

delta2323:
We decided not to backport this PR, as discussed in #5050. Please write comments or requests if you have ones.

PR #4675, opened 2018-04-24
Add training mode checking to `extract` vision models (ResNet, VGG16, GoogLeNet).

### LINKED ISSUE #3210: Behavior of extract() in ResNet and VGG models is different in v1 and v2
In chainer v1, it was assumed that `extract()` methods in ``ResNetLayers`` and ``VGG16Layers`` extract feature vectors in *test* mode.
```python
import chainer.links as L
model = L.ResNet50Layers()
feature = mode.extract([image])  # it is equivalent to extract([image], test=True)
```

However, in v2 we have to explicitly use ``with chainer.use_config()`` clause like that:
```python
with chainer.use_config('train', False):
    feature = mode.extract([image])
```
This specification is a bit confusing; users might expect that ``extract()`` literally extracts feature vectors regardless of the training mode, but it is not correct. Although I do not think that it is right thing to change this specification from now on, it would be better to add the above description to the comment.

delta2323:
Fixes #3210


kmaehashi:
Thanks for the PR.
During the discussion of #3210, we concluded that we should not change the behavior nor emit warnings, because (as you stated in the issue) this may be a false warning in some cases.
Could you update the docs?

delta2323:
Updated. Could you review it again?

chainer/links/model/vision/googlenet.py
```diff
@@ -255,10 +255,29 @@ def extract(self, images, layers=None, size=(224, 224), **kwargs):
 
            ``train`` and ``volatile`` arguments are not supported anymore since
            v2.
-           Instead, use ``chainer.using_config('train', train)`` and
-           ``chainer.using_config('enable_backprop', not volatile)``
-           respectively.
-           See :func:`chainer.using_config`.
+           Users should configure training and volatile mode
```
kmaehashi: I think that this issue is not only for users upgrading from v2. The method name `extract` sounds like saying `it does not update parameters at all`.

How about inserting sentence below, before the warning block?

```
Unlike `predict` method, this method does not override `train` and `enable_backprop` configuration.
If you want to extract features without updating model parameters, you need to manually set configuration when calling this method, as follows:

(code example)
```

chainer/links/model/vision/resnet.py
```diff
@@ -213,12 +213,31 @@ def extract(self, images, layers=None, size=(224, 224), **kwargs):
 
         .. warning::
 
-           ``test`` and ``volatile`` arguments are not supported anymore since
+           ``train`` and ``volatile`` arguments are not supported anymore since
```
kmaehashi: This fix seems invalid (the option name was ``test`` in v1.x).

chainer/links/model/vision/vgg.py
```diff
@@ -216,12 +216,31 @@ def extract(self, images, layers=None, size=(224, 224), **kwargs):
 
         .. warning::
 
-           ``test`` and ``volatile`` arguments are not supported anymore since
+           ``train`` and ``volatile`` arguments are not supported anymore since
```
kmaehashi: ditto (``test``)

kmaehashi:
@delta2323 ping? (I can takeover if you don't have time.)

delta2323:
Sorry, I think I can work on it tomorrow. If you cannot wait it. Please feel free to take it over.

kmaehashi:
LGTM! (will merge after CI pass)

kmaehashi:
LGTM.
AppVeyor failure seems unrelated to this PR.

Issue #3968, opened 2017-11-27
Automatic broadcast


beam2d:
In #228, we have once decided not to support automatic broadcast which is error prone for users. It's now time to rethink; it looks more important to have an array semantics closer to NumPy/CuPy, where the lack of automatic broadcast is one of the major gaps between ndarrays and Variables.

The scope would not be limited to basic math operators, but it seems good to start from functions in `basic_math.py`.

kmaehashi:
bump

PR #4679, opened 2018-04-25
Automatic broadcast in basic_math and matmul

beam2d:
This PR makes Variable operators broadcast operands automatically. To make the implementation simple, I added `utils.sum_to` and `F.sum_to`, which computes the gradient of broadcast. Note that other functions that should support automatic broadcast (e.g. `where`) are not covered in this PR. A part of #3968.

### DIFF
--- a/chainer/functions/math/basic_math.py
+++ b/chainer/functions/math/basic_math.py
@@ -148,6 +148,8 @@ class Add(function_node.FunctionNode):
         type_check.expect(
             in_types[0].dtype == in_types[1].dtype,
         )
+        type_check.expect_broadcast_shapes(
+            in_types[0].shape, in_types[1].shape)
 
     def forward(self, x):
         # may broadcast
@@ -236,6 +238,8 @@ class Sub(function_node.FunctionNode):
     def check_type_forward(self, in_types):
         type_check.expect(in_types.size() == 2)
         type_check.expect(in_types[0].dtype == in_types[1].dtype)
+        type_check.expect_broadcast_shapes(
+            in_types[0].shape, in_types[1].shape)
 
     def forward(self, x):
         # may broadcast
@@ -308,6 +312,8 @@ class Mul(function_node.FunctionNode):
             in_types[0].dtype.kind == 'f',
             in_types[0].dtype == in_types[1].dtype,
         )
+        type_check.expect_broadcast_shapes(
+            in_types[0].shape, in_types[1].shape)
 
     def forward(self, x):
         self.retain_inputs((0, 1))
@@ -368,6 +374,8 @@ class Div(function_node.FunctionNode):
             in_types[0].dtype.kind == 'f',
             in_types[0].dtype == in_types[1].dtype,
         )
+        type_check.expect_broadcast_shapes(
+            in_types[0].shape, in_types[1].shape)
 
     def forward(self, x):
         self.retain_inputs((0, 1))
@@ -532,6 +540,8 @@ class PowVarVar(function_node.FunctionNode):
             in_types[0].dtype.kind == 'f',
             in_types[0].dtype == in_types[1].dtype,
         )
+        type_check.expect_broadcast_shapes(
+            in_types[0].shape, in_types[1].shape)
 
     def forward(self, x):
         self.retain_inputs((0, 1))
--- a/chainer/functions/math/basic_math.py
+++ b/chainer/functions/math/basic_math.py
@@ -35,7 +35,7 @@ def _check_constant_type(value):
     elif isinstance(value, (numpy.ndarray, cuda.ndarray)):
         return
     else:
-        raise ValueError(
+        raise TypeError(
             'Value must be a scalar, `numpy.ndarray`, `cupy.ndarray` '
             'or a `Variable`.\nActual: {}'.format(type(value)))
 
@@ -49,6 +49,13 @@ def _preprocess_const(x, value):
     b = xp.broadcast(x, value)
     if b.shape != x.shape:
         raise ValueError('Failed to broadcast arrays')
+    return utils.force_type(x.dtype, value)
+
+
+def _preprocess_rhs(x, value):
+    if isinstance(value, chainer.Variable):
+        return value
+    _check_constant_type(value)
     return utils.force_type(x.dtype, value)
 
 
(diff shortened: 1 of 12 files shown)


toslunar:
I suggest `expect_broadcast_shapes` should accepts two **or more** shapes, if it'll be used everywhere.
`expect_broadcast_shapes(x1, x2) and expect_broadcast_shapes(x2, x3)` does not imply `(x1, x2, x3)` can be broadcasted, e.g. if their shapes are `(2,)`, `(1,)`, `(3,)`

chainer/utils/array.py
```diff
@@ -35,3 +36,16 @@ def empty_like(x):
         return cuda.cupy.empty_like(x)
     else:
         return numpy.empty_like(x)
+
+
+def sum_to(x, shape):
```
toslunar: Could you assert `not isinstance(x, Variable)`?  Once `Variable.sum` is implemented, `chainer.utils.sum_to` can compute `Variable`'s "sum_to" but its backward is slower than `F.sum_to`'s.

toslunar:
Could you support broadcast in `F.matmul` when one of the two operands is a vector? 
 See e.g. https://github.com/cupy/cupy/blob/2b56cea0007d2efd0aad2fecb4f543c08d711f9b/tests/cupy_tests/math_tests/test_matmul.py#L22-L23

chainer/utils/array.py
beam2d: Fixed to check if x is not Variable.

beam2d:
- Rewrote `expect_broadcast_shapes` to support more than two inputs. I now find reusing `expect` inside of it is useless, so I just make it directly check the shapes and raise `InvalidType` with a custom error message.
- Make `matmul` support broadcast of a vector operand.

chainer/utils/type_check.py
```diff
@@ -583,3 +585,33 @@ def prod(xs):
...
     else:
         return _prod(xs)
+
+
+def expect_broadcast_shapes(*shape_types, **kwargs):
+    """Checks if two shapes are broadcastable at first some axes.
+
+    Args:
+        shapes_types: Type-checked shapes of the arrays to broadcast.
+        ignore_tail (int): Number of axes at the tail to ignore.
```
toslunar: Should the function check the shapes have tail dimensions?  I prefer `type_check.expect_broadcast_shapes(a_type.shape[:-2], b_type.shape[:-2])` as the meaning of `ignore_tail` is not clear.

chainer/functions/math/matmul.py
```diff
@@ -88,48 +89,85 @@ def check_type_forward(self, in_types):
...
         if a_ndim == 0 or b_ndim == 0:
             pass
-        elif a_ndim == 1 or b_ndim == 1:
-            type_check.expect(
-                a_type.ndim == b_type.ndim,
-                a_type.shape == b_type.shape,
-            )
+        elif a_ndim == 1:
+            if b_ndim == 1:
+                type_check.expect(a_type.shape == b_type.shape)
```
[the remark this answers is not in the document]
beam2d: I don't think this case can be eliminated. It's a check for the case `a_ndim == 1 == b_ndim`.

chainer/utils/type_check.py
beam2d: Looks better. Fixed accordingly.

beam2d:
@toslunar Sorry for being late. I made a reply and a fix for a comment.

chainer/functions/math/matmul.py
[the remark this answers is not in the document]
toslunar: I meant `a_ndim` doesn't matter if `b_ndim == 1`:
```
        if a_ndim == 0 or b_ndim == 0:
            pass
        elif b_ndim == 1:
            type_check.expect(a_type.shape[-1] == b_type.shape[0])
        elif a_ndim == 1:
            type_check.expect(a_type.shape[0] == b_type.shape[-2])
        else:
```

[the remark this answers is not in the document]
beam2d: Ah, I see. OK, I wil fix it.

beam2d:
Simplified the logic based on the suggestions.

beam2d:
I found that `rollaxis` has a bug in backward. I will fix it in another PR.

beam2d:
Made a fix: #4836.

chainer/utils/type_check.py
```diff
@@ -5,9 +5,11 @@
 import threading
 
 import numpy
+import six
 
 import chainer
 from chainer.backends import cuda
+from chainer.utils import argument
```
toslunar: Remove unused import.

chainer/functions/math/matmul.py
```diff
@@ -88,48 +89,82 @@ def check_type_forward(self, in_types):
...
+            # may broadcast
             y = _matmul(a, b, self.transa, self.transb, self.transc)
-        if self.dtype is not None and y.dtype != self.dtype:
-            y = y.astype(self.dtype)
-        return utils.force_array(y),
+
+        if self.dtype is not None:
+            dtype = self.dtype
+        else:
+            dtype = a.dtype
```
toslunar: Is this a typo of `y.dtype`?

chainer/utils/type_check.py
beam2d: Fixed.

chainer/functions/math/matmul.py
beam2d: Oops, you're right. Fixed.

toslunar:
The travis failure seems due to numpy/numpy#10343.  Use `xp.matmul` whenever it exists.

beam2d:
Thank you, I fixed it to use matmul if available.

chainer/functions/math/basic_math.py
```diff
@@ -368,35 +392,55 @@ def forward_cpu(self, inputs):
...
+                gx0 = -ggx1 * gy / x1_square
+                ret.append(chainer.functions.sum_to(gx0, x0.shape))
+
         if 1 in indexes:
-            gx1 = -ggx0 * gy / x1_square + \
-                ggx1 * 2 * gy * x0 / (x1_square * x1)
-            ret.append(gx1)
+            gx1 = None if ggx0 is None else -ggx0 * gy / x1_square
+            gx1_1 = (None if ggx1 is None else
+                     ggx1 * 2 * gy * x0 * (x1_square * x1))
```
toslunar: `ggx1 * 2 * gy * x0 * (x1_square * x1)` should be
`ggx1 * 2 * gy * x0 / (x1_square * x1)`

beam2d: Thanks, I fixed it.

toslunar:
Could you resolve circular imports?

beam2d:
Made a fix for the circular imports.

toslunar:
A GPU test is failing.  Fix the test: #4863.

beam2d:
Rebased with the fix #4863.

toslunar:
`chainer-doc` test warns there's no doc for `chainer.functions.sum_to`.

beam2d:
Added the reference entry of `sum_to`.

toslunar:
@beam2d Could you check if my fix is ok?

beam2d:
The fix looks good.

PR #4680, opened 2018-04-25
assign user-friendly name for better type check exception message

kmaehashi:
Currently type check error report is difficult to understand for beginners, because:

* the name `in_types` is not intuitive
* the index `X` of `in_types[X]` in the error message is a X-th argument of the FunctionNode, so sometimes it does not match with the X-th argument of `F.xxx`.

This PR is a proposal of assigning a user-friendly name for each types.

Example output:

```
Invalid operation is performed in: SigmoidCrossEntropy (Forward)

Expect: x.shape == t.shape
Actual: (100, 1) != (100,)
```

### DIFF
--- a/chainer/functions/connection/linear.py
+++ b/chainer/functions/connection/linear.py
@@ -14,7 +14,7 @@ class LinearFunction(function_node.FunctionNode):
         n_in = in_types.size()
         type_check.expect(2 <= n_in, n_in <= 3)
         x_type, w_type = in_types[:2]
-        type_check.args((x_type, w_type), ('x', 'W'))
+        type_check.argname((x_type, w_type), ('x', 'W'))
 
         type_check.expect(
             x_type.dtype.kind == 'f',
@@ -25,7 +25,7 @@ class LinearFunction(function_node.FunctionNode):
         )
         if type_check.eval(n_in) == 3:
             b_type = in_types[2]
-            type_check.args((b_type,), ('b',))
+            type_check.argname((b_type,), ('b',))
             type_check.expect(
                 b_type.dtype == x_type.dtype,
                 b_type.ndim == 1,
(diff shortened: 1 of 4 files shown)


beam2d:
This feature looks good. Can we provide a more compact way of specifying all the input names, e.g. in one function or decorator? (sth like `type_check.name(in_types, 'x', 't')` or `@type_check.args('x', 't')`) In this case, I also think it may be possible to integrate the check of `in_types.size()`.

kmaehashi:
Thanks for the positive feedback!
I agree with you. I prefer to use function rather than decorator to make the way to declare names consistent between FunctionNodes whose number of arguments is fixed or not.

How about `type_check.name` to accept tuples so that the size is checked?

For FunctionNodes with fixed number of arguments:

```
type_check.name(in_types, ('x', 't'))
```

For FunctionNodes with dynamic number of arguments (e.g., [LinearFunction](https://github.com/chainer/chainer/blob/v5.0.0a1/chainer/functions/connection/linear.py#L13-L16)):
```
# (manually check in_types.size(), then)
type_check.name((x_type, t_type), ('x', 't'))
```

kmaehashi:
* Renamed to `type_check.args(in_types, names)` so that it can validate input length and assign names at once.
* Changed `F.linear` to use this feature as an example of variable-length FunctionNode.

kmaehashi:
This PR is ready for review (removed RFC)

niboshi:
It's just assigning variable names for error messages, but with the name `args`, I feel like it's something more essential.
How about `argname`?

niboshi:
jenkins, test this please

niboshi:
jenkins, test this please

okuta:
LGTM!

PR #4682, opened 2018-04-25
add ignore_names option to load_npz

yuyu2172:
This option was supported by `NpzDeserializer`, but not by `load_npz`.


okuta:
LGTM!

Issue #4272, opened 2018-02-05
pytest.skip() does not work with testing.condition


niboshi:
Test code:
```
import unittest                 
import pytest                   
from chainer.testing import condition                           

class TestAaa(unittest.TestCase):                               

    @condition.retry(32)        
    def test_aaa(self):         
        pytest.skip('aaa')      
        assert False            
```

Result:
```
________________________________________________________ TestAaa.test_aaa ________________________________________________________
chainer/testing/condition.py:74: in wrapper
    fail()
chainer/testing/condition.py:51: in fail
    instance.fail(msg)
E   AssertionError: 
E   Fail: 32, Success: 0
E   
E   The first error message:
E   Traceback (most recent call last):
E     File "/data/work/w/repos/chainer/chainer/testing/condition.py", line 60, in <lambda>
E       lambda: f(ins, *args[1:], **kwargs),
E     File "/data/work/w/repos/chainer/tests/chainer_tests/test_aaa.py", line 9, in test_aaa
E       pytest.skip('aaa')
E     File "/home/niboshi/anaconda/anaconda3/lib/python3.6/site-packages/_pytest/outcomes.py", line 72, in skip
E       raise Skipped(msg=msg)
E   Skipped: aaa
==================================================== 1 failed in 0.65 seconds ====================================================
```

After removing `retry`:
```
=================================================== 1 skipped in 0.60 seconds ====================================================
```

niboshi:
Currently `testing.condition` makes temporary test suites (`unittest.TestSuite`) in the loop, which seemingly does not work with PyTest.

toslunar:
`pytest.skip()` does not works with `testing.parameterize`.  https://travis-ci.org/chainer/chainer/jobs/370873926

To reproduce:
```
@testing.parameterize(
        {'a': 3}
)
class TestFail(unittest.TestCase):
        def test_a(self):
                pytest.skip()
                self.assertEqual(1, 0)
```

kmaehashi:
FYI, we have a discussion regarding use of pytest specific features: https://github.com/chainer/chainer/issues/4557

kmaehashi:
Note that #4685 does not work with `pytest.skip`, as it uses own exception type (`Skipped`) instead of `unittest.SkipTest`.

kmaehashi:
Closing as we decided not to use `pytest.skip`.
I'll update the contribution guide later.

niboshi:
[...]

PR #4685, opened 2018-04-26
Support skipping tests decorated by condition or parametize

kmaehashi:
Allow individual test cases to be skipped by `self.skipTest`.
Related to #4272.


okuta:
LGTM!

Issue #2748, opened 2017-05-18
Wrap Link.__call__


niboshi:
Currently users directly override `Link.__call__`.
I propose letting users override, say, `Link.call` instead, and call it from `Link.__call__`. It's just like how `Function.__call__` works.

This will introduce several benefits (At least I can think of):
* Link hooks
* Links will be able to be represented hierarchically in graph visualization. (See #2215)

Some candidate method names: `call`, `__call_link__`, `define`, `define_link`, any other?

Note: This will not break backward compatibility, as existing code (overriding `Link.__call__`) will just work, but with new features (e.g. link hooks) disabled.

beam2d:
It would be better to provide a context manager method or a decorator that inserts the pre/post processes. It will be used like follows.
```py
def __call__(self, x, ...):
    with self.forward_scope():  # run preprocess in __enter__, and postprocess in __exit__
        h = F.relu(...)
        ...
```
or
```py
@Link.forward_method
def __call__(self, x, ...):
    h = F.relu(...)
    ...
```
There are some benefits:

- It is straight-forward to understand the actual interface of the link implementation.
- There is no extra effort needed to support IDE correctly and provide a documentation that reflects the actual interface.
- Users can provide multiple methods that do forward computations for different purposes (e.g. one for computing a loss value and another for inference).

See #2657, too.

mitmul:
"Link hooks" may make ONNX export feature I'm trying to create as Function hooks more easier especially for LSTM-like modules which have multiple Functions inside.

beam2d:
We need to list up possible applications for this feature, then we can start discussing what the interface should be. Any ideas? Here is a list of ideas proposed in the offline discussions (including those that have alread appeared in this issue):
- Graph visualization improvement
- Link hooks
  - ONNX export code injection in Link level (e.g. LSTM)
  - Link-level profiling code injection

niboshi:
I, @bkvogel and @mitmul discussed the interface.
While @beam2d's suggestions above have some advantages, we have a concern that it may be difficult for users (especially those who are novice in Python language) to understand and track the code behind `Link`.
In the alternative (basically it's the same as the one I proposed above), usual user-defined links look just simple (the method name is borrowed from `nn.Module` in PyTorch), and complicated things are required only when multiple forward methods are defined.

Usual user-defined link would look like:
```py
def SomeLink(chainer.Link):
    def forward(self, x):
        h = F.relu(x)
        ....
````

[...]

PR #4701, opened 2018-05-01
[WIP] Add spectral normalization

tkerola:
This PR implements spectral normalization (https://arxiv.org/abs/1802.05957).
I just adapted the code from https://github.com/pfnet-research/sngan_projection/ and added tests, so credit is due to the original author.

- [x] Add spectral normalization for `Convolution2D`
- [ ] Add spectral normalization for `ConvolutionND`
- [ ] Add spectral normalization for `EmbedID`
- [ ] Add spectral normalization for `Linear`

One issue is that due to `self.u` being updated during each forward pass, writing clean forward tests is a bit cumbersome. Currently, I am just resetting `self.u` before each forward pass during testing. This, however, cannot be done for backward tests. Please let me know if you have any idea on how the function can be redesigned to better fit into Chainer's test framework.

### DIFF
--- a/chainer/links/__init__.py
+++ b/chainer/links/__init__.py
@@ -37,6 +37,7 @@ from chainer.links.connection.n_step_rnn import NStepRNNTanh  # NOQA
 from chainer.links.connection.parameter import Parameter  # NOQA
 from chainer.links.connection.peephole import StatefulPeepholeLSTM  # NOQA
 from chainer.links.connection.scale import Scale  # NOQA
+from chainer.links.connection.spectral_normalization_convolution_2d import SpectralNormalizationConvolution2D  # NOQA
 from chainer.links.connection.tree_lstm import ChildSumTreeLSTM  # NOQA
 from chainer.links.connection.tree_lstm import NaryTreeLSTM  # NOQA
 from chainer.links.connection.zoneoutlstm import StatefulZoneoutLSTM  # NOQA
--- a/chainer/links/connection/spectral_normalization_convolution_2d.py
+++ b/chainer/links/connection/spectral_normalization_convolution_2d.py
@@ -1,0 +1,144 @@
+import numpy as np
+
+import chainer
+from chainer import cuda
+import chainer.functions as F
+from chainer.functions.connection import convolution_2d
+from chainer.links.connection.convolution_2d import Convolution2D
+
+
+class SpectralNormalizationConvolution2D(Convolution2D):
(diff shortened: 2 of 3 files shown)


toslunar:
On backward tests, how about this?
```
    def forward(self, x):
        self.reset_u()
        return self.link(x)

    def check_backward(self, x_data, y_grad):
            self.forward, x_data, y_grad,
            eps=1e-2, **self.check_backward_options)

    # @condition.retry(3)
    def test_backward_cpu(self):
        self.check_backward(self.x, self.gy)
```

tkerola:
Thank you! That seems like a good way of implementing backward tests.

I would also like to discuss another issue before continuing with this PR.
Currently, the idea is to implement spectral normalization (SN) for the four set of links that were used in the author's original implementation (Conv2D, ConvND, Linear, EmbedID). While this works, I doesn't scale well if we want to add SN to another type of link in the future, since e.g. the code for `Convolution2D` and `SpectralNormalizationConvolution2D` are very similar, expect the weight normalization part, which goes against the DRY principle. In a sense, spectral normalization is a technique independent of the actual `FunctionNode` that uses the normalized weights.

It would be cleaner if we instead implemented SN as either a
* link hook. (If implemented in the future, see discussion in #2748). In this case, the weight of a link could be normalized with SN before the link calls the underlying `FunctionNode`.
* optimizer hook. Similar to weight decay, we could implement SN as a normalization of the weights before the optimizer update is done.
The disadvantage to the above is that it would not be possible to support the optional learnable parameter `gamma` in SN. Also, with the optimizer update variant, the implementation would be slightly different from the paper.

A way of supporting `gamma` while still making SN independent from other links could be to implement SN as a link that does not operate on a `Variable`, but rather a `Parameter` of another link.
The use case could be as below.
```
sn = L.SpectralNormalization()
conv = L.Convolution2D(...)

x = chainer.Variable(...)
y = sn(conv)(x)
```  
where `sn.__call__()` takes a link as input and will return the link with modified parameter `W`.

What do you think? Any other ideas of how to fit SN cleanly into the Chainer framework is welcome.

toslunar:
I'm interested in how to implement SN in Chainer since I had similar question when I added Noisy Networks to ChainerRL.  At that time, I wasn't able to come up with a better way to replace a `Parameter` with an output of a `Link`.
https://github.com/chainer/chainerrl/blob/7df91420bac63ba11353f09f728a59ff6d4d15cc/chainerrl/links/noisy_linear.py

PR #4706, opened 2018-05-03
Add `--test` option to seq2seq example

arisliang:
There's already a device allocation above the removed one.

### DIFF
--- a/examples/seq2seq/seq2seq.py
+++ b/examples/seq2seq/seq2seq.py
@@ -226,6 +226,8 @@ def main():
                         'with validation dataset')
     parser.add_argument('--out', '-o', default='result',
                         help='directory to output the result')
+    parser.add_argument('--test', dest='test', action='store_true')
+    parser.set_defaults(test=False)
     args = parser.parse_args()
 
     # Load pre-processed dataset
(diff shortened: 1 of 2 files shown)


kmaehashi:
Thanks!
Jenkins, test this please.

arisliang:
What's the best way to create multiple pull requests? I pushed another commit of the same file. It automatically combined in the previous pull request.

Should I change the name of the pull request to reflect it contains two merge? Or someway to separate the PR (but they're in the same file). What's the better way to do this?

kmaehashi:
I think `--test` option is not needed for seq2seq example, as data is given from external file (i.e., unlike other examples using built-in dataset functions, users can control the size of data.)
If you need to experiment with small dataset, you can just use the first N lines of the source/target sentences. You can even create your own small dataset in the same format, like this: https://github.com/chainer/chainer-test/tree/master/data/seq2seq

arisliang:
> use the first N lines of the source/target sentences

That's what I did with --test option. True, user can manipulate external data file, but it's just easier to toggle it with a switch.

arisliang:
`Could you please split the PR?` To be honest, I'm not sure how to do it through github ui. Is it something I'd have to go through command line? If that's the case, maybe I'll try recreate the PR, might be easier.

kmaehashi:
You cannot do that via Github UI. Feel free to close and recreate PRs!

kmaehashi:
@arisliang ping? (I can takeover this fix if you don't have time.)

arisliang:
@kmaehashi Yes, please, thank you.

kmaehashi:
I'll revisit after #4782

kmaehashi:
Redundant device allocation has been fixed in #4848.
As for ``--test`` argument, we should reconsider it as we now have `TextDataset` which does not load everything on memory.

Issue #3408, opened 2017-09-17
Feature request: logarithmic gamma function


muupan:
The logarithmic gamma function is missing in Chainer. It is required when parameterizing a Beta distribution.

Tensorflow implementation: https://www.tensorflow.org/api_docs/python/tf/lgamma
Pytorch implementation: https://github.com/rachtsingh/lgamma (unofficial)

muupan:
To implement backward computation of it, digamma and polygamma functions are required. https://github.com/cupy/cupy/issues/865

muupan:
bump

mitmul:
lgamma will come with many probability distribution implementations in this PR #4678:
https://github.com/chainer/chainer/pull/4678/files#diff-714b726ad7916530f4c83de6e33e3e9b

muupan:
Great, but its backward computation (digamma) is implemented only in cpu using scipy. I think it may be good to keep this issue open until its full gpu support. https://github.com/YoshikawaMasashi/chainer/blob/fb71a98cece3d23b73e90ecae4bc024c00dee60f/chainer/functions/math/digamma.py

beam2d:
This issue has been fixed via #4720.

PR #4720, opened 2018-05-07
logarithmic gamma function

YoshikawaMasashi:
#3408 
I implemented logarithmic gamma(lgamma) function. (along with digamma and polygamma)

As mentioned in this issue, I have not implemented effective calculation for with gpu.

### DIFF
--- a/chainer/functions/__init__.py
+++ b/chainer/functions/__init__.py
@@ -188,6 +188,7 @@ from chainer.functions.math.cumsum import Cumsum  # NOQA
 from chainer.functions.math.det import batch_det  # NOQA
 from chainer.functions.math.det import BatchDet  # NOQA
 from chainer.functions.math.det import det  # NOQA
+from chainer.functions.math.digamma import digamma  # NOQA
 from chainer.functions.math.erf import erf  # NOQA
 from chainer.functions.math.erfc import erfc  # NOQA
 from chainer.functions.math.exponential import exp  # NOQA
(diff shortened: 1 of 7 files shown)


YoshikawaMasashi:
`digamma` and `polygamma` don't exist in `numpy` and `math. Therefore I implemented like this for now.

Should I implement numerical calculation of digamma like scipy?
Or is there any good solution?

beam2d:
Could you make the tests being skipped when scipy is not available, and add a test to check if an appropriate error is raised on such environments? The skip can be done with `@chainer.testing.with_requires('scipy')`: https://github.com/chainer/chainer/blob/master/chainer/testing/helper.py#L20-L46.

beam2d:
> Should I implement numerical calculation of digamma like scipy?
I think it's good to provide such a simple CUDA implementation.

YoshikawaMasashi:
I implemented digamma and polygamma with gpu.

However, this implementation is same as `cupyx.special.digamma` and `cupyx.special.polygamma`. cupy/cupy#1232
So I want to use `cupyx.special.digamma` and `cupyx.special.polygamma` directly.
Can I do?

beam2d:
I'm amazed by your coding speed! Of course, it's better to avoid duplication with cupy, so it's totally fine to use `cupy` functions for digamma/polygamma.

beam2d:
cupy/cupy#1232 has been merged. Could you fix the GPU code to use them?

toslunar:
Jenkins, test this please.

PR #4717, opened 2018-05-07
Raise an error when CUDA has been initialized before MultiprocessParallelUpdater is initialized

mitmul:
This PR resolves #4709.

PR #4726, opened 2018-05-08
separate MultiprocessParallelUpdater tests

kmaehashi:
Fix in #4717 changed MPU to raise an error when CUDA is used in the parent process.
This PR changes the MPU tests to run in separate process to avoid test failure when #4717  got merged.

(discussion regarding this change: https://github.com/chainer/chainer/pull/4368#issuecomment-366934262)


mitmul:
@kmaehashi Thanks for the PR! Could you check the Jenkins tests? It fails due to ` AssertionError: chainer/tests/chainer_tests/training_tests/updaters_tests/snippets/raw_array.py is not runnable.`

kmaehashi:
Thanks, fixed!

mitmul:
LGTM

PR #4730, opened 2018-05-08
Add LinkHook

mitmul:
This PR adds `LinkHook` which enables to add `forward_preprocess` and `forward_postprocess` hooks to `chainer.Link`s. To enable this, all the `__call__` methods of provided `Link`s in Chainer are renamed to `forward`. `Link.__call__` is still available and it calls `forward_preprocess` first, then call `Link.forward`, then call `forward_postprocess`. Almost all design is come from `FunctionHook`.

### DIFF
--- a/chainer/link.py
+++ b/chainer/link.py
@@ -108,11 +108,11 @@ class Link(object):
                       self.b = chainer.Parameter(
                           initializers.Zero(), (n_out,))
 
-              def __call__(self, x):
+              def forward(self, x):
                   return F.linear(x, self.W, self.b)
 
        This example shows that a user can define arbitrary parameters and use
-       them in any methods. Links typically implement the ``__call__``
+       them in any methods. Links typically implement the ``forward``
        operator, although they can also provide other methods to implement the
        forward propagation.
 
@@ -129,6 +129,8 @@ class Link(object):
 
     """
 
+    _local_link_hooks = None
+
     def __init__(self, **params):
         self._params = set()
         self._persistent = set()
@@ -141,6 +143,41 @@ class Link(object):
             # Note: deprecation warning will be raised in add_param
             shape, dtype = _ensure_shape_dtype(value)
             self.add_param(name, shape, dtype=dtype)
+
+    @property
+    def local_link_hooks(self):
+        """Ordered dictionary of registered link hooks.
+
+        Contrary to ``chainer.thread_local.link_hooks``,
+        which registers its elements to all functions,
+        Function hooks in this property is specific to this function.
+
+        """
+        if self._local_link_hooks is None:
+            self._local_link_hooks = collections.OrderedDict()
+        return self._local_link_hooks
+
+    @property
+    def _n_local_link_hooks(self):
+        return (0 if self._local_link_hooks is None
+                else len(self._local_link_hooks))
+
+    def __call__(self, *args):
+        hooks = chainer.get_link_hooks()
+        if self._n_local_link_hooks > 0:
+            hooks = collections.OrderedDict(hooks)
+            hooks.update(self.local_link_hooks)
+        hooks = hooks.values()  # avoid six for performance
+
+        for hook in hooks:
+            hook.forward_preprocess(self, *args)
+
+        out = self.forward(*args)
+
+        for hook in hooks:
+            hook.forward_postprocess(self, *args)
+
+        return out
 
     @property
     def xp(self):
@@ -631,7 +668,7 @@ Assign a Parameter object directly to an attribute within a \
                                 None, 64, 3, 1, 1, nobias=True)
                             self.bn = L.BatchNormalization(64)
 
-                    def __call__(self, x):
+                    def forward(self, x):
                         return F.relu(self.bn(self.conv(x)))
 
                 net = ConvBNReLU().repeat(16, mode='init')
@@ -760,7 +797,7 @@ class Chain(Link):
                       self.layer2 = L.Linear(n_hidden, n_hidden)
                       self.layer3 = L.Linear(n_hidden, n_out)
 
-              def __call__(self, x):
+              def forward(self, x):
                   # Forward propagation
                   h1 = F.relu(self.layer1(x))
                   h2 = F.relu(self.layer2(h1))
@@ -768,7 +805,7 @@ class Chain(Link):
 
        Child links are registered via the assignment within a
        ``with self.init_scope():`` block. The forward propagation is often
-       implemented as the ``__call__`` operator as the above example, though
+       implemented as the ``forward`` operator as the above example, though
        it is not mandatory.
 
     Args:
--- a/chainer/link_hook.py
+++ b/chainer/link_hook.py
@@ -1,0 +1,139 @@
+import chainer
+
+
+class LinkHook(object):
+    """Base class of hooks for Links.
+
+    :class:`~chainer.LinkHook` is a callback object
+    that is registered to a :class:`~chainer.Link`.
+    Registered link hooks are invoked before calling
+    :meth:`forward` methodof each link.
+
+    Link hooks that derive :class:`LinkHook` are required
+    to implement two methods:
+    :meth:`~chainer.FunctionHook.forward_preprocess` and
+    :meth:`~chainer.FunctionHook.forward_postprocess`.
+    By default, these methods do nothing.
+
+    Specifically, when :meth:`~chainer.Link.__call__`
+    method of some function is invoked,
+    :meth:`~chainer.LinkHook.forward_preprocess`
+    (resp. :meth:`~chainer.LinkHook.forward_postprocess`)
+    of all function hooks registered to this function are called before
+    (resp. after) :meth:`forward` method of the link.
+
+    There are two ways to register :class:`~chainer.LinkHook`
+    objects to :class:`~chainer.Link` objects.
+
+    First one is to use ``with`` statement. Link hooks hooked
+    in this way are registered to all functions within ``with`` statement
+    and are unregistered at the end of ``with`` statement.
+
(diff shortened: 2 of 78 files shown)


corochann:
I came up in mind for the benefit of `LinkHook`.

I think we can get specified layer's output using `LinkHook` for the usual Chain class as well, link `PickableSequentialChain`. This is good for the research to analyze each layer's behavior.
https://github.com/chainer/chainercv/blob/master/chainercv/links/model/pickable_sequential_chain.py

chainer/link_hook.py
```diff
@@ -0,0 +1,139 @@
+import chainer
+
+
+class LinkHook(object):
+    """Base class of hooks for Links.
+
+    :class:`~chainer.LinkHook` is a callback object
+    that is registered to a :class:`~chainer.Link`.
+    Registered link hooks are invoked before calling
+    :meth:`forward` methodof each link.
```
crcrpar: nit-picking: "methodof" -> "method of"

toslunar:
User-defined Link that supports this feature cannot merely run in Chainer v4, right?

toslunar:
Sorry, we can with something like `if hasattr(super(MyLink), '__call__')`.

chainer/link_hook.py
mitmul: @crcrpar Thanks! I'll fix it.

chainer/link.py
```diff
@@ -142,6 +144,41 @@ def __init__(self, **params):
...
+
+    def __call__(self, *args, **kwargs):
+        hooks = chainer.get_link_hooks()
+        if self._n_local_link_hooks > 0:
+            hooks = collections.OrderedDict(hooks)
+            hooks.update(self.local_link_hooks)
+        hooks = hooks.values()  # avoid six for performance
+
+        for hook in hooks:
+            hook.forward_preprocess(self, *args, **kwargs)
```
niboshi: How about passing `args` and `kwargs` without `*`s for future extensibility?

chainer/link.py
```diff
@@ -142,6 +144,41 @@ def __init__(self, **params):
...
+
+        out = self.forward(*args, **kwargs)
+
+        for hook in hooks:
+            hook.forward_postprocess(self, *args, **kwargs)
```
niboshi: How about including `out`?

chainer/link_hook.py
```diff
@@ -0,0 +1,139 @@
...
+        pass
+
+    # forward
+    def forward_preprocess(self, link, *args, **kwargs):
+        """Callback function invoked before :meth:`forward`.
+
+        Args:
+            link(~chainer.Link): Link object to which
+                the link hook is registered.
+            in_data(tuple):
```
niboshi: Fix docstring

chainer/link_hook.py
```diff
@@ -0,0 +1,139 @@
...
+        """
+        pass
+
+    def forward_postprocess(self, link, *args, **kwargs):
+        """Callback function invoked after link propagation.
+
+        Args:
+            link(~chainer.Link): Link object to which
+                the link hook is registered.
+            in_data(tuple):
```
niboshi: Ditto.

chainer/link_hook.py
```diff
@@ -0,0 +1,139 @@
...
+               Input argument of :meth:`forward`.
+        """
+        pass
+
+    def forward_postprocess(self, link, *args, **kwargs):
+        """Callback function invoked after link propagation.
```
niboshi: "link propagation" seems an odd wording.

chainer/link_hooks/timer.py
```diff
@@ -0,0 +1,132 @@
+import sys
+import time
+
+import numpy
+
+from chainer.backends import cuda
+from chainer import link_hook
+
+
+class ForwardTimerHook(link_hook.LinkHook):
```
niboshi: Is the word `Forward` needed? Shouldn't it be limited to `forward` method?

niboshi: Replace "function" with "link" in this class.

niboshi:
I took over this PR.

chainer/link_hook.py
```diff
@@ -0,0 +1,139 @@
...
+        """Callback function invoked when a link hook is deleted
+
+        Args:
+            link(~chainer.Link): Link object to which
+                the link hook is deleted.
+        """
+        pass
+
+    # forward
+    def forward_preprocess(self, link, *args, **kwargs):
```
niboshi: It shouldn't be limited to `forward`.

chainer/link.py
```diff
@@ -142,6 +144,41 @@ def __init__(self, **params):
...
+        if self._local_link_hooks is None:
+            self._local_link_hooks = collections.OrderedDict()
+        return self._local_link_hooks
+
+    @property
+    def _n_local_link_hooks(self):
+        return (0 if self._local_link_hooks is None
+                else len(self._local_link_hooks))
+
+    def __call__(self, *args, **kwargs):
```
niboshi: Add TODO to support hooks in arbitrary functions, not limited to `forward`.

chainer/link_hook.py
niboshi: I re-checked https://github.com/chainer/chainer/issues/2748#issuecomment-383836796, and it seems we had the intention that arbitrary user-defined methods (other than `forward`) can also be considered as a kind of "forward" methods.

chainer/link_hook.py
```diff
@@ -0,0 +1,139 @@
...
+
+class LinkHook(object):
+    """Base class of hooks for Links.
+
+    :class:`~chainer.LinkHook` is a callback object
+    that is registered to a :class:`~chainer.Link`.
+    Registered link hooks are invoked before calling
+    :meth:`forward` methodof each link.
+
+    Link hooks that derive :class:`LinkHook` are required
```
niboshi: I think it's not required to implement the callback functions.

niboshi:
I think we should tell which "forward" function is called to the callback.
It may be enough to pass a function name (`str`, in this PR it's always `'forward'`.) as an additional argument.

But I think it's better to wrap such arugments as some object, to avoid breaking user code in the future.

niboshi:
Discussed with @beam2d:
* Use `forward` in callback names like `forward_preprocess`, as we use `forward` in the method name `Link.forwrad`. However not in the concrete hook name: `ForwardTimerHook` -> `TimerHook`, as it might support timing for other pairs of callbacks in the future (such as backward, serialization, etc.).
* Wrap arguments in a single object.

niboshi:
I believe I've finished major fixes.

One remaining thing to consider: `link_hooks.TimerHook.call_history` could be a list (perhaps iterable) of special objects rather than fixed-length tuples. It's just like arguments of the callbacks (`_CallbackArgs`) and provides flexibility for future extension. For example, it could provide record of forward method name, `xp`, etc, not just link name and elapsed time.

niboshi:
Ready for review, hopefully.

niboshi:
I have two concerns regarding current `link_hook.TimerHook`.

(1) I've come to think a time profiler for links should essentially support hierarchical call statistics (suc as both cumulative elapsed time including child links and exlusive elapsed time). Appropriate interface would be different from what it is now (which came from `function_hooks.TimerHook`).

(2) We would also want to support backward profiling. For that, the "hook" would have to implement both `LinkHook` and `FunctionHook`. We should re-think where it should be put into. Maybe we should provide a base-class implementation that implement both hooks and support link backward callbacks.

For now, I would like to dismiss the timer hook as the first link hook implementation.
The problem then is what we should implement as an alternative.

niboshi:
Discussed with @beam2d and concluded to remove `link_hook.TimerHook` for now.
I will make an issue later.

chainer/link.py
```diff
@@ -728,6 +775,39 @@ def count_params(self):
             size += param.size
         return size
 
+    def add_hook(self, hook, name=None):
+        """Registers a link hook.
+
+        Args:
+            hook (~chainer.LinkHook): Link hook to be registered.
+            name (str): Name of the link hook. The name must be unique
```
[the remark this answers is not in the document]
niboshi: If multiple hooks of the same type is to be registered, you need to specify this argument. It's the same as `FunctionHook`.

niboshi:
Thank you. Fixed following the review. PTAL

chainer/link.py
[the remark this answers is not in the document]
bkvogel: Thanks for the explanation. I think it seems reasonable.

bkvogel:
I think it looks good other than the TODO to add an example in the `link_hook` doc.

bkvogel:
@niboshi Do you prefer to have the PR merged as-is and add an example in `link_hook` doc as a later PR? Or do you prefer to add an example before merging?

niboshi:
@bkvogel 
Oh, sorry I thought it will be another PR.
Currently we don't have any link hook implementation, so we will need to define some simple link hook in example code, such as one which simply logs each callback invocation.
Do you think it's enough? Then I will work on that.

bkvogel:
Sure, I think that will be enough. If you can add a simple example like that, it would be nice.

mitmul:
Jenkins, test this please

niboshi:
@bkvogel 
Discussed with @mitmul, and we are thinking example code should be another PR.
The reason is that we want more complicated example than that of function hook, to show hierarchical nature of this hook, for example.

niboshi:
Btw, calling `link.forward` will not trigger link hooks. Is it ok?

bkvogel:
LGTM then. I think it's fine for now, but we can revisit when example code is added as a future PR. I see that Jenkins test `unary_math_function_test.py` failed, but that does not seem related to this PR so I think it is safe to ignore.

PR #4731, opened 2018-05-09
Fix hacking version

kmaehashi:
Our code is currently not compatible with `hacking` 1.1 released on May 9.

https://pypi.org/project/hacking/#history

This is a temporary fix to use old `hacking` for tests.


niboshi:
LGTM!

PR #4732, opened 2018-05-09
Update top sentences in documentation

niboshi:
Current top sentencese in the documentation is a bit too propagandic, and not necessarily accurate.
I replaced it with the sentences from the Chainer web site.


kmaehashi:
LGTM!

PR #4735, opened 2018-05-09
Replace all __call__ methods in Links with forward

mitmul:
By the request from @okuta , I picked up some part of changes from #4730 

**We need to check all the documents and examples carefully.**


niboshi:
This PR fixes #2748.

chainer/link.py
```diff
@@ -142,6 +142,10 @@ def __init__(self, **params):
             shape, dtype = _ensure_shape_dtype(value)
             self.add_param(name, shape, dtype=dtype)
 
+    def __call__(self, *args, **kwargs):
+        out = self.forward(*args, **kwargs)
```
okuta: How about one liner code?

mitmul: forward_preprocess and forward_postprocess will be inserted before and after this line. So I just want to leave it as is.

niboshi: >forward_preprocess and forward_postprocess will be inserted before and after this line. So I just want to leave it as is.

That can be done in that PR.
Each PR should be self-consistent.

mitmul: Why isn't this self-consistent? I don't think it breaks "self-consistency".

niboshi: >forward_preprocess and forward_postprocess will be inserted before and after this line. So I just want to leave it as is.

Doesn't this mean it's imperfect until forward_preprocess and forward_postprocess will be inserted?

okuta: OK, I accept current code.

okuta:
LGTM!

PR #4737, opened 2018-05-09
[WIP] Perform NaN check of output values in context of output array device

kmaehashi:


toslunar:
Are you going to add tests for the issue?  I'll close the PR otherwise.

PR #4740, opened 2018-05-10
Add small examples for serialization feature

mitmul:
- [ ] Clarify what can be serialized, and what can't.


examples/serialization/load.py
```diff
@@ -0,0 +1,28 @@
+import chainer
+import chainer.functions as F
+import chainer.links as L
+
+n_in = 784
+n_units = 100
+n_out = 10
+
+# Network definition
+class MLP(chainer.Chain):
```
niboshi: How about moving the model definition into a separate file?

examples/serialization/load.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def __init__(self, n_in, n_units, n_out):
+        super(MLP, self).__init__()
+        with self.init_scope():
```
niboshi: How about demonstrating `add_persistent`?

niboshi:
Other suggestions:
* User-defined serializable object, not just a link
* Serialization into other formats (e.g. hdf5)

niboshi:
@mitmul Do you have time to work on this PR?

examples/serialization/load.py
mitmul: Thanks. moved

examples/serialization/load.py
mitmul: That's nice idea. I added it.

mitmul:
@niboshi I came back to Chainer development, so restarted working on it. I added some of your suggestions. Could you check it again?

examples/serialization/load.py
```diff
@@ -0,0 +1,27 @@
+import chainer
+import numpy as np
+import model
+
+
+n_in = 784
+n_units = 100
+n_out = 10
```
niboshi: How about including these parameters in `model.py`?
We should demonstrate minimal code repetition is required in save/load scripts.

examples/serialization/load.py
```diff
@@ -0,0 +1,27 @@
...
+
+# Create model object first
```
niboshi: Logical separation of this script is not clear at first glance.
How about separating into functions, like `load_npz_file` and `load_hdf5_file` (or any better names)?

ditto. for `save` script.

examples/serialization/save.py
```diff
@@ -0,0 +1,47 @@
...
+
+    # Load the saved HDF5 using h5py
+    print('--- The list of saved params in model.h5 ---')
+    f = h5py.File('model.h5', 'r')
+    for param_key, param in f.items():
+        sys.stdout.write('{}:'.format(param_key))
+        if isinstance(param, h5py.Dataset):
+            print(' {}'.format(param.shape))
+        else:
+            print('')
```
niboshi: `sys.stdout` and `print` should not be mixed.
How about `print`ing the whole line at once?
```
if ...
    print('{}: {}'.format(param_key, param.shape)
else:
    print('{}:'.format(param_key)
```

niboshi:
LGTM!

PR #4742, opened 2018-05-10
Change the initial avg_var of BatchNormalization to 1

### LINKED ISSUE #879: BatchNormalization's avg_var initialized with zeros can cause division by zero
Currently, `BatchNormalization`'s `avg_var` is initialized with zeros. This will cause division by zero when called it with test=True before test=False, so it might be better we use another values, say, ones, to initialize `avg_var`.

Calling `BatchNormalization` with test=True before test=False can seem unnatural, but it can happen in the reinforcement learning context: an agent selects an action by evaluating its value function or policy represented by a NN, observes consequences, and then (if it has enough experiences) updates its value function or policy.

beam2d:
Fix #879. This PR introduces `initial_avg_mean` and `initial_avg_var` arguments to `BatchNormalization` and `BatchRenormalization`. The default value of `initial_avg_var` is 1 instead of 0, which was used as the initial avg_var in v4. It breaks the compatibility of learned models including these links, though the influence is very limited because the difference in the initial value is negligible after long enough training. This change should make inference at the early stage of training stabler. In particular, inference before the first iteration is made valid with this change.


toslunar:
git will also conflict with #4673

chainer/links/normalization/batch_normalization.py
```diff
@@ -168,3 +185,11 @@ def start_finetuning(self):
 
         """
         self.N = 0
+
+
+def _init_array(initializer, default_value, size, dtype):
+    if initializer is None:
+        initializer = default_value
+    initializer = initializers._get_initializer(initializer)
+    initializer.dtype = dtype
```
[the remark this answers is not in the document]
beam2d: `dtype` attribute is a part of the initializer interface (for example, see https://docs.chainer.org/en/stable/reference/generated/chainer.initializers.generate_array.html#chainer.initializers.generate_array). Maybe it's better to fix the documentation for initializers? (Btw, it does not raise `AttributeError` in general to set any attributes to callables; even function (including lambda) can have a custom attribute)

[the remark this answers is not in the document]
niboshi: Ah, I have missed that statement in `initializers.generate_array`. I understand.

(I thought it would raise `AttibuteError` because `a = object(); a.b = 3` did. It seems `object` is different...)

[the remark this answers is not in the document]
niboshi: But previously, `initializer.dtype` is not particularly meant to be writable (both in existing code and documentation).

Rather than forcing it to be writable, how about adding `dtype` (`default_dtype`?) argument to `initializers.generate_array()`?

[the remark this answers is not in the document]
beam2d: That sounds good. I will add `dtype` option to `generate_array`.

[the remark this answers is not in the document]
beam2d: fixed

chainer/initializers/__init__.py
```diff
@@ -29,14 +29,14 @@ def generate_array(initializer, shape, xp):
...
 
     Returns:
         numpy.ndarray or cupy.ndarray: An initialized array.
 
     """
-    dtype = numpy.float32
-    if hasattr(initializer, 'dtype') and initializer.dtype is not None:
-        dtype = initializer.dtype
+    if dtype is None:
+        dtype = getattr(initializer, 'dtype', numpy.float32)
```
niboshi: How about prefering `initializer.dtype` if both specified?
That's why I suggested `default_dtype` as the name of the argument.

beam2d: Hmm, I feel the argument should provide the way to override other configurations.

niboshi: I think that could be confusing for initializer side.
Initializers with concrete `dtype`s should expect arrays with that `dtype`.

If `initializer.dtype` is just a default dtype, it should have been named as `default_dtype`.

niboshi: Actually, initializer implementations have `if self.dtype is not None: assert array.dtype == self.dtype` in their `__call__` method.

beam2d: Hmmm, ok, I understand the situation, but I'm a little bit against using the name `default_dtype` because it conflicts with the "default dtype" concept introduced in #4510. Another solution is just raising an error when these two (`dtype` and `initializer.dtype`) conflict.

niboshi: That sounds good.

chainer/links/normalization/batch_normalization.py
```diff
@@ -204,9 +223,12 @@ def __init__(self, size=None, decay=0.9, eps=2e-5, dtype=None,
             self._initialize_params(size)
 
     def _initialize_params(self, shape):
-        self.avg_mean = numpy.zeros(shape, dtype=self._dtype)
+        dtype = self._dtype
+        self.avg_mean = _init_array(self._initial_avg_mean, 0, shape, dtype)
+        del self._initial_avg_mean
```
niboshi: How about assigning `None`?
I think it's not a good practice to add/remove this kind of attributes in ad-hoc manner.

chainer/links/normalization/batch_normalization.py
```diff
@@ -204,9 +223,12 @@ def __init__(self, size=None, decay=0.9, eps=2e-5, dtype=None,
...
         self.register_persistent('avg_mean')
-        self.avg_var = numpy.zeros(shape, dtype=self._dtype)
+        self.avg_var = _init_array(self._initial_avg_var, 1, shape, dtype)
+        del self._initial_avg_var
```
niboshi: ditto.

niboshi:
LGTM

PR #4743, opened 2018-05-10
Fix incorrect attributes and incorrect indentation in pooling tests

niboshi:
* Fix incorrect test attributes.
* `chainer.using_config('use_cudnn')` was not applied correctly.


kmaehashi:
LGTM!

PR #4744, opened 2018-05-10
Fix a stability issue in `LayerNormalization`

anaruse:
This PR aims to fix sort of divide-by-zero issue in `LayerNormalization`.


toslunar:
IMO, testing backward with large eps (as in #4190) would make it clear that the PR is a bugfix.

bkvogel:
Jenkins, test this please.

PR #4746, opened 2018-05-11
Support copying 1-dim array to ideep

kmaehashi:
Currently `to_intel64` does not copy 1-dim arrays to iDeep.
As a result, bias parameter of the Linear layer is not copied to iDeep.
This causes memory copy overhead every time Linear layer is evaluated.

`mnist.mnist.MLPPredict.time_predict` benchmark results over this PR is as follows:
https://github.com/chainer/chainer-benchmark/blob/master/benchmarks/mnist/mnist.py#L92

Before applying this PR:

```
=========== ========== ============
--             units / batchsize   
----------- -----------------------
  backend    1000 / 1   1000 / 100 
=========== ========== ============
 cpu-ideep   971.01μs     1.40ms   
=========== ========== ============
```

After applying this PR:

```
=========== ========== ============
--             units / batchsize   
----------- -----------------------
  backend    1000 / 1   1000 / 100 
=========== ========== ============
 cpu-ideep   830.86μs     1.29ms   
=========== ========== ============
```

### DIFF
--- a/chainer/backends/intel64.py
+++ b/chainer/backends/intel64.py
@@ -87,7 +87,7 @@ def should_use_ideep(level):
 
 
 def inputs_all_ready(inputs, supported_ndim=(2, 4)):
-    """Checks if input arrays are supported for ideep optimization.
+    """Checks if input arrays are supported for the iDeep primitive.
 
     Information to be checked includes array types, dimesions and data types.
     The function checks ``inputs`` info and ``supported_ndim``.
@@ -96,8 +96,7 @@ def inputs_all_ready(inputs, supported_ndim=(2, 4)):
         inputs (sequence of arrays or variables``):
             Inputs to be checked.
         supported_ndim (tuple of ints):
-            Supported ndim values.
-            iDeep supports array dimension in either 2 or 4 only.
+            Supported ndim values for the iDeep primitive.
 
     Returns:
         bool: ``True`` if all conditions meet.
--- a/chainer/backends/intel64.py
+++ b/chainer/backends/intel64.py
@@ -87,18 +87,18 @@ def should_use_ideep(level):
 
 
 def inputs_all_ready(inputs, supported_ndim=(2, 4)):
-    """Checks if input arrays are supported for the iDeep primitive.
+    """Checks if input arrays are supported for an iDeep primitive.
 
-    Before calling iDeep primitives (e.g., ``ideep4py.linear.Forward``), you
+    Before calling an iDeep primitive (e.g., ``ideep4py.linear.Forward``), you
     need to make sure that all input arrays are ready for the primitive by
     calling this function.
     Information to be checked includes array types, dimesions and data types.
     The function checks ``inputs`` info and ``supported_ndim``.
 
     Inputs to be tested can be any of ``Variable``, ``numpy.ndarray`` or
-    ``ideep4py.mdarray``. However, all inputs to the iDeep primitive must be
-    ``ideep4py.mdarray``. Callers of iDeep primitives are responsible to
-    convert all inputs as ``ideep4py.mdarray``.
+    ``ideep4py.mdarray``. However, all inputs to iDeep primitives must be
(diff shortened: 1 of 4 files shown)


chainer/link.py
```diff
@@ -416,8 +416,10 @@ def to_intel64(self):
             value = d[name]
             if isinstance(value, cuda.ndarray):
                 value = value.get()  # to numpy.ndarray
-            if (isinstance(value, numpy.ndarray) and
-                    intel64.inputs_all_ready((value,))):
+            if (isinstance(value, numpy.ndarray) and value.ndim in (1, 2, 4)):
```
niboshi: Why not use `intel64.inputs_all_ready`?

kmaehashi: `inputs_all_ready` is designed for testing if inputs are ready for iDeep primitives like `ideep4py.linear.Forward`. On the other hand, in this code we want to test if inputs can be converted to iDeep mdarray.

I think [`ideep4py.array`](https://github.com/intel/ideep/blob/bfdec56a9ecf21c85ae19c61e7c5b8554c56d12f/python/ideep4py/__init__.py#L35-L56) should be responsible for checking ndim is valid for mdarray, but such check is not implemented (python process dies if invalid ndim is passed to it).

Intel guy told me that iDeep is going to support arbitrary dimensions in the future version of iDeep. So, I'm considering this fix as a temporary workaround until iDeep is fixed.

niboshi: >`inputs_all_ready` is designed for testing if inputs are ready for iDeep primitives like `ideep4py.linear.Forward`.

OK. Please document it in `inputs_all_ready`.

>Intel guy told me that iDeep is going to support arbitrary dimensions in the future version of iDeep. So, I'm considering this fix as a temporary workaround until iDeep is fixed.

Also, please write a TODO about this.

kmaehashi: Thanks, fixed.

chainer/backends/intel64.py
```diff
@@ -87,17 +87,24 @@ def should_use_ideep(level):
 
 
 def inputs_all_ready(inputs, supported_ndim=(2, 4)):
-    """Checks if input arrays are supported for ideep optimization.
+    """Checks if input arrays are supported for the iDeep primitive.
```
niboshi: "an iDeep primitive"? or "iDeep primitives"?

chainer/backends/intel64.py
```diff
@@ -87,17 +87,24 @@ def should_use_ideep(level):
...
 
+    Before calling iDeep primitives (e.g., ``ideep4py.linear.Forward``), you
+    need to make sure that all input arrays are ready for the primitive by
```
niboshi: "iDeep primitives" -> "an iDeep primitive", OR "the primitive" -> "the primitives"

chainer/backends/intel64.py
```diff
@@ -87,17 +87,24 @@ def should_use_ideep(level):
...
+    calling this function.
     Information to be checked includes array types, dimesions and data types.
     The function checks ``inputs`` info and ``supported_ndim``.
 
+    Inputs to be tested can be any of ``Variable``, ``numpy.ndarray`` or
+    ``ideep4py.mdarray``. However, all inputs to the iDeep primitive must be
```
niboshi: the iDeep primitive -> "iDeep primitives"

chainer/backends/intel64.py
```diff
@@ -87,17 +87,24 @@ def should_use_ideep(level):
...
+    ``ideep4py.mdarray``. Callers of iDeep primitives are responsible to
+    convert all inputs as ``ideep4py.mdarray``.
```
niboshi: "responsible of converting all inputs to"

kmaehashi:
Thanks for comments, fixed accordingly.

niboshi:
Please fix `test_variable.py` as well.

kmaehashi:
Fixed tests.

niboshi:
jenkins, test this please

niboshi:
LGTM!

PR #4738, opened 2018-05-10
Fix regression in bilinear

toslunar:
close #4736.

PR #4752, opened 2018-05-11
Replace ternary numpy.einsum

toslunar:
Merge #4738 first.

Ternary numpy.einsum (without `optimize` option, which is not available in old numpy) is slow.  Do some optimization manually.

### DIFF
--- a/chainer/functions/connection/bilinear.py
+++ b/chainer/functions/connection/bilinear.py
@@ -81,7 +81,12 @@ class BilinearFunction(function_node.FunctionNode):
 
         xp = cuda.get_array_module(*inputs)
         if xp is numpy:
-            y = numpy.einsum('ij,ik,jkl->il', e1, e2, W)
+            # optimize: y = numpy.einsum('ij,ik,jkl->il', e1, e2, W)
+            y = numpy.einsum(
+                'ij,ijl->il',
+                e1,
+                numpy.einsum('ik,jkl->ijl', e2, W)
+            )
         else:
             i_len, j_len = e1.shape
             k_len = e2.shape[1]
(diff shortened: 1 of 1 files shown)


hvy:
Could you please rebase it on the master now that #4738 is merged?

toslunar:
Rebased.

hvy:
Thanks for the quick response. Do you have any benchmarks to back these changes up?

toslunar:
I've done only a quick check with
`CHAINER_TEST_GPU_LIMIT=0 pytest tests/chainer_tests/functions_tests/connection_tests/test_bilinear.py`.
The results `==== 12 passed, 11 skipped in * seconds ====` were {4.82, 4.74, 4.78} before the PR (53d0977fea95e0e5147dadd4a0fdd20ff7b9ef98), and are {4.00, 4.02, 4.07} with the PR.

hvy:
Could you verify this with a practical case with larger arrays and maybe different memory? I could unfortunately not observe the same significant improvement (I ran the same test in about half your time on my machine). At the same time, I noticed that replacing some of your einsums with tensordot gave a quite large improvement instead. Would you mid verifying it?

chainer/functions/connection/bilinear.py
```diff
@@ -81,7 +81,12 @@ def forward(self, inputs):
 
         xp = cuda.get_array_module(*inputs)
         if xp is numpy:
-            y = numpy.einsum('ij,ik,jkl->il', e1, e2, W)
+            # optimize: y = numpy.einsum('ij,ik,jkl->il', e1, e2, W)
+            y = numpy.einsum(
+                'ij,ijl->il',
+                e1,
+                numpy.einsum('ik,jkl->ijl', e2, W)
```
hvy: For instance, this line can be replaced with a single tensordot over axes 1 in both e2 and W.

toslunar:
Thanks a lot.  `tensordot` improved the performance.

It's possible that your environment installed numpy==1.14.0 (not other 1.14.*) or 1.15.0dev, where ternary `numpy.einsum` is optimized by default.

hvy:
Thanks for the fixes. Indeed, I am testing with numpy==1.14.0. Again, the results vary depending on the data but I ran a couple of benchmarks covering older numpy versions and could see some improvements, especially for backward. I'll merge this PR once merging is unblocked.

hvy:
jenkins, test this please.

hvy:
LGTM!

PR #4756, opened 2018-05-12
Fix hang on exceptions when MultiprocessParallelUpdater is used

### LINKED ISSUE #4709: Process does not exit when CUDARuntimeError occurs
Hi,

My problem is: If I misuse MultiprocessParallelUpdater such that an exception is thrown, the python process does not exit, but just hangs.

The code to reproduce is at the bottom. It causes `cupy.cuda.runtime.CUDARuntimeError: cudaErrorInitializationError: initialization error`, but the python process does not exit.

I'd like the process to exit, but I am not sure how -- I can't use `python -m mpi4py` or call `mpi4py.MPI.COMM_WORLD.abort()` as in https://github.com/chainer/chainermn/issues/236 since this code does not use mpi4py.

Thank you.

* Conditions
  - Chainer version 4.0.0
  - CuPy version 4.0.0
  - OS/Platform ubuntu
  - CUDA/cuDNN version 9.0
* Code to reproduce

```
import chainer
import chainer.functions as F
import chainer.links as L

class MLP(chainer.Chain):

    def __init__(self, n_units, n_out):
        super(MLP, self).__init__()
        with self.init_scope():
            # the size of the inputs to each layer will be inferred
            self.l1 = L.Linear(None, n_units)  # n_in -> n_units
            self.l2 = L.Linear(None, n_units)  # n_units -> n_units
            self.l3 = L.Linear(None, n_out)  # n_units -> n_out

    def __call__(self, x):
        h1 = F.relu(self.l1(x))
        h2 = F.relu(self.l2(h1))
        return self.l3(h2)

def train():
    train, test = chainer.datasets.get_mnist()

    batch_size = 64
    learning_rate = 0.05

    model = L.Classifier(MLP(1000, 10))

    optimizer = chainer.optimizers.MomentumSGD(learning_rate)
    optimizer.setup(model)
    optimizer.add_hook(chainer.optimizer.WeightDecay(5e-4))

    # Set up a trainer
    num_gpus = 2
    devices = range(num_gpus)

    # this is just to force the error
    chainer.cuda.get_device_from_id(0).use()

    train_iters = [chainer.iterators.MultiprocessIterator(i, batch_size, n_processes=num_gpus) \
                   for i in chainer.datasets.split_dataset_n_random(train, len(devices))]

    updater = training.updaters.MultiprocessParallelUpdater(train_iters, optimizer, devices=range(num_gpus))
    updater.setup_workers()


if __name__=="__main__":
    train()
```

Stacktrace:
```
algo-1_1  | Process _Worker-1:
algo-1_1  | Traceback (most recent call last):
algo-1_1  |   File "/usr/lib/python3.5/multiprocessing/process.py", line 249, in _bootstrap
algo-1_1  |     self.run()
algo-1_1  |   File "/usr/local/lib/python3.5/dist-packages/chainer/training/updaters/multiprocess_parallel_updater.py", line 45, in run
algo-1_1  |     dev.use()
algo-1_1  |   File "cupy/cuda/device.pyx", line 101, in cupy.cuda.device.Device.use
algo-1_1  |   File "cupy/cuda/device.pyx", line 107, in cupy.cuda.device.Device.use
algo-1_1  |   File "cupy/cuda/runtime.pyx", line 184, in cupy.cuda.runtime.setDevice
algo-1_1  |   File "cupy/cuda/runtime.pyx", line 136, in cupy.cuda.runtime.check_status
algo-1_1  | cupy.cuda.runtime.CUDARuntimeError: cudaErrorInitializationError: initialization error

```

I had a problem similar to this one, but with mpi4py / chainermn, so the answer might be similar: https://github.com/chainer/chainermn/issues/236

keisukefukuda:
This is a PR to fix #4709.

## The problem
When `MultiprocessParallelUpdater` is used and a worker process exists before or during a training process unexpectedly, the master process is not aware of the fail and still tries to communicate using NCCL. As a result, the whole processes hang and the worker process becomes a zombie.

## The solution
This PR adds a `SIGCHLD` signal handler in the master process of `MultiprocessParallelUpdater`. The handler checks if the child process exits abnormally and shut down the master process if necessary to avoid deadlock.

## Restriction
This PR uses `os.waitpid()` function, which behaves differently on Unix and Windows platforms. As of now, this PR only works on Unix.

### DIFF
--- a/chainer/training/updaters/multiprocess_parallel_updater.py
+++ b/chainer/training/updaters/multiprocess_parallel_updater.py
@@ -19,21 +19,6 @@ except ImportError:
 import numpy
 
 
-def _sigchld_handler(signo, stk):
-    import sys
-    import os
-
-    pid, stat = os.waitpid(-1, os.WNOHANG)
-
-    if stat != 0:
-        sys.stderr.write("******************************************\n")
-        sys.stderr.write("Chainer multiprocess parallel updater: \n")
-        sys.stderr.write("   It seems that an uncaught exception in a worker process\n")
-        sys.stderr.write("******************************************\n")
-        sys.stderr.write("\n\n")
-        sys.stderr.flush()
-        sys.exit(-1)
-
 class _Worker(multiprocessing.Process):
 
     def __init__(self, proc_id, pipe, master):
@@ -183,6 +168,7 @@ class MultiprocessParallelUpdater(standard_updater.StandardUpdater):
 
         self._pipes = []
         self._workers = []
+        self._worker_pids = []
         self.comm = None
 
     @staticmethod
@@ -193,8 +179,25 @@ class MultiprocessParallelUpdater(standard_updater.StandardUpdater):
         for pipe in self._pipes:
             pipe.send(message)
 
+    def _sigchld_handler(self, signo, stk):
+        import sys
+        import os
+
+        pid, stat = os.waitpid(-1, os.WNOHANG)
+
+        if pid in self._worker_pids and stat != 0:
+            sys.stderr.write("\n")
+            sys.stderr.write("*" * 70 + "\n")
+            sys.stderr.write("MultiprocessParallelUpdater: \n")
+            sys.stderr.write("   An uncaught exception occured in "
+                             "a worker process (PID {})\n".format(pid))
+            sys.stderr.write("*" * 70 + "\n")
+            sys.stderr.write("\n")
+            sys.stderr.flush()
+            sys.exit(-1)
+
     def setup_workers(self):
-        signal.signal(signal.SIGCHLD, _sigchld_handler)
+        signal.signal(signal.SIGCHLD, self._sigchld_handler)
 
         if self._initialized:
             return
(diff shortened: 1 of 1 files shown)


keisukefukuda:
@okuta I'm now investigating it. sorry for the inconvenience.

keisukefukuda:
I'm still working on it. It's still alive.

imos:
Just curious:
- Isn't setting a signal handler implicitly surprising to users?
- If a user uses a function using waitpid(-1), does it work properly?
- What if having a loop to check worker.is_alive() repeatedly with some intervals in finalize?

keisukefukuda:
Your question makes sense. I need to check them.

I think using signal is not portable, so simpler solution as you suggested
might be better.

2019年6月5日(水) 18:29 Kentaro IMAJO <notifications@github.com>:

keisukefukuda:
Although the issue is useful, I'm closing it because it's stale and it would be a non-trivial work to impliment the functionality on Windows.

PR #4757, opened 2018-05-13
Add abs to document of Variable

toslunar:
I found several users, including me, search for the absolute function and find `chainer.functions.absolute`, while python's `abs` is available.  Add `abs` to the docstring of `Variable` even though it does not have a "-tion" form in English.


kmaehashi:
LGTM!

PR #4758, opened 2018-05-14
Flush the output stream after PrintReport reports

ktns:
PrintReport extension now seems not flushing the output stream at all, and output comes after a significant delay unless ProgressBar or something is used together.

This patch would fix that, but I can't provide a test because I don't know how I can test the stream is flushed properly.

### DIFF
--- a/chainer/training/extensions/print_report.py
+++ b/chainer/training/extensions/print_report.py
@@ -24,13 +24,18 @@ class PrintReport(extension.Extension):
     """
 
     def __init__(self, entries, log_report='LogReport', out=sys.stdout):
-        if not hasattr(out, "write") or not hasattr(out, "flush"):
+        if not hasattr(out, "write"):
             raise TypeError(
-                "Output stream should support `write' and `flush' method!")
+                "Output stream should support `write' method!")
 
         self._entries = entries
         self._log_report = log_report
         self._out = out
+
+        if hasattr(out, "flush"):
+            self._flush = lambda: out.flush()
+        else:
+            self._flush = lambda: None
 
         self._log_len = 0  # number of observations already printed
 
@@ -51,7 +56,6 @@ class PrintReport(extension.Extension):
 
         if self._header:
             out.write(self._header)
-            out.flush()
             self._header = None
 
         log_report = self._log_report
@@ -88,4 +92,4 @@ class PrintReport(extension.Extension):
             else:
                 out.write(empty)
         out.write('\n')
-        out.flush()
+        self._flush()
--- a/chainer/training/extensions/print_report.py
+++ b/chainer/training/extensions/print_report.py
@@ -24,6 +24,9 @@ class PrintReport(extension.Extension):
     """
 
     def __init__(self, entries, log_report='LogReport', out=sys.stdout):
+        if not hasattr(out, "write") or not hasattr(out, "flush"):
+            raise TypeError("Output stream should support `write' and `flush' method!")
+
         self._entries = entries
         self._log_report = log_report
         self._out = out
--- a/chainer/training/extensions/print_report.py
+++ b/chainer/training/extensions/print_report.py
@@ -24,10 +24,6 @@ class PrintReport(extension.Extension):
     """
 
     def __init__(self, entries, log_report='LogReport', out=sys.stdout):
-        if not hasattr(out, "write"):
-            raise TypeError(
-                "Output stream should support `write' method!")
-
         self._entries = entries
         self._log_report = log_report
         self._out = out
--- a/chainer/training/extensions/print_report.py
+++ b/chainer/training/extensions/print_report.py
@@ -83,5 +83,5 @@ class PrintReport(extension.Extension):
             else:
                 out.write(empty)
         out.write('\n')
-        if hasattr(out, "flush"):
+        if hasattr(out, 'flush'):
             out.flush()
--- a/chainer/training/extensions/print_report.py
+++ b/chainer/training/extensions/print_report.py
@@ -47,6 +47,7 @@ class PrintReport(extension.Extension):
 
         if self._header:
             out.write(self._header)
+            out.flush()
             self._header = None
 
         log_report = self._log_report
@@ -83,3 +84,4 @@ class PrintReport(extension.Extension):
             else:
                 out.write(empty)
         out.write('\n')
+        out.flush()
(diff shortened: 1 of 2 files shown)


niboshi:
Thank you for reporting & making a PR!
`out` does not necessarily have `flush` method. How about checking it with `hasattr` before calling it?

ktns:
I believe all streams should support `flush` method as `IOBase` class has `flush` method. Should I add a duck typing check or `isinstance(out, IOBase)` check in `__init__` method of `PrintReport`?

niboshi:
`out` does not have to be inherited from `IOBase`. In Python, there's no concrete requirements for an object to be "file-like".
So I'd suggest duck typing check.

niboshi:
Thank you for fixing.
I think it should function for `out` obj without `flush`.
So rather than rejecting, how about just replacing `out.flush()` with `if hasattr(out, 'flush'): out.flush()`?

ktns:
Sorry for late response. I've modified the type checking and make it skip flushing if `out` does not have `flush` method. Please take a look at the code.

chainer/training/extensions/print_report.py
```diff
@@ -24,10 +24,19 @@ class PrintReport(extension.Extension):
     """
 
     def __init__(self, entries, log_report='LogReport', out=sys.stdout):
+        if not hasattr(out, "write"):
```
niboshi: Please remove this check.

chainer/training/extensions/print_report.py
```diff
@@ -24,10 +24,19 @@ class PrintReport(extension.Extension):
...
+                "Output stream should support `write' method!")
+
         self._entries = entries
         self._log_report = log_report
         self._out = out
 
+        if hasattr(out, "flush"):
+            self._flush = lambda: out.flush()
+        else:
+            self._flush = lambda: None
```
niboshi: You don't have to make an attribute here.

chainer/training/extensions/print_report.py
```diff
@@ -83,3 +92,4 @@ def _print(self, observation):
             else:
                 out.write(empty)
         out.write('\n')
+        self._flush()
```
niboshi: You can simply check here:
```
if hasattr(os, 'flush'):
    os.flush()
```

ktns: fixed.

chainer/training/extensions/print_report.py
ktns: fixed.

chainer/training/extensions/print_report.py
ktns: Sorry I've simply overlooked this comment. Now fixed.

tests/chainer_tests/training_tests/extensions_tests/test_print_report.py
```diff
@@ -0,0 +1,41 @@
...
+        self.trainer.extend(self.logreport)
+        self.trainer.extend(self.report)
+        self.logreport.log = [{'epoch': 0}]
+
+    def test_stream_with_flush_is_flushed(self):
+        self._setup(delete_flush=False)
+        self.assertTrue(hasattr(self.stream, 'flush'))
+        self.stream.flush.assert_not_called()
+        self.report(self.trainer)
+        self.stream.flush.assert_called_once_with()
```
niboshi: Use `assert_called_with`.
Even if it is called more than once, it doesn't mean the feature is broken.

tests/chainer_tests/training_tests/extensions_tests/test_print_report.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def test_stream_without_flush_raises_no_exception(self):
+        self._setup(delete_flush=True)
+        self.assertFalse(hasattr(self.stream, 'flush'))
+        try:
+            self.report(self.trainer)
+        except Exception as e:
```
niboshi: You don't have to catch exception.
Test framework (pytest) can handle exceptions.

chainer/training/extensions/print_report.py
```diff
@@ -83,3 +83,5 @@ def _print(self, observation):
             else:
                 out.write(empty)
         out.write('\n')
+        if hasattr(out, "flush"):
```
niboshi: Use single-quotes for string literals. (`''`)

tests/chainer_tests/training_tests/extensions_tests/test_print_report.py
```diff
@@ -0,0 +1,45 @@
+import sys
+import unittest
+
+from chainer import testing
+from chainer.training import extensions
+from mock import MagicMock
```
niboshi: Please separate `mock` from `chainer` imports, and please avoid directly importing a class (`MagicMock`).
Please see `tests/chainer_tests/test_variable.py`, for example.

tests/chainer_tests/training_tests/extensions_tests/test_print_report.py
```diff
@@ -0,0 +1,45 @@
...
+        self.report(self.trainer)
+        self.stream.flush.assert_called_with()
+
+    def test_stream_without_flush_raises_no_exception(self):
+        self._setup(delete_flush=True)
+        self.assertFalse(hasattr(self.stream, 'flush'))
+        self.report(self.trainer)
+
+    def test_real_stream_raises_no_exception(self):
+        self._setup(stream=sys.stderr)
```
niboshi: Please avoid output to stderr or stdout in unit tests.
Maybe this test is not needed, because it's covered in `test_stream_with_flush_is_flushed`?

ktns:
I've fixed the both. But why do you recommend to avoid using ```sys.stderr ```in pytest? I think pytest handles standard output/error beautifully while testing.

niboshi:
Thank you for the fix.
`stderr` may be used while debugging. Debugging might be a bit difficult if they're mixed.

niboshi:
Please remove `import sys`.
LGTM except that.

ktns:
Sorry for looking over the unused module. Now I've fixed it.

niboshi:
Jenkins, test this please

niboshi:
Thank you very much!

niboshi:
It has been merged directly to v4 branch.
I'll revert this PR tentatively and resubmit to master.

PR #4763, opened 2018-05-14
Fix dtype bug in F.normalize

toslunar:
The test added in the first commit failed because the output dtype was `float64`.

After the refactoring, `F.normalize` supports Frobenius norm, which is equivalent to L2 norm of flattened vector.

### DIFF
--- a/chainer/functions/normalization/l2_normalization.py
+++ b/chainer/functions/normalization/l2_normalization.py
@@ -26,6 +26,7 @@ class NormalizeL2(function_node.FunctionNode):
         self.retain_inputs((0,))
         x, = inputs
         xp = cuda.get_array_module(x)
+        # keep x.ndim >= 1 to avoid casting to self.eps' type
         norm = xp.sqrt(xp.sum(
             xp.square(x), axis=self.axis, keepdims=True)) + self.eps
         return x / norm,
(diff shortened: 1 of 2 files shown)


chainer/functions/normalization/l2_normalization.py
```diff
@@ -26,23 +26,21 @@ def forward(self, inputs):
         self.retain_inputs((0,))
         x, = inputs
         xp = cuda.get_array_module(x)
-        norm = xp.linalg.norm(x, axis=self.axis) + self.eps
-        norm = xp.expand_dims(norm, self.axis)
+        norm = xp.linalg.norm(x, axis=self.axis, keepdims=True) + self.eps
```
niboshi: According [the documentation](https://docs.scipy.org/doc/numpy-1.14.0/reference/generated/numpy.linalg.norm.html), `numpy.linalg.norm(keepdims=)` has been introduced in 1.10.0. We should avoid using it.

niboshi:
Could you explain why it was converted to `float64` previously, and how it is fixed now?
Does `numpy.expand_dims` convert to `float64` unconditionally?

chainer/functions/normalization/l2_normalization.py
```diff
@@ -12,6 +12,10 @@ class NormalizeL2(function_node.FunctionNode):
 
     def __init__(self, eps=1e-5, axis=1):
         self.eps = eps
+        if isinstance(axis, int):
+            axis = axis,
+        if len(axis) not in [1, 2]:
```
niboshi: Tuple is preferred (faster) to list.

niboshi:
LGTM! Thank you!

PR #4769, opened 2018-05-14
Avoid zero division in F.normalize

### LINKED ISSUE #4765: backward of F.normalize is not stable
`NormalizeL2.backward` computes 0/0 if the input contains a zero vector.  PR #4190, I wrote, caused this.  Sorry.

To begin with, x/(||x|| + eps) is C^1 but not C^2 (at x=0).  The correct backward might not be a good choice.

toslunar:
close #4765.
merge #4763 first.

I added `_where_indices`, which would be a good alternative to `F.where` when we keep from backwarding nan grads.


niboshi:
Please rebase to master, as #4763 has been merged.

niboshi:
Correct me if I'm wrong.

Given a positive constant `eps`, I think dy/dx approaches `1 / eps` as `x` approaches 0.
How about implementing like that, instead of treating it as nondifferentiable?

toslunar:
> dy/dx approaches `1 / eps` as `x` approaches 0.

Yes.  Therefore the backward of the function is continuous.  I've not come up with a way to compute it without zero-division.  It seems a similar situation that `numpy.sinc(x)` uses `where` and `sin(pi * x)/(pi * x)`.

Computing the norm by `F.sqrt(F.sum(...))` tends to round it down to `0.`.  This seems to me good enough to check `nonzero` later without epsilon.

I added tests with near-zero inputs.  `gradient_check.check_backward` is left having loose `atol, rtol` because numerical backward is inaccurate.
I additionally checked that the backward is more accurate by testing:
```
    def check_backward(self, x_data, axis, y_grad):
# ...
        if self.nonzeros == 0:
            x = chainer.Variable(x_data)
            gy = chainer.Variable(y_grad)
            gx, = chainer.grad([f(x)], [x], [gy])
            x_grad = gx.array
            xp = cuda.get_array_module(x_grad)
            max_norm_x = xp.max(xp.sqrt(
                xp.sum(x_data ** 2, axis=axis, keepdims=True)))
            max_norm_gy = xp.max(xp.sqrt(
                xp.sum(y_grad ** 2, axis=axis, keepdims=True)))
            tol = 2. * max_norm_x * max_norm_gy / self.eps ** 2
            print(max_norm_x, max_norm_gy, tol)
            xp.testing.assert_array_less(
                xp.abs(x_grad - y_grad/self.eps),
                max(1e-6, tol)
            )
```

niboshi:
Could you check test failure?

toslunar:
I replaced `nonzero` with boolean masks to fix a 0d-array problem.  This PR depends on #4845.

niboshi:
Could you rebase?

chainer/functions/normalization/l2_normalization.py
```diff
@@ -43,7 +67,11 @@ def backward(self, indexes, grad_outputs):
         norm = F.broadcast_to(norm, gy.shape)
 
         x_gy_reduced = F.sum((x * gy), axis=self.axis, keepdims=True)
-        x_gy_reduced /= norm_noeps
+        mask = norm_noeps.array != 0
```
niboshi: Please add brief comment to describe that the intention is to avoid zero division.

tests/chainer_tests/functions_tests/normalization_tests/test_l2_normalization.py
```diff
@@ -26,14 +45,32 @@
         {'shape': (), 'axis': ()},
     ],
     [
-        {'eps': 1e-5},
-        {'eps': 1e-1},
+        {'eps': 1e-5, 'nonzeros': None},
```
niboshi: Please add comments describing `nonzeros` and `truezero` parameters.

tests/chainer_tests/functions_tests/normalization_tests/test_l2_normalization.py
```diff
@@ -12,6 +13,24 @@
 from chainer.testing import attr
 
 
+def _skip_if(cond, reason):
```
niboshi: Please add comments for `_skip_if` and `_skip_if_many_zeros`, respectively.

tests/chainer_tests/functions_tests/normalization_tests/test_l2_normalization.py
```diff
@@ -26,14 +45,32 @@
...
+        {'eps': 1e-1, 'nonzeros': 0, 'truezero': False},
+        {'eps': 1e-1, 'nonzeros': 2, 'truezero': True},
+        {'eps': 1e-1, 'nonzeros': 2, 'truezero': False},
     ],
 ]))
 class TestL2Normalization(unittest.TestCase):
 
     def setUp(self):
         self.x = numpy.random.uniform(-1, 1, self.shape).astype(numpy.float32)
+        if self.nonzeros is not None:
```
niboshi: Please add comments for the logic.

toslunar:
The 0d-array problem cannot be fixed by using boolean mask if `numpy<1.13`.

toslunar:
I'd like to drop support for 0-dim input.

niboshi:
Could you describe the problem?

toslunar:
`np.ndarray.__getitem__` and `np.add.at` had been inconsistent with 0-dim mask:
- Old numpy allows `arr[mask]` if `arr.ndim == mask.ndim` or `mask.ndim > 0`.
- The backward of `GetItem` uses `numpy.add.at(arr, mask, grad)`, which raises in old numpy unless `mask.ndim > 0`.

niboshi:
It seems like we should fix `GetItem` first.
How about giving up supporting it in this PR and leave it TODO?

chainer/functions/array/get_item.py
```diff
@@ -52,7 +52,16 @@ def forward(self, inputs):
         xp = cuda.get_array_module(*inputs)
         gx = xp.zeros(self._in_shape, self._in_dtype)
         if xp is numpy:
-            numpy.add.at(gx, self.slices, inputs[0])
+            try:
```
niboshi: How about using `numpy.lib.NumpyVersion` to test NumPy version?

https://docs.scipy.org/doc/numpy-1.14.0/reference/generated/numpy.lib.NumpyVersion.html

(It's better to evaluate and hold the result in a module level variable.)

toslunar:
Fixed!

tests/chainer_tests/functions_tests/normalization_tests/test_l2_normalization.py
```diff
@@ -26,14 +47,33 @@
         {'shape': (), 'axis': ()},
     ],
     [
-        {'eps': 1e-5},
-        {'eps': 1e-1},
+        # nonzeros (optional int): max number of nonzero elems in input
+        # truezero (bool): flag whether zero elems are exactly zero
```
niboshi: Please write the behavior of the case when ths flag is `False`.

tests/chainer_tests/functions_tests/normalization_tests/test_l2_normalization.py
niboshi: Please write a brief comment here.
(Maybe single line is enough here, if the detail is covered in the comments of respective arguments.)

niboshi:
LGTM

Issue #3766, opened 2017-11-02
Too large error in numerical function tests


niboshi:
This is a list of tests I noticed failure but could not fix by adjusting tolerances, due to too large errors.

----
@niboshi
- [ ] TestBatchNormalization.test_double_backward_cpu
----
@bkvogel
- [ ] TestBatchRenormalization.test_backward_cpu
- [ ] TestBatchRenormalization.test_forward_cpu
- [ ] TestFixedBatchRenormalization.test_backward_cpu
----
@mitmul
- [ ] TestDeconvolution2DFunction.test_backward_cpu
- [ ] TestDeconvolutionND.test_backward_cpu
----
@beam2d
- [ ] TestNStepBiRNN.test_backward_cpu
- [ ] TestNStepRNN.test_backward_cpu
----
@kmaehashi #4838
- [x] TestMin.test_backward_multi_axis_invert_cpu
- [x] TestMin.test_backward_negative_multi_axis_invert_cpu
- [x] TestMax.test_backward_multi_axis_invert_cpu
- [x] TestMax.test_backward_negative_multi_axis_invert_cpu
----
@asi1024 #4771
- [x] TestAverage_param_26.test_backward_cpu
----
@toslunar #5210
- [x] TestL2Normalization.test_double_backward_cpu: https://ci.appveyor.com/project/pfnet/chainer/build/11072/job/3xwi9wm2ycjo8cqb, internal CI (daily_master_chainer-shuffled task 631, ID=10)
- [x] TestL2Normalization_param_18.test_backward_cpu: https://ci.appveyor.com/project/pfnet/chainer/build/11015/job/99ivaixp53op2ido
----
@delta2323
- [ ] TestMatMulVarVar_param_1.test_double_backward_cpu
----
@toslunar #5218
- [x] TestContrastive.test_double_backward_cpu
----
(unassigned)
- [ ] SpatialTransformerSampler. test_consistency_with_cudnn_gpu

niboshi:
Perhaps some of the failures are due to non-differentiable inputs.
#3551 could be helpful to solve them.

hvy:
`TestAverage_param_26.test_backward_cpu` generated quite a large error. Should this test be added to the list?

```
E                   AssertionError: 
E                   Not equal to tolerance rtol=0.1, atol=0.01
E                   
E                   (mismatch 100.0%)
E                    x: array(-0.2126242916502541)
E                    y: array(-0.13950088823685647)
```

[...]

PR #4771, opened 2018-05-14
Fix occasional test failure of `F.average` by changing lower-bound of weight.sum

asi1024:
Fixed the lower bound of `self.w.sum()`.
(#3766)

### DIFF
--- a/tests/chainer_tests/functions_tests/math_tests/test_average.py
+++ b/tests/chainer_tests/functions_tests/math_tests/test_average.py
@@ -47,7 +47,7 @@ class TestAverage(unittest.TestCase):
         # Sample weights. Weights should not sum to 0.
         while True:
             self.w = numpy.random.uniform(-2, 2, w_shape).astype(self.dtype)
-            w_sum_eps = 0.2 if self.dtype == numpy.float16 else 5e-2
+            w_sum_eps = 1.0 if self.dtype == numpy.float16 else 5e-2
             if abs(self.w.sum()) > w_sum_eps:
                 break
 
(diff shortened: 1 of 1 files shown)


tests/chainer_tests/functions_tests/math_tests/test_average.py
```diff
@@ -47,7 +47,8 @@ def setUp(self):
         # Sample weights. Weights should not sum to 0.
         while True:
             self.w = numpy.random.uniform(-2, 2, w_shape).astype(self.dtype)
-            if self.w.sum() > 1e-2:
+            w_sum_eps = 0.2 if self.dtype == numpy.float16 else 5e-2
+            if self.w.sum() > w_sum_eps:
```
toslunar: Test `abs(self.w.sum())` to fix the slip in #3704.

asi1024: Fixed.

kmaehashi:
I still observe error after applying this PR. Could you repeat-run the unittest test with `@condition.repeat(20000)` annotated to `test_backward_cpu`?

tests/chainer_tests/functions_tests/math_tests/test_average.py
```diff
@@ -47,7 +47,8 @@ def setUp(self):
         # Sample weights. Weights should not sum to 0.
         while True:
             self.w = numpy.random.uniform(-2, 2, w_shape).astype(self.dtype)
-            if self.w.sum() > 1e-2:
+            w_sum_eps = 1.0 if self.dtype == numpy.float16 else 5e-2
+            if abs(self.w.sum()) > w_sum_eps:
```
kmaehashi: As discussed offline, could you try sum with high precision?

asi1024: I tried `abs(self.w.astype(numpy.float64).sum()) > w_sum_eps`, but the test failed.

kmaehashi: I see.

kmaehashi:
LGTM!

PR #4772, opened 2018-05-14
New backward_accumulate

toslunar:
The current backprop implementations use a trick when a variable is passed to multiple input-indices of a function node.  This PR renew the trick and simplify codes.

This PR joins some backward grad summation without memory consumption drawbacks.

- [x] refactor `chainer.Variable.backward`
- [x] refactor `chainer.grad`
- [x] take care of `FunctionNode.backward_accumulate`


toslunar:
Some tests in #4779 fail with `CHAINER_LAZY_GRAD_SUM=1`.  Now the tests pass.

chainer/_backprop_utils.py
```diff
@@ -1,40 +1,112 @@
...
-        sum_gx += g_input
-    elif g_input is not None:
-        sum_gx += g_input,
+def backprop_step(
+        func, target_input_indexes, grad_outputs, grad_inputs):
+    assert isinstance(target_input_indexes, tuple)
+    assert target_input_indexes == tuple(sorted(target_input_indexes))
+    assert isinstance(grad_outputs, tuple)
+    if func.backward_accumulate.__code__ \
+            is not chainer.FunctionNode.backward_accumulate.__code__:
```
beam2d: I feel this dispatching is a bit tricky and makes it hard to customize the behavior (e.g. writing `backward_accumulate` that behaves almost same as the default one with slight differences). I also think the default implementation should be written to the default `backward_accumulate` code (instead of `raise NotImplementedError`; it makes the readers hard to follow the logic when they only override `backward`). How about the following fix:

1. Allow `backward_accumulate` return a tuple of tuples instead of a tuple of variables. Each tuple contains the given grad input and the newly computed gradient that will be accumulated by the backprop logic. It allows `backward_accumulate` to choose lazy accumulation if available.
2. Pop only one variable in the list to pass to `backward_accumulate` instead of reducing them everytime. It does not hurt the performance even when the function supports fused accumulation.

The default mode may be worth optimizing in the current way, though (skip preparing `grad_inputs_tuple`, reduce the number of function calls). Anyway, I feel it's better to allow users to emulate the same behavior as the default one.

beam2d:
The idea looks good to me. I added a comment on the dispatching part of the default implementation. The other parts look good to me.

chainer/_backprop_utils.py
beam2d: We discussed offline and found that the code is anyway customizable by patching some methods (including this `backprop_step`). We assume that there are few users that want to customize this point and take the design simplicity and performance gain here by dispatching in an ad-hoc way.

chainer/_backprop_utils.py
```diff
@@ -1,40 +1,120 @@
...
-    # gx is None and g_input is None
-    if len(sum_gx) == 0:
-        sum_gx = None
-    elif len(sum_gx) == 1:
-        sum_gx = sum_gx[0]
 
-    return sum_gx
+def backprop_step(
+        func, target_input_indexes, grad_outputs, grad_inputs):
+    """Accumulate gradients of a FunctionNode
```
beam2d: s/Accumulate/Accumulates/

toslunar:
Jenkins, test this please.

beam2d:
Jenkins, test this please.

beam2d:
LGTM

PR #4678, opened 2018-04-25
[WIP] Probability distribution implementations

YoshikawaMasashi:
#3622 

I am implementing probability distribution.

## Distribution
- [x] Bernoulli (merged)
- [x] Beta (merged)
- [x] Binomial (splited)
- [x] Categorical (merged)
- [x] Cauchy (splited)
- [x] Chi2 (splited)
- [x] Dirichlet (splited)
- [x] Distribution (marged)
- [x] Exponential (splited)
- [x] FisherSnedecor (splited)
- [x] Gamma (splited)
- [x] Geometric (splited)
- [x] Gumbel (splited)
- [x] Laplace (marged)
- [x] LogNormal (merged)
- [x] Multinomial (splited)
- [x] MultivariateNormal (merged)
- [x] Normal (marged)
- [x] OneHotCategorical (splited)
- [x] Pareto (splited)
- [x] StudentT (splited)
- [x] Poisson (splited)
- [x] Uniform (merged)

## Other
- [x] Bijector (splited)
- [x] TransformedDistribution (splited)
- [x] kl_divergence (marged)
- [x] register_kl (marged)
- [x] functions of kl divergence (marged)

## Error Handling
- [ ] type error

## Test
- [x] arch(CPU, GPU)
- [x] Parameter(Chainer.Variable, numpy.array)
- [ ] numerical integral test for kl

PR #4773, opened 2018-05-14
Normal Distribution

YoshikawaMasashi:
I am implementing many Distributions. #4678 
#4678 PR is too big to review.
Therefore I am dividing this PR into some PRs.

First, I opened this PR which contain normal distribution and utils related to normal distribution.


chainer/distribution.py
```diff
@@ -0,0 +1,260 @@
...
+
+        """
+        raise NotImplementedError
+
+    def cdf(self, x):
+        """Returns Cumulative Distribution Function for a input variable.
+
+        Args:
+            x(:class:`~chainer.Variable` or :class:`numpy.ndarray` or \
+            :class:`cupy.ndarray`): Input variable representing a random
```
beam2d: It is clear from the type information that it is a variable, so I think the documentation should be more direct (no need of writing "Input variable representing"). I also think "a random variable" is a little bit misleading because `x` can be a deterministic value. How about writing, e.g., "A data point in the domain of the distribution"?

chainer/distribution.py
```diff
@@ -0,0 +1,260 @@
...
+            variable.
+
+        Returns:
+            ~chainer.Variable: Output variable representing Cumulative
```
beam2d: Ditto: it is apparent that this is a variable, so you can just write "Cumulative distribution function value evaluated at `x`." Ditto for the docstrings of other methods.

chainer/distribution.py
beam2d: Another point: could you write the specification of input/output shapes? For example, can `x` be a batch of multiple examples, or should it represent only one example? Ditto for other methods (if most methods have the same specification, it is also ok to write that in the document of `Distribution` class itself instead of repeating it at every method).

chainer/distribution.py
```diff
@@ -0,0 +1,260 @@
...
+            variable.
+
+        Returns:
+            ~chainer.Variable: Output variable representing probability.
+
+        """
+        raise NotImplementedError
+
+    def sample(self, shape=()):
+        """Samples from this distribution.
```
beam2d: Be more explicit: samples what? My suggestion is "Samples[Draws] random points from the distribution".

chainer/distribution.py
```diff
@@ -0,0 +1,260 @@
...
+
+        Args:
+            shape(:class:`tuple` of :class:`int`): Sampling shape.
```
beam2d: Could you write how the output shape is built from this parameter in combination with `batch_shape` and `event_shape`?

chainer/distributions/kl_divergence.py
```diff
@@ -0,0 +1,69 @@
...
+
+       This test case runs only when `numpy>=1.10` is installed.
+
+       >>> from chainer import distributions
+       ... @register_kl(distributions.Normal, distributions.Normal)
+       ... def _kl_dist1_dist2(dist1, dist2):
+       ...     return KL
+
+    Args:
+        dist1(:class:`~chainer.Distribution`): Distribution to calculate KL
```
beam2d: Maybe better to write that it is the first (left) operand of the KL divergence because KL divergence is not symmetric against the operands. Ditto for `dist2`. It is also good to write the mathematical form here, too (like written in `kl_divergence` docstring below).

chainer/distributions/kl_divergence.py
```diff
@@ -0,0 +1,69 @@
+from chainer.distributions import normal
+from chainer.functions.math import exponential
+
+_KLDIVERGENCE = {}
+
+
+def register_kl(dist1, dist2):
+    """Decorator to register KL divergence function.
```
beam2d: Write when/how to use this decorator and what can be done with it. Hint:
- What should be the signature of the decorated function?
- The decorated function will be called by `kl_divergence` based on the argument types. Using it, we can write distribution-agnostic code to compute KL divergence.

chainer/distributions/kl_divergence.py
```diff
@@ -0,0 +1,69 @@
...
+            divergence.
+
+    """
+    def f(kl):
+        _KLDIVERGENCE[dist1, dist2] = kl
+    return f
+
+
+def kl_divergence(dist1, dist2):
+    """Kullback-Leibler divergence.
```
beam2d: "Computes" or "Returns". It is also better to refer to `register_kl`.

chainer/distributions/kl_divergence.py
```diff
@@ -0,0 +1,69 @@
...
+    Returns:
+        ~chainer.Variable: Output variable representing kl divergence
+            :math:`D_{KL}(dist1 || dist2)`.
+
+    """
+    return _KLDIVERGENCE[type(dist1), type(dist2)](dist1, dist2)
+
+
+def cross_entropy(dist1, dist2):
+    """Cross entropy.
```
beam2d: "Computes" or "Returns"

chainer/distributions/kl_divergence.py
```diff
@@ -0,0 +1,69 @@
...
+    Returns:
+        ~chainer.Variable: Output variable representing cross entropy
+            :math:`H(dist1, dist2)`.
+
+    """
+    return dist1.entropy() + kl_divergence(dist1, dist2)
+
+
+@register_kl(normal.Normal, normal.Normal)
+def _kl_normal_normal(dist1, dist2):
```
beam2d: How about writing it in `normal.py` instead of here? I feel this function is more tightly coupled with the `Normal` class.

chainer/distributions/normal.py
```diff
@@ -0,0 +1,239 @@
...
+
+
+class Normal(Distribution):
+
+    """Normal Distribution.
+
+    Args:
+        loc(:class:`~chainer.Variable` or :class:`numpy.ndarray` or \
+        :class:`cupy.ndarray`): Parameter of distribution representing the \
+        location :math:`\\mu`.
```
beam2d: Better to mention that this is the `mean` parameter of the normal distribution.

chainer/distributions/normal.py
```diff
@@ -0,0 +1,239 @@
...
+from chainer.functions.array import repeat
+from chainer.functions.math import erf
+from chainer.functions.math import erfinv
+from chainer.functions.math import exponential
+import numpy
+
+
+class Normal(Distribution):
+
+    """Normal Distribution.
```
beam2d: Better to write how the normal distribution is parameterized (e.g., write down the density function using the same variable names as the argument documentation: mu and sigma).

chainer/distributions/normal.py
```diff
@@ -0,0 +1,239 @@
...
+        if isinstance(loc, chainer.Variable):
+            self.loc = loc
+        else:
+            self.loc = chainer.Variable(loc)
+        if isinstance(scale, chainer.Variable):
+            self.scale = scale
+        else:
+            self.scale = chainer.Variable(scale)
+
+    def __copy__(self):
```
beam2d: Why do you need to explicitly write `__copy__`?

chainer/distributions/normal.py
```diff
@@ -0,0 +1,239 @@
...
+        return self._copy_to(Normal(self.loc, self.scale))
+
+    @property
+    def batch_shape(self):
+        return self.loc.shape
+
+    def cdf(self, x):
+        """Returns Cumulative Distribution Function for a input variable.
```
beam2d: No need to repeat the documentation if there is nothing special for this distribution (ditto for other methods and properties).

chainer/distribution.py
```diff
@@ -0,0 +1,260 @@
+import copy
+
+
+class Distribution(object):
+
+    """Interface of Distribution
+
+    `Distribution` is a bass class to treat probability distributions.
+    When initialization, it takes parameter as input.
```
beam2d: Could you also emphasize that the outputs of every method and property that computes them from `chainer.Variable` can basically be differentiated?

chainer/distributions/normal.py
```diff
@@ -0,0 +1,239 @@
...
+        """Returns logarithm logarithm of probability for a input variable.
+
+        Args:
+            x: Input variable representing a random variable.
+
+        Returns:
+            Output variable representing logarithm of probability.
+
+        """
+        return - 0.5 * numpy.log(2 * numpy.pi) \
```
beam2d: How about precomputing `-0.5 * numpy.log(2 * numpy.pi)` at the global scope and just using it here? It is also better ot use `math` instead of `numpy` for it (because numpy is slow for scalar computations).

chainer/distributions/normal.py
```diff
@@ -0,0 +1,239 @@
...
+        Args:
+            x(:class:`~chainer.Variable` or :class:`numpy.ndarray` or \
+            :class:`cupy.ndarray`): Input variable representing a random
+            variable.
+
+        Returns:
+            ~chainer.Variable: Output variable representing probability.
+
+        """
+        return 1. / (2 * numpy.pi) ** 0.5 \
```
beam2d: Ditto (precomputing the constant coefficient)

chainer/functions/math/erfinv.py
```diff
@@ -0,0 +1,63 @@
...
+        return 'erfinv'
+
+    def check_type_forward(self, in_types):
+        type_check.expect(in_types.size() == 1)
+        type_check.expect(in_types[0].dtype.kind == 'f')
+
+    def forward_cpu(self, x):
+        global _erfinv_cpu
+        if _erfinv_cpu is None:
+            try:
```
beam2d: Could you write try-import at the global scope? The exception can be cached and reraised in `forward_cpu`. (see e.g. https://github.com/chainer/chainer/blob/master/chainer/datasets/image_dataset.py#L4-L9)

chainer/functions/math/erfinv.py
```diff
@@ -0,0 +1,63 @@
...
+        self.retain_inputs((0,))
+        return cuda.elementwise(
+            'T x', 'T y',
+            'y = erfinv(x)',
+            'elementwise_erfinv',
+        )(x[0]),
+
+    def backward(self, indexes, gy):
+        x = self.get_retained_inputs()[0]
+        return numpy.pi ** 0.5 / 2 * chainer.functions.exp(erfinv(x) ** 2) \
```
beam2d: How about precomputing the coefficient? (ditto as above)

chainer/distributions/normal.py
YoshikawaMasashi: I made a mistake. I didn't need to.

YoshikawaMasashi:
Thank you for the comments.
I fixed the codes based on your comments.

beam2d:
Thank you. I will check the updates.

chainer/distributions/kl_divergence.py
```diff
@@ -0,0 +1,97 @@
...
+    returns KL divergence value. When `kl_divergence` will be called taking a
+    instance of `Dist1` and `Dist2` after registration, the decorated function
+    will be called.
+
+    .. admonition:: Example
+
+       >>> from chainer import distributions
+       ... @register_kl(Dist1, Dist2)
+       ... def _kl_dist1_dist2(dist1, dist2):
+       ...     return KL
```
beam2d: Make it correct code. You may skip some part with `...`. Also, `register_kl` should be `distributions.register_kl`. I feel it's better to use a code block here instead of the REPL-style document.

chainer/distributions/kl_divergence.py
```diff
@@ -0,0 +1,97 @@
+_KLDIVERGENCE = {}
+
+
+def register_kl(dist1, dist2):
```
beam2d: I feel it's better to define it in `distribution.py` instead of under `distributions`.

YoshikawaMasashi:
I fixed again.

chainer/distribution.py
```diff
@@ -0,0 +1,420 @@
...
+        >>> dist.event_shape
+        (2,)
+        >>> dist.batch_shape
+        (4, 3)
+        >>> sample = dist.sample(sample_shape=(6, 5))
+        >>> sample.shape
+        (6, 5, 4, 3, 2)
+
+    Every function about probability takes realization value whose shape is
+    `(sample_shape, batch_shape, event_shape)` and returns an evaluated value
```
beam2d: Maybe it's better to write "the concatenation of `sample_shape`, `batch_shape`, and `event_shape`" rather than `(sample_shape, batch_shape, event_shape)` because the latter is incorrect as a Python expression and thus confusing.

chainer/distribution.py
```diff
@@ -0,0 +1,420 @@
...
+            ~chainer.Variable: Probability evaluated at `x`.
+
+        """
+        raise NotImplementedError
+
+    def sample(self, sample_shape=()):
+        """Samples random points from the distribution.
+
+        This function call `sample_n` and reshape a result of `sample_n` to
+        `(sample_shape, batch_shape, event_shape)`. On implementing sampling
```
beam2d: Better to write `sample_shape + batch_shape + event_shape` rather than `(sample_shape, batch_shape, event_shape)`.

chainer/distribution.py
```diff
@@ -0,0 +1,420 @@
...
+                n *= shape_
+            final_shape = sample_shape + final_shape
+        samples = self.sample_n(n)
+        return samples.reshape(final_shape)
+
+    def sample_n(self, n):
+        """Samples n random points from the distribution.
+
+        This function returns sampled points whose shape is
+        `(n, batch_shape, event_shape)`. On implementing sampling codes in an
```
beam2d: Similarly to the above comment, it's better to write `(n,) + batch_shape + event_shape` instead of `(n, batch_shape, event_shape)`.

chainer/testing/distribution_test.py
```diff
@@ -0,0 +1,435 @@
...
+    if _error is not None:
+        raise RuntimeError('''\
+{} is not available.
+
+Reason: {}: {}'''.format(__name__, type(_error).__name__, _error))
+
+
+def distribution_unittest(dist, scipy_dist_name, params_init, sample_for_test,
+                          tests=set(), continuous=True, support="real",
+                          event_shape=(), scipy_onebyone=False):
```
beam2d: How about making it just a base class instead of a class decorator? Options can be provided as methods or (class) attributes of each child class. Conditional tests with `tests` may be provided by a method decorator that checks the `tests` attribute of the child class (btw, I think the name `tests` should be more explicit, like `test_targets`).

chainer/distribution.py
```diff
@@ -0,0 +1,420 @@
+import copy
+
+
+class Distribution(object):
+
+    """Interface of Distribution
+
+    `Distribution` is a bass class to treat probability distributions.
```
bkvogel: Consider rewording: examples:

`Distribution` is a bass class for dealing with probability distributions.
`Distribution` is a bass class for handling probability distributions.
`Distribution` is a bass class for a probability distribution.

chainer/distribution.py
```diff
@@ -0,0 +1,420 @@
...
+
+    This class provides a mean to perform following operations.
```
bkvogel: Wording: some examples:

This class provides the following capabilities.
This class can perform the following operations.

chainer/distribution.py
```diff
@@ -0,0 +1,420 @@
...
+    1. Sampling random points.
+    2. Evaluating a functions about probability at given realization values.
```
bkvogel: 2. Evaluating a functions about probability at given realization values. ->
2. Evaluating a probability-related function at a given realization value.

chainer/distribution.py
```diff
@@ -0,0 +1,420 @@
...
+        (e.g., probability density function, probability mass function)
+    3. Obtaining properties of distributions.
+        (e.g., mean, variance)
+
+    Note that every method and property that computes them from
+    `chainer.Variable` can basically be differentiated.
+
+    In this class, sampled random points and realization values given in
+    function about probability is called "sample".  Sample consists of batches,
```
bkvogel: Consider rewording "function about probability". How about "probability-related function"? or "a function related to probability"?

chainer/distribution.py
```diff
@@ -0,0 +1,420 @@
...
+        >>> dist = D.MultivariateNormal(loc, l)
+        >>> dist.event_shape
+        (2,)
+        >>> dist.batch_shape
+        (4, 3)
+        >>> sample = dist.sample(sample_shape=(6, 5))
+        >>> sample.shape
+        (6, 5, 4, 3, 2)
+
+    Every function about probability takes realization value whose shape is
```
bkvogel: Same here. Consider rewording "function about probability".

chainer/distribution.py
```diff
@@ -0,0 +1,420 @@
...
+        """Returns the shape of a batch.
+
+        Returns:
+            tuple: The shape of a sample that is not identical and indipendent.
+
+        """
+        raise NotImplementedError
+
+    def cdf(self, x):
+        """Evaluates the cumulative distribution function at a given points.
```
bkvogel: Evaluates the cumulative distribution function at a given points. ->
Evaluates the cumulative distribution function at the given points.

chainer/distribution.py
```diff
@@ -0,0 +1,420 @@
...
+        """Returns the shape of an event.
+
+        Returns:
+            tuple: The shape of a sample that is not identical and indipendent.
+
+        """
+        raise NotImplementedError
+
+    def icdf(self, x):
+        """Evaluates the inverse cumulative distribution function at a given points.
```
bkvogel: Evaluates the inverse cumulative distribution function at a given points. ->
Evaluates the inverse cumulative distribution function at the given points.

chainer/distribution.py
```diff
@@ -0,0 +1,420 @@
...
+
+        Returns:
+            ~chainer.Variable: Inverse cumulative distribution function value
+            evaluated at `x`.
+
+        """
+        raise NotImplementedError
+
+    def log_cdf(self, x):
+        """Evaluates the log of cumulative distribution function at a given points.
```
bkvogel: "at a given points" -> "at the given points". Please also make the same change in methods below.

chainer/distribution.py
```diff
@@ -0,0 +1,420 @@
...
+            final_shape = sample_shape + final_shape
+        samples = self.sample_n(n)
+        return samples.reshape(final_shape)
+
+    def sample_n(self, n):
+        """Samples n random points from the distribution.
+
+        This function returns sampled points whose shape is
+        `(n, batch_shape, event_shape)`. On implementing sampling codes in an
+        inherited ditribution class, it is preferable to override this method.
```
bkvogel: Consider rewording, such as: When implementing sampling code in a subclass, it is recommended to override this method.

YoshikawaMasashi:
I fixed.

beam2d:
Jenkins, test this please.

YoshikawaMasashi:
fixed.

beam2d:
Jenkins, test this please.

beam2d:
Could you update `setup.py` to install `chainer.distributions`? It looks the CI fails for it.

YoshikawaMasashi:
Done.

beam2d:
Thank you.
Jenkins, test this please.

YoshikawaMasashi:
I fixed.

beam2d:
Thanks. Jenkins, test this please again.

PR #4779, opened 2018-05-15
Add backprop tests

toslunar:
I found the tests on the backprop routines were not very strong.  This PR adds
- `chainer.grad` tests at intermediate variables,
- `FunctionNode.backward_accumulate` tests.

### DIFF
--- a/tests/chainer_tests/test_variable.py
+++ b/tests/chainer_tests/test_variable.py
@@ -71,94 +71,91 @@ def muladd(a, b, c):
     return MulAdd().apply((a, b, c))[0]
 
 
-@testing.parameterize(*testing.product({
-    'var_a': [False, True],
-    'var_b': [False, True],
-    'var_c': [False, True],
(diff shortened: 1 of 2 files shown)


tests/chainer_tests/test_variable.py
```diff
@@ -41,6 +41,122 @@ def constant(xs, value):
...
+        return a * b + c,
+
+    def backward_accumulate(self, target_input_indexes, grad_outputs,
+                            grad_inputs):
+        a, b = self.get_retained_inputs()
+        g, = grad_outputs
+        ret = []
+        for i, g_in in zip(target_input_indexes, grad_inputs):
+            if g_in is None:
+                g_in = g * 0.  # sorry, it's slow.
```
okuta: Why don't you use `cuda.get_array_module(g).zeros_like(g.data)` ?

okuta: OK

PR #4725, opened 2018-05-08
Select current device using grads of outputs in addition to inputs during backward

Hakuyume:
fix #4669

PR #4780, opened 2018-05-16
[backport] Select current device using grads of outputs in addition to inputs during backward

kmaehashi:
Backport of #4725


beam2d:
LGTM

PR #4786, opened 2018-05-17
Sparse matrix example docs

hvy:
Some documentations and examples related to newly introduced sparse matrices.

### DIFF
--- a/docs/source/reference/util/sparse.rst
+++ b/docs/source/reference/util/sparse.rst
@@ -1,9 +1,19 @@
 Sparse utilities
 -----------------------------------
 
-Sparse matrix representations that support back-propagation for e.g. efficient
-sparse matrix multiplication and conversion functions between sparse and dense
-representations.
+A :class:`chainer.Variable` can be converted into a sparse matrix in e.g.
+COO (Coordinate list) format.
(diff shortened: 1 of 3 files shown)


docs/source/reference/util/sparse.rst
```diff
@@ -1,6 +1,10 @@
 Sparse utilities
 -----------------------------------
 
+Sparse matrix representations that support back-propagation for e.g. efficient
+sparse matrix multiplication and conversion functions between sparse and dense
+representations.
+
```
kmaehashi: How about adding a note like this to avoid confusing users that sparse matrix computations are available everywhere?
`Not all functions support use of sparse matrix. See F.sparse_matmul for the example usage.`

hvy: Fixed.

hvy:
@kmaehashi thanks for the feedback. I added a note to address your comment but also made the documentation a bit more verbose, as the previous one was a bit too much on the short side.

kmaehashi:
Thanks, LGTM!
Jenkins, test this please.

kmaehashi:
LGTM!

PR #4789, opened 2018-05-17
chainer.grad without priority queue

toslunar:
Refactor `chainer.grad` with the observation that the set of all backward calls can be calculated in the earlier steps.
Possible drawbacks:
- will share less codes between `Variable.backward` and `chainer.grad`;
- order of `func.backward` of the same `func.rank` will be non-deterministic

I added some `assert` lines for the review.  I used tests in #4779.


niboshi:
Thank you for the PR.
Could you clarify the motivation of this PR? Does it improve efficiency of backward computation?

toslunar:
My motivation came from that I felt difficulty in maintaining long code of `chainer.grad`.  I'd like to remove unnecessary code in it derived from `Variable.backward`, unless their routines are going to be shared in a module.

I have no motivation in the performance in this PR.  I've not compared the performances (speed or memory).  I believe at least that this change does not degrade the performance much.

niboshi:
OK. Thank you for clarification.

chainer/function_node.py
```diff
@@ -837,19 +838,13 @@ def grad(outputs, inputs, grad_outputs=None, grad_inputs=None, set_grad=False,
...
 
-def _backprop(outputs, inputs, grad_required, retain_grad, grads, loss_scale):
-    candidate_funcs, push_candidate, pop_candidate = _get_ordered_func_heap()
-
-    for y in outputs:
-        creator = y.creator_node
-        if creator is not None:
-            push_candidate(creator)
+def _backprop(outputs, inputs, visited_funcs, visited_vars,
+              retain_grad, grads, loss_scale):
```
niboshi: The names `visited_funcs` and `visited_vars` might be misleading in the scope of this function.
How about other naming? Some candidates I'v come up with:

* `subgraph_funcs`, `subgraph_vars`
* `graph_funcs`, `graph_vars`

chainer/function_node.py
```diff
@@ -747,55 +746,57 @@ def grad(outputs, inputs, grad_outputs=None, grad_inputs=None, set_grad=False,
...
 
-            if not x.requires_grad:
-                continue
-            forward_graph[x].append(func)
-            creator = x.creator_node
-            if creator is not None and creator not in visited_funcs:
-                candidate_funcs.append(creator)
+        # Raise error here if y is created by Function.backward.
+        # In such case, we don't know exact inputs of the creator.
+        y._check_old_style_gradient()
```
niboshi: Why did you move this check?
Shouldn't we check for all of the `func.inputs`?

chainer/function_node.py
```diff
@@ -747,55 +746,57 @@ def grad(outputs, inputs, grad_outputs=None, grad_inputs=None, set_grad=False,
...
+
+        func = y.creator_node
+        if func is None:
+            continue
+        if func not in forward_func_to_var:  # func is not visited
+            for x in func.inputs:
+                if x.requires_grad:
+                    candidate_vars.append(x)
```
niboshi: How about checking `x not in visited_vars` before `candidate_vars.append(x)`?
(Instead of checking `y in visited_vars` after dequeuing?)

chainer/function_node.py
```diff
@@ -747,55 +746,57 @@ def grad(outputs, inputs, grad_outputs=None, grad_inputs=None, set_grad=False,
...
             if func in visited_funcs:
                 continue
             visited_funcs.add(func)
-            for y_ref in func.outputs:
-                y = y_ref()
-                if y is not None and y in forward_graph:
-                    candidate_vars.append(y)
+            assert func in forward_func_to_var
+            for y in forward_func_to_var[func]:
+                candidate_vars.append(y)
```
niboshi: How about checking `y not in visited_vars` before `candidate_vars.append(y)` instead of checking after dequeuing?

chainer/function_node.py
toslunar: I moved it because `FunctionAdapter` says
```
# Intentionally not passing requires_grad=False so that
# backprop routines can raise an error when a further backprop
# is attempted against this gradient variable.
```

The original code didn't depend on this, and it was safer....

niboshi: >The original code didn't depend on this, and it was safer....

Why do you mean by that the original code didn't depend on that?
Now I reread the logic and I think the check is not changed. Is this understanding not correct?

PR #4793, opened 2018-05-21
[backport] Warn only if the BN input tensor is 2-dimensional

kmaehashi:
Backport of #4663


kmaehashi:
This backport is not need, as `axis` option of BatchNorm (https://github.com/chainer/chainer/pull/4266) was not backported.

PR #4798, opened 2018-05-22
Improve performance of batch normalization

anaruse:
This PR aims to improve performance of batch normalization on GPU.

Sometimes cuDNN batch normalization may not be so fast specifically when batch/channel size is small and image size is big. The "CUDNN_BATCHNORM_SPATIAL_PERSISTENT" mode that is supported by cuDNN 7.0 or later can improve the performance for that cases. With this PR, you can apply this mode to batch normalization layers in your models by just adding the line below.

    with chainer.using_config('cudnn_fast_batch_normalization', True):

Note that this PR requires https://github.com/cupy/cupy/pull/1260.


okuta:
jenkins, test this please

okuta:
@anaruse Cloud you add test?

anaruse:
I've just added tests for this PR. Please double check.

okuta:
jenkins, test this please.

okuta:
LGTM except document.
Please edit `configuration.rst`.

anaruse:
I've added explanation on how to enable fast implementation for batch normalization in cuDNN. I've also added environment variables `CHAINER_CUDNN_FAST_BATCH_NORMALIZATION` and `CHAINER_CUDNN_FAST_BN` for the default setting. Please review again.

chainer/__init__.py
```diff
@@ -147,7 +147,11 @@ def is_arrays_compatible(arrays):
 global_config.use_ideep = os.environ.get('CHAINER_USE_IDEEP', 'never')
 global_config.lazy_grad_sum = bool(int(
     os.environ.get('CHAINER_LAZY_GRAD_SUM', '0')))
-global_config.cudnn_fast_batch_normalization = False
+global_config.cudnn_fast_batch_normalization = bool(int(
+    os.environ.get('CHAINER_CUDNN_FAST_BATCH_NORMALIZATION', '0')))
+if not global_config.cudnn_fast_batch_normalization:
+    global_config.cudnn_fast_batch_normalization = bool(int(
+        os.environ.get('CHAINER_CUDNN_FAST_BN', '0')))
```
okuta: Why do you add this environment variable?

anaruse: Is your question about `CHAINER_CUDNN_FAST_BN`? That is because I thought `CHAINER_CUDNN_FAST_BATCH_NORMALIZATION` is bit too long and easy to make a mistake. But, at the same time, I also feel having two env-vars for the same purpose is rather redundant. What would you think?

anaruse:
I've update the PR. Could you review it again?

chainer/__init__.py
okuta: Thanks

okuta:
jenkins, test this please.

okuta:
LGTM!

PR #4799, opened 2018-05-22
Allow more than 2 axis in `F.normalize`

niboshi:
As the use of `np.linalg.norm` has been eliminated in #4763, we no longer have to limit the number of axis in `F.normalize`.

### DIFF
--- a/chainer/functions/normalization/l2_normalization.py
+++ b/chainer/functions/normalization/l2_normalization.py
@@ -57,17 +57,17 @@ def normalize(x, eps=1e-5, axis=1):
     This function implements L2 normalization on a vector along the given axis.
     No reduction is done along the normalization axis.
 
-    This function computes an output vector :math:`y` by the following
-    equation:
+    In the case when :obj:`axis=1` and :math:`\\mathbf{x}` is a matrix of
+    dimension :math:`(N, K)`, where :math:`N` and :math:`K` denote mini-batch
+    size and the dimension of the input vectors, this function computes an
+    output matrix :math:`\\mathbf{y}` of dimension :math:`(N, K)` by the
+    following equation:
 
     .. math::
-       y_{i,j} = {x_{i,j} \\over \\| x_{i,*} \\|_2 + \\epsilon}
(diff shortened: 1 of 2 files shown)


tests/chainer_tests/functions_tests/normalization_tests/test_l2_normalization.py
```diff
@@ -21,6 +21,8 @@
         {'shape': (4, 3, 2, 5), 'axis': 2},
         {'shape': (4, 3, 2, 5), 'axis': 3},
         {'shape': (4, 3, 2), 'axis': (0, 1)},
+        {'shape': (4, 3, 2, 4, 3, 2, 2), 'axis': (1, 4, 3, 6)},
+        {'shape': (0, 2), 'axis': 1},
```
toslunar: `NormalizeL2.forward` needs `force_array` in order to support `len(axis) == 0`.  (Test `{'shape': (), 'axis': ()},`.)

chainer/functions/normalization/l2_normalization.py
```diff
@@ -59,13 +57,14 @@ def normalize(x, eps=1e-5, axis=1):
     This function implements L2 normalization on a vector along the given axis.
     No reduction is done along the normalization axis.
 
-    In the case when :obj:`axis=1` and :math:`x` is a vector of dimension
-    :math:`(N, K)`, where :math:`N` and :math:`K` denote mini-batch size and
-    the dimension of the input variable, this function computes an output
-    vector :math:`y` by the following equation:
+    This function computes an output vector :math:`y` by the following
+    equation:
```
toslunar: The original paragraph showed the most typical example.  I'd like to preserve it somewhere in this docs.

tests/chainer_tests/functions_tests/normalization_tests/test_l2_normalization.py
niboshi: Thank you for pointing out. Fixed.
(Casting `eps` is needed to support ()-shape.)

chainer/functions/normalization/l2_normalization.py
niboshi: I reverted to the original formula, and updated some notation to make it clear `x_i` and `y_j` are vectors.

PR #4814, opened 2018-05-28
Fix a documenation bug in `TransformDataset`

### LINKED ISSUE #4655: Need to update documentation for `Chainer.datasets.TransformDataset`
When using `Chainer.datasets.TransformDataset` with `Chainer.training.Evaluator`, if you set the trigger for the `Evaluator` to be, e.g., `(100, 'iteration')`, it will cause the performance to significantly drops (at least for the value computed by the evaluator). 

For an example code to reproduce the problem, see https://gist.github.com/jinjiren/b70ce3516d2790923f9f0ea4d9a18250
Sample results:
image

If you comment out the code related to `TransformDataset`, then it will behave normally:
image

If you change the interval for the trigger of `Evaluator`, the behavior will also change:
When changed to `(1, 'epoch')` (default value), result becomes:
image
Seems OK but actually is worse than no `TransformDataset`.

This is a bug confused me for long. And it took me several hours to identify the problem. Currently I found that `Evaluator` with `Transform Dataset` seems to affect the performance, and the more frequently you use the `Evaluator`, the worse performance you get.

bkvogel:
fix #4655.

### DIFF
--- a/chainer/datasets/transform_dataset.py
+++ b/chainer/datasets/transform_dataset.py
@@ -14,14 +14,19 @@ class TransformDataset(dataset_mixin.DatasetMixin):
 
     The function :obj:`transform` takes, as an argument, :obj:`in_data`, which
     is the output of the base dataset's :meth:`__getitem__`, and returns
-    the transformed arrays as output. Please see the following example.
+    the transformed arrays as output. Please see the following example. Since
+    :obj:`in_data` directly refers to the item in the dataset, take care that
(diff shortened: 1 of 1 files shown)


kmaehashi:
I think it is OK to just fix the documentation, because
* we don't know how to properly copy the element of upstream dataset
* copying introduces additional overhead

kmaehashi:
LGTM!

PR #4816, opened 2018-05-28
Test `F.layer_normalization` with large eps

toslunar:
Thanks to #4744, test with larger `eps` passes, which I mentioned at https://github.com/chainer/chainer/pull/4744#issuecomment-388045474.  I also confirmed the new test failed before merging the fix.


toslunar:
I fixed syntax errors in Python 2.

bkvogel:
LGTM!

PR #4818, opened 2018-05-28
Correct docs of batch_{re}normalization about running averages

grafi-tt:
Current document is quite misleading. The link `chainer.links.BatchNormalization` does not access to `running_mean` and/or `running_var`; instead it relies on in-place updates. But another link `chainer.links.BatchRenormalization` indeed accesses to the attributes.


toslunar:
Jenkins, test this please.

PR #4821, opened 2018-05-29
[WIP] Add `inplace` option to `F.relu`

mitmul:
This PR adds `inplace` option to `F.relu` function.

To remove [WIP] tag, we need to check the input is not used anywhere else in the computational graph. This is just a tweak for a benchmark, currently.


bkvogel:
I like this idea! But I think it might sometimes not be safe to replace the input array by the result. For example, what if the input is also needed by another function? Would it make sense to mark the input array (or data attribute of input variable) as read-only in that case to prevent modification?

PR #4822, opened 2018-05-29
Avoid mutable default arguments

kmaehashi:
Fix arguments not fixed in #4419.


niboshi:
How about using different default values from `None`?

niboshi:
@kmaehashi I think it's better to use some singleton value (like `numpy._NoValue`).

okuta:
How about using `()`?

niboshi:
I don't think `()` is good because one may think it has some meaning.

beam2d:
I prefer `'default'` in this case. It's much easier to use. We don't need to distinguish if the user passed the argument explicitly (distinguishing them is the purpose of `numpy._NoValue`).

kmaehashi:
Currently `variable_style` of `build_computational_graph` behaves as follows:

* When dict is specified: use the dict as option
* When None is specified: use Chainer v1-style option (ie `{}`)
* When no value is specified: use default option (ie `{'shape': 'octagon', 'fillcolor': '#E0E0E0', 'style': 'filled'}`)

My proposal is:

* When dict is specified: use the dict as option
* When None is specified: use default option (`{'shape': 'octagon', 'fillcolor': '#E0E0E0', 'style': 'filled'}`)
(ie, users currently specifying Chaienr-v1 style (None) should specify `{}` instead.)

@niboshi -san's proposal is to prepare singleton immutable object to keep the current spec while removing the mutable default arguments.

beam2d:
My suggestion was:
- When dict is specified: use the dict as option.
- When `None` is specified: use Chainer v1-style option, or raise an error (if we choose to remove this feature).
- When `'default'` is specified: use the default option.
- When nothing is specified: use the default option.
What I meant is that we do not need to distinguish between the last two cases.

beam2d:
I rather thought a "hidden default value" is far more complex for users. For example, consider a user who wants to write a wrapper of this function that passes through this argument. If the function is using hidden default value semantics, the user has to emulate this behavior using again a hidden default value (possibly using the "hidden" value by accessing an private API or inventing another default value which needs to be translated in the wrapper function).

kmaehashi:
Thanks for the discussion.
@niboshi If your point is to keep current spec as-is, I'd like to go with `'default'`. Do you think it's OK?

niboshi:
Sorry for delay.
I think it's ok to proceed with `'default'`.

kmaehashi:
Updated according to the discussion above.

chainer/computational_graph.py
```diff
@@ -53,7 +55,9 @@ def label(self):
 
 class ComputationalGraph(object):
 
-    """Class that represents computational graph.
+    """__init__(self, nodes, edges, variable_style=_var_style, function_style=_func_style, rankdir='TB', remove_variable=False, show_name=True)
```
toslunar: Could you avoid `NOQA` as #6184?

kmaehashi: Thanks, fixed.

chainer/computational_graph.py
```diff
@@ -190,10 +208,12 @@ def _skip_variable(nodes, edges):
 
 
 def build_computational_graph(
-        outputs, remove_split=True, variable_style=_var_style,
-        function_style=_func_style, rankdir='TB', remove_variable=False,
+        outputs, remove_split=True, variable_style=_default,
+        function_style=_default, rankdir='TB', remove_variable=False,
         show_name=True):
-    """Builds a graph of functions and variables backward-reachable from outputs.
+    """build_computational_graph(outputs, remove_split=True, variable_style=_var_style, function_style=_func_style, rankdir='TB', remove_variable=False, show_name=True)
```
toslunar: ditto 🙇

kmaehashi: fixed :bow:

chainer/computational_graph.py
```diff
@@ -190,10 +208,14 @@ def _skip_variable(nodes, edges):
 
 
 def build_computational_graph(
-        outputs, remove_split=True, variable_style=_var_style,
-        function_style=_func_style, rankdir='TB', remove_variable=False,
+        outputs, remove_split=True, variable_style=_default,
+        function_style=_default, rankdir='TB', remove_variable=False,
```
niboshi: Let's use the explicit value `'default'` here and make it documented.

chainer/training/extensions/parameter_statistics.py
```diff
@@ -8,8 +8,24 @@
...
+    'max': lambda x: backend.get_array_module(x).max(x),
+    'zeros': lambda x: backend.get_array_module(x).count_nonzero(x == 0),
+    'percentile': lambda x: backend.get_array_module(x).percentile(
+        x, (0.13, 2.28, 15.87, 50, 84.13, 97.72, 99.87))
+}
+
+
 class ParameterStatistics(extension.Extension):
-    """Trainer extension to report parameter statistics.
+    """__init__(self, links, statistics=_default_statistics, \
```
niboshi: Ditto.

chainer/computational_graph.py
```diff
@@ -53,7 +55,11 @@ def label(self):
 
 class ComputationalGraph(object):
 
-    """Class that represents computational graph.
+    """__init__(self, nodes, edges, variable_style=_var_style, \
```
niboshi: Ditto.

chainer/computational_graph.py
toslunar: Then I'd like to have the doc like
```
        function_style(dict or 'default'): Dot node style for function.
            The ``'default'`` style is
            ``{'shape': 'record', 'fillcolor': '#6495ED', 'style': 'filled'}``.
```

I'm wondering if we can do it without repeating the default values.

kmaehashi:
@niboshi I'd like to discuss *API expansion* (`'default'` support) in another issue if needed. It's out of the scope of this PR. I'd also like to backport this.

niboshi:
Hmm, in my understanding the purpose of using `str` argument is to make it explicit to users.

https://github.com/chainer/chainer/pull/4822#issuecomment-400913963
>I rather thought a "hidden default value" is far more complex for users. For example, consider a user who wants to write a wrapper of this function that passes through this argument. If the function is using hidden default value semantics, the user has to emulate this behavior using again a hidden default value (possibly using the "hidden" value by accessing an private API or inventing another default value which needs to be translated in the wrapper function).

Also I don't necessarily think this PR needs backporting, because it does not itself improve anything, but it's a safeguard for future bug introduction.

toslunar:
If #4413 will not be backported, it doesn't look too bad to change v7 only.  Almost all the new PRs are into `master` branch.

kmaehashi:
> Hmm, in my understanding the purpose of using str argument is to make it explicit to users.

Ok, then how about:
* renaming `_default` to `DEFAULT` so that users can use it
* make it documented
* do not backport
Or do you prefer to use `'default'` instead of `DEFAULT`?

> If #4413 will not be backported, it doesn't look too bad to change v7 only. Almost all the new PRs are into master branch.

I agree with this.

niboshi:
@kmaehashi 
Do you mean introducing `chainer.DEFAULT`?
Hmm, I don't think it's pythonic (or Chainer-ish. I think `str` is more common for this kind of arguments).

kmaehashi:
Updated according to the offline discussion.

chainer/training/extensions/parameter_statistics.py
```diff
@@ -30,9 +41,12 @@ class ParameterStatistics(extension.Extension):
         links (~chainer.Link or iterable of ~chainer.Link): Link(s) containing
             the parameters to observe. The link is expected to have a ``name``
             attribute which is used as a part of the report key.
-        statistics (dict): Dictionary with function name to function mappings.
+        statistics (dict or 'default'): Dictionary with function name to
+            function mappings.
             The name is a string and is used as a part of the report key. The
             function is responsible for generating the statistics.
+            If the special value ``'default'`` is specified, use the default
+            statistics functions.
```
niboshi: Tiny suggestion:
```diff
- use the default statistics functions.
+ the default statistics functions will be used.
```

chainer/computational_graph.py
```diff
@@ -205,9 +230,14 @@ def build_computational_graph(
             :class:`~chainer.FunctionNode` object.
         remove_split(bool): It must be ``True``. This argument is left for
             backward compatibility.
-        variable_style(dict): Dot node style for variable.
-            Possible keys are 'shape', 'color', 'fillcolor', 'style', and etc.
-        function_style(dict): Dot node style for function.
+        variable_style(dict or 'default'): Dot node style for variable.
+            Possible keys are 'shape', 'color', 'fillcolor', 'style' etc.
+            If the special value ``'default'`` is specified, use the default
```
niboshi: ditt.

chainer/computational_graph.py
```diff
@@ -205,9 +230,14 @@ def build_computational_graph(
...
+            configuration.
+        function_style(dict or 'default'): Dot node style for function.
+            Possible keys are 'shape', 'color', 'fillcolor', 'style' etc.
+            If the special value ``'default'`` is specified, use the default
```
niboshi: ditto.

chainer/training/extensions/parameter_statistics.py
```diff
@@ -52,17 +79,9 @@ class ParameterStatistics(extension.Extension):
...
-        'std': lambda x: backend.get_array_module(x).std(x),
-        'min': lambda x: backend.get_array_module(x).min(x),
-        'max': lambda x: backend.get_array_module(x).max(x),
-        'zeros': lambda x: backend.get_array_module(x).count_nonzero(x == 0),
-        'percentile': lambda x: backend.get_array_module(x).percentile(
-            x, (0.13, 2.28, 15.87, 50, 84.13, 97.72, 99.87))
-    }
-
-    def __init__(self, links, statistics=default_statistics,
+    default_statistics = _default_statistics
```
niboshi: Is this attribute needed?

chainer/training/extensions/parameter_statistics.py
kmaehashi: Fixed as suggested.

chainer/training/extensions/parameter_statistics.py
kmaehashi: I think its convenient for users.
Let's discuss the topic in the separate issue if needed.

niboshi: OK.

niboshi:
Jenkins, test this please

PR #4770, opened 2018-05-14
Add __call__ method to Link for backward compatibility

mitmul:
This PR adds `__call__` to Link. This is for backporting to v4, so once this is merged, the `__call__` method added to Link in v5 will be overwritten by #4735.

PR #4823, opened 2018-05-29
Revert "Replace all __call__ methods in Links with forward"

kmaehashi:
Reverts chainer/chainer#4735
https://github.com/chainer/chainer/pull/4770#issuecomment-392801354


niboshi:
LGTM!

PR #4825, opened 2018-05-31
adjust test input for F.min/max

kmaehashi:
Fix F.min/max error of #3766

### DIFF
--- a/tests/chainer_tests/functions_tests/math_tests/test_minmax.py
+++ b/tests/chainer_tests/functions_tests/math_tests/test_minmax.py
@@ -13,7 +13,7 @@ from chainer.testing import attr
 class TestMax(unittest.TestCase):
 
     def setUp(self):
-        eps = 1e-5
+        eps = 1e-6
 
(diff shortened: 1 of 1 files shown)


tests/chainer_tests/functions_tests/math_tests/test_minmax.py
```diff
@@ -13,7 +13,7 @@
 class TestMax(unittest.TestCase):
 
     def setUp(self):
-        eps = 1e-5
+        eps = 1e-6
```
toslunar: This makes the test harder.

toslunar: Sorry, my previous comment wasn't true.

- The precondition `(self.x > (numpy.max(self.x) - 2 * eps)).sum() == 1` becomes weaker.
- The backward check with `'eps': eps` might be more accurate.

toslunar:
@kmaehashi Could you review #4838 instead?

PR #4829, opened 2018-05-31
Fix ResNetLayers pass wrong kwargs to Convolution2D

kmaehashi:
`downsample_fb` should not be passed to `Convolution2D`.
This fixes error introduced in #4479.


okuta:
LGTM!

PR #4830, opened 2018-05-31
v4: fix required version of CuPy

### LINKED ISSUE #4084: Required CuPy version specification in the stable version looks incomplete
Current minor versions of Chainer requires CuPy (if installed) older than the unseen next minor version. For example, the recent release of Chainer v3.2.0 requires CuPy <2.3. I think it should be fixed so that 1) it does not allow too old CuPy, and 2) it allows any newer version of CuPy at the same major version.

kmaehashi:
This PR is for stable (v4) branch.

Fix #4084.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,7 +21,7 @@ set CHAINER_PYTHON_350_FORCE environment variable to 1."""
 
 
 def cupy_requirement(pkg):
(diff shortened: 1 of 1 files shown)


okuta:
LGTM!

PR #4831, opened 2018-05-31
Improve performance of sparse_matmul

anaruse:
This PR aims to improve the performance of `sparse_matmul`.

The current `sparse_matmul` for GPU are implemented on NO assumption about how sparse matrix is being sorted. Therefore, memory access conflicts by atomic operations may occur badly and the performance may decrease. This performance degradation can be mitigated if there is kind of information about how the sparse matrix is sorted.

This PR adds an option `row_major` to a class `CooMatrix` so that you can add a hint on how the sparse matrix is being sorted. If you specify `row_major=True` or `row_major=False` when you make a sparse matrix using `CooMatrix`, then the sparse matrix is assumed to be sorted in row-major or column-major format respectively, and internal functions will use that information to improve performance.


anaruse:
I'm attaching the performance measurement results on a P100 GPU of some GCN model which has sparse matrix with 60K nodes and 10M non-zero elements. As you can see, the performance is improved by about 2.5x with this PR.

[ with sparse_matmul in current master ]
image

[ with sparse_matmul in this PR ]
image

hvy:
Thanks for this PR. Rhe performance gain seems quite significant. Are there any trade-offs that the user should be aware of?

chainer/functions/math/sparse_matmul.py
```diff
@@ -13,8 +13,8 @@
     _scipy_available = False
 
 
-def _coo_matmul(sp_data, sp_row, sp_col, sp_shape, dn, transa, transb, transc,
-                dtype=None):
+def _coo_matmul(sp_data, sp_row, sp_col, sp_shape, sp_row_major,
```
hvy: How about using the term `c_contiguous` to align with numpy (i.e. `np.arange(...).flags.c_contiguous` ?

anaruse: Thanks for your suggestion. I'd like to use one that is easier to understand for python/numpy developers. Regarding an argument of class `CooMatrix`, would you think it is better to use `c_contiguous` instead of `row_major` for that as well?

anaruse:
Thanks for your comment. Re: trade-offs, I don't think there is remarkable performance drop even when users misuse the option `row_major` incorrectly.

chainer/functions/math/sparse_matmul.py
hvy: I think so, basically replacing all occurrences of `row_major` with `c_contiguous` ?

hvy: Assuming they actually mean the same thing of course. I have not understood the implementation details of this optimization but I also understood that you cannot simply look at the memory layout of the given `ndarray` ?

anaruse: Thank you! I did not think about checking if it is c_contiguous or f_contiguous from the given ndarray. Actually it is probably possible with O(N) complexity.

anaruse:
I changed the option name from `row_major` to `c_contiguous` so that is becomes more intuitive for python/numpy developers. I also added a method `_is_c_contiguous` which automatically check if a coo matrix is C-contiguous, Fortran-contiguous or another.

chainer/functions/math/sparse_matmul.py
hvy: So maybe we should reconsider the name, loop up how e.g. scipy does it. There's also reference from CuPy! https://docs-cupy.chainer.org/en/stable/reference/generated/cupy.sparse.csr_matrix.html#cupy.sparse.csr_matrix.toarray Some suggestions that has been brought up already are `c_order`.

anaruse: I went through scipy code (https://github.com/scipy/scipy/blob/v1.1.0/scipy/sparse/coo.py). Actually there is an argument named `order` in `def toarray(...)`, though it is about order of a dense matrix to be created from a coo sparse matrix.  Anyway, it seems to me that `order='C'/'F'/None` or `c_order=True/False/None` is good. Either would be fine to me, but my preference is rather `c_order`.

hvy: Could you explain why you'd prefer `c_order` ? I think it's good to follow scipy (and cupy). Then we don't have to "introduce" a new concept.

anaruse: I have no strong reason on that. I just did not think `c_order` was a really new concept. Anyway, I'm OK to go with `order`.

hvy: Sorry, maybe not a new concept. Neither do I have a very strong opinion. I think the main points are: 
- Will the default behavior be "None" / "no assumption"? 
- Are there possibilities for other orders except C and F?

hvy: I do think though that `order=None` looks more clear than `c_order=None`.

anaruse: I agree with you on that `order=None` looks more clear.

Re: possibilities for other orders, I don't think we need to consider other orders except **'C'** and **'F'** as long as it is a matrix (2D array). BTW, this is not order, but it may be better to add kind of **'auto'** to the options, and the order is automatically checked by looking at row and column index arrays only when **'auto'** is explicitly specified by users. What would you think?

anaruse:
I've just changed the argument name from `c_contiguous` to `order`, etc. for now. Please take a look at it.

chainer/functions/math/sparse_matmul.py
```diff
@@ -23,10 +23,16 @@ def _coo_matmul(sp_data, sp_row, sp_col, sp_shape, dn, transa, transb, transc,
         A_row = sp_col
         A_col = sp_row
         A_shape = (sp_shape[1], sp_shape[0])
+        A_order = sp_order
+        if sp_order is 'C':
+            A_order = 'F'
+        elif sp_order is 'F':
+            A_order = 'C'
```
hvy: How about replacing the line above (`A_order = sp_order`) with an `else:` here, avoiding possible reassignment. Is it also the case that it's always `None`? Then we could be explicit about that too.

chainer/functions/math/sparse_matmul.py
```diff
@@ -90,49 +97,70 @@ def _coo_matmul_gpu(A_data, A_row, A_col, A_shape, B, dtype):
...
 
-    nthreads = nb * ldnz * _n
-    _cupy_coo_matmul()(nb, _m, _n, _k, ldnz, A_data, A_row, A_col, B, C,
+    chunk = 1
+    if A_order is 'C':
+        # A chunk is the number of non-zero elements handled by a single GPU
+        # thread. If contiguous non-zero elemets are related to the same
+        # location of the output matrix and they are processed in the same
+        # thread, number of atomic-add operations can be reduced.
+        chunk = max(ldnz // _m, 1)
```
hvy: Again, let's stick to the `else: chunk = 1` idiom.

chainer/utils/sparse.py
```diff
@@ -14,6 +14,16 @@ class CooMatrix(object):
...
         shape (tuple of int): The shape of the matrix in dense format.
+        order ('C', 'F', 'auto', None): If ``'C'``, the matrix is assumed to be
+            made from row-major format (AKA C-contiguous) array, in other
+            words, the row indices are sorted. If ``'F'``, the matrix is
+            assumed to be made from column-major format (AKA
+            Fortran-contiguous) array, in other words, the column indices are
+            sorted. If ``None``, the matrix is assumed as neither 'C' order nor
+            'F' order (this is the default). If ``'auto'``, the matrix is
+            automatically checked if it is 'C' order, 'F' order or another.
+            This information will be used by some functions like sparse_matmul
```
hvy: Could you make `sparse_matmul` a doc reference?

chainer/functions/math/sparse_matmul.py
```diff
@@ -419,13 +454,13 @@ def sparse_matmul(a, b, transa=False, transb=False):
     """
```
hvy: Do you think we should document that 'C' order sparse matrices are faster on GPUs (for instance using `.. note::`)?

chainer/utils/sparse.py
```diff
@@ -106,7 +126,13 @@ def to_coo(x, ldnz=None, requires_grad=False):
         row[:nnz] = xp.array(_row).astype(xp.int32)
         col[:nnz] = xp.array(_col).astype(xp.int32)
         shape = x.shape
-        return CooMatrix(data, row, col, shape, requires_grad)
+        order = None
+        if x.flags['C_CONTIGUOUS']:
```
hvy: A small detail but you can also use the `x.flags.c_contiguous` attr.

chainer/utils/sparse.py
```diff
@@ -123,6 +149,36 @@ def to_coo(x, ldnz=None, requires_grad=False):
             row[i] = coo.row
             col[i] = coo.col
         shape = x.shape[1:]
-        return CooMatrix(data, row, col, shape, requires_grad)
+        order = None
+        if x.flags['C_CONTIGUOUS']:
```
hvy: Ditto. Or maybe better if you could unify the return statements for case ndim == 2 and 3. For DRY.

chainer/utils/sparse.py
```diff
@@ -123,6 +149,36 @@ def to_coo(x, ldnz=None, requires_grad=False):
...
+    row_diff = xp.zeros(row.shape, dtype=row.dtype)
+    row_diff[1:] = row[1:] - row[:-1]
+    if xp.amin(row_diff) < 0:
+        return False
+    col_diff = xp.zeros(col.shape, dtype=col.dtype)
+    col_diff[1:] = col[1:] - col[:-1]
+    col_diff[(row_diff > 0)] = 0
+    if xp.amin(col_diff) < 0:
+        return False
+    return True
```
hvy: How about `return xp.amin(col_diff) >= 0` ?

anaruse:
Thank you for your detail comments, @hvy . I've just updated the PR. Please review it again.

chainer/utils/sparse.py
```diff
@@ -32,10 +43,20 @@ def __init__(self, data, row, col, shape, requires_grad=False):
             raise ValueError('length of shape must be 2.')
         if not (shape[0] > 0 and shape[1] > 0):
             raise ValueError('numbers in shape must be greater than 0.')
+        if not (order in ('C', 'F', 'auto', None)):
```
hvy: A small detail but I don't think you need the outer parenthesis.

chainer/utils/sparse.py
```diff
@@ -123,6 +151,29 @@ def to_coo(x, ldnz=None, requires_grad=False):
...
             col[i] = coo.col
         shape = x.shape[1:]
-        return CooMatrix(data, row, col, shape, requires_grad)
+        return CooMatrix(data, row, col, shape, order=order,
+                         requires_grad=requires_grad)
     else:
         raise ValueError('ndim of x must be 2 or 3.')
+
+
+def _is_c_order(row, col):
```
hvy: I'm wondering if we should make this function public (and test it) and make it return the order instead, i.e. either `'C'` , `'F'` or `None`. User may call it outside once and reuse the order when constructing additional CooMatrices?

hvy: Not sure if it's going to be repeatedly constructed with same `row`, `col` though. What do you think? In any case, I'd personally like to test this function.

hvy:
Thanks for the fixes and sorry for the delayed response. I added some comments and verified with @beam2d that the current interface regarding `order` is good. 

Could you take another look at the comments and maybe resolve the conflicts?

chainer/utils/sparse.py
anaruse: Thanks! will remove the outer parenthesis.

chainer/utils/sparse.py
anaruse: Thank you for your comment. Actually, I though that `'auto'` did not align well with others like `'C'` or `'F'`, because it did not show anything about order. So, I think it is good idea to make the function public and let users call it. And probably it is better to change the name? How about `get_order`?

anaruse: Re: cost of computing the order, the compute complexity is O(nnz) as I mentioned before (nnz is number of non-zero elements), and I think the cost is smaller than the gain in most cases. I though `'auto'` was not your preference, but If you are OK with it, I will make it the default.

hvy: Yes, I first thought adding unintended overhead was not ideal but if the gain is in general larger then that might be better. Could you proceed with the change?

hvy: An idea that came up was making `None` behave like `'auto'` and instead introducing a new value such as `'other'`.

anaruse: I think it is good idea to use `'other'`. OK, I will change the code in that way.

hvy: Got it, thanks a lot.

anaruse:
@hvy : Could you take a look at it again? I've just finished the code fix.

anaruse:
It looks some errors happened where this PR is not related...

tests/chainer_tests/utils_tests/test_sparse.py
```diff
@@ -54,4 +54,58 @@ def test_to_dense(self):
...
+        assert(utils.get_order(col, row) == 'F')
+
+
+class TestGetOrder2(unittest.TestCase):
+
+    def test_other_order(self):
+        row = numpy.array((0, 2, 4, 1, 3), dtype=numpy.int32)
+        col = numpy.array((1, 3, 0, 2, 4), dtype=numpy.int32)
+        assert(utils.get_order(row, col) == 'other')
+        assert(utils.get_order(col, row) == 'other')
```
hvy: Could you prefer `assert utils.get_order(col, row) == 'other'`?

tests/chainer_tests/utils_tests/test_sparse.py
```diff
@@ -54,4 +54,58 @@ def test_to_dense(self):
...
+
+    def test_diag_order(self):
+        row = numpy.array((0, 0, 1, 1, 2), dtype=numpy.int32)
+        col = numpy.array((0, 1, 1, 2, 2), dtype=numpy.int32)
+        assert(utils.get_order(row, col) == 'C')
+        assert(utils.get_order(col, row) == 'C')
```
hvy: Ditto.

chainer/functions/math/sparse_matmul.py
```diff
@@ -90,49 +98,71 @@ def _coo_matmul_gpu(A_data, A_row, A_col, A_shape, B, dtype):
...
-    _cupy_coo_matmul()(nb, _m, _n, _k, ldnz, A_data, A_row, A_col, B, C,
+    if A_order is 'C':
+        # A chunk is the number of non-zero elements handled by a single GPU
+        # thread. If contiguous non-zero elemets are related to the same
+        # location of the output matrix and they are processed in the same
+        # thread, number of atomic-add operations can be reduced.
+        chunk = max(ldnz // _m, 1)
+    else:
+        chunk = 1
+    nthreads = ((nb * ldnz + chunk - 1) // chunk) * _n
```
hvy: I think you can remove the outer parenthesis.

chainer/utils/sparse.py
```diff
@@ -14,6 +14,14 @@ class CooMatrix(object):
         col (numpy.ndarray or cupy.ndarray): The column indices of the matrix
             entries.
         shape (tuple of int): The shape of the matrix in dense format.
+        order ('C', 'F', 'other', None): If ``'C'``, the maxtix is assumed that
```
hvy: How about adding "or" between `'other'` and `None` to be less ambiguous and more consistent.

anaruse:
Thank you for your comment. The branch has been updated. Please check it again.

hvy:
jenkins, test this please.

hvy:
jenkins, test this please.

chainer/utils/sparse.py
```diff
@@ -123,6 +138,55 @@ def to_coo(x, ldnz=None, requires_grad=False):
...
+    if _is_c_order(row, col):
+        return 'C'
+    if _is_c_order(col, row):
+        return 'F'
+    return 'other'
+
+
+def _is_c_order(row, col):
+    """Check if a coo matrix with given row and col is c_order"""
+    if not (row.shape == col.shape):
```
hvy: How about `if row.shape != col.shape:`?

tests/chainer_tests/utils_tests/test_sparse.py
```diff
@@ -54,4 +54,58 @@ def test_to_dense(self):
...
+            x0[0, self.shape[1]-1] = 1.
+            x0[self.shape[0]-1, 0] = 1.
+        sp_x = utils.to_coo(x0)
+        row = sp_x.row
+        col = sp_x.col
+        assert utils.get_order(row, col) == 'C'
+        assert utils.get_order(col, row) == 'F'
+
+
+class TestGetOrder2(unittest.TestCase):
```
hvy: Should we add tests with `F` order matrices?

hvy:
Again, sorry for the delayed response. I think it looks good except some small nitpicks mentioned above!

tests/chainer_tests/utils_tests/test_sparse.py
anaruse: I think that is already tested in `TestGetOrder`.

chainer/utils/sparse.py
anaruse: got it. will fix in that way.

anaruse:
Sorry for my delayed response. I've just updated the branch. Please check again.

hvy:
jenkins, test this please.

hvy:
LGTM!

PR #4837, opened 2018-06-01
Simplify classification custom loop example

okuta:
rel #3832

### DIFF
--- a/examples/mnist/train_mnist_custom_loop.py
+++ b/examples/mnist/train_mnist_custom_loop.py
@@ -73,12 +73,10 @@ def main():
 
         while train_iter.epoch < args.epoch:
             batch = train_iter.next()
-            x_array, t_array = convert.concat_examples(batch, args.gpu)
-            x = chainer.Variable(x_array)
-            t = chainer.Variable(t_array)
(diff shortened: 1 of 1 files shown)


examples/mnist/train_mnist_custom_loop.py
```diff
@@ -88,18 +86,14 @@ def main():
...
-                        for batch in test_iter:
-                            x, t = convert.concat_examples(batch,
-                                                           args.gpu)
-                            x = chainer.Variable(x)
-                            t = chainer.Variable(t)
-                            loss = model(x, t)
-                            sum_loss += float(loss.data) * len(t.data)
-                            sum_accuracy += (float(model.accuracy.data) *
-                                             len(t.data))
+                # This is optional but can reduce computational overhead.
```
kmaehashi: I think it's better to say that only `enable_backprop` is optional. (setting `train` to `False` is not optional)

okuta: Thanks! I will fix

okuta:
I fixed.

kmaehashi:
LGTM!

PR #4838, opened 2018-06-01
Fix input checks of `TestMin` and `TestMax`

toslunar:
The checks have to be aware of `axis`:
```
y = x.max(axis=self.axis, keepdims=True)
if numpy.all((x > y - 2 * eps).sum(axis=self.axis) == 1):
    # ok
```

### DIFF
--- a/tests/chainer_tests/functions_tests/math_tests/test_minmax.py
+++ b/tests/chainer_tests/functions_tests/math_tests/test_minmax.py
@@ -29,11 +29,13 @@ class TestMax(unittest.TestCase):
 
         # Sample x with single maximum value
         while True:
-            self.x = numpy.random.uniform(
-                -1, 1, (3, 2, 4)).astype(numpy.float32)
-            if (self.x > (numpy.max(self.x) - 2 * eps)).sum() == 1:
-                self.y_expect = numpy.max(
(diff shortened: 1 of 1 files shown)


tests/chainer_tests/functions_tests/math_tests/test_minmax.py
```diff
@@ -190,163 +108,80 @@ def test_duplicate_axis(self):
             functions.Max((0, 0))
 
     def test_pos_neg_duplicate_axis(self):
+        x_data = numpy.random.uniform(-1, 1, (3, 2, 4)).astype(numpy.float32)
+        x = chainer.Variable(x_data)
         with self.assertRaises(ValueError):
-            self.x.max(axis=(1, -2))
```
kmaehashi: Good catch...

kmaehashi:
LGTM!

kmaehashi:
This is a fix F.min/max error of #3766

tests/chainer_tests/functions_tests/math_tests/test_minmax.py
toslunar: Thanks.  `ValueError` has been raised by `numpy.ndarray.max`, not by `chainer.functions.max`.

PR #4845, opened 2018-06-04
Remove wrong length check from `GetItem.check_type_forward`

toslunar:
Because of boolean masks and ellipsis, it's not so easy to check if the index tuple of slices is valid.

### DIFF
--- a/chainer/functions/array/get_item.py
+++ b/chainer/functions/array/get_item.py
@@ -32,9 +32,6 @@ class GetItem(function_node.FunctionNode):
 
     def check_type_forward(self, in_types):
         type_check.expect(in_types.size() == 1)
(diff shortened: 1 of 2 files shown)


toslunar:
It costs to tell whether an index (of the indices) is a mask.  While `cupy.ndarray.__getitem__` does not (yet?) support a list index, the following code is valid in cpu mode.
```
>>> x = chainer.Variable(np.arange(24).reshape(4, 3, 2).astype('f'))
>>> x[([True, False, True, False],)]
variable([[[ 0.,  1.],
           [ 2.,  3.],
           [ 4.,  5.]],

          [[12., 13.],
           [14., 15.],
           [16., 17.]]])
```

niboshi:
OK.
LGTM!

PR #4846, opened 2018-06-04
Remove deprecated imp.load_source in setup.py

vilyaair:
I want to circumvent [imp.load_source](https://docs.python.org/3/library/imp.html), the deprecated module and undocumented function in Python3, via using exec, whereas such change may reduce code readability.


kmaehashi:
LGTM.
Jenkins, test this please.

@vilyaair Could you make a PR for CuPy to fix this?

kmaehashi:
Jenkins, test this please.

PR #4855, opened 2018-06-06
Improve function reference and documentation

kmaehashi:
Related to #4052.

### DIFF
--- a/chainer/links/connection/mlp_convolution_2d.py
+++ b/chainer/links/connection/mlp_convolution_2d.py
@@ -33,7 +33,12 @@ class MLPConvolution2D(link.ChainList):
         pad (int or pair of ints): Spatial padding width for input arrays at
             the first convolution layer. ``pad=p`` and ``pad=(p, p)`` are
             equivalent.
-        activation (:doc:`wrapper function </reference/functions>`): Activation function for internal hidden units.
+        activation (callable):
+            Activation function for internal hidden units.
+            You can pass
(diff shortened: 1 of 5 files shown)


niboshi:
It misses two requirements pointed out in https://github.com/chainer/chainer/pull/4052#issuecomment-350364224?

toslunar:
Yes.  I'd like to have something like the following:
1. "to backprop through the Link, the output of activation function should be a Variable whose gradient  is `backward`-ed to a gradient of the input"
2. "parameters in the activation function are not caught by the Link"

kmaehashi:
Thanks for comments!

> 1. "to backprop through the Link, the output of activation function should be a Variable whose gradient is backward-ed to a gradient of the input"

Fixed by saying: `Returned value must be a Variable derived from the input Variable to perform backpropagation on the variable.`

> 2. "parameters in the activation function are not caught by the Link"

Sorry, I couldn't get the point of this comment. What kind of mistakes are expected without this description?

toslunar:
>> 2.

`MLPConvolution2D(..., activation=chainer.links.PReLU)` is a mistake.

kmaehashi:
Thanks, I understood the concerns. I added

```
            It should not be an
            :doc:`activation functions with parameters </reference/links>`
            (i.e., :class:`~chainer.Link` instance).
```

toslunar:
Jenkins, test this please.

PR #4862, opened 2018-06-07
Step shift

jinjiren:
Add StepShift to chainer extensions. Unittests and docs have been added.


jinjiren:
I made a mistake, this PR contains code change for another PR...

PR #4863, opened 2018-06-07
Set output grad in `TestCUDAProfileHook`

toslunar:
Currently the test works because `Variable.__add__` doesn't check `None`.  Tests shouldn't depend on this behavior.


beam2d:
LGTM

beam2d:
Thanks!

Issue #4610, opened 2018-04-11
Avoid hacking for coding style check


niboshi:
Now that contributors can install style-check tools with `pip install -e .[stylecheck]`, we no longer have to use `hacking` in order to fix the tool versions. We can have finer control over tool versions than `hacking`.

One problem of `hacking` is that it fixes to rather old versions, which means we're unable to benefit from recent bugfixes.

Another problem is that its dependency can conflict with other tools (e.g. the case in #4606).

jinjiren:
Yesterday I actually met some bugs when using the old version of flake8 used in hacking.

niboshi:
(See conversation in #4674) hacking has its own style checks. It's not that simple as replacing hacking with explicit requirements.

kmaehashi:
`hacking` has been updated to use `flake8` 2.6.x.

https://github.com/openstack-dev/hacking/compare/1.0.0...1.1.0#files_bucket (requirements.txt)

Related: #4731

PR #4674, opened 2018-04-24
Replace hacking with explicit requirements

niboshi:
Fixes #4610.

PR #4864, opened 2018-06-07
Update style check tools

niboshi:
Fix: #4610
Related: #4674

### DIFF
--- a/docs/source/contribution.rst
+++ b/docs/source/contribution.rst
@@ -212,9 +212,15 @@ Coding Guidelines
 
 We use `PEP 8 <https://www.python.org/dev/peps/pep-0008/>`_ and a part of `OpenStack Style Guidelines <https://docs.openstack.org/developer/hacking/>`_ related to general coding style as our basic style guidelines.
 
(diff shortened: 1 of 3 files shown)


niboshi:
@kmaehashi @beam2d Can you merge this?

toslunar:
The Jenkins failure (`links_tests.normalization_tests.test_batch_renormalization.BatchRenormalizationTest_param_9.test_forward_multi_gpu`) is irrelevant to the PR.

kmaehashi:
## Context behind this PR

* We are planning to stop using ``hacking`` in v5 final release. The motivation is to make version of each tool under our control, so that we can use the latest tools (ie, which includes the latest bug fixes) as early as possible.
* As a first step, we substituted ``hacking`` package with explicit tool requirements with version specified. **Please note that this is currently under the trial phase; we may revise this later.**
* This PR does not intend to change our coding style.
* The `ignore` rules are adjusted so that code that passes check with toolset installed via ``hacking==1.0.0`` should be able to pass check with toolset installed via new dependencies proposed in this PR. However, the reverse may not be true; it will be confirmed during the trial period.

kmaehashi:
LGTM!

PR #4868, opened 2018-06-08
Stop using freeze_running_statistics in TestFixedBatchRenormalization

toslunar:
Avoid directly using `FunctionNode`.


okuta:
LGTM!

PR #4872, opened 2018-06-08
Mark multi_gpu tests as gpu tests

toslunar:


chainer/testing/attr.py
```diff
@@ -66,4 +74,4 @@ def gpu(f):
     """
 
     check_available()
-    return multi_gpu(1)(pytest.mark.gpu(f))
+    return _multi_gpu(1)(pytest.mark.gpu(f))
```
okuta: Why do not use `return multi_gpu(1)(f)` ?

okuta:
LGTM!

PR #4874, opened 2018-06-08
Improve Caffe exporter test to compare results

kmaehashi:
Improve caffe exporter test to compare results with re-imported link.

Blocked by #4875.


kmaehashi:
see #4884

PR #4875, opened 2018-06-08
Support importing Caffe Reshape layer

kmaehashi:
Motivation behind this PR: #4884 (Some functions like `F.linear` exports `Reshape` layer)


okuta:
LGTM!

PR #4876, opened 2018-06-08
Catch warnings in `test_count_params`

toslunar:


toslunar:
#3101

tests/chainer_tests/test_link.py
```diff
@@ -507,10 +507,9 @@ def test_update_enabled(self):
     def test_count_params(self):
         with warnings.catch_warnings(record=True) as w:
             warnings.simplefilter('always')
-            self.link.count_params()
+            self.assertEqual(self.link.count_params(), 8)
```
okuta: How about using `assert` ?

tests/chainer_tests/test_link.py
```diff
@@ -999,12 +998,13 @@ def test_serialize(self):
         mocks['l2'].assert_called_with('x', self.l2.x.data)
 
     def test_count_params(self):
+        self.assertEqual(self.c1.count_params(), 8)
```
okuta: ditto

PR #4879, opened 2018-06-09
Fix error message when checking types of grads

oggreo:
The error message for _check_grad_type prints out type (dtype, shape) of x and gx in the wrong order.


okuta:
Thanks!
jenkins, test this please.

PR #4880, opened 2018-06-10
Improve argument processing performance

okuta:
I want to remove overhead of argument processing in `chainer.functions`.


niboshi:
LGTM!

PR #4887, opened 2018-06-11
Simplify array type information fields in function documentation

### LINKED ISSUE #3929: Document: type specification of arrays
In function documentation, input arrays are described as **"`Variable` or `numpy.ndarray` or `cupy.ndarray`"** ([example](https://docs.chainer.org/en/stable/reference/generated/chainer.functions.squeeze.html)).

Currently we only support those two ndarray types, but we will need to generalize this to support more backend arrays (e.g. iDeep). It would be better to specify types in a general term and describe in detail in another documentation (just like initializer (example: [F.linear](https://docs.chainer.org/en/stable/reference/generated/chainer.links.Linear.html) links to [initializer documentation](https://docs.chainer.org/en/stable/reference/initializers.html))

mitmul:
This PR resolves #3929.

### DIFF
--- a/chainer/functions/activation/clipped_relu.py
+++ b/chainer/functions/activation/clipped_relu.py
@@ -129,8 +129,7 @@ def clipped_relu(x, z=20.0):
     .. math:: \\text{ClippedReLU}(x, z) = \\min(\\max(0, x), z).
 
     Args:
-        x (:class:`~chainer.Variable` or :class:`numpy.ndarray` or \
-        :class:`cupy.ndarray`):
+        x (:class:`~chainer.Variable` or `ndarray`):
             Input variable. A :math:`(s_1, s_2, ..., s_n)`-shaped float array.
         z (float): Clipping value. (default = 20.0)
 


niboshi:
I think it's better to make it a link to other documentation like ["wrapper functions"](https://docs.chainer.org/en/latest/reference/functions.html#module-chainer.functions), just like initializers.

mitmul:
Which do you think is preferred? to NumPy's doc or CuPy's doc.

niboshi:
A link to ["wrapper functions"](https://docs.chainer.org/en/latest/reference/functions.html#module-chainer.functions) is good. It has links to both ndarrays. (Although it seems it lacks module part...)

mitmul:
Ah, I see. I misread your comment that means making it a link to other doc as like "wrapper functions" does. OK, I'll make all into a link to that page! Thanks.

Crissman:
I believe @kmaehashi was doing re-work on the "wrapper functions" page to make it clearer.

niboshi:
Ref. #4855

@Crissman 
This PR can be done independent of it.

Crissman:
@mitmul Is this still WIP?

mitmul:
@niboshi Now the "wrapper functions" page does not have a link to both numpy and cupy ndarrays, what do you think which page is suitable for linking from "ndarray" in the docstring?

niboshi:
Hmm, I think we should describe the input specification anywhere in that page...

mitmul:
@niboshi How about linking to https://docs.chainer.org/en/latest/reference/generated/chainer.Variable.html#chainer.Variable which has the description about both numpy.ndarray and cupy.ndarray.

toslunar:
@mitmul I'd like https://docs.chainer.org/en/latest/reference/variable.html to have a "ndarray" documentation on the compatibility between `Variable` and `xp.ndarray` in Chainer's functions.

PR #4894, opened 2018-06-12
Add `StepShift` Extension.

jinjiren:
This PR adds StepShift Extension to `Trainer`.

For example, given ``k``, a multiplier ``gamma`` and an initial value ``init``, the optimizer attribute is set to ``init * gamma ^ (floor(iter / k))``.

Docs and simple unittests are added.

### DIFF
--- a/chainer/training/extensions/step_shift.py
+++ b/chainer/training/extensions/step_shift.py
@@ -10,7 +10,7 @@ class StepShift(extension.Extension):
     """Trainer extension to shift an optimizer attribute in "steps".
 
     This extension multiplies the specified attribute of the optimizer in
-    "steps". The typical use case is to scale the attribute at every ``k``th
+    "steps". The typical use case is to scale the attribute at every ``k``\ th
     iteration.
 
     For example, suppose that this extension is invoked at every iteration,
--- a/chainer/training/extensions/step_shift.py
+++ b/chainer/training/extensions/step_shift.py
@@ -23,25 +23,34 @@ class StepShift(extension.Extension):
 
     Args:
         attr (str): Name of the optimizer attribute to adjust.
-        init (float): The initial value of the attribute.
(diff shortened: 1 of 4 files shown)


niboshi:
Thank you for PR!
Implementation looks good.
Please fix flake8 failure.

jinjiren:
Fixed flake8 error and modified the doc string to be compatible with existing extensions.

niboshi:
Thanks!
Could you align the interface of `init` and `target` arguments with existing extensions as well?

kmaehashi:
@jinjiren ping?

jinjiren:
@kmaehashi Sorry for late reply. I have made the interfaces of init and target arguments to be compatible with existing extensions.

jinjiren:
I'm sorry that I cannot understand the error... I have run tests and flake8 (the version with `hacking`) locally and did not find this kind of error.

chainer/training/extensions/step_shift.py
```diff
@@ -0,0 +1,87 @@
...
+
+from chainer.training import extension
+
+
+class StepShift(extension.Extension):
+
+    """Trainer extension to shift an optimizer attribute in "steps".
+
+    This extension multiplies the specified attribute of the optimizer in
+    "steps". The typical use case is to scale the attribute at every ``k``th
```
niboshi: Thank you for the fix.
It seems ``` ``k``th ``` should be replaced with ``` ``k``\ th ``` here.

niboshi: By the way, this is an error in documentation.
Please check out [here](https://docs.chainer.org/en/stable/contribution.html#documentation) for details.

kmaehashi:
Thanks for the update! Jenkins, test this please.

niboshi:
LGTM. Thank you very much!

Issue #4056, opened 2017-12-07
Describe relation between alpha argument and learning rate in Adam optimizer


niboshi:
Chainer [documents](https://docs.chainer.org/en/v4.0.0b1/reference/generated/chainer.optimizers.Adam.html) this argument as "step size", but other major frameworks describe it as "learning rate". ([pytorch](http://pytorch.org/docs/master/optim.html#torch.optim.Adam), [tensorflow](https://www.tensorflow.org/api_docs/python/tf/train/AdamOptimizer), [keras](https://keras.io/optimizers/#adam), [mxnet](https://mxnet.incubator.apache.org/api/python/optimization.html#mxnet.optimizer.Adam)).

The original paper describes it as "step size", so maybe we should leave the current description, but it may be better to mention "learning rate" as well.

beam2d:
One complication is that `chainer.optimizers.Adam` also has `lr` property, which means "learning rate" and returns a *corrected version* of `alpha` used in the algorithm. I agree that we should mention how the `alpha` argument is called in other frameworks, but we need a careful explanation about the correspondence (I mean, we cannot just say "alpha is equivalent to learning rate").

niboshi:
I see. I changed the title accordingly.

niboshi:
bump

toslunar:
I'd like to hide `Adam.lr` property, because it's not a meaningful value to me.  "Correcting" `alpha` is just computationally cheaper than correcting momentums.

Crissman:
Merged.

PR #4895, opened 2018-06-12
Fix Adam alpha argument explanation

Crissman:
PR to address issue #4056


Crissman:
`Learning rate` seems preferred, according to other frameworks and your informal poll. I went back to the original AdamW paper, and did a search for "Step size" and couldn't find the term used. Seems like Learning Rate is best. If you would like to add a detailed explanation, please provide a text draft.

AdamW is not related only to `alpha`. It depends on `eta` and `weight_decay_rate`, if these are changed from the default value.

`AdamRule` docstring fixed. Please approve!

toslunar:
Jenkins, test this please.

toslunar:
Jenkins, test this please.

PR #4896, opened 2018-06-12
Remove experimental warning

niboshi:
We have changed the policy so that we no longer use the warning to
indicate that a feature is experimental, and instead we should write
the status in the documentation.

### DIFF
--- a/chainer/function_node.py
+++ b/chainer/function_node.py
@@ -11,7 +11,6 @@ from chainer import _backprop_utils
 from chainer.backends import cuda
 from chainer import configuration
 from chainer import function_hook
-from chainer.utils import experimental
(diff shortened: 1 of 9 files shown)


chainer/function_node.py
```diff
@@ -308,8 +307,6 @@ def apply(self, inputs):
                 self._retained_output_data = tuple(retained_data)
 
             self.lazy_grad_sum = configuration.config.lazy_grad_sum
-            if self.lazy_grad_sum:
-                experimental('config.lazy_grad_sum')
```
niboshi: 'config.lazy_grad_sum' is not documented as experimental, but maybe it's ok to just remove experimental status?

PR #4900, opened 2018-06-13
Improve the documentation of `F.moveaxis` and discourage `F.rollaxis`

toslunar:
`rollaxis` is not a good function when it is used to move an axis forward.


bkvogel:
Thanks for improving the documentation. LGTM!

PR #4907, opened 2018-06-13
Implement mask and return_mask option to `F.dropout`

kmaehashi:
Refs https://github.com/chainer/chainer/pull/4421#issuecomment-395377917

### DIFF
--- a/chainer/functions/noise/dropout.py
+++ b/chainer/functions/noise/dropout.py
@@ -108,11 +108,13 @@ def dropout(x, ratio=.5, **kwargs):
             Dropout ratio. The ``ratio`` must be ``0.0 <= ratio < 1.0``.
         mask (`ndarray` or None):
             The mask to be used for dropout.
-            The shape and dtype must be the same as ``x`` and should be on the
-            same device.
+            You do not have to specify this value, unless you need to make
+            results deterministic.
             If ``mask`` is not specified or set to ``None``, a mask will be
             generated randomly according to the given ``ratio``.
             If ``mask`` is specified, ``ratio`` will be ignored.
+            The shape and dtype must be the same as ``x`` and should be on the
+            same device.
             Note that iDeep will not be used for this function if mask is
             specified, as iDeep does not support it.
         return_mask (bool):
--- a/chainer/functions/noise/dropout.py
+++ b/chainer/functions/noise/dropout.py
@@ -13,12 +13,11 @@ class Dropout(function_node.FunctionNode):
 
     """Dropout regularization."""
 
-    mask = None
-
-    def __init__(self, dropout_ratio):
+    def __init__(self, dropout_ratio, mask=None):
         if not 0.0 <= dropout_ratio < 1.0:
             raise ValueError('dropout_ratio must be in the range [0, 1)')
         self.dropout_ratio = dropout_ratio
+        self.mask = mask
 
     def check_type_forward(self, in_types):
         type_check.expect(in_types.size() == 1)
@@ -26,7 +25,8 @@ class Dropout(function_node.FunctionNode):
 
     def forward(self, x):
         if (intel64.should_use_ideep('>=auto')
-                and intel64.inputs_all_ready(x)):
+                and intel64.inputs_all_ready(x)
+                and self.mask is None):
             return self._forward_ideep(x)
 
         if self.mask is not None:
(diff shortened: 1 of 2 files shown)


chainer/functions/noise/dropout.py
```diff
@@ -105,9 +106,28 @@ def dropout(x, ratio=.5, **kwargs):
             Input variable. A :math:`(s_1, s_2, ..., s_N)` -shaped float array.
         ratio (float):
             Dropout ratio. The ``ratio`` must be ``0.0 <= ratio < 1.0``.
+        mask (ndarray or None):
```
hvy: This "ndarray" is not a link, but did we agree that that was okay?

chainer/functions/noise/dropout.py
```diff
@@ -105,9 +106,28 @@ def dropout(x, ratio=.5, **kwargs):
...
+            The mask to be used for dropout.
+            The shape and dtype must be the same as ``x`` and should be on the
+            same device.
+            If ``mask`` is not specified or set to ``None``, a mask will be
+            generated randomly according to the given ``ratio``.
+            If ``mask`` is specified, ``ratio`` will be ignored.
+            Note that iDeep will not be used for this function if mask is
+            specified, as iDeep does not support it.
+        return_mask (bool):
+            If ``True``, the mask used for dropout is returned altogether with
```
hvy: A nitpick but could you change "altogether" to "together"?

tests/chainer_tests/functions_tests/noise_tests/test_dropout.py
```diff
@@ -129,4 +132,46 @@ def test_immutable(self, backend_config):
...
+        with chainer.using_config('train', self.train):
+            out, out_mask = functions.dropout(
+                self.x, 0.5, mask=mask, return_mask=True)
+
+        if self.train:
+            assert isinstance(out_mask, type(out.array))
+            if mask is None:
+                assert out_mask.shape == out.array.shape
+            else:
+                assert out_mask is mask
```
hvy: How about in both cases (mask is None and mask is not None), make sure we get the correct output by applying the returned mask to the input?

hvy:
Although these PRs won't break backward compatibility, they change the interface. Just wondering if they need to be backported.

kmaehashi:
I think it's better to backport these non-breaking changes, because it will make migration to v5  smooth for users.
Maybe it's better to add notes like this to class document?:

Direct access of ``mask`` attribute will be deprecated in Chainer v5.  Use  `return_mask`/`mask`  option of `F.dropout` to get/set mask used for dropout.

hvy:
I see you point regarding the backport. Let's do so. 

As for the the documentation you suggest, it could help users but if this wasn't a public API before I don't think we need to document it that way.

chainer/functions/noise/dropout.py
kmaehashi: The discussion is ongoing in #4887. I added quote, so that we can later find it by ``grep -F 'ndarray`'``

chainer/functions/noise/dropout.py
kmaehashi: Fixed.

tests/chainer_tests/functions_tests/noise_tests/test_dropout.py
kmaehashi: Test added.

kmaehashi:
> As for the the documentation you suggest, it could help users but if this wasn't a public API before I don't think we need to document it that way.

Ok, I agree with it.

hvy:
jenkins, test this please.

hvy:
jenkins, test this please.

chainer/functions/noise/dropout.py
```diff
@@ -105,9 +106,30 @@ def dropout(x, ratio=.5, **kwargs):
...
+            If ``True``, the mask used for dropout is returned together with
+            the output variable.
+            The returned mask can later be reused by passing it to ``mask``
+            argument.
 
     Returns:
-        ~chainer.Variable: Output variable.
+        ~chainer.Variable or tuple:
+            When ``return_mask`` is ``False`` (default), returns the output
+            variable. Otherwise returnes the tuple of the output variable and
```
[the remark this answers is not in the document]
kmaehashi: Fixed.

hvy:
jenkins, test this please.

kmaehashi:
Jenkins, test this please.

hvy:
LGTM!

PR #4910, opened 2018-06-13
Introduce CHAINER_SETUP_IGNORE_CUPY to ignore cupy in setup.py

niboshi:
Chainer depends on a very specific version of CuPy.
Installation of CuPy takes too much time.
I propose introduction of environment variable to ignore CuPy.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -101,20 +101,21 @@ def find_any_distribution(pkgs):
     return None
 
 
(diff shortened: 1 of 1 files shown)


niboshi:
It turned out `--no-deps` option can be used for this purpose:

```
pip install -e . --no-deps
```

Issue #4656, opened 2018-04-19
Careless use of `no_grads` in tests


toslunar:
The argument `no_grads` appears in the tests of
- `F.crf1d`
- `F.minimum`
- `F.sign`

This skips computing numerical gradients w.r.t. the no_grads inputs, while it seems authors just wanted "trivial" tests to pass.  I suggest `gx == None` should be checked as if `gx == xp.zeros_like(x)`, since such authors' math is not so "trivial" to everyone.

An exception is the test of `F.roi_pooling_2d` where `no_grads` is set to `rois` probably because indices are stored at `rois[:, 0]`:
```
            idx, xmin, ymin, xmax, ymax = bottom_rois[i_roi]
            xmin = int(round(xmin * self.spatial_scale))
            xmax = int(round(xmax * self.spatial_scale))
            ymin = int(round(ymin * self.spatial_scale))
            ymax = int(round(ymax * self.spatial_scale))
```

BTW, the tests of `L.SimplifiedDropconnect` specify default `no_grads`. I found removing them after the commit 56ed844 is OK:
```
% git diff
diff --git a/tests/chainer_tests/links_tests/activation_tests/test_simplified_dropconnect.py b/tests/chainer_tests/links_tests/activation_tests/test_simplified_dropconnect.py
index a7cf6a18c..564b442df 100644
--- a/tests/chainer_tests/links_tests/activation_tests/test_simplified_dropconnect.py
+++ b/tests/chainer_tests/links_tests/activation_tests/test_simplified_dropconnect.py
@@ -86,7 +86,7 @@ class TestSimplifiedDropconnect(unittest.TestCase):
         gradient_check.check_backward(
             self.link_wrapper, (x_data, mask), y_grad,
             (self.link.W, self.link.b), eps=2 ** -3,
-            no_grads=(False, True), **self.check_backward_options)
+             **self.check_backward_options)

     @condition.retry(3)
     def test_backward_cpu(self):
@@ -153,7 +153,7 @@ class TestSimplifiedDropconnectParameterShapePlaceholder(unittest.TestCase):
     def check_backward(self, x_data, y_grad, mask):
         gradient_check.check_backward(
             self.link_wrapper, (x_data, mask), y_grad,
-            (self.link.W, self.link.b), eps=1e-2, no_grads=(False, True))
+            (self.link.W, self.link.b), eps=1e-2, )
% nosetests -a '!gpu' chainer_tests/links_tests/activation_tests/test_simplified_dropconnect.py
[...]

PR #4914, opened 2018-06-13
Delete redundant `no_grads` from test

toslunar:
See https://github.com/chainer/chainer/issues/4656#issue-315705686


okuta:
LGTM

PR #4915, opened 2018-06-13
Fix `SigmoidCrossEntropy.backward`

### LINKED ISSUE #4877: Variable.backward causes an error when composing F.tile, F.sigmoid_cross_entropy, and F.sum.
An error occurs when we compose `F.tile`, `F.sigmoid_cross_entropy` (with `reduce=no`), and `F.sum` and do `Variable.backward` to the output variable.

* Target version: 3.2.0<= v <=master

Code to reproduce
```python
import chainer
import numpy as np

chainer.print_runtime_info()

y = chainer.Variable(np.random.standard_normal((12, 10)).astype(np.float32))
t = chainer.Variable(np.random.standard_normal((12, 1)).astype(np.int32))
size = 10

t = F.tile(t, (1, size))
loss = F.sigmoid_cross_entropy(y, t, reduce="no")
loss = F.sum(loss)
loss.backward()
```

Log
```
Chainer: 5.0.0b1
NumPy: 1.11.3
CuPy: Not Available
  File "test.py", line 14, in <module>
    loss.backward()
  File "/home/delta/dev/chainer/chainer/variable.py", line 966, in backward
    self._backward_main(retain_grad, loss_scale)
  File "/home/delta/dev/chainer/chainer/variable.py", line 1095, in _backward_main
    target_input_indexes, out_grad, in_grad)
  File "/home/delta/dev/chainer/chainer/function_node.py", line 548, in backward_accumulate
    gxs = self.backward(target_input_indexes, grad_outputs)
  File "/home/delta/dev/chainer/chainer/functions/array/tile.py", line 50, in backward
    for i in range(gy.ndim):
  File "/home/delta/dev/chainer/chainer/variable.py", line 703, in ndim
    return self.data.ndim
AttributeError: 'NoneType' object has no attribute 'ndim'
```

The error did not occur if we change in either way:
- Remove `F.tile`.
- Remove `F.sum` and change `reduce` option of `F.sigmoid_cross_entropy` to `mean`.

I did `git bisect` and found that the first bad commit was #3705. So, I suspect `F.sigmoid_cross_entropy`.

toslunar:
This fixes #4877, probably.
- [x] Fix backward
- [x] Fix tests

### DIFF
--- a/chainer/functions/loss/sigmoid_cross_entropy.py
+++ b/chainer/functions/loss/sigmoid_cross_entropy.py
@@ -60,8 +60,9 @@ class SigmoidCrossEntropy(function_node.FunctionNode):
     def backward(self, inputs, grad_outputs):
         x, t = self.get_retained_inputs()
         gy, = grad_outputs
-        return SigmoidCrossEntropyGrad(
+        gx, = SigmoidCrossEntropyGrad(
             self.reduce, self.count, self.ignore_mask, t.data).apply((x, gy))
+        return gx, None
 
 
 class SigmoidCrossEntropyGrad(function_node.FunctionNode):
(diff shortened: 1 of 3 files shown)


okuta:
LGTM!

PR #4922, opened 2018-06-14
Add WaveNet example

dhgrs:
Add an example of mel-spectrogram vocoder using WaveNet based on https://github.com/dhgrs/chainer-Fast-WaveNet

I tested training with docker image chainer/chainer:v4.1.0-python3. After 600k iterations with default parameters, we can get audible results like [here](https://github.com/dhgrs/chainer/blob/wavenet_result/examples/wavenet/result.wav). This result was generated from a mel-spectrogram of a speaker in CMU Arctic Dataset and never be used in training.

### DIFF
--- a/examples/wavenet/modules.py
+++ b/examples/wavenet/modules.py
@@ -1,4 +1,3 @@
-import numpy
 import chainer
 import chainer.functions as F
 import chainer.links as L
--- a/examples/wavenet/modules.py
+++ b/examples/wavenet/modules.py
@@ -1,0 +1,84 @@
+import numpy
+import chainer
+import chainer.functions as F
+import chainer.links as L
+
+
+class ResidualBlock(chainer.Chain):
+    def __init__(self, filter_size, dilation,
+                 residual_channels, dilated_channels, skip_channels):
+        super(ResidualBlock, self).__init__()
+        with self.init_scope():
+            self.conv = L.DilatedConvolution2D(
+                residual_channels, dilated_channels,
+                ksize=(filter_size, 1),
+                pad=(dilation * (filter_size - 1), 0), dilate=(dilation, 1))
+            self.res = L.Convolution2D(
+                dilated_channels // 2, residual_channels, 1)
+            self.skip = L.Convolution2D(
+                dilated_channels // 2, skip_channels, 1)
+
+        self.filter_size = filter_size
+        self.dilation = dilation
+        self.residual_channels = residual_channels
+
+    def __call__(self, x, condition):
+        length = x.shape[2]
+        h = self.conv(x)
+        h = h[:, :, :length]  # crop
+        h += condition
+        tanh_z, sig_z = F.split_axis(h, 2, axis=1)
+        z = F.tanh(tanh_z) * F.sigmoid(sig_z)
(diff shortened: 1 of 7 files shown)


examples/wavenet/modules.py
```diff
@@ -0,0 +1,83 @@
...
+            self.skip = L.Convolution2D(
+                dilated_channels // 2, skip_channels, 1)
+
+        self.filter_size = filter_size
+        self.dilation = dilation
+        self.residual_channels = residual_channels
+
+    def __call__(self, x, condition):
+        length = x.shape[2]
+        h = self.conv(x)
```
y-takashina: I just noticed that a single conv layer is applied before different activation functions in this implementation, while [the original paper](https://arxiv.org/pdf/1609.03499.pdf) looks like it uses different conv layers (W_{f,k} and W_{g,k} in Eq.(2)).

dhgrs: Two convolutions can be represented by single convolution and splitting in channel axis after convolution.

https://github.com/dhgrs/chainer/blob/48f7786371bf0b0c40a4504426ef4e18bc34e888/examples/wavenet/modules.py#L29

In figure 4 in the original paper looks like single convolution and splitting are used. And nv-wavenet, CUDA kernel for WaveNet inference also uses single convolution and splitting instead of two convolutions.

y-takashina: I see. Thank you for your response!

examples/wavenet/train.py
```diff
@@ -0,0 +1,125 @@
...
+
+if args.gpu >= 0:
+    chainer.global_config.autotune = True
+
+# Datasets
+paths = sorted([
+    str(path) for path in pathlib.Path(args.dataset).glob('wav48/*/*.wav')])
+preprocess = Preprocess(
+    sr=16000, n_fft=1024, hop_length=256, n_mels=128, top_db=20,
+    length=7680, quantize=args.a_channels)
```
keisuke-umezawa: `7680` should be `args.length`?

dhgrs: You are right! I fixed it. Thanks.

examples/wavenet/utils.py
```diff
@@ -0,0 +1,91 @@
...
+        if self.length is not None:
+            spectrogram = spectrogram[:, :self.length // self.hop_length]
+        spectrogram = spectrogram.astype(numpy.float32)
+
+        # expand dimensions
+        one_hot = numpy.identity(
+            self.quantize, dtype=numpy.float32)[quantized]
+        one_hot = numpy.expand_dims(one_hot.T, 2)
+        raw = numpy.expand_dims(raw, 0)  # expand channel
+        raw = numpy.expand_dims(raw, -1)  # expand height
```
keisuke-umezawa: These two lines changes ``raw``, but it is not used later. The lines can be deleted?

dhgrs: Sorry for late reply. I'll fix it. The transforms are for MoL WaveNet but no need for this PR.

dhgrs:
I think the test's failure does not has nothing to do with my PR. Please check it.

toslunar:
Jenkins, test this please.

PR #4923, opened 2018-06-14
Refactor distribution

toslunar:
- Make user pass variables to `Distribution` if grad is needed.
- Check `test_targets` members are valid.
- Don't override `setUp`.


okuta:
Thank you checking.

PR #4925, opened 2018-06-14
[RFC] Add `params` and `xp` to `Distribution`

toslunar:
I'd like to have `Distribution.params` to get all the parameters (in the statistics' sense) in a tuple or a dict.  This name can be confusing because `Link` has `params` method with different meaning.
In ChainerRL, `params` method returns a tuple of `Variable`s.

`params` can be used to give the default implementation of `Distribution.xp`.

### DIFF
--- a/chainer/distribution.py
+++ b/chainer/distribution.py
@@ -1,5 +1,7 @@
 import copy
 
+from chainer.backends import cuda
+
 
 class Distribution(object):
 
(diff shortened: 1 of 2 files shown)


okuta:
@YoshikawaMasashi What do you think about this PR?

YoshikawaMasashi:
I think it is good enhancement (especially `Distribution.xp`).
Can I review?

toslunar:
A set of parameters may depend on how `Distribution` was created, e.g. {`loc`, `scale`} or {`loc`, `log_scale`} for `D.Normal`.
Just holding variables in (kw)args of `__init__` and returning them seems OK.  Now I think `dict` is much better than `tuple`.

okuta:
@YoshikawaMasashi OK, Please review this.

PR #4934, opened 2018-06-15
Suggest users to use iDeep4py v1 for now

kmaehashi:
Refs #4933.

iDeep4py v2 seems to have different API than v1, and it also has a separate package for AVX2 environment and AVX512 environment.


mingxiaoh:
Once https://github.com/chainer/chainer/pull/4933 is merged to chainer, iDeep1.* will not be supported anymore.

okuta:
Once, we will merge this.
Cloud you fix this tips in your PR? @mingxiaoh

PR #4936, opened 2018-06-16
Add support for importing and exporting Caffe Deconvolution layer

tsurumeso:
I fixed Caffe Deconvolution layer to export properly `num_output` param. Also, I added support for importing the layer. I tested these features with [waifu2x-chainer](https://github.com/tsurumeso/waifu2x-chainer) that contains Deconvolution layer and I confirmed that it works correctly.


okuta:
jenkins, test this please

okuta:
LGTM!

PR #4949, opened 2018-06-18
Raise an error on 0-dim input to matmul

### LINKED ISSUE #4684: F.matmul allows scalar operand
PEP465 does not allow multiplying constant using the matmul operator `@`.
```
>>> import chainer; import numpy as np
>>> chainer.functions.matmul(np.zeros((), 'f'), np.zeros((), 'f'))
variable(0.)
>>> np.matmul(np.zeros((), 'f'), np.zeros((), 'f'))
ValueError: Scalar operands are not allowed, use '*' instead
>>> np.zeros((), 'f') @ np.zeros((), 'f')
ValueError: Scalar operands are not allowed, use '*' instead
Chainer: 4.0.0
NumPy: 1.14.2
CuPy: Not Available
>>>
```

toslunar:
Close #4684.

### DIFF
--- a/chainer/functions/math/matmul.py
+++ b/chainer/functions/math/matmul.py
@@ -83,13 +83,13 @@ class MatMul(function_node.FunctionNode):
         type_check.expect(
             a_type.dtype.kind == 'f',
             b_type.dtype.kind == 'f',
+            a_type.ndim >= 1,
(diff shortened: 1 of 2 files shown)


okuta:
LGTM!

PR #4960, opened 2018-06-19
[backport] Fix occasional test failure of `F.average` by changing lower-bound of weight.sum

kmaehashi:
Backport of #4771


asi1024:
LGTM.

Issue #2100, opened 2017-01-14
No document on usage of extension.name attribute by Trainer.extend


muupan:
`Trainer.extend` checks `extension.name` attribute before `extension.default_name` when `name` argument is `None`, but its docstring only mentions `extension.default_name`.

https://github.com/pfnet/chainer/blob/f34c20e45dc86dfa6bbc62e080be3fd97aa3f466/chainer/training/trainer.py#L202
https://github.com/pfnet/chainer/blob/f34c20e45dc86dfa6bbc62e080be3fd97aa3f466/chainer/training/trainer.py#L181

muupan:
I think this issue should be fixed because it means the current documentation is wrong.

muupan:
This is not yet resolved
https://github.com/chainer/chainer/blob/master/chainer/training/trainer.py#L194-L197
https://github.com/chainer/chainer/blob/master/chainer/training/trainer.py#L219

Crissman:
Fixed with #4980

PR #4980, opened 2018-06-21
Add extension.name clarification

Crissman:
Addressing issue #2100

### DIFF
--- a/chainer/training/extension.py
+++ b/chainer/training/extension.py
@@ -32,7 +32,9 @@ class Extension(object):
         ~Extension.priority: Default priority of the extension. It is set to
             ``PRIORITY_READER`` by default.
         ~Extension.name: Name of the extension. It is set to
-            ``None`` by default.
+            ``None`` by default. This value will be overwritten when
+            registering an extension to a trainer. See
+            :meth:`chainer.training.Trainer.extend` for details.
 
     """
     trigger = 1, 'iteration'
--- a/chainer/training/extension.py
+++ b/chainer/training/extension.py
@@ -31,10 +31,13 @@ class Extension(object):
             is set to ``(1, 'iteration')`` by default.
         ~Extension.priority: Default priority of the extension. It is set to
             ``PRIORITY_READER`` by default.
+        ~Extension.name: Name of the extension. It is set to
+            ``None`` by default.
 
     """
     trigger = 1, 'iteration'
     priority = PRIORITY_READER
+    name = None
 
     @property
     def default_name(self):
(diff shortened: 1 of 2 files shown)


kmaehashi:
@beam2d What was the motivation to introduce the concept of `default_name`? (in https://github.com/chainer/chainer/pull/1285/commits/09cc404d1c80a0daac778fbe5a7b3ecea93eb380)

niboshi:
Both `name` and `default_name` attributes are undocumented. How about documenting them and linking to them?

beam2d:
@kmaehashi It is the default value provided as a *property*. I wanted to override the behavior with an attribute, so I used `default_name` instead of `name` (note that we cannot override a property with an attribute with the same name due to the resolution order of `getattr`). It could be resolved in a different manner without using the name `default_name` (e.g. also writing a property setter to emulate attribute-like behavior, using non-data descriptor), but I did not choose them because I felt them a bit trickier.

Crissman:
@niboshi `default_name` is mentioned here. Is it enough to link to this page? https://docs.chainer.org/en/stable/reference/generated/chainer.training.Extension.html#chainer.training.Extension

kmaehashi:
@beam2d Thanks for clarification. I understood that this feature is basically for users who want to override the default name quickly (e.g., by calling `ext.name = 'MyExtension'` before calling `Trainer.extend`).

@Crissman I think @niboshi -san's comment includes the following:

* Add `name = None` to `Extension` class
* Add description of `name` under `Attributes:` of `Extension` class
* Fix docs to link to both attributes using `` :attr:`Extension.name` ``,  `` :attr:`Extension.default_name` ``

chainer/training/extension.py
```diff
@@ -31,10 +31,13 @@ class Extension(object):
             is set to ``(1, 'iteration')`` by default.
         ~Extension.priority: Default priority of the extension. It is set to
             ``PRIORITY_READER`` by default.
+        ~Extension.name: Name of the extension. It is set to
+            ``None`` by default.
```
kmaehashi: How about adding:

```
This value will be overwritten when registering an extension to a trainer. See :meth:`chainer.training.Trainer.extend` for details.
```

chainer/training/trainer.py
```diff
@@ -192,7 +192,9 @@ def extend(self, extension, name=None, trigger=None, priority=None,
         Args:
             extension: Extension to register.
             name (str): Name of the extension. If it is omitted, the
-                ``default_name`` attribute of the extension is used instead.
+                :attr:`Extension.name` attribute of the extension is used or
+                the :attr:`Extension.default_name` attribute of the extension
+                if ''name'' is also undefined.
```
kmaehashi: * `''name''` -> ``` `name` ```
* `also undefined` -> `is set to None or is undefined`

chainer/training/extension.py
Crissman: Done.

chainer/training/trainer.py
Crissman: Done.

kmaehashi:
Jenkins, test this please.

kmaehashi:

PR #4981, opened 2018-06-21
Fix docs of `chainer.config.dtype`

kmaehashi:


kmaehashi:
Sorry, that was my mistake. Removed and force-pushed.

okuta:
LGTM!

Issue #4115, opened 2017-12-17
Documentation of chainer.links.Linear has room for improvement


muupan:
https://docs.chainer.org/en/latest/reference/generated/chainer.links.Linear.html

- The `bias` parameter doesn't exist
- Explanation on how `W` and `b` are initialized ignores `initialW` and `initial_bias`
- There's no explanation on how `out_size=None` works (except in the example section)
- There's no explanation on how `initialW=None` works

In addition to the above, I think shapes and dtypes of `W` and `b` should be documented, otherwise they are out of the scope of the compatibility policy. However, this is not specific to `chainer.links.Linear`.

Crissman:
> @Crissman Shapes are documented in some parameters. Please see hx etc. in n_step_gru: https://docs.chainer.org/en/stable/reference/generated/chainer.functions.n_step_gru.html

Next step is to go through the documentation of all of the functions, and add shape and maybe dtype information to them. Reviewing some of the other frameworks to see how they handle this.

kmaehashi:
Let me close this issue and continue discussion in #2182.

PR #4983, opened 2018-06-21
 Clarifications for L.Linear

Crissman:
Adressing Issue #4115 

These items are done:

- Explanation on how W and b are initialized ignores initialW and initial_bias
- There's no explanation on how out_size=None works (except in the example section)
- There's no explanation on how initialW=None works

@muupan Can you provide clarification on what you mean by this?
> The `bias` parameter doesn't exist

Also, can you give a link to the specific compatibility policy or example of the format you're looking for for this?
> shapes and dtypes of `W` and `b` should be documented

Thanks,

### DIFF
--- a/chainer/links/connection/linear.py
+++ b/chainer/links/connection/linear.py
@@ -15,21 +15,26 @@ class Linear(link.Link):
     and holds a weight matrix ``W`` and optionally a bias vector ``b`` as
     parameters.
 
-    The weight matrix ``W`` is initialized with i.i.d. Gaussian samples, each
-    of which has zero mean and deviation :math:`\\sqrt{1/\\text{in_size}}`. The
-    bias vector ``b`` is of size ``out_size``. Each element is initialized with
-    the ``bias`` value. If ``nobias`` argument is set to ``True``, then this
-    link does not hold a bias vector.
+    If ``initialW`` is left to the default value of ``None``, the weight matrix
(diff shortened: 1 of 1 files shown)


muupan:
I mean ``bias`` in https://github.com/chainer/chainer/blob/master/chainer/links/connection/linear.py#L21. It seems already removed in this PR.

> Also, can you give a link to the specific compatibility policy or example of the format you're looking for for this?

I'm not sure if such a compatible policy or example already exists in Chainer. As a user, I just mean the shapes of `W` and `b` should be documented in some way so that users don't need to read the code.

Crissman:
@kmaehashi Let's merge these fixes separately from the shape/dtype discussion, since that will affect all of the function documentation, and I'd like to have some discussion on a standard format.

Let me know if there's anything else that should be done to for these specific changes.

kmaehashi:
> @kmaehashi Let's merge these fixes separately from the shape/dtype discussion, since that will affect all of the function documentation, and I'd like to have some discussion on a standard format.

Yes, let's consider the unified format like as we done for functions (#2182).
(As for `L.Linear` and many other links, I'm thinking that it's enough to just link to the corresponding function, e.g. `F.linear`: https://docs.chainer.org/en/latest/reference/generated/chainer.functions.linear.html, as the shape is already documented in there).

Jenkins, test this please.

Issue #4449, opened 2018-03-08
Improve support of higher-order derivatives for built-in functions


kmaehashi:
Although most of built-in functions now support higher-order derivative, some functions still do not support it.

Currently following functions do not support higher-order derivatives:

* ~~`F.slstm`~~ (#4824) merged!
* `F.tree_lstm`
* `F.spatial_transformer_grid`
* `F.spatial_transformer_sampler`
* `F.depthwise_convolution_2d`
* `F.n_step_bigru`
* `F.n_step_bilstm`
* `F.n_step_birnn`
* `F.n_step_gru`
* `F.n_step_lstm`
* `F.n_step_rnn`
* `F.accuracy`
* `F.binary_accuracy`
* `F.classification_summary`
* `F.r2_score`
* ~~`F.contrastive` (#3830)~~ merged!
* `F.connectionist_temporal_classification`
* `F.decov`
* `F.hinge`
* ~~`F.prod`~~ (#3764) merged!
* `F.batch_renormalization`
* `F.local_response_normalization`
* `F.roi_pooling_2d`
* ~~`F.forget`~~ (#3792) merged!
* `TheanoFunction`

The following functions only support second order derivatives:

* `F.lstm`
* `F.batch_normalization`
* `F.fixed_batch_normalization`

kmaehashi:
Guides for people considering contributions to do this work:

* To make function support higher-order derivative (aka double backprop), you need to reimplement `chainer.Function` as `chianer.FunctionNode`.
  [New-Style Function Implementation Guideline](https://gist.github.com/beam2d/dd87d97b8d7581f95cd19b1de86c9f4d) describes the detailed steps to do this work.
* You can find example of PRs in https://github.com/chainer/chainer/issues/3147 and https://github.com/chainer/chainer/issues?q=new+style+-label%3Abackport.

PR #4986, opened 2018-06-21
Write guides to implement new-style functions

kmaehashi:
I started to write docs on how to implement user-defined new-style functions.

I'd like to discuss two points:

## 1. How do we reorganize this docs?

Currently I'm planning to describe this way:

> In the following sections, we first introduce steps to implenent an old-style function. After that we describe the difference of implementation between old-style functions and new-style functions.

Or, should we completely rewrite the guide to use new-style functions?

I think old-style functions are still convenient for many people, as the implementation cost is less than new-style, so I'd like to proceed with the current style.

## 2. Can we give some better name to "old-style" functions?

I think the name "old-style" make users feel "it is not recommended to use".
How about giving less-aggressive name, like "classic-style" or "conventional"?

### DIFF
--- a/docs/source/guides/functions.rst
+++ b/docs/source/guides/functions.rst
@@ -488,7 +488,7 @@ You will need to check the value of the boolean flag ``chainer.config.train`` an
 For example, consider the following simple dropout function::
 
   def dropout(x):
-      xp = backend.get_array_module(x.data)
+      xp = backend.get_array_module(x.array)
       mask = 2 * (xp.random.rand(*x.shape) > 0.5).astype(x.dtype)
       return x * mask
 
@@ -500,7 +500,7 @@ We can fix it as follows::
       if not chainer.config.train:
           return x
 
-      xp = backend.get_array_module(x.data)
+      xp = backend.get_array_module(x.array)
       mask = 2 * (xp.random.rand(*x.shape) > 0.5).astype(x.dtype)
       return x * mask
 
--- a/docs/source/guides/functions.rst
+++ b/docs/source/guides/functions.rst
@@ -1,5 +1,5 @@
 Define your own function
-------------------------
+========================
 
 .. currentmodule:: chainer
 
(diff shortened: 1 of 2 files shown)


kmaehashi:
Refs #4449.
This document also intends to cover topics from https://gist.github.com/beam2d/dd87d97b8d7581f95cd19b1de86c9f4d.

niboshi:
>I think old-style functions are still convenient for many people, as the implementation cost is less than new-style, so I'd like to proceed with the current style.

Do you mean when optimized implementation is required?
Do normal users usually need optimized implementation?

>Can we give some better name to "old-style" functions?

How about "functions with/without higher-order backpropagation"?

kmaehashi:
As for users, it's simple that functions can simply implemented in one class using NumPy/CuPy. (like done in other frameworks).
I'm not sure if higher-order backprop will widely be used in many other domains. In that case I think it's the right direction (in terms of designing the ecosystem) to always recommend users to use FunctionNode so that community assets will become more valuable for longer period.

As for the second point, I overlooked that the terms are already defined here (Notes box), so I'll proceed with the current terms.
https://docs.chainer.org/en/stable/reference/generated/chainer.FunctionNode.html#chainer.FunctionNode

kmaehashi:
We discussed offline:

* We recommend users to use FunctionNode in the first place.
* However, still leave documentation for Function (expected readers are users who needs to read old code)
* Also add guidance to help users implement functions without having to write backward (i.e., "non-double-backpropable" Function) to reduce implementation cost.
  (It may be better to have a built-in feature for this; this point is out of scope of this PR.)

toslunar:
I expect this PR closes #5707.

docs/source/guides/functions.rst
```diff
@@ -17,13 +17,38 @@ After reading this section, you will be able to:
...
+
+In Chainer, you can define a function in two ways: old-style and new-style.
+
+* Old-style functions inherit from :class:`chainer.Function` class.
+  Forward and backward computation can be implemented using NumPy/CuPy.
+* New-style functions inherit from :class:`chainer.FunctionNode` class (introduced in Chainer v3).
+  Forward computation can be implemented using NumPy/CuPy.
+  Backward computation needs to be implemented using other new-style functions.
+
+The primary advantage of using new-style functions is that they support computation of higher-order gradient (a.k.a. higher-order derivative or double back-propagation).
```
Crissman: gradients

docs/source/guides/functions.rst
```diff
@@ -17,13 +17,38 @@ After reading this section, you will be able to:
...
+Higher-order gradient is used in some models e.g., recently-proposed GAN architectures.
```
Crissman: gradients are used

docs/source/guides/functions.rst
```diff
@@ -17,13 +17,38 @@ After reading this section, you will be able to:
...
+
+Currently, most of built-in functions are implemented as :class:`~chainer.FunctionNode` (with a few exceptions listed in `#4449 <https://github.com/chainer/chainer/issues/4449>`__).
+
+Basically, we suggest you to use new-style when implementing new functions.
```
Crissman: We recommend you use

docs/source/guides/functions.rst
Crissman:
```suggestion
The primary advantage of using new-style functions is that they support computation of higher-order gradients (a.k.a. higher-order derivative or double back-propagation).
```

docs/source/guides/functions.rst
Crissman:
```suggestion
Higher-order gradients are used in some models e.g., recently-proposed GAN architectures.
```

docs/source/guides/functions.rst
Crissman:
```suggestion
Basically, we recommend you use new-style when implementing new functions.
```

kmaehashi:
imosCI, test this please.

kmaehashi:
I've finished working on this PR. Ready for another review.

docs/source/guides/functions.rst
```diff
@@ -420,143 +485,38 @@ You can refer to :ref:`configuration` to see how to configure this flag as well
...
-
-.. testcode::
+We can fix it as follows::
 
   def dropout(x):
       if not chainer.config.train:
           return x
 
-      xp = backend.get_array_module(x.array)
+      xp = backend.get_array_module(x.data)
```
toslunar:
```suggestion
      xp = backend.get_array_module(x.array)
```

docs/source/guides/functions.rst
```diff
@@ -420,143 +485,38 @@ You can refer to :ref:`configuration` to see how to configure this flag as well
...
 You will need to check the value of the boolean flag ``chainer.config.train`` and branch appropriately.
 
-For example, consider the following simple dropout function:
-
-.. testcode::
+For example, consider the following simple dropout function::
 
   def dropout(x):
-      xp = backend.get_array_module(x.array)
+      xp = backend.get_array_module(x.data)
```
toslunar:
```suggestion
      xp = backend.get_array_module(x.array)
```

kmaehashi:
Jenkins, test this please.

PR #4989, opened 2018-06-22
Add link from installation guide to upgrade guide

niboshi:
Add link from installation guide to upgrade guide.

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -125,6 +125,10 @@ Just use ``pip`` with ``-U`` option::
 
   $ pip install -U chainer
 
+.. seealso::
+
+   To fix the code which uses Chainer, please refer to the :doc:`upgrade`.
+
 
 Reinstall Chainer
 -----------------


docs/source/install.rst
```diff
@@ -125,6 +125,10 @@ Just use ``pip`` with ``-U`` option::
 
   $ pip install -U chainer
 
+.. seealso::
+
+   To fix the code which uses Chainer, please refer to the :doc:`upgrade`.
```
kmaehashi: As the Upgrade Guide covers topics other than fixing code (e.g., change to CuPy version requirement), how about saying like this?

> If you are migrating between major versions (e.g., from Chainer v3.x to v4.x), please check out the :doc:`upgrade` for important changes.

niboshi: Requirements should be resolved by `pip`, so I don't think normal users have to refer to the guide for that reason. Is there anything they have to care about?

kmaehashi: Yes. As CuPy is a separate package, it may be used by user, outside of Chainer.

niboshi: What do you mean? I think CuPy is upgraded by `pip` automatically if it's installed.

kmaehashi: I'd like to document changes of requirements to important component in the Upgrade Guide, so that users can track the changes.

kmaehashi: > What do you mean?

I meant, some users wants to be aware of important dependency changes even though it is automatically resolved.

> I think CuPy is upgraded by `pip` automatically if it's installed.

As CuPy package no longer updates automatically, I think it's worth mentioning.

niboshi: I don't want the documentation too noisy.
How about simply stating "Also refer to the upgrade guide."?
I just want to place a link to the guide. Anyone can follow the link to check what is written.

Issue #4902, opened 2018-06-13
ParallelUpdater does not share persistent values


iwiwi:
https://github.com/chainer/chainer/blob/06554143569ff0e78e7c1b37904dc20a2cde3d33/chainer/training/updaters/parallel_updater.py#L149

`copyparams` does not copy persistent values, so models in different GPUs do not share persistent values.

iwiwi:
Addressed by #4997

PR #4997, opened 2018-06-25
Copy persistent values in `Link.copyparams`

### LINKED ISSUE #4903: Should `copyparams` copy persistent values as well?
Currently, `{Link, Chain}.copyparams` does not care persistent values and it only copies parameters. Due to this behavior, for example, *every* model using batch normalization cannot be correctly copied using this method.

See #4902 for an example of errors due to this problem.

Do we have situations where we want to copy parameter values but we don't want to copy persistent values? I propose to change the behavior of `copyparams` to also copy persistent values.

beam2d:
Fixes #4903. This change makes `Link.copyparams` also copies the persistent values, notably the `avg_mean`, `avg_var`, and `N` of `links.BatchNormalization`.

Note that this change is backward incompatible since the default behavior is changed. The original behavior is confusing, and the new behavior should be beneficial in most cases (including existing code). The original behavior can be reproduced by passing `copy_persistent=False`.

### DIFF
--- a/chainer/link.py
+++ b/chainer/link.py
@@ -515,6 +515,11 @@ Assign a Parameter object directly to an attribute within a \
         copy is even done across the host and devices. Note that this method
         does not copy the gradient arrays.
 
+        *From v5.0.0:* this method also copies the persistent values (e.g. the
+        moving statistics of :class:`~chainer.links.BatchNormalization`). If
+        the persistent value is an ndarray, the elements are copied. Otherwise,
+        it is copied using :func:`copy.deepcopy`.
+
         Args:
             link (Link): Source link object.
 
@@ -523,6 +528,14 @@ Assign a Parameter object directly to an attribute within a \
         dst = self.__dict__
         for name in self._params:
             dst[name].copydata(src[name])
+        array_types = numpy.ndarray, cuda.ndarray
+        for name in self._persistent:
+            d = dst[name]
+            s = src[name]
+            if isinstance(d, array_types) and isinstance(s, array_types):
+                cuda.copyto(d, s)
+            else:
+                dst[name] = copy.deepcopy(s)
 
     def cleargrads(self):
         """Clears all gradient arrays.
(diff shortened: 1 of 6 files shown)


beam2d:
This PR also includes a utility function `cuda.copyto`, which simplifies the implementation.

chainer/link.py
```diff
@@ -508,21 +508,38 @@ def children(self):
...
+            copy_persistent (bool): If ``True``, persistent values are also
+                copied. ``True`` by default.
 
         """
         src = link.__dict__
         dst = self.__dict__
         for name in self._params:
             dst[name].copydata(src[name])
+        if copy_persistent:
+            array_types = numpy.ndarray, cuda.ndarray
```
kmaehashi: Use `chainer.get_array_types()`.

kmaehashi:
During the review I found another issue regarding iDeep: #5009. (Let's fix it after merging this PR)

chainer/link.py
beam2d: fixed

beam2d:
I'll fix for ideep after merging #5009.

beam2d:
I fixed for the comment. The iDeep issue is left in this PR.

tests/chainer_tests/test_link.py
```diff
@@ -397,6 +401,22 @@ def test_copyparams(self):
...
+        l, gs = self._setup_test_copyparams()
+        self.link.copyparams(l)
+        self._check_copyparams(l, gs)
+        numpy.testing.assert_array_equal(self.link.p, l.p)
+
+    def test_copyparams_no_copy_persistent(self):
+        orig_p = self.link.p.copy()
+
+        l, gs = self._setup_test_copyparams()
+        numpy.testing.assert_array_equal(False, orig_p == l)
```
kmaehashi: `l` -> `l.p`

kmaehashi:
LGTM!

iwiwi:
Thank you so much!!

kmaehashi:
For later reference, this fixes the issue (#4902) that ParallelUpdater does not share persistent values.

PR #5002, opened 2018-06-26
Fix verb error in chainer.functions.fft docstring

butsugiri:
I found a minor grammatical error in the docstring of chainer.functions.fft
https://github.com/chainer/chainer/blob/4bd5237f45f384a05490e92aa6c58bd0d7e134e6/chainer/functions/math/fft.py#L57


okuta:
Thanks. LGTM!

PR #5010, opened 2018-06-26
Add notes about relationship between dilated_convolution_2d and convolution_2d

kmaehashi:


kmaehashi:
Note: `dilate` option has been introduced in #2858.

okuta:
LGTM!

PR #5011, opened 2018-06-26
Fix broken notations in `F.linear` docs

kmaehashi:
Fix screen shot 2018-06-26 at 20 44 22
https://docs.chainer.org/en/latest/reference/generated/chainer.functions.linear.html

### DIFF
--- a/chainer/functions/connection/linear.py
+++ b/chainer/functions/connection/linear.py
@@ -201,21 +201,21 @@ def linear(x, W, b=None, n_batch_axes=1):
             ..., s_n)`-shaped float array. Its first ``n_batch_axes``
             dimensions are handled as *minibatch dimensions*. The
             other dimensions are handled as concatenated one dimension whose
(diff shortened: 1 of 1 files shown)


hvy:
jenkins, test this please.

hvy:
LGTM, but I'm wondering whether we should remove the parenthesis too. What do you think?

kmaehashi:
Do you mean parenthesis after these two 
> The other dimensions are handled as concatenated one dimension whose size must be
and 
> Weight variable of shape (M,N), where
? I think that makes sense.

hvy:
Indeed, I meant those two. They simply look like conditions that must hold true, rather than explaining shapes. But let's leave it as is if you think it looks okay (and since the've been written in that way since before).

hvy:
Removed `to-be-backported` since https://github.com/chainer/chainer/pull/4204 wasn't.

PR #5012, opened 2018-06-26
Add rules regarding use of pytest module

### LINKED ISSUE #4557: Policy on whether to use PyTest features
We currently have no policy on whether to use PyTest-specific features in unit tests.
In order to avoid confusion, we want a policy.

(previous discussion: https://github.com/chainer/chainer/pull/4534#discussion_r177633250)

kmaehashi:
Fixes #4557.

### DIFF
--- a/docs/source/contribution.rst
+++ b/docs/source/contribution.rst
@@ -346,10 +346,15 @@ They simply use the :mod:`unittest` package of the standard library, while some 
 In addition to the :ref:`coding-guide` mentioned above, the following rules are applied to the test code:
 
 * All test classes must inherit from :class:`unittest.TestCase`.
-* In addition to the :mod:`unittest` features, you can also use the following ``pytest`` features.
+* Use :mod:`unittest` features to write tests, except for the following cases:
 
-    * ``assert`` statement instead of ``self.assert*`` methods (e.g., you can write ``assert x == 1`` instead of ``self.assertEqual(x, 1)``).
-    * ``with pytest.raises(...):`` instead of ``with self.assertRaises(...):``.
+    * Use ``assert`` statement instead of ``self.assert*`` methods (e.g., write ``assert x == 1`` instead of ``self.assertEqual(x, 1)``).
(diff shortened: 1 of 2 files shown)


beam2d:
LGTM. @niboshi Could you double-check it?

docs/source/contribution.rst
```diff
@@ -341,7 +341,15 @@ How to Write Tests
 ~~~~~~~~~~~~~~~~~~
 
 There are many examples of unit tests under the :tree:`tests` directory, so reading some of them is a good and recommended way to learn how to write tests for Chainer.
-They simply use the ``unittest`` package of the standard library, while some tests are using utilities from :mod:`chainer.testing`.
+They simply use the :mod:`unittest` package of the standard library, while some tests are using utilities from :mod:`chainer.testing`.
+
+In addition to the :ref:`coding-guide` mentioned above, the following rules are applied to the test code:
+
+* All test classes must inherit from :class:`unittest.TestCase`.
+* In addition to the :mod:`unittest` features, you can also use the following ``pytest`` features.
```
niboshi: "should" or "must" instead of "can"?

niboshi: I don't think so. New tests should be written in the new convention. We will simply loosen the review criteria.
Also it will take too long (perhaps never) to rewrite all the existing tests.

kmaehashi: OK. I think `should` can be used along with the following excuse. Do you think it's OK?

```
We're incrementally applying the above style.
Some existing test codes may using old style (`self.assertRaises`, etc.), but all newly written test code should follow the above style.
```

niboshi: Yes, I think it's good to include that excuse.

Some minor corrections:
* "test codes" -> "tests"
* "may using" -> "may be using"
* "old style" -> "the old style"
* "newly written test code" -> "newly written tests"

kmaehashi: Thanks! I updated accordingly.

kmaehashi:
Jenkins, test this please.

niboshi:
LGTM

PR #5025, opened 2018-06-27
Add Bernoulli distribution

YoshikawaMasashi:
split from #4678.

### DIFF
--- a/chainer/distributions/bernoulli.py
+++ b/chainer/distributions/bernoulli.py
@@ -3,7 +3,6 @@ from chainer.backends import cuda
 from chainer import distribution
 from chainer.functions.activation import sigmoid
 from chainer.functions.array import broadcast
-from chainer.functions.array import reshape
 from chainer.functions.array import where
 from chainer.functions.math import exponential
 from chainer.functions.math import logarithm_1p
--- a/chainer/distributions/bernoulli.py
+++ b/chainer/distributions/bernoulli.py
@@ -19,8 +19,9 @@ class BernoulliLogProb(chainer.function_node.FunctionNode):
         self.retain_inputs((0, 1))
         xp = cuda.get_array_module(x)
         y = logit * (x - 1) - xp.log(xp.exp(-logit) + 1)
+        y = utils.force_array(y)
 
-        self.invalid = xp.bitwise_and(x != 0, x != 1)
+        self.invalid = utils.force_array(xp.bitwise_and(x != 0, x != 1))
         y[self.invalid] = - xp.inf
 
         # extreme logit
@@ -56,8 +57,8 @@ class ModifiedXLogX(chainer.function_node.FunctionNode):
 
     def forward(self, inputs):
         x, = inputs
-        self.x_zero = (x == 0)
-        y = x * self._logx.array
+        self.x_zero = utils.force_array(x == 0)
+        y = utils.force_array(x * self._logx.array)
         y[self.x_zero] = 0.
         return y,
 
@@ -72,10 +73,6 @@ class ModifiedXLogX(chainer.function_node.FunctionNode):
 
 
 def _bernoulli_log_prob(logit, x):
-    if x.ndim == 0:
-        x = reshape.reshape(x, (1,))
-    if logit.ndim == 0:
-        logit = reshape.reshape(logit, (1,))
     y, = BernoulliLogProb().apply((logit, x))
     return y
 
@@ -83,10 +80,9 @@ def _bernoulli_log_prob(logit, x):
 def _modified_xlogx(x):
     x = chainer.as_variable(x)
     xp = x.xp
-    if x.ndim == 0:
-        x = reshape.reshape(x, (1,))
     return ModifiedXLogX(exponential.log(
-        where.where(x.array > 0, x, xp.ones_like(x.array)))).apply((x,))[0]
+        where.where(utils.force_array(x.array > 0),
+                    x, xp.ones_like(x.array)))).apply((x,))[0]
 
 
 class Bernoulli(distribution.Distribution):
--- a/chainer/distributions/bernoulli.py
+++ b/chainer/distributions/bernoulli.py
@@ -32,7 +32,7 @@ class BernoulliLogProb(chainer.function_node.FunctionNode):
         y[self.to_zero] = 0.
         y[self.to_m_inf] = - xp.inf
 
-        return utils.force_array(y, x.dtype),
+        return utils.force_array(y, logit.dtype),
 
     def backward(self, indexes, grad_outputs):
         gy, = grad_outputs
--- a/chainer/distributions/bernoulli.py
+++ b/chainer/distributions/bernoulli.py
@@ -3,10 +3,13 @@ from chainer.backends import cuda
 from chainer import distribution
 from chainer.functions.activation import sigmoid
 from chainer.functions.array import broadcast
+from chainer.functions.array import reshape
+from chainer.functions.array import where
 from chainer.functions.math import exponential
 from chainer.functions.math import logarithm_1p
 from chainer import utils
 import numpy
+import warnings
 
 
 class BernoulliLogProb(chainer.function_node.FunctionNode):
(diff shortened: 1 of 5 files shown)


YoshikawaMasashi:
jenkins, test this please.

chainer/distributions/bernoulli.py
```diff
@@ -0,0 +1,100 @@
...
+from chainer.backends import cuda
+from chainer import distribution
+from chainer.functions.activation import sigmoid
+from chainer.functions.array import broadcast
+from chainer.functions.math import exponential
+import numpy
+
+
+class Bernoulli(distribution.Distribution):
+
```
toslunar: Could you add a docstring with meaning of arguments `p`, `logits`?

chainer/distributions/bernoulli.py
```diff
@@ -0,0 +1,119 @@
...
+                "Either `p` or `logit` (not both) must have a value.")
+
+        with chainer.using_config('enable_backprop', True):
+            if p is None:
+                self.logit = chainer.as_variable(logit)
+                self.p = sigmoid.sigmoid(self.logit)
+            else:
+                self.p = chainer.as_variable(p)
+                self.logit = exponential.log(self.p) - \
+                    exponential.log(1. - self.p)
```
toslunar: To avoid casting, `log1p(-self.p)` can be used.  Maybe you'd like to cache `p.dtype.type(1) - p` in the both cases.

chainer/distributions/bernoulli.py
```diff
@@ -0,0 +1,119 @@
...
+
+    @property
+    def batch_shape(self):
+        return self.p.shape
+
+    @property
+    def entropy(self):
+        return -self.p * exponential.log(self.p) - \
```
toslunar: While `x * log(x)` converges to `0` as x → +0, the result is `0 * inf = nan`.

chainer/distributions/bernoulli.py
```diff
@@ -0,0 +1,119 @@
...
+            (numpy.float32(1.) - self.p) \
```
toslunar: How about the following?
```
p = self.p
q = p.dtype.type(1.) - p
```

chainer/distributions/bernoulli.py
```diff
@@ -0,0 +1,119 @@
...
+    @property
+    def event_shape(self):
+        return ()
+
+    @property
+    def _is_gpu(self):
+        return isinstance(self.p.data, cuda.ndarray)
+
+    def log_prob(self, x):
+        if type(x) == chainer.Variable:
```
toslunar: Use `isinstance` to support subclasses (e.g. `chainer.Parameter`).

chainer/distributions/bernoulli.py
```diff
@@ -0,0 +1,119 @@
...
+            x = x.data
```
toslunar: Since `distribution` is new, `.array` is preferred to `.data`.

chainer/distributions/bernoulli.py
```diff
@@ -0,0 +1,119 @@
...
+    def stddev(self):
+        return (self.p * (1 - self.p)) ** 0.5
+
+    @property
+    def support(self):
+        return '{0, 1}'
+
+    @property
+    def variance(self):
+        return (self.p * (1 - self.p))
```
toslunar: The outer parentheses is unnecessary.

tests/chainer_tests/distributions_tests/test_bernoulli.py
```diff
@@ -0,0 +1,39 @@
+from chainer import distributions
+from chainer import testing
+import numpy
+
+
+@testing.parameterize(*testing.product({
+    'shape': [(3, 2), (1,)],
+    'is_variable': [True, False],
+    'sample_shape': [(3, 2), ()],
+}))
```
toslunar: Add a boolean parameter to test with `p` that contains `0` and `1`.
(You don't have to add it to the KL divergence test.)

chainer/distributions/bernoulli.py
```diff
@@ -0,0 +1,119 @@
...
+    def _is_gpu(self):
+        return isinstance(self.p.data, cuda.ndarray)
+
+    def log_prob(self, x):
+        if type(x) == chainer.Variable:
+            x = x.data
+        x = x.astype(self.p.dtype)
+        if self._is_gpu:
+            invalid_inf = cuda.cupy.zeros_like(x)
+            invalid = cuda.cupy.bitwise_and(x != 0, x != 1)
```
toslunar: Why do you want to validate `x`?  Do you expect other integers, or any floats?

chainer/distributions/bernoulli.py
```diff
@@ -0,0 +1,119 @@
...
+
+    @property
+    def batch_shape(self):
+        return self.p.shape
+
+    @property
+    def entropy(self):
+        return -self.p * exponential.log(self.p) - \
+            (numpy.float32(1.) - self.p) \
+            * exponential.log(numpy.float32(1.) - self.p)
```
toslunar: Be consistent in breaking a line after/before binary operator (pycodestyle's W503 or W504).  I prefer linebreaks before binary operators (as this line).

chainer/distributions/bernoulli.py
YoshikawaMasashi: Oh...
As a matter of fact, when `p=0`, entropy is `-q * log(q) = 0`.
I'll fix. (but when `p=0`, entropy is not differentiable.)

chainer/distributions/bernoulli.py
YoshikawaMasashi: Yes. When x is not 0 or 1, this function returns -inf.

tests/chainer_tests/distributions_tests/test_bernoulli.py
YoshikawaMasashi: For now, I don't expect `p` whose type is `bool`, and we get error when boolean `p` is used (Because `F.log` does not support bool.).
And scipy.stats also don't support boolean `p`.
I think we don't have to support boolen `p`. How about this?

toslunar: What I tried to say is
```
if self.p_has_extreme_values:
    some_indices = ...
    p[some_indices] = np.random.randint(2, size=...)
```

chainer/distributions/bernoulli.py
YoshikawaMasashi: Um.., I also think that it is better to support an option not to validate as like `F.bernoulli_nll`.

chainer/distributions/bernoulli.py
```diff
@@ -0,0 +1,186 @@
...
+        return gy * dx,
+
+
+def _bernoulli_log_prob(logit, x):
+    y, = BernoulliLogProb().apply((logit, x))
+    return y
+
+
+def _modified_xlogx(x):
+    return ModifiedXLogX(exponential.log(x)).apply((x,))[0]
```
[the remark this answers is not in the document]
YoshikawaMasashi: Oh, `where` is useful..

YoshikawaMasashi:
jenkins, test this please.

chainer/distributions/bernoulli.py
```diff
@@ -59,27 +56,40 @@ def __init__(self, logx):
...
     return y
 
 
 def _modified_xlogx(x):
-    return ModifiedXLogX(exponential.log(x)).apply((x,))[0]
+    xp = cuda.get_array_module(x)
+    if x.ndim == 0:
+        x = reshape.reshape(x, (1,))
+    if isinstance(x, chainer.Variable):
+        return ModifiedXLogX(exponential.log(
```
toslunar: Do not repeat the application of `ModifiedXLogX`. How about storing logx, or
```
x = chainer.as_variable(x)
xp = x.xp
```

YoshikawaMasashi: Oh yes!

YoshikawaMasashi:
> Could you explain why reshape is needed if ndim == 0?

```
return ModifiedXLogX(exponential.log(
            where.where(x.array > 0, x, xp.ones_like(x.array)))).apply((x,))[0]
```
When `x.ndim == 0`, `x.array > 0` become scalar. And `where` does not work.

toslunar:
How about `utils.force_array(x.array > 0)`?

YoshikawaMasashi:
fixed. thanks.

YoshikawaMasashi:
jenkins, test this please.

YoshikawaMasashi:
jenkins, test this please.

PR #5027, opened 2018-06-27
Move add to mathematical functions in docs

Crissman:
chainer.functions.add looks lonely by itself as the only "arithmetic" function. I think it fits better with "mathematical" functions.


kmaehashi:
> chainer.functions.add looks lonely by itself as the only "arithmetic" function.

That's why I added the description here; https://github.com/chainer/chainer/commit/29a7d5195579aae01d4ad6868a3c5e7c838b0c67#diff-4801516dbf888a00ecf4b2b270ce2f83R38

I think `add` should not be categorized into activation function.

Crissman:
Ok. Re-reading the note, I understand the reason for the separate category. Closing PR.

PR #5037, opened 2018-06-28
Fix attribute name collisions in docstring

toslunar:
The attribute `gamma` in https://docs.chainer.org/en/v5.0.0b2/reference/generated/chainer.links.BatchNormalization.html#chainer.links.BatchNormalization links to the document of `BatchRenormalization`.  Fix this kind of errors.

cf. #3068


toslunar:
Now we can revert some changes in #3068.
Jenkins, test this please.

toslunar:
Jenkins, test this please.

chainer/_sphinx_ext.py
```diff
@@ -0,0 +1,40 @@
...
+        except AttributeError:
+            q = klass.__name__
+        return '~%s.%s' % (q, attr_name)
+    return attr_name
+
+
+def setup(app):
+    app.setup_extension('sphinx.ext.napoleon')
+
+    def _parse_attributes_section(self, section):
```
kmaehashi: I haven't looked into the details yet, but could you write down some comment about the difference to the original `_parse_attributes_section` function?

chainer/_sphinx_ext.py
```diff
@@ -0,0 +1,40 @@
+import sphinx
```
kmaehashi: I think it's better to place this file under `docs/source` with name like `_napoleon_patch.py` or sth.
Then `extensions = ['sphinx.ext.autodoc', ... , '_napoleon_patch']` in `conf.py` should work.

kmaehashi:
Also to-be-ported to CuPy.

toslunar:
I submitted the same fix to the official repo since the change might be acceptable to others.
sphinx-doc/sphinx#5129

kmaehashi:
Jenkins, test this please.

kmaehashi:

PR #5042, opened 2018-06-30
Balanced order sampler

corochann:
This is related to below issue and PR
 - https://github.com/chainer/chainer/issues/2860
 - https://github.com/chainer/chainer/pull/3429

Now `OrderSampler` is introduced and we can customize the internal sampling strategy of `Iterator`.
This PR introduces `BalancedOrderSampler` (Actually the name "Order" is not necessary but I keep it since base class name is "OrderSampler"), which received `labels` information in its constructor and it will sample each examples evenly according to the label.

I also changed the following for further order_samplers development:
 - structure of order_samplers: I introduced `order_samplers` directory to separate each module in each file.
 - serialize order_sampler: `BalancedOrderSampler` has internal value to be serialized. So I changed 1. `OrderSampler` class to have `serialize` interface. 2. each iterator to serialize `order_sampler`.

### DIFF
--- a/chainer/functions/array/get_item.py
+++ b/chainer/functions/array/get_item.py
@@ -8,6 +8,10 @@ from chainer.utils import type_check
 from chainer import variable
 
 
+_numpy_supports_0d_bool_index = \
+    numpy.lib.NumpyVersion(numpy.__version__) >= '1.13.0'
+
+
 class GetItem(function_node.FunctionNode):
 
     """Function that slices array and extract elements."""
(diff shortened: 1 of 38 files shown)


corochann:
FYI: @yuyu2172 @hvy

corochann:
I tested `trainer` serialization with `BalancedOrderSampler` in below branch of chainer chemistry.
https://github.com/corochann/chainer-chemistry/blob/tox21_balanced_order_sampler/examples/tox21/train_tox21_balanced_order_sampler.py

```
# training with BalancedOrderSampler
python train_tox21_balanced_order_sampler.py -f 1
# resume training with BalancedOrderSampler
python train_tox21_balanced_order_sampler.py -f 1 --out=result_resume --resume=result/snapshot_iter_562
```

corochann:
I think the current error occurs at `TestAbsoluteError_param_5` and `BatchNormalization` which are not related to this PR.
Can someone check it?

corochann:
Your code is short and seems nice.
As a user point of view, at least it is nice to have this kind of balancing/adjusting label weighting functionality.
I feel having both are better than having neither of them, but yes the functionality is little bit overlapped as you wrote. 
I want to know chainer team's opinion.

In my opinion, except `batch_balancing` option, your code includes my desire to balance the label of dataset. However it might be easier to have `BalancedOrderSampler` so that user can omit calculating `weights` when user want to evenly balance label. So I feel it is ok to consider having both sampler in core library.

grafi-tt:
@corochann @yuyu2172 I think we need not to address serialization in this PR. Correct serialization of `order_sampler` is rather difficult because `MultiprocessIterator` and `MultithreadIterator` performs prefetching.

To support both of serialization and prefetching, order sampler should not overwrite old states. But current design assumes that an order sampler updates the internal PRNG states when it is called.

As a user, I personally prefer to the weighted order sampler, rather than the balanced one. It is easy to calculate labels' frequency by `numpy.bincount`. Sometimes I'd like to manually tune the weights.

@okuta I suppose a decision has to be made by Chainer team.

okuta:
@beam2d Do you have any comments about this PR?

grafi-tt:
bump

PR #5046, opened 2018-07-02
Let `F.depthwise_convolution_2d` use `F.convlution_2d` as an implementation

### LINKED ISSUE #4985: Replace the implementation of `depthwise_convolution_2d` with grouped convolution
The implementation of `depthwise_convolution_2d` is currently a plain im2col+matmul, which is not efficient in both speed and memory consumption. Depthwise convolution is a special case of grouped convolution supported by `convolution_2d`, which should be much faster and more memory efficient because it uses cuDNN. We should make `depthwise_convolution_2d` delegate the computation to `convolution_2d`.

It's also straightforward to support `nd` variant of depthwise convolution, thanks to recently-added grouped convolution support in `convolution_nd` (#4591).

beam2d:
Fix #4985. It enables `F.depthwise_convolution_2d` to use cuDNN if available, with which both speed and memory efficiency are improved.

There is one caveat in this PR: due to the differences in the weight layout between `F.depthwise_convolution_2d` and `F.convolution_2d`, the new implementation transposes and reshapes the weight array, which causes copy when the input weight is C-contiguous. The time/space overhead incurred by this subtle problem should be much smaller than that of im2col, but it is not zero. To remove this overhead, we have to change the weight layout from `(multiplier, in_channels, kh, kw)` to `(in_channels, multiplier, kh, kw)`, which breaks compatibility. The current PR does not include such a change to avoid compatibility breakage.


tests/chainer_tests/functions_tests/connection_tests/test_depthwise_convolution_2d.py
```diff
@@ -79,8 +77,8 @@ def check_backward(self, x_data, W_data, b_data, y_grad):
             args = args + (b_data,)
 
         gradient_check.check_backward(
-            depthwise_convolution_2d.DepthwiseConvolution2D(
-                self.stride, self.pad),
+            lambda *x: functions.depthwise_convolution_2d(
+                *x, stride=self.stride, pad=self.pad),
```
toslunar: Use `functools.partial` or avoid the misleading name `x`.

chainer/functions/connection/depthwise_convolution_2d.py
```diff
@@ -1,124 +1,4 @@
...
-            gx = conv.col2im_gpu(gcol, self.sy, self.sx,
-                                 self.ph, self.pw, h, w)
-
-        if b is None:
-            return gx, gW
-        else:
-            gy = xp.rollaxis(gy, 1, 4)
-            gb = gy.sum(axis=(0, 1, 2))
-            return gx, gW, gb
+from chainer import functions
```
toslunar: Fix circular import (python<3.5) by importing the module `chainer.functions.connection.convolution_2d`.

tests/chainer_tests/functions_tests/connection_tests/test_depthwise_convolution_2d.py
beam2d: fixed by renaming it

chainer/functions/connection/depthwise_convolution_2d.py
beam2d: fixed by only importing `chainer`

toslunar:
Jenkins, test this please.

PR #5017, opened 2018-06-26
Avoid hasattr in `L.BatchNormalization`

toslunar:
See #3952.

PR #5050, opened 2018-07-03
[backport] Infer input size in batchnorm using aggregate axes

delta2323:
Backport of #4673

### DIFF
--- a/chainer/links/normalization/batch_normalization.py
+++ b/chainer/links/normalization/batch_normalization.py
@@ -33,8 +33,9 @@ class BatchNormalization(link.Link):
     fine-tuning mode.
 
     Args:
-        size (int or tuple of ints): Size (or shape) of channel
-            dimensions.
+        size (int, tuple of ints, or None): Size (or shape) of channel
+            dimensions.  If ``None``, the size will be determined from
+            dimension(s) of the input batch during the first forward pass.
         decay (float): Decay rate of moving average. It is used on training.
         eps (float): Epsilon value for numerical stability.
         dtype (numpy.dtype): Type to use in computing.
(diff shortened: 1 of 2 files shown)


delta2323:

delta2323:
According to @niboshi, this PR should be merged in the same release as the backport of #5017. Should we wait to merge it until the backport for #5017?

niboshi:
We should merge in this order.
1. #5017
2. This PR (#5050)
3. Backport of #5017

1\. should be in v5b3.
2\. and 3. should be in the same (corresponding) releases.

toslunar:
Without a backport of #4266, the inference of parameter sizes seems difficult.
🤔

delta2323:
I tried and was able to make the backport PR of #4266 (#5060). But it is a different question if we do merge it.

delta2323:
When I merged the original PR (#4266), I didn't make a backport PR as it was addition of new features. But I also think it is OK to do so. How do you think? @niboshi @toslunar

toslunar:
I'd rather not backport #4266.

BTW, lazy initialization of persistent arrays is new.  This backport PR (of #4673) itself is somewhat dangerous.

niboshi:
I don't fully understand the detail of related PRs. I'm going to leave the decision to both of you.

So it's better to skip backporting?

delta2323:
@toslunar Actually, I also think we do not have to backport #4266 and #4673 unless there is strong need for it because they modify the internal structure of `L.BatchNormalization` too much. Then, let's close the backport PRs.

delta2323:
Please reopen the PR if you have some comments or requests.

Issue #4582, opened 2018-04-09
FP16 support in various functions and links


beam2d:
Related to #4510. It looks the following functions and links should be fixed somehow to support fp16 (I checked them quickly, so if you find mistakes/lacking items, fix the list or let me know).

functions:
- [x] spatial_transformer_grid(#5114)
- [x] sparial_transformer_sampler(#5751)
- [x] n_step_rnn(#5808)
- [x] absolute_error(#5145)
- [x] contrastive(#5152)
- [x] cross_covariance(#5158)
- [x] connectionist_temporal_classification(#5680)
- [x] decov(#5174)
- [x] hinge(#5175)
- [x] huber_loss(#5176)
- [x] mean_absolute_error(#5053)
- [x] mean_squared_error(#5052)
- [x] negative_sampling(#5336)
- [x] sigmoid_cross_entropy(#5211)
- [x] squared_error(#5212)
- [x] triplet(#5214)
- [x] batch_l2_norm_squared(#5235)
- [x] det / batch_det(#5323)
- [x] inv / batch_inv(#5323)
- [x] gaussian(#5284)
- [x] normalize(#5256)
- [x] roi_pooling_2d(#5281)
- [x] embed_id(#5624)
- [x] scatter_add(#5335)
- [x] get_item(#5335)

links:
- [x] Maxout(#5058)
- [x] InceptionBN(#5062)
- [x] StatefulMGU(#5084)
- [x] BlackOut(#5638)
- [x] BinaryHierarchicalSoftmax(#5648)
- [x] NegativeSampling(#5664)
- [x] GoogLeNet(#5099)
- [x] ResNetLayers(#5101)
- [x] VGG16Layers(#5107)
- [x] BatchNormalization(#5034)
- [x] BarchRenormalization(#5034)

kmaehashi:
I tentatively set assignee, but still welcoming others to divide the work.

takagi:
I will check if any other functions and links are left to need work.

gwtnb:
It seems that functions and links in ChainerMN are omitted. e.g. https://github.com/chainer/chainer/blob/f4ba08a521b9154f619b70634b7c5a4f7249944e/chainermn/links/batch_normalization.py#L48

takagi:
Currently the original check list is completed. Can I close this issue as solved?

kmaehashi:
Everything listed has been merged. Thanks!

PR #5053, opened 2018-07-03
Support all float dtypes in `F.mean_absolute_error`

takagi:
This PR is a part of #4582 and generalizes `F.mean_absolute_error` to support all float dtypes other than `float32`, with losing the condition of inputs.


beam2d:
Looks good.
Jenkins, test this please.

PR #5058, opened 2018-07-04
Support default dtype in `L.Maxout`

takagi:
This PR is a part of #4582, makes `L.Maxout` use the default dtype to initialize its initial bias when a scalar value is supplied for it.

### DIFF
--- a/chainer/links/activation/maxout.py
+++ b/chainer/links/activation/maxout.py
@@ -1,5 +1,6 @@
 import numpy
 
+import chainer
 from chainer.backends import cuda
 from chainer.functions.activation import maxout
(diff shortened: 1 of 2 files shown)


beam2d:
It is, strictly speaking, not related to this change, but the documentation says `initialW` and `initial_bias` can be initializers, but actually the code looks assuming that they are scalars or ndarrays. Could you fix it? It's ok to do it either in this PR or in another PR.

takagi:
Yes, I will do it in another PR!

okuta:
jenkins, test this please.

okuta:
LGTM!

PR #5061, opened 2018-07-05
Fix kernels not memorized

kmaehashi:
Some functions do not cache kernels correctly, causing memory leak.


niboshi:
Why weren't they released?

kmaehashi:
There is another issue in CuPy. I'll send PR to CuPy.

niboshi:
LGTM

PR #5064, opened 2018-07-05
Initialize gradient of uninitialized parameter with default dtype when initializer is callable

takagi:
This PR follows up #4510.

The gradient of an uninitialized parameter was always initialized with NaN of `numpy.float32` when its initializer was a callable, in contrast to its data that was initialized with the default dtype. This caused mismatch of dtypes between the data and the gradient.

This PR fixes it to use the default dtype to initialize the gradient in such case so that the data and the gradient have the same dtype.

### DIFF
--- a/chainer/variable.py
+++ b/chainer/variable.py
@@ -1331,7 +1331,7 @@ class Parameter(Variable):
             else:
                 # uninitialized parameter
                 super(Parameter, self).__init__(name=name)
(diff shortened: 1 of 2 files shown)


mitmul:
Jenkins, test this please.

takagi:
Jenkins, test this please.

okuta:
LGTM!

PR #5067, opened 2018-07-06
Support DALI

anaruse:
This PR aims to support NVIDIA DALI to accelerate image data loading to GPU.

In this PR, a iterator for DALI named `DaliIterator` has been added. Current iterators like `SerialIterator` and `MultiprocessIterator` are created with an instance of `DatasetMixin`, and training data, in other words, examples are loaded through the dataset instance with explicit index. In contract, `DaliIterator` is created with an instance of DALI `Pipeline` and training examples are pre-processed (decode, normalization, augmentation, etc) in the pipeline according to its configurations and the data is passed to the iterator in the form of "tuple of DALI array". That means there are big difference between outputs of current iterators and `DaliIterator`. The difference is solved by a converter named `dali_converter` in which an output of DALI `Pipeline` is converted to "tuple of numpy/cupy array". I've also edited imagenet example so that you can test DALI with actual training dataset.

There are still several things to fix/add (tests for example), but first I would like to hear your thought on whether this basic design is OK or not.


examples/imagenet/train_imagenet.py
```diff
@@ -24,6 +25,78 @@
...
+        return x
+    return x, x
+
+
+class ImagenetDaliPipeline(pipeline.Pipeline):
+
+    def __init__(self, file_list, file_root, crop_size,
+                 batch_size, num_threads, device_id,
+                 random_shuffle=False, seed=-1,
+                 mean=[0.485 * 255, 0.456 * 255, 0.406 * 255],
```
okuta: Please use tuple or `None`.

chainer/iterators/dali_iterator.py
```diff
@@ -0,0 +1,91 @@
...
+
+    Args:
+        pipeline: DALI pipeline.
+        repeat (bool): If ``True``, it infinitely loops over the dataset.
+            Otherwise, it stops iteration at the end of the first epoch.
+
+    """
+
+    def __init__(self, pipeline, repeat=True):
+        pipeline.build()
```
kmaehashi: Does this operation cause CUDA context initialization? In that case I think it's better to move it to `__next__` to defer initialization.

examples/imagenet/train_imagenet.py
```diff
@@ -24,6 +25,78 @@
...
+
+    def define_graph(self):
+        jpegs, labels = self.loader()
+        images = self.decode(jpegs)
+        images = self.rrcrop(images.gpu())
+        images = self.cmnorm(images, mirror=self.coin())
+        return [images, labels]
+
+
+def dali_converter(inputs, device=None):
```
kmaehashi: I think it's better to provide this feature in the framework, rather than in an example.
I even think it's possible to merge this into existing `concat_examples`, but wondering if it is too aggressive... (how do you think? @okuta)

examples/imagenet/train_imagenet.py
```diff
@@ -24,6 +25,78 @@
...
+import ctypes
+
+
+def _pair(x):
+    if hasattr(x, '__getitem__'):
+        return x
+    return x, x
+
+
+class ImagenetDaliPipeline(pipeline.Pipeline):
```
kmaehashi: nitpick: it will not work if `_dali_available == False` as `pipeline` is undefined.

examples/imagenet/train_imagenet.py
anaruse: Got it. will use `None` there.

chainer/iterators/dali_iterator.py
anaruse: Though I'm not sure if CUDA context is created there, I agree with you that it's safer to move it to `__next__`.

examples/imagenet/train_imagenet.py
anaruse: Thanks, I did not notice that. will add some workaround.

anaruse:
I've just updated the branch based on your comments, except the place of `dali_converter`. Please take a look at it again.

examples/imagenet/train_imagenet.py
anaruse: I agree with you that it is better to put `dali_converter` in the framework. But, I'm not sure whether it is good idea to merge this feature into `concat_examples`.

anaruse:
I've added DaliConvertor to imagenet example in order to support per-pixel subtraction that is used as default augmentation in your imagenet example. Now, you can convert images with DALI as you do with current imagenet example.

okuta:
Do you have a benchmark result?

examples/imagenet/train_imagenet.py
```diff
@@ -24,6 +25,191 @@
 import resnet50
 import resnext50
 
+try:
```
okuta: I feel that adding to much.
- Create new train script as `train_imagenet_dali.py` .
- Move these code to separate utility file `dali_util.py` .

okuta:
This PR has no test. But we want to merge this PR in v5rc1.
So, cloud you use `chainer.utils.experimental` ?

kmaehashi:
This PR has been taken over to #5387.

PR #5069, opened 2018-07-07
Add a comment that explains why the Classifier chain clears Variable attribues.

grafi-tt:
This PR makes the Classifier chain better explains the importance to clear the Variable attirubtes. It prevents sloppy users defines their own chain that consumes unnecessary memory. 

The docstring of `chainer.links.model.Classifier` says:

    This is an example of chain that wraps another chain. It computes the
    loss and accuracy based on a given input/label pair.

So it is reasnoable to assume that users often view and modify the source to define a custom chain. In fact, such a double memory allocation bug has actually happend in a project I'm work for.

### DIFF
--- a/chainer/links/model/classifier.py
+++ b/chainer/links/model/classifier.py
@@ -128,9 +128,15 @@ class Classifier(link.Chain):
             t = kwargs[self.label_key]
             del kwargs[self.label_key]
 
+        # The model keeps these variables on its attributes for ease of later
+        # accesses. In such a case, we shall clear the attributes before
+        # running the predictor on another batch. If we didn't, memory
+        # would be doubly allocated, because the variables retain the whole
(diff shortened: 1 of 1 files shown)


mitmul:
Thank you for the PR! How about putting the explanation into the docstring of the class `Classifier`? If you put the explanation to the docstring, both users who are modifying this class and those who will modify the code can see the explanation. So I think it's good to have it in the documentation to make it visible to all.

grafi-tt:
@mitmul Sorry for late reply. I moved the comment to a note on docstrings (with change of wordings).

mitmul:
Jenkins, test this please

mitmul:
Welcome back! Jenkins, test this please

mitmul:
LGTM

PR #5057, opened 2018-07-04
Clarify distinction between `chainer.dataset` and `chainer.datasets`

toslunar:

PR #5072, opened 2018-07-09
[backport] Clarify distinction between `chainer.dataset` and `chainer.datasets`

okuta:
Backport of #5057


toslunar:
Jenkins, test this please.

PR #5077, opened 2018-07-10
Support 0-dim params in distributions

toslunar:
This PR uses #5075.


toslunar:
Jenkins, test this please.

kmaehashi:
According to @toslunar, this feature does not require #5075 (which was reverted in #5087.)

Issue #5026, opened 2018-06-27
Remove deprecation messages from Chainer Docs


Crissman:
As I read through the Chainer docs, many pages feel overly long due to the repeated deprecation and other warnings, often from v1 or v2 of Chainer. I feel this gets in the way of clean explanations, such as on this page: https://docs.chainer.org/en/stable/reference/generated/chainer.links.GRU.html

How about we remove the deprecations, instead displaying as code warnings or version release notes?

kmaehashi:
I'm thinking about the same thing. How about cleaning up deprecations since v2 (or earlier) from docs in v5?
cc/ @beam2d @niboshi

In that case it's better to make sure that everything is covered in the Upgrade Guide.
e.g. https://docs.chainer.org/en/stable/upgrade_v2.html#the-interfaces-of-gru-and-lstm-are-aligned

It's also related to #3233: inherited methods (altogether with docs including deprecation warnings) are displayed in subclasses.

beam2d:
+1 for removing documentation of methods deprecated for long time.

We have to be careful about issues related to un-deprecate some features of `Link` and `Chain` (#4106); these are deprecated since v2 but we are planning to make them non-deprecated again. We should not remove documents of these methods.

Crissman:
Completed.

PR #5081, opened 2018-07-10
remove v1 v2 references

Crissman:
Remove deprecation notices and style change notices for v1 and v2 of Chainer as per #5026

### DIFF
--- a/chainer/__init__.py
+++ b/chainer/__init__.py
@@ -229,21 +229,6 @@ def set_debug(debug):
 
 
 class DebugMode(object):
-    """Debug mode context.
-
-    This class provides a context manager for debug mode. When entering the
-    context, it sets the debug mode to the value of `debug` parameter with
-    memorizing its original value. When exiting the context, it sets the debug
-    mode back to the original value.
-
-    .. deprecated:: v2.0.0
-
-        Use :func:`chainer.using_config` instead. See :ref:`debug` for details.
-
-    Args:
-        debug (bool): Debug mode used in the context.
-    """
-
     def __init__(self, debug):
         warnings.warn('chainer.DebugMode is deprecated. '
                       'Use chainer.using_config("debug", ...) instead.',
--- a/chainer/__init__.py
+++ b/chainer/__init__.py
@@ -229,6 +229,21 @@ def set_debug(debug):
 
 
 class DebugMode(object):
+    """Debug mode context.
+
+    This class provides a context manager for debug mode. When entering the
+    context, it sets the debug mode to the value of `debug` parameter with
+    memorizing its original value. When exiting the context, it sets the debug
+    mode back to the original value.
+
+    .. deprecated:: v2.0.0
+
+        Use :func:`chainer.using_config` instead. See :ref:`debug` for details.
+
+    Args:
+        debug (bool): Debug mode used in the context.
+    """
+
     def __init__(self, debug):
         warnings.warn('chainer.DebugMode is deprecated. '
                       'Use chainer.using_config("debug", ...) instead.',
--- a/chainer/backends/cuda.py
+++ b/chainer/backends/cuda.py
@@ -266,10 +266,9 @@ def get_device_from_array(*arrays):
 def get_device(*args):
     """Gets the device from a device object, an ID integer or an array object.
 
-    .. note::
+    .. deprecated:: v2
 
-        This API is deprecated since v3.0.0. Please use
-        :func:`~chainer.backends.cuda.get_device_from_id`
+        Please use :func:`~chainer.backends.cuda.get_device_from_id`
         or :func:`~chainer.backends.cuda.get_device_from_array` instead.
 
     This is a convenient utility to select a correct device if the type of
--- a/chainer/backends/cuda.py
+++ b/chainer/backends/cuda.py
@@ -268,7 +268,7 @@ def get_device(*args):
 
     .. note::
 
-        This API is deprecated. Please use
+        This API is deprecated since v3.0.0. Please use
         :func:`~chainer.backends.cuda.get_device_from_id`
         or :func:`~chainer.backends.cuda.get_device_from_array` instead.
 
--- a/chainer/backends/cuda.py
+++ b/chainer/backends/cuda.py
@@ -266,9 +266,10 @@ def get_device_from_array(*arrays):
 def get_device(*args):
     """Gets the device from a device object, an ID integer or an array object.
 
-    .. deprecated:: v2
+    .. note::
 
-        Please use :func:`~chainer.backends.cuda.get_device_from_id`
+        This API is deprecated since v3.0.0. Please use
+        :func:`~chainer.backends.cuda.get_device_from_id`
         or :func:`~chainer.backends.cuda.get_device_from_array` instead.
 
     This is a convenient utility to select a correct device if the type of
--- a/chainer/gradient_check.py
+++ b/chainer/gradient_check.py
@@ -290,18 +290,6 @@ def numerical_grad(
 
 
 def assert_allclose(x, y, atol=1e-5, rtol=1e-4, verbose=True):
-    """Asserts if some corresponding element of x and y differs too much.
-
-    This function can handle both CPU and GPU arrays simultaneously.
-
-    Args:
-        x: Left-hand-side array.
-        y: Right-hand-side array.
-        atol (float): Absolute tolerance.
-        rtol (float): Relative tolerance.
-        verbose (bool): If ``True``, it outputs verbose messages on error.
-
-    """
     warnings.warn(
         'chainer.gradient_check.assert_allclose is deprecated.'
         'Use chainer.testing.assert_allclose instead.',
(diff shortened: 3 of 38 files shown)


docs/source/reference/debug.rst
```diff
@@ -34,13 +34,5 @@ You can check if debug mode is enabled with :func:`chainer.is_debug` function.
 Deprecated interface
```
kmaehashi: I think the section itself can be removed.

chainer/link.py
```diff
@@ -551,15 +551,6 @@ def cleargrads(self):
             param.cleargrad()
 
     def zerograds(self):
-        """Initializes all gradient arrays by zero.
```
kmaehashi: I think it's better not to remove the existing docs for deprecated methods.
It will show up as an undocumented method in the docs, which may confuse users.

kmaehashi:
There are some others that need to be removed:

https://docs.chainer.org/en/latest/reference/generated/chainer.Deserializer.html?highlight=v2.0.0
https://docs.chainer.org/en/latest/reference/generated/chainer.serializers.HDF5Serializer.html?highlight=v2.0.0

Crissman:
Thanks, @kmaehashi for pointing out the missing old version notes. I had renamed my local branch, and so the commits were not updating this branch. Fixed.

As for removing existing docs for deprecated methods, I think it would be best to remove those methods entirely from the docs. I looked for a way to do this with Read the Docs, but couldn't find one. Is there a way to do this?

If not, I'll add comments to those functions that they're deprecated.

Thanks!

kmaehashi:
As for deprecated module members like `DebugMode` and `assert_allclose`, just remove the reference from docs (https://github.com/chainer/chainer/pull/5081/files#diff-d7aa349852b1ed75022a62937030d452L46). Let's leave the pydocs (docs in the source) so that users can still understand why they're undocumented by looking into the code.

As for deprecated class members like `Link.zerograds`, I think there's no way to hide it, unfortunately.

niboshi:
>As for deprecated class members like `Link.zerograds`, I think there's no way to hide it, unfortunately.

It's a bit tricky but maybe we can blacklist in `docs/source/_templates/autosummary/class.rst`?

Crissman:
@niboshi I checked, and blacklisting in `class.rst` works! But, I don't think we should do that given @kmaehashi 's comments about leaving the docs for `zerograds`, so I've left it for now.

I think this is ok to merge. Please confirm!

chainer/backends/cuda.py
```diff
@@ -196,7 +196,7 @@ def get_device(*args):
 
     .. note::
 
-        This API is deprecated. Please use
+        This API is deprecated since v3.0.0. Please use
```
kmaehashi: It seems its already deprecated in v2.

chainer/link.py
```diff
@@ -554,13 +554,11 @@ def cleargrads(self):
...
         """Initializes all gradient arrays by zero.
 
-        This method can be used for the same purpose of cleargrads, but less
-        efficient. This method is left for backward compatibility.
+         .. deprecated:: v1.15
+            Use :meth:`cleargrads` instead.
 
-        .. deprecated:: v1.15
-           Use :meth:`cleargrads` instead.
+         """
```
kmaehashi: * Indentation seems broken.
* How about saying `Use :meth:`cleargrads` instead, which is more efficient.`? So users can understand this is not just a rename.

chainer/links/connection/mlp_convolution_2d.py
```diff
@@ -54,10 +54,8 @@ class MLPConvolution2D(link.ChainList):
             a keyword argument.
 
     .. note:
-        From v2, `conv_init` and `bias_init` arguments must be specified as
-        keyword arguments only. We impose this restriction to forbid
-        users to assume the API for v1 and specify `wscale` option,
-        that had been between `activation` and `conv_init` arguments in v1.
+        `conv_init` and `bias_init` arguments must be specified as
```
kmaehashi: I think the entire note block can be removed, as there are many other keyword-arguments-only methods.

chainer/links/connection/parameter.py
```diff
@@ -7,10 +7,6 @@ class Parameter(link.Link):
 
     """Link that just holds a parameter and returns it.
 
-    .. deprecated:: v1.5
```
kmaehashi: This is a documented API and thus cannot be removed.
https://docs.chainer.org/en/stable/reference/generated/chainer.links.Parameter.html?highlight=Parameter

chainer/optimizer.py
```diff
@@ -690,21 +690,6 @@ def update(self, lossfun=None, *args, **kwds):
         self.call_hooks('post')
 
     def use_cleargrads(self, use=True):
-        """Enables or disables use of :func:`~chainer.Link.cleargrads` in `update`.
```
kmaehashi: This is a documented api and thus cannot be removed.
https://docs.chainer.org/en/stable/reference/generated/chainer.GradientMethod.html#chainer.GradientMethod

chainer/variable.py
```diff
@@ -805,16 +800,6 @@ def cleargrad(self):
         self._grad_var = None
 
     def zerograd(self):
-        """Initializes the gradient array by zeros.
```
kmaehashi: This is an documented API. https://docs.chainer.org/en/stable/reference/generated/chainer.Variable.html#chainer.Variable.zerograd

How about fixing this as you did in Link?
https://github.com/chainer/chainer/pull/5081/files#diff-349b58823620f69f6f4ad574377a654d

docs/source/reference/debug.rst
kmaehashi: This section is a list of "Deprecated interface", and there's nothing to list as `DebugMode` has been removed.
So I think it's OK to remove the section.

chainer/backends/cuda.py
Crissman: Fixed in master already.

chainer/link.py
Crissman: Updated.

Crissman:
@kmaehashi Please confirm review ok, thanks.

chainer/__init__.py
```diff
@@ -229,21 +229,6 @@ def set_debug(debug):
 
 
 class DebugMode(object):
-    """Debug mode context.
```
kmaehashi: https://github.com/chainer/chainer/pull/5081#issuecomment-406478374

chainer/backends/cuda.py
```diff
@@ -266,10 +266,9 @@ def get_device_from_array(*arrays):
 def get_device(*args):
     """Gets the device from a device object, an ID integer or an array object.
 
-    .. note::
+    .. deprecated:: v2
 
-        This API is deprecated. Please use
-        :func:`~chainer.backends.cuda.get_device_from_id`
+        Please use :func:`~chainer.backends.cuda.get_device_from_id`
```
kmaehashi: This docs should not be changed as it is listed on the reference.

chainer/gradient_check.py
```diff
@@ -290,18 +290,6 @@ def iterate_single_input(i_in, x, orig_x, i):
 
 
 def assert_allclose(x, y, atol=1e-5, rtol=1e-4, verbose=True):
-    """Asserts if some corresponding element of x and y differs too much.
```
kmaehashi: https://github.com/chainer/chainer/pull/5081#issuecomment-406478374

chainer/optimizer.py
```diff
@@ -757,16 +757,16 @@ def use_cleargrads(self, use=True):
...
-                `cleargrads`. If ``False``, disables use of `cleargrads`
-                (`zerograds` is used).
-
-        .. deprecated:: v2.0
-           Note that :meth:`update` calls :meth:`~Link.cleargrads` by default.
-           :meth:`~Link.cleargrads` is more efficient than
-           :meth:`~Link.zerograds`, so one does not have to call
-           :meth:`use_cleargrads`. This method remains for backward
-           compatibility.
+             use (bool): If ``True``, this function enables use of
```
kmaehashi: Space unexpectedly inserted?

Crissman:
Review addressed. Please confirm, @kmaehashi -san!

chainer/optimizer.py
```diff
@@ -761,12 +761,12 @@ def use_cleargrads(self, use=True):
                 `cleargrads`. If ``False``, disables use of `cleargrads`
                 (`zerograds` is used).
 
-        .. deprecated:: v2.0
-           Note that :meth:`update` calls :meth:`~Link.cleargrads` by default.
-           :meth:`~Link.cleargrads` is more efficient than
-           :meth:`~Link.zerograds`, so one does not have to call
-           :meth:`use_cleargrads`. This method remains for backward
-           compatibility.
+         .. deprecated:: v2.0
```
kmaehashi: Space unexpectedly inserted?

Crissman:
Fixed! Thanks. @kmaehashi

kmaehashi:
LGTM! Jenkins, test this please.

kmaehashi:

Issue #3622, opened 2017-10-20
Probability distribution implementations


muupan:
Official implementations of various probability distributions that supports batch computation and differentiation would be useful.

Tensorflow has a thorough set of distributions under contrib. https://www.tensorflow.org/api_guides/python/contrib.distributions PyTorch seems to have started supporting distributions very recently, still limited though. http://pytorch.org/docs/master/distributions.html

ChainerRL has a set of distribution implementations, but they are still limited and designed for RL. http://chainerrl.readthedocs.io/en/latest/distributions.html I think it would be better to have a general-purpose and thorough set of distributions in Chainer.

muupan:
PyTorch now supports a lot of distributions: http://pytorch.org/docs/master/distributions.html

mitmul:
- Bernoulli
- Beta
- Binomial
- Categorical
- Cauchy
- Chi2
- Dirichlet
- Distribution
- Exponential
- FisherSnedecor
- Gamma
- Geometric
- Gumbel
- Laplace
- LogNormal
- Multinomial
- Normal
- OneHotCategorical
- Pareto
- StudentT
- Poisson
- Uniform
- TransformedDistribution
- biject_to
- kl_divergence
- register_kl

kmaehashi:
bump

PR #5085, opened 2018-07-11
Add transformed distribution

YoshikawaMasashi:
split from #4678  (#3622 ).
By using `TransformedDistribution` and `Bijector`, we can implement (normalizing) flow such as MAF and Glow easier.

### DIFF
--- a/chainer/distributions/bijector.py
+++ b/chainer/distributions/bijector.py
@@ -1,8 +1,8 @@
+import chainer
 from chainer.functions.math import exponential
-from chainer.functions.math import identity
 
 
-class Bijector():
+class Bijector(object):
 
     """Interface of Bijector.
 
@@ -14,30 +14,31 @@ class Bijector():
     def __init__(self):
         self.cache_x_y = (None, None)
 
-    def forward(self, x):
+    def forward_(self, x):
         old_x, old_y = self.cache_x_y
-        if id(x) == id(old_x):
+        if x is old_x:
             return old_y
-        y = self._forward(x)
+        y = self.forward(x)
         self.cache_x_y = x, y
         return y
 
-    def inv(self, y):
+    def inv_(self, y):
         old_x, old_y = self.cache_x_y
-        if id(y) == id(old_y):
+        if y is old_y:
             return old_x
-        x = self._inv(y)
+        x = self.inv(y)
         self.cache_x_y = x, y
         return x
 
-    def logdet_jac(self, x):
-        return self._logdet_jac(x)
+    def logdet_jac_(self, x):
+        return self.logdet_jac(x)
 
-    def _forward(self, x):
+    def forward(self, x):
         """Forward computation
 
         Args:
-            x(:class:`~chainer.Variable`): Data points in the domain of the
+            x(:class:`~chainer.Variable` or :class:`numpy.ndarray` or \
+            :class:`cupy.ndarray`): Data points in the domain of the
             based distribution.
 
         Returns:
@@ -46,11 +47,12 @@ class Bijector():
         """
         raise NotImplementedError
 
-    def _inv(self, y):
+    def inv(self, y):
         """Inverse computation
 
         Args:
-            x(:class:`~chainer.Variable`): Data points in the domain of the
+            y(:class:`~chainer.Variable` or :class:`numpy.ndarray` or \
+            :class:`cupy.ndarray`): Data points in the domain of the
             transformed distribution.
 
         Returns:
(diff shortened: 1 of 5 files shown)


chainer/distributions/transformed_distribution.py
```diff
@@ -0,0 +1,41 @@
...
+    def event_shape(self):
+        return self.base_distribution.event_shape
+
+    def cdf(self, x):
+        return self.base_distribution.cdf(self.bijector.inv(x))
+
+    def log_prob(self, x):
+        return self.base_distribution.log_prob(self.bijector.inv(x)) \
+            - self.bijector.logdet_jac(self.bijector.inv(x))
+
```
toslunar: Could you also implement `prob`?

chainer/distributions/transformed_distribution.py
```diff
@@ -0,0 +1,41 @@
...
+    @property
+    def event_shape(self):
+        return self.base_distribution.event_shape
+
+    def cdf(self, x):
+        return self.base_distribution.cdf(self.bijector.inv(x))
+
+    def log_prob(self, x):
+        return self.base_distribution.log_prob(self.bijector.inv(x)) \
```
toslunar: Do not repeat `self.bijector.inv(x)`.

tests/chainer_tests/distributions_tests/test_transformed_distribution.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+@testing.parameterize(*testing.product({
+    'shape': [(3, 2), (1,)],
+    'is_variable': [True, False],
+    'sample_shape': [(3, 2), ()],
+}))
+@testing.fix_random()
+@testing.with_requires('scipy')
+class TestLogTransformedDistribution(testing.distribution_unittest):
```
toslunar: This class tests `Exp`-transformed `Normal` distribution, which is a log-normal distribution.

chainer/distributions/bijector.py
```diff
@@ -0,0 +1,88 @@
...
+            return old_y
+        y = self._forward(x)
+        self.cache_x_y = x, y
+        return y
+
+    def inv(self, y):
+        old_x, old_y = self.cache_x_y
+        if id(y) == id(old_y):
+            return old_x
+        x = self._inv(y)
```
toslunar: `y` can be a `ndarray` with the current implementation of `TransformedDistribution.cdf`, while `self._inv` requires `Variable`.

chainer/distributions/bijector.py
```diff
@@ -0,0 +1,88 @@
...
+
+    """Interface of Bijector.
+
+    `Bijector` is implementation of bijective (invertible) function that is
+    used by `TransformedDistribution`. The three method `_forward`, `_inv` and
+    `_logdet_jac` have to be defined in inhereted class.
+    """
+
+    def __init__(self):
+        self.cache_x_y = (None, None)
```
toslunar: This caching mechanism seems a trade-off between speed and memory.  Users may want to turn off this feature.
This issue can be addressed in another PR.

toslunar: Wait.  There's a more significant issue.

In the current implementation of `TransformedDistribution.cdf`, we cannot compute the following.
```
d = TransformedDistribution(...)
y = d.sample(...)
q = d.cdf(y)
g, = chainer.grad(outputs=(q,), inputs=(y,))
```

YoshikawaMasashi: Oh, no...
Caching does not retain computational graph between input and output. But in many cases, `inv`  calculation is heavy and computational graph between input and output is needless.
Is it good solution that we add option to using cache? or is there any good solution that we can calculate gradient and cache?

chainer/distributions/bijector.py
YoshikawaMasashi: I fix docstring as like that `self._inv` requires `Variable` and `ndarray`.

YoshikawaMasashi:
jenkins, test this please.

ganow:
This PR is very attractive to me, but the development seems to be stopped due to the following problem (https://github.com/chainer/chainer/pull/5085#discussion_r201650176).

> Wait. There's a more significant issue.
> 
> In the current implementation of `TransformedDistribution.cdf`, we cannot compute the following.
> 
> ```
> d = TransformedDistribution(...)
> y = d.sample(...)
> q = d.cdf(y)
> g, = chainer.grad(outputs=(q,), inputs=(y,))
> ```

Pytorch avoids this problem by setting the flag for caching, and I think we can use the same technique.
https://github.com/pytorch/pytorch/blob/6a297b86754324d850d1595947c5e5559b51acc2/torch/distributions/transforms.py#L51-L52

toslunar:
It's fine to drop the cache feature.  IMO, unsafe feature should be added with very good documentation.

PR #5100, opened 2018-07-13
Fix dataset path to support Windows

okuta:
We have to use `os.path.join`.


kmaehashi:
I have no objection to merge this, but it seems the current code also works well on Windows.

Jenkins, test this please.

kmaehashi:
LGTM!

PR #5104, opened 2018-07-16
[WIP] Make `LocalConfig` enable to keep configuration history

takagi:
This PR makes `LocalConfig` enable to keep configuration history by adding `push` and `pop` methods to make it easier to use in object-oriented manner, especially in `setUp` / `tearDown` methods in unit test.

Note that thread local configuration is still `setattr` able so existing codes should not be affected.


takagi:
I close this because https://github.com/chainer/chainer/pull/5114#discussion_r203917883 do the same thing.

PR #5106, opened 2018-07-17
Add raw kernel function

kmaehashi:
This exposes Chainer interface for RawKernel (https://github.com/cupy/cupy/pull/1398).


beam2d:
The code looks good. Could you add it to the reference manual?

kmaehashi:
Thanks, I added it to docs.

beam2d:
LGTM

PR #5114, opened 2018-07-18
Support default dtype in `F.spatial_transformer_grid`

takagi:
This PR is a part of #4582, makes `F.spatial_transformer_gird` use the default dtype.

### DIFF
--- a/chainer/functions/array/spatial_transformer_grid.py
+++ b/chainer/functions/array/spatial_transformer_grid.py
@@ -24,7 +24,7 @@ class SpatialTransformerGrid(function.Function):
 
         theta_type = in_types[0]
         type_check.expect(
-            theta_type.dtype == self._dtype,
+            theta_type.dtype.kind == 'f',
             theta_type.ndim == 3,
             theta_type.shape[1] == 2,
             theta_type.shape[2] == 3,
--- a/chainer/functions/array/spatial_transformer_grid.py
+++ b/chainer/functions/array/spatial_transformer_grid.py
@@ -16,7 +16,6 @@ class SpatialTransformerGrid(function.Function):
 
     def __init__(self, output_shape):
         self.output_shape = output_shape
-        self._dtype = chainer.get_dtype()
 
     def check_type_forward(self, in_types):
         n_in = in_types.size()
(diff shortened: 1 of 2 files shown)


tests/chainer_tests/functions_tests/array_tests/test_spatial_transformer_grid.py
```diff
@@ -10,21 +10,45 @@
...
+      'forward_options': {},
+      'backward_options': {}},
+     ],
+    [{'use_cudnn': 'always'},
+     {'use_cudnn': 'never'}]
+))
 class TestSpatialTransformerGrid(unittest.TestCase):
 
     def setUp(self):
+        self._old_dtype = None
```
beam2d: How about using `using_config` by calling `__enter__` and `__exit__` directly?
```
    def setUp(self):
        self._config_user = chainer.using_config('dtype', self.dtype)
        self._config_user.__enter__()
        ...
    def tearDown(self):
        self._config_user.__exit__(None, None, None)
```

chainer/functions/array/spatial_transformer_grid.py
```diff
@@ -16,14 +16,15 @@ class SpatialTransformerGrid(function.Function):
 
     def __init__(self, output_shape):
         self.output_shape = output_shape
+        self._dtype = chainer.get_dtype()
```
beam2d: I overlooked it; I think it's better to accept `theta` of any dtype with kind `f` like many other functions do instead of restricting it to the current dtype.

tests/chainer_tests/functions_tests/array_tests/test_spatial_transformer_grid.py
beam2d: (I added another comment below; that has higher priority so please look that first!)

chainer/functions/array/spatial_transformer_grid.py
takagi: Fixed!

tests/chainer_tests/functions_tests/array_tests/test_spatial_transformer_grid.py
takagi: Fixed!

chainer/functions/array/spatial_transformer_grid.py
```diff
@@ -62,13 +63,13 @@ def _forward(self, inputs):
         xp = cuda.get_array_module(theta)
 
         ys, xs = xp.meshgrid(
-            xp.linspace(-1, 1, H, dtype=numpy.float32),
-            xp.linspace(-1, 1, W, dtype=numpy.float32), indexing='ij',
+            xp.linspace(-1, 1, H, dtype=self._dtype),
+            xp.linspace(-1, 1, W, dtype=self._dtype), indexing='ij',
```
beam2d: Use `theta.dtype` instead of `self._dtype`? (ditto for other parts, too.) We could remove `self._dtype`, too.

kmaehashi:
@takagi ping?

takagi:
Now I touch this.

takagi:
Jenkins, test this please.

takagi:
Jenkins, test this please.

okuta:
jenkins, test this please.

okuta:
LGTM!

PR #5118, opened 2018-07-19
Add isort settings in setup.cfg

ktns:
For ease of making PRs, I've added isort configuration which seems to generate import lines that conform to the convention in this repository.

If you don't like this, please close it silently.


kmaehashi:
How do you run `isort`? I couldn't get the output compatible with the current rule.

PR #5119, opened 2018-07-19
Return params and children as ordered by name in Link and Chain

### LINKED ISSUE #4547: About the unordered behavior of `params`, `namedparams` method.
For `Link` class, the implementation of `params` and `namedparams` methods depends on the `self._params` attribute. Similarly, For `Chain` class, the implementation of `params` and `namedparams` also depends on the `self._children` attribute. The problem is that both `_params` and `_children` is type `set`, which causes the unordered behavior when calling `params` or `namedparams` methods.

This problem can be inconvenient when we want to use an exact copy (deepcopy) of a model, expecially when we expect the `params` or `namedparams` are well aligned between the original model and the duplicate version (using copy.deepcopy).

Suppose we have a model composed of three convolutional layers: conv1, conv2, conv3.
```
for name, param in original_model.namedparams():
    print(name)
```
Above code gives us the results:
```
/conv1/W
/conv1/b
/conv2/W
/conv2/b
/conv3/W
/conv3/b
```

However, if we use `copied_model = copy.deepcopy(original_model)`, the results can happen to be:
```
/conv1/W
/conv1/b
/conv3/W
/conv3/b
/conv2/W
/conv2/b
```

For this problem, first, I'm wondering why a `set` is used for storing `_params` and `_children` attribute (maybe for avoiding the duplicate params/child links?). Second, is it possible to change the current code to make the `_params` and `_children` have a definite order? Finally, even if such change is not proper, I think it is necessary to add some note in the documentation to warn users about this kind of behavior.

At least for myself, I was expecting the `params` and `namedparams` to return a fixed order of parameters before reading details in the source code...

kmaehashi:
Fixes #4547.
The following methods now returns params and children in consistent order.

* `Link`: `params`, `namedparams`
* `Chain`: `params`, `namedparams`, `links`, `namedlinks`, `children`

### DIFF
--- a/chainer/link.py
+++ b/chainer/link.py
@@ -442,7 +442,7 @@ Assign a Parameter object directly to an attribute within a \
 
         """
         d = self.__dict__
-        for name in self._params:
+        for name in sorted(self._params):
             if include_uninit or d[name].data is not None:
                 yield d[name]
 
@@ -459,7 +459,7 @@ Assign a Parameter object directly to an attribute within a \
 
         """
         d = self.__dict__
-        for name in self._params:
+        for name in sorted(self._params):
             if include_uninit or d[name].data is not None:
                 yield '/' + name, d[name]
 
@@ -906,7 +906,7 @@ Assign a Link object directly to an attribute within a \
         for param in super(Chain, self).params(include_uninit):
             yield param
         d = self.__dict__
-        for name in self._children:
+        for name in sorted(self._children):
             for param in d[name].params(include_uninit):
                 yield param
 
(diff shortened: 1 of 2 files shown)


kmaehashi:
Jenkins, test this please.

kmaehashi:
Jenkins, test this please.

okuta:
Why don't you use `OrderedDict` (Python3.7+ `dict`)?

toslunar:
A reason I'd not use insertion orders is that the result of v1-style initialization, e.g. `chainer.Chain(child1=chainer.Link(), child2=chainer.Link())`, is unstable before Python 3.6 (PEP 468).

okuta:
OK, I understood.

toslunar:
Jenkins, test this please.

PR #5095, opened 2018-07-12
Move `backends.cuda.copyto` to `backends.copyto` and make it work with iDeep

kmaehashi:
This fixes #5009.
This is not to-be-backported as `backends.cuda.copyto` is introduced in #4997.

PR #5121, opened 2018-07-19
Add iDeep to backend docs

kmaehashi:
* Separate "CUDA Utilities" section and add iDeep.
  Context: in #5095 (move ``chainer.backends.cuda.copyto`` to ``chainer.backend.copyto``), I want a generic section to add backend-agnostic functions.
* Expose `intel64.is_ideep_available` as public API, as it is convenient because users can use this to decide whether to call ``to_intel64`` or not.


okuta:
jenkins, test this please.

okuta:
LGTM!

PR #5132, opened 2018-07-22
modified the iterator section of the tutorial document.

fiarabbit:
removed grammatical error from the section.
removed irrational description about repeating and shuffling.
clarify the difference between training and testing iterator (repeat=False, shuffle=False).


kmaehashi:
I'll backport this after #5170.

PR #5137, opened 2018-07-23
Fix L.Convolution2D error message

fiarabbit:
The ValueError message of L.Convolution2D is not splitted by a space.

Before:
the number of input channels must **_bedivisible_** by the number of groups
After
the number of input channels must **_be divisible_** by the number of groups


kmaehashi:
(for record; taken over to #5138)

PR #5138, opened 2018-07-23
Fix L.Convolution2D error message

fiarabbit:
The ValueError message of L.Convolution2D is not splitted by a space.

Before:
the number of input channels must bedivisible by the number of groups
After
the number of input channels must be divisible by the number of groups


okuta:
LGTM!

PR #5139, opened 2018-07-23
Remove doctest dependency on ideep4py

### LINKED ISSUE #5134: doctest requires ideep4py
doctest of `docs/source/tips.rst` fails without ideep4py installed.
We should either require it explicitly or fix the test.

niboshi:
Fixes #5134


kmaehashi:
It cannot be annotated as such as this is not doctest.
Could you check #5148?

niboshi:
OK, I close this.

PR #5125, opened 2018-07-20
Use default dtype in `F.n_step_rnn` and `F.n_step_birnn`

takagi:
Merge cupy/cupy#1471  for appropriate cuDNN RNN support, and #5143 first

This PR is a part of #4582, makes `F.n_step_rnn` support the default dtype.

PR #5143, opened 2018-07-25
Use data type rather than data type object for default dtype

takagi:
This PR fixes the default dtype to use data type, such as `numpy.float32`, rather than data type object (an instance of `numpy.dtype`) as the configuration reference expects.

### DIFF
--- a/chainer/__init__.py
+++ b/chainer/__init__.py
@@ -154,7 +154,7 @@ _chainer_dtype = os.environ.get('CHAINER_DTYPE', 'float32')
 if _chainer_dtype not in ('float16', 'float32', 'float64'):
     raise TypeError('incorrect dtype name in CHAINER_DTYPE: "{}". '
                     'Only float16/32/64 are allowed.'.format(_chainer_dtype))
-global_config.dtype = numpy.dtype(_chainer_dtype)
+global_config.dtype = numpy.dtype(_chainer_dtype).type
 
 
 def is_debug():


kmaehashi:
I think it's LGTM.
@beam2d Could you confirm?

kmaehashi:
Jenkins, test this please.

takagi:
Jenkins, test this please.

takagi:
I posted this change for the branches below, but perhaps comparing equality with a data type object and a data type rather than compairng identity, as `dtype == numpy.float32`, is enough?

https://github.com/chainer/chainer/pull/5125/files#diff-f79fb989aafdfa735d21eac69a25d020R56

kmaehashi:
After reading [numpy docs](https://docs.scipy.org/doc/numpy-1.13.0/reference/arrays.dtypes.html#specifying-and-constructing-data-types):

> Whenever a data-type is required in a NumPy function or method, either a dtype object or something that can be converted to one can be supplied.

It seems the most "formal" way of specifying dtype to function seems to use data type object; so I think it's better to return data type object (current implementation in master branch) from `get_dtype()`, instead of data type (this PR).

> I posted this change for the branches below, but perhaps comparing equality with a data type object and a data type rather than compairng identity, as dtype == numpy.float32, is enough?

Or `dtype.type is numpy.float32`?

beam2d:
Users may overwrite `config.dtype`. We should support both dtype objects and data types in that case. When the value of `config.dtype` may be any of data type specifiers (dtype objects, data types, str, ...), we anyway have to once convert it to dtype objects (by passing it to `np.dtype()`), so I think using dtype objects as normalized values is reasonable. `get_dtype` is a good place to normalize the value.

BTW, I also think it's great if we can normalize config values at assignments. We may think about that as a feature of our config system.

takagi:
Okay, then I make `chainer.get_dtype()` normalize the value and fix `_get_cudnn_data_type` to compare equality to data type.

kmaehashi:
Fixed via #5167

PR #5150, opened 2018-07-25
Implement eps_inside_sqrt option to RMSprop

### LINKED ISSUE #4754: Difference in the implementation of rmsprop  with Tensor Flow
There is a small difference between Chainer's implementation of rmsprop and the one in Tensor Flow which I thought worth pointing out. In [TF](https://github.com/tensorflow/tensorflow/blob/r1.8/tensorflow/python/training/rmsprop.py), the epsilon is inside the sqrt function while this is not the case for Chainer's [implementation](https://github.com/chainer/chainer/blob/v4.0.0/chainer/optimizers/rmsprop.py). This does not make much difference when you pick very small (compared to mean grad**2) values for epsilon but it does when epsilon is large (i.e eps=1.0) as it is used in training Inception [V3](https://arxiv.org/pdf/1512.00567.pdf) and [V4](https://arxiv.org/pdf/1602.07261v1.pdf).

kmaehashi:
Fix #4754.

### DIFF
--- a/chainer/optimizers/rmsprop.py
+++ b/chainer/optimizers/rmsprop.py
@@ -8,6 +8,7 @@ _default_hyperparam = optimizer.Hyperparameter()
 _default_hyperparam.lr = 0.01
 _default_hyperparam.alpha = 0.99
 _default_hyperparam.eps = 1e-8
+_default_hyperparam.eps_inside_sqrt = False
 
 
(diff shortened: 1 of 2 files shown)


toslunar:
Jenkins, test this please.

kmaehashi:
Jenkins, test this please.

PR #4153, opened 2017-12-26
Fix to make it consistent with numpy.split and cupy.split

ken-nakanishi:
Fix to make it consistent with numpy.split and cupy.split

before
```
>>> import numpy as np
>>> import cupy as cp
>>> import chainer.functions as F
>>> np.split(np.arange(4), indices_or_sections=[0,1,1], axis=0)
[array([], dtype=int64), array([0]), array([], dtype=int64), array([1, 2, 3])]
>>> cp.split(cp.arange(4), indices_or_sections=[0,1,1], axis=0)
[array([], dtype=int64), array([0]), array([], dtype=int64), array([1, 2, 3])]
>>> F.split_axis(np.arange(4), indices_or_sections=[0,1,1], axis=0)
(variable([]), variable([0]), variable([]), variable([1, 2, 3]))
>>> np.split(np.arange(4), indices_or_sections=[0,1,1,4], axis=0)
[array([], dtype=int64), array([0]), array([], dtype=int64), array([1, 2, 3]), array([], dtype=int64)]
>>> cp.split(cp.arange(4), indices_or_sections=[0,1,1,4], axis=0)
[array([], dtype=int64), array([0]), array([], dtype=int64), array([1, 2, 3]), array([], dtype=int64)]
>>> F.split_axis(np.arange(4), indices_or_sections=[0,1,1,4], axis=0)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/***/chainer/functions/array/split_axis.py", line 92, in split_axis
    res = SplitAxis(indices_or_sections, axis).apply((x,))
  File "/***/chainer/function_node.py", line 230, in apply
    self._check_data_type_forward(in_data)
  File "/***/chainer/function_node.py", line 298, in _check_data_type_forward
    self.check_type_forward(in_type)
  File "/***/chainer/functions/array/split_axis.py", line 37, in check_type_forward
    type_check.expect(in_types[0].shape[self.axis] > max_index)
  File "/***/chainer/utils/type_check.py", line 524, in expect
    expr.expect()
  File "/***/chainer/utils/type_check.py", line 482, in expect
    '{0} {1} {2}'.format(left, self.inv, right))
chainer.utils.type_check.InvalidType:
Invalid operation is performed in: SplitAxis (Forward)

Expect: in_types[0].shape[0] > max_index
Actual: 4 <= 4
```

after
```
>>> import numpy as np
>>> import chainer.functions as F
[...]

PR #5157, opened 2018-07-26
Support old numpy in `F.split_axis`

toslunar:
This PR fixes #4153.

Outputs of `np.split` differ among numpy versions if there's an output with size 0.  It's possible that  intensionally such case had been unsupported.  #4153 added the support, but didn't work with numpy<1.11.

### DIFF
--- a/chainer/functions/array/split_axis.py
+++ b/chainer/functions/array/split_axis.py
@@ -188,11 +188,6 @@ def split_axis(x, indices_or_sections, axis, force_tuple=True):
         When ``force_tuple`` is ``True``, returned value is always a tuple
         regardless of the number of outputs.
 
-    .. note::
-        This function raises :class:`ValueError` if at least
-        one of the outputs is split to zero-size
-        (i.e. ``axis``-th value of its shape is zero).
-
     """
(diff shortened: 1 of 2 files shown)


chainer/functions/array/split_axis.py
```diff
@@ -10,6 +10,23 @@
 from chainer.utils import type_check
 
 
+def _split_and_fix_shape(xp, x, indices_or_sections, axis):
+    ret = xp.split(x, indices_or_sections, axis)
+    if all(r.ndim == x.ndim for r in ret):
```
beam2d: How about returning quickly when numpy>=1.11?

okuta:
jenkins, test this please.

beam2d:
GPU tests do not pass in my environment. Could you check that? (It looks cupy does not support split for zero-sized arrays.)

toslunar:
@beam2d `cupy.split` is fixed in cupy/cupy#1459, after v5.0.0b3.

beam2d:
Oh, I missed that. OK, I'll try that.

okuta:
LGTM!

PR #5159, opened 2018-07-26
Override `Link.forward` in MNIST model parallel example

niboshi:
Extracted from #4730.


okuta:
jenkins, test this please.

PR #5163, opened 2018-07-29
Include platform information to `chainer.print_runtime_info()`

himkt:
Inspired by [SpaCy](https://github.com/explosion/spaCy/blob/master/spacy/cli/info.py).

I add a script in `/tools`.
This script returns environmental information which are needed to create new issue.
(currently it does not support CuDNN... I will really appreciate any feedback)

### usage
`python -m chainer.tools.info` 

### returns
In my laptop (macOS), I get...

```
- Chainer version:  5.0.0b3
- CuPy version:  not installed
- OS/Platform:  Darwin-17.7.0-x86_64-i386-64bit
- CUDA version:  not installed
```

And in my server (Ubuntu), I get

```
- Chainer version:  4.1.0
- CuPy version:  4.3.0
- OS/Platform:  Linux-4.4.0-112-generic-x86_64-with-debian-stretch-sid
- CUDA version:  8.0, V8.0.61
```


kmaehashi:
Thanks for contribution!
You can use `chainer.print_runtime_info()` for this purpose.
https://docs.chainer.org/en/latest/performance.html

(Maybe we should add it to our issue template...)

kmaehashi:
Adding `platform.platform()` to `print_runtime_info` seems nice.

himkt:
Thanks for sharing and I'm sorry that I couldn't find the method :bow:
It seems to have sufficient features to fill the issue template of environmental information.

Because I remain this PR to be opened because I want to get the runtime information by CLI.
But I also think it may not be necessary because we can easily get using interactive shell.
So please close this PR if dev-teams think modifying the issue template is enough for helping users to create issues.

(Currently, this PR enables users to get runtime information by `python -m chainer.tools.runtime_info`)

toslunar:
I'd rather not introduce `chainer.tools` namespace just for making an alias of
```
python -c 'import chainer; chainer.print_runtime_info()'
```

himkt:
@toslunar thanks, I updated!

himkt:
Ah sorry, I misunderstood.
Updated PR: does this change fix what you point out?

chainer/_runtime_info.py
```diff
@@ -16,13 +17,15 @@ class _RuntimeInfo(object):
...
         self.numpy_version = numpy.__version__
+        self.platform_version = platform.platform()
         if cuda.available:
             self.cuda_info = cuda.cupyx.get_runtime_info()
         else:
             self.cuda_info = None
 
     def __str__(self):
         s = six.StringIO()
+        s.write('''OS/Platform: {}\n'''.format(self.platform_version))
```
kmaehashi: How about `Platform` instead of `OS/Platform`?

himkt:
Thanks for the feedback, I updated :pray:

kmaehashi:
LGTM!

PR #5171, opened 2018-07-31
Add "copy and set item"

toslunar:


okuta:
@bkvogel @beam2d Is `copied_set_item` ok?
Another ideas are `copied_setitem`  `setitem_with_copy`.

toslunar:
How about adding `set_item(x, slices, rhs, *, inplace=True)` and supporting only `inplace=False` in this PR.

okuta:
@beam2d How about this design?

okuta:
jenkins, test this please.

okuta:
@beam2d Please check API design.

PR #5172, opened 2018-08-01
[WIP] Fix DeprecationWarnings in Python 3.7

### LINKED ISSUE #5097: Using Sequence from collections is deprecated in Python 3.7
It should be `collections.abc.Sequence`.  However `collections.abc` isn't there in Python 2.

kmaehashi:
This fixes #5097.


kmaehashi:
I locally confirmed tests pass. Note that https://github.com/cupy/cupy/pull/1515 is also needed to pass test on py37.

PR #5179, opened 2018-08-06
Fix backward computation in F.forget

ir5:
Current `F.forget` doesn't do backward computation appropriately. If `func` includes some model parameters, their gradient parameters may not be set.

Example:
```python
import chainer.functions as F
import chainer.links as L
import numpy as np

class MLP(chainer.Chain):
    def __init__(self):
        super(MLP, self).__init__()
        with self.init_scope():
            self.lin = L.Linear(None, 10)

    def forward(self, x):
        return F.forget(self.lin, x)

def main():
    model = MLP()
    model.cleargrads()
    x = np.random.uniform(-1, 1, (64, 768)).astype(np.float32)
    x = chainer.Variable(x, requires_grad=True)

    y = F.sum(model(x))
    y.backward()
    print(model.lin.W.grad)  # => None
    print(model.lin.b.grad)  # => None

if __name__ == '__main__':
    main()
```

This issue exists because current backward implementation in `F.forget` calls `chainer.grad` function. This PR fixes the issue by replacing the `chainer.grad` with `*.backward()`.

### DIFF
--- a/chainer/functions/util/forget.py
+++ b/chainer/functions/util/forget.py
@@ -54,9 +54,10 @@ class Forget(function_node.FunctionNode):
         with function.force_backprop_mode():
             outs = _call_func(self.func, dummy_inputs)
             assert len(outs) == len(grad_outputs)
-            outs = chainer.functions.identity(*outs)
-        if not isinstance(outs, tuple):
-            outs = (outs,)
+            if len(outs) > 1:
+                # Avoid doing backward multiple times when `outs` is a tuple
+                outs = chainer.functions.identity(*outs)
+
         for out, grad_output in zip(outs, grad_outputs):
             out.grad_var = grad_output
         outs[0].backward()
--- a/chainer/functions/util/forget.py
+++ b/chainer/functions/util/forget.py
@@ -53,10 +53,14 @@ class Forget(function_node.FunctionNode):
 
         with function.force_backprop_mode():
             outs = _call_func(self.func, dummy_inputs)
-        assert len(outs) == len(grad_outputs)
+            assert len(outs) == len(grad_outputs)
+            outs = chainer.functions.identity(*outs)
+        if not isinstance(outs, tuple):
+            outs = (outs,)
         for out, grad_output in zip(outs, grad_outputs):
             out.grad_var = grad_output
-            out.backward()
+        outs[0].backward()
+
         return tuple([inp.grad_var for inp in dummy_inputs])
 
 
--- a/chainer/functions/util/forget.py
+++ b/chainer/functions/util/forget.py
@@ -1,3 +1,4 @@
+import chainer
 from chainer import function
 from chainer import function_node
 from chainer import variable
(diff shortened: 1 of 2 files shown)


ir5:
Note: To keep implementation simple, this PR will discard double-backpropagation support of `F.forget`.

chainer/functions/util/forget.py
```diff
@@ -43,12 +42,16 @@ def forward(self, inputs):
...
-            outs = _call_func(self.func, inputs)
-        # Return gradients that are further backproable
-        return chainer.grad(
-            outs, inputs, grad_outputs=grad_outputs,
-            enable_double_backprop=True)
+            outs = _call_func(self.func, dummy_inputs)
+        assert len(outs) == len(grad_outputs)
+        for out, grad_output in zip(outs, grad_outputs):
+            out.grad_var = grad_output
+            out.backward()
```
beam2d: When there are N>1 outputs, it looks backward is called N times, which is redundant when these outputs share a part of the computational graph. Could you reduce it to only one backward call? For example, all the paths are covered using `F.identity`:

```py3
outs = F.identity(*outs)
for out, grad_output in zip(outs, grad_outputs):
    out.grad_var = grad_output
outs[0].backward()  # other outputs are taken into account through the backward of F.identity
```

chainer/functions/util/forget.py
```diff
@@ -105,6 +108,10 @@ def forget(func, *xs):
         This conversion takes place to ensure that this function is included
         in the computational graph to enable backward computations.
 
+    .. note::
+
+        ``F.forget`` does not support double backpropagation.
```
beam2d: Can we raise an error when users try to use `forget` with double backprop, e.g. by checking `not chainer.config.enable_backprop`?

ir5: I added code that raises an error when `chainer.config.enable_backprop == True`. I added a test as well.

chainer/functions/util/forget.py
ir5: Thank you for the suggestion. Unfortunately, the code you suggested didn't work well. I couldn't fix it. One error is that the test `TestForgetGrad` didn't pass.

ir5: Oh, I noticed that we need to call `F.identity` in `with function.force_backprop_mode():` block. In that case, your code works well.

beam2d: Ahhh, sorry for my incomplete code.

chainer/functions/util/forget.py
```diff
@@ -42,13 +42,26 @@ def forward(self, inputs):
...
-            outs = _call_func(self.func, inputs)
-        # Return gradients that are further backproable
-        return chainer.grad(
-            outs, inputs, grad_outputs=grad_outputs,
-            enable_double_backprop=True)
+            outs = _call_func(self.func, dummy_inputs)
+            assert len(outs) == len(grad_outputs)
+            outs = chainer.functions.identity(*outs)
+        if not isinstance(outs, tuple):
+            outs = (outs,)
```
beam2d: How about writing it as follows to avoid re-wrapping by tuple:
```py
assert len(outs) == len(grad_outputs)
if len(outs) > 1:
    outs = chainer.functions.identity(*outs)
```

beam2d:
Could you add some tests that use forget with Link?

ir5:
Thank you for the suggestion. I added tests with Links.

beam2d:
Thank you!

PR #5182, opened 2018-08-06
Fix doctests of `open_pickle_dataset`

toslunar:

PR #5184, opened 2018-08-06
Fix doctest for #4581

beam2d:
#4581 introduced doc test regression. This PR fixes it.


beam2d:
Duplicating with #5182.

Issue #5183, opened 2018-08-06
Normal distribution test error


niboshi:
Version: v5.0.0b3 master 313083e2f (with unrelated local changes)

I happen to encountered this error once. Several (~10) retries did not replicate the error.

```
self = <chainer.testing.parameterized.TestNormal_param_0 testMethod=test_log_survival_cpu>

    @skip_not_in_test_target('log_survival')
    def test_log_survival_cpu(self):
>       self.check_log_survival(False)

chainer/testing/distribution_test.py:217:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
chainer/testing/distribution_test.py:213: in check_log_survival
    array.assert_allclose(log_survival1, log_survival2)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

x = array([[[[ -2.6995385 ,  -4.594169  ,  -1.725811  ],
         [        -inf,  -1.1890926 ,  -0.3637382 ]],

        [[...    [[ -8.725308  ,  -0.10234644,  -1.169075  ],
         [ -7.818679  ,  -0.30798176,  -0.5043421 ]]]], dtype=float32)
y = array([[[[ -2.69953811,  -4.59416883,  -1.72581108],
         [-19.45005923,  -1.18909271,  -0.36373823]],

        [[...628318]],

        [[ -8.72533196,  -0.10234649,  -1.16907499],
         [ -7.81867423,  -0.30798173,  -0.50434207]]]])
atol = 1e-05, rtol = 0.0001, verbose = True

[...]

PR #5185, opened 2018-08-06
Avoid cancellation in Normal

gwtnb:
Use `erfc` to avoid cancellation. This may relate to #5183.


gwtnb:
Added another fix.

gwtnb:
`1 + erf(x) = 1 - erf(-x) = erfc(-x)`

toslunar:
Jenkins, test this please.

PR #5188, opened 2018-08-06
Fix `FunctionHook` documentation

niboshi:
Fixed some incorrect/missing information.


chainer/function_hook.py
```diff
@@ -76,15 +84,24 @@ class FunctionHook(object):
        as a thread local object. So, function hooks registered
        are different depending on threads.
 
+    If the hook is registered in this way, :meth:`~chainer.FunctionHook.added`
```
beam2d: It takes time for me to notice that this sentence is a continuation of the last paragraph above the long example/note sections. How about moving it to the end of the last paragraph?

chainer/function_hook.py
```diff
@@ -76,15 +84,24 @@ class FunctionHook(object):
...
     The other one is to register directly to
     :class:`~chainer.FunctionNode` object with
-    :meth:`~chainer.Function.add_hook` method.
+    :meth:`~chainer.FunctionNode.add_hook` method.
     Function hooks registered in this way can be removed by
-    :meth:`~chainer.Function.delete_hook` method.
+    :meth:`~chainer.FunctionNode.delete_hook` method.
     Contrary to former registration method, function hooks are registered
     only to the function which :meth:`~chainer.FunctionNode.add_hook`
     is called.
```
beam2d: It's also not related to this change, but I think it's a good chance to fix some expressions:
- The other one is to `register directly` to... -> `register it directly`
- Contrary to `former registration method`, ... -> `the former registration method`
- only to the function which add_hook is called. -> only to the function whose add_hook method is called.

chainer/function_hook.py
```diff
@@ -76,15 +84,24 @@ class FunctionHook(object):
        as a thread local object. So, function hooks registered
        are different depending on threads.
 
+    If the hook is registered in this way, :meth:`~chainer.FunctionHook.added`
+    and :meth:`~chainer.FunctionHook.deleted` are given ``None`` as the
+    ``function`` argument.
```
beam2d: How about rephrasing (matching with the below one):
None is passed to the function argument of added and deleted.

chainer/function_hook.py
```diff
@@ -76,15 +84,24 @@ class FunctionHook(object):
...
+    :meth:`~chainer.FunctionNode.add_hook` method.
     Function hooks registered in this way can be removed by
-    :meth:`~chainer.Function.delete_hook` method.
+    :meth:`~chainer.FunctionNode.delete_hook` method.
     Contrary to former registration method, function hooks are registered
     only to the function which :meth:`~chainer.FunctionNode.add_hook`
     is called.
 
+    If the hook is registered in this way, the :class:`~chainer.FunctionNode`
+    instance is given as the ``function`` argument of
```
beam2d: Suggestion (similar to the above one): given as -> passed to

chainer/function_hook.py
```diff
@@ -118,7 +136,8 @@ def deleted(self, function=None):
 
         Args:
             function(~chainer.FunctionNode): Function object to which
-                the function hook is deleted.
+                the function hook is deleted. ``None`` if the function hook
+                had been registered globally.
```
beam2d: has been?

chainer/function_hook.py
```diff
@@ -118,7 +136,8 @@ def deleted(self, function=None):
 
         Args:
             function(~chainer.FunctionNode): Function object to which
```
beam2d: (again, suggestion outside of this change)
to which -> from which

chainer/function_hook.py
niboshi: I'm not confident about my expression either, but the hook is unregistered (the state of "being registered" is negative) at the moment of this callback.

niboshi: According to @hvy, "had been" is okay but "was" is more natural expression. I fixed as such.

chainer/function_hook.py
niboshi: Fixed using "passed as".

chainer/function_hook.py
niboshi: Fixed using "passed as".

chainer/function_hook.py
niboshi: Fixed.

chainer/function_hook.py
niboshi: Fixed.

chainer/function_hook.py
```diff
@@ -76,14 +84,24 @@ class FunctionHook(object):
        as a thread local object. So, function hooks registered
        are different depending on threads.
 
-    The other one is to register directly to
+    The other one is to register it directly to
     :class:`~chainer.FunctionNode` object with
```
hvy: Nitpick but 
with -> with the

chainer/function_hook.py
```diff
@@ -76,14 +84,24 @@ class FunctionHook(object):
...
-    :meth:`~chainer.Function.add_hook` method.
+    :meth:`~chainer.FunctionNode.add_hook` method.
     Function hooks registered in this way can be removed by
```
hvy: Nitpick, actually not directly related to this PR...
registered in this way ... by -> registered this way ... by the

chainer/function_hook.py
```diff
@@ -76,14 +84,24 @@ class FunctionHook(object):
...
-    :meth:`~chainer.Function.delete_hook` method.
-    Contrary to former registration method, function hooks are registered
-    only to the function which :meth:`~chainer.FunctionNode.add_hook`
-    is called.
+    :meth:`~chainer.FunctionNode.delete_hook` method.
+    Contrary to the former registration method, function hooks are registered
+    only to the function whose :meth:`~chainer.FunctionNode.add_hook`
+    method is called.
+
+    If the hook is registered globally using ``with`` statement, ``None`` is
```
hvy: I'd place an "a" before the "``with``". What do you think?

chainer/function_hook.py
```diff
@@ -59,9 +66,10 @@ class FunctionHook(object):
         >>> with chainer.function_hooks.TimerHook() as m:
         ...   _ = model1(x)
         ...   y = model2(x)
```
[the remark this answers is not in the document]
beam2d: Yes, if you prefix the blank line with `>>>` or `...` like follows.

```py
"""
        >>> class Model(chainer.Chain):
        ...   def __init__(self):
        ...     super(Model, self).__init__()
        ...     with self.init_scope():
        ...       self.l = L.Linear(10, 10)
        ...
        ...   def __call__(self, x1):
        ...     return F.exp(self.l(x1))
        ...
        >>> model1 = Model()
        >>> model2 = Model()
        >>>
        >>> x = chainer.Variable(np.zeros((1, 10), np.float32))
        >>>
        >>> with chainer.function_hooks.TimerHook() as m:
        ...   _ = model1(x)
        ...   y = model2(x)
        ...
        >>> model3 = Model()
        >>> z = model3(y)
        >>>
        >>> print('Total time : {}'.format(m.total_time()))
        ... # doctest:+ELLIPSIS
        Total time : ...
"""
```

By the way, is it necessary to write this example in doctest style?

chainer/function_hook.py
beam2d: Ah, yes you are right. I was confused with "pass to function as an argument".

toslunar:
Jenkins, test this please.

Issue #5015, opened 2018-06-26
`F.logsumexp` fails on 0-dim input


toslunar:
Use `chainer.utils.force_array`.
```
>>> chainer.functions.logsumexp(chainer.Variable(np.array(3, 'f')), axis=None)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/tos/GitHub/chainer/chainer/functions/math/logsumexp.py", line 91, in logsumexp
    return LogSumExp(axis).apply((x,))[0]
  File "/Users/tos/GitHub/chainer/chainer/function_node.py", line 257, in apply
    outputs = self.forward(in_data)
  File "/Users/tos/GitHub/chainer/chainer/functions/math/logsumexp.py", line 47, in forward
    xp.exp(y, out=y)
TypeError: return arrays must be of ArrayType
```

toslunar:
Fixed by #5190.

PR #5190, opened 2018-08-06
Support 0-dim input in `F.logsumexp`

cadenacchi:
Related to #5015 

* Used `chainer.utils.force_array`
* Added shape parameter for tests
* Skiped some tests which is specified axis other than None if input shape is 0-dim
  * referred  to `functions_tests.math_tests.test_einsum._skip_if`
* But does not skip following tests
  * `test_invalid_axis_type`
  * `test_invalid_axis_type_in_tuple`

### DIFF
--- a/chainer/functions/math/logsumexp.py
+++ b/chainer/functions/math/logsumexp.py
@@ -1,6 +1,7 @@
 import chainer
 from chainer.backends import cuda
 from chainer import function_node
+from chainer import utils
 from chainer.utils import type_check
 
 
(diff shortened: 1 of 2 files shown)


chainer/functions/math/logsumexp.py
```diff
@@ -1,6 +1,7 @@
 import chainer
 from chainer.backends import cuda
 from chainer import function_node
+from chainer.utils import force_array
```
toslunar: Could you import the module rather than importing the function?
```
from chainer import utils
...
        y = utils.force_array(x - m)
```

cadenacchi: Ok, I will fix!

cadenacchi:
Fixed.

toslunar:
Jenkins, test this please.

PR #5192, opened 2018-08-07
Document attributes before methods in class documentation

niboshi:
I think it's more convenient.
(I'm not sure why I've written in the oppsite way in the first place)


kmaehashi:
I think this convention is coming from the default autosummary template.
https://raw.githubusercontent.com/sphinx-doc/sphinx/master/sphinx/ext/autosummary/templates/autosummary/class.rst

(I don't have a strong preference for this.)

PR #5193, opened 2018-08-07
Add .pytest_cache/ to .gitignore

gwtnb:
Add `.pytest_cache/` to `.gitignore`.


kmaehashi:
LGTM, makes sense.
I think it can further improved by rewriting based on https://github.com/github/gitignore/blob/master/Python.gitignore, though.

PR #5202, opened 2018-08-09
Implement erfcinv

gwtnb:
Implement `erfcinv`

### DIFF
--- a/chainer/functions/__init__.py
+++ b/chainer/functions/__init__.py
@@ -126,6 +126,7 @@ from chainer.functions.math.digamma import digamma  # NOQA
 from chainer.functions.math.einsum import einsum  # NOQA
 from chainer.functions.math.erf import erf  # NOQA
 from chainer.functions.math.erfc import erfc  # NOQA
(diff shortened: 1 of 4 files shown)


toslunar:
How about implementing it in `erfinv.py` to avoid repeating `available_cpu` and `BACKWARDC`

gwtnb:
If `erfinv` is used, the error message contains `erfinv` (not `erfcinv`). It may be confusing.

toslunar:
Do you mean that the traceback will contain `erfinv`?  I think the error message itself will contain `erfinvc`.

gwtnb:
I mean this message. https://github.com/chainer/chainer/blob/737625d1aac59c05c7241ec9d038ea2ec218c039/chainer/functions/math/erfinv.py#L32

toslunar:
My suggestion is https://github.com/toslunar/chainer/commit/8c6e31e04cb6865176047146fdef21de64e57c71

toslunar:
I rethought it.  OK to merge.

toslunar:
Jenkins, test this please.

Issue #5082, opened 2018-07-11
Numpy LSTM very slow in chainer 4


gwichern:
I recently upgraded from v3.4, and the following code runs about 20x slower in v4.2.

```
import numpy as np
import chainer.links as L

n_inputs = 120
n_steps = 500
n_out = 600
dropout = 0.0
n_layers = 3

x = [np.random.rand(n_steps, n_inputs).astype(np.float32)]
net = L.NStepLSTM(n_layers, n_inputs, n_out, dropout)
h, c, y = net(None, None, x)
```
With a quick profile it seems like the stacking of the weight matrices for the different LSTM gates is now happening in the inner time-step loop, instead of once per batch as in v3.4.

https://github.com/chainer/chainer/blob/master/chainer/functions/connection/n_step_lstm.py#L450-L457

niboshi:
Thank you for reporting.
It seems at least a part of slowdown is caused by ideep4py.
How about updating to ideep4py>=2.0.0?

### Some more detail
The problem started in #4276.
After this PR, this simple code takes ~100 times slower.
```
import chainer
import numpy as np
X = np.ones((1, 50), 'f')
W = np.ones((200, 50), 'f')
X.dot(W.T)
```
Note that in this code neither ideep4py nor chainer is actually used.

Conditions of slowdown:
* ideep4py 1.0.3 or 1.0.4 is installed
* `import chainer` - This is the mysterious part. If ideep4py is directly imported as `import ideep4py` without chainer, performance degradation is not observed.

kmaehashi:
It seems iDeep v1.x overwrites the default thread number to the number of cores upon initialization.
https://github.com/intel/ideep/blob/v1.0.4/python/ideep4py/common/cpu_info.cc#L408-L411

The code does not exist in iDeep v2.x.

Setting environment variable `export OMP_NUM_THREADS=1` seems to resolve the issue on my environment.

> How about updating to ideep4py>=2.0.0, if it's installed?

Please do not do this, as iDeep4py v2.0 is incompatible with the current Chainer.
It will be supported (possibly) in the next release.

toslunar:
Chainer v5 is stable.  Feel free to open a new issue if it's slow in v5+.

PR #5231, opened 2018-08-15
Unsupport ideep4py 2+ in Chainer v4

niboshi:
https://github.com/chainer/chainer/issues/5082#issuecomment-413084915

It's better to detect and warn ideep4py 2+ in Chainer v4.


niboshi:
Closed because ideep4py 2+ will be supported in the next minor v4 release (v4.4.0).

PR #5235, opened 2018-08-16
Support all float dtypes in `F.batch_l2_norm_squared`

takagi:
This PR follows #4582 to support all float dtypes in `F.batch_l2_norm_squared`.


takagi:
Jenkins, test this please.

PR #5239, opened 2018-08-16
Add log_ndtr

gwtnb:
Add `log_ndtr`. This PR depends on #5237.

PR #5237, opened 2018-08-16
Add `F.ndtr`

gwtnb:
Add cumulative distribution function of normal distribution.

PR #5240, opened 2018-08-16
Use ndtr and log_ndtr in normal distribution

gwtnb:
`log_ndtr` is used to avoid overflow and underflow. `ndtr` is for consistency.


gwtnb:
This PR depends on #5237 and #5239.

toslunar:
Jenkins, test this please.

PR #5249, opened 2018-08-20
Improve type_check.argname for optional arguments

niboshi:
This PR improves `type_check.argname()` (#4680) interface for handling optional arguments nicely.

### DIFF
--- a/chainer/utils/type_check.py
+++ b/chainer/utils/type_check.py
@@ -555,6 +555,9 @@ class _MissingArgument(object):
     def __div__(self, other):
         return self
 
+    def __truediv__(self, other):
+        return self
+
 
 _missing_argument = _MissingArgument()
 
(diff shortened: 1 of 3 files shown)


takagi:
I finished the implementation of this PR.

The function name `argname()` may be better to change, for example `destructure_types()`, before v5 release to reflect the feature introduced here.

takagi:
Jenkins, test this please.

chainer/utils/type_check.py
```diff
@@ -504,26 +504,103 @@ def __init__(self, expect, actual, msg=None):
...
         if isinstance(in_type, Variable):
             in_type.name = name
 
+    # Assign the names for the optional arguments.
+    if actual_len > min_len:
+        for in_type, name in zip(in_types[len(names):], optional_names):
+            if isinstance(in_type, Variable):
+                in_type.name = name
+
+    # For missing optional arguments, fill with _missing_argument which
```
okuta: This function is called many times. So, This code is only for test.
Can you remove it?

PR #5254, opened 2018-08-21
Use ndtri in normal distribution

gwtnb:
Use `ndtri` in normal distribution.


toslunar:
Jenkins, test this please.

PR #5198, opened 2018-08-08
add roi_max_align_2d

knorth55:
~~Merge after #5070~~
~~[WIP] I haven't done GPU test yet.~~
Update: GPU tested.

add `roi_max_align_2d`.
#5070 is `average` pooling of `roi_align`, and this PR add `max` pooling of `roi_align`.
mainly `max` is for detection, and `average` for instance segmentation.

BTW, `roi_pooling_2d` is `max` pooling, but default `roi_align_2d` should be `average`?

cc. @wkentaro

PR #5259, opened 2018-08-21
Rename roi_align_2d to roi_average_align_2d

kmaehashi:
This is a tentative fix (should be removed in #5198) to keep compatibility between betas.


kmaehashi:
@toslunar Can you check?

PR #5269, opened 2018-08-24
Use `with cuda_device`

toslunar:
This fixes the issue that `chainer.grad` may change the current device.

I'd like to use `with device` instead of `device.use()`.  I left `dev.use()` in the worker of `MultiprocessParallelUpdater` because it's clear the process does not do other tasks.


toslunar:
Jenkins, test this please.

okuta:
LGTM!

PR #5270, opened 2018-08-24
Add `contains_nan`

toslunar:
The NaN check is not trivial so that Chainer could offer such a function.

The PR may supersede #4737.


okuta:
Do you want to this as public method?

toslunar:
OK, it's better to be private at the moment.

okuta:
jenkins, test this please.

okuta:
LGTM!

toslunar:
`_contains_nan` is private.

PR #5268, opened 2018-08-24
[backport] Include platform information to `chainer.print_runtime_info()`

kmaehashi:
Backport of #5163

PR #5271, opened 2018-08-24
Fix import order

toslunar:


chainer/_runtime_info.py
```diff
@@ -1,8 +1,8 @@
 import sys
 
 import numpy
-import six
 import platform
+import six
```
kmaehashi: Thanks for the catch, but `import platform` should be above `import sys`.

kmaehashi:
LGTM!

kmaehashi:
This will be backported with #5268

PR #5228, opened 2018-08-14
NaN check in `chainer.grad` in debug mode

toslunar:
Move the NaN check from `Variable.backward_main` to `_backprop_utils.backprop_step`.

Merge #3983 first.

PR #5274, opened 2018-08-27
Backprop from moved variables

### LINKED ISSUE #5266: Backprop sometimes needs additional memory
Revisit https://github.com/chainer/chainer/pull/4301#issuecomment-392438444

# Status

I measured how much memory is used in back-propagation.  The peak consumptions of gpu memory (max of `cuda.memory_pool.used_bytes`) are

|model|initialize|forward|backward|fwd/bwd|
|-|-|-|-|-|
|ResNet50|`0127614464`|`3626270720`|`4164088320`|0.8708|
|U-Net|`0081935360`|`0785561088`|`1011115520`|0.7769|

- ResNet50
  - code: `chainer.links.model.vision.resnet.ResNet50Layers`
  - input: `(32, 3, 256, 256), float32`
- U-Net
  - code: `UNET` in https://github.com/pfnet/PaintsChainer/blob/master/cgi-bin/paint_x2_unet/unet.py
  - input: `(16, 4, 128, 128), float32`

# Feature request

Such memory would be reduced by a feature that allows `del loss` before executing `loss.backward()`.  I'd like to implement it.

toslunar:
Close #5266.

The idea of this PR is to move the "ownership" of a `Variable` `y` to its method `y.backward`.
Because Python does not have this kind of move semantics, I propose:
```python
with chainer.variable.delay_backward():
    y.backward()
    del y
```


toslunar:
In the settings of the experiments in https://github.com/chainer/chainer/issues/5266#issue-353347476, peak memory consumption is

before

|model|initialize|forward|backward|fwd/bwd|
|-|-|-|-|-|
|ResNet50|`0127614464`|`3626270720`|`4164088320`|0.8708|
|U-Net|`0081935360`|`0785561088`|`1011115520`|0.7769|

after

|model|initialize|forward|backward|fwd/bwd|
|-|-|-|-|-|
|ResNet50|`0127614464`|`3626270720`|`3667630592`|0.9887|
|U-Net|`0081935360`|`0785561088`|`0852677120`|0.9213|

toslunar:
I need an idiom to safely delete a loop variable.   Is the following OK?
```python
for x in xs:
    pass
x = None
del x
```
Note that `x` will not be defined by the loop if `xs` is empty.

It was so surprising that `flake8` on
```python
def f():
    if True:
        x = None
        del x
        print([1 for x in []])
```
generates the warning
```
a.py:5:22: F812 list comprehension redefines 'x' from line 3
```

toslunar:
Merge #5228 first.  (Resolving the conflicts with `master` using the PR was easier.)

beam2d:
I think, instead of letting user write `with` and `del` in such a way, it's better to provide an option to `backward` to let the method unchain the loss variable from the graph before backprop, like `loss.backward(unchain_self=True)` or more aggressively `loss.backward(unchain=True)` (which may unchain every node encountered during backprop). What do you think? Also, is this feature essentially same with #4301?

toslunar:
```
h = f(x)
y0 = g0(h)
y1 = g1(h)
del h
with chainer.variable.delay_backward():
    y0.backward()
    del y0
y1.backward()
```
correctly computes `x.grad`, while
```
y0.backward(unchain_self=True)
y1.backward()
```
will not.

toslunar:
As I wrote in https://github.com/chainer/chainer/pull/4301#issuecomment-389464326, just writing
```
cont = y.backward(execute=False)
del y
cont()
```
is OK.  This is better if it's weird to use with-statement.  I think `del y` is quite natural.

beam2d:
I understood the difference between move and unchain. How about adding a method to move the value, say `Variable.move()`? I'm imagining a method that create a fresh temporary Variable, swapping its dict with self's, and returns the temporary that now looks like the original one. We should also take care of the weak reference from the node. The implementation would look like this.

```py
def move(self):
    new = Variable(None)
    d = self.__dict__
    self.__dict__ = new.__dict__
    new.__dict__ = d

    new_node = new._node
    self_node = self._node
    r = new_node._variable
    new_node._variable = self_node._variable
    self_node._variable = r
    return new
```

It could be used like `y.move().backward()`. It would be also fine to provide an option to simply do that in backward: `y.backward(consume_self=True)`.

toslunar:
If it's OK to make `Variable.backward` an "attribute", we can write
```
cont = y.backward
del y
cont(**kwargs)
```

beam2d:
`backward` as an attribute is strange, so I want to avoid that.

toslunar:
Now `Variable.backward` is a method and the following works.
```
backward = y.backward
del y
backward(**kwargs)
```

beam2d:
I now come to feel the `execute=False` version is enough for now. Writing it with a short comment is enough for readers to understand the intention of the idiom, e.g. in `Optimizer`. We may want a cleaner interface without such an idiom in the future, though.

beam2d:
`movemethod` is interesting, but the implementation feels too complicated and hacky.

toslunar:
`chainer.backward` as in https://github.com/chainer/chainer/issues/5709#issuecomment-441972123 is an alternative place to implement the move semantics.
```python
chainer.backward(outputs, grad_outputs, move=True)
```

toslunar:
For simplicity, this PR will depend on #5741.

chainer/variable.py
```diff
@@ -1346,8 +1356,32 @@ def backward(self, retain_grad=False, enable_double_backprop=False,
...
+
+        ref_self = [self]
+
+        def cont():
+            if not ref_self:
+                raise RuntimeError(
+                    'the continuation Variable.backward(_return_cont=True) '
+                    'has been consumed')
+            outputs = ref_self
+            if sys.getrefcount(ref_self[0]) > 2:
```
beam2d: Is it a good way to assert it here? How about keeping the week ref to an output and checking that it is invalid after `_backprop_to_all`?

chainer/variable.py
```diff
@@ -1464,15 +1498,43 @@ def __bool__(self):
     __hash__ = None
 
 
+if six.PY3:
+    OrderedDict = collections.OrderedDict
+else:
+    # Reference counting cannot free keys in old `collections.OrderedDict`,
+    # where a doubly linked list is used to maintain the order.
+    class OrderedDict(object):
```
beam2d: How about moving it to `utils`?
Another (more minor) point: can we use a plain `dict` for py>3.6?

chainer/utils/__init__.py
```diff
@@ -1,4 +1,4 @@
-import collections
+import collections as _collections
```
beam2d: I just noticed; does it work correctly on py2? It would be better to use `from __future__ import absolute_import` to avoid conflicts with `./collections.py`.

toslunar:
The approach does not seem to work well with the `FooGrad(self).apply(...)` pattern.
https://github.com/chainer/chainer/blob/8c51c6be1a1b4f4e61677c8c05dbec782e2809bf/chainer/functions/pooling/max_pooling_2d.py#L124-L125

niboshi:
ping

PR #5287, opened 2018-09-02
fix inner bounds +1 before spatial_scale multiplication

knorth55:
This is a bug fix.
fix bounding box inner bounds +1.
I think the previous code is wrong for bounding box boundings,
but @yuyu2172, can you check this PR?


knorth55:
@yuyu2172 can you check this?
I think I should keep old function way, too.

yuyu2172:
> I think I should keep old function way, too.

Yes. 

I will discuss with Chainer team on whether to change functions in Chainer, or make new ones in ChainerCV. Please wait until then.

Hakuyume:
better API was implemented in #5304

PR #5288, opened 2018-09-02
add bounds arg for roi_pooling

knorth55:
DO NOT MERGE YET
MERGE AFTER #5287 

there are 2 format of bounding box representation; inner_bounds and
outer_bounds.
For inner_bounds, `roi_height = y_max - y_min + 1` and `roi_width =
x_max - x_min + 1`.
But for outer_bounds, `roi_height = y_max - y_min` and `roi_width =
x_max - x_min` 
I set arg for this two representations.

### DIFF
--- a/chainer/functions/pooling/roi_pooling_2d.py
+++ b/chainer/functions/pooling/roi_pooling_2d.py
@@ -53,7 +53,7 @@ class ROIPooling2D(function.Function):
     def __init__(self, outh, outw, spatial_scale, bounds):
         self.outh, self.outw = outh, outw
         self.spatial_scale = spatial_scale
-        if not isinstance(bounds, str) or bounds not in ['inner', 'outer']:
+        if bounds not in ['inner', 'outer']:
             raise TypeError(
                 'bounds must be string and \'inner\' or \'outer\': {}, {}'
                 .format(type(bounds), bounds))
--- a/chainer/functions/pooling/roi_pooling_2d.py
+++ b/chainer/functions/pooling/roi_pooling_2d.py
@@ -141,8 +141,10 @@ class ROIPooling2D(function.Function):
             int roi_batch_ind = bottom_rois[num * 5 + 0];
             int roi_start_w = round(bottom_rois[num * 5 + 1] * spatial_scale);
             int roi_start_h = round(bottom_rois[num * 5 + 2] * spatial_scale);
-            int roi_end_w = round((bottom_rois[num * 5 + 3] + 1.0) * spatial_scale);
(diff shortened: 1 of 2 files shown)


chainer/functions/pooling/roi_align_2d.py
```diff
@@ -540,6 +568,8 @@ def roi_align_2d(x, rois, outh, outw, spatial_scale, sampling_ratio=0):
             It must meet >=0 and is automatically decided when 0 is passed.
             Use of different ratio in height and width axis is also supported
             by passing tuple of int as (sampling_ratio_h, sampling_ratio_w).
+        inner_bounds (bool): Bounding box represents inner bounds or not.
+            Default is `False`.
```
beam2d: Can we use str instead of bool to represent the bounds mode, like `'inner'` and `'outer'`? (I thought a bool flag should be used when the argument represents on/off of a feature, but the case here is not like that)

chainer/functions/pooling/roi_pooling_2d.py
```diff
@@ -338,6 +356,8 @@ def roi_pooling_2d(x, rois, outh, outw, spatial_scale):
         outh (int): Height of output image after pooled.
         outw (int): Width of output image after pooled.
         spatial_scale (float): Scale of the roi is resized.
+        inner_bounds (bool): Bounding box represents inner bounds or not.
+            Default is `True`.
```
beam2d: ditto

chainer/functions/pooling/roi_align_2d.py
knorth55: Updated. I modified `bounds` and check the input.

chainer/functions/pooling/roi_pooling_2d.py
```diff
@@ -50,9 +50,14 @@ class ROIPooling2D(function.Function):
 
     """RoI pooling over a set of 2d planes."""
 
-    def __init__(self, outh, outw, spatial_scale):
+    def __init__(self, outh, outw, spatial_scale, bounds):
         self.outh, self.outw = outh, outw
         self.spatial_scale = spatial_scale
+        if not isinstance(bounds, str) or bounds not in ['inner', 'outer']:
```
beam2d: Just checking `bounds not in ['inner', 'outer']` looks enough.

knorth55: OK, I will update.

knorth55:
DO NOT MERGE YET
MERGE AFTER #5287

Hakuyume:
better API was implemented in #5304

PR #5289, opened 2018-09-02
add roi_indices arg for roi_align_2d and roi_pooling_2d

knorth55:
current inteface accepts `rois` formatted as
`(batch_index, x_min, y_min, x_max, y_max)`, but there is no need to
concatenate `batch_index` and `(x_min, y_min, x_max, y_max)`.
I think it is better to move `roi_pooling` and `roi_align` interfaces
to accept `rois` and `roi_indices` separately.
This PR is the first step of the inteface change, and it keeps
back-compatibilty with `roi_indices=None (default)`.
When you pass `roi_indices`, you can use new interface.

cc. @wkentaro

### DIFF
--- a/chainer/functions/pooling/roi_align_2d.py
+++ b/chainer/functions/pooling/roi_align_2d.py
@@ -20,6 +20,7 @@ import collections
 import numpy
 import six
 
+import chainer
 from chainer.backends import cuda
 from chainer import function
 from chainer.utils import type_check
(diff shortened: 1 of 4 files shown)


knorth55:
we have discussed and add new interface functions `roi_average_pooling_2d` `roi_max_pooling_2d` `roi_average_align_2d` `roi_max_align_2d`

PR #5291, opened 2018-09-03
Fix Dockerfile in v4 branch to use CuPy minor versions

kmaehashi:
**This PR is for v4 branch.**

This intends to use CuPy minor versions (hot fix).


okuta:
LGTM!

PR #5294, opened 2018-09-05
Amend chainer.Chain document

### LINKED ISSUE #5293: `Chain.children` not returns its children in registered order
## Version
chainer==5.0.0b4

## Problem

In the documentation of `chainer.Chain`, `children` is explained as follows:
```
    The child link is accessible via :meth:`children`
    generator, which returns a generator running through the children in
    registered order.
```

However, `Chain.children` returns its children sorted by their names.

```
    def children(self):
        d = self.__dict__
        for name in sorted(self._children):
            yield d[name]
```

## Cause
I think the documentation is wrong.

fiarabbit:
fixed #5293.


kmaehashi:
LGTM!

PR #5226, opened 2018-08-14
Integrate ChainerMN to Chainer repository

keisukefukuda:
This PR integrates ChainerMN, an add-on package to support multinode training.
Background:
 - Distributed deep learning is getting more popular, thus Chainer should natively support it
 - As both of Chainer and ChainerMN are continuously updated, it is getting harder to support all version combinations of the softwares.

What this PR does:
 - Move ChainerMN to Chainer's repository and synchronize the release cycle.
 - Integrate `.travis.yml` setup

What this PR doesn't 
 - It does not change the library usage. The `chainermn` package remains as is and existing ChainerMN programs work without modification.
 - It doesn't am to integrate various components of Chainer and ChainerMN. There are several duplicate building blocks between Chainer and ChainerMN, but we focus on integrating the repositories and do not step further for now.

Major TODOs:
 - Integrate `.travis.yml` 
 - Integrate documents.
 - Check coding styles of both projects.
 - Improve error messages when `mpi4py` is not installed

PR #5296, opened 2018-09-06
Fix location of chainermn docs in sidebar

kmaehashi:
I think it's better to place ChainerMN docs after Chainer Reference.

Ref #5226.


kmaehashi:
Jenkins, test this please.

PR #3628, opened 2017-10-21
Improve version embedding

okuta:
Current Chainer requires installed state.
This fix relaxing limitation of environment. For example, [Pyinstaller.](http://www.pyinstaller.org/)

PR #5298, opened 2018-09-06
Avoid using `pkg_resources` to retrieve Chainer version

kmaehashi:
`chainermn.__version__` will be kept for backward compatibility, and now it points to the version of Chainer.
However, we should avoid using `pkg_resources` (c.f. #3628)

Ref #5226.


keisukefukuda:
Jenkins, test this please.

PR #5306, opened 2018-09-08
Introduce `utils.mixed_presision` decorator

takagi:
This PR introduces `utils.mixed_presision` decorator that makes forward computation perform in FP32 for FP16 inputs, returning outputs casted back to FP16. It does nothing for FP32 and FP64 inputs.

```Python
class BatchDet(function_node.FunctionNode):
    ...
    @utils.mixed_precision
    def forward_cpu(self, x):
        self.retain_inputs((0,))
        self.retain_outputs((0,))
        detx = utils.force_array(numpy.linalg.det(x[0]))  # numpy.linalg.det does not support FP16
        return detx,
    ...
```

### DIFF
--- a/chainer/utils/precision.py
+++ b/chainer/utils/precision.py
@@ -9,7 +9,8 @@ def mixed_precision(fn):
     """
     @functools.wraps(fn)
     def wrapper(self, in_data):
-        flag = any([x.dtype == numpy.float16 for x in in_data])
+        flag = all([x.dtype.kind != 'f' or x.dtype == numpy.float16
+                    for x in in_data])
 
         in_data1 = []
         for x in in_data:
--- a/chainer/utils/precision.py
+++ b/chainer/utils/precision.py
@@ -2,7 +2,7 @@ import numpy
 import functools
 
 
-def mixed_precision(fn):
+def _fp16_mixed_precision_helper(fn):
     """Decorator to perform forward computation in FP32 for FP16 inputs,
        returning outputs casted back to FP16. Do nothing for FP32 and FP64
        inputs.
--- a/chainer/utils/precision.py
+++ b/chainer/utils/precision.py
@@ -10,13 +10,25 @@ def mixed_precision(fn):
     @functools.wraps(fn)
     def wrapper(self, in_data):
         flag = any([x.dtype == numpy.float16 for x in in_data])
-        in_data = tuple([
-            x.astype(numpy.float32) if x.dtype == numpy.float16 else x
-            for x in in_data])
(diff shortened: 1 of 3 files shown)


takagi:
Jenkins, test this please.

tests/chainer_tests/utils_tests/test_precision.py
```diff
@@ -0,0 +1,58 @@
...
+        assert f.x[1].dtype == numpy.float32
+        assert y[0].dtype == numpy.float16
+        assert y[1].dtype == numpy.float16
+
+    def test_fp32(self):
+        x = (numpy.zeros((1, 2, 3), dtype=numpy.float32),) * 2
+        f = F()
+        y = f.apply(x)
+        assert f.x[0] is x[0]
+        assert f.x[1] is x[1]
```
toslunar: This test is weak since `x[0]` is `x[1]`.

takagi: oh...

takagi: Fixed!

chainer/utils/precision.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+def mixed_precision(fn):
+    """Decorator to perform forward computation in FP32 for FP16 inputs,
+       returning outputs casted back to FP16. Do nothing for FP32 and FP64
+       inputs.
+    """
+    @functools.wraps(fn)
+    def wrapper(self, in_data):
+        flag = any([x.dtype == numpy.float16 for x in in_data])
```
okuta: Why `any` ?

takagi: I think of cases `float16` data and that of other dtype, e.g. integer for labels, are supplied in `in_data`.

takagi: I got your point. Makes sense.

takagi: Fixed!

okuta:
jenkins, test this please.

toslunar:
The name of the decorator could be discussed, if it'll be a public one.

Within this decorator everything is computed in float32 (or float64), but by the name `mixed_precision` a user might expect a decorator that replaces some of float16 operations inside the function by inspecting it.

okuta:
OK, cloud you change to private method.
How about `_fp16_helper` ?

takagi:
I agree that its name would not be clear for users about how much "mixed". I will make it private for now.

okuta:
jenkins, test this please.

okuta:
LGTM!

PR #5307, opened 2018-09-08
remove deprecated is_linear arg

knorth55:
it seems deprecated arg `is_linear` causes test fail in `AppVeyor`.
https://github.com/chainer/chainer/pull/5198#issuecomment-419634854


toslunar:
Jenkins, test this please.

PR #5312, opened 2018-09-10
Remove unnecessary version checking in ChainerMN

keisukefukuda:
This PR removes unnecessary version checking in ChainerMN.

Background:
Before ChainerMN was merged into Chainer repository, ChainerMN has to check Chainer version installed in the environment. This was because some features in ChainerMN depended on Chainer's internal structures. However, ChainerMN is now integrated to Chainer and installed together. These version checking code is no longer necessary.


kmaehashi:
LGTM!

PR #5313, opened 2018-09-10
Add discriminative margin based clustering loss

dBeker:
Cleanup of the previous PR. History can be found at https://github.com/chainer/chainer/pull/4525


delta2323:
Regarding [this comment](https://github.com/chainer/chainer/pull/4525#discussion_r215881971) about the error propagation through mean values, sometimes we intentionally stop error propagation from the intermediate values (e.g. Target Q-Network in Deep Q Network). So, it could be possible the current implementaion is correct. I'm wondering how the original implementation deal with it.

chainer/functions/loss/discriminative_loss.py
```diff
@@ -0,0 +1,168 @@
...
+        l_var = c_sum(l_var) / self.max_embedding_dim
+
+        # Calculate distance loss
+        for c_a in range(len(mns)):
+            for c_b in range(c_a + 1, len(mns)):
+                m_a = mns[c_a]
+                m_b = mns[c_b]
+                dist = self.norm(m_a - m_b, 1)  # N
+                l_dist += c_sum((relu(2 * self.delta_d - dist)) ** 2)
+                count += 1
```
delta2323: If we change the code so that computation of regularization and variance is free from for-loops, it would be better to change the computation of distance loss, too, because it takes O(N^2), which is dominant in computation time. But I also think it is ok that we change the implementation after we merge this PR.

delta2323: As I wrote in the previous PR, there is a trade-off between memory consumption and computation time. So, it is up to actual usecases which is better. Do you have any idea?

dBeker: Actually, I couldn't come up with a proper idea about dist term with matrix multiplication manner. We need to filter zero-means distance calculation (just noticed that current implementation has a bug about it). With matrix multiplication manner, we need some sort of mask to achieve it. Hence, I wonder if it will be much more efficient or not. So, I would like to keep it with current implementation as the current one is more like O(N * log(N)) rather than O(N ^ 2) - second for loop starts from the value of first for the loop -

delta2323: OK, I understand it. Let's reconsider the improvement when we encounter issues in current implementation.

dBeker:
@delta2323 I fixed the build issues

delta2323:
OK, I'll take a look.

delta2323:
jenkins, test this please.

dBeker:
jenkins, test this please.

delta2323:
jenkins, test this please.

delta2323:
LGTM

Issue #5301, opened 2018-09-07
Flake8 environment setup


Crissman:
Chainer docs don't clarify how to set up flake8 so that it doesn't report irrelevant errors, like line length. If the `stylecheck` requirements from `setup.py` are installed manually, then `flake8` gives appropriate messages.

Need to explain to beginners how to install these in the `contribution` docs.

kmaehashi:
Fixed via #5318

PR #5318, opened 2018-09-10
Add quotes to stylecheck install

Crissman:
Add quotation marks to `pip install -e '.[stylecheck]'` to allow it to work on shells like zsh that allow brace expansion.

Backwards compatible with bash shell.

Addresses issue #5301


kmaehashi:
Jenkins, test this please.

PR #5322, opened 2018-09-11
Make link.to_*pu return self

### LINKED ISSUE #5321: Links to_gpu return None
I'm trying to construct an LSTM network with multi GPUs using Chainer (v4.0.0b1). As in the following code.

```
import numpy as np
import chainer
from chainer import optimizers, Chain, training, iterators, serializers, cuda, Variable
import chainer.functions as F
import chainer.links as L

...

class Network(Chain):
    def __init__(self):
        super(Network, self).__init__()
        with self.init_scope():
            ...
            self.fc1  = L.Liner(3000, 1000).to_gpu(1)
            self.lstm = L.LSTM(1000, 1000).to_gpu(1)
            self.fc2  = L.Liner(1000, 3000).to_gpu(1)
            ...

    def __call__(self, x, t):
        ...

...
```
However, the LSTM link becomes "NoneType". As in the following error in call.
>TypeError: 'NoneType' object is not callble

I thought it was strange so I displayed 'self.lstm'. As a result, "None" was displayed. For example, fc1 which is "Link" is displayed as follows.

><chainer.links.connection.linear.Linear object at hogehoge>

I found out that 'self.lstm' could not be declared as Link in 'self.lstm = L.LSTM(1000, 1000).to_gpu(1)'. However, I don't know why I can't declare it.

I use [Chainer's Docker](https://ngc.nvidia.com/registry/partners-chainer) as the execution environment.

- Chainer 4.0.0b1
- CuPy 4.0.0b1
- Ubuntu 16.04.3 LTS
- Cuda 9.0.176
- cudnn 7.0.5

toslunar:
Fix #5321

### DIFF
--- a/chainer/links/connection/gru.py
+++ b/chainer/links/connection/gru.py
@@ -205,11 +205,13 @@ class StatefulGRU(GRUBase):
         super(StatefulGRU, self).to_cpu()
         if self.h is not None:
             self.h.to_cpu()
+        return self
 
     def to_gpu(self, device=None):
         super(StatefulGRU, self).to_gpu(device)
         if self.h is not None:
             self.h.to_gpu(device)
+        return self
 
     def set_state(self, h):
         assert isinstance(h, variable.Variable)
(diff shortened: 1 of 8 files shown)


okuta:
jenkins, test this please.

okuta:
LGTM!

PR #5328, opened 2018-09-12
Disallow summing over ellipsis in `F.einsum`

toslunar:
This feature is controversial and I'd rather not support it in the next stable release.


okuta:
LGTM!

PR #5330, opened 2018-09-12
Automatic broadcast: minimum, maximum, where

### LINKED ISSUE #5324: Automatic broadcast for F.maximum, F.minimum
These functions should support automatic broadcast.

toslunar:
Close #5324.

### DIFF
--- a/chainer/functions/array/where.py
+++ b/chainer/functions/array/where.py
@@ -26,8 +26,6 @@ class Where(function_node.FunctionNode):
         self.retain_inputs((0,))
         xp = cuda.get_array_module(*inputs)
         condition, x, y = inputs
-        self.x_shape = x.shape
-        self.y_shape = y.shape
         return xp.where(condition, x, y),
 
     def backward(self, indexes, grad_outputs):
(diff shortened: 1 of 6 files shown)


toslunar:
Jenkins, test this please.

toslunar:
Passed `test_(double_)backward_cpu` with `@condition.repeat(30000)`.

chainer/functions/array/where.py
```diff
@@ -16,30 +17,33 @@ def check_type_forward(self, in_types):
...
+        type_check.expect_broadcast_shapes(
+            c_type.shape, x_type.shape, y_type.shape)
 
     def forward(self, inputs):
+        # may broadcast
         self.retain_inputs((0,))
         xp = cuda.get_array_module(*inputs)
         condition, x, y = inputs
+        self.x_shape = x.shape
+        self.y_shape = y.shape
```
beam2d: You can get them in backward with `self.inputs[i].shape` (i=1,2), so no need to save them manually.

chainer/functions/array/where.py
```diff
@@ -16,30 +17,33 @@ def check_type_forward(self, in_types):
...
 
     def backward(self, indexes, grad_outputs):
         condition = self.get_retained_inputs()[0]
         xp = cuda.get_array_module(condition.data)
         g, = grad_outputs
-        zeros = xp.zeros(g.shape, dtype=g.dtype)
+        zero = xp.zeros((), dtype=g.dtype)
         ret = []
         if 0 in indexes:
             ret.append(None)
```
beam2d: It's not related to this PR, but I found that the first argument (condition) could be an `__init__` parameter instead of the `forward` argument, then there is no need of constructing Variable object for it (and no need to retain). It's ok to leave it in this PR.

beam2d:
Jenkins, test this please.

beam2d:
LGTM

PR #5335, opened 2018-09-12
Support all float dtypes in `F.scatter_add` and `F.get_item`

takagi:
Merge #5306 first.

~https://github.com/takagi/chainer/compare/mixed-precision-decorator...takagi:dtype-scatter-add~

This PR follows #4582 to support all float dtypes in `F.scatter_add` as well as `F.get_item` using each other in backward computation. ~Note that because some routines used do not support FP16 data, they compute FP16 inputs in FP32.~

Merge https://github.com/cupy/cupy/pull/1684 also before this PR.

### DIFF
--- a/chainer/functions/array/get_item.py
+++ b/chainer/functions/array/get_item.py
@@ -42,20 +42,20 @@ class GetItem(function_node.FunctionNode):
 
     def backward(self, indexes, gy):
         return GetItemGrad(
-            self.slices, self.inputs[0].shape, self.inputs[0].dtype).apply(gy)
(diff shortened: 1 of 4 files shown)


toslunar:
#5306 is merged.

takagi:
I think of using https://github.com/cupy/cupy/pull/1684 .

toslunar:
The Jenkins failure (`tests.chainer_tests.functions_tests.loss_tests.test_contrastive.TestContrastive_param_63.test_double_backward_gpu`) is unrelated to the PR.

okuta:
Thanks.
LGTM!

PR #5344, opened 2018-09-13
Move `copyto` to `chainer.backend`

toslunar:
Convention on naming modules is discussed by @niboshi and @beam2d.


niboshi:
Jenkins failure is irrelevant.

niboshi:
LGTM

PR #5350, opened 2018-09-13
[WIP] Support all float dtypes in `F.connectionist_temporal_classification`

takagi:
Merge #5306 first.

https://github.com/takagi/chainer/compare/mixed-precision-decorator...takagi:dtype-ctc

This PR follows #4582 to support all float dtypes in `F.connectionist_temporal_classification`. Note that `atomicAdd` used does not support FP16 data, it computes FP16 inputs in FP32.


takagi:
I will make a PR for this issue again.

PR #5353, opened 2018-09-14
For proper resuming, don't raise KeyError at UpdateRule deserialization

grafi-tt:
When a network is trained with some parameters set `update_rule.enabled=False`, the snapshot of an optimizer can't be deserialized if `strict=False` option isn't set to the deserializer.

This PR fixes the problem by checking `update_rule.enabled` at the time of deserialization. Some tests are also added.

The following is the minimal example that reproduce the issue.

```python
import chainer
from chainer import functions, links, optimizers, serializers
import numpy


class Model(chainer.Chain):
    def __init__(self):
        super().__init__()
        with self.init_scope():
            self.l1 = links.Linear(10, 10)
            self.l2 = links.Linear(10, 10)

    def __call__(self, x, t):
        h = self.l1(x)
        h = self.l2(h)
        h = functions.mean(h, axis=1)
        return functions.sigmoid_cross_entropy(h, t)


def main():
    model = Model()
    optimizer = optimizers.MomentumSGD()
    optimizer.setup(model)
    model.l1.W.update_rule.enabled = False

    x = numpy.random.random((4, 10)).astype(numpy.float32)
    t = numpy.asarray([0, 1, 0, 1], dtype=numpy.int32)
    optimizer.update(model, x, t)

    serializers.save_npz('optimizer.npz', optimizer)

    model = Model()
    optimizer = optimizers.MomentumSGD()
    optimizer.setup(model)
    model.l1.W.update_rule.enabled = False

    serializers.load_npz('optimizer.npz', optimizer)


if __name__ == '__main__':
    main()
```

Even though the `update_rule.enabled` options are kept to be **the same** across two models, `strict=False` is required for the `load_npz` function. I'd like to say this is a bug.

I checked a related issue #3125. Its fix doesn't check `update_rule.enabled` at the time of deserialization. I think `strict=False` should be required **only** if the `update_rule.enabled` options are **changed** when loading the snapshot.

### DIFF
--- a/chainer/optimizer.py
+++ b/chainer/optimizer.py
@@ -292,13 +292,20 @@ class UpdateRule(object):
                 self_copy.init_state(variable.Variable(arr, grad=arr))
 
                 for key in self._state:
-                    self._state[key] = serializer(key, None)
+                    try:
+                        value = serializer(key, None)
+                    except KeyError:
+                        if self.enabled:
+                            raise
+                        value = None
                     # leave the update rule state as `None` if the keys are not
                     # contained in the snapshot, so that these states can be
                     # automatically initialized with the `_prepare` method
-                    if self._state[key] is None:
+                    if value is None:
                         self._state = None
                         break
+                    else:
+                        self._state[key] = value
         else:
             for key in self._state:
                 self._state[key] = serializer(key, self._state[key])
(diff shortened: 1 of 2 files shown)


toslunar:
Jenkins, test this please.

toslunar:
The Jenkins failure (`tests.chainer_tests.functions_tests.pooling_tests.test_average_pooling_nd.TestAveragePoolingND_param_0.test_double_backward_cpu`) is unrelated to the PR.

PR #5356, opened 2018-09-15
Rewrite the example of VAE using Chainer distributions

ganow:
I rewrote the official example of Variational Autoencoder in `examples/vae` by using Chainer distributions.

This code records 83.1864 validation loss by 100 epochs and 20 latent variable dimensions, the parameter setting of which is the same as the default behavior of the previous code. A visualization of the reconstructed images is shown below.

The figure shown below represents the random samples from the validation dataset.
test

The images reconstructed from these random samples are shown below.
test_reconstruction

I have to note that there are two breaking changes from the previous codes.

First, I change the dataset from MNIST to binarized MNIST. It is because Variational Autoencoder typically models each pixel of the MNIST images as Bernoulli distribution so as this code does. Bernoulli distribution is defined on a discrete variable {0, 1}, and `D.Bernoulli(p).log_prob(x)` returns `0` or `-inf` when x is not in {0, 1}. This behavior collapses training for continuous input. For this reason, we need to discretize pixel values when we want to use Bernoulli distribution.

Second, I removed `train_vae_custom_loop.py`. The `train_vae_custom_loop.py` is the version which does not use a trainer and an updater. I found that this code basically does the same thing as `train_vae.py`, and I felt that explaining how to use a trainer and an updater is outside the scope of this example. I can add train_vae_custom_loop.py again if it is needed.


crcrpar:
Quick look your code and I like this but some parameters seem to be changed or removed.

For example, `beta` and `k` are missed.

examples/vae/net.py
```diff
@@ -1,67 +1,100 @@
...
+        q_z = self.encoder(x)
+        z = q_z.sample()
+        p_x = self.decoder(z)
+        p_z = self.prior()
+
+        self.loss = None
+        self.rec = None
+        self.penalty = None
+        self.rec = F.mean(F.sum(p_x.log_prob(x), axis=-1))
+        self.penalty = F.mean(
```
YoshikawaMasashi: I think it is better to use `kl_divergence`.
https://docs.chainer.org/en/latest/reference/generated/chainer.kl_divergence.html#chainer.kl_divergence

examples/vae/train_vae.py
```diff
@@ -20,60 +21,72 @@ def main():
...
+        chainer.serializers.load_npz(args.initmodel, avg_elbo_loss)
 
     # Load the MNIST dataset
     train, test = chainer.datasets.get_mnist(withlabel=False)
+
+    # Binarize dataset
+    train[train >= 0.5] = 1.0
+    train[train < 0.5] = 0.0
+    test[test >= 0.5] = 1.0
+    test[test < 0.5] = 0.0
```
YoshikawaMasashi: I'll implement not strict option for `D.Bernoulli`.

ganow:
> Quick look your code and I like this but some parameters seem to be changed or removed.
> 
> For example, `beta` and `k` are missed.

Thank you for mentioning them. I missed these parameters because the previous code did not provide any methods to change them via CLI. I think it's better to incorporate them and provide the command line options, so I'll fix it.

examples/vae/net.py
ganow: I'll change it, thank you.

examples/vae/train_vae.py
YoshikawaMasashi: #5363

ganow:
> > Quick look your code and I like this but some parameters seem to be changed or removed.
> > For example, `beta` and `k` are missed.
> 
> Thank you for mentioning them. I missed these parameters because the previous code did not provide any methods to change them via CLI. I think it's better to incorporate them and provide the command line options, so I'll fix it.

I changed the code so that to incorporate `beta`, but I realized that to incorporate `k` is a little difficult. If we simply put `k` into the posterior distribution like `z = q_z.sample(k)`, the shape of `z` becomes `(k, batch_size, dim)`. This modification changes `z.ndim` and following operation `decoder(z)` raises an error shown below:

```
...
Invalid operation is performed in: LinearFunction (Forward)

Expect: x.shape[1] == W.shape[1]
Actual: <batch_size * dim> != <dim>
```

I have some idea to solve this problem dirtily, like using `F.reshape()` several times. If someone has any better, or appropriate way to solve, I would like you to ask.

ganow:
> Does `n_batch_axes=2` work?

It worked, thank you, but `D.Bernoulli(logit=self.linear(z, n_batch_axes=2))` raised following error. `D.Bernoulli` seems to only accept two-dimensional array in total.

```
Exception in main training loop: boolean index did not match indexed array along dimension 0; dimension is 10 but corresponding boolean dimension is 100
...
  File "/Users/ganow/local/src/github.com/ganow/chainer/examples/vae/net.py", line 34, in __call__
    self.rec = F.mean(F.sum(p_x.log_prob(x), axis=-1))
  File "/Users/ganow/.pyenv/versions/anaconda3-5.2.0/lib/python3.6/site-packages/chainer/distributions/bernoulli.py", line 113, in log_prob
    return _bernoulli_log_prob(self.logit, x)
  File "/Users/ganow/.pyenv/versions/anaconda3-5.2.0/lib/python3.6/site-packages/chainer/distributions/bernoulli.py", line 53, in _bernoulli_log_prob
    y, = BernoulliLogProb().apply((logit, x))
  File "/Users/ganow/.pyenv/versions/anaconda3-5.2.0/lib/python3.6/site-packages/chainer/function_node.py", line 257, in apply
  File "/Users/ganow/.pyenv/versions/anaconda3-5.2.0/lib/python3.6/site-packages/chainer/distributions/bernoulli.py", line 23, in forward
    y[self.invalid] = - xp.inf
...
```

ganow:
In VAE, we typically do `D.Bernoulli(logit=NN(q_z.sample(k))).log_prob(x)` such that `x.shape == (batch_size, dim)` and `NN(q_z.sample(k)).shape == (k, batch_size, dim)`. Currently `D.Bernoulli` seems not to work when `logit.ndim > x.ndim`, so I implemented `k` option with `F.broadcast_to()`.

examples/vae/train_vae.py
YoshikawaMasashi: this PR is merged. so you can use `D.Bernoulli` without binarization.

examples/vae/train_vae.py
```diff
@@ -20,60 +21,79 @@ def main():
...
     if args.initmodel:
-        chainer.serializers.load_npz(args.initmodel, model)
+        chainer.serializers.load_npz(args.initmodel, avg_elbo_loss)
 
     # Load the MNIST dataset
     train, test = chainer.datasets.get_mnist(withlabel=False)
+
+    # Binarize dataset
+    train[train >= 0.5] = 1.0
+    train[train < 0.5] = 0.0
```
toslunar: How about `train = (train >= 0.5).astype(np.float32)`?

examples/vae/net.py
```diff
@@ -1,67 +1,103 @@
...
+        self.k = k
+
+        with self.init_scope():
+            self.encoder = encoder
+            self.decoder = decoder
+            self.prior = prior
+
+        self.loss = None
+        self.rec = None
+        self.penalty = None
```
toslunar: It seems these loss values don't need to be saved to attributes of the chain.  How about `reporter.report({'loss':, loss}, self)` etc?

examples/vae/train_vae.py
```diff
@@ -20,60 +21,79 @@ def main():
...
-                                        eval_func=model.get_loss_func(k=10)))
+    trainer.extend(extensions.Evaluator(
+        test_iter, avg_elbo_loss, device=args.gpu))
     trainer.extend(extensions.dump_graph('main/loss'))
     trainer.extend(extensions.snapshot(), trigger=(args.epoch, 'epoch'))
     trainer.extend(extensions.LogReport())
     trainer.extend(extensions.PrintReport(
         ['epoch', 'main/loss', 'validation/main/loss',
-         'main/rec_loss', 'validation/main/rec_loss', 'elapsed_time']))
+         'main/rec', 'main/penalty', 'elapsed_time']))
```
toslunar: `rec` might be too short to guess its meaning.
I'd write `reconstr` and `kl_div`.  What do you think?

examples/vae/net.py
```diff
@@ -1,67 +1,103 @@
...
+    def forward(self, z, inference=False):
+        n_batch_axes = 1 if inference else 2
+        h = F.tanh(self.linear(z, n_batch_axes=n_batch_axes))
+        h = self.output(h, n_batch_axes=n_batch_axes)
+        return D.Bernoulli(logit=h)
+
+
+class Prior(chainer.Chain):
+
+    def __init__(self, n_latent, dtype=np.float32, device=-1):
```
toslunar: Please do not specify `device` here.  Why don't you make `Prior` inherit `chainer.Link` and add `loc` and `scale` as `chainer.Parameter`s?

examples/vae/net.py
```diff
@@ -1,67 +1,103 @@
...
+import chainer.distributions as D
 import chainer.functions as F
-from chainer.functions.loss.vae import gaussian_kl_divergence
 import chainer.links as L
+from chainer import reporter
 
 
-class VAE(chainer.Chain):
-    """Variational AutoEncoder"""
+class AvgELBOLoss(chainer.Chain):
```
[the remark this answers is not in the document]
ganow: I've added it: 3caaae1

examples/vae/train_vae.py
ganow: Thank you. It seems smarter than the previous way. I've changed it: 099a98c

examples/vae/net.py
ganow: fixed them: d20a9b9

examples/vae/train_vae.py
ganow: Thank you for your recommendation. I think `kl_div` is still vague because this name does not specify KL divergence for what. Since I thought that the variable name of KL penalty should imply that it is the penalty term of the loss function, I have changed `rec` and `penalty` to `reconstr` and `kl_penalty`: add8bfe

examples/vae/net.py
ganow: I fixed it: 39f8620

ganow: I have just realized that the prior distribution of the original VAE paper is not trainable. My current implementation (which is based on your advice) makes `loc` and `scale` as `chainer.Parameter`s and it makes these variables trainable. I couldn't find how to allocate non-trainable variable without using `device` option. Are there any ways to do that?

toslunar: Could you try `register_persistent`?

ganow: parameters for prior distribution now seem to be non-trainable, thank you. 8294d6f

examples/vae/net.py
```diff
@@ -1,67 +1,112 @@
...
+
+
+class Prior(chainer.Link):
+
+    def __init__(self, n_latent):
+        super(Prior, self).__init__()
+
+        with self.init_scope():
+            self.loc = chainer.Parameter(0, n_latent)
+            self.scale = chainer.Parameter(1, n_latent)
```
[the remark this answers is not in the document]
ganow: c107dda

examples/vae/train_vae.py
```diff
@@ -90,26 +112,27 @@ def save_images(x, filename):
             ai.imshow(xi.reshape(28, 28))
         fig.savefig(filename)
 
-    model.to_cpu()
+    encoder.to_cpu()
+    decoder.to_cpu()
+    prior.to_cpu()
```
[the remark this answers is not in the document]
ganow: 4fc6912

toslunar:
Jenkins, test this please.

examples/vae/net.py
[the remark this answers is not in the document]
toslunar: The lines
```
        self.register_persistent('loc')
        self.register_persistent('scale')
```
are needed.

examples/vae/net.py
```diff
@@ -1,67 +1,108 @@
...
+        return D.Bernoulli(logit=h, binary_check=self.binary_check)
+
+
+class Prior(chainer.Link):
+
+    def __init__(self, n_latent):
+        super(Prior, self).__init__()
+
+        self.loc = np.zeros(n_latent, np.float32)
+        self.scale = np.ones(n_latent, np.float32)
```
toslunar:
```suggestion
        self.scale = np.ones(n_latent, np.float32)
        self.register_persistent('loc')
        self.register_persistent('scale')
```

examples/vae/net.py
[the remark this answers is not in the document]
ganow: I misunderstood your recommendation. I fixed it, thank you. 61565f4

toslunar:
Jenkins, test this please.

toslunar:
The Jenkins failure (`BatchRenormalizationTest_param_7.test_forward_gpu`) is unrelated to the PR.

PR #5373, opened 2018-09-19
Make deserializers overwrite `intel64.mdarray`

toslunar:
Make deserializer overwrite arrays of any backend.

I locally tested the added tests fail without this fix.  Note that `type(value)(numpy.asarray(dataset))` created new `intel64.mdarray`.


beam2d:
LGTM

PR #5384, opened 2018-09-22
Mention chainer.backend in upgrade guide

hvy:
Mentioned the `chainer.backend` subpackage and the new location of `get_array_module` in the upgrade guide.


takagi:
LGTM!

PR #5385, opened 2018-09-22
Add missing test in Variable test

hvy:
Added missing grad attribute tests (there was an unused `check_grad` method).


toslunar:
Jenkins, test this please.

PR #5316, opened 2018-09-10
Update Chainer at a Glance

Crissman:
* Improve resolution of images--addresses #5130 
* Add glance.py code to examples directory
* Change Glance docs to refer to glance.py code instead of copy/paste segments
* Convert from `import chainer` to `import chainer as ch` style
* Use chainer.Sequential for defining network model
* Move to one-line format for setting up optimizer

PR #5388, opened 2018-09-26
Improve code sample appearance in docs

niboshi:
- Removes weird blank lines
- Adds line numbers

### DIFF
--- a/docs/source/glance.rst
+++ b/docs/source/glance.rst
@@ -34,16 +34,16 @@ Let's start the program. Here are the typical imports for a Chainer program. :mo
 .. literalinclude:: ../../examples/glance/glance.py
    :language: python
    :linenos:
-   :lines: 1-8
-   :lineno-start: 1
+   :lines: 6-13
+   :lineno-start: 6
 
 We'll use Matplotlib for the graphs to show training progress.
 
 .. literalinclude:: ../../examples/glance/glance.py
    :language: python
    :linenos:
-   :lines: 10-11
-   :lineno-start: 10
+   :lines: 15-16
+   :lineno-start: 15
 
 Trainer Structure
 ~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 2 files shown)


niboshi:
jenkins, test this please

niboshi:
jenkins, test this please

niboshi:
jenkins, test this please

toslunar:
On blank lines, do you have an idea to keep maintainability?  See https://github.com/chainer/chainer/pull/5316#issuecomment-420489939

niboshi:
Oh, I didn't notice it has already discussed.

I know it's less maintainable and I can't find any better way, but for me, current appearance is simply not acceptable. I would have the impression that this documentation is poorly written, which in turn means Chainer itself is poorly maintained.

kmaehashi:
Using `start-after` and `lines` may be (a bit) better option.

> Changed in version 1.6: With both start-after and lines in use, the first line as per start-after is considered to be with line number 1 for lines.

http://www.sphinx-doc.org/en/1.8/usage/restructuredtext/directives.html?highlight=literalinclude#directive-literalinclude

niboshi:
Thank you for suggestion.

If we're going to add line numbers, the maintenance cost is same: we need both the starting line number and the number of lines to show.

And I think it's better to show line numbers: (1) it's easier to view code and text in parallel (2) it implies the code fragments are extracts from external code (for those who missed the "The code below is from..." statement).

niboshi:
How do you think? I want to fix the appearance before v5 release.

toslunar:
How about adding some comments to `examples/glance/glance.py`?
```
# This example is referred by "Chainer at a Glance".
# If the file is modified, please update `docs/source/glance.rst`.
```

niboshi:
I think that's helpful :+1:

niboshi:
PTAL.

toslunar:
Jenkins, test this please.

PR #5394, opened 2018-09-27
Travis test against v5 branch

beam2d:
Stable branch has been switched to `v5`. This change lets Travis look that.


kmaehashi:
LGTM!

PR #5404, opened 2018-09-29
Add Eve optimizer

muupan:
This PR implements Eve optimizer following the revised version of the paper (https://arxiv.org/abs/1611.01505v3), while https://github.com/chainer/chainer/pull/1847 was based on the older version of the paper.

Because Eve requires loss values, `tests/chainer_tests/optimizers_tests/test_optimizers_by_linear_model.py` is modified to use `lossfun`.

See https://github.com/muupan/chainer-eve for experimental results on mnist.


chainer/optimizers/eve.py
```diff
@@ -0,0 +1,182 @@
...
+        assert self.d_tilde is not None
+        return _learning_rate(self.hyperparam, self.t, self.d_tilde)
+
+
+class Eve(optimizer.GradientMethod):
+
+    """Eve optimizer.
+
+    See: https://arxiv.org/abs/1611.01505v3
+
```
toslunar: On `weight_decay_rate` and `amsgrad`, could you refer the documentation of `Adam`?

chainer/optimizers/eve.py
```diff
@@ -0,0 +1,182 @@
...
+        eps (float): Small value for the numerical stability.
+        eta (float): Schedule multiplier, can be used for warm restarts.
+        weight_decay_rate (float): Weight decay rate.
+        amsgrad (bool): Whether to use the AMSGrad variant of Eve.
+    """
+
+    d_tilde = None
+
+    @property
+    def lr(self):
```
toslunar: `lr` is renamed to `alpha_t` in Chainer v6.  (#5420)

PR #5409, opened 2018-10-01
Use automatic broadcast and `force_array`

toslunar:


toslunar:
Jenkins, test this please.

PR #5420, opened 2018-10-04
Rename Adam.lr

toslunar:
The name `lr` is too dangerous to denote `\alpha_t`, the corrected `\alpha`, especially for users of other frameworks.

### DIFF
--- a/chainer/optimizers/adam.py
+++ b/chainer/optimizers/adam.py
@@ -189,7 +189,8 @@ class AdamRule(optimizer.UpdateRule):
     @property
     def lr(self):
         warnings.warn(
-            'AdamRule.lr is deprecated. Use AdamRule.alpha_t instead.',
+            'AdamRule.lr has been renamed to AdamRule.alpha_t. '
+            'Use of AdamRule.lr is deprecated in Chainer v6.',
             DeprecationWarning)
         return self.alpha_t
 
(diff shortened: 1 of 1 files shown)


toslunar:
See also #4056.

beam2d:
I agree that the name `lr` is much confusing for users of other frameworks as well as Chainer users who try to reproduce results from other frameworks, so it's good to rename.

chainer/optimizers/adam.py
```diff
@@ -148,41 +151,48 @@ def update_core_gpu(self, param):
...
 
     @property
-    def lr(self):
+    def alpha_t(self):
         return _learning_rate(self.hyperparam, self.t)
 
+    @property
+    def lr(self):
+        warnings.warn(
+            'AdamRule.lr is deprecated. Use AdamRule.alpha_t instead.',
```
kmaehashi: I think it's better to improve the message so that users can clearly understand this is just a rename.
e.g., `AdamRule.lr has been renamed to AdamRule.alpha_t. Use of AdamRule.lr is deprecated in Chainer v6.`

toslunar:
Fixed.

kmaehashi:
Jenkins, test this please.

kmaehashi:
LGTM!

PR #5424, opened 2018-10-04
Fix FunctionNode.retained_outputs

toslunar:
`func.outputs` is a tuple of `weakref`s of `VariableNode`s.

- [x] Add test

### DIFF
--- a/chainer/function_node.py
+++ b/chainer/function_node.py
@@ -645,7 +645,7 @@ Use apply() method instead.\
                 # Variable object.
                 output_var = variable.Variable(data)
                 output_var.creator_node = self
-                new_outputs[index] = weakref.ref(output_var)
+                new_outputs[index] = weakref.ref(output_var.node)
                 outputs_modified = True
             else:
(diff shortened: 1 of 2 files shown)


toslunar:
Added a test.

Without the fix, the test fails with the following message, while I've not investigated why the grad wasn't computed.
```
        # (x: Variable) requires grad
        # (y1_grad: ndarray) does not require grad
        gx, = chainer.grad([y1], [x], [y1_grad], enable_double_backprop=True)

        # assert gx == exp(x) * y1_grad
        xp.testing.assert_allclose(
            gx.array,
            xp.exp(x.array) * y1_grad)

        gx_, = chainer.grad([gx], [x], [x_grad_grad])
        xp.testing.assert_allclose(
>           gx_.array,
            gx.array * x_grad_grad)
E       AttributeError: 'NoneType' object has no attribute 'array'
```

okuta:
jenkins, test this please.

okuta:
LGTM!

PR #5427, opened 2018-10-04
Fix docstrings to follow OpenStack Style Guidelines 

toslunar:


toslunar:
Could you check English (if the new summaries are accurate)?

takagi:
How about `Trainer extension to gradually initialize an attribute of a given optimizer.` ? (I think what the original two-line description says is how it does.)

takagi:
LGTM!

PR #5428, opened 2018-10-04
Simplify scalar handling in `basic_math`

niboshi:
* Removed unnecessary preprocessing
* Use `math.log` in preference to `numpy.log` (`math` is faster)
* ~~Changed class names~~

### DIFF
--- a/chainer/functions/math/basic_math.py
+++ b/chainer/functions/math/basic_math.py
@@ -30,17 +30,6 @@ def _convert_value_to_string(value):
             'or a `Variable`.\nActual: {}'.format(type(value)))
 
 
-def _check_constant_type(value):
-    if numpy.isscalar(value):
-        return
-    elif isinstance(value, (numpy.ndarray, cuda.ndarray)):
-        return
-    else:
-        raise TypeError(
-            'Value must be a scalar, `numpy.ndarray`, `cupy.ndarray` '
-            'or a `Variable`.\nActual: {}'.format(type(value)))
-
-
 def _preprocess_const(x, value):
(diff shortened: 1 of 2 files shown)


toslunar:
> Use math.log in preference to numpy.log (math is faster)

I confirmed:
```
In [8]: dtype
Out[8]: dtype('float32')

In [10]: %timeit np.log(2.0, dtype=dtype)
1.29 µs ± 17.6 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [12]: %timeit dtype.type(math.log(2.0))
451 ns ± 13.2 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

toslunar:
Please explain why you want to change the name of `FunctionNode`s.

- The names `*Const` were clearer than `*Scalar` in that the `FunctionNode`s don't take the corresponding operands.
- Prefixing `_` to the names is acceptable because they have `label` attributes.  (Otherwise I'd rather not.)

niboshi:
I reverted class renames.

>- The names `*Const` were clearer than `*Scalar` in that the `FunctionNode`s don't take the corresponding operands.

I thought the opposite. These `FunctionNode`s takes only scalars, not other constants (ndarrays).

niboshi:
If I can't rename the classes, can I rename `self.value` as `self.scalar`? It took me a while to understand they're actually scalars.

okuta:
Please rebase this because I merged #5439 .

toslunar:
Jenkins, test this please.

PR #5430, opened 2018-10-04
Remove debug print

niboshi:
Just a minor fix.


toslunar:
Jenkins, test this please.

PR #5434, opened 2018-10-05
Include ChainerCV doc

yuyu2172:


takagi:
Would you check the CI failure?

kmaehashi:
Makefiles are not used when publishing to RTD. Another hack may be needed.
https://readthedocs.org/projects/chainer/builds/7918379/

yuyu2172:
jenkins, test this please.

yuyu2172:
jenkins, test this please.

docs/source/conf.py
```diff
@@ -365,6 +368,23 @@ def setup(app):
     app.connect('autodoc-process-docstring', _autodoc_process_docstring)
     app.connect('build-finished', _build_finished)
 
+    chainercv_version = '0.10.0'
```
takagi: How about drawing chainercv version from the installed one? (also document sources?)

yuyu2172: OK. I will change the code to use the version of ChainerCV that is installed.
Is it OK to hardcode version in `setup.py` for `docs` install mode?

yuyu2172:
Can someone review https://github.com/chainer/chainer-test/pull/457?

docs/source/conf.py
takagi: I think it is Okay with leaving a comment as `# see chainer/5434` .

yuyu2172:
jenkins, test this please.

beam2d:
Jenkins, test this please.

beam2d:
Jenkins, test this please.

takagi:
Currently no doctest runs on chainercv repo?

yuyu2172:
No..

With the current implementation, Chainer would need to be able to import the latest released version of ChainerCV. This assumption can fail when functions imported by ChainerCV get deleted in Chainer's master branch.
I talked with beam2d-san offilne, and we decided not to integrate docs. Instead I will add links to ChainerCV's docs hosted here http://chainercv.readthedocs.io/.

docs/source/reference/links.rst
```diff
@@ -126,6 +126,10 @@ VGG Networks
    chainer.links.VGG19Layers
    chainer.links.model.vision.vgg.prepare
 
+.. note::
+   ChainerCV contains implementation of VGG networks as well (i.e.,
+   :class:`chainercv.links.model.vgg.VGG16`).
```
kmaehashi: Could you clarify the difference from Chainer's one? Users may want to know which one is what they want.

yuyu2172: Thank you for the comment.

yuyu2172:
Jenkins, test this please.

takagi:
LGTM!

PR #5445, opened 2018-10-09
Make build_computational_graph accept single output

toslunar:


toslunar:
I'd like to hear your opinion whether to allow it or not.

cf.
- `chainer.grad` is strict.
- `check_backward` is not strict.

okuta:
Please add test.

toslunar:
Added.

okuta:
jenkins, test this please.

okuta:
LGTM!

PR #5425, opened 2018-10-04
Check optional dependencies at runtime

kmaehashi:
Optional dependencies should not be decleared as requirements, as it makes difficult for users to uninstall the optional dependencies.

One of the common pitfall (when user upgraded their CUDA version):

```
pip install cupy-cuda90
pip install chainer
# -> pip recognizes the dependency from chainer to cupy-cuda90

pip uninstall cupy-cuda90
pip install cupy-cuda92

pip install chainer
# -> pip tries to "fix" broken dependency, causing installation of cupy-cuda90
```

PR #5472, opened 2018-10-11
[backport] Check optional dependencies at runtime

takagi:
Backport of #5425

### DIFF
--- a/chainer/_environment_check.py
+++ b/chainer/_environment_check.py
@@ -3,6 +3,9 @@ import sys
 import warnings
 
 import numpy.distutils.system_info
+import pkg_resources
+
+import chainer
 
 
 def _check_python_350():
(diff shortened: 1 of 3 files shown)


takagi:

niboshi:
Another thing to point out is that this implementation depends on `setuptools`, which is not a standard library  (although chainer has already been depending on `setuptools` via `chainer.testing.helper`.)

kmaehashi:
@niboshi 

> Is it ok to backport this PR?
> Change of dependency is not a trivial change.
> Some existing system which (perhaps incorrectly) rely on this dependency will break with this.

I consider this case as a bug rather than just an enhancement, as behavior described in #5425 is obviously unwanted for users.
I'd like to know if you have any concrete use-case that is affected by this change.

> Another thing to point out is that this implementation depends on setuptools, which is not a standard library (although chainer has already been depending on setuptools via chainer.testing.helper.)

I think we can assume use of `setuptools` is safe because it is a common convention to use it for package installation. (note that it's already required in Chainer installation and no one claims it https://github.com/chainer/chainer/blob/7565ecb1a49a396118527fe0db87bdba1d4c4680/setup.py#L7)

niboshi:
> I'd like to know if you have any concrete use-case that is affected by this change.

Yes.
In a team I belong to, CI system stopped working by this change.
In that system CuPy is first installed without dependency followed by Chainer.
The fault was the incorrectly written CI script, but that's what happened to us and could happen to external users as well.

>I think we can assume use of `setuptools` is safe because it is a common convention to use it for package installation.

Discussed with @beam2d and the dependency itself is OK, but we should explicitly write it in `requirements` in `setup.py` as well as documentation.

kmaehashi:
> The fault was the incorrectly written CI script, but that's what happened to us and could happen to external users as well.

OK. In addition, as users are now required to manually update CuPy after updating Chainer, this may be a bit large change for users. I'd avoid backporting this.

> Discussed with @beam2d and the dependency itself is OK, but we should explicitly write it in requirements in setup.py as well as documentation.

I also feel it makes sense, will work on separate PR.

PR #5444, opened 2018-10-09
Define `Variable.__iter__`

toslunar:
Fix #5443.

PR #5477, opened 2018-10-12
Avoid Iterable in CaffeFunction

toslunar:
The current logic of the internals of `L.CaffeFunction` depends on `Variable` is not `Iterable`, which I'd like to avoid.

Split from #5444.

### DIFF
--- a/chainer/links/caffe/caffe_function.py
+++ b/chainer/links/caffe/caffe_function.py
@@ -14,7 +14,6 @@ from chainer.links.connection import linear
 from chainer.links.connection import scale
 from chainer.links.normalization import batch_normalization
 from chainer.utils import argument
-from chainer.utils import collections_abc
 
 
 try:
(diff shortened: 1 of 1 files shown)


chainer/links/caffe/caffe_function.py
```diff
@@ -221,7 +220,7 @@ def forward(self, inputs, outputs, disable=(), **kwargs):
             func = self.forwards[func_name]
             input_vars = tuple(variables[blob] for blob in bottom)
             output_vars = func(*input_vars)
-            if not isinstance(output_vars, collections_abc.Iterable):
+            if not isinstance(output_vars, tuple):
```
kmaehashi: I guess it's safer to use `(tuple, list)` here, considering issues like #1624 exists.

okuta:
Jenkins, test this please.

okuta:
LGTM!

PR #5478, opened 2018-10-12
Implement no-double-backprop version of F.softmax_cross_entropy using FunctionNode

niboshi:
Implements `SoftmaxCrossEntropy` with `FunctionNode`.

This execution path is for single-backward only, so double-backward is not implemented.
This PR does not enhance functionality, but the overhead of `FunctionAdapter` could be reduced.

### DIFF
--- a/chainer/functions/loss/softmax_cross_entropy.py
+++ b/chainer/functions/loss/softmax_cross_entropy.py
@@ -3,7 +3,7 @@ import six
 
 import chainer
 from chainer.backends import cuda
-from chainer import function
+from chainer import function_node
 from chainer.functions.activation import log_softmax
 from chainer.utils import type_check
 from chainer import variable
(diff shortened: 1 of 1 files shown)


chainer/functions/loss/softmax_cross_entropy.py
```diff
@@ -250,6 +269,11 @@ def backward_gpu(self, inputs, grad_outputs):
 
         return gx, None
 
+    def backward(self, input_indexes, grad_outputs):
+        assert False, (
+            'This execution path of F.softmax_cross_entropy does not support '
+            'double-backprop. There must be a bug.')
```
beam2d: Better to advise the user to use `enable_double_backprop=True`.

niboshi:
PTAL

beam2d:
Jenkins, test this please.

beam2d:
LGTM

Issue #5474, opened 2018-10-11
MPICommunicatorBase.allreduce does not work with 0-d array or cupy


Hakuyume:
chainer/cupy == 5.0.0rc1

```python
import chainermn
import cupy
import numpy as np

comm = chainermn.create_communicator()

# numpy scalar -> NG
# TypeError: 'numpy.float64' object cannot be interpreted as an integer
comm.allreduce(np.array(1))

# numpy array -> OK
comm.allreduce(np.array((2, 3)))

# cupy scalar -> NG
# TypeError: 'numpy.float64' object cannot be interpreted as an integer
comm.allreduce(cupy.array(1))

# cupy array -> NG
# AttributeError: 'tuple' object has no attribute 'reshape'
comm.allreduce(cupy.array((2, 3)))
```

Hakuyume:
In the case of cupy array, `_memory_utility.array_to_buffer_object` returns a tuple but `allreduce` calls `reshape` on it.
https://github.com/chainer/chainer/blob/master/chainermn/communicators/_memory_utility.py#L121
https://github.com/chainer/chainer/blob/master/chainermn/communicators/mpi_communicator_base.py#L492-L496

kuenishi:
@Hakuyume For a scaler use case, I think you can use just `allreduce_obj` instead. I saw your code in https://github.com/chainer/chainercv/pull/705 .

```python
        positive = gt_mb_labels.array > 0
        n_positive = xp.array(positive.sum())
        if comm:
            n_positive = cuda.to_cpu(n_positive).reshape(-1)
            n_positive = comm.allreduce(n_positive) / comm.size
```

can I think be replaced with

```python
        positive = gt_mb_labels.array > 0
        if comm:
            n_positive = comm.allreduce_obj(positive.sum()) / comm.size
```

Hakuyume:
@kuenishi Thank you! `allreduce_obj` worked well for my case.
However, I will leave this PR open since `allreduce` still has some problem.

kuenishi:
Dekita

kuenishi:
merge to chainer/chainer will be via #5535 .

kuenishi:
#5535 was merged.

PR #5479, opened 2018-10-12
Return right pointer from communicator allreduce()

kuenishi:
Partial fix for #5474 .


toslunar:
We should not create a PR branch in chainer/chainer.

PR #5481, opened 2018-10-12
 Fix errors on 0-d array input to Communicator APIs

kuenishi:
Complementary fix for #5474 .


kuenishi:
I've tested both on CPUs and on GPUs. Once it passes all CIs It'll be ready for review.

PR #5482, opened 2018-10-13
Fix `reporter.Summary` float value deserialization

### LINKED ISSUE #5401: chainer.reporter.Summary does not properly deserialize float values from NpzDeserializer
Hi, I've noticed that ```chainer.reporter.Summary.serialize``` convert all float values into int if it is called with ```chainer.serializers.NpzDeserializer``` because the deserializer convert the deserialized values into the same types as the original values and ```._x``` and ```._x2``` are initialized as ```0```, which are integer values.

Apparently, this problem will be fixed if you make them initialized as ```0.0```, but I'm not sure whether it is applicable. What do you think?

* Conditions
  - Platform: Linux-4.4.0-97-generic-x86_64-with-Ubuntu-16.04-xenial
  - Chainer: 5.0.0rc1
  - NumPy: 1.15.2
  - CuPy:
    - CuPy Version          : 5.0.0rc1
    - CUDA Root             : /usr/local/cuda
    - CUDA Build Version    : 9000
    - CUDA Driver Version   : 9000
    - CUDA Runtime Version  : 9000
    - cuDNN Build Version   : 7300
    - cuDNN Version         : 7300
    - NCCL Build Version    : 2213

* Code to reproduce
```python
import tempfile

import chainer

summary = chainer.reporter.Summary()
summary.add(0.5)
print('before:', summary.compute_mean())
with tempfile.NamedTemporaryFile() as npz:
    chainer.serializers.save_npz(npz, summary)
    npz.flush()
    summary = chainer.reporter.Summary()
    chainer.serializers.load_npz(npz.name, summary)
print('after:', summary.compute_mean())
```
* Output
```
before: 0.5
after: 0.0
```

niboshi:
Fixes #5401

### DIFF
--- a/tests/chainer_tests/test_reporter.py
+++ b/tests/chainer_tests/test_reporter.py
@@ -10,6 +10,7 @@ from chainer import configuration
 from chainer import functions
 from chainer import testing
 from chainer.testing import attr
+from chainer.testing import backend
 
 
 class TestReporter(unittest.TestCase):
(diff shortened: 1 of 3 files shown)


tests/chainer_tests/test_reporter.py
```diff
@@ -163,27 +164,17 @@ def test_report_scope(self):
...
-        testing.assert_allclose(mean, numpy.array(-0.5, 'f'))
-        testing.assert_allclose(std, numpy.array(1.5, 'f'))
-
-    @attr.gpu
-    def test_cupy(self):
-        xp = cuda.cupy
-        self.summary.add(xp.array(1, 'f'))
-        self.summary.add(xp.array(-2, 'f'))
+    def test_basic(self, backend_config):
+        self.summary.add(backend_config.get_array(numpy.array(1, 'f')))
```
takagi: I think that `BackendConfig.get_array` is too general, at least for current backend candidates of Chainer. How about making `BackendConfig` have `xp` property that is either `numpy` or `cupy`?

takagi: Okay.

takagi:
Jenkins, test this please.

takagi:
LGTM! Waiting for CI finish.

takagi:
Yes, I will merge this after code freezing period for the release tomorrow.

PR #5486, opened 2018-10-14
Relax tolerance of check_double_backward test

niboshi:
This test sometimes fails due to the strict tolerance.


kmaehashi:
Jenkins, test this please.

PR #5504, opened 2018-10-16
Fix dead fragments to CuPy docs

kmaehashi:


toslunar:
Why don't you backport this PR?

kmaehashi:
Yes this needs to be backported (please exclude ChainerMN one).

PR #5507, opened 2018-10-17
Change default scales to 1.0

toslunar:
Default scales should be `1.0`.

Just in case, I added a config option to make them legacy.


toslunar:
The tests are too weak.  Merge #5511 first.

kmaehashi:
We should discuss if it is ok to introduce this change immediately. Raising a warning when `scale` is not specified is one idea.

toslunar:
Added warnings.

PR #5508, opened 2018-10-17
Fix SoftmaxCrossEntropy grad function node

beam2d:
Fix a bug in #5478 (sorry, it was my review fault). FunctionNode.forward should not return None as an output, so I fixed it. It should fix the current doctest failure in master.


niboshi:
LGTM
jenkins, test this please

toslunar:
LGTM, too.  Let's merge it before the daily tests start.

PR #5509, opened 2018-10-17
Unify implementations on computing size from shape

toslunar:


chainer/utils/array.py
```diff
@@ -39,6 +41,10 @@ def empty_like(x):
         return numpy.empty_like(x)
 
 
+def size_of_shape(shape):
+    return functools.reduce(operator.mul, shape, 1)
```
okuta: Is this first?
Cloud you check speed?
```
size = 1
for i in shape:
    size *= i
return size
```

okuta:
jenkins, test this please.

kuenishi:
Good catch! A lot of same idioms occur at https://github.com/chainer/chainer/blob/master/chainermn/communicators/mpi_communicator_base.py , can you also apply this changes to them?

toslunar:
@kuenishi Yes, I will.

okuta:
OK, I will merge this PR.

okuta:
LGTM!

PR #5512, opened 2018-10-18
Modify ChainerMN bcast to retain delegate variable

levelfour:
## Problem

The current implementation of `chainermn.functions.bcast` fails to retain delegate variables, which means non-root processes fails to backprop their gradients in the middle if non-connected components are defined as models.

## Example: gather -> bcast

After calling `chainermn.functions.gather`, models defined on non-root processes do not retain the computational graph to the succeeding `bcast`.
Please refer to the test `check_gather_bcast` added in this PR.

## Solution

The cause is that `Bcast.backward` returns `None` for certain cases.
It's modified to return a dummy `chainer.Variable`:
```python
return tuple([xp.zeros(x.shape, dtype=xp.float32) for x in inputs])
```
and delegate variables are carefully handled.

### DIFF
--- a/chainermn/functions/collective_communication.py
+++ b/chainermn/functions/collective_communication.py
@@ -71,15 +71,18 @@ class Bcast(chainer.Function):
         return self.comm.bcast(x, self.root),
 
     def backward(self, inputs, grad_outputs):
+        xp = backend.get_array_module(*grad_outputs)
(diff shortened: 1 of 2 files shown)


shu65:
Rebased at https://github.com/chainer/chainer/pull/6155

PR #5531, opened 2018-10-19
Chainer v5 requires CuPy v5

kmaehashi:


toslunar:
Jenkins, test this please.

PR #5532, opened 2018-10-19
Require setuptools and add docs for #5425

kmaehashi:
* Upgrade guide for #5425
* Add requirement to setuptools as discussed in #5472


takagi:
Jenkins, test this please.

PR #5536, opened 2018-10-19
Fix `dump_graph` not to memory leak

hitsgub:
## Problem
The current version of `dump_graph` set a flag `keep_graph_on_report` in `initializer()` even when restored from a snapshot.
It will keep every variables in each iterations and it cause a memory leak problem.

## Solution
Changing the extension definition from using decorator to direct class definition.
I added the new attribute `self._flag_called` in order to run the extension surely only once.
The flag is the execution condition for `initialize` and `__call__`,  disabled in `trigger`, and serialized.


hitsgub:
Sorry I made some mistakes in process of opening PR

PR #5538, opened 2018-10-19
Fix `dump_graph` not to memory leak

hitsgub:
## Problem
The current version of `dump_graph` set a flag `keep_graph_on_report` in `initializer()` even when restored from a snapshot.
It will keep every variables in each iterations and it cause a memory leak problem.

## Solution
Changing the extension definition from using decorator to direct class definition.
I added the new attribute `self._flag_called` in order to run the extension surely only once.
The flag is the execution condition for `initialize` and `__call__`,  disabled in `trigger`, and serialized.

### DIFF
--- a/chainer/training/extensions/computational_graph.py
+++ b/chainer/training/extensions/computational_graph.py
@@ -10,9 +10,9 @@ _var_style = {'shape': 'octagon', 'fillcolor': '#E0E0E0', 'style': 'filled'}
 _func_style = {'shape': 'record', 'fillcolor': '#6495ED', 'style': 'filled'}
 
 
-def dump_graph(root_name, out_name='cg.dot',
-               variable_style=None, function_style=None):
-    """Returns a trainer extension to dump a computational graph.
+class dump_graph(extension.Extension):
+
+    """Trainer extension to dump a computational graph.
 
     This extension dumps a computational graph. The graph is output in DOT
     language.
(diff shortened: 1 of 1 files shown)


chainer/training/extensions/computational_graph.py
```diff
@@ -60,37 +60,45 @@ def dump_graph(root_name, out_name='cg.dot',
...
+    def __init__(self, root_name, out_name='cg.dot',
+                 variable_style=None, function_style=None):
+        self._root_name = root_name
+        self._out_name = out_name
+        self._variable_style = variable_style or _var_style
+        self._function_style = function_style or _func_style
+        self._original_flag = None
+        self._flag_called = False
+
+    def initialize(self, _):
```
crcrpar: Could you tell me why `_` exists here?

chainer/training/extensions/computational_graph.py
```diff
@@ -60,37 +60,45 @@ def dump_graph(root_name, out_name='cg.dot',
...
-        original_flag[0] = configuration.config.keep_graph_on_report
-        configuration.config.keep_graph_on_report = True
 
-    @extension.make_extension(trigger=trigger, initializer=initializer)
-    def dump_graph(trainer):
+    def __init__(self, root_name, out_name='cg.dot',
+                 variable_style=None, function_style=None):
+        self._root_name = root_name
+        self._out_name = out_name
+        self._variable_style = variable_style or _var_style
```
crcrpar:
```python
If variable_style is None:
    variable_style = _var_style
self._variable_style = variable_style
```
or using kwargs_parser is better for me.

chainer/training/extensions/computational_graph.py
```diff
@@ -60,37 +60,45 @@ def dump_graph(root_name, out_name='cg.dot',
...
+        self._function_style = function_style or _func_style
```
crcrpar: ditto

chainer/training/extensions/computational_graph.py
hitsgub: The methods are called by https://github.com/chainer/chainer/blob/master/chainer/training/trainer.py#L304 with one argument.
Do you recommend fix the name to `trainer`?

chainer/training/extensions/computational_graph.py
hitsgub: Thank you for your recommendation, I will fix here.

chainer/training/extensions/computational_graph.py
crcrpar: Thank you, I just didn't know how `initialize` is invoked.

> Do you recommend fix the name to trainer?
I think it's up to core members.

chainer/training/extensions/computational_graph.py
```diff
@@ -10,9 +10,9 @@
 _func_style = {'shape': 'record', 'fillcolor': '#6495ED', 'style': 'filled'}
 
 
-def dump_graph(root_name, out_name='cg.dot',
-               variable_style=None, function_style=None):
-    """Returns a trainer extension to dump a computational graph.
+class dump_graph(extension.Extension):
```
beam2d: How about using a class-like name here (`DumpGraph`) and leave `dump_graph` a thin function that constructs and returns it? It may be also good to set `default_name = 'dump_graph'` as a class attribute for backward compatibility.

chainer/training/extensions/computational_graph.py
beam2d: It's safer to use `trainer` as the argument name here because, in Python, the argument name is a part of API (due to keyword argument call).

chainer/training/extensions/computational_graph.py
hitsgub: Thank you, I renamed it according to naming rule of `class` and add `default_name` attribute.

chainer/training/extensions/computational_graph.py
hitsgub: OK, I reflected it.

chainer/training/extensions/__init__.py
```diff
@@ -1,7 +1,7 @@
 # import classes and functions
 from chainer.training.extensions._snapshot import snapshot  # NOQA
 from chainer.training.extensions._snapshot import snapshot_object  # NOQA
-from chainer.training.extensions.computational_graph import dump_graph  # NOQA
+from chainer.training.extensions.computational_graph import DumpGraph  # NOQA
```
beam2d: Could you leave the old name for backward compatibility? It's ok to define it as an alias to `DumpGraph`.

hitsgub: Thank you.
I added the areas for backward compatibility in last commit.

chainer/training/extensions/__init__.py
```diff
@@ -20,3 +20,6 @@
 from chainer.training.extensions.value_observation import observe_value  # NOQA
 from chainer.training.extensions.variable_statistics_plot import VariableStatisticsPlot  # NOQA
 from chainer.training.extensions.warmup_shift import WarmupShift  # NOQA
+
+# Aliase
```
toslunar: Fix typo:
```suggestion
# Alias
```

okuta:
LGTM!

kmaehashi:
For record, this issue was originally reported in Chainer-JP Slack (https://chainer-jp.slack.com/archives/CBT83K011/p1539680156000100?thread_ts=1535179361.000100&cid=CBT83K011)

kmaehashi:
In the offline discussion in #6079, we decided to backport this PR as this should be considered as bug rather than feature.

PR #5539, opened 2018-10-19
Unchain all variables after running extensions

hitsgub:
When the `configuration.config.keep_graph_on_report` flag is turned on, the computational graph created by the updater is kept *and seems not to be released* in current implementation.
I added the `unchain_backward()` after all extensions are executed.

### DIFF
--- a/chainer/training/extensions/variable_unchain.py
+++ b/chainer/training/extensions/variable_unchain.py
@@ -8,7 +8,9 @@ class unchain_variables(extension.Extension):
     """Trainer extension to unchain all comptational graphs.
 
     This extenstion unchains all comptational graphs after all extensions are
-    run to avoid any memory problem.
+    run to release memory and to avoid memory leak.
+    This extension can be used as a last resort when there is an extension that
+    use a variable graph and cannot release the graph in itself.
     It observes the previous ``chainer.config.keep_graph_on_report`` flag.
     The extension is triggered when the flag is turned on.
 
--- a/chainer/training/extensions/variable_unchain.py
+++ b/chainer/training/extensions/variable_unchain.py
@@ -10,7 +10,7 @@ class unchain_variables(extension.Extension):
     This extenstion unchains all comptational graphs after all extensions are
     run to avoid any memory problem.
     It observes the previous ``chainer.config.keep_graph_on_report`` flag.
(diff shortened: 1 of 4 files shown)


toslunar:
Why don't you implement this feature as an `Extension`?

hitsgub:
Thank you for your recommendation, I just didn't reach this idea.
I will re-implement as an `Extension` and commit.

chainer/training/extensions/variable_unchain.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def trigger(self, _):
+        flag = self._prev_flag
+        self._prev_flag = configuration.config.keep_graph_on_report
+        return flag
+
+    def __call__(self, trainer):
+        for var in six.itervalues(trainer.observation):
+            if isinstance(var, variable.Variable):
+                var.unchain_backward()
```
beam2d: A minor point: isn't it enough to call `unchain` instead of `unchain_backward` here, if the motivation is to release the graph kept by the observation?

toslunar: I took the feature as a last resort when an extension has some memory issue that a user cannot debug by themselves.  `unchain_backward` looks suitable here to me just because it's more powerful.

chainer/training/extensions/variable_unchain.py
```diff
@@ -0,0 +1,32 @@
...
+from chainer.training import extension
+from chainer import variable
+import six
+
+
+class unchain_variables(extension.Extension):
+    """Trainer extension to unchain all comptational graphs.
+
+    This extenstion unchain all comptational graphs after all extensions is
+    run to avoid any memory problem.
```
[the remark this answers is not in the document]
hitsgub: Thank you, please show new commit.

chainer/training/extensions/variable_unchain.py
hitsgub: As you say, this PR is for some memory issue.
I added this mention to the explanation, thank you.

beam2d:
Jenkins, test this please.

toslunar:

beam2d:
LGTM

beam2d:
Thanks!

PR #5548, opened 2018-10-22
Fix `D.Uniform.log_prob` to avoid returning -inf at boundary

toslunar:
I found `dist.log_prob(dist.sample())` is returning `-inf`.


toslunar:
Jenkins, test this please.

PR #5553, opened 2018-10-23
Un-deprecate `Link.add_param` and `Link.add_link`

### LINKED ISSUE #4106: A concern about deprecating link-adding interfaces except init_scope
From Chainer v2, `init_scope` is introduced as a new interface to add parameters and child links. While I think this interface is nice in a sense that it helps IDEs and users understand they can be used as attributes, I have a concern about deprecating the old interfaces: `Chain.__call__` and `add_link`.

Let's say I want to wrap two links to make a single link.
Old interface:
```
chain = chainer.Chain(a=LinkA(), b=LinkB())
```

New interface:
```
chain = chainer.Chain()
with chain.init_scope():
    chain.a = LinkA()
    chain.b = LinkB()
```
It's lengthy and unnecessarily stateful.

Another example is to construct a Chain when child links and names are determined at runtime. I guess in this situation IDEs cannot infer attributes anyway.
Old interface:
```
chain = chainer.Chain(**children)  # children is a dict holding child links by their names
# or
chain = chainer.Chain()
for name, link in children:
    chain.add_link(name, link)
```

New interface:
```
chain = chainer.Chain()
with chain.init_scope():
    for name, link in children:
        setattr(chain, name, link)
```
I feel `setattr` is less readable to humans than `add_link`.

I guess Chain is mainly designed to be used via inheritance, not by direct instantiation, but defining a new class for just wrapping other links seems overkill to me. I want it to work as a container as well.

I don't mean Chainer should stop deprecating the old interfaces, as I guess the developer team already had discussions on it. I just want to show my concern and hear opinions on it from other users as well as from the developer team.

kmaehashi:
Fix #4106.

### DIFF
--- a/chainer/link.py
+++ b/chainer/link.py
@@ -268,26 +268,6 @@ class Link(object):
                   initializer=None):
         """Registers a parameter to the link.
 
-        .. deprecated:: v2.0.0
-
-           Assign a :class:`~chainer.Parameter` object directly to an
-           attribute within :meth:`~chainer.Link.init_scope` instead.
-           For example, the following code
-
-           .. code-block:: python
-
-               link.add_param('W', shape=(5, 3))
-
-           can be replaced by the following assignment.
-
-           .. code-block:: python
-
-               with link.init_scope():
-                   link.W = chainer.Parameter(None, (5, 3))
-
-           The latter is easier for IDEs to keep track of the attribute's
-           type.
-
         Args:
             name (str): Name of the parameter. This name is also used as the
                 attribute name.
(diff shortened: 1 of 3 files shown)


beam2d:
Could you de-deprecate the arguments of `Link.__init__` and `Chain.__init__` as well? You also need to fix the tests that check these deprecation warnings.

kmaehashi:
Jenkins, test this please.
Jenkins, test this please.

beam2d:
Jenkins, test this please.

beam2d:
LGTM

PR #5557, opened 2018-10-23
Fix BN+F.forget

### LINKED ISSUE #5555: The behavior of F.forget is not appropriate for BatchNormalization
The combination of `F.forget` and `BatchNormalization` will cause wrong computation result.
`BatchNormalization` link updates its internal statistics (avg_mean and avg_var) every time the forward computation is called during training time, on the other hand, `F.forget` will call forward computation twice.  So the combination of them will lead to wrong statistics result.

Example:
```py
import chainer
import chainer.functions as F
import chainer.links as L

import numpy as np


class Model(chainer.Chain):

    def __init__(self):
        super().__init__()
        with self.init_scope():
            self.bn1 = L.BatchNormalization(5)

    def forward(self, x):
        h = self.bn1(x)
        # h = F.forget(self.bn1, x)
        return h


np.random.seed(666)
mlp = Model()

x = np.random.uniform(size=(2, 5)).astype('f')
y = mlp(x)
F.sum(y).backward()

with chainer.using_config('train', False):
    # test time
    y = mlp(x)
    print(y)
```

Output (without F.forget):
```
variable([[ 0.6917034   0.8193421   0.667606    0.71579635  0.9209751 ]
          [-0.02388343  0.36777574  0.01308255  0.06104017  0.45613596]])
```
Output (with F.forget):
```
variable([[ 0.6842567   0.79659545  0.6600341   0.70522314  0.89286864]
          [-0.05953163  0.32327673 -0.02182698  0.02312531  0.4058005 ]])
```

This unmatched behavior is not described in document.

ir5:
Fix #5555 by simply adding new global config.

TODO:
- [x] Write test
- [x] Fix F.BatchReNormalization as well?
- [x] Write document somewhere

### DIFF
--- a/chainer/links/normalization/batch_normalization.py
+++ b/chainer/links/normalization/batch_normalization.py
@@ -304,9 +304,7 @@ class BatchNormalization(link.Link):
                 # Do not update statistics when extra forward computation is
                 # called.
                 if finetune:
-                    # Revert the count
-                    # (TODO: mkusumoto) Test finetuning in recomputation.
-                    self.N -= 1
+                    self.N -= 1  # Revert the count
                 avg_mean = None
                 avg_var = None
 
--- a/chainer/links/normalization/batch_renormalization.py
+++ b/chainer/links/normalization/batch_renormalization.py
@@ -66,6 +66,8 @@ class BatchRenormalization(BatchNormalization):
             if chainer.config.in_recomputing:
                 # Do not update statistics when extra forward computation is
                 # called.
+                if finetune:
+                    self.N -= 1  # Revert the count
                 avg_mean = self.xp.zeros_like(self.avg_mean)
                 avg_var = self.xp.zeros_like(self.avg_var)
 
--- a/tests/chainer_tests/functions_tests/util_tests/test_forget.py
+++ b/tests/chainer_tests/functions_tests/util_tests/test_forget.py
@@ -1,3 +1,4 @@
+import functools
 import unittest
 
 import numpy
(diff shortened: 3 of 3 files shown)


kmaehashi:
Jenkins, test this please.

chainer/__init__.py
```diff
@@ -169,6 +169,7 @@ def is_arrays_compatible(arrays):
     raise TypeError('incorrect dtype name in CHAINER_DTYPE: "{}". '
                     'Only float16/32/64 are allowed.'.format(_chainer_dtype))
 global_config.dtype = numpy.dtype(_chainer_dtype)
+global_config.recompute = False
```
kmaehashi: * The configuration key `recompute` sounds too general and a bit confusing to users. How about naming it like `in_recomputation`?
* Could you add the description like this to https://docs.chainer.org/en/latest/reference/configuration.html#configuration-keys?

```rst
This flag is automatically set by :func:`chainer.functions.forget` and not intended to be changed by users.
You can use this flag when implementing your own Link to avoid updating the internal states during recomputation done by `F.forget`.
See the documentation of :func:`chainer.functions.forget` for details.
```

ir5: Thank you for review. I fixed code.

ir5:
I fixed the problems, so this PR is ready for review.

ir5:
OK, I'll check.

ir5:
I fixed when `self.N` is incremented.
However, it seems that this problem occurs only when `BatchNormalization` takes `finetune=True` argument in `**kwargs`, which we cannot pass to the arguments of `F.forget`. Further, current `F.forget` does not take boolean arguments. So, while I think I have fixed the problem, I didn't write test for it.

toslunar:
Could you test it with something like `F.forget(functools.partial(self.link, finetune=True), x)`?

ir5:
Sounds nice. I added test and fixed bug in `BatchRenormalization`.

ir5:
Why Travis CI build failed? It indicates flake8 error at strange point...

toslunar:
I restarted Travis after #5764 is merged.

Jenkins, test this please.

toslunar:
Jenkins, test this please.

kmaehashi:
After discussion, concluded that this issue should be rather considered as bug instead of enhancement.

PR #5565, opened 2018-10-23
Add extension to be executed only once

hitsgub:
Basic `trainer.Extension` class is implemented for iterated extensions.
However, sometimes we need the extensions to be executed only once.
(There is only one such official extension `dump_graph`.)

This PR provide the base class for such extensions.

### DIFF
--- a/chainer/training/triggers/once_trigger.py
+++ b/chainer/training/triggers/once_trigger.py
@@ -4,15 +4,15 @@ class OnceTrigger(object):
 
     This trigger accepts only once at starting point of the iteration. There
     are two ways to specify the starting point: only starting point in whole
-    iteration or recalled when training resumed.
+    iteration or called again when training resumed.
 
     Args:
-        recall_on_resume (bool): Whether the extension is recalled or not when
-            restored from a snapshot. It is set to ``False`` by default.
+        call_on_resume (bool): Whether the extension is called again or not
+            when restored from a snapshot. It is set to ``False`` by default.
     """
 
-    def __init__(self, recall_on_resume=False):
-        self._recall_on_resume = recall_on_resume
+    def __init__(self, call_on_resume=False):
+        self._call_on_resume = call_on_resume
         self._flag_called = False
 
     def trigger(self, trainer):
@@ -32,5 +32,5 @@ class OnceTrigger(object):
         return self._flag_called
 
     def serialize(self, serializer):
-        if not self._recall_on_resume:
+        if not self._call_on_resume:
             self._flag_called = serializer('_flag_called', self._flag_called)
--- a/chainer/training/triggers/once_trigger.py
+++ b/chainer/training/triggers/once_trigger.py
@@ -10,10 +10,6 @@ class OnceTrigger(object):
         call_on_resume (bool): Whether the extension is called again or not
             when restored from a snapshot. It is set to ``False`` by default.
 
-    Attributes:
-        finished (bool): Flag that triggered when this trigger called once.
-            The flag helps decision to call `Extension.initialize` or not
-            in `trainer`.
     """
 
     def __init__(self, call_on_resume=False):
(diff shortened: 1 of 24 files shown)


hitsgub:
> Sorry for the late response. I thought this feature should be covered by a trigger instead of an Extension subclass. What do you think about introducing such a trigger, say `TriggerOnce`?

Thanks to your recommendation. I re-implemented this feature as trigger instead of extension, and use name `OnceTrigger` for consistency with other triggers.

chainer/training/triggers/once_trigger.py
```diff
@@ -0,0 +1,36 @@
+class OnceTrigger(object):
+
+    """Trigger based on the starting point of the iteration.
+
+    This trigger accepts only once at starting point of the iteration. There
+    are two ways to specify the starting point: only starting point in whole
+    iteration or recalled when training resumed.
+
+    Args:
+        recall_on_resume (bool): Whether the extension is recalled or not when
```
beam2d: How about naming it `call_on_resume`? I'm wondering if "recall" could mean "call again".

chainer/training/triggers/once_trigger.py
```diff
@@ -0,0 +1,36 @@
...
+
+            If this flag is exist and set `True`, `Extension.initialize` is
+            skipped.
+
+        """
+        return self._flag_called
+
+    def serialize(self, serializer):
+        if not self._recall_on_resume:
+            self._flag_called = serializer('_flag_called', self._flag_called)
```
beam2d: Can we also support the case that `recall_on_resume` flag changes after suspend/resume?

chainer/training/triggers/once_trigger.py
```diff
@@ -0,0 +1,36 @@
...
+        self._flag_called = False
+
+    def trigger(self, trainer):
+        if self._flag_called:
+            return False
+        self._flag_called = True
+        return True
+
+    @property
+    def skip_initialize(self):
```
beam2d: I understood this feature would be useful, but I feel it inconsistent with the existing triggers' behavior. Extensions with other triggers initialize even if they will not be triggered after resume. Do you think this feature is mandatory for `OnceTrigger`?

chainer/training/triggers/once_trigger.py
hitsgub: OK, I will change this argument.

chainer/training/triggers/once_trigger.py
hitsgub: Certainly, some other triggers can know easily that they will never triggered, such as `manual_schedule_trigger`.
And I wonder if there may be other uses of this information that they will never triggered.
How about adding property `never_triggered` to trigger objects, instead of `skip_initialize`?
- `trainer` observe `never_triggered` for initialize the extension or not
- Default value of `never_triggered` is `False` for almost triggers, such as iterative triggers.

hitsgub: Ah, but now I think that the name `never_triggered` is confusing it means whether "the trigger will be never triggered" or "the trigger have never been triggered".
I will think better name for this property.

chainer/training/triggers/once_trigger.py
hitsgub: Thanks, I rewrite the code for support this case.

chainer/training/triggers/once_trigger.py
hitsgub: Last commit adopts above property, `finished`.

beam2d: The property `finished` looks good. I would like to take this approach, but maybe it is safer to work on that in another PR. I want to keep this PR simply covering the new "once" trigger, and let the change on the base trigger API be separated into a dedicated issue/PR.

hitsgub: OK, I agree with you.
Now I deleted `finished` property, and I will make another PR for this property.

hitsgub:
Sorry for late response.
I updated training.rst.

beam2d:
Looks good. Could you add simple tests for this trigger? (sorry for mentioning it late)[

beam2d:
Jenkins, test this please.

beam2d:
The Jenkins failure looks irrelevant to this PR. LGTM!

PR #5586, opened 2018-10-26
Fix docstring of `F.forget`

### LINKED ISSUE #5528: Docstring of `functions.forget` is incorrect as `+` doesn't retain inputs anymore
The docstring says that `(x + y) + x` retains the immediate variable holding `x + y`. 

```
       Let ``f`` be a function defined as:
       >>> def f(a, b):
       ...   return a + b + a
       and, ``x`` and ``y`` be :class:`~chainer.Variable`\\ s:
       >>> x = chainer.Variable(np.random.uniform(-1, 1, 5).astype(np.float32))
       >>> y = chainer.Variable(np.random.uniform(-1, 1, 5).astype(np.float32))
       When ``z`` is calculated as ``z = f(x, y)``, its intermediate result
       ``x + y`` is stored in memory. Instead, if you call ``f`` with
       ``F.forget``:
       >>> z = F.forget(f, x, y)
       intermediate ``x + y`` is forgotten.
```

But this isn't true for new-style function of `+`, because addition don't requires book-kept inputs for backpropagation.

I checked the behavior by the following script, which traverses retained variables.

```python
import chainer
import chainer.functions as F
import numpy as np 


def f(a, b):
    return (a + b) + a


def recur_check_vars(v, x, y):
    creator = v.creator_node
    if creator is None:
        return
    for pnode in creator.inputs:
        p = pnode.get_variable()
        assert p.data is None or p is x or p is y
        print(p)
        recur_check_vars(p, x, y) 


def main():
    x = chainer.Variable(np.random.uniform(-1, 1, 5).astype(np.float32))
    y = chainer.Variable(np.random.uniform(-1, 1, 5).astype(np.float32))
    print(x)
    print(y)
    print()
    z = f(x, y) 
    recur_check_vars(z, x, y) 


if __name__ == '__main__':
    main()
```

The script doesn't fail, and the output is as follows. We can see that`x + y` is discarded. Living variables `x` and `y` are retrieved, as each `VariableNode` instance has a weakref to the corresponding variable.

```
variable([-0.7699733  -0.50523347 -0.20869003 -0.7912116   0.92058474])
variable([ 0.58832335 -0.06183117  0.1939743   0.9021316  -0.19973369])

variable(None)
variable([-0.7699733  -0.50523347 -0.20869003 -0.7912116   0.92058474])
variable([ 0.58832335 -0.06183117  0.1939743   0.9021316  -0.19973369])
variable([-0.7699733  -0.50523347 -0.20869003 -0.7912116   0.92058474])
```

fiarabbit:
fixed #5528

### DIFF
--- a/chainer/functions/util/forget.py
+++ b/chainer/functions/util/forget.py
@@ -89,7 +89,7 @@ def forget(func, *xs):
        Let ``f`` be a function defined as:
 
        >>> def f(a, b):
-       ...   return a + b + a
+       ...   return (a + b) * a
 
        and, ``x`` and ``y`` be :class:`~chainer.Variable`\\ s:
 


kmaehashi:
Jenkins, test this please.

kmaehashi:

Issue #5590, opened 2018-10-28
text_classification example fails on Python 3


koreyou:
# Problem description

Running text_classification example fails with dbpedia dataset.

```
# in examples/text_classification
$ python train_text_classifier.py --dataset  dbpedia
{
  "batchsize": 64,
  "epoch": 30,
  "gpu": -1,
  "out": "result",
  "unit": 300,
  "layer": 1,
  "dropout": 0.4,
  "dataset": "dbpedia",
  "model": "cnn",
  "char_based": false
}
read dbpedia
Traceback (most recent call last):
  File "train_text_classifier.py", line 138, in <module>
    main()
  File "train_text_classifier.py", line 50, in main
    char_based=args.char_based)
  File "/home/koreyou/work/chainer-tmp/examples/text_classification/text_datasets.py", line 45, in get_dbpedia
    train = read_dbpedia(tf, 'train', shrink=shrink, char_based=char_based)
  File "/home/koreyou/work/chainer-tmp/examples/text_classification/text_datasets.py", line 32, in read_dbpedia
    for i, (label, title, text) in enumerate(csv.reader(f)):
_csv.Error: iterator should return strings, not bytes (did you open the file in text mode?)
```

## Environment

```
Platform: Linux-4.4.0-137-generic-x86_64-with-debian-stretch-sid
Chainer: 5.0.0
NumPy: 1.15.3
CuPy: Not Available
iDeep: Not Available
```

I am running on Python 3.6.4.

# Reason for failure

This does not reproduce on python 2.7.12, so it is due to str-bytes problem.

I will send a merge request to fix this problem.

PR #5591, opened 2018-10-28
Fix text_classification example fails on Python 3 (#5590)

koreyou:
Fix text_classification example fails on Python 3 #5590

I have confirmed that it runs fine on Python 2.7.12, 3.4.3, 3.6.4.

### DIFF
--- a/examples/text_classification/text_datasets.py
+++ b/examples/text_classification/text_datasets.py
@@ -31,7 +31,7 @@ def read_dbpedia(tf, split, shrink=1, char_based=False):
     dataset = []
     f = tf.extractfile('dbpedia_csv/{}.csv'.format(split))
     if sys.version_info > (3, 0):
-         f = io.TextIOWrapper(f, encoding='utf-8')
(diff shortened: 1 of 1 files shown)


examples/text_classification/text_datasets.py
```diff
@@ -29,6 +30,8 @@ def download_dbpedia():
 def read_dbpedia(tf, split, shrink=1, char_based=False):
     dataset = []
     f = tf.extractfile('dbpedia_csv/{}.csv'.format(split))
+    if sys.version_info > (3, 0):
+         f = io.TextIOWrapper(f, encoding='utf-8')
```
takagi: Would you fix the indentation of this line?
```suggestion
        f = io.TextIOWrapper(f, encoding='utf-8')
```

koreyou:
I've pushed the fix.
Can you review the change as is, or do I have to make another PR?

takagi:
LGTM! I'll add a test for text_classification example in another PR.

PR #5595, opened 2018-10-29
Support soft target in softmax_cross_entropy

anaruse:
This PR aims to support "soft target" in softmax_cross_entropy.

Current softmax_cross_entropy implementation support "hard target" but does not support "soft target" that is becoming popular as a method to mitigate over-fitting. This PR allows users to use "soft target" in softmax_cross_entropy as follows.

    soft_target_loss = F.softmax_cross_entropy(x, t, soft_target=soft_target)

The soft target loss is KL divergence.


anaruse:
Or, would it be better to implement this as different function such as softmax_kl_divergence?

anaruse:
I've fixed the PR, so that it uses argument `t` for both hard and soft target. Whether it is hard or soft target is determined by ndim and shape of `x` and `t`.

tests/chainer_tests/functions_tests/loss_tests/test_softmax_cross_entropy.py
```diff
@@ -572,4 +572,107 @@ def test_consistency_gpu_never(self):
...
+    'reduce': ['mean', 'no'],
+})))
+class TestSoftTargetCompareToHard(BaseSoftTarget, unittest.TestCase):
+
+    def setUp(self):
+        BaseSoftTarget.setUp(self)
+        t_hard_shape = (self.nb,) + self.shape[1:]
+        self.t_hard = numpy.random.randint(
+            0, self.shape[0], t_hard_shape).astype(numpy.int32)
+        t = numpy.zeros(numpy.prod(self.x.shape)).astype(self.dtype)
```
beam2d:
```suggestion
        t = numpy.zeros(self.x.size).astype(self.dtype)
```

anaruse:
Thanks for your comments. I've just fixed the branch based on your feedback. Please review it again.

beam2d:
Thank you for the updates. Looks good to me. Could you add a description of the soft target support to the docstring?

toslunar:
Could you resolve merge conflicts?

chainer/functions/loss/softmax_cross_entropy.py
```diff
@@ -72,14 +74,22 @@ def check_type_forward(self, in_types):
...
+                t_type.ndim == x_type.ndim - 1,
+                x_type.shape[0] == t_type.shape[0],
+                x_type.shape[2:] == t_type.shape[1:],
+            )
+        else:
+            # assume t is soft_target
+            type_check.expect(
+                x_type.dtype.kind == 'f',
+                t_type.dtype.kind == 'f',
+                t_type.ndim == x_type.ndim,
```
toslunar: The following line covers this check.

tests/chainer_tests/functions_tests/loss_tests/test_softmax_cross_entropy.py
```diff
@@ -681,4 +681,107 @@ def test_consistency_gpu_never(self):
...
+        self.check_backward(numpy)
+
+    @attr.gpu
+    def test_backward_gpu(self):
+        self.check_backward(cuda.cupy)
+
+
+@testing.parameterize(*(testing.product({
+    'nb': [1, 2, 4],
+    'shape': [(3,), (3, 2), (3, 2, 2)],
```
toslunar: Could you add `()` to the shapes?

anaruse: Well, is there a use case when shape is () ?

anaruse: Actually, this test code does not work well with shape = ().

beam2d:
Bump. Sorry for the late response. Could you resolve the conflict again?

tests/chainer_tests/functions_tests/loss_tests/test_softmax_cross_entropy.py
beam2d: I thought there should be at least one axis in this shape following the documentation of the original hard-target softmax cross entropy.

anaruse:
I've just resolved the conflicts with the master branch. Could you check it?

beam2d:
Sorry for the late reply. The code looks good to me, but I have a bit concern on the naming now. The function named "cross entropy" that computes KL divergence is confusing and it will quite likely be misused. You asked above if it is better to name it `softmax_kl_divergence`, and that sounds better to me, but I also think it's overkill to add a new function, based on the current implementation. How about adding an option to turn on/off the negative entropy term (i.e., switching between cross-entropy mode and KL divergence mode)? The option only affects the soft target case because the cross-entropy and KL divergence are the same for the hard target case.

chainer/functions/loss/softmax_cross_entropy.py
```diff
@@ -223,21 +244,32 @@ def forward_gpu(self, inputs):
             ret = ret.reshape(t.shape)
         return ret,
 
+    def _soft_target_loss(self, xp, x, t, log_y):
+        kl_d = xp.sum(t * (xp.log(t + self.eps) - log_y), axis=1)
```
toslunar: To compute the cross entropy, 
```suggestion
        ____ = -xp.sum(t * log_y, axis=1)
```

toslunar:
BTW, `chainer.distributions.Categorical` is available.

anaruse:
All right, I added an option 'soft_target_loss' so that you can opt which loss calculation method to use for soft target loss: 'cross-entropy' or 'kl-divergence'.  What would you think about this option?

beam2d:
Thanks for the fix!
CI, test this please.

chainer/functions/loss/softmax_cross_entropy.py
```diff
@@ -223,21 +246,35 @@ def forward_gpu(self, inputs):
             ret = ret.reshape(t.shape)
         return ret,
 
+    def _soft_target_loss(self, xp, x, t, log_y):
+        if self.soft_target_loss == 'kl-divergence':
+            ret = xp.sum(t * (xp.log(t + self.eps) - log_y), axis=1)
+        else:
+            ret = -xp.sum(t * log_y), axis=1)
```
beam2d: It looks there is a syntax error here.

anaruse: Ah, sorry, I was careless..  will fix it soon.

chainer/functions/loss/softmax_cross_entropy.py
anaruse: Now, I'm wondering if this is correct, since it fails at the test below. 
https://github.com/chainer/chainer/blob/1ab26322767bec5fe9b0c6b5da702b95cb1ef3bb/tests/chainer_tests/functions_tests/loss_tests/test_softmax_cross_entropy.py#L618
This test uses output of softmax as soft target label, so output of softmax_cross_entropy becomes zero or almost zero. But it becomes non-zero when 'cross-entropy' is used as soft target loss calculation.

beam2d:
Jenkins, test this please.

beam2d:
It looks the test is still failing. Could you check it?

anaruse:
The CI test fails at `TestSoftTargetExpectNearZero` (https://github.com/chainer/chainer/blob/f84721460f96e4074135ae28970f8ad5e4ceb734/tests/chainer_tests/functions_tests/loss_tests/test_softmax_cross_entropy.py#L621) when `'cross-entropy'` is used to compute soft target loss.

The test above expects a loss value to be zero, but it becomes non-zero when `'cross-entropy'` is used. I think the cause is that this test is not appropriate for `'cross-entropy'` or computation method  of `'cross-entropy'` is not correct.

What would you think on this?

anaruse:
Sorry for being very late.
I fixed an issue of unit test `TestSoftTargetExpectNearZero` when `cross-entropy` is selected as soft target loss calculation by dividing the test to two tests, one for `kl-divergence` and another for `cross-entropy`. Perhaps there is no problem remaining. Could you check this again?

beam2d:
Jenkins and flexCI, test this please.

chainer/functions/loss/softmax_cross_entropy.py
toslunar: Done in 1ab26322767bec5fe9b0c6b5da702b95cb1ef3bb.

beam2d:
Thank you!!!

anaruse:
Thank you for merging the PR !

PR #5597, opened 2018-10-29
Add `return_samples` argument to `F.negative_sampling` and `L.NegativeSampling`

niboshi:
Refactored test of `L.NegativeSampling` as well.

### DIFF
--- a/chainer/functions/loss/negative_sampling.py
+++ b/chainer/functions/loss/negative_sampling.py
@@ -5,6 +5,7 @@ import chainer
 from chainer import backend
 from chainer.backends import cuda
 from chainer import function_node
+from chainer.utils import argument
 from chainer.utils import type_check
 
 
@@ -16,6 +17,7 @@ def _sigmoid_grad(x, y, gy):
 class NegativeSamplingFunction(function_node.FunctionNode):
 
     ignore_label = -1
+    _samples = None
 
     def __init__(self, sampler, sample_size, reduce='sum'):
         if reduce not in ('sum', 'no'):
@@ -29,14 +31,11 @@ class NegativeSamplingFunction(function_node.FunctionNode):
         self.wx = None
 
     def _make_samples(self, t):
-        if hasattr(self, 'samples'):
-            return self.samples  # for testing
-
         size = int(t.shape[0])
         # first one is the positive, and others are sampled negatives
         samples = self.sampler((size, self.sample_size + 1))
         samples[:, 0] = t
-        self.samples = samples
+        return samples
 
     def check_type_forward(self, in_types):
         type_check._argname(in_types, ('x', 't', 'W'))
@@ -57,9 +56,9 @@ class NegativeSamplingFunction(function_node.FunctionNode):
         x, t, W = inputs
 
         self.ignore_mask = (t != self.ignore_label)
-        self._make_samples(t)
+        samples = self._make_samples(t)
 
-        w = W[self.samples]
+        w = W[samples]
         wx = numpy.einsum(
             'ij,ikj->ik', x[self.ignore_mask], w[self.ignore_mask])
         wx[:, 0] *= -1
@@ -69,6 +68,8 @@ class NegativeSamplingFunction(function_node.FunctionNode):
 
         if self.reduce == 'sum':
             loss = numpy.array(loss.sum(), 'f')
+
+        self._samples = samples
         return loss,
 
     def forward_gpu(self, inputs):
@@ -76,7 +77,7 @@ class NegativeSamplingFunction(function_node.FunctionNode):
         x, t, W = inputs
 
         self.ignore_mask = (t != self.ignore_label)
-        self._make_samples(t)
+        samples = self._make_samples(t)
 
         n_in = x.shape[1]
         self.wx = cuda.elementwise(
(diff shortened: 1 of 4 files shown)


chainer/functions/loss/negative_sampling.py
```diff
@@ -314,8 +317,10 @@ def backward(self, indexes, grad_outputs):
         return ret
 
 
-def negative_sampling(x, t, W, sampler, sample_size, reduce='sum'):
-    """Negative sampling loss function.
+def negative_sampling(x, t, W, sampler, sample_size, reduce='sum', **kwargs):
```
beam2d: Why is `return_samples` keyword only?

chainer/functions/loss/negative_sampling.py
```diff
@@ -16,6 +17,7 @@ def _sigmoid_grad(x, y, gy):
 class NegativeSamplingFunction(function_node.FunctionNode):
 
     ignore_label = -1
+    _samples = None
```
beam2d: Do we need to change the attribute name from `samples` to `_samples`?

chainer/functions/loss/negative_sampling.py
```diff
@@ -373,8 +378,15 @@ def negative_sampling(x, t, W, sampler, sample_size, reduce='sum'):
 
     .. seealso:: :class:`~chainer.links.NegativeSampling`.
 
-    """
-    return (
-        NegativeSamplingFunction(sampler, sample_size, reduce)
-        .apply((x, t, W))
-    )[0]
+    """  # NOQA
```
beam2d: Add a documentation of the `return_samples` argument.

tests/chainer_tests/links_tests/loss_tests/test_negative_sampling.py
```diff
@@ -34,14 +30,49 @@ def setUp(self):
...
+
+        assert saved_samples[0] is not None
+        return out, saved_samples[0]
+
+    def check_forward(self, link, x_data, t_data):
         x = chainer.Variable(x_data)
         t = chainer.Variable(t_data)
-        y = self.link(x, t, reduce=self.reduce)
+        y, samples = self.call_link_and_return_samples(
+            lambda: link(x, t, reduce=self.reduce))
```
beam2d: How about adding `return_samples` argument to `L.NegativeSampling.forward`? I like such a consistency and we can avoid the patching trick in the tests.

chainer/functions/loss/negative_sampling.py
niboshi: To be conservative: We don't have to define fixed ordering. It's benefitial for both users and us.

* For users: Forces easier-to-understand keyword notation.
* For us: Easier to deprecate/add options in the future.

tests/chainer_tests/links_tests/loss_tests/test_negative_sampling.py
niboshi: Added the option as suggested.

chainer/functions/loss/negative_sampling.py
niboshi: Added documentation.

chainer/functions/loss/negative_sampling.py
niboshi: Reverted to `samples`.

niboshi:
PTAL

chainer/functions/loss/negative_sampling.py
beam2d: OK.

chainer/functions/loss/negative_sampling.py
```diff
@@ -359,22 +364,37 @@ def negative_sampling(x, t, W, sampler, sample_size, reduce='sum'):
         sample_size (int): Number of samples.
         reduce (str): Reduction option. Its value must be either
             ``'sum'`` or ``'no'``. Otherwise, :class:`ValueError` is raised.
+        return_samples (bool):
+            If ``True``, the sample array taken by ``sampler`` is also
+            returned.
```
beam2d: Could you add a description of "the sample array"? It is a (batch size, sample size+1)-array of integers whose first column is fixed to the ground truth labels and the other columns are drawn from the sampler.

chainer/functions/loss/negative_sampling.py
```diff
@@ -359,22 +364,37 @@ def negative_sampling(x, t, W, sampler, sample_size, reduce='sum'):
...
-            above equation.
+        ~chainer.Variable or tuple:
+            If ``return_samples`` is ``False`` (default), the output
+            variable holding the loss value(s) calculated by the
+            above equation is returned.
             If ``reduce`` is ``'no'``, the output variable holds array
             whose shape is same as one of (hence both of) input variables.
             If it is ``'sum'``, the output variable holds a scalar value.
 
+            If ``return_samples`` is ``True``, a tuple of the output
```
beam2d: How about putting this sentence right after the "False" case description, above the description of "reduce" option? I was first confused by the combinations of interleaved "if" sentences.

beam2d:
Jenkins, test this please.

chainer/functions/loss/negative_sampling.py
niboshi: Added detailed description.

chainer/functions/loss/negative_sampling.py
niboshi: Fixed.

niboshi:
PTAL

beam2d:
LGTM

Issue #5592, opened 2018-10-28
ideep4py availability check is broken


koreyou:
# Problem description

Chainer 5.0.0 does not work with ideep4py 2.0.0 (which is currently the latest version).

There is a bug (?) in ideep availability check. ideep4py 2.0.0 does not yet have `__version__` attribute.

https://github.com/chainer/chainer/blob/v5.0.0/chainer/backends/intel64.py#L11-L13

```python
try:
    import ideep4py as ideep  # NOQA
    from ideep4py import mdarray  # NOQA
    _ideep_version = 2 if hasattr(ideep, '__version__') else 1
...
def is_ideep_available():
    """Returns if iDeep is available.
    Returns:
        bool: ``True`` if the supported version of iDeep is installed.
    """
    return _ideep_version is not None and _ideep_version == 2
```

You can check that this is problematic by running `pip install ideep4py==2.0.0` as recommended by [the official documentation](https://docs.chainer.org/en/v5.0.0/install.html) and by running `python -c "import ideep4py; print(hasattr(ideep4py, '__version__'))"`.

## Environment

```
Platform: Linux-4.4.0-137-generic-x86_64-with-debian-stretch-sid
Chainer: 5.0.0
NumPy: 1.15.3
CuPy: Not Available
iDeep: Not Available
```

(iDeep *is* available. Availability check if broken.)

kmaehashi:
Thanks for pointing that out.
Please use the latest `2.0.0post1` (the latest stable) instead of `2.0.0`.
https://pypi.org/project/ideep4py/2.0.0.post1/#history

I'll update the docs to clarify this.

koreyou:
Thanks for the prompt reply. It worked out with ideep4py=2.0.0.post1 .

toslunar:
- #5600 fixed the check in v6
- #5601 fixed the doc in v5

PR #5600, opened 2018-10-31
Improve iDeep version checking

kmaehashi:
Improve documentation and environment check for ideep (see #5592).

### DIFF
--- a/chainer/_version.py
+++ b/chainer/_version.py
@@ -19,7 +19,7 @@ _optional_dependencies = [
         'packages': [
             'ideep4py',
         ],
-        'specifier': '>=2.0, <2.1',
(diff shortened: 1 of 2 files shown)


toslunar:
Is it ok to backport `>=2.0.0.post3`, which will require updating ideep4py if an older version is installed?

kmaehashi:
Ok, I'll make a separate PR for v4 (which requires `>=2.0.0.post1`).

toslunar:
> for v4

for v5

Jenkins, test this please.

kmaehashi:
Ah, I noticed that `_optional_dependencies` was finally not backported to v5. Anyway I'll send separate PR for docs to v5.

toslunar:
The Jenkins failures are unrelated to the PR.

PR #5606, opened 2018-11-01
Overlap compute and transfer more

anaruse:
This PR aims to overlap GPU kernels and data transfer more.

Current `ConcatWithAsyncTransfer` implementation uses global synchronization to prevent a GPU kernel read arrays that might be still being transferred to GPU. This implementation is safe but it blocks a CPU thread to launch GPU kernels until data transfer to GPU is finished.   With this PR, the global synchronized is removed and CUDA events are used instead to further overlap GPU kernels and data transfers w/o blocking a CPU thread to launch GPU kernels. You should specify CUDA stream used for compute kernels by an argument `compute_stream` like below, or global synchronization is used as before.

    converter = dataset.ConcatWithAsyncTransfer(compute_stream=chainer.backends.cuda.Stream.null)

### DIFF
--- a/chainer/dataset/convert.py
+++ b/chainer/dataset/convert.py
@@ -205,13 +205,30 @@ class ConcatWithAsyncTransfer(object):
         stream (cupy.cuda.Stream): CUDA stream. If ``None``, a stream is
             automatically created on the first call. Data transfer operation
             is launched acynchrnously using the stream.
+        compute_stream(cupy.cuda.Stream): CUDA stream used for compute kernels.
+            If not ``None``, CUDA events are created/used to avoid global
+            synchronization and overlap execution of compute kernels and data
+            transfers as much as possible. If ``None``, global synchronization
+            is used instread.
     """
 
-    def __init__(self, stream=None):
+    def __init__(self, stream=None, compute_stream=None):
         self._stream = stream
+        self.compute_stream = compute_stream
+
         self._device = None
         self._conveyor = collections.defaultdict(
             lambda: Conveyor(self._device, self._stream))
+        if compute_stream is not None:
+            # * event1 prevents a CPU thread to update arrays that might be
+            #   still being used by GPU kernels.
+            # * event2 prevents a GPU kernel to read arrays that might be
+            #   still being transfered to GPU.
+            self.event1 = cuda.Event()
+            self.event2 = cuda.Event()
+            self.sync_get = False
+        else:
+            self.sync_get = True
 
     def __call__(self, batch, device=None, padding=None):
         """Concatenate data and transfer them to GPU asynchronously.
@@ -239,6 +256,10 @@ class ConcatWithAsyncTransfer(object):
         if device is not self._device:
             raise ValueError('device is different')
 
+        if self.compute_stream is not None:
+            self.event1.synchronize()
+            self.event1.record(stream=self.compute_stream)
+
         with cuda.get_device_from_id(device):
             if isinstance(first_elem, tuple):
                 result = []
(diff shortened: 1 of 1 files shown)


chainer/dataset/convert.py
```diff
@@ -205,13 +205,30 @@ class ConcatWithAsyncTransfer(object):
         stream (cupy.cuda.Stream): CUDA stream. If ``None``, a stream is
             automatically created on the first call. Data transfer operation
             is launched acynchrnously using the stream.
+        compute_stream(cupy.cuda.Stream): CUDA stream used for compute kernels.
+            If not ``None``, CUDA events are created/used to avoid global
+            synchronization and overlap execution of compute kernels and data
+            transfers as much as possible. If ``None``, global synchronization
+            is used instread.
```
mitmul:
```suggestion
            is used instead.
```

chainer/dataset/convert.py
```diff
@@ -205,13 +205,30 @@ class ConcatWithAsyncTransfer(object):
         stream (cupy.cuda.Stream): CUDA stream. If ``None``, a stream is
             automatically created on the first call. Data transfer operation
             is launched acynchrnously using the stream.
```
mitmul:
```suggestion
            is launched asynchronously using the stream.
```

anaruse:
Thanks, @mitmul . Typos were fixed.

chainer/dataset/convert.py
```diff
@@ -239,6 +256,10 @@ def __call__(self, batch, device=None, padding=None):
         if device is not self._device:
             raise ValueError('device is different')
 
+        if self.compute_stream is not None:
+            self.event1.synchronize()
```
okuta: Why do you use synchronize?

anaruse: This synchronization is necessary to ensure that a CPU thread won't break the data of arrays that might be still being used by GPU kernels for previous iteration.

Suppose you are preparing input data for iteration _t_. Considering double-buffering scheme is used here, the array to be used for iteration _t_ was/is also used for iteration _t-2_. If you update/write the array for iteration _t_ while the array still being used by GPU kernels for iteration _t-2_, the results of iteration _t-2_ might get wrong.

How can we avoid this possible data corruption? The safe way is to make sure that all GPU kernels for iteration _t-2_ are finished, and the following codes are for that purpose.

    259        if self.compute_stream is not None:
    260            self.event1.synchronize()
    261            self.event1.record(stream=self.compute_stream)

As you can see, `self.event1.synchronize()` wait for the event that is "recorded" at previous iteration, more specifically, at the beginning of iteration _t-1_, in other word, the end of iteration _t-2_, on "compute stream". With this event synchronization, you can safely update the array.

chainer/dataset/convert.py
```diff
@@ -345,19 +375,23 @@ def put(self, array):
...
-    def get(self):
+    def get(self, sync=True):
         """Returns the array of data transferred to a target device asynchronously.
 
-        This method first waits for completion of asynchrnous data trasfer
-        initiated by :meth:`put`, then returns the array on the target
-        device.
+        If sync is ``True``, the data of returned array is available in GPU
+        kernels. If sync is ``False``, the data of returned array might be
+        being transfered to GPU, so synchronizeaion must be done carefully by
```
grafi-tt: synchronizeaion -> synchronization

anaruse: Thanks!

anaruse:
Thanks for your comment, @okuta .
I've just changed variable names based on your suggestion.

okuta:
Thanks
jenkins, test this please.

okuta:
jenkins, test this please.

chainer/dataset/convert.py
okuta: OK, I understood.

okuta:
LGTM!

PR #5511, opened 2018-10-18
Test statistics of initializers

toslunar:
Test `scale` arg of `chainer.initializers`.

This PR adds statistical tests ~and fixes bias of sign of `Orthogonal` initializer~.

PR #5615, opened 2018-11-02
Make signs generated by `initializers.Orthogonal` unbiased

toslunar:
This PR fixes the issue that signs of vectors `initializers.Orthogonal` generated were biased.

Merge #5511 first.


grafi-tt:
It might be helpful for readers if you would add a reference to the paper: Francesco Mezzadri, "How to generate random matrices from the classical compact groups",  <https://arxiv.org/abs/math-ph/0609050>, p11.

toslunar:
@grafi-tt Could you tell me the reason why you chose the paper?
My googling "generate random orthogonal matrix" shows https://math.stackexchange.com/questions/2098897/generating-random-orthogonal-matrices at the top and it cites the paper you mentioned.
The paper (or the *lecture note*) cites the *survey* "Random matrix theory" https://web.eecs.umich.edu/~rajnrao/Acta05rmt.pdf, which cites G. W. Stewart (1980) https://www.jstor.org/stable/2156882, which could be the original while I've not confirmed it.

grafi-tt:
@toslunar Many thanks for your investigation.

I did found the paper via the same StackExchange page when I had wondered how to initialize orthogonal matrix a short while ago. Basically just because "hit by Google search". 

I think referring to an internet-accessible resource is good as a programming practice, but I understand it's not academically fair. If such fairness is concerned, I withdraw my opinion.

Still I believe the algorithm is mystic for the most programmers, not math-inclined. Just adding a comment kind of "take the canonical decomposition for unbiased sampling" would give some clues for understanding. (Anyway docstrings need to be updated if QR decomposition is used instead of SVD.)

beam2d:

PR #5618, opened 2018-11-02
Fix test of `F.rrelu`

toslunar:
- Fix typos, including `'train'` config.
- The shape of `r` should be the same as the shape of `x`.
- Remove `retry` from tests.  I confirmed that the cpu tests pass with `repeat(10000)`.


okuta:
jenkins, test this please.

okuta:
LGTM!

PR #5547, opened 2018-10-21
Avoid unnecessary copy in `ndarray.astype`

toslunar:
This would improve the performance.

PR #5623, opened 2018-11-05
[backport] Avoid unnecessary copy in `ndarray.astype`

beam2d:
Backport of #5547


beam2d:
I excluded the change on `basic_math.py` because it depends on #5439 which is not to be backported.

PR #5568, opened 2018-10-24
Simplify batch normalization with cuDNN

okuta:
Please merge cupy/cupy#1737 first.

PR #5567, opened 2018-10-24
Simplify pooling with cuDNN

okuta:
Please merge cupy/cupy#1735 first.

PR #5566, opened 2018-10-23
Simplify softmax with cuDNN

okuta:
Please merge cupy/cupy#1736 first.

PR #5626, opened 2018-11-06
improve cuDNN call performance by caching ndarray.ctypes objects

grafi-tt:
Retrieving `numpy.ndarray.ctypes` is slow because it imports `numpy.core._internal` module. I checked the behavior with numpy 1.14.5. 

Batchnorm and pooling routines uses the property for passing constants 0 and 1 to cuDNN. If CPU is bottleneck, this slows down training.

You can check the behavior by `cProfile` module and/or the following code that hooks imports.

```python
import builtins
ip_old = builtins.__import__
npi_count = 0
def ip(name, globals=None, locals=None, fromlist=(), level=0):
    global mp_count
    print(name)
    if name == 'numpy.core._internal':
        npi_count +=1
        if npi_count == 1000:
            raise
    return ip_old(name, globals, locals, fromlist, level)
builtins.__import__ = ip
```

I performed a benchmark for `chainer/examples/cifar` by the command below. The sample code is modified to disable snapshots.

```sh
python train_cifar.py --epoch 2 --batchsize 32
```

The results with i5-3550 CPU and GTX 1080 Ti GPU are as follows. If you have faster CPU or slower GPU, you may have to clock-down the CPU or decrease the number of filters. Though the difference is small, results have been consistent on several runs. I also confirmed ~5-10% speed-up on a real-world task.

Before:

```
1           2.44298     2.2634                0.112944       0.138578                  60.286
2           1.90411     1.77155               0.239937       0.295028                  118.455
```

After:

```
1           2.24371     1.88177               0.178003       0.242013                  59.4717
2           1.74626     1.54112               0.310559       0.3751                    116.899
```

### DIFF
--- a/chainer/functions/activation/log_softmax.py
+++ b/chainer/functions/activation/log_softmax.py
@@ -8,10 +8,21 @@ import chainer.functions
 from chainer.functions.activation import softmax
 from chainer.utils import type_check
 
+
 if cuda.cudnn_enabled:
     cudnn = cuda.cudnn
     libcudnn = cuda.cuda.cudnn
     _algorithm = libcudnn.CUDNN_SOFTMAX_LOG
+
+
+_singleton_ndarray_ctypes_cache = {}
+def _singleton_ndarray_ctypes(v, dtype):
+    if (v, dtype) in _singleton_ndarray_ctypes_cache:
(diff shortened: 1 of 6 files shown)


grafi-tt:
Just now I noticed this PR is not required when #5566, #5567 and #5568 are merged.

PR #5629, opened 2018-11-07
Add Code of Conduct

Crissman:
Add Code of Conduct for contributors


kmaehashi:
Fyi, `CODE_OF_CONDUCT.md` is the standard filename. https://help.github.com/articles/adding-a-code-of-conduct-to-your-project/
(I prefer to put it under `docs/` if we add one)

kmaehashi:
Ok, I think root directory is also fine to make it more visible to users.

PR #5634, opened 2018-11-07
support spatial_scale >= 1.0 in roi_average_align_2d.py

knorth55:
currently, when `spatial_scale > 1.0` is given, the function hangs.

PR #5635, opened 2018-11-07
support spatial_scale >= 1.0 in roi_max_align_2d.py

knorth55:
~Merge after #5634~

- test spatial_scale >= 1.0 in test_roi_max_align_2d.py
- support spatial_scale >= 1.0 in roi_max_align_2d.py
- fix typo in roi_max_align_2d.py


Hakuyume:
Thanks. I'll check.

chainer/functions/pooling/roi_max_align_2d.py
```diff
@@ -112,31 +113,32 @@ def forward_cpu(self, inputs):
...
-                roi_bin_grid_w = numpy.ceil(roi_width / pooled_width)
+                roi_bin_grid_w = int(numpy.ceil(roi_width / pooled_width))
             else:
                 roi_bin_grid_w = self.sampling_ratio[1]
 
-            max_val = None
-            max_index = None
-            iy = 0
-            while iy < roi_bin_grid_h:
+            max_val = -1e20
```
Hakuyume: How about using `-math.inf` or `-np.inf` ?

Hakuyume: For CUDA kernel, `-FLT_MAX`? (I'm not sure)

Hakuyume:
jenkins, test this please.

Hakuyume:
jenkins, test this please.

mitmul:
Jenkins, test this please.

mitmul:
Jenkins, test this please.

niboshi:
Is it necessary to treat NaNs as zeroes in `numerical_grad`?
I don't understand the source of NaNs, but if a function returned outputs with NaN, it would be an error in [debug mode](https://docs.chainer.org/en/latest/reference/debug.html).
Even if NaN needs be an output of a function, I think it's better to do the replacement of NaN values in the test rather than `numerical_grad`.

knorth55:
@niboshi it returns `-np.inf` as return values https://github.com/chainer/chainer/blob/a437a9f39e22ecd0c671ff2e98a2372dcdef0e29/chainer/functions/pooling/roi_max_align_2d.py#L124
this is the reason why numerical grad has `NaN`
Return value =-np.inf` is already discussed in https://github.com/chainer/chainer/pull/5635#discussion_r237037529

niboshi:
@knorth55 
Thanks, I understand the source of NaN.
However I think there's no mathematically sound definition for numerical gradients of functions with infinite numbers. I think it's better to handle such values in the test, as I stated above.
Specifically, `inf`s in the function outputs can be replaced with 0 by using `F.where`.

knorth55:
@niboshi OK, you mean solving the problem like this?
https://github.com/chainer/chainer/pull/5635/commits/12a82e6f0dd20aa1802b9b97204297b440f801d8

niboshi:
@knorth55 Yes, exactly. Thank you so much for the quick fix!

niboshi:
@mitmul LGTM

mitmul:
Thanks

PR #5624, opened 2018-11-05
Support all float dtypes in `F.embed_id`

takagi:
This PR follows #4582, generalizes `F.embed_id` to support all float data types.

PR #5638, opened 2018-11-08
Dtype blackout

takagi:
Merge #5624 first.

`L.BlackOut` will work with the default dtype as it is after `F.embed_id`, on which `L.BlackOut` depends, will support all float dtypes in #5624.

This PR just adds tests for the default dtype in `L.BlackOut`.


takagi:
Jenkins, test this please.

Issue #5585, opened 2018-10-26
BestValueTrigger should repeat the same result within an iteration


niboshi:
If an instance of `triggers.BestValueTrigger` (`MinValueTrigger` / `MaxValueTrigger`) is called multiple times in a single iteration, it can evaluate to `True` only at the first call, even when the best value is achieved.

### Cause

The cause is that the trigger only checks the value per-call.
Subsequent calls within an iteration do not update the best value, so it's not triggered.

### Possible fix

(Discussed with @beam2d )
All builtin triggers, not only `BestValueTrigger`, are not assumed to be used multiple times in an iteration. It's better to provide a general fix.

A possible fix is to introduce a base trigger class that returns the cached result in subsequent calls, and let concrete triggers implement their `__call__` logic in an overridden method.

All builtin triggers should derive from it.

### See also

(japanese) https://groups.google.com/forum/#!topic/chainer-jp/QWwQYdt-syM

PR #5641, opened 2018-11-08
Iteration-aware trigger

niboshi:
Related: #5585

### DIFF
--- a/chainer/training/trigger.py
+++ b/chainer/training/trigger.py
@@ -1,4 +1,107 @@
-# For backward compatibility
-from chainer.training.triggers.interval_trigger import IntervalTrigger  # NOQA
-from chainer.training.util import _never_fire_trigger  # NOQA
-from chainer.training.util import get_trigger  # NOQA
+import functools
+
+
+class _IterationAwareTriggerWrapper(object):
+
+    _prev_epoch_detail = None
+    _prev_result = None
+
+    def __init__(self, original_trigger):
+        self._original_trigger = original_trigger
+
+    def __call__(self, trainer):
+        prev_epoch_detail = self._prev_epoch_detail
+        epoch_detail = trainer.updater.epoch_detail
+        if prev_epoch_detail == epoch_detail:
+            result = self._prev_result
+        else:
+            result = self._original_trigger(trainer)
+            self._prev_result = result
+            self._prev_epoch_detail = epoch_detail
+        return result
+
+
+def iteration_aware(_arg=None):
+    """iteration_aware()
+Utility to make a trigger aware of training iterations.
+
+    This utility makes the trigger cache its evaluation and return the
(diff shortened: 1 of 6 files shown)


niboshi:
This PR is blocked by #5645.

niboshi:
PTAL

niboshi:
>cached_through_iteration

Doesn't that sound like the cache lives across iterations?
Perhaps `cached_during_iteration`? (I tempted to include `each` or `every`, but actually the cache does not necessarily exist in every iteration).

hvy:
I have not followed the discussion behind this PR so excuse me if this is nonsense. May I ask why this issue was generalized to all triggers? It sounds like a bug in only some of the triggers incl. best value trigger to me? 🤔  IMHO the concept of "iteration" is necessarily inherent to all triggers (e.g. the `TimeTrigger`) and this PR sort of introduces an assumption that a trigger are always evaluated on an iteration basis. Although true, and I understand it from a practical point of view, but am curious from a design perspective. I guess one way to see it would be to have a `Trigger` class that doesn't care about iterations, and a `IterationAwareTrigger(Trigger)`, instead of a flag that is?

niboshi:
@hvy All triggers receive a trainer. I think it's natural that the concepts of triggers and iterations are coupled.

hvy:
For the record, discussed offline and agreed that this flag is `cached_during_iteration` is not needed. Rather we should make sure all triggers behave as if "iteration aware" by default.

chainer/training/trigger.py
```diff
@@ -1,4 +1,146 @@
...
+        @chainer.trigger
+        def random_trigger(trainer):
+            # A trigger that triggers randomly at 10% of all the iterations.
+            return random.randint(0, 9) == 0
+
+    See also:
+        :func:`chainer.trigger`
+    """
+
+    cached_during_iteration = True
```
hvy: Agreed to remove this flag as per earlier comment.

chainer/training/trigger.py
```diff
@@ -1,4 +1,146 @@
...
+
+    # None if the result is NOT cached during an iteration.
+    # Otherwise a tuple containing two elements:
+    # [0] epoch detail and [1] the evaluation (True/False).
+    __cache_during_iteration = None
+
+    def evaluate(self, trainer):
```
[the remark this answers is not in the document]
niboshi: Hmm, I don't think `trigger.evaluate()` is not abstract at all.

chainer/training/trigger.py
niboshi: I don't remember the argument, but should `TimeTrigger` for example be also iteration-aware?

Saanidhyavats:
Whether someone else is working on this issue or shall I start contributing on it

PR #5643, opened 2018-11-09
Update style check tools to the versions compatible with pycodestyle 2.4

toslunar:
The newest flake8 (3.6.0) requires pycodestyle 2.4, which has the `W504` check.
It seems `W504` needs to be added to `ignore` in `setup.py`, while the check is disabled by default.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,8 +59,7 @@ install:
 
 script:
   - flake8
-  - autopep8 -r . --diff | tee check_autopep8
-  - test ! -s check_autopep8
+  - autopep8 -r . --diff --exit-code
   # To workaround Travis issue (https://github.com/travis-ci/travis-ci/issues/7261),
   # ignore DeprecationWarning raised in `site.py`.
   - python -Werror::DeprecationWarning -Wignore::DeprecationWarning:site -m compileall -f -q chainer chainermn examples tests docs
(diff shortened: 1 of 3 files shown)


niboshi:
Why does it have to be explicitly `ignore`d if it is disabled by default?

niboshi:
Also, I wonder if there is any motivation to upgrade flake8 in the first place.

toslunar:
My motivation to update flake8 is a better support for Python 3.7.
With `flake8==3.5`, the following warning message is shown.
```
% flake8 chainer
/Users/tos/.pyenv/versions/3.7.1/envs/n/lib/python3.7/site-packages/pycodestyle.py:113: FutureWarning: Possible nested set at position 1
  EXTRANEOUS_WHITESPACE_REGEX = re.compile(r'[[({] | []}),;:]')
```

toslunar:
`W503` and `W504` are mutually exclusive.  I prefer the style of
```
a
+ b
```
while I don't request enabling `W504` in Chainer.

On `setup.cfg`,
http://pycodestyle.pycqa.org/en/latest/intro.html#error-codes
> Please note that if the option `–ignore=errors` is used, the default configuration will be overridden and ignore only the check(s) you skip.

niboshi:
Ah, OK. That answers my question. Thanks.

okuta:
LGTM!

PR #5646, opened 2018-11-10
Avoid using dype char codes

niboshi:
NumPy dtype char codes (such as `'f'`, `'i'`) are expressly discouraged in NumPy documentation:
https://docs.scipy.org/doc/numpy/reference/arrays.scalars.html#built-in-scalar-types

They represent corresponding C types (such as `float` and `int`), whose actual bit widths may be dependent on environment.

### DIFF
--- a/chainer/functions/activation/rrelu.py
+++ b/chainer/functions/activation/rrelu.py
@@ -126,7 +126,7 @@ def rrelu(x, l=1. / 8, u=1. / 3, **kwargs):
 
     .. admonition:: Example
 
-        >>> x = np.array([[-1, 0], [2, -3], [-2, 1]], 'f')
+        >>> x = np.array([[-1, 0], [2, -3], [-2, 1]], numpy.float32)
         >>> x
         array([[-1.,  0.],
                [ 2., -3.],
--- a/chainer/functions/activation/rrelu.py
+++ b/chainer/functions/activation/rrelu.py
@@ -126,7 +126,7 @@ def rrelu(x, l=1. / 8, u=1. / 3, **kwargs):
 
     .. admonition:: Example
 
-        >>> x = np.array([[-1, 0], [2, -3], [-2, 1]], numpy.float32)
+        >>> x = np.array([[-1, 0], [2, -3], [-2, 1]], np.float32)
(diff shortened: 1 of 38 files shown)


toslunar:
See also #4350, which changed char codes in the documents and the examples.

toslunar:
Jenkins, test this please.

chainer/links/connection/convolution_nd.py
```diff
@@ -63,7 +63,7 @@ class ConvolutionND(link.Link):
 
         Let an input vector ``x`` be:
 
-        >>> x = np.arange(2 * 5 * 5 * 5, dtype='f').reshape(1, 2, 5, 5, 5)
+        >>> x = np.arange(2 * 5 * 5 * 5, dtype=numpy.float32).reshape(1, 2, 5, 5, 5)
```
kmaehashi: s/numpy/np/

chainer/links/connection/deconvolution_nd.py
```diff
@@ -61,7 +61,7 @@ class DeconvolutionND(link.Link):
 
         Let an input vector ``x`` be:
 
-        >>> x = np.arange(2 * 5 * 5 * 5, dtype='f').reshape(1, 2, 5, 5, 5)
+        >>> x = np.arange(2 * 5 * 5 * 5, dtype=numpy.float32).reshape(1, 2, 5, 5, 5)
```
kmaehashi: ditto

chainer/functions/activation/rrelu.py
```diff
@@ -126,7 +126,7 @@ def rrelu(x, l=1. / 8, u=1. / 3, **kwargs):
 
     .. admonition:: Example
 
-        >>> x = np.array([[-1, 0], [2, -3], [-2, 1]], 'f')
+        >>> x = np.array([[-1, 0], [2, -3], [-2, 1]], numpy.float32)
```
kmaehashi: s/numpy/np/

niboshi:
Thank you for pointing out.
Fixed as suggested.

toslunar:
Jenkins, test this please.

PR #5662, opened 2018-11-14
Use automatic broadcasting instead of `F.repeat`

toslunar:


toslunar:
Jenkins, test this please.

PR #5665, opened 2018-11-14
Fix typo in resnet

takaaki82:


kmaehashi:
Thanks, LGTM!

kmaehashi:
Hmm, there are many others left to fix...

```
chainer/exporters/caffe.py:            argment ``args`` is directly given to the model by the ``()``
chainer/exporters/caffe.py:        args (list of ~chainer.Variable): The argments which are given to the
chainer/links/model/vision/resnet.py:            with the specified number of layers given as the first argment to
chainer/sequential.py:    method in the same order as the given argments.
chainer/sequential.py:        order as the order in which the argments are given to the constructor.
chainer/sequential.py:                'All elements of the argment should be callable. But '
chainer/sequential.py:        :class:`~Link`, the argment ``type_name`` should be its class name,
chainer/sequential.py:        the argment ``type_name``. For example, if you want to know the number
```

PR #5669, opened 2018-11-15
Refactor the statemachine of iterators that iterates indices

grafi-tt:
Iterators have had almost same routines to enumerate indices. This PR factor out the routine.

Further refactoring isn't done, because it may bloat up diff.


grafi-tt:
@toslunar Thank you for your suggestion. I changed to raise an error if `repeat` is set and return an empty batch otherwise.

ChainerMN related test errors are fixed. The errors occurred where a master iterator sends `self._order` to slave iterators.

toslunar:
Jenkins, test this please.

PR #5681, opened 2018-11-19
Add optional property `finished` to trigger object

hitsgub:
This PR is separated from #5565.
New optional property `finished` in `trigger` object helps to avoid unnecessary call of `Extension.initialize`.
The main target of this `finished` property is `once_trigger` in PR #5565, but the PR is now working.
So I first apply it to `manual_schedule_trigger` instead.

TODO:
- [ ] Add property `finished` to `once_trigger` in PR #5565

### DIFF
--- a/chainer/training/triggers/manual_schedule_trigger.py
+++ b/chainer/training/triggers/manual_schedule_trigger.py
@@ -18,12 +18,18 @@ class ManualScheduleTrigger(object):
         unit (str): Unit of the time specified by ``points``. It must be
             either ``'iteration'`` or ``'epoch'``.
 
+    Attributes:
+        finished (bool): Flag that triggered when this trigger will never
+            called again. The flag helps decision to call
+            `Extension.initialize` or not in `trainer`.
+
     """
 
     def __init__(self, points, unit):
         assert unit == 'epoch' or unit == 'iteration'
         self.points = (points if isinstance(points, list) else [points])
         self.unit = unit
+        self.finished = False
 
         self._previous_iteration = 0
         self._previous_epoch_detail = 0.
@@ -54,6 +60,9 @@ class ManualScheduleTrigger(object):
             fire = any(
                 previous_epoch_detail < p <= epoch_detail
                 for p in self.points)
+
+            if fire and epoch_detail >= max(self.points):
+                self.finished = True
         else:
             iteration = updater.iteration
             previous_iteration = self._previous_iteration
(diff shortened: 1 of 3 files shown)


chainer/training/triggers/manual_schedule_trigger.py
```diff
@@ -18,12 +18,18 @@ class ManualScheduleTrigger(object):
         unit (str): Unit of the time specified by ``points``. It must be
             either ``'iteration'`` or ``'epoch'``.
 
+    Attributes:
+        finished (bool): Flag that triggered when this trigger will never
+            called again. The flag helps decision to call
+            `Extension.initialize` or not in `trainer`.
```
beam2d: Suggestion to improve:
"Flag that indicates whether or not this trigger will fire in the future. This flag is used to determine if the extension should be initialized after resume."

hitsgub: Sorry for the late response, and thank you for your suggestion.
I reflected it to the latest version.

beam2d:
Thanks for the update. Could you add checks for the `finished` property to the tests of `ManualScheduleTrigger`?

hitsgub:
> Thanks for the update. Could you add checks for the `finished` property to the tests of `ManualScheduleTrigger`?

Thank you.
Is it OK with the following way? (I added only one test case for preview.)
https://github.com/chainer/chainer/pull/5681/files#diff-aff69d90a7835db143b7a9215d6cf16f

beam2d:
Looks good. It is also good to use a simpler way to write down the expected values, e.g. `[False] * n + [True] * m` or `[i >= n for i in range(m)]` or writing and using a utility function for that.

hitsgub:
> Looks good. It is also good to use a simpler way to write down the expected values, e.g. `[False] * n + [True] * m` or `[i >= n for i in range(m)]` or writing and using a utility function for that.

Thank you for your opinion. A utility function alike `[i >= n for i in range(m)]` is added and added test cases using this function.

beam2d:
Could you check the Travis failure? It looks the compatibility test is failing. I think it's fine to follow the same way as other attributes like `_previous_iteration`.

hitsgub:
I have already checked, just I have been in office.
I will fix it as same way as `_previous_iteration`.

beam2d:
The jenkins failure looks irrelevant to this PR. LGTM!

PR #5692, opened 2018-11-22
Fix docs for backprop_step

### LINKED ISSUE #5690: `backprop_step` only partially documented
The `backprop_step` function in `chainer/_backprop_utils.py` is documented but misses an explanation for the `func` argument. https://github.com/chainer/chainer/blob/master/chainer/_backprop_utils.py#L73-L89

hvy:
Fixes #5690 and a typo.

### DIFF
--- a/chainer/_backprop_utils.py
+++ b/chainer/_backprop_utils.py
@@ -78,8 +78,8 @@ def backprop_step(
     :func:`chainer.grad`.
 
     Args:
-        func (:class:`~chainer.FunctionNode`): The function for which gradients
-            are accumulated.
+        func (~chainer.FunctionNode): The function for which gradients are
+            accumulated.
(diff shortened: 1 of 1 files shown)


chainer/_backprop_utils.py
```diff
@@ -78,13 +78,16 @@ def backprop_step(
     :func:`chainer.grad`.
 
     Args:
-        target_input_indexes (tuple of int): Sorted indices of the input
-            variables w.r.t. which the gradients are required. It is
-            guaranteed that this tuple contains at least one element.
+        func (:class:`~chainer.FunctionNode`): The function for which gradients
```
toslunar: `func (~chainer.FunctionNode):` would work.

toslunar:
Jenkins, test this please.

PR #5699, opened 2018-11-23
Refactor gradient setter in `gradient_check`

niboshi:
Allow `y_grad=None` in any target functions, not only loss functions.


niboshi:
Rebased.
PTAL

chainer/gradient_check.py
```diff
@@ -21,6 +22,34 @@ def _copy_arrays(xs):
...
+def _check_outputs_and_grad_outputs(outputs, grad_outputs):
+    if len(outputs) != len(grad_outputs):
+        raise ValueError(
+            'Output gradients must contain equally as many elements as '
+            'the number of output elements.\n'
+            'Outputs         : {}\n'
+            'Output gradients: {}'.format(
+                ', '.join(str(y.shape) for y in outputs),
+                ', '.join(str(None if gy is None else gy.shape)
+                          for gy in grad_outputs)))
```
toslunar: Why are the `shape`s printed on this error?

chainer/gradient_check.py
```diff
@@ -345,21 +375,23 @@ def run(self):
...
         # otherwise the shapes of uninitialized parameters wouldn't be
         # determined.
-        xs_backward, y_backward, y0_data, y_grad = (
+        xs_backward, ys = (
             self._forward_for_backward_gradients())
-        self.y_grad = y_grad
+        # Keep output arrays to save computation in numerical gradients
+        y0_data = tuple([y.array for y in ys])
+
+        # If y_grad is not given, generate the all-1 gradients.
```
toslunar: The current behavior is compatible with `Variable.backward`.  Do you want to change it to be compatible with `chainer.grad`?

niboshi: Are you referring to the `Variable.backward` behavior where it fills grad with 1 only for those with shape `()`? Actually it's not so clear to me why the behavior of `Variable.backward` is like that.
Also, considering the purpose of `gradient_check`, I think there's no reason for uers of it to expect the same behavior.

chainer/gradient_check.py
niboshi: I thought it would be helpful for those who read this message to spot the cause.

toslunar: I got your intention.  Could you make the error message clearer in that the shapes are printed?

chainer/gradient_check.py
```diff
@@ -21,6 +22,34 @@ def _copy_arrays(xs):
...
+                          for gy in grad_outputs)))
+    if not all([gy is None or y.shape == gy.shape
+                for y, gy in zip(outputs, grad_outputs)]):
+        raise ValueError(
+            'Shapes of outputs and output gradients do not match.\n'
+            'Outputs         : {}\n'
+            'Output gradients: {}\n'.format(
+                ', '.join(str(y.shape) for y in outputs),
+                ', '.join(str(None if gy is None else gy.shape)
+                          for gy in grad_outputs)))
```
toslunar: I'm wondering if `dtype` can be checked in this function, too.

chainer/gradient_check.py
toslunar: random is a better default

niboshi: That's also fine for me, but isn't that compatibility breaking?
In that case ()-shape grads should be initialized in random for consistency.

chainer/gradient_check.py
```diff
@@ -71,6 +100,7 @@ def numerical_grad(
 
     """
     assert eps > 0
+    assert isinstance(inputs,  (tuple, list))
```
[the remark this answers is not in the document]
niboshi: Fixed.

chainer/gradient_check.py
niboshi: Fixed the messages.

chainer/gradient_check.py
niboshi: After discussion, I reverted the behavior change. That can be separated from this PR.

chainer/gradient_check.py
niboshi: I thought I encountered some error by doing this, but now I can't find it.
Fixed so that dtypes are compared as well.

niboshi:
PTAL.

Currently the error message looks like this:
```
E   ValueError: Shapes and/or dtypes of outputs and output gradients do not match.            
E   Output shapes and dtypes         : (2, 2, 3):float16, (2, 3, 3):float16, (2, 2, 3):float16                     
E   Output gradient shapes and dtypes: (12,):float16, (18,):float16, (12,):float16
```
(I wonder if there is any canonical way to present shapes and dtypes🤔)

toslunar:
The Jenkins failure (`TestDeconvolutionND_param_19.test_forward_consistency_cudnn`) seems unrelated to the PR.

PR #5721, opened 2018-11-27
[WIP] Raise TypeError if None is applied to FunctionNode

toslunar:
Make `Func(...).apply((None,))` immediately fail.

The motivation is the following issue I claimed. (https://chainer.slack.com/archives/C5P26TUR1)/p1542017121027400
> ```
> >>> chainer.as_variable(None)
> variable(None)
> ```
> Is this intended?
> `Func().apply((None,))` don't fail immediately because of this

Other possible fixes
- (More compatible, but hardly effective) Just warn it at `forward` and raise an error at `backward`
- (Less compatible, and possibly fixes other issues) Return `None` for `as_variable(None)`

### DIFF
--- a/chainer/function_node.py
+++ b/chainer/function_node.py
@@ -223,6 +223,10 @@ Use apply() method instead.\
             A tuple of output :class:`~chainer.Variable` objects.
 
         """
+        if any([x is None for x in inputs]):
+            raise TypeError(
+                'inputs to FunctionNode.apply contain None')
+
         input_vars = [chainer.as_variable(x) for x in inputs]
         in_data = tuple([x.data for x in input_vars])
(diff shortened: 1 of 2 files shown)


niboshi:
Note that `FunctionNode.forward()` can return `None` as an output and in that case `Variable(None)` is returned. Even build-in functions do this.

I'm in favor of forbidding `Variable(None)` but I think whether to allow `Variable(None)` and whether to allow `FunctionNode.apply((None,))` are two independent issues.

toslunar:
It's possible that allowing `None` inputs makes the implementation of `LSTMGrad.forward` easier.
On `NegativeSamplingFunctionGrad.forward`, I'd like to avoid returning `None` by fixing `NegativeSamplingFunction.backward`.

PR #5722, opened 2018-11-27
Stop `Variable.backward` from gathering other gradients

### LINKED ISSUE #5709: Document upstream gradient behavior for functions with multiple outputs
When a function has several outputs and the user backwards through one of them, it is not obvious that all of the other upstream gradients are collected. This is probably most true when the last function in the graph has multiple outputs. The following is such a case.

```python
import numpy as np
import chainer
from chainer import Variable

x = chainer.Variable(np.arange(4, dtype='f'))
ys = chainer.functions.split_axis(x, 2, axis=0)
for y in ys:
    y.grad_var = chainer.Variable(np.full_like(y, 3, dtype='f'))
ys[0].backward()
x.grad  # [3, 3, 3, 3]. Some users might expect [3, 3, 0, 0] since ys[1] is not involved?
```

How about document this behavior, or is it already?

toslunar:
Make the semantics compatible with `chainer.grad`.

Close #5709.


kmaehashi:
Blocked by #5952

toslunar:
#5952 is merged.  I'll resolve conflicts.

toslunar:
TODO:

- Add docs of the compatibility option (`reproduce_backward_output_bug`).
- Recommend `chainer.backward` in the warning message.

niboshi:
Just confirming, are you going to do the TODOs (https://github.com/chainer/chainer/pull/5722#issuecomment-500359777) in this PR?

PR #5724, opened 2018-11-27
Detect disabled backprop

### LINKED ISSUE #5457: Semantics of `enable_backprop=False` is not safe
`chainer.using_config('enable_backprop', False)`, or equivalently `chainer.no_backprop_mode()`, is used for two different purposes:
- For performance: e.g. default option `enable_double_backprop=False`  of `backward`;
- For partial gradient: e.g. target Q-network.

I'd like to have safe option to assert that a block of computation will not be `backward`-ed later, for the first purpose above.

related: #5454

toslunar:
Use `v.array` if partial gradient is needed.

Close #5457.


toslunar:
Jenkins, test this please.

toslunar:
Jenkins, test this please.

niboshi:
I think the change itself is reasonable, but my concern is that it might not be trivial to change chainerx-wrapped `Variable`s to behave in the similar way.

It might be safer to change chainerx variables altogether, perhaps after other aspects of backprop are first aligned.

PR #5725, opened 2018-11-27
ChainerX

beam2d:
ChainerX is a new ndarray+autograd implementation for Chainer. It aims at reducing Python overhead for host-heavy workloads, providing Python-free basis for application coding, and making device support pluggable.

This PR is to move the development to the main repository. It includes:

- C++ implementation of the new ndarray and related routines (`chainerx_cc`),
- Python binding of ChainerX (`chainerx_cc/python` and `chainerx`),
- Modifications on Chainer to support the new ndarray, and
- Documentation of ChainerX

We are planning to include this feature in the next beta release.

This PR is very big so that it's impossible to run review here, but we still use PR instead of silently merging it for visibility of changes included in the release. I'd like to merge it once CI is made ready and we confirm that this PR passes that.

PR #5729, opened 2018-11-28
[DO NOT MERGE] ChainerX: Fix Travis CI documentation build failure

niboshi:
The purpose of this PR is to check a fix for Travis CI failure, where `make html` fails due to missing ChainerX objects in Python 2.

Related: #5725


hvy:
Verified and merged into the main ChainerX PR so I think we can close this one.

PR #5736, opened 2018-11-29
Change link.add_hook to return self

crcrpar:
This change makes it easier to define original models when their links have `LinkHook`.

One expected usage is below.

```python
class MyModel(chainer.Chain):
    def __init__(self, *args, **kwargs):
        super(MyModel, self).__init__()
        with self.init_scope():
            self.linear = L.Linear(10).add_hook(SomeLinkHook)
            self.conv = L.Convolution2D(10, 3, 1, pad=1).add_hook(AnotherLinkHook)
...
```

### DIFF
--- a/tests/chainer_tests/test_link_hook.py
+++ b/tests/chainer_tests/test_link_hook.py
@@ -179,11 +179,11 @@ class TestLinkHook(unittest.TestCase):
     def test_local_hook_unnamed(self):
         self._check_local_hook(None, 'MyLinkHook')
 
-    def test_link_addhook(self):
+    def test_addhook_returns_self(self):
         model, x, dot = self._create_model_and_data()
         hook = MyLinkHook()
-        model = model.add_hook(hook)
-        assert model is not None and isinstance(model, chainer.Chain)
+        ret = model.add_hook(hook)
+        assert ret is model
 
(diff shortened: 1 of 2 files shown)


tests/chainer_tests/test_link_hook.py
```diff
@@ -179,6 +179,12 @@ def test_local_hook_named(self):
     def test_local_hook_unnamed(self):
         self._check_local_hook(None, 'MyLinkHook')
 
+    def test_link_addhook(self):
```
niboshi: Let's give a more explicit name.
```suggestion
    def test_addhook_returns_self(self):
```

tests/chainer_tests/test_link_hook.py
```diff
@@ -179,6 +179,12 @@ def test_local_hook_named(self):
...
+        model, x, dot = self._create_model_and_data()
+        hook = MyLinkHook()
+        model = model.add_hook(hook)
+        assert model is not None and isinstance(model, chainer.Chain)
```
niboshi: How about check the reference identity instead of just the type?
```
ret = model.add_hook(hook)
assert ret is model
```

niboshi:
LGTM. Thanks!

PR #5737, opened 2018-11-29
fix usage of downsample_fb in resnet

### LINKED ISSUE #5700: Wrong usage of downsample_fb in ResNet
The output of model `chainer.links.ResNet50Layers` seems to be wrong in Chainer 5.0.0, in contrast to 4.4.0.

```
import chainer
import scipy.misc
model = chainer.links.ResNet50Layers()
img = scipy.misc.imread("/some/image.jpg")
prep = img.copy()
y=(prep.shape[0]-224)//2
x=(prep.shape[1]-224)//2
prep = prep[y:y+224,x:x+224,::-1].astype(np.float32)
prep -= [103.063, 115.903, 123.152]
prep *= 1
prep = np.transpose(prep, (2, 0, 1))
chainer.config.use_cudnn='never'
chainer.config.train = False

rets=model(prep[np.newaxis, ...],model.available_layers)
for k, v in rets.items():
    print(k)
    print(v.data.flatten()[:100])
```

The output of above code differs between Chainer 5.0.0 and Chainer 4.4.0.
Also, validation accuracy for each version is 62% and 70% on the pretrained model.

In Chainer 5.0.0, I can get the same result as 4.4.0 by `chainer.links.ResNet50Layers(downsample_fb=True)`.

It seems that stride contradicts to the documentation.
https://github.com/chainer/chainer/blob/ee0b971547de243077f3afe98bbee45db630de6a/chainer/links/model/vision/resnet.py#L619

It should be
```
stride_1x1, stride_3x3 = (1, stride) if downsample_fb else (stride, 1)
```

```
Platform: Linux-4.4.0-64-generic-x86_64-with-debian-jessie-sid
Chainer: 5.0.0
NumPy: 1.12.1
CuPy:
  CuPy Version          : 5.0.0
  CUDA Build Version    : 9000
  CUDA Driver Version   : 9010
  CUDA Runtime Version  : 9000
  cuDNN Build Version   : 7300
  cuDNN Version         : 7300
  NCCL Build Version    : None
iDeep: Not Available
```

milhidaka:
This fixes #5700.

Test `tests/chainer_tests/links_tests/model_tests/test_vision.py` passes regardless of this fix.
Is there any good idea to add test without need of pretrained model and image?

### DIFF
--- a/chainer/links/model/vision/resnet.py
+++ b/chainer/links/model/vision/resnet.py
@@ -616,7 +616,7 @@ class BottleneckA(link.Chain):
         # In the original MSRA ResNet, stride=2 is on 1x1 convolution.
         # In Facebook ResNet, stride=2 is on 3x3 convolution.
 
-        stride_1x1, stride_3x3 = (stride, 1) if downsample_fb else (1, stride)
+        stride_1x1, stride_3x3 = (1, stride) if downsample_fb else (stride, 1)
         with self.init_scope():
             self.conv1 = Convolution2D(
                 in_channels, mid_channels, 1, stride_1x1, 0, initialW=initialW,


toslunar:
It's also fine to leave it untested.

toslunar:
Jenkins, test this please.

PR #5739, opened 2018-11-29
fix side effect of MultiprocessParallelUpdater

### LINKED ISSUE #5738: MultiprocessParallelUpdater modifies `devices` argument as side effect
The `MultiprocessParallelUpdater` init method has `devices` argument that takes as input a list, a tuple or a dictionnary.

In the latter case, the creator pops the `'main'` key from it, thus modifies the object given as argument.
https://github.com/chainer/chainer/blob/12340007efee499ddcb53120938cc23c8d96c5f9/chainer/training/updaters/multiprocess_parallel_updater.py#L173-L176

If the user wants to reuse the devices dictionary later (e.g. when specifying `Evaluator`'s device), the user has to create a copy of it. 
It is in a way inconsistent with the behaviour of `ParallelUpdater` (which does not modify the `devices` dictionary), and can confuse the user.

I think the `MultiprocessParallelUpdater` should copy the `devices` argument in case of dictionary to avoid this.
What are your insights about this?

Guriido:
Fixes #5738

### DIFF
--- a/chainer/training/updaters/multiprocess_parallel_updater.py
+++ b/chainer/training/updaters/multiprocess_parallel_updater.py
@@ -171,6 +171,7 @@ class MultiprocessParallelUpdater(standard_updater.StandardUpdater):
         )
 
         if isinstance(devices, dict):
+            devices = devices.copy()
             main = devices.pop('main')
             devices = list(six.itervalues(devices))
             devices = [main] + devices


kmaehashi:
Thanks! Jenkins, test this please.

kmaehashi:
Thanks, LGTM!

PR #5743, opened 2018-11-29
Build ChainerX documentation on ReadTheDocs

kmaehashi:
See https://docs.readthedocs.io/en/latest/yaml-config.html for the reference of RTD YAML file.


kmaehashi:
I confirmed the docs can build successfully.

docs/environment.yml
```diff
@@ -0,0 +1,6 @@
+# Conda environment definition to build documentation on RTD
```
niboshi: How about something like `rtd_conda.yml` to make it more explicit about the purpose?

readthedocs.yml
```diff
@@ -1,5 +1,8 @@
 name: chainer
 type: sphinx
 base: docs/source
+conda:
+  file: docs/environment.yml
 python:
+  version: 3
```
niboshi: How about `3.5`?

setup.py
```diff
@@ -187,6 +187,10 @@ def find_any_distribution(pkgs):
 
 
 build_chainerx = 0 != int(os.getenv('CHAINER_BUILD_CHAINERX', '0'))
+if os.getenv('READTHEDOCS', None):
```
niboshi: How about `== 'True'`?

kmaehashi:
Jenkins, test this please.

hvy:
LGTM

PR #5744, opened 2018-11-29
Fixes for clang

niboshi:
Checked without CUDA backend with:
* clang version 5.0.1-4
* clang version 6.0.0-1ubuntu2

### DIFF
--- a/chainerx_cc/CMakeLists.txt
+++ b/chainerx_cc/CMakeLists.txt
@@ -108,6 +108,21 @@ if("${CMAKE_CXX_COMPILER_ID}" MATCHES "GNU|Clang|Intel")
         set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=thread")
     endif()
 endif()
+
+# Compiler-specific workarounds
+
+if(CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+    EXECUTE_PROCESS(COMMAND ${CMAKE_CXX_COMPILER} --version OUTPUT_VARIABLE clang_full_version_string)
+    string(REGEX REPLACE ".*clang version ([0-9]+\\.[0-9]+).*" "\\1" CLANG_VERSION_STRING ${clang_full_version_string})
+
+    if(CLANG_VERSION_STRING VERSION_LESS 6.0)
+        # clang<6.0 suggests superfluous braces for std::array initialization.
+        # https://bugs.llvm.org/show_bug.cgi?id=21629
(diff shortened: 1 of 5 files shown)


chainerx_cc/chainerx/routines/manipulation_test.cc
```diff
@@ -839,7 +839,9 @@ TEST_THREAD_SAFE_P(ManipulationTest, SplitDefaultAxesIndices) {
     Array e1 = testing::BuildArray({1, 4}).WithData<int32_t>({0, 1, 2, 3});
     Array e2 = testing::BuildArray({1, 4}).WithData<int32_t>({4, 5, 6, 7});
 
-    Run([&]() { testing::CheckForward([](const std::vector<Array>& xs) { return Split(xs[0], {2}); }, {a}, {e1, e2}); });
+    Run([&]() {
+        testing::CheckForward([](const std::vector<Array>& xs) { return Split(xs[0], std::vector<int64_t>{1}); }, {a}, {e1, e2});
```
asi1024: Typo.
```suggestion
        testing::CheckForward([](const std::vector<Array>& xs) { return Split(xs[0], std::vector<int64_t>{2}); }, {a}, {e1, e2});
```

niboshi: This fix is intentional.
Please check the shapes.

hvy:
Have we decided which versions of clang to support?

hvy:
Note that it still fails to build with Apple clang. This conversion (to the return type defined by the signature) seems invalid. Could you take a look? https://github.com/chainer/chainer/blob/7f539ef53c3b2ff22e4203ea7f08250889e29545/chainerx_cc/chainerx/native/native_device/memory.cc#L18

hvy:
Sorry forgot this comment. Could you address it? It LGTM otherwise. https://github.com/chainer/chainer/pull/5744#pullrequestreview-181162212

chainerx_cc/CMakeLists.txt
```diff
@@ -108,6 +108,21 @@ if("${CMAKE_CXX_COMPILER_ID}" MATCHES "GNU|Clang|Intel")
         set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=thread")
     endif()
 endif()
+
+# Compiler-specific workarounds
+
```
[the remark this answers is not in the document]
hvy: Discussed offline to leave it as is since we're expecting other workarounds to be placed here in the future.

hvy:
LGTM

PR #5747, opened 2018-11-29
ChainerX backward w.r.t. inputs (C++ chainerx.grad) 

hvy:
This PR introduces `void Backward(const std::vector<ConstArrayRef>& inputs, ...)` which is similar to `chainer.grad` but does not return anything but instead sets the gradients of the specified inputs. It's better keep the behavior consistent with the previous `Backward` since a graph is consumed, in contrast to Chainer which allows repeated backprops. Basically, a function to extract a subgraph is introduced and this subgraph is checked for at certain places in backward to skip computations that are not required.

1. Traverse backwards (from given outputs) to create an ArrayNode -> OpNode mapping where the ArrayNode is an input to the OpNode.
2. Traverse forwards (from given inputs) to extract a subgraph (OpNode -> boolean "input included in subgraph" flags).
3. Then standard `Backward` but
3.a. A backward entry is skipped if none of the inputs are involved in the subgraph.
3.b After a backward entry is backproped, gradients are ignored if it is not included in the subgraph.
3.c. There are no optimizations **within** a backward entry. This would e.g. require some additional index arguments to the backward lambdas.

1 and 2 are similar to the Chainer implementation of `chainer.grad`.

The signature of the new `Backward` is open for discussion. `inputs` should maybe be the second/last argument?

### DIFF
--- a/chainerx_cc/chainerx/backward.cc
+++ b/chainerx_cc/chainerx/backward.cc
@@ -3,7 +3,9 @@
 #include <algorithm>
 #include <functional>
 #include <memory>
+#include <type_traits>
 #include <unordered_map>
+#include <unordered_set>
 #include <utility>
 #include <vector>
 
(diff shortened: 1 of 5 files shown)


niboshi:
I haven't looked into the implementation in detail but regarding the interface, I think we should not give the empty inputs `{}` the special meaning like this. Considering the behavior of non-empty cases, users would expect it to behave as no-op.

hvy:
Discussed offline within the team and decided to align the behavior to `chainer.grad`. I.e. to not modify the grads of the inputs and return them instead.

hvy:
jenkins, test this please.

hvy:
jenkins, test this please.

hvy:
Test failure seems unrelated. PTAL.

niboshi:
LGTM

PR #5759, opened 2018-12-03
Fix style of `setup.py`

toslunar:


toslunar:
https://github.com/chainer/chainer/tree/chainerx-docs seems ready (https://travis-ci.org/chainer/chainer/builds/462707799).

PR #5760, opened 2018-12-03
Implement BFC algorithm

asi1024:
Implemented BFC algorithm in ChainerX CUDA memory pool.


chainerx_cc/chainerx/cuda/memory_pool.cc
```diff
@@ -37,109 +37,245 @@ MallocStatus PinnedMemoryAllocator::Malloc(void** ptr, size_t bytesize) {
...
+        void* ptr{nullptr};
         CudaSetDeviceScope scope{device_index_};
         MallocStatus status = allocator_->Malloc(&ptr, allocation_size);
         if (status == MallocStatus::kErrorMemoryAllocation) {
             FreeUnusedBlocks();
             status = allocator_->Malloc(&ptr, allocation_size);
             if (status == MallocStatus::kErrorMemoryAllocation) {
                 // TODO(sonots): Include total pooled bytes in the error message
-                throw OutOfMemoryError{bytesize};
+                throw OutOfMemoryError{allocation_size};
```
niboshi: Why this change? I think `bytesize` is what this class tries to allocate. `allocation_size` does not mean anything to the caller.

chainerx_cc/chainerx/cuda/memory_pool.cc
```diff
@@ -37,109 +37,245 @@ MallocStatus PinnedMemoryAllocator::Malloc(void** ptr, size_t bytesize) {
...
         in_use_.erase(it);
     }
 
+    CHAINERX_ASSERT(chunk != nullptr);
     {
         std::lock_guard<std::mutex> lock{free_bins_mutex_};
-        if (free_bins_.size() <= index) {
-            free_bins_.resize(index + 1);
+
+        if (chunk->next() != nullptr) {
```
niboshi: Please add comment before this line, like
```
// If there are free chunks before/after this chunk, merge it with them.
```

chainerx_cc/chainerx/cuda/memory_pool.cc
```diff
@@ -37,109 +37,245 @@ MallocStatus PinnedMemoryAllocator::Malloc(void** ptr, size_t bytesize) {
...
+        if (chunk->prev() != nullptr) {
+            std::unique_ptr<Chunk> chunk_prev = RemoveChunkFromFreeList(chunk->prev());
+            if (chunk_prev != nullptr) {
+                chunk_prev->MergeWithNext();
+                chunk = std::move(chunk_prev);
+            }
         }
-        std::vector<void*>& free_list = free_bins_[index];
-        free_list.emplace_back(ptr);
+        PushIntoFreeList(std::move(chunk));
```
niboshi: Add a blank line and a comment before this line, like
```
// Push this chunk into the free list.
```

chainerx_cc/chainerx/cuda/memory_pool.h
```diff
@@ -19,6 +20,7 @@ class MemoryPoolTest;  // for unit-tests
 }  // namespace cuda_internal
 
 constexpr size_t kAllocationUnitSize = 512;
+constexpr size_t kCompactionThreashold = 512;
```
niboshi: Write a comment.

chainerx_cc/chainerx/cuda/memory_pool.cc
```diff
@@ -37,109 +37,245 @@ MallocStatus PinnedMemoryAllocator::Malloc(void** ptr, size_t bytesize) {
...
+// Pushes a chunk into an appropriate free list
+//
+// Not thread-safe
+void MemoryPool::PushIntoFreeList(std::unique_ptr<Chunk> chunk) {
+    FreeList& free_list = free_bins_[chunk->bytesize()];
+    free_list.emplace_back(std::move(chunk));
+}
+
+void MemoryPool::CompactFreeBins(std::map<size_t, FreeList>::iterator it_start, std::map<size_t, FreeList>::iterator it_end) {
+    for (;; --it_start) {
```
niboshi: How about rewriting this loop with `std::find_if` and `std::make_reverse_iterator`?

chainerx_cc/chainerx/cuda/memory_pool.cc
```diff
@@ -37,109 +37,245 @@ MallocStatus PinnedMemoryAllocator::Malloc(void** ptr, size_t bytesize) {
...
+
+// Pushes a chunk into an appropriate free list
+//
+// Not thread-safe
+void MemoryPool::PushIntoFreeList(std::unique_ptr<Chunk> chunk) {
+    FreeList& free_list = free_bins_[chunk->bytesize()];
+    free_list.emplace_back(std::move(chunk));
+}
+
+void MemoryPool::CompactFreeBins(std::map<size_t, FreeList>::iterator it_start, std::map<size_t, FreeList>::iterator it_end) {
```
niboshi: This function is a bit too difficult to understand.
Please write more comments, perhaps before each of the loops and `erase`.

chainerx_cc/chainerx/cuda/memory_pool.cc
```diff
@@ -37,109 +37,245 @@ MallocStatus PinnedMemoryAllocator::Malloc(void** ptr, size_t bytesize) {
...
+        FreeList& free_list = it_start->second;
+        if (!free_list.empty()) {
+            ++it_start;
+            break;
+        }
+        if (it_start == free_bins_.begin()) {
+            break;
+        }
+    }
+    for (; it_end != free_bins_.end(); ++it_end) {
```
niboshi: Ditt. How about `std::find_if`?

chainerx_cc/chainerx/cuda/memory_pool.h
```diff
@@ -74,10 +117,17 @@ class MemoryPool {
 private:
     friend class cuda_internal::MemoryPoolTest;  // for unit-tests
 
+    // Rounds up the memory size to fit memory alignment of cudaMalloc.
+    size_t GetAllocationSize(size_t bytesize) { return ((bytesize + kAllocationUnitSize - 1) / kAllocationUnitSize) * kAllocationUnitSize; }
+    void PushIntoFreeList(std::unique_ptr<Chunk> chunk);
+    std::unique_ptr<Chunk> PopFromFreeList(size_t allocation_size);
+    std::unique_ptr<Chunk> RemoveChunkFromFreeList(Chunk* chunk);
+    void CompactFreeBins(std::map<size_t, FreeList>::iterator it_start, std::map<size_t, FreeList>::iterator it_end);
```
niboshi: Write the expected behavior as a comment.

chainerx_cc/chainerx/cuda/memory_pool.cc
```diff
@@ -37,109 +37,245 @@ MallocStatus PinnedMemoryAllocator::Malloc(void** ptr, size_t bytesize) {
...
+
+    // Find an appropriate free list
+    auto free_bins_it = free_bins_.find(chunk->bytesize());
+    if (free_bins_it == free_bins_.end()) {
+        return nullptr;
+    }
+    FreeList& free_list = free_bins_it->second;
+
+    // Remove the given chunk from the found free list
+    for (auto it = free_list.begin(); it != free_list.end(); ++it) {
```
niboshi: How about `std::find_if`?

chainerx_cc/chainerx/cuda/memory_pool.cc
```diff
@@ -37,109 +37,245 @@ MallocStatus PinnedMemoryAllocator::Malloc(void** ptr, size_t bytesize) {
...
 }
 
 void MemoryPool::FreeUnusedBlocks() {
     CudaSetDeviceScope scope{device_index_};
-    std::lock_guard<std::mutex> lock{free_bins_mutex_};
-    for (const std::vector<void*>& free_list : free_bins_) {
-        for (void* ptr : free_list) {
-            allocator_->Free(ptr);
+
+    for (auto free_bins_it = free_bins_.begin(); free_bins_it != free_bins_.end();) {
```
niboshi: How about performing compaction in a separate loop? That way I think we can make use of range-based for loop, making the code much easier to read.

chainerx_cc/chainerx/cuda/memory_pool.cc
```diff
@@ -37,109 +37,245 @@ MallocStatus PinnedMemoryAllocator::Malloc(void** ptr, size_t bytesize) {
...
         }
+        // Compact free_bins
+        if (free_list.empty()) {
+            free_bins_it = free_bins_.erase(free_bins_it);
+        } else {
+            ++free_bins_it;
+        }
+    }
+    if (free_bins_.empty()) {
+        free_bins_.clear();
```
niboshi: Isn't it no-op if it's already empty?

chainerx_cc/chainerx/cuda/memory_pool.cc
asi1024: It may not be good to compact `free_bins_` by erasing empty lists, because it costs O(n^2) time complexity to compact each free list (n is the length of free list).
I think it is better to make new free bins and replace older one with it.

chainerx_cc/chainerx/cuda/memory_pool.h
```diff
@@ -74,10 +117,17 @@ class MemoryPool {
 private:
     friend class cuda_internal::MemoryPoolTest;  // for unit-tests
 
+    // Rounds up the memory size to fit memory alignment of cudaMalloc.
+    size_t GetAllocationSize(size_t bytesize) { return ((bytesize + kAllocationUnitSize - 1) / kAllocationUnitSize) * kAllocationUnitSize; }
```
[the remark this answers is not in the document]
asi1024: This function prevents `free_bins_` being too huge size.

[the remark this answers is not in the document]
niboshi: Yes, I understand that. But it does not sound related to CUDA.

chainerx_cc/chainerx/cuda/memory_pool.cc
niboshi: `allocation_size` is just an internal implementation. It does not represent any meaningful number to users. It does not necessarily represent a low-level allocation size either (which is passed to cudaMemAlloc for example), because allocators are abstracted.

Presenting this number could confuse users. (Actually I was confused when I saw this number).

asi1024:
jenkins, test this please.

chainerx_cc/chainerx/cuda/memory_pool.cc
niboshi: Hmm, sorry what do you mean? What I meant was to write this after the current loop.

```
std::remove_if(
    free_bins_.begin(),
    free_bins_.end(),
    [](const auto& pair) { return pair.second.empty(); });
```

asi1024: `std::remove_if` cannot erase the elements of `std::map` because `std::pair<const Key, T>` is not move-assignable type.

asi1024: `std::remove_if` cannot be used, but range-based-for loop can be used.  I fixed.

chainerx_cc/chainerx/cuda/memory_pool.cc
asi1024: Fixed.

chainerx_cc/chainerx/cuda/memory_pool.cc
```diff
@@ -37,66 +39,174 @@ MallocStatus PinnedMemoryAllocator::Malloc(void** ptr, size_t bytesize) {
...
     CudaSetDeviceScope scope{device_index_};
-    std::lock_guard<std::mutex> lock{free_bins_mutex_};
-    for (const std::vector<void*>& free_list : free_bins_) {
-        for (void* ptr : free_list) {
-            allocator_->Free(ptr);
+
+    // Frees unused memory blocks
+    for (auto& pair : free_bins_) {
+        FreeList& free_list = pair.second;
+        FreeList new_free_list;
```
niboshi: This requires extra malloc for the new vector. I don't think that's worthwhile.
Rather than that, how about this logic:
```
for (auto& chunk: free_list) {
    if (...) {
        allocator_->Free(chunk->ptr());
        chunk.reset();
    }
}
free_list.erase(
    std::remove_if(free_list.begin(), free_list.end(), [](const auto& chunk) { return chunk == nullptr; }),
    free_list.end());
```

chainerx_cc/chainerx/cuda/memory_pool.cc
```diff
@@ -107,39 +217,57 @@ void* MemoryPool::Malloc(size_t bytesize) {
...
         in_use_.erase(it);
     }
 
+    CHAINERX_ASSERT(chunk != nullptr);
     {
         std::lock_guard<std::mutex> lock{free_bins_mutex_};
-        if (free_bins_.size() <= index) {
-            free_bins_.resize(index + 1);
+
+        // If there are free chunks before this chunk, merges it with them.
```
sonots: after?

chainerx_cc/chainerx/cuda/memory_pool.cc
```diff
@@ -107,39 +217,57 @@ void* MemoryPool::Malloc(size_t bytesize) {
...
+            std::unique_ptr<Chunk> chunk_prev = RemoveChunkFromFreeList(chunk->prev());
+            if (chunk_prev != nullptr) {
+                chunk_prev->MergeWithNext();
+                chunk = std::move(chunk_prev);
+            }
         }
-        std::vector<void*>& free_list = free_bins_[index];
-        free_list.emplace_back(ptr);
+
+        // Pushes this chunk into the free list.
```
sonots: NITPICK: This comment says same thing with the function name. So, I prefer to remove the redundant information, but this is just my preference.

chainerx_cc/chainerx/cuda/memory_pool.cc
sonots: This logic here merges with only one next chunk, not chunks.

chainerx_cc/chainerx/cuda/memory_pool.h
```diff
@@ -53,13 +57,56 @@ class PinnedMemoryAllocator : public Allocator {
...
 
+// A chunk that points to a device memory.
+//
+// A chunk might be a splitted memory block from a larger allocation.
+// The prev/next pointers construct a doubly-linked list of contiguous memories
+// sorted by those base addresses.
+class Chunk {
+public:
+    Chunk(void* ptr, size_t offset, size_t bytesize)
+        : ptr_{reinterpret_cast<void*>(reinterpret_cast<intptr_t>(ptr) + offset)}, bytesize_(bytesize) {}
```
sonots: Maybe, `NOLINT(cppcoreguidelines-pro-type-reinterpret-cast)` is needed here, too.

chainerx_cc/chainerx/cuda/memory_pool.h
```diff
@@ -53,13 +57,56 @@ class PinnedMemoryAllocator : public Allocator {
     void Free(void* ptr) noexcept override { cudaFreeHost(ptr); }
 };
 
+// A chunk that points to a device memory.
+//
+// A chunk might be a splitted memory block from a larger allocation.
+// The prev/next pointers construct a doubly-linked list of contiguous memories
+// sorted by those base addresses.
+class Chunk {
```
sonots: I should've pointed before, but this class should be cuda_internal?

chainerx_cc/chainerx/cuda/memory_pool.cc
```diff
@@ -37,66 +39,174 @@ MallocStatus PinnedMemoryAllocator::Malloc(void** ptr, size_t bytesize) {
...
+    free_bins_.erase(it_start, it_end);
+}
+
+// Finds best-fit, or a smallest larger allocation if available
+//
+// Not thread-safe
+std::unique_ptr<Chunk> MemoryPool::PopFromFreeList(size_t allocation_size) {
+    auto it_start = free_bins_.lower_bound(allocation_size);
+    size_t distance{0};
+    for (auto it = it_start; it != free_bins_.end(); ++it, ++distance) {
```
niboshi: This loop could also written with `std::find_if` and `std::distance`. The complexity of `std::distance` is O(N), so it would not increase.

chainerx_cc/chainerx/cuda/memory_pool.cc
```diff
@@ -37,66 +39,174 @@ MallocStatus PinnedMemoryAllocator::Malloc(void** ptr, size_t bytesize) {
...
+    // Find an appropriate free list
+    auto free_bins_it = free_bins_.find(chunk->bytesize());
+    if (free_bins_it == free_bins_.end()) {
+        return nullptr;
+    }
+    FreeList& free_list = free_bins_it->second;
+
+    // Remove the given chunk from the found free list
+    auto it = std::find_if(free_list.begin(), free_list.end(), [chunk](const auto& ptr) { return ptr.get() == chunk; });
+    if (it == free_list.end()) return nullptr;
```
niboshi: Add `{ }` around the return statement.

chainerx_cc/chainerx/cuda/memory_pool.cc
```diff
@@ -37,66 +39,174 @@ MallocStatus PinnedMemoryAllocator::Malloc(void** ptr, size_t bytesize) {
...
+
+    // Find an appropriate free list
+    auto free_bins_it = free_bins_.find(chunk->bytesize());
+    if (free_bins_it == free_bins_.end()) {
+        return nullptr;
+    }
+    FreeList& free_list = free_bins_it->second;
+
+    // Remove the given chunk from the found free list
+    auto it = std::find_if(free_list.begin(), free_list.end(), [chunk](const auto& ptr) { return ptr.get() == chunk; });
```
niboshi: `auto` -> How about typing explicitly?

chainerx_cc/chainerx/cuda/memory_pool.cc
```diff
@@ -37,66 +39,174 @@ MallocStatus PinnedMemoryAllocator::Malloc(void** ptr, size_t bytesize) {
...
+    if (it == free_list.end()) return nullptr;
+    std::unique_ptr<Chunk> chunk_uniq_ptr = std::move(*it);
```
niboshi: A nitpick, but how about `removed_chunk` instead of `chunk_uniq_ptr`?
It's clear from the type it's a unique pointer. It's better to encode the logic instead.

chainerx_cc/chainerx/cuda/memory_pool.cc
```diff
@@ -37,66 +39,174 @@ MallocStatus PinnedMemoryAllocator::Malloc(void** ptr, size_t bytesize) {
...
 MemoryPool::~MemoryPool() {
     // NOTE: CudaSetDeviceScope is not available at dtor because it may throw
     int orig_device_index{0};
     cudaGetDevice(&orig_device_index);
     cudaSetDevice(device_index_);
 
-    for (const std::vector<void*>& free_list : free_bins_) {
-        for (void* ptr : free_list) {
-            allocator_->Free(ptr);
+    for (auto& item : free_bins_) {
```
niboshi: How about typing explicitly?

chainerx_cc/chainerx/cuda/memory_pool.cc
```diff
@@ -37,66 +39,174 @@ MallocStatus PinnedMemoryAllocator::Malloc(void** ptr, size_t bytesize) {
...
-    std::lock_guard<std::mutex> lock{free_bins_mutex_};
-    for (const std::vector<void*>& free_list : free_bins_) {
-        for (void* ptr : free_list) {
-            allocator_->Free(ptr);
+
+    // Frees unused memory blocks
+    for (auto& pair : free_bins_) {
+        FreeList& free_list = pair.second;
+        FreeList new_free_list;
+        for (auto& chunk : free_list) {
```
niboshi: How about typing explicitly?

chainerx_cc/chainerx/cuda/memory_pool.cc
```diff
@@ -107,39 +217,57 @@ void* MemoryPool::Malloc(size_t bytesize) {
                 throw OutOfMemoryError{bytesize};
             }
         }
+        chunk = std::make_unique<Chunk>(ptr, 0, allocation_size);
```
niboshi: how about `CHAINERX_ASSERT(ptr != nullptr)` before this line?

chainerx_cc/chainerx/cuda/memory_pool.cc
```diff
@@ -107,39 +217,57 @@ void* MemoryPool::Malloc(size_t bytesize) {
...
     }
 
+    CHAINERX_ASSERT(chunk != nullptr);
+    void* ptr = chunk->ptr();
```
niboshi: The variable name duplicates with the previous one. How about something different, e.g. `chunk_ptr`?

chainerx_cc/chainerx/cuda/memory_pool.h
niboshi: `bytesize_{bytesize}`

chainerx_cc/chainerx/cuda/memory_pool.cc
asi1024: Hmm, I think `const std::unique_ptr<cuda_internal::Chunk>&` is a little bit long...

chainerx_cc/chainerx/cuda/memory_pool.h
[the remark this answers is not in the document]
sonots: CUDA aligns to multiples of 512 (when Iimplemented BFC in CuPy, so maybe CUDA8?).
You can see it with cudaMalloc(1) twice and prints the address of ptrs.

But, it might be changed in the future, say, to multiples of 1024.
kAllocationUnitSize should be changed to 1024 in such future.
Better to change this comment to describe this fact.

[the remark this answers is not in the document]
sonots: Of course, increasing kAllocationUnitSize more is another choice, but at least, decreasing kAllocationUnitSize to 256 or less is nonsense because of the fact.

chainerx_cc/chainerx/cuda/memory_pool.cc
```diff
@@ -37,66 +39,178 @@ MallocStatus PinnedMemoryAllocator::Malloc(void** ptr, size_t bytesize) {
...
 MemoryPool::~MemoryPool() {
     // NOTE: CudaSetDeviceScope is not available at dtor because it may throw
     int orig_device_index{0};
     cudaGetDevice(&orig_device_index);
     cudaSetDevice(device_index_);
 
-    for (const std::vector<void*>& free_list : free_bins_) {
-        for (void* ptr : free_list) {
-            allocator_->Free(ptr);
+    for (std::pair<const size_t, cuda_internal::FreeList>& pair : free_bins_) {
```
sonots: It is encouraged to use auto in ranged-for loop https://google.github.io/styleguide/cppguide.html#auto

asi1024: `std::pair<const size_t, cuda_internal::FreeList>` appearants fequently, it might be better to alias this type.

chainerx_cc/chainerx/cuda/memory_pool.cc
```diff
@@ -37,66 +39,178 @@ MallocStatus PinnedMemoryAllocator::Malloc(void** ptr, size_t bytesize) {
...
         }
     }
     // Ideally, in_use_ should be empty, but it could happen that shared ptrs to memories allocated
     // by this memory pool are released after this memory pool is destructed.
     // Our approach is that we anyway free CUDA memories held by this memory pool here in such case.
     // Operators of arrays holding such memories will be broken, but are not supported.
-    for (const auto& item : in_use_) {
-        void* ptr = item.first;
-        allocator_->Free(ptr);
+    for (const std::pair<void* const, std::unique_ptr<cuda_internal::Chunk>>& pair : in_use_) {
```
sonots: ditto

chainerx_cc/chainerx/cuda/memory_pool.cc
```diff
@@ -37,66 +39,178 @@ MallocStatus PinnedMemoryAllocator::Malloc(void** ptr, size_t bytesize) {
...
 
 void MemoryPool::FreeUnusedBlocks() {
     CudaSetDeviceScope scope{device_index_};
-    std::lock_guard<std::mutex> lock{free_bins_mutex_};
-    for (const std::vector<void*>& free_list : free_bins_) {
-        for (void* ptr : free_list) {
-            allocator_->Free(ptr);
+
+    // Frees unused memory blocks
+    for (std::pair<const size_t, cuda_internal::FreeList>& pair : free_bins_) {
```
sonots: ditto

chainerx_cc/chainerx/cuda/memory_pool.cc
```diff
@@ -37,66 +39,178 @@ MallocStatus PinnedMemoryAllocator::Malloc(void** ptr, size_t bytesize) {
...
+        cuda_internal::FreeList& free_list = pair.second;
+        for (std::unique_ptr<cuda_internal::Chunk>& chunk : free_list) {
```
sonots: ditto

chainerx_cc/chainerx/cuda/memory_pool.cc
sonots: It might be another choice, but here we aliases the second with a local variable, so auto does not decrease readability. With this sense, I agree with the google guide (using auto is fine enough). But, I respect your decision.

asi1024: With this sense, using auto is indeed fine enough, but this type appears also in a parameter of lambda expression (currently, written with `auto`), so I want to define an alias of this type and use it in this ranged-for loop.

sonots: okay

chainerx_cc/chainerx/cuda/memory_pool.h
```diff
@@ -74,10 +126,21 @@ class MemoryPool {
...
+    // Rounds up the memory size to fit memory alignment of memory allocation.
+    size_t GetAllocationSize(size_t bytesize) { return ((bytesize + kAllocationUnitSize - 1) / kAllocationUnitSize) * kAllocationUnitSize; }
+    void PushIntoFreeList(std::unique_ptr<cuda_internal::Chunk> chunk);
+    std::unique_ptr<cuda_internal::Chunk> PopFromFreeList(size_t allocation_size);
+    std::unique_ptr<cuda_internal::Chunk> RemoveChunkFromFreeList(cuda_internal::Chunk* chunk);
+
+    // Finds the longest consecutive empty free lists that include the section between `it_start` and `it_end`, and removes them from free
+    // bins.
+    void CompactFreeBins(
+            std::map<size_t, cuda_internal::FreeList>::iterator it_start, std::map<size_t, cuda_internal::FreeList>::iterator it_end);
```
sonots: Use FreeBinsMap here.

chainerx_cc/chainerx/cuda/memory_pool.h
```diff
@@ -74,10 +126,21 @@ class MemoryPool {
...
+
     int device_index_;
     std::unique_ptr<Allocator> allocator_;
-    std::unordered_map<void*, size_t> in_use_;
-    std::vector<std::vector<void*>> free_bins_;
+    std::unordered_map<void*, std::unique_ptr<cuda_internal::Chunk>> in_use_;  // ptr => cuda_internal::Chunk
+    std::map<size_t, cuda_internal::FreeList> free_bins_;  // allocation size => cuda_internal::FreeList
```
sonots: ditto?

chainerx_cc/chainerx/cuda/memory_pool.cc
```diff
@@ -37,66 +39,179 @@ MallocStatus PinnedMemoryAllocator::Malloc(void** ptr, size_t bytesize) {
...
+//
+// Not thread-safe
+void MemoryPool::PushIntoFreeList(std::unique_ptr<cuda_internal::Chunk> chunk) {
+    cuda_internal::FreeList& free_list = free_bins_[chunk->bytesize()];
+    free_list.emplace_back(std::move(chunk));
+}
+
+void MemoryPool::CompactFreeBins(FreeBinsMap::iterator it_start, FreeBinsMap::iterator it_end) {
+    auto it_start_rev = std::make_reverse_iterator(it_start);
+    it_start = std::find_if(it_start_rev, free_bins_.rend(), [](const FreeBinsMap::value_type& p) { return !p.second.empty(); }).base();
```
sonots: `p.second` is not obvious on its type. How about aliasing to FreeList as a local variable as we do in ranged-for loop.

chainerx_cc/chainerx/cuda/memory_pool.cc
```diff
@@ -37,66 +39,179 @@ MallocStatus PinnedMemoryAllocator::Malloc(void** ptr, size_t bytesize) {
...
+// Pushes a chunk into an appropriate free list
+//
+// Not thread-safe
+void MemoryPool::PushIntoFreeList(std::unique_ptr<cuda_internal::Chunk> chunk) {
+    cuda_internal::FreeList& free_list = free_bins_[chunk->bytesize()];
+    free_list.emplace_back(std::move(chunk));
+}
+
+void MemoryPool::CompactFreeBins(FreeBinsMap::iterator it_start, FreeBinsMap::iterator it_end) {
+    auto it_start_rev = std::make_reverse_iterator(it_start);
```
sonots: You probably prefer `FreeBinsMap::iterator` here?

chainerx_cc/chainerx/cuda/memory_pool.cc
```diff
@@ -37,66 +39,179 @@ MallocStatus PinnedMemoryAllocator::Malloc(void** ptr, size_t bytesize) {
...
+
+    // Find an appropriate free list
+    FreeBinsMap::iterator free_bins_it = free_bins_.find(chunk->bytesize());
+    if (free_bins_it == free_bins_.end()) {
+        return nullptr;
+    }
+    cuda_internal::FreeList& free_list = free_bins_it->second;
+
+    // Remove the given chunk from the found free list
+    auto it = std::find_if(free_list.begin(), free_list.end(), [chunk](const auto& ptr) { return ptr.get() == chunk; });
```
sonots: You probably prefer `FreeList::iterator` here?

chainerx_cc/chainerx/cuda/memory_pool.cc
```diff
@@ -37,66 +39,179 @@ MallocStatus PinnedMemoryAllocator::Malloc(void** ptr, size_t bytesize) {
...
+// Not thread-safe
+void MemoryPool::PushIntoFreeList(std::unique_ptr<cuda_internal::Chunk> chunk) {
+    cuda_internal::FreeList& free_list = free_bins_[chunk->bytesize()];
+    free_list.emplace_back(std::move(chunk));
+}
+
+void MemoryPool::CompactFreeBins(FreeBinsMap::iterator it_start, FreeBinsMap::iterator it_end) {
+    auto it_start_rev = std::make_reverse_iterator(it_start);
+    it_start = std::find_if(it_start_rev, free_bins_.rend(), [](const FreeBinsMap::value_type& p) { return !p.second.empty(); }).base();
+    it_end = std::find_if(it_end, free_bins_.end(), [](const FreeBinsMap::value_type& p) { return !p.second.empty(); });
```
sonots: ditto

chainerx_cc/chainerx/cuda/memory_pool.cc
```diff
@@ -37,66 +39,179 @@ MallocStatus PinnedMemoryAllocator::Malloc(void** ptr, size_t bytesize) {
...
+    free_bins_.erase(it_start, it_end);
+}
+
+// Finds best-fit, or a smallest larger allocation if available
+//
+// Not thread-safe
+std::unique_ptr<cuda_internal::Chunk> MemoryPool::PopFromFreeList(size_t allocation_size) {
+    FreeBinsMap::iterator it_start = free_bins_.lower_bound(allocation_size);
+    FreeBinsMap::iterator non_empty_it =
+            std::find_if(it_start, free_bins_.end(), [](const FreeBinsMap::value_type& pair) { return !pair.second.empty(); });
```
sonots: ditto

chainerx_cc/chainerx/cuda/memory_pool.cc
sonots: Okay, but after reviewing, I felt same sense can be applied for lambda, too.
`pair->second` is not obvious, so we should alias it as a local variable, then `auto& pair` looked fine enough.

Anyway, I respect your decision.

chainerx_cc/chainerx/cuda/memory_pool.cc
sonots: Also, I feel `pair` improves readability than `p`.

chainerx_cc/chainerx/cuda/memory_pool.cc
```diff
@@ -107,39 +222,57 @@ void* MemoryPool::Malloc(size_t bytesize) {
...
         in_use_.erase(it);
     }
 
+    CHAINERX_ASSERT(chunk != nullptr);
     {
         std::lock_guard<std::mutex> lock{free_bins_mutex_};
-        if (free_bins_.size() <= index) {
-            free_bins_.resize(index + 1);
+
+        // If the next chunk is free, merges it with them.
```
sonots: not `them`

chainerx_cc/chainerx/cuda/memory_pool.cc
```diff
@@ -107,39 +222,57 @@ void* MemoryPool::Malloc(size_t bytesize) {
...
+        if (chunk->next() != nullptr) {
+            std::unique_ptr<cuda_internal::Chunk> chunk_next = RemoveChunkFromFreeList(chunk->next());
+            if (chunk_next != nullptr) {
+                chunk->MergeWithNext();
+            }
         }
-        std::vector<void*>& free_list = free_bins_[index];
-        free_list.emplace_back(ptr);
+
+        // If the previous chunk is free, merges it with them.
```
sonots: not `them`.

chainerx_cc/chainerx/cuda/memory_pool.cc
```diff
@@ -37,66 +39,179 @@ MallocStatus PinnedMemoryAllocator::Malloc(void** ptr, size_t bytesize) {
...
+        }
+        free_list.erase(std::remove(free_list.begin(), free_list.end(), nullptr), free_list.end());
+    }
+
+    // Erase empty free lists from free bins.
+    for (FreeBinsMap::iterator free_bins_it = free_bins_.begin(); free_bins_it != free_bins_.end();) {
+        if (free_bins_it->second.empty()) {
+            free_bins_it = free_bins_.erase(free_bins_it);
+        } else {
+            free_bins_it++;
```
sonots: ++free_bins_it

chainerx_cc/chainerx/cuda/memory_pool.cc
```diff
@@ -258,15 +258,15 @@ void MemoryPool::Free(void* ptr) {
     {
         std::lock_guard<std::mutex> lock{free_bins_mutex_};
 
-        // If the next chunk is free, merges it with them.
+        // If the next chunk is free, merges it with it.
```
sonots: I don't understand `it with it` 😅

asi1024: `merges them`?

chainerx_cc/chainerx/cuda/memory_pool.cc
```diff
@@ -37,66 +39,179 @@ MallocStatus PinnedMemoryAllocator::Malloc(void** ptr, size_t bytesize) {
...
+    FreeBinsMap::iterator non_empty_it =
+            std::find_if(it_start, free_bins_.end(), [](const FreeBinsMap::value_type& pair) { return !pair.second.empty(); });
+    if (non_empty_it == free_bins_.end()) {
+        return nullptr;
+    }
+    cuda_internal::FreeList& free_list = non_empty_it->second;
+    std::unique_ptr<cuda_internal::Chunk> chunk = std::move(free_list.back());
+    CHAINERX_ASSERT(chunk != nullptr);
+    free_list.pop_back();
+    if (static_cast<size_t>(std::distance(it_start, non_empty_it)) > kCompactionThreashold) {
```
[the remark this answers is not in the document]
asi1024: This line is correct because the range [it_start, non_emtpy_it) represents a half-closed interval.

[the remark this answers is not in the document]
asi1024: Sorry, The condition should be `>= kCompactionThreashold`.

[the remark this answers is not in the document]
sonots: Ah, okay.

chainerx_cc/chainerx/cuda/memory_pool.cc
sonots: `merge them` or `merges the chunk with the next chunk`?

chainerx_cc/chainerx/cuda/memory_pool.h
```diff
@@ -53,13 +57,65 @@ class PinnedMemoryAllocator : public Allocator {
...
+    size_t bytesize_{0};  // Chunk bytesize.
+    Chunk* prev_{nullptr};  // Prev memory pointer if splitted from a larger allocation
+    Chunk* next_{nullptr};  // Next memory pointer if splitted from a larger allocation
+};
+
+using FreeList = std::vector<std::unique_ptr<Chunk>>;  // List of free chunks with the same size.
+
+}  // namespace cuda_internal
+
+namespace {
```
sonots: Unnamed scope is not available in header.

sonots: https://ttsuki.github.io/styleguide/cppguide.ja.html#Unnamed_Namespaces_and_Static_Variables

chainerx_cc/chainerx/cuda/memory_pool_test.cc
```diff
@@ -12,14 +12,94 @@ namespace cuda_internal {
 
 class MemoryPoolTest {
 public:
-    static const std::vector<std::vector<void*>>& GetFreeBins(const MemoryPool& pool) { return pool.free_bins_; }
+    static const std::map<size_t, FreeList>& GetFreeBins(const MemoryPool& pool) { return pool.free_bins_; }
```
sonots: FreeBinsMap

sonots:
LGTM except https://github.com/chainer/chainer/pull/5760#discussion_r238944802

chainerx_cc/chainerx/cuda/memory_pool.h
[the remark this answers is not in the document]
asi1024: I think that it might be better to define two constant values: "unit size of CUDA malloc" and "unit size of MemoryPool malloc".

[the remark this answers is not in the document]
sonots: It would be nice, but "unit size of CUDA malloc" is not used in codes actually.
Writing it in comments would be enough.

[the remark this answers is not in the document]
sonots: BTW, I discovered my gist https://gist.github.com/sonots/41daaa6432b1c8b27ef782cd14064269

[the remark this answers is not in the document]
niboshi: It seems it's dependent on the underlying allocator functions.
In my environment, `cudaMallocManaged` gives the allocation unit size of 4096.
Perhaps it's better to make it determined at runtime, instead of a constant.


Maybe current implementation is fine, but could you at least add TODO?

>I think that it might be better to define two constant values: "unit size of CUDA malloc" and "unit size of MemoryPool malloc".

Write about this, too.

chainerx_cc/chainerx/cuda/memory_pool.h
```diff
@@ -18,8 +19,14 @@ class MemoryPoolTest;  // for unit-tests
 
 }  // namespace cuda_internal
 
+// TODO(imanishi): Determine this value at runtime.
+// TODO(imanishi): Define two constant values: the unit size of allocator and uint size of memory pool.
+// CUDA aligns to multiples of 512.
 constexpr size_t kAllocationUnitSize = 512;
```
niboshi: How about something like this?
```
// Allocation unit size. This value should be a multiple of the allocation unit size of underlying CUDA allocator.
// TODO(imanishi): The unit sizes of CUDA allocators are not documented.
// It may be dependent on various factors, such as CUDA runtime version or allocation functions.
// It has been observed that `cudaMallocManaged` has an allocation unit size of 4096 bytes in certain environment.
// We should revisit this number later, or perhaps better to determine it at runtime.
// TODO(imanishi): This number is currently used as both the underlying allocation unit size and the allocation unit size of the memory pool (the unit size of splitted chunks).
// We could separate them and fine-tune to optimize to the typical memory usage.
```

chainerx_cc/chainerx/cuda/memory_pool_test.cc
```diff
@@ -12,14 +12,94 @@ namespace cuda_internal {
...
 
 }  // namespace cuda_internal
 
 namespace {
 
+void* AddOffset(void* ptr, size_t offset) {
+    return reinterpret_cast<void*>(reinterpret_cast<intptr_t>(ptr) + offset);  // NOLINT(cppcoreguidelines-pro-type-reinterpret-cast)
+}
+
+TEST(ChunkTest, Split) {
```
niboshi: I'm having hard time understanding these unit tests.
Could you add some comments describing the objectives?

niboshi:
Jenkins, test this please

niboshi:
Please fix clang-tidy errors:
```
14:56:42 /repo/chainerx_cc/chainerx/cuda/memory_pool.cc:98:5: error: use auto when declaring iterators [modernize-use-auto,-warnings-as-errors]
14:56:42     FreeBinsMap::reverse_iterator it_start_rev = std::make_reverse_iterator(it_start);
14:56:42     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
14:56:42     auto
14:56:42 /repo/chainerx_cc/chainerx/cuda/memory_pool.cc:108:5: error: use auto when declaring iterators [modernize-use-auto,-warnings-as-errors]
14:56:42     FreeBinsMap::iterator it_start = free_bins_.lower_bound(allocation_size);
14:56:42     ^~~~~~~~~~~~~~~~~~~~~
14:56:42     auto
14:56:42 /repo/chainerx_cc/chainerx/cuda/memory_pool.cc:109:5: error: use auto when declaring iterators [modernize-use-auto,-warnings-as-errors]
14:56:42     FreeBinsMap::iterator non_empty_it = std::find_if(it_start, free_bins_.end(), IsFreeListNonEmpty);
14:56:42     ^~~~~~~~~~~~~~~~~~~~~
14:56:42     auto
14:56:42 /repo/chainerx_cc/chainerx/cuda/memory_pool.cc:130:5: error: use auto when declaring iterators [modernize-use-auto,-warnings-as-errors]
14:56:42     FreeBinsMap::iterator free_bins_it = free_bins_.find(chunk->bytesize());
14:56:42     ^~~~~~~~~~~~~~~~~~~~~
14:56:42     auto
14:56:42 /repo/chainerx_cc/chainerx/cuda/memory_pool.cc:137:5: error: use auto when declaring iterators [modernize-use-auto,-warnings-as-errors]
14:56:42     FreeList::iterator it =
14:56:42     ^~~~~~~~~~~~~~~~~~
14:56:42     auto
14:56:42 /repo/chainerx_cc/chainerx/cuda/memory_pool.cc:192:10: error: use auto when declaring iterators [modernize-use-auto,-warnings-as-errors]
14:56:42     for (FreeBinsMap::iterator free_bins_it = free_bins_.begin(); free_bins_it != free_bins_.end();) {
14:56:42          ^~~~~~~~~~~~~~~~~~~~~
14:56:42          auto
```

niboshi:
Jenkins, test this please

niboshi:
Failure in jenkins is irrelevant.

niboshi:
LGTM

PR #5762, opened 2018-12-03
Warn `Link.to_gpu` unless compatible with `to_device`

toslunar:
Takeover #4139.


toslunar:
I resolved the conflicts with #5986.

niboshi:
Is this feature still need?
We now have `to_device` and the original issue #4129 has been closed.

toslunar:
- `to_gpu` should be warned anyway.
- We no longer need `sticky` option.

chainer/device_resident.py
```diff
@@ -14,6 +15,29 @@
...
+        return False
+    if isinstance(dst, cuda.GpuDevice) and isinstance(arr, cuda.ndarray):
+        src_id = arr.device.id
+        dst_id = dst.device.id
+        if src_id == dst_id:
+            # The link is already on the requested device; nothing
+            # to do.
+            return True
+        else:
+            # Sticky option is omitted. As the default behavior is
```
niboshi: This comment is obsolete: there will be no "sticky" option and the behavior won't change. Please fix.

chainer/device_resident.py
```diff
@@ -14,6 +15,29 @@
...
+            # planned to be changed, raise a warning.
+            warnings.warn('''\
+You are trying to transfer a Link to GPU-{dst} which is already on GPU-{src}.
```
niboshi: It's a `DeviceResident`: not necessarily a `Link`.

chainer/device_resident.py
```diff
@@ -260,17 +289,17 @@ def _device_to_method_name_and_kwargs(self, device):
 
     def visit_array(self, arr):
         assert isinstance(arr, chainer.get_array_types())
-        if not (self._skip_between_cupy_devices
-                and self._device.xp is cuda.cupy
-                and isinstance(arr, cuda.ndarray)):
+        skip = _warn_to_gpu(
+            self._device, arr, legacy=self._skip_between_cupy_devices)
+        if not skip:
```
niboshi: It's a bit hard to understand this code.
How about this:
```py
        if (self._skip_between_cupy_devices
                and self._device.xp is cuda.cupy
                and isinstance(arr, cuda.ndarray)):
            return _warn_to_gpu(arr, self._device)  # note: argument reverted because it's more natural
        return self._device.send(arr)

File: chainer/device_resident.py [hunk 4]
```diff
@@ -14,6 +15,29 @@
 import chainerx
 
 
+def _warn_to_gpu(dst, arr, legacy):
```
niboshi: How about moving this into `_ToDeviceVisitor` as a method?
It's only used in the class.

niboshi: How about this comment?

chainer/device_resident.py
niboshi: How about moving the condition outside?
I feel the current implementation is still difficult to understand.

niboshi: Specifically, in the current implementation the condition is hidden (from the calling code) and not clear what (and when) `_warn_to_gpu` does.

toslunar:
I'm afraid if I understand the review comments well, but I addressed them.

chainer/device_resident.py
```diff
@@ -278,18 +284,38 @@ def _device_to_method_name_and_kwargs(self, device):
...
-        if not (self._skip_between_cupy_devices
-                and self._device.xp is cuda.cupy
-                and param.device.xp is cuda.cupy):
-            param.to_device(self._device)
+        if self.skip_visiting(param):
+            self._warn_to_gpu(param.array, self._device)
+            return
+        param.to_device(self._device)
+
+    def skip_visiting(self, obj):
```
niboshi: How about `_skip_visiting`?

chainer/device_resident.py
```diff
@@ -278,18 +284,38 @@ def _device_to_method_name_and_kwargs(self, device):
 
     def visit_array(self, arr):
         assert isinstance(arr, chainer.get_array_types())
-        if not (self._skip_between_cupy_devices
-                and self._device.xp is cuda.cupy
-                and isinstance(arr, cuda.ndarray)):
-            return self._device.send(arr)
-        return arr
+        if self.skip_visiting(arr):
```
toslunar:
```suggestion
        if self._skip_visiting(arr):
```

chainer/device_resident.py
```diff
@@ -278,18 +284,38 @@ def _device_to_method_name_and_kwargs(self, device):
...
+            return arr
+        return self._device.send(arr)
 
     def visit_variable(self, param):
         assert isinstance(param, chainer.Variable)
-        if not (self._skip_between_cupy_devices
-                and self._device.xp is cuda.cupy
-                and param.device.xp is cuda.cupy):
-            param.to_device(self._device)
+        if self.skip_visiting(param):
```
toslunar:
```suggestion
        if self._skip_visiting(param):
```

chainer/device_resident.py
toslunar:
```suggestion
    def _skip_visiting(self, obj):
```

toslunar:
The force-push did nothing important.  Just github.com was broken.

chainer/device_resident.py
toslunar: Done.

niboshi:
Jenkins, test this please

toslunar:
Jenkins, test this please.

niboshi:
Please check test failure:
> TestStandardUpdaterDeviceArgumentFallback_param_2_{omit_device=False, omit_input_device=True}.test_gpu_to_gpu_transfer _

niboshi:
Jenkins, test this please

PR #5768, opened 2018-12-04
Add robots.txt to hide older versions from search results

kmaehashi:
After merging this we need to setup redirect from `/robots.txt` to `/en/latest/_static/robots.txt` in Cloudflare.


okuta:
LGTM!

kmaehashi:
Deployed https://docs.chainer.org/robots.txt

Issue #5773, opened 2018-12-05
ChainerX OSX Travis failure


hvy:
OSX Travis CI is failing and should be fixed. It currently uses Apple clang which ChainerX does not yet support and it fails in `memory.cc` because of it. Please see https://github.com/chainer/chainer/pull/5744#issuecomment-444327421

ktnyt:
Just a quick note while looking to make a personal patch for this issue.
Apple LLVM linker doesn't seem to support the `--(no-)whole-archive` flag so this might need to be patched as weel.

PR #5776, opened 2018-12-05
Patch files for macOS

ktnyt:
Attempt to fix issue #5773 

Changes are as follows:
- Remove `--(no-)whole-archive` when building on macOS platforms.
- Change native memory allocation logic for macOS platforms.
- Cosmetic changes so tests build successfully. (Will revert if necessary.)


chainerx_cc/chainerx/native/native_device/memory.cc
```diff
@@ -15,7 +15,11 @@ std::shared_ptr<void> NativeDevice::Allocate(size_t bytesize) {
     if (bytesize == 0) {
         return std::shared_ptr<void>{nullptr};
     }
+#ifdef __APPLE__
+    return std::shared_ptr<uint8_t>(new uint8_t[bytesize], std::default_delete<uint8_t[]>());
```
hvy: Actually, you can skip the branching and simply replace the existing `std::make_unique` with the one you're proposing.

chainerx_cc/chainerx/testing/array.h
```diff
@@ -57,7 +57,11 @@ class ArrayBuilder {
             Strides strides = builder.GetStrides<T>();
             int64_t total_size = shape.GetTotalSize();
             size_t n_bytes = internal::GetRequiredBytes(shape, strides, sizeof(T));
+#ifdef __APPLE__
+            auto ptr = std::shared_ptr<uint8_t>(new uint8_t[n_bytes], std::default_delete<uint8_t[]>());
```
hvy: Ditto.

chainerx_cc/chainerx/CMakeLists.txt
```diff
@@ -84,20 +84,34 @@ add_library(chainerx SHARED
...
-    PUBLIC
-    ${CMAKE_DL_LIBS}
-    PRIVATE
-    -Wl,--whole-archive)
+if(${APPLE})
+  target_link_libraries(chainerx
+      PUBLIC
+      ${CMAKE_DL_LIBS}
+      PRIVATE
+      -Wl)
```
hvy: Is it okay to skip `--whole-archive` instead of using e.g. `-force_load` or `-all_load`?

chainerx_cc/chainerx/CMakeLists.txt
```diff
@@ -129,6 +143,12 @@ if(${CHAINERX_BUILD_TEST})
         stack_vector_test.cc
         strides_test.cc
         thread_local_state_test.cc
+        testing/array_check.cc
+        testing/array_test.cc
+        testing/gtest_main.cc
+        testing/routines.cc
+        testing/routines_test.cc
+        testing/util.cc
```
hvy: I'm wondering if these additions are really necessary. 🤔

hvy:
Thanks @ktnyt a lot for addressing my issue. I verified that it worked but left some comments. Would you mind taking a look at them?

ktnyt:
I'm sorry for the late response I hadn't received the notifications from GitHub somehow... I'll get to work ASAP.

chainerx_cc/chainerx/native/native_device/memory.cc
ktnyt: To be insanely picky `std::make_unique` will allocate the memory regions for the `uint8_t*` and `std::unique_ptr` simultaneously so it should in theory provide a slight performance edge (I'm not sure as to exactly how much of an impact this will have in real-life situations).

chainerx_cc/chainerx/CMakeLists.txt
ktnyt: I'm not too bright on the subject so I will have to do some research. It would be great if there is somebody who can point out a workaround for this.

chainerx_cc/chainerx/CMakeLists.txt
ktnyt: These changes were actually committed and pushed by mistake (I needed them in my local environment for the tests to compile) so I can just revert them if that would be better.

ktnyt: On second thought this seems related to the `--whole-archive` shenanigan.

ktnyt:
For the moment I removed the source code branching and unified the logic to using the `std::shared_ptr` ctor in favor over to `std::make_unique`. The test related linking issues were in fact a side effect of me removing the `--(no-)whole-archive` option from the linking step. The CMake configuration has been modified to link the libraries between the `--whole-archive` and `--no-whole-archive` using the macOS linker alternative `-force_load` and `-noall_load` instead. The linking pattern should be unaltered for non-Apple platforms but I haven't been able to test it on another machine.

chainerx_cc/chainerx/CMakeLists.txt
ktnyt: Yes indeed it was...

chainerx_cc/chainerx/CMakeLists.txt
```diff
@@ -84,20 +84,38 @@ add_library(chainerx SHARED
...
+
+  target_link_libraries(chainerx PRIVATE chainerx_routines)
+  target_link_libraries(chainerx PRIVATE chainerx_native)
+  target_link_libraries(chainerx PRIVATE chainerx_testing)
+
+  if(${CUDA_FOUND})
+      target_link_libraries(chainerx PRIVATE chainerx_cuda)
+  endif()
+
+  target_link_libraries(chainerx PRIVATE -Wl,--no-whole-archive)
```
hvy: What do you think of unifying the `target_link_libraries` as follows, for instance.
```cmake
if(${APPLE})
  target_link_libraries(chainerx
      PUBLIC
      ${CMAKE_DL_LIBS}
      PRIVATE
      -Wl,-all_load)
else()
  target_link_libraries(chainerx
      PUBLIC
      ${CMAKE_DL_LIBS}
      PRIVATE
      -Wl,--whole-archive)
endif()

target_link_libraries(chainerx PRIVATE chainerx_routines)
target_link_libraries(chainerx PRIVATE chainerx_native)
target_link_libraries(chainerx PRIVATE chainerx_testing)

if(${CUDA_FOUND})
    target_link_libraries(chainerx PRIVATE chainerx_cuda)
endif()

if(${APPLE})
  target_link_libraries(chainerx PRIVATE -Wl,-noall_load)
else()
  target_link_libraries(chainerx PRIVATE -Wl,--no-whole-archive)
endif()
```

hvy: A nitpick but could you align the indentation? It might be the case that it's not correct everywhere else but we are trying to stick to 4 spaces.

hvy:
Also, another fix has been introduced since you branched I think. Could you merge with the latest master and update this line https://github.com/chainer/chainer/blob/master/chainerx_cc/CMakeLists.txt#L114 to also check for `AppleClang`? I.e. change the line to `if(CMAKE_CXX_COMPILER_ID MATCHES "Clang|AppleClang")` ?

hvy:
jenkins, test this please.

chainerx_cc/chainerx/CMakeLists.txt
ktnyt: FWIU the `-all_load` flag will load all members from all static libraries linked to the binary and not just the ones that follow the flag. Hence the cluttering. I'm not too sure if my understanding is correct though.

From the ld manual:

```
     -all_load   Loads all members of static archive libraries.

     -ObjC       Loads all members of static archive libraries that implement an Objective-C class or category.

     -force_load path_to_archive
                 Loads all members of the specified static archive library.  Note: -all_load forces all members of all archives to be loaded.  This option allows
                 you to target a specific archive.
```

Because the documentation for `-force_load` explicitly states that it is for designating specific archives I assumed that `-all_load` will probably behave in an unwanted way.

ktnyt:
The upstream master has been merged into the branch.
I also fixed the indentation and added `AppleClang` to the specified line.

chainerx_cc/CMakeLists.txt
```diff
@@ -111,7 +111,7 @@ endif()
 
 # Compiler-specific workarounds
 
-if(CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+if(CMAKE_CXX_COMPILER_ID MATCHES "^(Clang|AppleClang)$")
```
hvy: Is the following not sufficient?
```suggestion
if(CMAKE_CXX_COMPILER_ID MATCHES "Clang|AppleClang")
```

chainerx_cc/chainerx/native/native_device/memory.cc
hvy: Ah yes but the return type of this function is a `std::shared_ptr<void>` so there're still addition allocations.

ktnyt: Oh yes that is true I don't know why that didn't occur to me...

chainerx_cc/CMakeLists.txt
ktnyt: I don't really have any preference here. I'll commit the suggestion if it's better for you.

hvy: That'd be good. For consistency with https://github.com/chainer/chainer/pull/5776/files/c6f2d9feae51ae5379e4dd0a9b8d116debd6bf6b#diff-1e5eafb3b2b05e7c616558d2154480a1R86

hvy: Sorry, please ignore this comment as `... MATCHES "Clang"` will match "AppleClang" too.

chainerx_cc/chainerx/CMakeLists.txt
hvy: Thanks for the reference. I checked the sizes which seemed to be identical in this case, but I see that `-force_load` would be preferred. Let's keep it as is.

hvy:
jenkins, test this please.

hvy:
@ktnyt I can merge this PR once CI looks green. However, it seems like some additional changes are required at the Python binding layer in order to use ChainerX via Python. Do you want me to handle those issues in a separate PR on in this one?

ktnyt:
Is it macOS specific? If that is the case I would be happy to work on it.

hvy:
Yup, if you have macOS, you can probably reproduce the error trying to build the Python module.

```shell
CHAINER_BUILD_CHAINERX=1 pip install -e .
```

ktnyt:
OK I'll check it out.

hvy:
That sounds great. If you'll get an error from gls-lite, try synching with the upstream again https://github.com/chainer/chainer/pull/5849 .

ktnyt:
I got the sources to compile. I have a number of suggestions as to what can be done to the code but I opted to just editing the binding code.
I'm having problems at the linking step where the `-flto` flag is't a linker option but compiler option and `-R` is not supported in the LLVM linker.

ktnyt:
These fixes have worked for me but I'm not too confident about the linker options.

ktnyt:
@hvy Hi! I'm sorry if I interrupted you. I was just wondering how the review on the latest commit was going.

hvy:
Sorry for the delayed response and thanks for you fixes. I'll be taking a look.

chainerx_cc/chainerx/python/routines.cc
```diff
@@ -58,7 +58,10 @@ ArrayBodyPtr MakeArrayFromBuffer(py::buffer buffer, py::handle dtype, int64_t co
         throw ChainerxError{"offset must be non-negative and no greater than buffer length (", n_bytes, ")"};
     }
 
-    if (!internal::IsContiguous(Shape{info.shape}, Strides{info.strides}, info.itemsize)) {
+    Shape info_shape{info.shape.begin(), info.shape.end()};
+    Strides info_strides{info.strides.begin(), info.strides.end()};
+
+    if (!internal::IsContiguous(info_shape, info_strides, info.itemsize)) {
```
hvy: I think you can inline `info_{shape,strides}` without giving them names (but use the same ctors) since they're not used afterwards.

chainerx_cc/chainerx/python/testing/device_buffer.cc
```diff
@@ -59,7 +59,7 @@ void InitChainerxDeviceBuffer(pybind11::module& m) {
               Dtype dtype = python_internal::GetDtype(dtype_handle);
               int64_t item_size = GetItemSize(dtype);
               int64_t bytes = item_size * total_size;
-              std::shared_ptr<void> host_data = std::make_unique<uint8_t[]>(bytes);
+              std::shared_ptr<void> host_data(new uint8_t[bytes], std::default_delete<uint8_t[]>());
```
hvy: Could you prefer list initialization, for consistency.
```suggestion
              std::shared_ptr<void> host_data{new uint8_t[bytes], std::default_delete<uint8_t[]>()};
```

hvy:
As for constructing the `info_shape` and `info_strides` I think current approach of reusing the existing ctor is completely fine since there are no performance differences (the span ctor calls begin and end internally).

chainerx_cc/chainerx/python/routines.cc
ktnyt: Yes you're right. I don't know why that didn't occur to me.

ktnyt:
I thought it might come in handy when users want to access the C++ interface directly. Performance-wise it makes no difference it would be nice if any iterable object can be given (I do acknowledge that is out of the scope of this issue though).

hvy:
As for the `-R` flag, it seems okay to skip. At least after Mac OS X 10.5, `@rpath` is available and it looks to me like it's properly appended to the list of shared object (`otool -L _core.so`).

chainerx_cc/chainerx/testing/array.h
```diff
@@ -57,7 +57,7 @@ class ArrayBuilder {
             Strides strides = builder.GetStrides<T>();
             int64_t total_size = shape.GetTotalSize();
             size_t n_bytes = internal::GetRequiredBytes(shape, strides, sizeof(T));
-            auto ptr = std::make_unique<uint8_t[]>(n_bytes);
+            auto ptr = std::shared_ptr<uint8_t>(new uint8_t[n_bytes], std::default_delete<uint8_t[]>());
```
hvy: This one as well. Sorry I forgot to send it.
```suggestion
            std::shared_ptr<uint8_t> ptr{new uint8_t[n_bytes], std::default_delete<uint8_t[]>()};
```

hvy:
I agree that the interfaces could be discussed. Would you like to file another issue for that?

hvy:
Added one more comment but it LGTM otherwise.

chainerx_cc/chainerx/python/routines.cc
```diff
@@ -58,7 +58,10 @@ ArrayBodyPtr MakeArrayFromBuffer(py::buffer buffer, py::handle dtype, int64_t co
         throw ChainerxError{"offset must be non-negative and no greater than buffer length (", n_bytes, ")"};
     }
 
-    if (!internal::IsContiguous(Shape{info.shape}, Strides{info.strides}, info.itemsize)) {
+    Shape info_shape{info.shape.begin(), info.shape.end()};
+    Strides info_strides{info.strides.begin(), info.strides.end()};
```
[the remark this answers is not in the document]
ktnyt: I completely missed it. My apologies.

hvy:
jenkins, test this please.

chainerx_cc/chainerx/python/routines.cc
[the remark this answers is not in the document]
hvy: Thanks a bunch!

chainerx_cc/chainerx/python/routines.cc
```diff
@@ -58,7 +58,7 @@ ArrayBodyPtr MakeArrayFromBuffer(py::buffer buffer, py::handle dtype, int64_t co
         throw ChainerxError{"offset must be non-negative and no greater than buffer length (", n_bytes, ")"};
     }
 
-    if (!internal::IsContiguous(Shape{info.shape}, Strides{info.strides}, info.itemsize)) {
+    if (!internal::IsContiguous(Shape{info.shape.begin(), info.shape.end()}, Strides{info.strides.begin(), info.strides.end()}, info.itemsize)) {
```
hvy: Could you fix the format of this line? Seems like a line break is missing.

hvy:
LGTM!

hvy:
@ktnyt thank a lot for the PR. It passed the CI (some unrelated failures) and it LGTM so let me merge this one.

PR #5779, opened 2018-12-05
Fix `testing.BackendConfig` context for repeated use

niboshi:
Fix `testing.BackendConfig` so that it can be used as a context manager multiple times.

Previously, the following code caused an error:
```
a = chainer.testing.BackendConfig({'use_cuda': True})
with a:
    with a:
        pass
```
```
...
  File "cupy/cuda/device.pyx", line 113, in cupy.cuda.device.Device.__exit__
IndexError: pop from empty list
```

### DIFF
--- a/chainer/testing/backend.py
+++ b/chainer/testing/backend.py
@@ -83,8 +83,6 @@ class BackendConfig(object):
         return self._device
 
     def __enter__(self):
-        if self._contexts is None:
-            self._contexts = []
         contexts = [
             chainer.using_config(
                 'use_cudnn', self.use_cudnn),
--- a/chainer/testing/backend.py
+++ b/chainer/testing/backend.py
@@ -83,7 +83,9 @@ class BackendConfig(object):
         return self._device
 
     def __enter__(self):
-        self._contexts = [
+        if self._contexts is None:
+            self._contexts = []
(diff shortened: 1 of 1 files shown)


hvy:
Would you mind recording where this is needed?

chainer/testing/backend.py
```diff
@@ -83,7 +83,9 @@ def device(self):
         return self._device
 
     def __enter__(self):
-        self._contexts = [
+        if self._contexts is None:
```
hvy: I think we can skip this check since it's initialized to a non-`None` in `__init__`.

hvy:
Wrote some minor comments, but it LGTM otherwise.

chainer/testing/backend.py
niboshi: Fixed.

niboshi:
>Would you mind recording where this is needed?

Sorry, I forgot where I noticed the bug.

hvy:
Thanks, LGTM!

hvy:
jenkins, test this please.

hvy:
jenkins, test this please.

hvy:
jenkins, test this please.

hvy:
LGTM!

PR #5787, opened 2018-12-06
Fix to test against new chainerx.GradientError

hvy:
A new error type `chainerx.GradientError` was introduced but some Python tests were not updated accordingly. This PR fixes those failing tests.


niboshi:
LGTM

PR #5788, opened 2018-12-06
Fix scalar handling of `indices_and_sections` in `chainerx.split`

### LINKED ISSUE #5785: ChainerX's split is wrong for one element np.array
Apparently, chx.split seems to think `np.array([2])` is a scalar. Repro:

```
import numpy as np
import chainerx as chx

def test(split):
    a = [x.shape for x in np.split(np.zeros((6,)), split, axis=0)]
    b = [x.shape for x in chx.split(chx.zeros((6,)), split, axis=0)]
    if a == b:
        print('OK', a, b)
    else:
        print('ERR', a, b)

test(np.array([2]))  # Error
# OK for the rest.
test(2)
test([2])
test(np.array(2))
test(np.array([2, 3]))
```

niboshi:
Fixes #5785

### DIFF
--- a/chainerx_cc/chainerx/python/routines.cc
+++ b/chainerx_cc/chainerx/python/routines.cc
@@ -410,6 +410,9 @@ void InitChainerxManipulation(pybind11::module& m) {
           py::arg("axis") = 0);
     m.def("split",
           [](const ArrayBodyPtr& ary, py::handle indices_or_sections, int8_t axis) {
+              // TODO(niboshi): Perhaps we would want more general approach to handle multi-type arguments like indices_or_sections to
+              // provide more helpful error message for users.
+
               auto split_sections = [](const ArrayBodyPtr& ary, int64_t sections, int8_t axis) {
                   return MoveArrayBodies(Split(Array{ary}, sections, axis));
               };
--- a/chainerx_cc/chainerx/python/routines.cc
+++ b/chainerx_cc/chainerx/python/routines.cc
@@ -409,13 +409,86 @@ void InitChainerxManipulation(pybind11::module& m) {
           py::arg("arrays"),
           py::arg("axis") = 0);
     m.def("split",
-          [](const ArrayBodyPtr& ary, int64_t sections, int8_t axis) { return MoveArrayBodies(Split(Array{ary}, sections, axis)); },
-          py::arg("ary"),
-          py::arg("indices_or_sections"),
-          py::arg("axis") = 0);
-    m.def("split",
-          [](const ArrayBodyPtr& ary, std::vector<int64_t> indices, int8_t axis) {
-              return MoveArrayBodies(Split(Array{ary}, indices, axis));
+          [](const ArrayBodyPtr& ary, py::handle indices_or_sections, int8_t axis) {
+              auto split_sections = [](const ArrayBodyPtr& ary, int64_t sections, int8_t axis) {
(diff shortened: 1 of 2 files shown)


chainerx_cc/chainerx/python/routines.cc
```diff
@@ -409,13 +409,86 @@ void InitChainerxManipulation(pybind11::module& m) {
...
+              // Raises ValueError if the value has non-zero fraction.
+              auto pyfloat_to_sections_or_value_error = [](py::handle num) {
+                  CHAINERX_ASSERT(py::isinstance<py::float_>(num));
+                  double num_fp = py::cast<double>(num);
+                  auto num_int = static_cast<int64_t>(num_fp);
+                  if (static_cast<double>(num_int) != num_fp) {
+                      throw py::value_error{"Sections must be an integer."};
+                  }
+                  return num_int;
+              };
```
hvy: How about defining these as free functions since they're only used conditionally?

chainerx_cc/chainerx/python/routines.cc
```diff
@@ -409,13 +409,86 @@ void InitChainerxManipulation(pybind11::module& m) {
...
+                  return split_sections(ary, sections, axis);
+              }
+              // sections: float
+              if (py::isinstance<py::float_>(indices_or_sections)) {
+                  int64_t sections = pyfloat_to_sections_or_value_error(indices_or_sections);
+                  return split_sections(ary, sections, axis);
+              }
+              // numpy.ndarray
+              if (py::isinstance<py::array>(indices_or_sections)) {
+                  py::array np = py::cast<py::array>(indices_or_sections);
```
hvy: I'm not sure about this variable name `np` since the code is quite long and you might further down wonder if it's the numpy module. How about `array` or `np_array`?

hvy:
Hm, wouldn't a third overload with `py::array` be a bit simpler, that creates a vector out of it and calls the routine? Or am I missing something.

chainerx_cc/chainerx/python/routines.cc
niboshi: I assume their cost for definition is zero because they have no capture.
I think it's better to define here because they're currently used only in this function.

chainerx_cc/chainerx/python/routines.cc
niboshi: How about `np_ios`? Both `array` and `np_array` are prone to be confused with `ary`.

niboshi:
We would need so many overloads only for `indices_and_sections`.

`int`, `float`, `list`, `tuple`, `np.ndarray`, ... perhaps `chx.ndarray` in the future.

chainerx_cc/chainerx/python/routines.cc
hvy: Maybe that's better, `np_ios` sounds good to me.

niboshi: Fixed.

hvy:
jenkins, test this please.

hvy:
Discussed offline about how to process complicated arguments such as `indices_or_sections` and also a bit about that we should better support more types to in general (not just `split`) be compatible with NumPy. https://github.com/chainer/chainer/issues/5802

chainerx_cc/chainerx/python/routines.cc
hvy: Okay, let's leave it as is.

hvy:
LGTM.

PR #5797, opened 2018-12-06
Add variable.item()

crcrpar:
This PR adds `item()` method to `chainer.Variable`.


niboshi:
What's the motivation of this method? 🤔 Isn't `var.array.item()` enough?

crcrpar:
main motivation is to provide users with one concrete and simple way to convert variables with 1 value such as variables representing loss to python scalar.

IMO, currently, it's a bit demanding and not clear to do this. ref: DictSummary

niboshi:
OK, thank you. I think we should

* mention that it may incur host-device synchronozation,
* mention that ChainerX is not supported,
* raise `NotImplementedError` with appropriate message for ChainerX.

chainer/variable.py
```diff
@@ -1424,6 +1424,18 @@ def backward(self, retain_grad=False, enable_double_backprop=False,
         with chainer.using_config('enable_backprop', enable_double_backprop):
             _backprop_to_all([self], retain_grad, loss_scale)
 
+    def item(self):
+        """Converts the variable with one element to a Python scalar.
+
+        This will incur host-device synchronization. Note that ChainerX is not
```
toslunar: ChainerX is supported :smile:

crcrpar: Thanks. fixed

tests/chainer_tests/test_variable.py
```diff
@@ -2329,6 +2329,53 @@ def test_backward_chainerx(self):
         self.check_backward(chainerx.array(self.x))
 
 
+@testing.parameterize(*testing.product({
+    'shape': [(), (1,), (1, 1), (1, 1, 1), (2,), (2, 3)],
```
toslunar: Why don't you test empty shapes (`x.size == 0`)?

crcrpar: I missed them. Thank you.

toslunar:
Jenkins, test this please.

toslunar:
The Jenkins failures are unrelated to the PR.

PR #5809, opened 2018-12-08
Fix bug when CuPy not installed and cuda.fuse decorator used without parenthesis

grafi-tt:
On an environment CuPy is not installed, the following code has caused error.

```python
import numpy as np

@cuda.fuse
def f(x, y):
    return x + y

x = np.array([1,2,3])
y = np.array([1,2,3])
print(f(x, y))
```

```
  File "fusesample.py", line 10, in <module>
    print(f(x, y))
TypeError: <lambda>() takes 1 positional argument but 2 were given
```

### DIFF
--- a/chainer/backends/cuda.py
+++ b/chainer/backends/cuda.py
@@ -670,6 +670,8 @@ def fuse(*args, **kwargs):
     """
     if available:
         return cupy.fuse(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


grafi-tt:
The test failure is a random numerical error. Not related to the code change.

asi1024:
The Jenkins test failure seems not related to this PR.

PR #5814, opened 2018-12-09
Fix error message of `testing.assert_allclose`

niboshi:
Currently `testing.assert_allclose` can show an error that is within the tolerance.
That's because the element to show is chosen by using `numpy.argmax(err)`, taking only the absolute error into account.

This PR also improves the message so that both relative and absolute errors are shown.

### DIFF
--- a/chainer/testing/array.py
+++ b/chainer/testing/array.py
@@ -33,15 +33,18 @@ def assert_allclose(x, y, atol=1e-5, rtol=1e-4, verbose=True):
             '  shape: {} {}\n'.format(x.shape, y.shape) +
             '  dtype: {} {}\n'.format(x.dtype, y.dtype))
         if x.shape == y.shape:
-            xx = x if x.ndim != 0 else x.reshape((1,))
-            yy = y if y.ndim != 0 else y.reshape((1,))
+            xx = numpy.atleast_1d(x)
+            yy = numpy.atleast_1d(y)
             err = numpy.abs(xx - yy)
-            i = numpy.unravel_index(numpy.argmax(err), err.shape)
+            tol_err = atol + rtol * numpy.abs(yy).astype(numpy.float64)
+            i = numpy.unravel_index(
+                numpy.argmax(err.astype(numpy.float64) - tol_err), err.shape)
             f.write(
                 '  i: {}\n'.format(i) +
                 '  x[i]: {}\n'.format(xx[i]) +
                 '  y[i]: {}\n'.format(yy[i]) +
-                '  err[i]: {}\n'.format(err[i]))
+                '  relative error[i]: {}\n'.format(err[i] / numpy.abs(yy[i])) +
+                '  absolute error[i]: {}\n'.format(err[i]))
         opts = numpy.get_printoptions()
         try:
             numpy.set_printoptions(threshold=10000)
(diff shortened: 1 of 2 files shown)


chainer/testing/array.py
```diff
@@ -33,15 +33,18 @@ def assert_allclose(x, y, atol=1e-5, rtol=1e-4, verbose=True):
...
         if x.shape == y.shape:
-            xx = x if x.ndim != 0 else x.reshape((1,))
-            yy = y if y.ndim != 0 else y.reshape((1,))
+            xx = numpy.atleast_1d(x)
+            yy = numpy.atleast_1d(y)
             err = numpy.abs(xx - yy)
-            i = numpy.unravel_index(numpy.argmax(err), err.shape)
+            tol_err = atol + rtol * err.astype(numpy.float64)
+            i = numpy.unravel_index(
+                numpy.argmax(err.astype(numpy.float64) - tol_err), err.shape)
```
okuta: Why do you subtract `tol_err`?

niboshi: Ah, `tol_err = atol + rtol * err.astype(numpy.float64)` should be `tol_err = atol + rtol * numpy.abs(y)` I guess.

It's according to this equation in NumPy documentation:
>absolute(a - b) <= (atol + rtol * absolute(b))

https://docs.scipy.org/doc/numpy/reference/generated/numpy.allclose.html

niboshi: >Ah, `tol_err = atol + rtol * err.astype(numpy.float64)` should be `tol_err = atol + rtol * numpy.abs(y)` I guess.

Fixed this.

okuta: Thanks

okuta:
jenkins, test this please.

toslunar:
Jenkins, test this please.

chainer/testing/array.py
```diff
@@ -33,15 +33,18 @@ def assert_allclose(x, y, atol=1e-5, rtol=1e-4, verbose=True):
...
-            i = numpy.unravel_index(numpy.argmax(err), err.shape)
+            tol_err = atol + rtol * numpy.abs(yy).astype(numpy.float64)
+            i = numpy.unravel_index(
+                numpy.argmax(err.astype(numpy.float64) - tol_err), err.shape)
             f.write(
                 '  i: {}\n'.format(i) +
                 '  x[i]: {}\n'.format(xx[i]) +
                 '  y[i]: {}\n'.format(yy[i]) +
-                '  err[i]: {}\n'.format(err[i]))
+                '  relative error[i]: {}\n'.format(err[i] / numpy.abs(yy[i])) +
```
toslunar: Zero-division at this line should be ignored.  Is there a test for the case?

niboshi: Currently there is not test at all for `assert_allclose`.
It prints `inf` if `yy = 0`.

toslunar: Should the line raise inside `with np.errstate(divide='raise')`?  (I'd rather not.)

niboshi: I see. I think we need some tests.

niboshi: Hmm, this line already causes `FloatingPointError` if `y` is small enough and `with np.errstate(under='raise')`.

https://github.com/chainer/chainer/blob/f02fde2390ab72a0e117aa6fbe4f1171b106e10b/chainer/testing/array.py#L26

niboshi: Anyway, added a test for zero division particularly.

PR #5745, opened 2018-11-29
`FunctionNode` automatic fallback of array attributes in forward

niboshi:
In ChainerX FunctionNode fallback, type mismatch will happen if the function assigns an array as its attribute.
This PR implements automatic conversion of such attributes.

PR #5816, opened 2018-12-10
`L.NegativeSampling` ChainerX support

niboshi:
Fixes `L.NegativeSampling` for ChainerX.

### DIFF
--- a/chainer/backends/_chainerx.py
+++ b/chainer/backends/_chainerx.py
@@ -1,6 +1,5 @@
 import numpy
 
-import chainer
 from chainer import _backend
 from chainer.backends import _cpu
 from chainer.backends import cuda
@@ -165,11 +164,7 @@ def _array_from_chainerx(array):
     if array is None:
         return None
     if not isinstance(array, chainerx.ndarray):
-        if isinstance(array, chainer.get_array_types()):
-            return array
-        raise TypeError(
-            'Tried to convert to a non-ChainerX array from an invalid type: '
-            '{}'.format(type(array)))
+        return array
 
     backend_name = array.device.backend.name
     if backend_name == 'native':
--- a/chainer/backends/_chainerx.py
+++ b/chainer/backends/_chainerx.py
@@ -1,5 +1,6 @@
 import numpy
 
+import chainer
 from chainer import _backend
 from chainer.backends import _cpu
 from chainer.backends import cuda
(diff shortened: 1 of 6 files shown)


niboshi:
`TestNegativeSampling.test_backward_compare_with_numpy` causes a strange failure due to invalid dtype.
I susupect that's caused from `using_config('dtype', ...)` in `setUp`/`tearDown` pair: if the test is skipped, the `tearDown` is never called.

niboshi:
Well, now the problem is gone for some reason...

The problem I mentioned above, that `tearDown` is not called if the test is skipped, does not seem persistent. The `--pdb` option of `pytest` could possibly be the fault: without `--pdb` the teardown seems to be called no matter whether the test is skipped, but with it, the opposite.

tests/chainer_tests/links_tests/loss_tests/test_negative_sampling.py
```diff
@@ -4,27 +4,28 @@
 
 import chainer
 from chainer import backend
-from chainer.backends import cuda
+from chainer.backends import _cpu
```
hvy: How about using `chainer.backend.CpuDevice.send` for cpu transfer preferring public APIs?

chainer/backends/_chainerx.py
```diff
@@ -165,11 +164,7 @@ def _array_from_chainerx(array):
     if array is None:
         return None
     if not isinstance(array, chainerx.ndarray):
-        if isinstance(array, chainer.get_array_types()):
-            return array
-        raise TypeError(
-            'Tried to convert to a non-ChainerX array from an invalid type: '
-            '{}'.format(type(array)))
+        return array
```
hvy: I see this behavior is convenient (for slices?) but it'll stop being symmetrical to "to_chainerx". Is it difficult to keep the old behavior and check the type before calling this function?

chainer/functions/array/get_item.py
```diff
@@ -38,7 +38,7 @@ def check_type_forward(self, in_types):
         type_check._argname(in_types, ('x',))
 
     def forward(self, xs):
-        return utils.force_array(xs[0][self.slices]),
+        return utils.force_array(xs[0][backend.from_chainerx(self.slices)]),
```
hvy: I now feel the attribute fallback https://github.com/chainer/chainer/pull/5745 wasn't robust enough. Don't you think we should extend it to support contained ndarrays too?

asi1024: Beyond the scope of this PR.

asi1024:
jenkins, test this please.

asi1024:
jenkins, test this please.

hvy:
CI failure seems unrelated, LGTM (and @asi1024 ).

PR #5817, opened 2018-12-10
Simplify `F.batch_renormalization` test 

niboshi:
Applies the function test case https://github.com/chainer/chainer/pull/3499.

~`chainer.Function` also needs automatic attribute fallback #5745 for F.batch_renormalization to support ChainerX.~

`chainer.Function` attribute fallback is implemented in #5828.

### DIFF
--- a/tests/chainer_tests/functions_tests/normalization_tests/test_batch_renormalization.py
+++ b/tests/chainer_tests/functions_tests/normalization_tests/test_batch_renormalization.py
@@ -140,7 +140,7 @@ class TestBatchRenormalizationForward(testing.FunctionTestCase):
         assert isinstance(gy, chainer.Variable)
 
         import chainerx
-        if isinstance(x.array, chainerx.ndarray):
+        if x.xp is chainerx:
             # TODO(niboshi): ChainerX does not support grad yet
             y.grad = gy.array.copy()
             y.backward()
--- a/tests/chainer_tests/functions_tests/normalization_tests/test_batch_renormalization.py
+++ b/tests/chainer_tests/functions_tests/normalization_tests/test_batch_renormalization.py
@@ -1,142 +1,218 @@
-import unittest
-
 import numpy
 import six
 
 import chainer
-from chainer.backends import cuda
 from chainer.functions.normalization import batch_renormalization
-from chainer import gradient_check
 from chainer import testing
-from chainer.testing import attr
-from chainer.testing import condition
-
-
-def _batch_renormalization(expander, gamma, beta, x, mean, var, r, d):
-    mean = mean[expander]
-    std = numpy.sqrt(var)[expander]
-    y_expect = (gamma[expander] * ((x - mean) / std * r + d) + beta[expander])
-    return y_expect
 
 
 # naive implementation of differentiable batch renormalization
 def _naive_batch_renormalization(
-        x, gamma, beta, rmax, dmax, eps, avg_mean, avg_std, axis):
-    shape = x.shape
-    stat_shape = list(shape)
-    for i in axis:
-        stat_shape[i] = 1
-    stat_shape = tuple(stat_shape)
-    gamma = chainer.functions.reshape(gamma, stat_shape)
-    beta = chainer.functions.reshape(beta, stat_shape)
-    avg_mean = avg_mean.reshape(stat_shape)
-    avg_std = avg_std.reshape(stat_shape)
+        x, gamma, beta,  # variables
+        mean, var,  # variables
+        running_mean, running_var,  # arrays
+        rmax, dmax, eps, decay):
+    # If decay is not None, the running stats are updated.
 
-    mean = chainer.functions.mean(x, axis=axis, keepdims=True)
-    std = chainer.functions.sqrt(
-        eps +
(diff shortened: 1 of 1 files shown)


hvy:
jenkins, test this please.

niboshi:
Rebased, and removed float16 skip.

tests/chainer_tests/functions_tests/normalization_tests/test_batch_renormalization.py
```diff
@@ -4,215 +4,267 @@
...
+            {'use_chainerx': True, 'chainerx_device': 'cuda:1'},
+        ])
+
+
+@parameterize_batch_renormalization()
+@inject_backend_tests_batch_renormalization()
+class TestBatchRenormalizationForward(testing.FunctionTestCase):
+    # As F.batch_renormalization includes a calculation in which the outcome
+    # depends on x but it's gradient-stopped w.r.t. x,
+    # gradient_check.check_backword cannot be used, because numerical
```
crcrpar: nit-picking: `check_backword` -> `check_backward`

niboshi: Thanks!

hvy:
@niboshi could you fix the style error importing `unittest` which is no longer needed?

niboshi:
Oh, fixed.

tests/chainer_tests/functions_tests/normalization_tests/test_batch_renormalization.py
```diff
@@ -1,218 +1,268 @@
...
-                'dtype': numpy.float64, 'atol': 1e-3, 'rtol': 1e-2}
-
-    def _forward(self, *args):
+            self.check_forward_options.update({'atol': 5e-3, 'rtol': 1e-2})
+            self.check_backward_options.update({'atol': 5e-3, 'rtol': 1e-2})
+            self.check_double_backward_options.update(
+                {'atol': 5e-3, 'rtol': 1e-2})
+
+    def generate_inputs(self):
+        d = 3
```
hvy: A detail but how about using a different variable name such as `channels` since `d` is used in the batch renormalization algorithm for a different meaning.

niboshi:
@hvy Fixed the reviewed points.

tests/chainer_tests/functions_tests/normalization_tests/test_batch_renormalization.py
```diff
@@ -1,218 +1,268 @@
...
+
+    def _compute_backward(self, x, gamma, beta, y, gy):
+        assert isinstance(x, chainer.Variable)
+        assert isinstance(gamma, chainer.Variable)
+        assert isinstance(beta, chainer.Variable)
+        assert isinstance(y, chainer.Variable)
+        assert isinstance(gy, chainer.Variable)
+
+        import chainerx
+        if isinstance(x.array, chainerx.ndarray):
```
hvy: A detail but is this a "cleaner" way to check?
```suggestion
        if x.xp is chainerx:
```

tests/chainer_tests/functions_tests/normalization_tests/test_batch_renormalization.py
```diff
@@ -1,218 +1,268 @@
...
-    'update_statistics': [True, False],
-})))
-class TestBatchRenormalization(unittest.TestCase):
+        x, gamma, beta,  # variables
+        mean, var,  # variables
+        running_mean, running_var,  # arrays
+        rmax, dmax, eps, decay):
+    # If decay is not None, the running stats are updated.
+
+    F = chainer.functions
```
hvy: I don't think this declaration is needed since "F" is only used once within this scope.

hvy:
Sorry, basically LGTM but I added some minor comments. Could you take a look?

hvy:
jenkins, test this please.

tests/chainer_tests/functions_tests/normalization_tests/test_batch_renormalization.py
niboshi: It's for readability. The number of occurrences should not matter.

hvy: It's a detail but this way of declaring "F" is rather uncommon and it's declaration is several lines aways from its usage which doesn't really speak for readability IMHO.

hvy:
jenkins, test this please.

tests/chainer_tests/functions_tests/normalization_tests/test_batch_renormalization.py
toslunar: - To me, `F` is so common that it's readable in Chainer.
- Using `F.batch_renormalization` and `F.sqrt` with `import chainer.functions as F` is another option because there's no issue with circular dependency in the test.

hvy: Let me merge it as is. I think it's mostly a cosmetic matter. I agree though that it'd be fine or maybe cleaner to declare "F" at the top.

hvy:
LGTM!

PR #5821, opened 2018-12-10
CuPy/ChainerX memory pool sharing

hvy:
Currently, CuPy and ChainerX both implement and use their own memory pools which leads to poor utilization of the available GPU memory. This PR introduces a mechanism to share the same pool in order to avoid it.

Requires ~https://github.com/cupy/cupy/pull/1883~ ~https://github.com/cupy/cupy/pull/1904~


asi1024:
jenkins, test this please.

asi1024:
jenkins, test this please.

hvy:
jenkins, test this please.

hvy:
jenkins, test this please.

hvy:
jenkins, test this please.

hvy:
jenkins, test this please.

hvy:
jenkins, test this please.

hvy:
jenkins, test this please.

hvy:
jenkins, test this please.

hvy:
jenkins, test this please.

asi1024:
jenkins, test this please.

hvy:
jenkins, test this please.

hvy:
jenkins, test this please.

hvy:
jenkins, test this please.

hvy:
jenkins, test this please.

hvy:
jenkins, test this please.

niboshi:
We will separate CUDA backend from ChainerX very soon.
Instead of integrating CUDA-specific interface in ChainerX, shouldn't we consider a way to allow backends to expose their own features dynamically?

hvy:
I agree that that's a cleaner approach. I don't think on the other hand that generalizing the exposure of allocation routines is really worth the effort if we're likely to do an overhauling change of how backend-specific features are exposed more in general, in the near future. An options is of course to do that first.

niboshi:
I don't get the point. What do you mean by the "overhauling change"?

hvy:
What I meant with "overhauling change" was, introducing a mechanism to expose backend-specific features in some general way, not limited to malloc and free. What you wrote here that is.
> Instead of integrating CUDA-specific interface in ChainerX, shouldn't we consider a way to allow backends to expose their own features dynamically?

hvy:
Trying to generalize this PR to other backends, without doing the "overhauling change".

hvy:
Ah, no. It's not worthwhile, that's what I meant.

niboshi:
Hmm, I'm asking WHY, not WHAT.

>Trying to generalize this PR to other backends, without doing the "overhauling change".

But wait, what do you mean by "without"?
You said the "overhauling change" refers to the "generalization":

>What I meant with "overhauling change" was, introducing a mechanism to expose backend-specific features in some general way, not limited to malloc and free. What you wrote here that is.

I'm confused...

hvy:
Depends on https://github.com/cupy/cupy/pull/1958

hvy:
jenkins, test this please.

asi1024:
jenkins, test this please.

hvy:
jenkins, test this please.

hvy:
jenkins, test this please.

hvy:
jenkins, test this please.

chainerx/_cuda.py
```diff
@@ -0,0 +1,38 @@
...
+except Exception:
+    _available = False
+
+try:
+    import cupy
+    _cupy_available = True
+except Exception:
+    _cupy_available = False
+
+_chainerx_allocator = None
```
niboshi: I think this variable is not needed.

chainerx/_cuda.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def cupy_share_allocator(owner=chainerx._global_context):
```
niboshi: Is there any reason to make `owner` customizable?

chainerx_cc/chainerx/python/cuda/cuda_module.cc
```diff
@@ -0,0 +1,55 @@
...
+
+std::pair<intptr_t, intptr_t> GetBackendMallocFreePtrs() {
+    return std::make_pair(
+            reinterpret_cast<intptr_t>(&Malloc), reinterpret_cast<intptr_t>(&Free));  // NOLINT(cppcoreguidelines-cppcoreguidelines);
+}
+
+void InitChainerxCudaModule(py::module& m) {
+    // Exposes pointers to objects and functions so that external sources can allocate CUDA memory via ChainerX.
+    m.def("get_backend_ptr", []() -> intptr_t { return GetBackendPtr(); });
+    m.def("get_backend_malloc_free_ptrs", []() -> std::pair<intptr_t, intptr_t> { return GetBackendMallocFreePtrs(); });
```
niboshi: Using the backend pointer as `param` is an implementation detail of `malloc` and `free`.
I think it's better to avoid exposing "`backend_ptr`" in the function name.
Also the "param" and malloc/free pointers are tightly coupled.
I think it's better to make them in a single function.
How about something like
`get_c_allocator() -> tuple[param: void*, malloc_ptr: void*, free_ptr: void*]"`

chainerx_cc/chainerx/python/cuda/cuda_module.cc
```diff
@@ -0,0 +1,55 @@
...
+namespace chainerx {
+namespace python {
+namespace cuda {
+namespace cuda_internal {
+
+namespace py = pybind11;
+
+using CudaDevice = chainerx::cuda::CudaDevice;
+
+void* Malloc(void* backend, size_t bytesize, int device_id) {
```
niboshi: Better to avoid `backend` as the parameter name. (because it being a backend is an implementation detail of `Malloc` and `Free`.)

chainerx_cc/chainerx/python/cuda/cuda_module.cc
```diff
@@ -0,0 +1,55 @@
...
+    CudaDevice* device = dynamic_cast<CudaDevice*>(
```
niboshi: Reference is preferable in this case:
`CudaDevice& device = dynamic_cast<CudaDevice&>(reinterpret_cast<..>(..)->GetDevice(..));`

https://en.cppreference.com/w/cpp/language/dynamic_cast
>If the cast is successful, dynamic_cast returns a value of type new_type. If the cast fails and new_type is a pointer type, it returns a null pointer of that type. If the cast fails and new_type is a reference type, it throws an exception that matches a handler of type std::bad_cast.

chainerx/_cuda.py
hvy: I don't have a strong opinion since contexts aren't tested much but I see a couple of reasons. Do you have any opinions to why it should not be?
- As this feature is "experimental" I think it's okay to allow some degree of freedom. Of course, the degree can be discussed, but my understanding is that it is difficult at this point. 
- It default to the default global context and does not make it more difficult to use.
- To allow users to create new contexts and still use this feature.

chainerx/_cuda.py
hvy: Must not the instance of `CFunctionAllocator` be kept alive? Only its `malloc` is passed to `set_allocator`.

chainerx_cc/chainerx/python/cuda/cuda_module.cc
hvy: What do you think would be a suitable name? It looks to me like a requirement by the caller rather than an implementation detail though.

chainerx_cc/chainerx/python/cuda/cuda_module.cc
hvy: Thanks for the commend and reference, fixed.

chainerx_cc/chainerx/python/cuda/cuda_module.cc
hvy: Fixed as suggested.

chainerx/_cuda.py
niboshi: I think it's kept alive via the bound method `malloc`.

chainerx/_cuda.py
niboshi: I believe the CUDA backend instance is *always* used as the owner and there will never be a requirement to change that.

chainerx_cc/chainerx/python/cuda/cuda_module.cc
niboshi: I prefer `param`.
The requirement is to pass the `param` returned by `get_c_allocator()`, not matter what it actually represents.

chainerx/_cuda.py
hvy: Ah it seems to work. Fixed.

chainerx_cc/chainerx/python/cuda/cuda_module.cc
hvy: I see, I changed as your suggestion above.

chainerx/_cuda.py
niboshi: Yes. What's the concern?
Even if it's the context instead of the backend, my argument above does not change (no need to customize it).

hvy: Okay I removed it as you suggested.

chainerx/__init__.py
```diff
@@ -51,6 +52,10 @@
     from chainerx import _docs
     _docs.set_docs()
 
+    from chainerx import _cuda
+    # Share memory pool with CuPy.
+    if bool(int(os.getenv('CHAINERX_CUDA_CUPY_SHARE_ALLOCATOR', '0'))):
+        _cuda.cupy_share_allocator(_global_context)
```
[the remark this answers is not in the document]
hvy: Sorry, removed the argument.

niboshi:
Jenkins, test this please

hvy:
Jenkins, test this please.

hvy:
The lack of documentation is addressed in https://github.com/chainer/chainer/pull/6017.

niboshi:
LGTM

PR #5825, opened 2018-12-10
[backport] Fix bug when CuPy not installed and cuda.fuse decorator used without parenthesis

asi1024:
Backport of #5809


okuta:
LGTM!

PR #5640, opened 2018-11-08
Support unretained `Variable`s in `_check_grad_type`

toslunar:
Close #5608.

PR #5826, opened 2018-12-10
[backport] Support unretained `Variable`s in `_check_grad_type`

takagi:
Backport of #5640


takagi:
Jenkins, test this please.

takagi:
Jenkins, test this please.

toslunar:
I'd like to know if it's OK to add checks to the stable version.  This backport PR makes a check stricter, while it leaves `None` and `Variable(None)` unchecked.  Some users might intentionally set `grad` with different `dtype` or `shape` but such code fails anyway unless all the relevant variables are unretained.

takagi:
I suppose it is Okay because we can make use of "Undocumented behaviors" section of API Compatibility Policy in the reference. Additionally, the case you mention is theoretically possible but not a natural usage, so we could easily claim that it would be out of our responsibility.

PR #5828, opened 2018-12-11
Automatic ChainerX fallback of array attributes in `Function` 

niboshi:
ChainerX array attribute fallback for `chainer.Function` similar to #5745

### DIFF
--- a/chainer/function_node.py
+++ b/chainer/function_node.py
@@ -1,4 +1,5 @@
 import collections
+import contextlib
 import heapq
 import inspect
 import traceback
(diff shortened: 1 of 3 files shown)


chainer/function_node.py
```diff
@@ -146,7 +147,11 @@ class FunctionNode(object):
     _retained_output_data = None
     _local_function_hooks = None
     _supports_static_optimizations = False
-    _is_chainerx = None
+    # True if the function node is operating on ChainerX arrays and it falls
+    # back to NumPy/CuPy implementation.
+    _is_chainerx_fallback = False
```
asi1024: In this implementation, `False` will be set to this variable whenever `is_chainerx` is `False` (even if chainerx implementation of the function is not supported).  I think it seems a little bit confusing.  How about assign `None` or anything like that which means "unknown state" in this line (or delete this line)?

niboshi: Renamed to `_is_chainerx_fallback_mode` as discussed offline

asi1024:
jenkins, test this please.

asi1024:
LGTM.

Issue #5732, opened 2018-11-28
Allreduce in Chainermn which mixed parameters of FP16 and FP32


shu65:
To improve the performance of batch normalization,  We need to reduce the conversion of FP32 and FP16 in it.
But, ChainerMN do not support allreduce mixed parameters of FP16 and FP32. 
So, we require to support it.

shu65:
Solved by https://github.com/chainer/chainer/pull/6448 https://github.com/chainer/chainer/pull/6548

PR #5829, opened 2018-12-11
Batched pack/unpack params before/after allreduce

anaruse:
This PR aims to improve performance of multi-nodes training by reducing number of GPU kernel launched for packing/unpacking parameter gradients before/after allreduce communication among nodes. You can use this feature by just adding a argument `batched_pack_unpack=True` to `create_communicator` like below.

    comm =  chainermn.create_communicator('pure_nccl', batched_pack_unpack=True)

In addition, this also supports mixed data type of parameters (only fp16 and fp32 for now). When dtype of a parameter and dtype for allreduce are different, the dtype of parameter is automatically converted to/from allreduce dtype by GPU kernel for packing/unpacking only when needed. (this is related to #5732).

Note that this feature is only available with 'pure_nccl' communicator for now.


shu65:
Thank you! How fast will ChainerMN be by this PR?

anaruse:
> How fast will ChainerMN be by this PR?

It depends on training configuration, and this feature will be effective when local batch size is small or a network/model has lots of layers, in other word, CPU is performance limiter. For your reference, performance when training resnet50 with local batch size of 32 was as follows.

[ performance (iterations/sec) ]
  16.98  (w/o this feature)
  17.54  (with this feature)

[ testing configs ]
  GPU: 2x Tesla V100
  Model: Resnet50
  Dtype: FP16
  Autotune: Enabled
  Local batch size: 32
  IO pipeline: DALI is used
  Convolution: NHWC format is used
  Batch normalization: NHWC format is used

chainermn/communicators/_memory_utility.py
```diff
@@ -112,6 +112,170 @@ def unpack_params(params, itemsize, attr_name, buffer, stream=None):
         offset += size
 
 
+class ParamsData(object):
```
kuenishi: Could you move all these NCCL-only class and methods to `pure_nccl_communicator.py` ? Then I think you can get rid of re-implementing `_get_nccl_type_id` .

chainermn/communicators/pure_nccl_communicator.py
```diff
@@ -123,6 +127,12 @@ def _assign_for_allreduce_grad(self, grad_dtype, allreduce_grad_dtype,
 
     def _pack_params_to_buffer(self, params, grad_dtype, allreduce_grad_dtype,
                                n_elems, stream):
+        if self.batched_pack_unpack:
+            if self.params_data is None:
+                self.params_data = _memory_utility.ParamsData(params, 'grad')
```
kuenishi: Based on Chainer's design principle define-by-run, the network potentially may change at any iteration. Thus we cannot have any assumption on caching or remembering gradient arrays. Given that this params_data may have N params in iteration X, we can't assume params have N params in iteration X+1.

I'm not sure how much cost we pay at rebuilding params_data here, but there should be a mechanism to detect the grads' change of size, shape or numbers (e.g. compare n_params and all pointers to arrays), or possibly rebuilding params_data every iteration.

chainermn/communicators/__init__.py
```diff
@@ -3,7 +3,7 @@
 
 def create_communicator(
         communicator_name='hierarchical', mpi_comm=None,
-        allreduce_grad_dtype=None):
+        allreduce_grad_dtype=None, batched_pack_unpack=False):
```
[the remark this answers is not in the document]
anaruse: Got it. I will change the name to `batched_copy`.

chainermn/communicators/_memory_utility.py
anaruse: Got it. I will move some NCCL related stuff to `pure_nccl_communicator.py`.

chainermn/communicators/pure_nccl_communicator.py
anaruse: I understand your concern. But this is not for that purpose, say, sharing gradient array sizes and pointers over iteration. This is just for sharing them between `_pack_params_to_buffer` and `_unpack_params_from_buffer` at the same iteration, so that `self.params_data` is set `None` after `_unpack_params_from_buffer` in `_allreduce_grad_async`. But I think current design is a bit confusing, so will change it to avoid confusion.

anaruse:
Thank you for your feedback. I've fixed the branch based on your feedback. Could you check it again?

kuenishi:
Thank you for updating the patch. I and @shu65 discussed on this a bit, and it's almost OK, but still one thing we need further discussion on supporting float64. We tend to not to support float64, but where to put numpy type check. Inserting it as `else` clause inside cuda kernel, but it'll have issue on how to pass error from cuda to Python. It might be good to add type check before entering into cuda kernel, right before packing. Do you have any good idea?

chainermn/communicators/pure_nccl_communicator.py
kuenishi: Sorry! My eyes needed some fix.

anaruse:
Regarding type checks, how about doing that in `__init__` of `_ParamsData`? Then, you can do that with small overheads.

anaruse:
I've added type checks in object _ParamData, and also fixed some issues in GPU kernels for unpacking parameters and tests for batched_copy. Could you check it again?

chainermn/communicators/pure_nccl_communicator.py
```diff
@@ -193,3 +212,159 @@ def _get_nccl_type_id(dtype):
...
+        n_params = len(params)
+        params_dptr = np.empty(n_params, dtype=np.int64)
+        params_dtype = np.empty(n_params, dtype=np.int32)
+        params_size_csum = np.empty(n_params+1, dtype=np.int32)
+        params_size_csum[0] = 0
+        for i, param in enumerate(params):
+            v = getattr(param, attr_name)
+            params_dptr[i] = v.data.ptr
+            if v.dtype not in [np.float16, np.float32]:
+                raise ValueError('dtype musb be float16 or float32.')
```
kuenishi: Typo here :'(

anaruse: Thanks (^^;)

kuenishi:
The code looks good, but we need a bit more time to test this.

kuenishi:
Jenkins, test this please.

PR #5832, opened 2018-12-11
Converter decorator for ChainerX device support

niboshi:
~Introduces `chainer.dataset.Converter` class.~
Introduces `chainer.dataset.converter` decorator.

In the existing converter convention, users can make their custom converters by defining a function with two arguments: a batch and a device, where "device" is an integer specifying either CPU (-1) or CUDA devices (>=0).

Now Chainer has introduced `chainer.Device` class.
To support this, a new interface for converters is required becaue existing converter functions can't handle this class.

In this PR, a new decorator `chainer.dataset.converter` is being introduced. This decorator marks the subject converter to indicate that it supports `chainer.backend.Device` argument. If `int` is given when the converter is being called, the decorator converts the argument to `chainer.backend.Device` instance and passed to the subject converter.

##### (obsolete description)
~In this PR, a new class `chainer.dataset.Converter` is being introduced for Chainer to be able to distinguish between the existing converters and those aware of `chainer.Device`.~

~For ease of defining custom converters, a decorator `chainer.dataset.converter` is also being introduced.~

### DIFF
--- a/chainer/dataset/convert.py
+++ b/chainer/dataset/convert.py
@@ -1,4 +1,5 @@
 import collections
+import functools
 
 import numpy
 import six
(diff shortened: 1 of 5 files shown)


chainer/dataset/convert.py
```diff
@@ -9,6 +10,40 @@
 from chainer import utils
 
 
+class Converter(object):
```
hvy: How about documenting this class and the `convert` method?

hvy:
jenkins, test this please.

hvy:

chainer/dataset/convert.py
niboshi: Wrote docs.

hvy:
jenkins, test this please.

chainer/dataset/convert.py
```diff
@@ -9,6 +10,74 @@
...
+            return func(*args, **kwargs)
+
+        cls = type(
+            func.__name__,
+            (Converter,),
+            {
+                'convert': wrapped_converter,
+                '__call__': wrapped_call,
+            })
+        return cls()
```
beam2d: I guess that the docstring is not appropriately shown with the above code. I feel it's a bit tricky to wrap a function with a callable object which itself is not a function. How about returning a function with a special attribute to distinguish it from a non-decorated one?

niboshi: Hmm, as for the latter part, I think it's not so uncommon to convert a function to non-function with a decorator (`@cupy.fuse` and `@numba.jit` are such examples). What's going on under the hood will be clearer for users. Also, the check (`isinstance(convert, Converter)`) is much easier.

beam2d: As for the compatibility of `concat_examples`, I now think it's enough to document it in the release notes and the upgrade guide. So then, what we need now is a good document rendering.

niboshi: We can render it correctly as a function, by using a custom documenter feature in sphinx.

https://github.com/sphinx-doc/sphinx/blob/d8b69c7f98c1dade6ae1e6b1ddb35d323cf177cd/tests/test_autodoc.py#L524

niboshi: I've made a commit to enable that.
(Currently for some reason the first argument `batch` is omitted from the generated signature. We can override it manually though.)

beam2d: (Sorry for late reply) The documenter feature looks useful for us. Is it needed for users as well when they want to correctly render the documentation for their own converters? That would be better avoided.

niboshi: After offline discussion, I removed `Converter` class and introduce an internal flag to indicate new-style (decorated) converter.

niboshi:
Fixed.

chainer/dataset/convert.py
```diff
@@ -8,6 +9,81 @@
...
+    return wrap
+
+
+def _call_converter(converter, batch, device):
+    # Calls the converter.
+    # Converter can be either new-style (accepts chainer.backend.Device) or
+    # old-style (accepts int as device).
+    assert device is None or isinstance(device, backend.Device)
+
+    if getattr(converter, '__converter_protocol_version', 0) == 1:
```
niboshi: Perhaps the attribute name is too "formal"?

beam2d: I don't think versioning is useful here. Functions are not pickled, so there is no concern on compatibility nor mixed use of "versions". The attribute name would be made simpler, too.

niboshi: OK, changed the attribute name.

beam2d:
Jenkins, test this please.

chainer/dataset/convert.py
```diff
@@ -8,6 +9,81 @@
...
+
+    The type of the device argument is :class:`chainer.backend.Device`.
+
+    The types and values of the batches (the first argument and the return
+    value) are not specified: they depend on how the converter is used (e.g.
+    by updaters).
+
+    .. admonition:: Example
+
+        >>> @chainer.dataset.converter()
```
beam2d: Doctest fails here. I guess you need to use `...` as the prompt for the following lines.

beam2d:
Jenkins, test this please.

niboshi:
Ready to merge?

beam2d:
sorry for being late!

PR #5836, opened 2018-12-11
Fix gitignore to avoid ignoring some necessary files

asi1024:
Currently, `.gitignore` includes `docs/source/reference/util/*.rst` as files to be ignored.


niboshi:
LGTM

niboshi:
jenkins, test this please

PR #5840, opened 2018-12-12
Move workaround `utils._getitem`/`_setitem` to `chainerx`

niboshi:


hvy:
Is `_fallback_workarounds` missing?

niboshi:
Oops, added.

chainerx/_fallback_workarounds.py
```diff
@@ -0,0 +1,143 @@
...
+                array,
+                array.device.index),
+            0),
+        strides=array.strides)
+
+
+def _from_cupy(array):
+    if cupy is None:
+        raise RuntimeError(
+            'Currently cupy is required in this operation.')
```
hvy: A detail but you can assert that `cupy is not None`. Maybe this error should be thrown by the caller. That would be friendlier than "ChainerX fallback implementation only supports native or cuda backends.", in case cupy isn't available.

chainerx/_docs/__init__.py
```diff
@@ -10,7 +10,8 @@
 
 
 def set_doc(obj, docstring):
-    if inspect.ismethoddescriptor(obj) or inspect.isroutine(obj):
+    if ((inspect.ismethoddescriptor(obj) or inspect.isroutine(obj))
+            and not inspect.isfunction(obj)):
```
hvy: Sorry but could you explain why this change is needed?

chainerx/_fallback_workarounds.py
hvy: Okay, let's leave it here as is if it was meant to be reused. It is currently only called from `_to_chainerx` (which I mean by "caller", sorry if it was unclear) and this code is never executed so that's why I asked.

hvy:
jenkins, test this please.

hvy:
Could you check the Travis failure?

chainerx/_fallback_workarounds.py
niboshi: Ah, I see. Then I think that's better.

niboshi: Fixed as suggested.

niboshi:
Fixed.

chainerx/_fallback_workarounds.py
```diff
@@ -0,0 +1,141 @@
...
+    return chainerx.to_numpy(array, copy=False)
+
+
+def _from_numpy(array):
+    assert isinstance(array, numpy.ndarray)
+    return chainerx.array(array, copy=False)
+
+
+def _to_cupy(array):
+    assert cupy is not None
```
hvy: I think this assertion may fail (although we can assume it in `_from_cupy`), i.e. ChainerX with CUDA backend is used but CuPy is unavailable?

hvy: I must have missed the check prior to the call. This doesn't seem like an issue.

hvy:
LGTM!

hvy:
jenkins, test this please.

hvy:
jenkins, test this please.

Issue #5731, opened 2018-11-28
Improve performance of optimization step


shu65:
When using FP16 mode, one of CPU bottlenecks is an optimization step. 
The main reason is that there are many calls of CUDA kernel.

As a method to solve this, I think that it is good to concatenate the array of parameters, then to perform optimization with one CUDA kernel.
What do you think?

LennMars:
At first I will test performance of some possible solutions to reducing the number of CUDA kernels.

PR #5841, opened 2018-12-13
Support batched update of weights by optimizer

anaruse:
This PR aims to support a batched update of weights by optimizer.

Current optimizers in Chainer update weights in DL model in layer-by-layer, more specifically, parameter-by-parameter. This is good when mini-batch size is big enough, but this could be a problem in terms of performance when mini-batch size is small and CPU is a performance limiter, because lots of GPU kernels are launched with current parameter-by-parameter update approach.

The purpose of this PR is to mitigate this problem by updating whole weights in the DL model by a single GPU kernel, in other words, batched GPU kernel. With this PR, whole weights are updated in batched way by just adding a line below (after optimizer setup) in your training script.

    optimizer.use_batched_update()

Please note that "batched update" is enabled only when a optimizer has a method `batched_update()`. Currently, `batched_update()` is implemented only in momentum sgd. So, you cannot try this feature with other optimizers for now.

### DIFF
--- a/chainer/optimizer.py
+++ b/chainer/optimizer.py
@@ -680,6 +680,7 @@ class GradientMethod(Optimizer):
         super(GradientMethod, self).__init__()
         self.hyperparam = Hyperparameter()
         self._use_fp32_update = False
+        self._batched_update_rule = None
 
     def setup(self, link):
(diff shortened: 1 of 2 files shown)


anaruse:
This is related to #5731

PR #5851, opened 2018-12-14
LocalResponseNormalization -> FunctionNode implementation

ir5:
Current implementation of `LocalResponseNormalization` is based on obsolete `chainer.Function` class.
This pull request changes it to be based on `chainer.FunctionNode`.

### DIFF
--- a/chainer/functions/normalization/local_response_normalization.py
+++ b/chainer/functions/normalization/local_response_normalization.py
@@ -45,6 +45,10 @@ class LocalResponseNormalization(function_node.FunctionNode):
         self.alpha = alpha
         self.beta = beta
 
+        self.scale = None
+        self.indexes = None
+        self.unit_scale = None
+
     def check_type_forward(self, in_types):
         type_check.expect(in_types.size() == 1)
         x_type, = in_types
@@ -110,13 +114,9 @@ class LocalResponseNormalization(function_node.FunctionNode):
         y, = self.get_retained_outputs()
         gy, = grad_outputs
 
-        scale = getattr(self, 'scale', None)
-        indexes = getattr(self, 'indexes', None)
-        unit_scale = getattr(self, 'unit_scale', None)
-
         f = LocalResponseNormalizationGrad(
             self.n, self.k, self.alpha, self.beta, self._use_ideep,
-            scale, indexes, unit_scale,)
+            self.scale, self.indexes, self.unit_scale,)
         return f.apply((x, y, gy))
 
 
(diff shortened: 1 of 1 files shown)


chainer/functions/normalization/local_response_normalization.py
```diff
@@ -54,86 +54,133 @@ def check_type_forward(self, in_types):
...
+                       self.y, self.scale)
         return self.y,
 
-    def backward_cpu(self, x, gy):
+    def backward(self, indexes, grad_outputs):
+        x, = self.get_retained_inputs()
+        y, = self.get_retained_outputs()
+        gy, = grad_outputs
+
+        scale = getattr(self, 'scale', None)
```
takagi: Would you initialize these instance variables with `None` in `__init__()` for efficiency?

takagi:
Jenkins, test this please.

chainer/functions/normalization/local_response_normalization.py
ir5: Thank you for quick review. I fixed this point.

chainer/functions/normalization/local_response_normalization.py
```diff
@@ -54,86 +58,129 @@ def check_type_forward(self, in_types):
...
         cuda.elementwise(
             ' T x, T gy, T scale, T beta, T coeff', 'T gx',
             'gx = pow(scale, -beta) * gy - coeff * x * gx',
-            'lrn_bwd')(x[0], gy[0], self.scale,
+            'lrn_bwd')(x, gy, self.scale,
                        self.beta, 2 * self.alpha * self.beta, gx)
         return gx,
 
+    def backward(self, indexes, grad_outputs):
+        raise NotImplementedError
```
takagi: Let me confirm that there is no essential reason to leave its double backward not implemented?

takagi:
Jenkins, test this please.

takagi:
LGTM except a comment above!

takagi:
The CI failure is not relevant to this PR.

chainer/functions/normalization/local_response_normalization.py
ir5: Thank you for your review, again. Well, I put `NotImplementedError` here just because there was no trivial way to implement double-backward for this function.
If there is more appropriate way, I would like to follow such method.

takagi: Thanks, then would you leave a comment as such because, I think, it is even valuable information?
``` Python
# No trivial way to implement double-backward for this function.
raise NotImplementedError
```

ir5: Thank you for advice. I added the comment.

PR #5852, opened 2018-12-14
Write "Wx + b" in the document of `Linear`.

toslunar:
`xW^\top + b` is so confusing.


kmaehashi:
Jenkins, test this please.

toslunar:
`y_i = W \mathrm{vec}(x_i)` if it's worth mentioning each input `x_i` in the batch is flattened.

okuta:
LGTM!

PR #5868, opened 2018-12-19
Allow skipping linkcode in docs with environment variable

niboshi:
Currently docs cannot be built in non-editable installation.

This fix allows that by setting an environment variable.

```
$ cd docs
$ CHAINER_DOCS_SKIP_LINKCODE=1 make html
```


okuta:
LGTM!

PR #5800, opened 2018-12-06
Code enhancements: avoid array copies

niboshi:
Move whenever possible.

PR #5870, opened 2018-12-19
Fix clang-tidy error

niboshi:
`out` is a `const` variable (because it's a capture).
clang-tidy complains about moving from `const`.

Introduced in #5800.


hvy:
LGTM!

hvy:
jenkins, test this please.

PR #5871, opened 2018-12-19
Do not retain input on iDeep `F.relu`

### LINKED ISSUE #5866: iDeep F.relu retains input array unnecessarily
If iDeep is used, `F.relu` retains both of input and output during training, even though iDeep implementation of `F.leaky_relu` retains outputs only. This can cause memory exhaust.

Apart from the issue itself, it is questionable whether iDeep improves `F.relu` performance, because it is band-width limited and trivially written in NumPy. So it might be better to change

```python
intel64.should_use_ideep('>=auto')
```

to 

```python
intel64.should_use_ideep('==always')
```

grafi-tt:
Fixes #5866. Some refactoring of code style is done as well.

### DIFF
--- a/chainer/functions/activation/relu.py
+++ b/chainer/functions/activation/relu.py
@@ -24,7 +24,6 @@ class ReLU(function_node.FunctionNode):
     """Rectified Linear Unit."""
 
     _use_cudnn = False
-    _use_ideep = False
 
     def check_type_forward(self, in_types):
         type_check._argname(in_types, ('x',))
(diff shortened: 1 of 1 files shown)


niboshi:
Thank you for the PR!
The test error indeed seems unrelated (issued: #5874).
I'll take a look at the changes soon.

grafi-tt:
@niboshi Thank you for helpful feedback! I had left those codes conservatively because I wasn't confident there are unneeded. Following your feedback, I removed them.

niboshi:
Thanks.
Jenkins, test this please.

niboshi:
LGTM

PR #5660, opened 2018-11-14
Add documentation of `ndarray`

toslunar:
Takeover #4887.
Close #5498.

Changes to each doc of function will be made after the idea is reviewed.

PR #5875, opened 2018-12-19
Make docs of connection functions refer `ndarray`

toslunar:
See #5660.


niboshi:
Before making backport, let's merge #5886 first.
(Otherwise there would be some conflicts.)

PR #5876, opened 2018-12-19
Set grad of `F.clip` 1 at `x_min` and `x_max`

grafi-tt:
Chainer's `F.clip` hasn't propagate gradient at `x_min` and `x_max`. This behavior is different to TF and recent PyTorch's. (https://github.com/pytorch/pytorch/issues/7002). This PR makes it consistent with TF and PyTorch. I also added docstrings and tests.

Note that `F.clip` has retained input, though it could retain output. This PR makes it necessary that `F.clip` retains input.

### DIFF
--- a/tests/chainer_tests/functions_tests/math_tests/test_clip.py
+++ b/tests/chainer_tests/functions_tests/math_tests/test_clip.py
@@ -101,4 +101,30 @@ class TestClipInvalidInterval(unittest.TestCase):
             functions.clip(self.x, 1.0, -1.0)
 
 
+@testing.parameterize(*testing.product({
+    'dtype': [numpy.float16, numpy.float32, numpy.float64],
+}))
+class TestClipBorderGrad(unittest.TestCase):
+
+    def setUp(self):
+        self.x = numpy.arange(6, dtype=self.dtype)
+        self.x_min = 1.0
+        self.x_max = 4.0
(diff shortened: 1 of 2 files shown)


tests/chainer_tests/functions_tests/math_tests/test_clip.py
```diff
@@ -101,4 +101,30 @@ def test_invalid_interval(self):
...
+}))
+class TestClipBorderGrad(unittest.TestCase):
+
+    def setUp(self):
+        self.x = numpy.arange(6, dtype=self.dtype)
+        self.x_min = 1.0
+        self.x_max = 4.0
+        self.expected = numpy.asarray([0, 1, 1, 1, 1, 0], dtype=self.dtype)
+
+    def check_border_grad(self, x, expected):
```
toslunar: If your primary motivation is CPU/GPU consistency, could you test the consistency instead?

grafi-tt: My primary motivation of this PR is not CPU/GPU consistency. The motivation is

* consistency with other frameworks, and
* clarity of corner case behavior.

I checked [MXNet also works in this manner](https://github.com/apache/incubator-mxnet/blob/0d480fbe1534d1c1228bf2c7470018ae06cbac37/src/operator/tensor/matrix_op-inl.h#L1449-L1462), as well as TF and PyTorch.

toslunar:
Jenkins, test this please.

toslunar:

PR #5893, opened 2018-12-20
Allow mulitiple application of testing.parameterized

niboshi:
Currently `testing.parameterize` cannot be used multiple times in a single test case, because the wrap function returns `None` to erase the original class.
This fix enables it.


asi1024:
jenkins, test this please.

asi1024:
jenkins, test this please.

toslunar:
Will the PR replace tests with `chainer.testing.product_dict`?

niboshi:
@toslunar 
This PR is not intended to replace (=deprecate) `testing.product_dict`, but yes it can be used like that.

More detailed background: this is a by-product of `testing.inject_backend_tests` refactoring.
We're considering refactoring `testing.inject_backend_tests` to behave more similarly to `testing.parameterize`: generate many classes rather than methods in a single class (like it currently does).
The purpose of that refactoring is to allow inheritance in the source test case class. But because the previous `parameterize` implementation returned `None`, we wouldn't be able to combine `parameterize` and `inject_bakend_tests` without a fix like this PR.

asi1024:
jenkins, test this please.

asi1024:
LGTM. I manually confirmed that all combinations are visited, and also that duplicated attribute names are prohibited.

asi1024:
jenkins, test this please.

asi1024:
jenkins, test this please.

asi1024:
The Jenkins test failure is not related to this PR.

PR #5895, opened 2018-12-20
Accept splitting at the tail of dataset in `split_dataset`

### LINKED ISSUE #5890: `split_dataset` doesn't accept `split_at == n_examples`
https://github.com/chainer/chainer/blob/master/chainer/datasets/sub_dataset.py#L104-L105

`split_dataset` accepts `split_at=0` and the first return value will be a SubDataset with length zero.
However, it raises an exception when `split_at` is exactly the same as `n_examples`.

I found it inconvenient when I had to manually split my dataset for cross validation (I couldn't use `get_cross_validation_datasets` for some reason). I first tried to split it into three pieces and concatenate the first and third ones, but I got an error for the last fold.

Is there any reason it declines when `split_at` is equal to `n_examples`?

Code to reproduce:
```
from chainer.dataset import DatasetMixin
from chainer.datasets import split_dataset


class Dataset(DatasetMixin):

    def __len__(self):
        return 3

    def get_example(self, i):
        return 42


dataset = Dataset()

a, b = split_dataset(dataset, 0)
print('a: {}, b: {}'.format(len(a), len(b)))

a, b = split_dataset(dataset, 3)
print('a: {}, b: {}'.format(len(a), len(b)))
```

nai62:
Fixes #5890

### DIFF
--- a/chainer/datasets/sub_dataset.py
+++ b/chainer/datasets/sub_dataset.py
@@ -101,7 +101,7 @@ def split_dataset(dataset, split_at, order=None):
                         .format(type(split_at)))
     if split_at < 0:
         raise ValueError('split_at must be non-negative')
-    if split_at >= n_examples:
+    if split_at > n_examples:
         raise ValueError('split_at exceeds the dataset size')
     subset1 = SubDataset(dataset, 0, split_at, order)
     subset2 = SubDataset(dataset, split_at, n_examples, order)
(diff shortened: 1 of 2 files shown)


niboshi:
Thank you for the fix!

jenkins, test this please

niboshi:
LGTM!

Issue #5877, opened 2018-12-19
Gradient of Clipped ReLU inconsistent if cuDNN used


grafi-tt:
* OS: Linux x86_64 64bit
* CUDA version: 10.0.130
* CUDNN version: 7.4.1.5

Apparently cuDNN's clipped relu propagates a gradient at the clipping threshold. I confirmed the behavior by the following code.

```python
import chainer
import chainer.functions as F
import numpy as np
from chainer.backends import cuda

a = np.array([0, 10, 20], dtype=np.float32)

v = chainer.Variable(a)
l = F.sum(F.clipped_relu(v, z=20.0))
l.backward()
print(v.grad)

a = cuda.cupy.asarray(a)

with chainer.using_config('use_cudnn', 'never'):
    v = chainer.Variable(a)
    l = F.sum(F.clipped_relu(v, z=20.0))
    l.backward()
    print(v.grad)

with chainer.using_config('use_cudnn', 'always'):
    v = chainer.Variable(a)
    l = F.sum(F.clipped_relu(v, z=20.0))
    l.backward()
    print(v.grad)
```

Output

```
[0. 1. 0.]
[0. 1. 0.]
[0. 1. 1.]
```

I think avoiding cuDNN is better, because it allows change `F.clipped_relu` to retain output. Retaining output is consistent with `F.relu` and usually reduces memory consumption. TensorFlow's ReLU6 do this. (https://github.com/tensorflow/tensorflow/blob/96237f7b7ae6b7b8a2cbcf6d64312906b96f060b/tensorflow/core/kernels/relu_op_functor.h#L84-L87)

Note that for `F.clip` I think it is better to propagate gradients at the borders (#5876). The difference in my proposals reflects the difference of expected use cases.

grafi-tt:
Now ChainerX implementation is added (#6627). Apparently it would return `[1., 1., 1.]` on the test snippet above.

grafi-tt:
bump

PR #5898, opened 2018-12-21
Fix broken `F.leaky_relu` grad when slope = 0

grafi-tt:
When slope = 0, `F.leaky_relu` backpropagated all gradients, even at x < 0. This PR fixes this.

The following script shows the problem.

```python
import chainer
import chainer.functions as F
import numpy as np
x = np.arange(-5, 5, dtype=np.float32)
v = chainer.Variable(x)
l = F.sum(F.leaky_relu(v, slope=0.0))
l.backward()
print(v.grad)
```

```output
[1, 1, 1, 1, 1, 1, 1, 1, 1, 1]
```

Note that this PR is not about the behavior at `x = 0`, contrary to #5876 #5877.

### DIFF
--- a/chainer/functions/activation/leaky_relu.py
+++ b/chainer/functions/activation/leaky_relu.py
@@ -12,7 +12,7 @@ def _get_kern():
     if _kern is None:
         _kern = cuda.elementwise(
             'T cond, T x, T slope', 'T y',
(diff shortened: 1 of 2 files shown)


okuta:
jenkins, test this please.

okuta:
LGTM!

PR #5911, opened 2018-12-22
Update third-party.cmake to follow the recent way

shinh:
suggested by gtest:

https://github.com/google/googletest/commit/c45631823c983935d228557b124eff05263b2660

This change makes it possible for other projects to use
ChainerX as a cmake sub-directory.


niboshi:
LGTM

PR #5914, opened 2018-12-22
Add copyparams method to Sequential

### LINKED ISSUE #5824: `Link.copyparams` does not work with `Sequential` with non-links
* Conditions
```
Platform: Darwin-18.2.0-x86_64-i386-64bit
Chainer: 6.0.0b1
NumPy: 1.15.4
CuPy: Not Available
iDeep: Not Available
```

* Code to reproduce
```python
l1 = chainer.Sequential(chainer.links.Linear(3, 4), chainer.functions.relu, chainer.links.Linear(4, 5))
l2 = chainer.Sequential(chainer.links.Linear(3, 4), chainer.functions.relu, chainer.links.Linear(4, 5))
l1.copyparams(l2)
```

```
  File "/Users/tos/GitHub/chainer/chainer/link.py", line 1227, in copyparams
    child.copyparams(link[idx], copy_persistent)
  File "/Users/tos/GitHub/chainer/chainer/link.py", line 611, in copyparams
    dst[name].copydata(src[name])
KeyError: 'b'
```

mitmul:
This PR fixes #5824

### DIFF
--- a/chainer/sequential.py
+++ b/chainer/sequential.py
@@ -6,7 +6,7 @@ import chainer
 from chainer import function
 
 
-class Sequential(chainer.ChainList):
+class Sequential(chainer.link.ChainList):
 
     """Sequential model which has a single-stream forward pass.
 
@@ -145,7 +145,7 @@ class Sequential(chainer.ChainList):
 
     def __delitem__(self, i):
         layer = self._layers.pop(i)
-        if isinstance(layer, chainer.Link):
+        if isinstance(layer, chainer.link.Link):
             for i, _link in enumerate(self._children):
                 if _link.name == layer.name:
                     del self._children[i]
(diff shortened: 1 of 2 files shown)


chainer/sequential.py
```diff
@@ -381,6 +381,11 @@ def copy(self, mode='share'):
                 ret.append(copy.copy(layer))
         return ret
 
+    def copyparams(self, sequential, copy_persistent=True):
```
niboshi: >`sequential`

How about not changing the parameter name from the base method? I don't think it's a good practice.
By the way, can we assume the source link is of `Sequential`? (There's no such mention of the assumption in the documentation).

mitmul: OK, I’ll fix the name of this argument.

> can we assume the source link is of Sequential?
Ah, I guess it’s better to raise an error when the given source link is not a Sequential object?
And at least I need to add a description about it to the documentation.

mitmul:
@niboshi Hi, I fixed the argument name and added a check for the object type of source link. Could you review this again?

takagi:
Jenkins, test this please.

chainer/sequential.py
niboshi: Thanks.

mitmul:
Jenkins, test this please.

mitmul:
@takagi could you merge this PR? All tests have succeeded.

chainer/sequential.py
```diff
@@ -2,11 +2,11 @@
 import functools
 import inspect
 
+import chainer
 from chainer import function
-from chainer import link
 
 
-class Sequential(link.ChainList):
+class Sequential(chainer.ChainList):
```
toslunar: This change relies on that `chainer.sequential` is imported later than `chainer.link.ChainList` in `chainer/__init__.py`.

mitmul: Thanks, I updated the code by replacing `chainer.*` with `chainer.link.*`. Is it what you meant to fix?

mitmul:
Jenkins, test this please

chainer/sequential.py
niboshi: I don't think it fixes.
`chainer.link` might not be imported.

mitmul:
Eh, jenkins, test this please

chainer/sequential.py
mitmul: @toslunar @niboshi Could you check this again?

takagi:
I now merge this!

PR #5920, opened 2018-12-24
[backport] Remove some CI requirements

okuta:
Backport of #5865

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -56,10 +56,6 @@ requirements = {
         '-r stylecheck',
         '-r test',
         '-r docs',
-        # pytest-timeout>=1.3.0 requires pytest>=3.6.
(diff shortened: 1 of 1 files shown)


okuta:

niboshi:
I think we don't have to backport this fix.

okuta:
OK.

niboshi:
How about restricting to `pytest-cov<=2.6.0` instead?
We should be careful about changing requirements (even removal) in the stable branch.
pytest-cov 2.6.1 was released just a day ago, so it should be relatively safer to change `pytest-cov` to `pytest-cov<=2.6.0`.

...But as they're requirements for `travis` and `appveyor`, it's unlikely that any third-party depend on it.
So it's not a strong opinion.

niboshi:
#5968 has been merged.

PR #5924, opened 2018-12-25
improve `F.group_normalization` via cuDNN call

grafi-tt:
Memory consumption reduced and speed slightly improved.

* cuDNN's batch normalization routine is directly called to simplify computation graph.
* `x_hat = (x - mean) / std` is re-computed at backward phase to reduce memory consumption.
* Some `ElementwiseKernel` and `ReductionKernel` are used to reduce kernel calls.

Note that if we cache `x_hat`, speed can be further improved. But I think memory consumption is more crucial, because Group Normalization is particularly useful when batch size can't be increased due to memory limit.

Regarding speed, time to train ResNet50 with 224 x 224 input images becomes 94.5%. Validation was done at each epoch, and train/valid split ratio was 8/2. Batch size was 32. The benchmark environment is as follows.

* CPU: Ryzen7 2700X
* GPU: ZOTAC GTX 1080 Ti Amp Edition
* DRAM: DDR4-3200 8GiB x 2 dual channel
* SSD: WD Black NVMe 1TB
* OS: Linux 64bit
* CUDA version: 10.0.130
* cuDNN version: 7.4.1.5

Add to that, as Group Normalization is a generalization of Layer Normalization and Instance Normalization, current `F.layer_normalization` implementation can be replaced in future.

### DIFF
--- a/chainer/functions/normalization/group_normalization.py
+++ b/chainer/functions/normalization/group_normalization.py
@@ -282,8 +282,16 @@ class _XHatGrad(function_node.FunctionNode):
         if 0 in indexes:
             # gx = inv_std * gx_std
             # dgx = dinv_std * gx_std + inv_std * dgx_std
+            #
             # -gx2l = (ggx * dinv_std * gx_std) / dx
-            # -gx_hat2r = (inv_std * ggx * dgx_std) / dx_hat
+            #       = mean(ggx * gx_std) * (dinv_std / dx)
+            #       = -mean(ggx * gx_std) * inv_std^2 * x_hat
+            #       = -inv_std * x_hat * mean(ggx * gx)
+            #
+            # By `gx_std = gx_hat - gx_hat_avg - x_hat * gx_hat_x_hat_avg`,
+            # -gx_hat2r = (ggx * inv_std * dgx_std) / dx_hat
+            #           = -inv_std * (ggx * mean(gx_hat * x_hat) +
+            #                         gx_hat * mean(ggx * x_hat))
 
             gx2l_std = x_hat * F.mean(ggx * gx, axis=1, keepdims=True)
             gx2l, = _MulInvStd(
--- a/chainer/functions/normalization/group_normalization.py
+++ b/chainer/functions/normalization/group_normalization.py
@@ -35,12 +35,16 @@ class GroupNormalization(function_node.FunctionNode):
             beta_type.ndim == 1,
             gamma_type.dtype == x_type.dtype,
             beta_type.dtype == x_type.dtype,
-            x_type.shape[1] % self.groups == 0,
             x_type.shape[1] == gamma_type.shape[0],
             gamma_type.shape == beta_type.shape,
         )
 
     def forward(self, inputs):
+        if inputs[0].shape[1] % self.groups != 0:
+            raise ValueError('The number of channels {} is not divisible by '
+                             '\'groups\' argument {}.'
+                             .format(inputs[0].shape[1], self.groups))
+
         xp = backend.get_array_module(*inputs)
         if xp is not numpy and chainer.should_use_cudnn('>=auto', 5000):
             return self.forward_cudnn(inputs)
--- a/chainer/functions/normalization/group_normalization.py
+++ b/chainer/functions/normalization/group_normalization.py
@@ -116,7 +116,7 @@ class GroupNormalization(function_node.FunctionNode):
         batch_size = orig_shape[0]
         groups = self.groups
         reduced_shape = (batch_size * groups, -1)
-        x = chainer.functions.reshape(x, reduced_shape)
+        x = x.reshape(reduced_shape)
 
         x_hat, = _XHat(
             self.eps, self.mean, self.inv_std,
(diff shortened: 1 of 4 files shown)


hvy:
Thanks for this PR and sorry for the delayed response. Could you post the memory usage improvement here as well, in addition to the speed improvements?

chainer/functions/normalization/group_normalization.py
```diff
@@ -1,10 +1,348 @@
...
+        gx = gx.reshape(orig_shape)
+        return gx, ggamma, gbeta
+
+
+class _GradHelper(function_node.FunctionNode):
+
+    def forward(self, inputs):
+        self.retain_inputs((0, 1, 2))
+        x_hat, gamma, gy = inputs
+        xp = backend.get_array_module(x_hat)
```
hvy: A detail but let's declare `xp` closer to where it's used. Also, it could be removed since it's only used once.

chainer/functions/normalization/group_normalization.py
```diff
@@ -1,10 +1,348 @@
...
+        gy = gy.reshape((batch_size, channels, -1))
+        ggx_hat = ggx_hat.reshape((batch_size, channels, -1))
+
+        gx_hat2 = gggamma[:, None] * gy
+        ggamma2 = chainer.functions.sum(ggx_hat * gy, axis=(0, 2))
+        ggy = (ggx_hat * gamma[:, None] + gggamma[:, None] * x_hat +
+               ggbeta[:, None])
+
+        gx_hat2 = chainer.functions.reshape(gx_hat2, reduced_shape)
+        ggy = chainer.functions.reshape(ggy, orig_shape)
```
hvy: Just curious but how come you're using the free function `reshape` instead of the method here (you're mostly using the method in other places)?

chainer/functions/normalization/group_normalization.py
```diff
@@ -1,10 +1,348 @@
...
+
+        self.groups = groups
+        self.eps = eps
+
+    def check_type_forward(self, in_types):
+        type_check.expect(in_types.size() == 3)
+        x_type, gamma_type, beta_type = in_types
+        type_check.expect(
+            x_type.dtype.kind == 'f',
+            x_type.ndim >= 2,
```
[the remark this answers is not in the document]
toslunar: The existing check is too strict.  My suggestion: grafi-tt#2

grafi-tt:
@toslunar Thank you for your PR!

@hvy Thank you for your review, and sorry for late reply. I reflected your suggestions. Could you make review again?

chainer/functions/normalization/group_normalization.py
```diff
@@ -1,10 +1,360 @@
...
+
+        batch_size, channels = gy.shape[:2]
+        gy = gy.reshape((batch_size, channels, -1))
+
+        reduced_shape = x_hat.shape
+        x_hat = x_hat.reshape((batch_size, channels, -1))
+
+        gx_hat = gy * gamma[:, None]
+        gbeta = gy.sum(axis=(0, 2))
+        if backend.get_array_module(x_hat) is numpy:
```
toslunar: ditto

chainer/functions/normalization/group_normalization.py
```diff
@@ -1,10 +1,360 @@
...
+        gx_hat, ggamma, gbeta = _GradHelper().apply((x_hat, gamma, gy))
+        gx, = _XHatGrad(
+            self.eps, self.mean, self.inv_std,
+            self.dummy_gamma, x_hat.array).apply((x, gx_hat))
+
+        gx = gx.reshape(orig_shape)
+        return gx, ggamma, gbeta
+
+
+class _GradHelper(function_node.FunctionNode):
```
toslunar: I'd mention the function node computes grad of `y = gamma * x_hat + beta`.  Maybe it's sufficient to rename to a more expressive name like `_ScaleShiftGrad`.

chainer/functions/normalization/group_normalization.py
```diff
@@ -1,10 +1,360 @@
...
+        self.dummy_gamma = dummy_gamma
+
+    def forward(self, inputs):
+        self.retain_inputs((0,))
+        _, y = inputs
+        z = self.inv_std[:, None] * y
+        self.retain_outputs((0,))
+        return z,
+
+    def backward(self, indexes, grad_outputs):
```
toslunar: It seems the method is not tested because it's used for >=3rd order grad.

hvy:
Thanks for the fixes, they look good to me. Could you also take a look at the remaining ones? Also, could you post some benchmarks on the memory consumption improvements (as requested in a comment above)?

chainer/functions/normalization/group_normalization.py
```diff
@@ -1,10 +1,360 @@
...
+        ggx, = grad_outputs
+
+        x_hat, = _XHat(
+            self.eps, self.mean, self.inv_std,
+            self.dummy_gamma).apply((x,))
+
+        ret = []
+
+        if 0 in indexes:
+            # gx = inv_std * gx_std
```
crcrpar: I think these commented out lines can be removed.

grafi-tt: Those lines are *not* commented out source codes. They are here to sketch the derivation of double backprop formula.

I'd like to make the comments less confusing, but I don't have an idea...

crcrpar: apology.

How about adding a single comment that clarifies below lines are not comment, but the sketch of double backprop formula?
For me, just adding some comments at the beginning and end of the sketch is enough.

grafi-tt:
@toslunar @hvy @crcrpar Thank you very much. I fixed the problems commented.

* rebase on v6.0.0b3
* compare `xp` to `cuda.cupy`
* rename `_GradHelper` to `_ScaleBiadGrad`
* fix `_MulInvStd` backward computation (sign flipped)
* make comments about `gx2` less confusing
* simplify tests using `FunctionTestCase`
* add test for `_MulInvStd`
* test with large `eps` value

I checked runtime memory consumption of ResNet50 by `nvidia-smi`. It reduced from 7473MiB to 3977MiB. Though this is a rough measure, I think it is enough to tell the improvement.

toslunar:
Great!

toslunar:
Jenkins, test this please.

toslunar:
cupy/cupy#2060 broke this PR.  Let me wait for cupy/cupy#2072.

takagi:
@grafi-tt Would you merge the current master to your branch? The master is based on the change of cupy/cupy#2060, so your PR should pass the CI after merging it.

grafi-tt:
@takagi Thanks. I merged the master branch #a8f9cb5. But If I understand correctly it doesn't fix the issue, because the GroupNormalization implementation directly calls `cudnn.batch_normalization_forward_training`. Should I modify the code according to #2060?

toslunar:
As #6323 fixed a fp16 issue, the PR needs the same fix (to stop using `F.batch_normalization`).

takagi:
Ah, I read `GroupNormalization ` of the master, where `cudnn.batch_normalization_forward_training` is not called directly. Yes, it would be great if you modified your code as well.

grafi-tt:
@takagi Thank you. I added the modification like #6323.

toslunar:
Jenkins, test this please.

hvy:
Sorry for the late response, and thanks a lot for working on this long running PR. LGTM!

PR #5927, opened 2018-12-25
Enforce `Link.__init__` in subclasses

niboshi:
When implementing a link, it's very easy to forget `super().__init__()`, and that results in difficult to understand errors (the error could be delayed, to make things worse).

```py
import chainer
import numpy


class A(chainer.Link):
    def __init__(self):
        #super().__init__()
        with self.init_scope():
            self.a = chainer.Parameter(1, (2, 3))

    def forward(self, x):
        return x * self.a


l = A()
x = numpy.ones((2, 3), numpy.float32)
y = l(x)
```
```
AttributeError: 'A' object has no attribute '_device'
```

This PR turns the errors into more easy to understand one, and raise it earlier. 

```
RuntimeError: Link.__init__() has not been called.
```

### DIFF
--- a/tests/chainer_tests/test_link.py
+++ b/tests/chainer_tests/test_link.py
@@ -4,6 +4,7 @@ import warnings
 
 import mock
 import numpy
+import pytest
 
 import chainer
 from chainer import backend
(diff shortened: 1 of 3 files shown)


niboshi:
Now ready for review.

beam2d:
The design and code look good. Could you add a small test?

niboshi:
Wrote some tests.

beam2d:

tests/chainer_tests/test_link.py
```diff
@@ -649,6 +650,51 @@ def test_to_chainerx(self, backend_config):
...
+        # The fault should be detected no later than __call__().
+
+        class Derived(chainer.Link):
+            def __init__(self):
+                pass
+
+            def forward(self, x):
+                return x
+
+        with pytest.raises(RuntimeError, message=self.expected_message):
```
beam2d: It looks `match` is the correct option to use instead of `message` (the latter just prints the message when the test fails). CI failed due to the warning.

tests/chainer_tests/test_link.py
```diff
@@ -649,6 +650,51 @@ def test_to_chainerx(self, backend_config):
...
+    def test_missing2(self):
+        # init_scope is called.
+        # The fault should be detected at init_scope.
+
+        class Derived(chainer.Link):
+            def __init__(self):
+                with self.init_scope():
+                    pass
+
+        with pytest.raises(RuntimeError, message=self.expected_message):
```
beam2d: ditto

tests/chainer_tests/test_link.py
```diff
@@ -649,6 +650,51 @@ def test_to_chainerx(self, backend_config):
...
+
+    def test_missing3(self):
+        # add_param is called.
+        # The fault should be detected at add_param.
+
+        class Derived(chainer.Link):
+            def __init__(self):
+                self.add_param('p1', (2, 3), numpy.float32)
+
+        with pytest.raises(RuntimeError, message=self.expected_message):
```
beam2d: ditto

beam2d:
Jenkins, test this please.

beam2d:
Jenkins, test this please.

beam2d:
Jenkins, test this please.

beam2d:
LGTM

PR #5936, opened 2018-12-27
Fix visibility for pybind exception registration for macOS

hvy:
Fix that's required to correctly bind ChainerX custom exceptions on macOS (libc++).

More precisely, custom exceptions raised by the C++ code were treated as different types from the ones registered here https://github.com/chainer/chainer/blob/master/chainerx_cc/chainerx/python/error.cc due to the default visibility configuration of pybind11 which is `hidden` and set here https://github.com/pybind/pybind11/blob/master/include/pybind11/detail/common.h#L24. 

This PR changes the visibility to `default` and also includes changes to utilize `pybind11_add_module` which also seems to be required.

#### Reference
"Problems with C++ exceptions (please read!)" in https://gcc.gnu.org/wiki/Visibility

### DIFF
--- a/chainerx_cc/chainerx/macro.h
+++ b/chainerx_cc/chainerx/macro.h
@@ -36,3 +36,11 @@
     } while (false)
 #endif  // NDEBUG
 #endif  // CHAINERX_NEVER_REACH
+
+#ifndef CHAINERX_VISIBILITY_HIDDEN
+#if defined(WIN32) || defined(_WIN32)
+#define CHAINERX_VISIBILITY_HIDDEN
+#else
+#define CHAINERX_VISIBILITY_HIDDEN __attribute__((visibility("hidden")))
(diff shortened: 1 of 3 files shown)


hvy:
jenkins, test this please.

hvy:
For the record, please refer to the exceptions section here https://gcc.gnu.org/wiki/Visibility.

hvy:
jenkins failure seems unrelated.

hvy:
Specify rpath so that ChainerX is correctly imported (`_core.so` finds `libchainerx.dylib`) when installed in a non-editable mode (e.g. via sdist).

hvy:
jenkins, test this please.

hvy:
jenkins, test this please.

hvy:
jenkins, test this please.

hvy:
jenkins, test this please.

asi1024:
LGTM.

PR #5938, opened 2018-12-27
Transform dataset documentation fix

23pointsNorth:
Fix errors in grammar and specify the correct range after the transformation.


toslunar:
Jenkins, test this please.

PR #5949, opened 2018-12-30
Do not ignore FutureWarning other than experimental features

toslunar:
This PR changes only tests, while defining `class ExperimentalWarning(FutureWarning): pass` might resolve some other issues.

PR #5950, opened 2018-12-30
Suppress FutureWarning from NumPy that Chainer fixes

toslunar:
Outputs of `F.split_axis` don't depend on NumPy version.

### DIFF
--- a/chainer/functions/array/split_axis.py
+++ b/chainer/functions/array/split_axis.py
@@ -1,3 +1,5 @@
+import warnings
+
 import numpy
 import six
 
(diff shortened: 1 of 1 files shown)


toslunar:
#5949 tests this.

chainer/functions/array/split_axis.py
```diff
@@ -13,8 +15,14 @@
 _numpy_split_ok = numpy.lib.NumpyVersion(numpy.__version__) >= '1.11.0'
 
 
-def _fix_numpy_split(ys, x, indices_or_sections, axis):
+def _numpy_1_11_split(x, indices_or_sections, axis):
     """Make the output of np.split compatible with numpy >= 1.11"""
+    with warnings.catch_warnings():
```
kmaehashi: Use of `catch_warnings` should be avoided because it is thread unsafe.

As this affects users using NumPy 1.10.x or earlier, I feel it's ok to leave as it is.

toslunar:
OK

PR #5952, opened 2018-12-31
Add interface to backprop from multiple variables

toslunar:
https://github.com/chainer/chainer/issues/5709#issuecomment-441972123


kmaehashi:
We discussed this with @beam2d in the daily meeting.
The interface sounds OK as it has the same signature as `chainerx.backward`.
https://docs.chainer.org/en/latest/chainerx/reference/generated/chainerx.backward.html

Could you proceed?

toslunar:
Should we give default `y.grad` for `y` in `outputs`?

kmaehashi:
I think it's not needed (unless it is

kmaehashi:
I think it's not needed considering the current design of `Variable.backward()`.
c.f. https://stackoverflow.com/questions/50969009/why-need-we-assign-the-grad-of-y-in-chainer/50979188#50979188

kmaehashi:
Blocked by #6348

chainer/_backprop.py
```diff
@@ -0,0 +1,262 @@
...
+
+    if grad_outputs is None:
+        grad_outputs = []
+        for y in outputs:
+            grad_var = y.grad_var
+            if grad_var is None:
+                warnings.warn(
+                    'outputs contains a Variable without grad, or '
+                    'duplicate outputs. Note that'
+                    'chainer.backward does not have default grad.',
```
kmaehashi:
```suggestion
                    'chainer.backward does not set default grad.',
```

kmaehashi:
Jenkins, test this please.

kmaehashi:
Jenkins, test this please.

toslunar:
Jenkins, test this please.

toslunar:
Note:
The PR includes moves of files.  Most of the changes are available at
https://github.com/chainer/chainer/pull/5952/files/d24dd57005cb575635538d8b477352266031252f.

PR #5954, opened 2019-01-07
WIP: Bypass `_concat_arrays` to improve performance

okapies:
*This PR is a work in progress.*

This patch bypasses `_concat_arrays` when loading examples into a device. It increases the performance of the MNIST example by 7~10% compared with the current implementation using chainerx on GPU (`--batch-size=100 --device=cuda:0`) in my environment. I believe it is especially useful for experiments with large mini-batch.

- Add `BatchableDatasetMixin`, which has `get_examples` in addition to the `DatasetMixin` operators
    - The existing `get_example` produces an example, or a list of {tuples, dicts} of examples (e.g. `List[Tuple[ndarray, int]]`). In this setting, you need to convert them (e.g. `concat_examples`) to feed them into a device
    - `get_examples` produces `Examples` instance unlike `get_example`
- Add `Examples`, which is a `Sequence`-like collection of examples
    - It keeps the examples as a dataset, or a {tuple, dict} of datasets (e.g. `Tuple[List[ndarray], List[int]]`)
    - Its `to_dataset` method transfers the examples into the specified `device` in batch without any conversions
    - If the dataset is `ndarray`, `to_dataset` tries to slice it by using array indexing (`dataset[indices]`) instead of picking each example
    - (It has a same interface with `collections.abc.Sequence` but is not an actual instance of it to avoid the overhead of `isinstance`)
- Add `sample_from_dataset` in `chainer.dataset.examples`, which is a convenient function to implement `get_examples`
    - e.g.:
```python
class MyDataset(BatchableDatasetMixin):
    ...
    def get_examples(self, indices):
        order = ...
        padding_spec = ...
        return examples.sample_from_dataset((self._images, self._labels), indices, order, padding_spec)
```
- Fix `Iterator`s to handle datasets that have a `get_examples` method
    - `next()` of the iterator will return an `Examples` for a "batchable" dataset instead of an actual list of examples
- Fix `Updater`s and `Evaluator`s to skip running `convert` (e.g. `concat_examples`) if `Iterator.next()` returns `Examples`
    - This patch keeps backward compatibility with the customized `Updaters` in the wild, thanks to `__iter__` method in `Examples`
- Add `resolve_device_spec` function to reuse a part of `to_device` code in `Examples`

## TODO
- Write tests
- Implement `get_examples` in `TransformDataset`
- Fix more iterators in Chiainer Core and examples to support `Examples`
- Fix iterators in ChainerMN
    - Some codes assume `Iterator.next()` returns `list` directly
- Fix imagenet example (which uses DALI for pre-processing)
    - Maybe need to extend `sample_from_dataset` or `to_dataset` to accept a custom convert function
- Confirm the specification
    - It seems some functions and links assume its inputs are `ndarray` instead of dataset objects (which just have `__getitem__` and `__len__`). I can't judge `Examples` need to convert their types on behalf of `concat_examples` or not
    - padding in parallel updater


okapies:
@hvy I believe I've done the PoC of `concat_examples` free iterator. Could you check the concept and design of this patch before proceeding with further fixes? Please feel free to ask me if you have any questions.

okapies:
It seems that preprocessing and transmission to the device should be done in the iterator because `Updater` can be a bottleneck.

okapies:
It will be superseded by #7117. Close this after it is merged.

PR #5972, opened 2019-01-09
improve `F.elu` memory consumption by retaining output

grafi-tt:
When `alpha >= 0.0`, `F.elu` can modified to retain output. Gradient computation becomes easier.

It might degrade the precision of gradients, but it won't be worse than `F.sigmoid`, `F.tanh`, `F.swish` or `F.softplus`. [TF also computes gradients from the output.](https://github.com/tensorflow/tensorflow/blob/96237f7b7ae6b7b8a2cbcf6d64312906b96f060b/tensorflow/core/kernels/relu_op_functor.h#L149)

### DIFF
--- a/chainer/functions/activation/elu.py
+++ b/chainer/functions/activation/elu.py
@@ -2,6 +2,7 @@ import numpy
 
 from chainer.backends import cuda
 from chainer import function_node
+from chainer import utils
 from chainer.utils import type_check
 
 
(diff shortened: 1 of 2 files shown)


hvy:
@toslunar can we assign you on this ticket?

grafi-tt:
@toslunar Thanks. I modified test code to parameterize alpha with a positive random value, zero and a negative random value.

toslunar:
LGTM.

Jenkins, test this please.

toslunar:

Issue #4134, opened 2017-12-21
When and how one can set a random seed for Chainer


muupan:
I cannot find any guarantee on how to set a random seed to make Chainer deterministic. I want Chainer to clarify when and how it is deterministic.

`numpy.random.seed` and `cupy.random.seed` may work, but there seems no guarantee that Chainer has no other random sources, and I can find some exceptions: cudnn, atomicAdd (https://groups.google.com/forum/#!topic/chainer-jp/JN9H9aF4TA8), https://github.com/chainer/chainer/blob/master/chainer/datasets/sub_dataset.py, https://github.com/chainer/chainer/blob/master/chainer/functions/connection/n_step_rnn.py.

kmaehashi:
We discussed offline:
* We need to document how to make Chainer computation deterministic.
* We need system to show warnings for non-deterministic computations, e.g.,
    * introduce decorator (like `@chainer.non_deterministic`) to annotate known non-deterministic functions (e.g., using `atomicAdd`)
    * introduce configuration option (like `chainer.config.deterministic`)
    * if the configuration is True and the function with the annotation is invoked, raise warning.

Crissman:
@muupan I've made a PR to handle the atomicAdds. I took a look at the two blobs you linked above, and both of them seem not to have atomicAdds, but rather `random` functions. Those functions allow setting the `seed` and already seem fairly well labelled. Do you want to add the `@non_deterministic` decorator to those random functions as well?

Crissman:
After discussion, we've decided that the PRs done covers the issue of flagging code which is non-deterministic, and cannot be given a seed. Closing this issue.

PR #5977, opened 2019-01-10
Add non-deterministic warning

Crissman:
To address issue #4134.

Created @non_deterministic decorator.
Added decorator to functions that use atomicAdd


niboshi:
I think it's better to implement it with a function rather than a decorator, because in general determinism depends on runtime information. I think there's no advantage of a decorator over a function.

By the way, shouldn't we also annotate nondeterminism in numpy?

Crissman:
Actually, I started replacing the decorator with a function, and I don't think it's better. Grammatically, it's cleaner to have a `@non_deterministic` decorator above the relevant functions. If instead of the decorator, I added a function call within the function, what would I call it? `check_deterministic()`? If so, that feels confusing with `config.check_deterministic`. I think it's clearer to use the decorator, since it makes for a cleaner name space.

niboshi:
First of all, I don't think `config.check_deterministic` is good. Its behavior is ambiguous. After checking determinism, what does it do? Keep in a log? Raise warning? Does it warn if it's deterministic? Or opposite?

I would prefer something like `config.warn_nondeterministic`.

As for the function name, how about simply `util.nondeterministic()`?
The role of this function would be "to declare the execution path is nondeterministic".

Crissman:
@niboshi I'm convinced. Converted to a function instead of a decorator.

Crissman:
Travis error on Mac appears unrelated.

niboshi:
We won't be able to assure 100% detection rate. We should mention that this feature is provided as best-effort and users should not rely on it.
I think we should also mention the possibilities of false-positives: for example, reduction on a single element is deterministic but I don't think we will eliminate every such conditions.

chainer/functions/connection/embed_id.py
```diff
@@ -6,6 +6,7 @@
 from chainer.backends import cuda
 from chainer import function_node
 from chainer.utils import type_check
+from chainer.utils import non_deterministic
```
niboshi: "Do not import objects, only modules".
https://docs.openstack.org/hacking/latest/user/hacking.html#imports

chainer/utils/non_deterministic.py
```diff
@@ -0,0 +1,15 @@
+import warnings
+from chainer import configuration
+
+
+def non_deterministic(f_name):
```
niboshi: How about `nondeterministic`?

Crissman:
@niboshi Please review revisions.

chainer/functions/connection/embed_id.py
```diff
@@ -59,6 +60,7 @@ def __init__(self, w_shape, ignore_label=None):
         self.ignore_label = ignore_label
 
     def forward(self, inputs):
+        utils.nondeterministic('atomicAdd')
```
niboshi: It seems this should be placed in the cupy code path.

chainer/functions/loss/ctc.py
```diff
@@ -154,6 +154,7 @@ def log_matrix(self, x, xp):
     # path probability to label probability
     def label_probability(self, label_size, path, path_length,
                           multiply_seq, xp):
+        utils.nondeterministic('atomicAdd')
```
niboshi: ditto.

docs/source/reference/configuration.rst
```diff
@@ -66,6 +66,11 @@ Configuration Keys
 
    You can change the default value to ``True`` by setting ``CHAINER_KEEP_GRAPH_ON_REPORT`` environment variable to ``1``.
 
+* ``warn_nondeterministic`` (default: ``False``)
+   Flag to give warning when a non-deterministic function is used. This function is experimental.
+
+   If it is true, then functions that use non-deterministic functions, such as atomicAdd, will give a warning when executed. Note the determinisitic outputs may still result, despite using a nondeterministic function, depending on usage.
```
niboshi: We should mention false negatives first, which are more important.
Let's discuss the expression offline later.

chainer/utils/nondeterministic.py
```diff
@@ -0,0 +1,15 @@
+import warnings
+from chainer import configuration
```
niboshi: Please add a blank line before importing chainer module.

niboshi:
Jenkins, test this please

Crissman:
Jenkins error looks unrelated. Retest?

niboshi:
Discussed:
We may raise warning for features that use random numbers by the following additions.
* Introduce an interface to set random seeds
  * Make it possible to enumerate backends (numpy, cupy, chainerx) and set random seeds in a unified API.
* Raise warning in features that use random numbers only if the seed is not set via the above interface.

However we don't do that for now, and make it users' responsibility to set seeds.

@Crissman Let's document that it won't raise warning for such features.

Crissman:
Documentation added that setting seed is responsibility of user.

niboshi:
Jenkins, test this please

Crissman:
Jenkins issue looks unrelated. Please merge.

niboshi:
@Crissman 
It seems you did the merge wrong. Please fix it.

Crissman:
Ok. I think it's clean now. Please take a look.

niboshi:
Jenkins, test this please

niboshi:
LGTM

PR #5986, opened 2019-01-13
Refactor `Link.to_device` family

niboshi:
Separates the structures and the algorithms of `{Link,Chain,ChainList,...}` × `{to_device,to_cpu,to_gpu,to_intel64,to_chainerx,from_chainerx}`.


niboshi:
Awaiting comments on the design (including class/method names) and then I will work on documentation and tests.

chainer/_link.py
```diff
@@ -0,0 +1,271 @@
...
+
+import chainer
+from chainer import backend
+from chainer.backends import cuda
+from chainer.backends import intel64
+from chainer import types  # NOQA
+import chainerx
+
+
+class DeviceObjectsVisitableMixin(object):
```
beam2d: I want some new concept and a name for it that indicates "a class/object that supports the canonical device identification/transfer APIs" if you expect that this feature will be public and used by non-link classes.

chainer/_link.py
```diff
@@ -0,0 +1,271 @@
...
+    def visit_param(self, param):
+        """Processes a parameter"""
+        raise NotImplementedError()
+
+    def visit_link(self, link):
+        """Processes a link"""
+        raise NotImplementedError()
+
+
+class _GetSoleDeviceVisitor(DeviceObjectsVisitor):
```
beam2d: Is this visitor needed to be dedicated to getting the "sole" device? Can we generalize it to a visitor that collects all devices used in the object and let the user of this visitor determine if it is "sole" from the length of the result?

chainer/_link.py
niboshi: Yes, I want to make it more easily understood concept than "device objects visitable", but I don't come up with a better name/generalization.

chainer/_link.py
niboshi: Yes, we can, but I think there's no need for generalization at the moment because it's a private class.

beam2d: Hmm, ok. I just thought the generalized one is easier to understand for me.

chainer/_link.py
niboshi: Some random ideas
* `DeviceObject`
* `DeviceResident`
* `DeviceOccupant`

niboshi:
Done:
* Renamed to `DeviceResidentMixin` (from `DeviceObjectVisitableMixin`).
* Overriding `to_device()` causes an error.
* Overriding `to_cpu/to_gpu/to_intel64` causes a warning, but they're still called (even if `to_device` was originally called).

Remaining:

* Where to put `DeviceResidentMixin` and visitor implementations? `_link.py` is rather temporary.
* I'm going to remove `_GetSoleDeviceVisitor` from this PR. It's implemented only for demonstration.

chainer/_link.py
```diff
@@ -0,0 +1,373 @@
...
+from chainer.backends import _cpu
+from chainer.backends import cuda
+from chainer.backends import intel64
+from chainer import types  # NOQA
+import chainerx
+
+
+class DeviceResidentMixin(object):
+    """A mix-in class to support multi-device hierarchy.
+    """
```
beam2d: Put a blank line before the docstring. I think it's also ok to put the closing quotes at the same line (not mandatory, though):
```py3
class DeviceResidentMixin(object):

    """A mix-in class to support multi-device hierarchy."""
```

chainer/_link.py
```diff
@@ -0,0 +1,373 @@
...
+
+    _device = _cpu.CpuDevice()
+
+    # Indicates the object is mixed-device.
+    _MIXED_DEVICE = object()
+
+    def __init__(self):
```
beam2d: Can we check the overrides at the inheritance time instead of instantiation time? Maybe it needs metaclass.

chainer/_link.py
```diff
@@ -0,0 +1,373 @@
...
+        visitor = _ToDeviceVisitor(
+            backend.CpuDevice(),
+            entry_method_info=('to_cpu', {}),
+            skip_between_cupy_devices=True)
+        self.__to_device(visitor)
+        return self
+
+    def to_gpu(
+            self,
+            device=None  # type: tp.Optional[types.CudaDeviceSpec]
```
beam2d: It is a matter of taste, but I prefer putting a trailing comma here (to keep the diff simple on adding a new argument).

chainer/_link.py
```diff
@@ -0,0 +1,373 @@
...
+    if to_method is not to_method_orig:
+        return True  # overridden
+    return False
+
+
+class DeviceResidentsVisitor(object):
+    # Base class of visitors.
+
+    def visit_visitable(self, visitable, visitor):
+        assert isinstance(visitable, DeviceResidentMixin)
```
beam2d: If the type hierarchy means something, I don't think it is good to call the superclass "mixin". How about just naming it `DeviceResident`?

chainer/link.py
```diff
@@ -47,7 +46,7 @@ def _ensure_shape_dtype(value):
         return value  # type: ignore
 
 
-class Link(object):
+class Link(_link.DeviceResidentMixin, object):
```
beam2d: No need to list `object` here.

chainer/utils/walker_alias.py
```diff
@@ -1,11 +1,11 @@
 import numpy
 
 import chainer
-from chainer import backend
 from chainer.backends import cuda
+from chainer import _link
 
 
-class WalkerAlias(object):
+class WalkerAlias(_link.DeviceResidentMixin, object):
```
beam2d: ditto

chainer/_link.py
beam2d: I just wrote a simple implementation for such a check. It should be simple to add warning version too. It may be good to put this kind of things to utils. https://gist.github.com/beam2d/3d1cd3baca30793ff804e7075051d6da

chainer/_link.py
niboshi: Fixed as suggested.

chainer/_link.py
niboshi: I agree. Fixed.

chainer/link.py
niboshi: Fixed.

chainer/utils/walker_alias.py
niboshi: Fixed.

chainer/_link.py
niboshi: Fixed.

chainer/_link.py
niboshi: Hmm, it seems it causes metaclass conflict at the declaration of `Chain`:
```
chainer/link.py:998: in <module>
    class ChainList(Link, collections_abc.MutableSequence):
E   TypeError: metaclass conflict: the metaclass of a derived class must be a (non-strict) subclass of the metaclasses of all its bases
```

beam2d: It looks `collections_abc.MutableSequence` is a metaclass of type `abc.ABCMeta`, so making `enable_final` a subclass of `ABCMeta` resolves the conflict.

niboshi: Fixed as suggested.
Interface has been slightly changed from the original snippet, so please take a look.

niboshi:
Hmm, there's an import cycle:
`utils/__init__.py` -> `utils/walker_alias.py` -> `_link.py` -> `utils/__init__.py`

niboshi:
Fixed.

chainer/_link.py
```diff
@@ -0,0 +1,370 @@
...
+        # type: () -> 'DeviceResident'
+        """Copies parameter variables and persistent values to CPU."""
+        intel64.check_ideep_available()
+        visitor = _ToDeviceVisitor(
+            chainer.get_device(intel64),
+            entry_method_info=('to_intel64', {}))
+        self.__to_device(visitor)
+        return self
+
+    def to_chx(self):
```
beam2d: Isn't it needed to be final? (ditto for from_chx)

chainer/utils/meta.py
```diff
@@ -0,0 +1,79 @@
...
+    base class returned by :meth:`~chainer.utils.enable_final`.
+
+    Args:
+        action: Specifies what happens when the decorated method is being
+            overridden. If `'error'`, :class:`TypeError` is raised. If
+            :class:`DeprecationWarning`, a deprecation warning is raised.
+    """
+    def wrap(f, action='error'):
+        assert callable(f)
+        f.__is_final = (action,)
```
beam2d: How about naming it `f.__override_action`?

chainer/utils/meta.py
```diff
@@ -0,0 +1,79 @@
...
+    return w
+
+
+class _EnableFinal(type):
+    def __new__(cls, name, bases, d):
+        for k in d:
+            for base in bases:
+                f = getattr(base, k, None)  # base method
+                if hasattr(f, '__is_final'):
+                    action, = getattr(f, '__is_final')
```
beam2d: Just write `f.__is_final`.

chainer/utils/meta.py
```diff
@@ -0,0 +1,79 @@
...
+    By default, :class:`TypeError` is raised when the decorated method is being
+    overridden.
+
+    The class in which the decorated method is defined must inherit from a
+    base class returned by :meth:`~chainer.utils.enable_final`.
+
+    Args:
+        action: Specifies what happens when the decorated method is being
+            overridden. If `'error'`, :class:`TypeError` is raised. If
+            :class:`DeprecationWarning`, a deprecation warning is raised.
```
beam2d: How about align the type of both values, i.e., either using `TypeError` for the error case or using `'warning'` for the warning case?

chainer/utils/meta.py
niboshi: It's not possible. It would be regarded as `f._EnableFinal__is_final`.

chainer/utils/meta.py
niboshi: Fixed by using types.

chainer/utils/meta.py
niboshi: Fixed.

chainer/_link.py
niboshi: Fixed.

niboshi:
>As for the name of _link.py, how about just naming it _device_resident.py?

Sounds good. Renamed to `device_resident.py` (without underscore).

niboshi:
Removed `_GetSoleDeviceVisitor`.

Now ready for review.

niboshi:
Renamed `DeviceResidentsVisitor.visit_visitable` to `visit_device_resident`.

chainer/utils/meta.py
beam2d: Oh, ok.

beam2d:
I found that it would be dangerous to mark the conventional `to_cpu/gpu` overriding as deprecated at the same time of introducing a new way (`to_device`). In my understanding, `to_device` interface is not mature yet and the desired API would be changed a bit in the future, so it's safer to wait for it maturing and then mark the old ways deprecated.

niboshi:
Removed deprecation.

beam2d:
Thanks, and sorry for the late response. LGTM.  Could you resolve the conflicts?

niboshi:
Resolved conflict.

beam2d:
Jenkins, test this please.

beam2d:
Jenkins, test this please.

beam2d:
Could you resolve flake8 errors?

niboshi:
Oops, fixed.

beam2d:
Jenkins, test this please.

beam2d:
LGTM!

PR #5987, opened 2019-01-13
Move mypy to static checks

niboshi:
As mypy is a static checker, I moved it to the "style check" matrix (and renamed it to "static check").


toslunar:
Do you want to keep the name of the key of `requirements['stylecheck']`?

toslunar:
Jenkins, test this please.

Issue #5991, opened 2019-01-14
CUDA memory pool thread safety


niboshi:
Memory pool in CUDA backend seems to have thread safety issues:

* `FreeUnusedBlocks` is not protected by a mutex. This can be easily fixed.
* `Malloc`-`Free` cycle also causes errors.

The test shown below would fail. (note that it would fail even if `FreeUnusedBlocks` is commented out).

```c++
TEST(MemoryPoolTest, FreeUnusedBlocksThreadSafe) {
    static constexpr size_t kRepeat = 100U;
    MemoryPool memory_pool{0, std::make_unique<FixedCapacityDummyAllocator>(0xffffffffU)};

    testing::RunThreads(3, [&memory_pool](size_t thread_index) {
        for (size_t i = 0; i < kRepeat; ++i) {
            switch (thread_index % 3) {
                case 0: {
                    void* ptr = memory_pool.Malloc(1U);
                    memory_pool.Free(ptr);
                } break;
                case 1: {
                    void* ptr = memory_pool.Malloc(1U);
                    memory_pool.Free(ptr);
                } break;
                case 2:
                    // memory_pool.FreeUnusedBlocks();
                    break;
            }
        }
    });
}
```

Error:
```
[ RUN      ] MemoryPoolTest.FreeUnusedBlocksThreadSafe
unknown file: Failure
C++ exception with description "Cannot free out-of-pool memory" thrown in the test body.
```

niboshi:
As for the second problem, it seems `FixedCapacityDummyAllocator` was the cause.
As the allocation sizes are encoded in the pointer values, it returned the same pointer for different allocations. That was the cause.

PR #5993, opened 2019-01-14
Fix FixedCapacityDummyAllocator in CUDA memory pool test

niboshi:
Fixes the second problem of #5991

### DIFF
--- a/chainerx_cc/chainerx/cuda/memory_pool_test.cc
+++ b/chainerx_cc/chainerx/cuda/memory_pool_test.cc
@@ -1,10 +1,13 @@
 #include "chainerx/cuda/memory_pool.h"
 
+#include <map>
 #include <memory>
+#include <mutex>
 
 #include <gtest/gtest.h>
 
 #include "chainerx/error.h"
+#include "chainerx/testing/threading.h"
 
 namespace chainerx {
 namespace cuda {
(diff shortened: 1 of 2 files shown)


hvy:
How about instead keep a dictionary of the sizes allocated by each pointer in the `FixedCapacityDummyAllocator` to simplify the logic?

niboshi:
Fixed.

chainerx_cc/chainerx/cuda/memory_pool_test.cc
```diff
@@ -116,20 +119,28 @@ class FixedCapacityDummyAllocator : public Allocator {
...
         return MallocStatus::kSuccess;
     }
     void Free(void* ptr) noexcept override {
-        intptr_t i = reinterpret_cast<intptr_t>(ptr);  // NOLINT(cppcoreguidelines-pro-type-reinterpret-cast)
-        capacity_ += static_cast<size_t>(i);
-        ++free_called_;
+        uint8_t* mem = static_cast<uint8_t*>(ptr);
+        {
+            std::lock_guard<std::mutex> lock{sizes_mutex_};
+            capacity_ += sizes_.at(mem);
```
hvy: How about deleting the entry from `sizes_` after `Free`? Not sure if we need to add an existence check but I think that's logically more correct (avoiding it from keep growing).

hvy:
Thanks for the fix! I added another small comment. Could you take a look?

chainerx_cc/chainerx/cuda/memory_pool_test.cc
niboshi: Fixed.

hvy:
jenkins, test this please.

hvy:
LGTM!

PR #5999, opened 2019-01-15
Fix python 3.5 incompatibility in Chainer X train_mnist.py example

### LINKED ISSUE #5925: Chainer X MNIST example not compatible with python 3.5
https://github.com/chainer/chainer/blob/ee05a5eacc593d77e8f2dbbaffc6487605938510/chainerx_cc/examples/mnist/train_mnist.py#L170-L172

In Python 3.5 (I used 3.5.2), the Path in pathlib needs to be explicitly converted to string before using it for `open` call.
A workaround could simply be:

```
    x_path = str(path / '{}-images-idx3-ubyte.gz'.format(name))
    y_path = str(path / '{}-labels-idx1-ubyte.gz'.format(name))
```

Also, the flags settings needed to be commented out in order to run properly.
https://github.com/chainer/chainer/blob/ee05a5eacc593d77e8f2dbbaffc6487605938510/chainerx_cc/examples/mnist/train_mnist.py#L179
https://github.com/chainer/chainer/blob/ee05a5eacc593d77e8f2dbbaffc6487605938510/chainerx_cc/examples/mnist/train_mnist.py#L184

Guriido:
Fixes #5925.

### DIFF
--- a/chainerx_cc/examples/mnist_py/train_mnist.py
+++ b/chainerx_cc/examples/mnist_py/train_mnist.py
@@ -181,7 +181,6 @@ def get_mnist(path, name):
     with gzip.open(y_path, 'rb') as fy:
         fy.read(8)  # skip header
         y = np.frombuffer(fy.read(), dtype=np.uint8)
-        # y.flags.writeable = True  # TODO(beam2d): remove this workaround
 
     assert x.shape[0] == y.shape[0]
 
--- a/chainerx_cc/examples/mnist_py/train_mnist.py
+++ b/chainerx_cc/examples/mnist_py/train_mnist.py
@@ -168,20 +168,20 @@ def main():
 
 def get_mnist(path, name):
     path = pathlib.Path(path)
-    x_path = path / '{}-images-idx3-ubyte.gz'.format(name)
-    y_path = path / '{}-labels-idx1-ubyte.gz'.format(name)
+    x_path = str(path / '{}-images-idx3-ubyte.gz'.format(name))
+    y_path = str(path / '{}-labels-idx1-ubyte.gz'.format(name))
 
     with gzip.open(x_path, 'rb') as fx:
         fx.read(16)  # skip header
         # read/frombuffer is used instead of fromfile because fromfile does not
         # handle gzip file correctly
         x = np.frombuffer(fx.read(), dtype=np.uint8).reshape(-1, 784)
-        x.flags.writeable = True  # TODO(beam2d): remove this workaround
+        # x.flags.writeable = True  # TODO(beam2d): remove this workaround
 
     with gzip.open(y_path, 'rb') as fy:
         fy.read(8)  # skip header
         y = np.frombuffer(fy.read(), dtype=np.uint8)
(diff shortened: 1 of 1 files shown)


niboshi:
Thank you so much for the PR!

chainerx_cc/examples/mnist_py/train_mnist.py
```diff
@@ -168,20 +168,20 @@ def main():
...
+    x_path = str(path / '{}-images-idx3-ubyte.gz'.format(name))
+    y_path = str(path / '{}-labels-idx1-ubyte.gz'.format(name))
 
     with gzip.open(x_path, 'rb') as fx:
         fx.read(16)  # skip header
         # read/frombuffer is used instead of fromfile because fromfile does not
         # handle gzip file correctly
         x = np.frombuffer(fx.read(), dtype=np.uint8).reshape(-1, 784)
-        x.flags.writeable = True  # TODO(beam2d): remove this workaround
+        # x.flags.writeable = True  # TODO(beam2d): remove this workaround
```
niboshi: Could you explain why this fix is needed? (What would happen without this fix?)

Guriido: This will raise a ValueError: `ValueError: cannot set WRITEABLE flag to True of this array`.

When checking the flags of the ndarray, the array does not seems to hold the data, so the array is read-only.

```
C_CONTIGUOUS : True
F_CONTIGUOUS : False
OWNDATA : False
WRITEABLE : False
ALIGNED : True
WRITEBACKIFCOPY : False
UPDATEIFCOPY : False
```

hvy: Thanks @Guriido for this patch. Would you mind sharing the version of Python and NumPy you used to produce the above error? I tried locally with Python 3.5.1 and NumPy 1.15.4 and could not reproduce it.

hvy: Oh, I reproduced it with NumPy 1.16.

hvy: Also noticed that these arrays are actually copied before being modified further down. It's okay to remove these lines. Could you do that instead of commenting out?

hvy: This line seems no longer needed.
```suggestion
```

chainerx_cc/examples/mnist_py/train_mnist.py
```diff
@@ -168,20 +168,20 @@ def main():
...
 
     with gzip.open(y_path, 'rb') as fy:
         fy.read(8)  # skip header
         y = np.frombuffer(fy.read(), dtype=np.uint8)
-        y.flags.writeable = True  # TODO(beam2d): remove this workaround
+        # y.flags.writeable = True  # TODO(beam2d): remove this workaround
```
hvy: Ditto.
```suggestion
```

hvy:
jenkins, test this please.

hvy:
Thanks, LGTM!

PR #6006, opened 2019-01-16
[WIP] Add NumPy compatibility module

toslunar:


kmaehashi:
I talked with @okuta regarding this, and we thought it is better to drop support for some older NumPy versions, because installation/upgrade of NumPy became easier as they now provides wheels.

toslunar:
I'm glad to hear that the draft PR leads to the discussion to drop old NumPy.

kmaehashi:
Regarding NumPy compatibility, we decided to ask users if they have any special reasons to use older NumPy (1.9/1.10) and drop the support starting in v7 if there's no problem.

PR #6010, opened 2019-01-16
Fix to use generator in `all` and `any` functions

okapies:
FIx to use `all` and `any` with generator expression instead of list. These functions have a functionality to abort iteration when the condition is (not) met like the following:

```python
>>> class A:
...   def __iter__(self):
...     for i in range(5):
...       print(i)
...       yield i
... 
>>> any(i > 0 for i in A())  # generator
0
1
True
>>> any([i > 0 for i in A()])  # list
0
1
2
3
4
True
```

### DIFF
--- a/chainer/__init__.py
+++ b/chainer/__init__.py
@@ -170,15 +170,15 @@ def is_arrays_compatible(arrays):
     # TODO(niboshi): intel64.mdarray is not supported yet.
     # TODO(niboshi): Delegate array compatibility check to chainerx.
     if (chainerx.is_available()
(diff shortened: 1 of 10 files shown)


kmaehashi:
Jenkins, test this please.

niboshi:
The baseline overhead of a generator `all(...)` is greater than that of list comprehension `all([...])`. Of cource there's a tradeoff to the cost of each single comparison, if it would abort in the middle.

Note also that, `assert all ...` should never abort in the middle. List comprehension should always be preferred in this case.

PR #6016, opened 2019-01-17
WIP: Prevent access to `Variable.device.xp` intensively

okapies:
Prevent access to `Variable.device.xp` as much as possible because the initialization of `device` is really costly.

TODO:
- Writing additional tests

## Micro benchmarking
- number of calls: `10000000`
- device: `native:0`
- batch size: `1`

The script is [here](https://gist.github.com/okapies/303c6316ae11987c17eb7a512dfb067c).

### `inputs: List[chainerx.array]`
| `nargs`        | 1                  | 2                  | 10                 |
|:---------------|:-------------------|:-------------------|:-------------------|
| current        | 22.89022358506918 | 32.13856142386794 | 97.47575618699193 |
| *this fix*     | 13.060540024191141 | 17.535107117146254 | 49.30647446960211 |

### `inputs: List[Variable[chainerx.array]]`
| `nargs`       | 1                  | 2                  | 10                |
|:--------------|:-------------------|:-------------------|:------------------|
| current       | 29.15448709577322 | 42.62322876602411 | 147.7505254484713 |
| *this fix*    | 22.7897168956697 | 31.85619568452239 | 49.87078559026122 |

### `inputs: List[numpy.array]`
| `nargs`        | 1                  | 2                  | 10                 |
|:---------------|:-------------------|:-------------------|:-------------------|
| current        | 11.929559241980314 | 16.040022619068623 | 44.611796759068966 |
| *this fix*     | 12.811775293201208 | 16.961403861641884 | 48.98415441438556 |

### `inputs: List[Variable[numpy.array]]`
| `nargs`        | 1                  | 2                  | 10                 |
|:---------------|:-------------------|:-------------------|:-------------------|
| current        | 22.65941659733653 | 36.52261972054839 | 148.5802201665938 |
| *this fix*     | 14.232808604836464 | 20.443320330232382 | 66.03057060018182 |


okuta:
Please resolve conflicts.

chainer/variable.py
```diff
@@ -603,7 +607,7 @@ def _ensure_grad_var_up_to_date(self):
         # _grad_var._data[0] and self._data[0].grad and recreates _grad_var
         # as necessary. (chainerx.ndarray.grad can be altered independently
         # from chainer)
-        if self.xp is chainerx:
+        if self._has_chainerx_array:
```
niboshi: Is this change needed? I think `self.xp` is now fast enough?

okapies: IMO, we need to do it for two reasons:

1. `xp` is a `property` of the `Variable` which generates further overhead than a function
2. It is safe to refer the independent flag instead of using `array`, `device`  and `xp`, because there is a chance that the attributes are configured as `chainerx` mode but `array` is `None`)

How do you think?

niboshi: >1. `xp` is a `property` of the `Variable` which generates further overhead than a function

Hmm, is it true? In my benchmark, a property is faster than a method and as fast as a global function.
* property/function: ~65ns
* method: ~80ns

>2. It is safe to refer the independent flag instead of using `array`, `device`  and `xp`, because there is a chance that the internal attributes are configured as `chainerx` mode but `array` is `None`)

I think it's a separate issue. We have to reconsider how a variable should behave if the array is `None`.

okapies: Ah, sorry, you are right. Accessing property is faster than the method call, but accessing attribute is still two or three times faster than other ways.

Anyway, I agree to fix it later as another issue. I'll revert them.

```python
>>> import timeit
>>> setup = """
... class A:
...   def __init__(self):
...     self._f = False
... 
... class B:
...   def f(self):
...     return False
... 
... class C:
...   @property
...   def f(self):
...     return False
... 
... a = A()
... b = B()
... c = C()
... """
>>> timeit.timeit("a._f", setup=setup, number=100000000)
2.5808276534080505
>>> timeit.timeit("b.f()", setup=setup, number=100000000)
8.198761623352766
>>> timeit.timeit("c.f", setup=setup, number=100000000)
7.625560574233532
>>> timeit.timeit("f()", setup="def f(): return False", number=100000000)
5.695897337049246
```

niboshi: >accessing attribute is still two or three times faster than other ways.

Yes, but the difference is only ~40ns (in my benchmark). I doubt it's worthwhile.

okapies:
Jenkins, test this please.

chainer/variable.py
```diff
@@ -854,6 +862,7 @@ def array(self, d):
         else:
             self._node._update_data_info(d)  # type: ignore # _node doesn't have value when xp is chainerx # NOQA
             self._data[0] = d
+            self._has_chainerx_array = isinstance(d, chainerx.ndarray)
```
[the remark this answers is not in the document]
okapies: I'll fix it to always assign `False` but it should be fixed in the future. As reported in #6116, `array` has a chance to receive `chainerx.ndarray` as `d` while the variable itself holds another type of ndarray.

okapies:
Thanks!

PR #6026, opened 2019-01-18
Fix warning message for backward on a scalar array

niboshi:
It seems the order is reversed.

But what does this advice mean?
> If the size of this variable accidentally becomes one, set zero to grad.

### DIFF
--- a/chainer/variable.py
+++ b/chainer/variable.py
@@ -1339,7 +1339,7 @@ class Variable(object):
         if self.array.size == 1 and self.grad_var is None:
             if self.array.ndim != 0:
                 warnings.warn(
-                    'Treating a scalar as a variable with only one element'
+                    'Treating a variable with only one element as a scalar'
                     ' in Variable.backward is deprecated. A scalar variable'
                     ' must be a 0-dimensional array. Apply'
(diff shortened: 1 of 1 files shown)


niboshi:
Hmm, I think I understand: we will eventually treat `grad` as zero for single-element arrays, but currently it's 1. This suggestion is to mimick the future behavior. Is that correct?
I think it's better to describe that in detail in order to let users understand, instead of removing the message.
How do you think?

niboshi:
Or perhaps we can simply move to the new behavior now (it has been deprecated since v4).

toslunar:
> This suggestion is to mimick the future behavior. Is that correct?

Yes.

My idea is to show the latter part of the message only if the creator of the variable has more than one output.

niboshi:
Hmm, I think I don't understand yet.
Does this have to do with multiple outputs?

```
x = chainer.Variable(numpy.array([1,], 'f'))
y = x + 1
y.backward()
print(x.grad)
```
```
[1.]
```
We're going to make this `None`, right?

toslunar:
> If the size of this variable accidentally becomes one, set zero to grad.

On this point, let's continue working in a new issue/PR.

PR #5904, opened 2018-12-21
Allow mixing `testing.inject_backend_tests` and `testing.parameterize`

niboshi:
Follow-up #5893

With this fix, it will be possible to mix `inject_backend_tests` and `parameterize` any times in any order.

**no-compat**: With this change, it is no longer possible to inherit from a base class with `@inject_backend_tests` applied.

PR #6028, opened 2019-01-18
Fix `inject_backend_tests` `multi_gpu` test mark

niboshi:
#5904 has caused `testing.attr.multi_gpu` (put by `inject_backend_tests`) to ignore `CHAINER_TEST_GPU_LIMIT` environment variable. This PR fixes it.

(just for reference)
Another fix would be to change `unittest.skipIf` to `pytest.mark.skipif` at this line (I don't know why):
https://github.com/chainer/chainer/blob/9f243e66fc5b6a337273d965b431644d846a9cc7/chainer/testing/attr.py#L60

### DIFF
--- a/chainer/testing/backend.py
+++ b/chainer/testing/backend.py
@@ -185,7 +185,7 @@ def _test_case_generator(base, method_names, params):
 
             # Apply test marks
             for mark in marks:
-                mark(new_method)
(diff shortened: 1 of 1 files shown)


okuta:
jenkins, test this please.

okuta:
LGTM!

PR #6001, opened 2019-01-15
Skip test on Theano<=1.0.2 and NumPy>=1.16.0

niboshi:
Fixes #5997.

PR #6029, opened 2019-01-18
Allow doctest to run in single-GPU environment

niboshi:
Currently doctest can't be run on a single-GPU environment.
As we have increased the required sphinx version in #6001, we can use `:skipif:` option now.

I think `testing.is_requires_satisfied()` can be removed in favor of `doctest_helper.skipif_requires_satisfied()`. It has not been released yet.


okuta:
jenkins, test this please.

okuta:
LGTM!

PR #6030, opened 2019-01-18
Remove `parse_kwargs` from `Variable`

okapies:
`parse_kwargs` is useful functionality but generates non-trivial overhead especially in hotspots. This patch unrolls `parse_kwargs` and place it directly in `Variable`. It also fixes `get_array_module` to reduce the function calls.

### DIFF
--- a/chainer/backend.py
+++ b/chainer/backend.py
@@ -165,15 +165,20 @@ def get_array_module(*args):
         on the types of the arguments.
 
     """
-    if chainerx.is_available() or cuda.available:
-        args = [arg.data if isinstance(arg, chainer.variable.Variable) else arg
-                for arg in args]
+    is_chainerx_available = chainerx.is_available()
+    if is_chainerx_available or cuda.available:
+        arrays = []
+        for arg in args:
+            # Unwrap arrays
+            if isinstance(arg, chainer.variable.Variable):
+                array = arg.data
+            else:
+                array = arg
+            if is_chainerx_available and isinstance(array, chainerx.ndarray):
+                return chainerx
+            arrays.append(array)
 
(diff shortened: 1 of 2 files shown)


okapies:
@toslunar Thanks for pointing out. What is the keyword-only argument in this sense?

toslunar:
This PR enables
```
chainer.Variable(np.array([3, 4], np.float32), 'foo')
```
while Chainer forced users to write
```
chainer.Variable(np.array([3, 4], np.float32), name='foo')
```

niboshi:
I want to drop Python 2 as soon as possible...

I agree `parse_kwargs` is slow and can be optimized.
How about something like caching the default values (instead of creating a tuple at every call) and if `kwargs` is empty, just return a cached tuple?

okapies:
Hmm. Unfortunately, it seems there is no one-size-fits-all solution because the most high-traffic spot requires `requires_grad`. https://github.com/chainer/chainer/blob/88d0d1ae50d7fc2592672c19d9f639406873c42f/chainer/function_node.py#L275-L278

niboshi:
Even in that case, I think `*names_and_values` and `**unexpected` passed to `parse_kwargs` could be cached.
(I don't know if it's worthwhile, though)

okapies:
Do you say the solution like this? https://github.com/chainer/chainer/pull/6030/commits/d02551a276e8ef453969bae3226075c377dfdcac

okapies:
It seems unrolling the loops in `parse_kwargs` is slightly better than the original (about ~10%).

toslunar:
Unspecified unexpected kwarg should raise an error, too.

okapies:
This is superseded by #6033.

PR #6033, opened 2019-01-19
Add `Variable._init_unchecked()` static method for faster instantiation

okapies:
`parse_kwargs` is essential for our user's experience but generates non-trivial overhead especially in hotspots. One problem is that this validation mechanism is hard to optimize or remove as discussed in #6030.

This patch introduces `_unsafe_variable`, which is an optimized factory method of `Variable` only for internal use. It skips the validation to reduce the overhead generated in the framework like `FunctionNode.apply` ([here](https://github.com/chainer/chainer/blob/88d0d1ae50d7fc2592672c19d9f639406873c42f/chainer/function_node.py#L275-L278)) and `UpdateRule.update_core_chainerx` ([here](https://github.com/chainer/chainer/blob/88d0d1ae50d7fc2592672c19d9f639406873c42f/chainer/optimizer.py#L308)).

It also fixes `get_array_module` to reduce the number of function calls.

Note: It is renamed to `Variable._init_unchecked()`.

## Benchmarks of `Variable` constructors
- number of calls: `10000000`
- device: `native:0`
- batch size: `1`
- require_grad: `False`

The script is [here](https://gist.github.com/okapies/207001ce67994fc7bd02736ef232aaab).

## `data` is `chainerx.ndarray`
| | `Variable` | `_unsafe_variable` |
|:--|:--|:--|
| current | 49.95680534094572 | - |
| *this* | 43.18582313135266 | 31.15744174271822 |

## `data` is `numpy.ndarray`
| | `Variable` | `_unsafe_variable` |
|:--|:--|:--|
| current | 38.3368557356298 | - |
| *this* | 32.35968181490898 | 19.19199386611581 |

### DIFF
--- a/chainer/backend.py
+++ b/chainer/backend.py
@@ -177,9 +177,8 @@ def get_array_module(*args):
             if is_chainerx_available and isinstance(array, chainerx.ndarray):
                 return chainerx
             arrays.append(array)
-        else:
-            if cuda.available:
-                return cuda.cupy.get_array_module(*arrays)
+        if cuda.available:
+            return cuda.cupy.get_array_module(*arrays)
     return numpy
 
 
--- a/chainer/backend.py
+++ b/chainer/backend.py
@@ -165,15 +165,20 @@ def get_array_module(*args):
         on the types of the arguments.
 
     """
-    if chainerx.is_available() or cuda.available:
-        args = [arg.data if isinstance(arg, chainer.variable.Variable) else arg
-                for arg in args]
+    is_chainerx_available = chainerx.is_available()
+    if is_chainerx_available or cuda.available:
+        arrays = []
+        for arg in args:
+            # Unwrap arrays
+            if isinstance(arg, chainer.variable.Variable):
+                array = arg.data
+            else:
+                array = arg
+            if is_chainerx_available and isinstance(array, chainerx.ndarray):
+                return chainerx
+            arrays.append(array)
 
-    if (chainerx.is_available()
-            and any([isinstance(a, chainerx.ndarray) for a in args])):
-        return chainerx
-    elif cuda.available:
-        return cuda.cupy.get_array_module(*args)
+        return cuda.cupy.get_array_module(*arrays)
     else:
         return numpy
 
--- a/chainer/backend.py
+++ b/chainer/backend.py
@@ -177,10 +177,10 @@ def get_array_module(*args):
             if is_chainerx_available and isinstance(array, chainerx.ndarray):
                 return chainerx
             arrays.append(array)
-
-        return cuda.cupy.get_array_module(*arrays)
-    else:
-        return numpy
+        else:
+            if cuda.available:
+                return cuda.cupy.get_array_module(*arrays)
+    return numpy
 
 
 def get_device_from_array(*arrays):
--- a/chainer/variable.py
+++ b/chainer/variable.py
@@ -538,7 +538,10 @@ class Variable(object):
 
         self._init_attrs(data, name, grad, requires_grad, None)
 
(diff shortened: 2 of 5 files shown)


okapies:
One idea is that makes the constructor of `Variable` a factory by overriding `__new__` like [this](https://github.com/chainer/chainer/pull/6033/commits/597a792b053b1435971a33c2d582879095118b8e#diff-e131a8dfc2ea641e7e7c63597b9b0d8eL512), and split all logic of `Variable` into `ChainerxVariable` and `NonChainerxVariable` to remove `xp is chainerx.ndarray` condition *if the array type held by `Variable` is never changed*.

chainer/backend.py
```diff
@@ -165,17 +165,22 @@ def get_array_module(*args):
...
+        for arg in args:
+            # Unwrap arrays
+            if isinstance(arg, chainer.variable.Variable):
+                array = arg.data
+            else:
+                array = arg
+            if is_chainerx_available and isinstance(array, chainerx.ndarray):
+                return chainerx
+            arrays.append(array)
+        else:
```
niboshi: This `else` is not needed?

chainer/variable.py
```diff
@@ -1578,6 +1600,23 @@ def add_cand(cand):
     grads.assert_no_grads()
 
 
+def _unsafe_variable(data, name=None, grad=None, requires_grad=True,
```
niboshi: How about making this a `@staticmethod` of `Variable`?
The name could be arguable. How about `Variable._init_unchecked`? (`init` in order to express it's an alternative to `__init__`. `unchecked` because it's more clearly express the behavior than `unsafe`).

chainer/variable.py
```diff
@@ -508,12 +510,23 @@ class Variable(object):
...
-        # type: (types.NdArray, **tp.Any) -> None
+        # type: (tp.Optional[types.NdArray], **tp.Any) -> None
 
-        name, grad, requires_grad = argument.parse_kwargs(
-            kwargs, ('name', None), ('grad', None), ('requires_grad', True),
-            volatile='volatile argument is not supported anymore. '
-            'Use chainer.using_config')
+        # Retrieve keyword-only arguments as done in parse_kwargs()
+        # (These checks are required to keep backward compatibility)
+        name = kwargs.pop('name', None)
```
niboshi: Is this unrolling of kwargs handling still needed?

okapies:
@niboshi I added `_unsafe_variable` factory method instead of `UnsafeVariable` class. It seems to both improve the performance and keep the sanity at the same time. Could you please check it?

chainer/variable.py
```diff
@@ -523,6 +536,12 @@ def __init__(self, data=None, **kwargs):
                     array_types[-1], type(data))
                 raise TypeError(msg)
 
+        self._init_attrs(data, name, grad, requires_grad, None)
+
+    def _init_attrs(self,
```
niboshi: Fix the position of line break. Either before `self` or before the parameter that does not fit the line.

niboshi: The name of this method also has a room to discuss.
I think "attributes" is nonessential. How about `_init_common`?

chainer/variable.py
okapies: I see. Fixed to use the original.

chainer/variable.py
okapies: Hmm, what about `_init_impl`?

niboshi: That also sounds good to me.

niboshi:
Jenkins, test this please

okapies:
Jenkins, test this please

niboshi:
LGTM

PR #6043, opened 2019-01-21
Fix `chainerx.maximum` to restore CUDA device

asi1024:
Use `CudaSetDeviceScope` instead of `cudaSetDevice()`.


niboshi:
Jenkins, test this please

niboshi:
Jenkins, test this please

PR #6044, opened 2019-01-21
Test if the default CUDA device keeps being 0 after each test

asi1024:
~Merge after #6043.~

PR #6047, opened 2019-01-22
[backport] Test if the default CUDA device keeps being 0 after each test

niboshi:
Backport of #6044


niboshi:

PR #6048, opened 2019-01-22
Fix chainermn tests

shu65:
This PR fixes chainermn tests.
The causes of the chainermn test clash are here
1.  use `MultiprocessIterator` without forkserver
2. Test skip mistake

### DIFF
--- a/tests/chainermn_tests/communicator_tests/test_communicator.py
+++ b/tests/chainermn_tests/communicator_tests/test_communicator.py
@@ -300,6 +300,9 @@ class TestPureNcclCommunicator(unittest.TestCase):
 class TestDifferentDtype(unittest.TestCase):
 
     def setup(self, gpu):
+        if self.communicator.size != 2:
+            pytest.skip('This test is for two processes')
+
         if gpu:
             self.communicator = chainermn.create_communicator('hierarchical')
             self.device = self.communicator.rank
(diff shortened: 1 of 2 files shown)


shu65:
Jenkins, test this please.

tests/chainermn_tests/communicator_tests/test_communicator.py
```diff
@@ -302,15 +302,18 @@ class TestDifferentDtype(unittest.TestCase):
...
-            self.device = self.communicator.rank
-            chainer.cuda.get_device_from_id(self.device).use()
         else:
             self.communicator = chainermn.create_communicator('naive')
-            self.device = -1
 
         if self.communicator.size != 2:
             pytest.skip('This test is for two processes')
 
+        if gpu:
```
keisukefukuda: What if the test pattern is "1 GPU x 2 nodes = 2 processes" ?
I think the `rank` should always be `intra_rank`. What do you think?

shu65: In the case, it passes without problems because 'rank' is 0 or 1.

I think that the pr version is better than using `intra_rank`. 
Because, we can not check the test case with 1 GPU x 2 nodes if we use  `intra_rank`.

keisukefukuda: Hmm.
I would like to consider the following two cases:
(A) `2x1`: Run the test on 1 machine, which has 2 GPUs, so `communicator.size == 2`
(B) `1x2`: Run the test on 2 machines, each of which has 1 GPU, so `communicator.size == 2`

In the case of (A), device IDs should be 0 and 1 respectively, so the test succeeds without problem, as you pointed out.
In the case of (B), both device IDs must be 0, because each machine has only 1 GPU. Test test does not work properly in this case.

shu65: I understood. I will fix it.

PR #6050, opened 2019-01-22
chainerx.ndarray.item

### LINKED ISSUE #6019: Add chainerx.ndarray.item
Background: #5797, #6007.

niboshi:
Fixes #6019 

Arguments are not supported yet.


tests/chainerx_tests/unit_tests/test_array.py
```diff
@@ -42,6 +44,81 @@ def test_init_with_device(shape, dtype_spec, device):
...
+    # Cast to int
+    if should_cast_succeed(int):
+        assert type(int(a_chx)) is int
+        assert int(a_np) == int(a_chx)
+    # Cast to bool
+    if should_cast_succeed(bool):
+        assert type(bool(a_chx)) is bool
+        assert bool(a_np) == bool(a_chx)
+
+    # item()
```
hvy: How about using `chainerx.testing.numpy_chainerx_array_equal` since this is a NumPy compat operation and we'll probably want to use this decorator when adding support for the arguments?

niboshi: What do you mean? We're not comparing arrays.

hvy: Can you not write the test in this fashion https://github.com/chainer/chainer/blob/master/tests/chainerx_tests/unit_tests/routines_tests/test_math.py#L33 ? Python scalars can apparently also be seen as "array_like" (see https://stackoverflow.com/questions/40378427/numpy-formal-definition-of-array-like-objects).

niboshi: Such generalization would only introduce risks. We can simply separate tests.

hvy: Okay, let's leave it as is. We discussed offline that this decorator shouldn't be used for non-array types, including Python scalars.

hvy:
jenkins, test this please.

Issue #6025, opened 2019-01-18
chainerx.max_pool floating point exception for stride=0


hvy:
Passing `stride` as `0` to `chainerx.max_pool` causes a crash. The strides should be validated prior to this. This issue was reported by @shinh . Reproduction code follows.

```python
import chainerx
chainerx.max_pool(chainerx.random.uniform(-1, 1, (1, 3, 32, 32)), ksize=3, stride=0)
# 48279 floating point exception (core dumped) 
```

As for reference, Chainer raises a  `ZeroDivisionError`.

PR #6064, opened 2019-01-23
DimsFormatter to print StackVector<int64_t, kMaxNdim>

hvy:
This PR makes `StackVector` printable using `operator<<`. It is convenient when generating informative error messages since these objects are commonly passed between functions and often need to be validated (e.g. the stride in convolution).

### DIFF
--- a/chainerx_cc/chainerx/stack_vector.h
+++ b/chainerx_cc/chainerx/stack_vector.h
@@ -5,10 +5,8 @@
 #include <cstddef>
 #include <initializer_list>
 #include <iterator>
-#include <sstream>
 #include <type_traits>
 
-#include "chainerx/constant.h"
 #include "chainerx/macro.h"
 
 namespace chainerx {
@@ -195,30 +193,5 @@ private:
     size_type n_{0};
     BaseContainer d_{};
 };
-
-// Formatter to print StackVector containing integral elements as e.g. '[]' or '[1, 2, 3]'.
-class DimsArgFormatter {
-public:
-    explicit DimsArgFormatter(const StackVector<int64_t, kMaxNdim>& stack_vector) : stack_vector_{stack_vector} {}
-
-    void Print(std::ostream& os) const {
-        os << "[";
-        for (auto iter = stack_vector_.begin(); iter != stack_vector_.end(); ++iter) {
-            if (iter != stack_vector_.begin()) {
-                os << ", ";
-            }
-            os << *iter;
-        }
-        os << "]";
-    }
-
-private:
-    const StackVector<int64_t, kMaxNdim>& stack_vector_;
-};
-
-inline std::ostream& operator<<(std::ostream& os, const DimsArgFormatter& formatter) {
-    formatter.Print(os);
-    return os;
-}
 
 }  // namespace chainerx
--- a/chainerx_cc/chainerx/stack_vector.h
+++ b/chainerx_cc/chainerx/stack_vector.h
@@ -195,16 +195,33 @@ private:
     BaseContainer d_{};
 };
 
+// Formatter to print StackVector containing integral elements as e.g. '[]' or '[1, 2, 3]'.
 template <typename T, stack_vector_detail::size_type N>
-std::ostream& operator<<(std::ostream& os, const StackVector<T, N>& stack_vector) {
-    os << "[";
-    for (auto iter = stack_vector.begin(); iter != stack_vector.end(); ++iter) {
-        if (iter != stack_vector.begin()) {
-            os << ", ";
+class DimArgFormatter {
+public:
+    static_assert(std::is_integral<T>::value, "DimArgFormatter requires intergral elements.");
+
+    DimArgFormatter(const StackVector<T, N>& stack_vector) : stack_vector_{stack_vector} {}
+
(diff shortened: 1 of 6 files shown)


hvy:
Motivated by https://github.com/chainer/chainer/issues/6025.

hvy:
Thanks for the suggestion. That makes sense. We can also specialize it for specific types (i.e. `int64_t`)? Then it can be abstracted and reused beyond "DimArg"s or you think we might want to have different formatters depending on usage even when the type is the same?

```c++
template <stack_vector_detail::size_type N>
std::ostream& operator<<(std::ostream& os, const StackVector<int64_t, N>& stack_vector) {
    os << "[";
    for (auto iter = stack_vector.begin(); iter != stack_vector.end(); ++iter) {
        if (iter != stack_vector.begin()) {
            os << ", ";
        }
        os << *iter;
    }
    return os << "]";
}
```

niboshi:
Yes. Consider if `T` were time_t`, for example.

hvy:
Could you elaborate on the comment above?

hvy:
Introduced `DimArgFormatter` with your former ctor suggestion.

hvy:
jenkins, test this please.

chainerx_cc/chainerx/stack_vector.h
```diff
@@ -194,4 +195,33 @@ class StackVector {
     BaseContainer d_{};
 };
 
+// Formatter to print StackVector containing integral elements as e.g. '[]' or '[1, 2, 3]'.
```
niboshi: Do not implement here: this is not a feature of `StackVector`.

chainerx_cc/chainerx/stack_vector.h
```diff
@@ -194,4 +195,33 @@ class StackVector {
...
+template <typename T, stack_vector_detail::size_type N>
+class DimArgFormatter {
```
niboshi: How about `DimsArgFormatter`? A single argument holds multiple dimensions.

chainerx_cc/chainerx/stack_vector.h
```diff
@@ -194,4 +195,33 @@ class StackVector {
     BaseContainer d_{};
 };
 
+// Formatter to print StackVector containing integral elements as e.g. '[]' or '[1, 2, 3]'.
+template <typename T, stack_vector_detail::size_type N>
```
niboshi: There's no need of template. Just use `T=int64_t` and `N=kMaxNdim`.

chainerx_cc/chainerx/stack_vector.h
```diff
@@ -194,4 +195,33 @@ class StackVector {
...
+class DimArgFormatter {
+public:
+    static_assert(std::is_integral<T>::value, "DimArgFormatter requires intergral elements.");
+
+    explicit DimArgFormatter(const StackVector<T, N>& stack_vector) : stack_vector_{stack_vector} {}
+
+    void Print(std::ostream& os) const {
```
niboshi: Make this function private, and make `operator<<` friend.

niboshi: How about `Format` instead of `Print`?
It matches the class name.

chainerx_cc/chainerx/stack_vector.h
hvy: How moving it to a new `chainerx_cc/chainerx/stack_vector_formatter.h`?

chainerx_cc/chainerx/stack_vector.h
hvy: Okay, will fix.

chainerx_cc/chainerx/stack_vector.h
hvy: Will fix too.

chainerx_cc/chainerx/stack_vector.h
hvy: Hm, I agree `Format` better matches the class name but I think `Print` is appropriate. The formatting is already implied by the class name while what's important to convey is that it also prints (of course this is also clear from the argument but still more important that reemphasizing the formatting). Actually, these "naming conventions" were borrowed from `array_repr.h`.

chainerx_cc/chainerx/stack_vector.h
hvy: Ah, that was also you suggestion. Changed to it.

chainerx_cc/chainerx/stack_vector.h
hvy: Fixed.

chainerx_cc/chainerx/stack_vector.h
hvy: Fixed.

chainerx_cc/chainerx/stack_vector.h
hvy: Fixed.

niboshi:
Jenkins, test this please

hvy:
Jenkins, test this please.

hvy:
Sorry added missing test file.

niboshi:
LGTM

PR #6066, opened 2019-01-23
Validate `{Max,Average}Pool` `kernel_size` and `stride`

hvy:
Fixes https://github.com/chainer/chainer/issues/6025.

~Depends on https://github.com/chainer/chainer/pull/6064.~


niboshi:
LGTM

PR #6067, opened 2019-01-23
Validate Conv, ConvTranspose stride

hvy:
Extension of https://github.com/chainer/chainer/pull/6066 (pooling) to convolutions.

~Depends on https://github.com/chainer/chainer/pull/6064.~


niboshi:
LGTM

PR #6068, opened 2019-01-23
[WIP] Fix Travis script

toslunar:
The current script only checks last lines of steps, because `set -e` is ignored.

> The -e setting shall be ignored when executing the compound list following the while, until, if, or elif reserved word, a pipeline beginning with the ! reserved word, or any command of an AND-OR list other than the last.

### DIFF
--- a/scripts/ci/travis/run-tests.sh
+++ b/scripts/ci/travis/run-tests.sh
@@ -9,7 +9,10 @@ phase="$1"
 
 # Check phase argument
 case "$phase" in
-    before_install|install|script)
+    before_install|install)
+        ;;
+    script)
+        step_num="$2"
         ;;
     *)
         echo "Unknown phase: ${phase}" >&2
@@ -243,8 +246,10 @@ case "${CHAINER_TRAVIS_TEST}" in
                 run_step python_mypy_check_deps
             ;;
             script)
-                run_step python_style_check
-                run_step python_mypy_check
+                case "$step_num" in
+                    1) run_step python_style_check ;;
+                    2) run_step python_mypy_check ;;
+                esac
             ;;
         esac
         ;;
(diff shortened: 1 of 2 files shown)


niboshi:
Could you post a link to the text you quoted?

niboshi:
Also could you point out the exact location of the problem?

toslunar:
The quote I pasted seems old.

https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html may be the newest.

> `-e`
Exit immediately if a pipeline (see Pipelines), which may consist of a single simple command (see Simple Commands), a list (see Lists), or a compound command (see Compound Commands) returns a non-zero status. The shell does not exit if the command that fails is part of the command list immediately following a while or until keyword, part of the test in an if statement, part of any command executed in a && or || list except the command following the final && or ||, any command in a pipeline but the last, or if the command’s return status is being inverted with !. If a compound command other than a subshell returns a non-zero status because a command failed while -e was being ignored, the shell does not exit. A trap on ERR, if set, is executed before the shell exits.

toslunar:
https://travis-ci.org/chainer/chainer/jobs/483290510 succeeded while `autopep8` failed.

niboshi:
Hmm, I don't think shell functions are included in the "compound command". (is that what you assume?)
I locally checked shell functions exit on failure as expected.

scripts/ci/travis/run-tests.sh
```diff
@@ -231,29 +227,12 @@ step_chainerx_clang_tidy() {
...
 run_step() {
-    # In case of failure, CHAINER_TEST_STATUS is incremented by 1.
-
     step="$1"
     shift
     echo "=== Step: $step $@"
-
-    source "$CHAINER_BASH_ENV"
-
-    step_"$step" "$@" || CHAINER_TEST_STATUS=$((CHAINER_TEST_STATUS + 1))
```
toslunar: `step_"$step" "$@"` does not stop on error.

niboshi:
That's what's expected.
Each step should not exit on failure immediately. `CHAINER_TEST_STATUS` is incremented instead.
I think the source of the problem is something different.

toslunar:
Other fixes I came up are
1. Join commands in a step with `&&`
2. Use `run_step` innermost: `for f in "${check_targets[@]}"; do run_step foo test -e "$f"; done`

niboshi:
I  think I figured out the cause.

(From the link you pasted)
>If a compound command or shell function executes in a context where -e is being ignored, none of the commands executed within the compound command or function body will be affected by the -e setting, even if -e is set and a command returns a failure status. If a compound command or shell function sets -e while executing in a context where -e is ignored, that setting will not have any effect until the compound command or the command containing the function call completes.

That is, in the compound `func1 || func2`, `func1` is executed in a context where `-e` is ignored.
In this context, commands in the body of `func1` ignores `-e` and there's no way to make them behave as if `-e` were set.

niboshi:
As a fix, perhaps the cleanest way is to run each step in a separate shell, just as in ChainerX tests (`scripts/ci/chainerx/jenkins`).

niboshi:
OK, I will.

niboshi:
@toslunar 
I made a PR (#6082). Could you take a look?

niboshi:
Fixed via #6082.

Issue #4074, opened 2017-12-10
when v is empty variable, `v.copydata(v1)` is calling `Variable.initialize` **bug**


EmbraceLife:
**environment**
MacOS Higher Sierra 10.13.1
chainer version 3.0.0
numpy version 1.13.3

**example code**
```python
import chainer
import chainer.cuda as cuda
import numpy as np
v = chainer.Variable()
v1 = chainer.Variable(np.array([1,3]))
v.copydata(v1)
```
**error message**
```(Pdb++) v.copydata(v1)
*** AttributeError: 'Variable' object has no attribute 'initialize'
```

**cause of error**
located at https://github.com/chainer/chainer/blob/master/chainer/variable.py#L782
```
self.initialize(src.shape)
```
I guess this line of code expect `self` to be a `Parameter` instance, but it is actually a `Variable` instance here, which does not have `initialize` method.

**proposed solution**
- specify what kind of input is required, for example, when using an empty Variable instance, this instance has to be Parameter instance too.

PR #6072, opened 2019-01-24
Fix to refuse an array from a different device in `Variable.array`

### LINKED ISSUE #6058: Variable.array doesn't handle an array on a different device
The setter of `array` in `Variable` doesn't handle an array coming from a different device. For example, if you assign `chainerx.ndarray` into `array` property of a `Variable` which is initialized with a `numpy.ndarray`, its `device` property still claims it is numpy array.

```python
>>> import numpy as np
>>> import chainerx as chx
>>> import chainer
>>> npa = np.array([0.], dtype=np.float32)
>>> cha = chx.array([0.], dtype=chx.float32)
>>> v = chainer.Variable(npa)
>>> v.device
<CpuDevice (numpy)>
>>> v.array = cha
>>> v.device
<CpuDevice (numpy)>  # should be <ChainerxDevice native:0>
```

okapies:
This patch adds `Device.is_compatible_array()` method and fixes to reject an array coming from a non-compatible device in the setter of `Variable.array`.

Fixes #6058. Related to #6116. Blocked by ~#7487~.


okapies:
Jenkins, test this please.

chainer/variable.py
```diff
@@ -1792,18 +1808,19 @@ def initialize(self, shape):
 
         data = initializers.generate_array(
             self.initializer, shape, xp, device=device)
+        # Convert the array for iDeep.
+        # (generate_array produces numpy array even if device is Intel64)
+        # TODO(niboshi): This could be done in generate_array().
+        if isinstance(device, intel64.Intel64Device):
+            data = device.send_array(data)
```
okapies: It converts `data` to `ideep.mdarray` before `self.array = data` because now `array()` rejects a numpy array if `device is Intel64Device`.

okapies:
Jenkins, test this please.

chainer/variable.py
```diff
@@ -1143,21 +1160,22 @@ def _to_device(self, device, allow_unchaining):
...
+        # Note: It affects the whole behavior of Variable's methods.
+        #       We must place them in this place.
         self._device = device
         self._has_chainerx_array = is_chainerx
 
-        if arr is None:
-            return
-
-        if backend.get_device_from_array(arr) == device:
+        if arr is None or old_device == device:
```
niboshi: How about checking `old_device == device` earlier?

tests/chainer_tests/test_variable.py
```diff
@@ -285,6 +286,35 @@ def test_uninitialized(self):
         assert a.xp is np
         assert a._has_chainerx_array is False
 
+    def check_array(self, xp, x, y, requires_grad=True):
```
niboshi: Some suggestions for the test:
* We should test for every possible pairs of devices. How about making a separate `TestCase` and parameterizing using two `inject_backend_config`s?
* The code inside `pyteset.raises` should be minimal. Enclosing the entire call of `check_array` could make the test pass unexpectedly by an assertion failure from some unexpected place.

chainer/variable.py
```diff
@@ -511,6 +511,10 @@ class Variable(object):
     # TODO(hvy): Avoid modifying this variable from outside this class.
     _chainerx_fallback_array = None
 
+    _data = None  # type: tp.List[tp.Optional[types.NdArray]]
```
niboshi: I think it's better to place these attribute declarations before ChainerX-specific attributes.

niboshi: Please remove type hints from internal interfaces and code.

chainer/variable.py
okapies: It depends on the conclusion of #6116 because `Parameter` expects `self._device` has a valid configuration even if `arr is None`. We can move `old_device == device` independently but I’d like to wait until the discussion is finished.

chainer/variable.py
okapies: Hmm, it is required to remove type errors in the user-facing methods. How do you think about this kind of cases?

chainer/variable.py
okapies: I'd like to leave it because #6116 has not yet been concluded.

chainer/variable.py
niboshi: Could you elaborate that? I thought internal attributes should not affect users' typing checks.

chainer/variable.py
niboshi: OK.

chainer/variable.py
okapies: If we specify types for arguments, mypy try to check if all variables (including the arguments) in the function scope are sound. It doesn't affect the user's type checking, but leaving these errors are not convenient because the type error output is useful for adding new type hints even if CI doesn't run type checking.

niboshi: Mmm, I don't still understand. What do you mean by it doesn't affect user's type checking? It sounds like it won't cause any errors.
Could you post actual commands?

okapies: I tend to run `mypy chainer` in the `chainer` directory when adding type hints to Chainer for inspecting *type bugs*.

niboshi: Is that for checking chainer's internal type consistency?
As discussed previously (#6146), we're not going to guarantee such check would succeed.

okapies: Yes, but sometimes, in my experience, the types of interface method are not obvious from the documentation. For example, I noticed #4074 independently because of the mypy error. In other case, the specification of `Initializer` [allows a scalar value](https://docs.chainer.org/en/stable/reference/initializers.html) but it is not explicitly documented in [some APIs](https://github.com/chainer/chainer/blob/7e4f072c6d42cd345ac216cfdf353bce891140e0/chainer/links/connection/linear.py#L34-L38). I noticed `InitializerSpec` should contain `ScalarValue` when running mypy.

I agree that non-public interfaces and variables should prevent from using type hints, but minimum usage is very useful for detecting problems described like above. How do you think?

niboshi: So do you mean you want to restart using mypy?
It's an option of course, but the team should agree with that first, as the current concensus is the opposite.

By the way, `L.Linear` docs has a link to initializers, so I'd say it's documented.

okapies: Ok, I'm honestly wondering what means "use type hints only in the public interfaces," but I understood that the type hints basically should not be used in Chainer. I'll remove this hint and create a PR for the documentations. Thanks for the discussion.

niboshi:
Jenkins, test this please

niboshi:
It seems some tests failed.

okapies:
Thanks for the catch. I'll check it.

niboshi:
@okapies How is it going?

okapies:
Jenkins, test this please.

okapies:
I remember that what are problems. For example, the following failure is *reasonable* because `Intel64Device` is a different instance from `CpuDevice`. Should we accept such the case? @niboshi 

```
ValueError: The specified array is invalid. Variable is configured against @intel64 but it is on @numpy.
```

niboshi:
I see. It should be accepted because `Intel64Device` handles `numpy.ndarray` as well.
This seems to be a difficult problem.
You could use `Device.supported_array_types` to check the array type, but it's not enough because `ChainerxDevice` with `cuda:0` should not accept `chainerx.ndarray` on `cuda:1`.

niboshi:
Hmm, this fix looks a bit too adhoc.
`ChainerxDevice` and other devices should behave similarly regarding this point.
One possible solution would be to introduce somethig like `Device.is_arrays_compatible(arrs)` (and perhaps remove `Device.supported_array_types`).

okapies:
Added `Device.is_compatible_array` method to check if the specified array is compatible with the device. I also found that some tests try to do `variable.array = None` as a result of invoking `copy`. I changed to accept these cases but it should be solved in (3.) of #6116.

okapies:
I found that `tests/chainer_tests/test_variable.py` fails in single GPU environment (maybe this PR add the check for `Device`). Is there good workaround for it?

okapies:
Jenkins, test this please.

okapies:
I got the same errors in `master` branch. I'll post an issue about it. https://github.com/chainer/chainer/pull/6072#issuecomment-496884300

okapies:
Jenkins, test this please.

okapies:
@niboshi All issues are solved I think. Could you please check it?

chainer/variable.py
```diff
@@ -1185,15 +1203,15 @@ def _to_device(self, device, allow_unchaining):
         elif not was_chainerx and is_chainerx:
             self._chainerx_name = self._node.name
 
+        # Note: It affects the whole behavior of Variable's methods.
+        #       We must place them in this place.
```
niboshi: Hmm, I don't get what this comment is trying to convey.

chainer/variable.py
```diff
@@ -849,27 +851,36 @@ def array(self):
...
+        # type: (tp.Optional[types.NdArray]) -> None
+
+        d_old = self._data[0]
+        if d_old is not None and d is not None:
+            old_device = self.device
+            if not old_device.is_compatible_array(d):
+                new_device = backend.get_device_from_array(d)
+                raise ValueError(
+                    "The specified array is incompatible. Variable is "
+                    "configured against {} but it is on {}.".format(
```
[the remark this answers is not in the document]
okapies: Fixed. But can we use multiple lines in the exception message?

chainer/variable.py
okapies: `self._device` and `self._has_chainerx_array` is referenced in many places both explicitly and implicitly. Reassining them leads to change the behavior of some methods and properties. For example, `grad_var = self.grad_var` doesn't work as expected if we place them before it.

chainer/variable.py
```diff
@@ -951,6 +962,7 @@ def _set_grad_without_check(self, g):
 
     @property
     def grad(self):
+        # type: () -> tp.Optional[types.NdArray]
```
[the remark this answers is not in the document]
okapies: Moved to #7400.

chainer/variable.py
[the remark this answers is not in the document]
niboshi: Yes. Multiline message is used in many places.

chainer/variable.py
niboshi: I don't see any particular reason to make such comment to these lines.
It's in fact very common that timing of attribute assignment matters.

chainer/_backend.py
```diff
@@ -105,3 +105,16 @@ def use(self):
         """Makes the device current in the current thread.
          """
         pass
+
+    def is_array_compatible(self, array):
```
niboshi: `is_array_compatible` needs its own tests.
Could you separate a PR?

chainer/variable.py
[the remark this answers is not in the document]
okapies: Ok, I'll fix.

chainer/variable.py
okapies: Hmm, OK. I'll remove it.

chainer/_backend.py
okapies: I see. Moved to #7487.

chainer/variable.py
```diff
@@ -858,18 +860,27 @@ def array(self):
     def array(self, d):
         # type: (tp.Optional[types.NdArray]) -> None
 
+        d_old = self._data[0]
+        if d_old is not None and d is not None:
+            old_device = self.device
+            if not old_device.is_array_supported(d):
+                new_device = backend.get_device_from_array(d)
+                raise ValueError(
```
niboshi: This check won't be triggered if `self._data[0] is None`. I think it should.

okapies: `d_old is None` can be `True` as a *normal* case when unpickle a `Parameter`.

1. A `Parameter` instance is created in uninitialized state
2. The unpickler calls `__reduce__` and `_recover_parameter`
3. `_recover_parameter` assigns recovered data to `array` property

niboshi: Yes, that's true. Is it valid to assign an array of different device in above cases?

okapies: Hmm, do you mean the case when `self._data[0] is None` and `self.device is not None`? In that case, we should check `old_device` instead of `d_old`.

okapies: Ah, I just remember that `self.device` (= `old_device`) returns `CpuDevice` if `self._data[0] is None`.

niboshi: Can you add a TODO? We should check the device in such cases anyway.

tests/chainer_tests/test_variable.py
```diff
@@ -1420,6 +1420,71 @@ def test_bool_inconvertible(self):
...
+        requires_grad_y = self.requires_grads[1]
+
+        x = self._gen_array(device_x, self.x_shape, requires_grad_x)
+        y = self._gen_array(device_y, self.x_shape, requires_grad_y)
+
+        v = chainer.Variable(x, requires_grad=requires_grad_x)
+
+        should_fail = (
+            (
+                not device_x == device_y
```
niboshi: `device_x != device_y`

tests/chainer_tests/test_variable.py
```diff
@@ -1420,6 +1420,71 @@ def test_bool_inconvertible(self):
...
+                and not device_x == intel64.Intel64Device) or
```
niboshi: Do you mean `isinstance`?
Currently this expression should always evaluate to `True`.

tests/chainer_tests/test_variable.py
```diff
@@ -1420,6 +1420,71 @@ def test_bool_inconvertible(self):
...
+            (xp_x is chainerx and (requires_grad_x or requires_grad_y)))
+
+        if should_fail:
+            # should not accept an array from a different device
+            # or `x` or `y` is a chainerx array which requires gradient
+            with pytest.raises(ValueError):
+                v.data = y
```
niboshi: `v.array = y`

tests/chainer_tests/test_variable.py
```diff
@@ -1420,6 +1420,71 @@ def test_bool_inconvertible(self):
...
+        else:
+            # Should succeed
+            v.data = y
```
niboshi: Ditto.

tests/chainer_tests/test_variable.py
```diff
@@ -1420,6 +1420,71 @@ def test_bool_inconvertible(self):
...
+            xp_x.testing.assert_array_equal(v.data, y)
```
niboshi: Ditto.

tests/chainer_tests/test_variable.py
```diff
@@ -1420,6 +1420,71 @@ def test_bool_inconvertible(self):
...
+            (False, False), (True, False), (False, True), (True, True),
+        ],
+    }))
+@testing.backend.inject_backend_tests(None, _set_data_backend_params)
+@testing.backend.inject_backend_tests(None, _set_data_backend_params)
+class TestVariableSetData(unittest.TestCase):
+
+    def _gen_array(self, device, shape, requires_grad):
+        if shape is None:
+            a = None
```
niboshi: return `None`

tests/chainer_tests/test_variable.py
```diff
@@ -1420,6 +1420,71 @@ def test_bool_inconvertible(self):
...
+@testing.parameterize(*testing.product(
+    {
+        'x_shape': [(10,), ()],  # TODO(okapies): add tests for None
+        'requires_grads': [
+            (False, False), (True, False), (False, True), (True, True),
+        ],
+    }))
+@testing.backend.inject_backend_tests(None, _set_data_backend_params)
+@testing.backend.inject_backend_tests(None, _set_data_backend_params)
+class TestVariableSetData(unittest.TestCase):
```
niboshi: `TestVariableSetArray`

tests/chainer_tests/test_variable.py
okapies: Good catch, fixed.

tests/chainer_tests/test_variable.py
okapies: Fixed (it was to handle #6335).

chainer/variable.py
okapies: I tried to fix the case like the following:

1. Disable caching in the getter of `device` property when `self._data[0] is None`
    - It means `self._device` is always `None` before `self._data[0]` is assigned once
2. Check `_device` field directly in the setter of `array` property instead of accessing `device` property

I'm running the tests...

niboshi: Hmm, disabling cache might have performance implication.

okapies: I think it is not a major issue because:

1. `device` will not be called frequently when `Variable`/`Parameter` is uninitialized
2. Creating `CpuDevice` instance incurs relatively low overhead

niboshi: Is that really true?

>1. `device` will not be called frequently when `Variable`/`Parameter` is uninitialized

I don't have a total picture of how `device` attribute is used by Chainer itself or users.
The only thing I can say is `device` is meant to be lightweight.

>2. Creating `CpuDevice` instance incurs relatively low overhead

We have made `device` attribute cached because there was a performance issue.
I don't think the overhead for creating `CpuDevice` is negligible.

okapies: Hmm. In this sense, maybe we need another flag like `self._initialized` (it is the same as `self._is_cached_real_device`) or something to distinct `self._device == CpuDevice()` is a real value. How do you think?

okapies: Another idea is to change the spec of `device` as returning `None` when `Variable` is uninitialized.

niboshi: Yes, we need some change as you described above, and that should be in another PR.
That's why I proposed a TODO.

okapies: I see. Fixed and added some tests for unpickling uninitialized `Parameter`.

tests/chainer_tests/test_variable.py
```diff
@@ -1420,6 +1420,84 @@ def test_bool_inconvertible(self):
...
+            should_fail = (
+                device_x != device_y or requires_grad_x or requires_grad_y)
+        elif (isinstance(device_x, backend.Intel64Device)
+              or isinstance(device_y, backend.Intel64Device)):
+            # Check the array type because Intel64Device.send_array() may
+            # return numpy.ndarray when its shape doesn't match
+            if isinstance(x, intel64.mdarray):
+                should_fail = not isinstance(y, (np.ndarray, intel64.mdarray))
+            else:
+                should_fail = type(x) != type(y)
```
niboshi: `(type(x), type(y)) == (numpy.ndarray, intel64.mdarray)`
Shouldn't this case succeed?

tests/chainer_tests/test_variable.py
```diff
@@ -1854,6 +1932,43 @@ def test_copy_to_initialize(self):
         x.initialize((3, 2))
         assert x.data is y.data
 
+    def test_set_array_numpy(self):
```
niboshi: Please separate those tests and use `inject_backend_tests` just like `TestVariableSetArray`.

niboshi:
Could you separate unrelated changes?
Every time non-trivial change is made in this PR, I have to check every single line.

okapies:
@niboshi I'm very sorry to late response. Could you check it?

PR #6091, opened 2019-01-25
Fix seq2seq example

mitmul:
This PR is quick fix for a bug in the seq2seq example. The `__call__` method of a trainer extension defined in the seq2seq example is mistakenly renamed to `forward` by this commit for LinkHook: https://github.com/chainer/chainer/pull/4912/files#diff-b41c7c0a2fadcae4e15f628ccfe20936L139
I appreciate that @niboshi and @asi1024 found this bug. Thanks both, sincerely.

I think the example tests have some room for improvement to be able to check this kind of bugs.

PR #6095, opened 2019-01-26
Raise `NotImplementedError` if `Extension.__call__` is not overridden

niboshi:
Detects bugs like the one fixed in #6091.

### DIFF
--- a/chainer/training/extension.py
+++ b/chainer/training/extension.py
@@ -61,7 +61,8 @@ class Extension(object):
             trainer (Trainer): Trainer object that calls this operator.
 
         """
-        pass
+        raise NotImplementedError(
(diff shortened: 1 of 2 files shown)


Crissman:
Attempted to trigger Error using the original seq2seq.py file, by replacing `__call__` in CalculateBleu class with `forward`, but Error was not raised. When correctly labelled as `__call__`, code was executed, but apparently `__call__` method of Extension not called if renamed `forward`.

Was this PR intended to give an error in those circumstances?

Crissman:
Tested with original seq2seq.py and confirmed error is raised.

Crissman:
Jenkins, please test this.

Crissman:
Jenkins, test this please.

PR #6096, opened 2019-01-26
Improve static typing to avoid `type: ignore`

niboshi:
Fixed some typing to avoid `type: ignore`.
Also added explanations to uses of them.

### DIFF
--- a/chainer/variable.py
+++ b/chainer/variable.py
@@ -511,6 +511,8 @@ class Variable(object):
     # instance.
     _grad = None
 
+    _data = None  # type: tp.List[tp.Optional[types.NdArray]]
+
     def __init__(self, data=None, **kwargs):
         # type: (tp.Optional[types.NdArray], **tp.Any) -> None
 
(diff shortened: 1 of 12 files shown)


chainer/variable.py
```diff
@@ -843,16 +846,18 @@ def array(self, d):
 
         if self.xp is chainerx:
             d_old = self._data[0]
+            assert d_old is None or isinstance(d_old, chainerx.ndarray)
             if (d_old is not None
-                    and (d_old.is_backprop_required()  # type: ignore
+                    and (d_old.is_backprop_required()
                          or d.is_backprop_required())):
```
okapies: I'm very sorry that the signature of this method should be `(types.NdArray) -> None`... I fixed it in #6016.

okapies:
@niboshi It seems it is outdated due to the policy change for type hinting. Is going to close it?

niboshi:
Yes, I'll close.

PR #6100, opened 2019-01-27
Reduce memory usage in `MultiprocessParallelUpdater`

okuta:
This PR reduces memory usage.


niboshi:
How does this fix reduce memory usage?

niboshi:
LGTM

PR #5888, opened 2018-12-20
[WIP] Add Windows matrix in Travis CI

niboshi:
Fixes #5887

PR #6103, opened 2019-01-28
Enable Python chainerx tests in Travis Windows matrix 

### LINKED ISSUE #6104: Enable ChainerX test in Travis Windows CI
Related: #5887

PR: #6103

niboshi:
Fixes #6104.
Depends on #5888.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,7 +62,6 @@ matrix:
       python: "3.7"
       env:
       - CHAINER_TRAVIS_TEST="chainer"
-      - SKIP_CHAINERX=1
       - SKIP_CHAINERMN=1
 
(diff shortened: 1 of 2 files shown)


niboshi:
@durswd Could you take a look at the Travis error?
We don't have an environment to fix it by ourselves.
Please branch from this PR and submit another PR if the fix is done.

hvy:
It seems like the LLVM toolchain is not made available. This might not be a trivial fix with the current Travis. An option could be to to modify the source code such that it can be build without the LLVM extension on Windows. We'd need an Windows environment to actually do that 🤔

niboshi:
We decided we do not officially support Windows at the moment.

Issue #5621, opened 2018-11-04
Out of memory error with Chainer-5.0-Cupy-5.0-Python2.7


surajkota:
Hi,

I am trying to run synthetic data training with Chainer on a 8 GPU machine. The [scripts](https://github.com/surajkota/deep-learning-benchmark-mirror/blob/master/chainer_examples/imagenet/train_imagenet_data_parallel.py) I am running are modification of [imagenet example scripts from chainers official repository](https://github.com/chainer/chainer/tree/v5/examples/imagenet) to run synthetic data.

For, python2.7
Training processes errors out with **out of memory error for more than 16 batch size** and below is the stack trace

Python3.6
I am able to go upto **128** batch size with same setup

Same scripts run fine for Chainer==4.5 & CuPy==4.5 for both python2.7 and python3.6 upto 128 batch size
Note that running mnist passes but resnet fails for batch size more than 16 within first few epochs. 

**Stack trace:**

```
Epoch: [3] 	 Batch_speed: 3158.273 samples/sec
Epoch: [3] 	 Speed: 3099.142 samples/sec 	 Time cost=0.323
Exception in main training loop: out of memory to allocate 205520896 bytes (total 16068929536 bytes)
Traceback (most recent call last):
  File "/home/ubuntu/anaconda3/envs/chainer_p271/lib/python2.7/site-packages/chainer/training/trainer.py", line 315, in run
    update()
  File "/home/ubuntu/anaconda3/envs/chainer_p271/lib/python2.7/site-packages/chainer/training/updaters/standard_updater.py", line 165, in update
    self.update_core()
  File "/home/ubuntu/anaconda3/envs/chainer_p271/lib/python2.7/site-packages/chainer/training/updaters/multiprocess_parallel_updater.py", line 237, in update_core
    loss = _calc_loss(self._master, batch)
  File "/home/ubuntu/anaconda3/envs/chainer_p271/lib/python2.7/site-packages/chainer/training/updaters/multiprocess_parallel_updater.py", line 271, in _calc_loss
    return model(*in_arrays)
  File "/home/ubuntu/anaconda3/envs/chainer_p271/lib/python2.7/site-packages/chainer/link.py", line 242, in __call__
    out = forward(*args, **kwargs)
  File "/home/ubuntu/chainer/examples/imagenet/resnet50.py", line 101, in forward
    h = self.res2(h)
[...]

PR #6105, opened 2019-01-28
Fix memory leak during backprop in Python 2

toslunar:
This PR wraps a use of `collections.OrderedDict` in order to address a memory issue with Python 2.

Split from #5274.

### DIFF
--- a/chainer/utils/_collections.py
+++ b/chainer/utils/_collections.py
@@ -5,11 +5,17 @@ import six
 
 
 if six.PY3:
-    RefCountFriendlyOrderedDict = collections.OrderedDict
+    OrderedDict = collections.OrderedDict
 else:
     # Reference counting cannot free keys in old `collections.OrderedDict`,
     # where a doubly linked list is used to maintain the order.
-    class RefCountFriendlyOrderedDict(object):
+    class OrderedDict(object):
+        """Dictionary that remembers insertion order
+
+        This class wraps `collections.OrderedDict` to free keys by reference
+        counting.
+        """
+
         def __init__(self):
             self.keys = set()
             self.dict = collections.OrderedDict()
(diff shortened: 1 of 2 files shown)


chainer/utils/_collections.py
```diff
@@ -0,0 +1,31 @@
...
+
+import six
+
+
+if six.PY3:
+    RefCountFriendlyOrderedDict = collections.OrderedDict
+else:
+    # Reference counting cannot free keys in old `collections.OrderedDict`,
+    # where a doubly linked list is used to maintain the order.
+    class RefCountFriendlyOrderedDict(object):
```
niboshi: How about simply naming this as `OrderedDict`? (because semantically they're equivalent.)

okuta:
I tested this PR with #5621. Memory leaking is stopped.

chainer/utils/_collections.py
toslunar: Thanks.  Renamed it.

chainer/variable.py
```diff
@@ -1500,6 +1501,8 @@ def __bool__(self):
 
 
 def _backprop_to_all(outputs, retain_grad, loss_scale):
+    OrderedDict3 = chainer.utils._collections.OrderedDict
```
niboshi: This one should be renamed to `OrderedDict`, too?

toslunar:
It seems #4772 introduced the regression.

chainer/variable.py
toslunar: I left this name to make it clear that there is overhead in Python 2.  In the function, both `collections.OrderedDict` and `chainer.utils._collections.OrderedDict` are used.

toslunar: I'll fix this line, too, because the point I wrote seems clear from the coding style disallows `from collections import OrderedDict`.

niboshi: Maybe a single comment on this line might be helpful?

okuta:
LGTM!

Issue #6120, opened 2019-01-28
Handle None as FunctionNode input with ChainerX


niboshi:
Required in seq2seq example.

PR #6121, opened 2019-01-28
Fix `get_retained_{in/out}puts` to return `None` for `None` inputs/outputs

niboshi:
This PR originally intended to support `None` as `FunctionNode` inputs in ChainerX (issue: #6120) , but during that I found that existing semantics should also change.

Previously, `get_retained_{in,out}puts` returns `Variable(None)` if the input/output is `None`.
In this PR it has changed to return `None`.

(Depends on #6118)

### DIFF
--- a/tests/chainer_tests/test_backprop.py
+++ b/tests/chainer_tests/test_backprop.py
@@ -458,7 +458,7 @@ class TestFunctionOutputNone(unittest.TestCase):
                 retained_outputs = self.get_retained_outputs()
                 assert isinstance(retained_outputs, tuple)
                 y2, y3 = retained_outputs
-                assert y3.array is None
+                assert y3 is None
                 assert isinstance(y2.array, backward_xp.ndarray)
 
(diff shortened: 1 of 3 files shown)


tests/chainer_tests/test_backprop.py
```diff
@@ -228,6 +228,138 @@ def test_backprop(self, backend_config):
...
+@testing.backend.inject_backend_tests(
+    None,
+    [
+        # CPU
+        {},
+        # CUDA
+        {'use_cuda': True, 'cuda_device': 0},
+        {'use_cuda': True, 'cuda_device': 1},
+        # ChainerX
+        # TODO(niboshi): Fix ChainerX for None inputs
```
niboshi: ChainerX test is disabled for now. We need more fix.

asi1024:
Jenkins, test this please.

hvy:
jenkins, test this please.

hvy:
LGTM!

PR #6124, opened 2019-01-28
Move Jenkins mypy check to misc matrix

niboshi:


asi1024:
Jenkins, test this please.

PR #6126, opened 2019-01-28
Do not compare `xp` with `numpy` for cupy code path

niboshi:
`xp` can be `chainerx`, too.
Testing `xp is numpy` for cupy should be wrong.


asi1024:
Jenkins, test this please.

hvy:
LGTM!

PR #6127, opened 2019-01-28
Enable `use_ideep` and `lazy_grad_sum` by default

niboshi:
They were introduced in v4.

### DIFF
--- a/chainer/__init__.py
+++ b/chainer/__init__.py
@@ -192,9 +192,9 @@ global_config.use_cudnn = os.environ.get('CHAINER_USE_CUDNN', 'auto')
 global_config.use_cudnn_tensor_core = 'auto'
 global_config.autotune = False
 global_config.schedule_func = None
-global_config.use_ideep = os.environ.get('CHAINER_USE_IDEEP', 'never')
+global_config.use_ideep = os.environ.get('CHAINER_USE_IDEEP', 'auto')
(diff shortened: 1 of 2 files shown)


Crissman:
Need to add the documentation change for ideep default to `auto` to `configuration.rst`.

Similarly, documentation needs to be updated for the default value of ideep in `docs/source/tips.rst`.

Crissman:
Discussed among team, and agreed to proceed with making `use_ideep` on by default.

niboshi:
Conclusion:
* Make `config.use_ideep = 'auto'` by default
  * Currently `'auto'` is equivalent to `'always'`.
* Make `config.lazy_grad_sum = 'auto'` by default
  * If it is `'auto'`, the feature is enabled in CPU mode only and not in GPU mode, because in GPU mode memory consumption is more likely to be a problem than CPU.
  * Currently it takes `bool` values. They're mapped to `never` and `always` respectively.

Crissman:
Jenkins, test this please.

Crissman:
Jenkins issues unrelated. Merging!

niboshi:
@Crissman Oh, this PR at the current state does not reflect the discussion: https://github.com/chainer/chainer/pull/6127#issuecomment-465888560

Also, please refrain from self-merging!

kmaehashi:
This PR has been reverted in #6794

PR #6138, opened 2019-01-29
CuPy cannot be enabled when cuDNN is unavailable

### LINKED ISSUE #6134: cuDNN required in v6?
Chainer code that used to work fine with 5.2 does not work any more with 6.0.0b2, presumably because I don't have cuDNN installed.

* Conditions
  - Chainer version: 6.0.0b2
  - CuPy version: 6.0.0b2
  - OS/Platform: Ubuntu 16.04
  - CUDA/cuDNN version: CUDA 8.0/no cuDNN
* Code to reproduce: `python3 -c "from chainer.backends import cuda; cuda.check_cuda_available()"`
* Error messages, stack traces, or logs:
  ```
    File "<string>", line 1, in <module>
    File "/usr/local/lib/python3.5/dist-packages/chainer/backends/cuda.py", line 135, in 
  check_cuda_available
      raise RuntimeError(msg)
  RuntimeError: CUDA environment is not correctly set up
  (see https://github.com/chainer/chainer#installation).module 'cupy.cuda' has no attribute 'cudnn'
  ```

@hvy pointed out that this change https://github.com/chainer/chainer/commit/91878a45fe40fb2ea70ff9bbf3e9a9ef4ee6df56#diff-b74802247d24e96cf7d5d8742480530dR69 may be the cause.

kmaehashi:
Fixes #6134.

### DIFF
--- a/chainer/backends/cuda.py
+++ b/chainer/backends/cuda.py
@@ -65,7 +65,6 @@ try:
     from cupy.cuda import Event  # type: ignore # NOQA
     from cupy.cuda import Stream  # type: ignore # NOQA
 
-    libcudnn = cuda.cudnn  # type: tp.Any # NOQA
     available = True
 except Exception as e:
     _resolution_error = e
@@ -108,11 +107,17 @@ except Exception as e:
     class Stream(object):  # type: ignore # for type testing
         pass
 
-    # for `xp is cuda.cupy` and `cuda.libcudnn` to always work
+    # for `xp is chainer.backends.cuda.cupy` to always work
     cupy = object()
-    libcudnn = object()
+
 
 if available:
+    try:
+        libcudnn = cupy.cuda.cudnn  # type: tp.Any # NOQA
+    except Exception as e:
+        # for `chainer.backends.cuda.libcudnn` to always work
+        libcudnn = object()
+
     _cudnn_disabled_by_user = int(os.environ.get('CHAINER_CUDNN', '1')) == 0
     try:
         import cupy.cudnn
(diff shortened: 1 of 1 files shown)


chainer/backends/cuda.py
```diff
@@ -108,11 +107,17 @@ class Event(object):  # type: ignore # for type testing
...
-    # for `xp is cuda.cupy` and `cuda.libcudnn` to always work
+    # for `xp is chainer.backends.cuda.cupy` to always work
     cupy = object()
-    libcudnn = object()
+
 
 if available:
+    try:
+        libcudnn = cupy.cuda.cudnn  # type: tp.Any # NOQA
+    except Exception as e:
```
niboshi: flake8 complains `e` is never used.

chainer/backends/cuda.py
```diff
@@ -108,11 +107,17 @@ class Event(object):  # type: ignore # for type testing
...
 
-    # for `xp is cuda.cupy` and `cuda.libcudnn` to always work
+    # for `xp is chainer.backends.cuda.cupy` to always work
     cupy = object()
-    libcudnn = object()
+
 
 if available:
+    try:
+        libcudnn = cupy.cuda.cudnn  # type: tp.Any # NOQA
```
niboshi: How about doing this only if `cudnn_enabled is True`?
(In that case I think this assignment should succeed?)

niboshi: Ok, it could raise an error even if `cudnn = cupy.cudnn` succeeds
(because `cupy.cuda.cudnn` link to external libraries).

niboshi:
Jenkins, test this please

niboshi:
LGTM

PR #5619, opened 2018-11-02
Add type check for `r` arg of `F.rrelu`

toslunar:
Fix its test (#5618) first.

PR #6139, opened 2019-01-29
Fix double-backprop of `F.rrelu`

toslunar:
#5619 broke `_RReLUGrad.backward`.

```
>>> x = chainer.Variable(np.random.randn(2, 3).astype('f'))
>>> y = F.rrelu(x)
>>> y.grad = np.random.randn(2, 3).astype('f')
>>> y.backward(enable_double_backprop=True)
>>> x.grad_var.grad = np.random.randn(2, 3).astype('f')
>>> x.grad_var.backward()
  File "/Users/tos/GitHub/chainer/chainer/variable.py", line 1386, in backward
  File "/Users/tos/GitHub/chainer/chainer/variable.py", line 1577, in _backprop_to_all
    func, target_input_indexes, out_grad, in_grad)
  File "/Users/tos/GitHub/chainer/chainer/_backprop_utils.py", line 116, in backprop_step
    target_input_indexes, grad_outputs)
  File "/Users/tos/GitHub/chainer/chainer/functions/activation/rrelu.py", line 89, in backward
    return _RReLUGrad(self.x, self.y, self.r).apply(grad_outputs)
AttributeError: '_RReLUGrad' object has no attribute 'x'
```

### DIFF
--- a/tests/chainer_tests/functions_tests/activation_tests/test_rrelu.py
+++ b/tests/chainer_tests/functions_tests/activation_tests/test_rrelu.py
@@ -30,7 +30,7 @@ class TestRReLU(testing.FunctionTestCase):
 
     dodge_nondifferentiable = True
 
-    def before_test(self, test_name):
+    def setUp(self):
         # Assumption l < u
         self.l = numpy.random.uniform(0, 1)
         self.u = numpy.random.uniform(0, 1)
--- a/tests/chainer_tests/functions_tests/activation_tests/test_rrelu.py
+++ b/tests/chainer_tests/functions_tests/activation_tests/test_rrelu.py
@@ -3,10 +3,8 @@ import unittest
 import numpy
 
 import chainer
-from chainer import backend
 from chainer import cuda
 from chainer import functions
-from chainer import gradient_check
 from chainer import testing
 from chainer.testing import attr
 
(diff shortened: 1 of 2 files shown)


niboshi:
Could you add some description (also make the title more descriptive)?

tests/chainer_tests/functions_tests/activation_tests/test_rrelu.py
```diff
@@ -3,82 +3,76 @@
...
+    + testing.product({
+        'use_cuda': [True],
+        'cuda_device': [0, 1],
+    })
+)
+class TestRReLU(testing.FunctionTestCase):
+
+    dodge_nondifferentiable = True
+
+    def before_test(self, test_name):
```
niboshi: I think this can be done in `setUp`, as usual.
(`before_test` was introduced for the case it would need access to `self.backend_config`, which are not available from `setUp`.)

tests/chainer_tests/functions_tests/activation_tests/test_rrelu.py
```diff
@@ -3,82 +3,76 @@
...
     'dtype': [numpy.float16, numpy.float32, numpy.float64],
 }))
-class TestRReLU(unittest.TestCase):
-
-    def setUp(self):
-        # Avoid unstability of numerical grad
-        self.x = numpy.random.uniform(-1, 1, self.shape).astype(self.dtype)
-        self.x[(-0.05 < self.x) & (self.x < 0.05)] = 0.5
-        self.gy = numpy.random.uniform(-1, 1, self.shape).astype(self.dtype)
+@testing.inject_backend_tests(
```
niboshi: I think it's better to put `@testing.inject_backend_tests` before `@testing.parameterize`, because parameters are more closely related to the test logic.

tests/chainer_tests/functions_tests/activation_tests/test_rrelu.py
```diff
@@ -3,82 +3,76 @@
...
         if self.dtype == numpy.float16:
             self.check_forward_options = {'atol': 1e-4, 'rtol': 1e-3}
             self.check_backward_options = {'atol': 5e-4, 'rtol': 5e-3}
+            self.check_double_backward_options = {'atol': 5e-4, 'rtol': 5e-3}
 
-    def check_forward(self, x_data):
-        x = chainer.Variable(x_data)
-        xp = backend.get_array_module(x)
-        with chainer.using_config('train', self.train):
-            y, r = functions.rrelu(x, l=self.l, u=self.u, return_r=True)
```
niboshi: It seems the test of `return_r` has been removed.

toslunar: - `return_r` is tested in `TestRReLUR`.
- There's no longer a test with `r=None, return_r=False`, (which are defaults!)  I'll add a test to confirm `F.rrelu` just runs with that options.

niboshi: I didn't notice it.
It still lacks a test for backward/double-backward for generated `r`s, but it's another issue.

niboshi:
Jenkins, test this please

niboshi:
LGTM

PR #6142, opened 2019-01-29
Reorganize documentation for easier access to tutorials and examples

### LINKED ISSUE #6056: Consider moving `Notebook Examples` documentation link to be easier for new users to find
In the current Chainer documentation, `Chainer at a Glance` is followed by `Installation` and then `Guides`. I think it would be good if we could provide some additional tutorials/examples for new users before jumping into the reference guide.

Actually, there already seem to be some nice examples in the hands-on which are currently linked to near the bottom of the main documentation page: `Community` -> `Notebook Examples`.
However, with the current organization, it is easy for new users not to notice them. I therefore suggest moving the `Notebook Examples link to somewhere near or immediately following the `Chainer at a Glance` section.

niboshi:
Fixes #6056 

sections

### DIFF
--- a/docs/source/guides/index.rst
+++ b/docs/source/guides/index.rst
@@ -1,5 +1,5 @@
-Guides
-======
+Concepts Walkthrough
+====================
 
 .. toctree::
    :maxdepth: 1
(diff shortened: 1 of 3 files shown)


Crissman:
LGTM. Documentation only, so not doing Jenkins. Merging!

PR #6145, opened 2019-01-30
[backport] Fix `FunctionNode.get_retained_outputs` to return `()` if no output is retained

hvy:
Backport of #6118


hvy:

niboshi:
We decided to not backport this PR:
documentation says those methods can return `None` and we guess only a few peope complain this behavior, if any.

Issue #6131, opened 2019-01-29
Python 3.4 to reach end-of-life


hvy:
Python 3.4. will reach end-of-life (EOL) by 2019-03-16 and will no longer be maintained.
https://devguide.python.org/#status-of-python-branches  

Chainer should thus accordingly update the requirements,  CI, docs, deprecations/warnings and maybe type hints (introduced in Python 3.5) (https://www.python.org/dev/peps/pep-0484/  .

beam2d:
Besides chainer-test fix, we only need documentation fix for now.

hvy:
Remaining tasks are
- [ ] Remove from https://github.com/chainer/chainer-test
- [ ] Check upgrade guide.

toslunar:
The message "Partially fixes #6131" automatically closed this, but it's partial, of course.
https://help.github.com/en/articles/closing-issues-using-keywords

hvy:
Thanks, I should stop writing that. Although late, I changed the description to "Part of ...".

toslunar:
The priority of the remaining task is not high.

toslunar:
https://github.com/chainer/chainer-test/pull/567 is on-going.

kmaehashi:
This is done in chainer-test.

PR #6149, opened 2019-01-30
Remove Travis Ubuntu Py34 job

hvy:
Related to Python 3.4 support drop https://github.com/chainer/chainer/issues/6131.

This PR simply removes the Ubuntu Py34 job from the Travis matrix. It is maybe worth discussing if we should bump the Python version to 3.5 of this job instead of removing it, to test Chainer without ChainerX.


niboshi:
Jenkins, test this please

niboshi:
LGTM

PR #6156, opened 2019-01-31
Check Array constructor for nullptr

### LINKED ISSUE #6115: Throw exception on chainerx::Array{nullptr}
An exception should be raised here, instead of using `gsl::not_null`.

https://github.com/chainer/chainer/blob/8a3849f5366f2fa11ff3f641d52e63b00e7af9b4/chainerx_cc/chainerx/array.h#L51

Currently passing `None` as an array leads to SIGABRT.
```py
>>> chainerx.max(None)
terminate called without an active exception
Aborted (core dumped)
```

hvy:
Fixes #6115.

### DIFF
--- a/chainerx_cc/chainerx/array.h
+++ b/chainerx_cc/chainerx/array.h
@@ -48,7 +48,12 @@ class Array {
 public:
     Array() = default;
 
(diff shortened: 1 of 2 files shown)


niboshi:
LGTM

niboshi:
Jenkins, test this please

PR #6164, opened 2019-02-01
Fix unit test collection timeout

hvy:
Since `pytest==4.2.0`, collection of tests has become slow (it gets slower and slower over time) and times out certain CI builds. Pinning `pytest` to the previous versions gets around the problem but I'm not sure it's the right thing to do.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,7 +38,7 @@ requirements = {
         'pycodestyle>=2.4,<2.5',
     ],
     'test': [
-        'pytest',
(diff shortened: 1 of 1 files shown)


niboshi:
Jenkins, test this please

niboshi:
LGTM!

hvy:
@niboshi Could you create a backport for this PR? It seems like a master-v5-independent matter.

niboshi:
Ah, I think this fix is too strict to backport to v5.
How about backporting #6502 instead?

hvy:
That looks better. Thanks for the PR.

PR #6165, opened 2019-02-01
let backticks embrace outputs of chainer.print_runtime_info

crcrpar:
This PR places three backticks at the beginning and end of outputs of `chainer.print_runtime_info` because some people (at least I) always do this.


crcrpar:
Yes, I assume environments that support backticks formatting.

OK, then I'll close this.

PR #6169, opened 2019-02-02
Check input size consistency in RNN and LSTM when using cuDNN

### LINKED ISSUE #5436: NStepLSTM silently fails with invalid in_size when cuDNN is enabled
## Problem description

When cuDNN is enabled NStepLSTM (and its variants), does not raise warning or error when `in_size` is incosistent with the actual input. It will run as if nothing is wrong, and then fails with a CUDA error. A random CUDA error seems to be raised (including `CUDNN_STATUS_EXECUTION_FAILED` and `CUDA_ERROR_ILLEGAL_ADDRESS`). 

## How to reproduce

1. Download and setup `text_classification` example

https://github.com/chainer/chainer/tree/v5.0.0rc1/examples/text_classification

2. Modify `in_size` of NStepLSTM to incosistent value.

```diff
-  self.encoder = L.NStepLSTM(n_layers, n_units, n_units, dropout)
+  self.encoder = L.NStepLSTM(n_layers, 50, n_units, dropout)
```

https://github.com/chainer/chainer/blob/v5.0.0rc1/examples/text_classification/nets.py#L129

3. Run `train_text_classifier.py`

```bash
python train_text_classifier.py --gpu 0 --model rnn --batchsize 256 --epoch 100
```

I set larger `batchsize` so that failure comes earlier.
It usually reproduces within first few epochs.

4. Confirmed that it does not reproduce without cuDNN

```bash
CHAINER_CUDNN=0 python train_text_classifier.py --gpu 0 --model rnn --batchsize 256 --epoch 100
```

It should (properly) raise error that input is incosistent.

I confirmed this behavior in Chainer v5.0.0rc1 and cupy v5.0.0rc1 with cuDNN 7.1.3.16-1+cuda9.1 and CUDA 9.1 on Ubuntu 16.04.

## Suggestion

I suggest we make NStepLSTM to do input size check or make documentation that recommend user to turn off cuDNN if weird CUDA error is raised, or possibly both.

It was super hard to figure out what is wrong and seriously need solving. (cuda-memcheck seems to point at random function as well)

takagi:
Close #5436.

This PR makes RNN and LSTM functions check input size consistency with their input values and weights when using cuDNN.

- `F.n_step_rnn`
- `F.n_step_birnn`
- `F.n_step_lstm`
- `F.n_step_bilstm`

### DIFF
--- a/chainer/functions/connection/n_step_lstm.py
+++ b/chainer/functions/connection/n_step_lstm.py
@@ -411,6 +411,13 @@ def n_step_lstm_base(
         lengths = [len(x) for x in xs]
         xs = chainer.functions.concat(xs, axis=0)
 
+        # Check input size consistency with xs and ws here.
+        x_in = xs[0].shape[0]
+        w_in = ws[0][0].shape[1]
+        if x_in != w_in:
+            raise ValueError('Inconsistent input size in input values and '
+                             'weight parameters')
+
         w = n_step_rnn.cudnn_rnn_weight_concat(
             n_layers, states, use_bi_direction, 'lstm', ws, bs)
 
--- a/chainer/functions/connection/n_step_lstm.py
+++ b/chainer/functions/connection/n_step_lstm.py
@@ -403,6 +403,13 @@ def n_step_lstm_base(
             'Use chainer.using_config')
         argument.assert_kwargs_empty(kwargs)
 
+    # Check input size consistency with xs and ws here.
+    x_in = xs[0].shape[1]
+    w_in = ws[0][0].shape[1]
+    if x_in != w_in:
+        raise ValueError('Inconsistent input size in input values and weight '
+                         'parameters')
+
     xp = backend.get_array_module(hx, hx.data)
 
     if xp is cuda.cupy and chainer.should_use_cudnn('>=auto', 5000):
@@ -410,13 +417,6 @@ def n_step_lstm_base(
         states.set_dropout_ratio(dropout_ratio)
         lengths = [len(x) for x in xs]
         xs = chainer.functions.concat(xs, axis=0)
-
-        # Check input size consistency with xs and ws here.
-        x_in = xs[0].shape[0]
-        w_in = ws[0][0].shape[1]
-        if x_in != w_in:
-            raise ValueError('Inconsistent input size in input values and '
-                             'weight parameters')
 
         w = n_step_rnn.cudnn_rnn_weight_concat(
             n_layers, states, use_bi_direction, 'lstm', ws, bs)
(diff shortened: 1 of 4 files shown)


hvy:
Should we do this check for non-cuDNN paths as well (i.e. CPU or CUDA without cuDNN) or are the size constraints different in those cases?

takagi:
In such case, `F.linear` used in these functions causes an error in its `check_type_forward` method, so we need no additional check.

hvy:
Ah I see, thanks for the explanation. I think though that the error message that you introduce is a lot friendlier to the to user in these cases as well, since the linear functions is an implementation detail. I also think that the code is more readable with a unified shape validation before entering the device branch logic as well.

takagi:
I got it, will fix the code!

hvy:
jenkins, test this please.

chainer/functions/connection/n_step_lstm.py
```diff
@@ -403,6 +403,13 @@ def n_step_lstm_base(
             'Use chainer.using_config')
         argument.assert_kwargs_empty(kwargs)
 
+    # Check input size consistency with xs and ws here.
+    x_in = xs[0].shape[1]
+    w_in = ws[0][0].shape[1]
+    if x_in != w_in:
+        raise ValueError('Inconsistent input size in input values and weight '
+                         'parameters')
```
hvy: How about including the actual values to make the message a bit more friendly?

chainer/functions/connection/n_step_rnn.py
```diff
@@ -633,6 +633,13 @@ def n_step_rnn_base(n_layers, dropout_ratio, hx, ws, bs, xs,
         raise ValueError('Invalid activation: "%s". Please select from [%s]'
                          % (activation, candidate))
 
+    # Check input size consistency with xs and ws.
+    x_in = xs[0].shape[1]
+    w_in = ws[0][0].shape[1]
+    if x_in != w_in:
+        raise ValueError('Inconsistent input size in input values and weight '
+                         'parameters')
```
hvy: Ditto.

chainer/functions/connection/n_step_lstm.py
takagi: Fixed!

chainer/functions/connection/n_step_rnn.py
takagi: Fixed!

hvy:
jenkins, test this please.

chainer/functions/connection/n_step_lstm.py
hvy: Thanks for the quick fix!

takagi:
I will fix the testing tolerance of `F.triplet` double backward as another issue.

hvy:
LGTM!

PR #6171, opened 2019-02-04
Fix `Optimizer` to convert state arrays back to ChainerX

asi1024:
Currently, `update_core_chainerx` in `Optimizer` class assumes that state arrays are updated in-place, but `optimizers.SMORM3` replaces the state arrays instead.


niboshi:
Jenkins, test this please

niboshi:
LGTM

PR #6176, opened 2019-02-05
make chainerx compiled on Windows

durswd:
I fixed compiling errors.
I made chainerx compiled on Windows.

### DIFF
--- a/chainerx_cc/chainerx/context.cc
+++ b/chainerx_cc/chainerx/context.cc
@@ -1,6 +1,9 @@
 #include "chainerx/context.h"
 
-#ifndef _WIN32
+#ifdef _WIN32
+#define RTLD_LOCAL 0
+#define RTLD_NOW 0x00002
+#else  // _WIN32
 // Windows doesn't support it currently
 #include <dlfcn.h>
 #endif  // _WIN32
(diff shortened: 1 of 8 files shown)


chainerx_cc/chainerx/dynamic_lib.h
```diff
@@ -9,4 +9,6 @@ void* DlOpen(const std::string& filename, int flags);
 
 void DlClose(void* handle);
 
+void* DlSym(void* handle, const char* name);
```
hvy: Can you take the `name` by `const std::string&` for consistency?

chainerx_cc/chainerx/context.cc
```diff
@@ -1,6 +1,9 @@
 #include "chainerx/context.h"
 
-#ifndef _WIN32
+#ifdef _WIN32
+#define RTLD_LOCAL 0
+#define RTLD_NOW 0x00002
```
hvy: Could you write a comment above these two lines why they're required (and that they're expected to be temporary)?

hvy:
Thanks for the PR and the fixes for `dlsys`. I added a few minor comments but it LGTM otherwise.

chainerx_cc/chainerx/context.cc
niboshi: Rather than defining them here, I think it's better to enclose their usage as an implementation detail of platforms.
In other words, how about removing `flags` argument from `chainerx::DlOpen` and `chainerx::platform::DlOpen`?

chainerx_cc/chainerx/context.cc
```diff
@@ -1,6 +1,10 @@
 #include "chainerx/context.h"
 
-#ifndef _WIN32
+#ifdef _WIN32
+// RTLD_LOCAL and RTLD_NOW are not defined on Windows
+#define RTLD_LOCAL 0
+#define RTLD_NOW 0x00002
+#else  // _WIN32
 // Windows doesn't support it currently
 #include <dlfcn.h>
```
niboshi: It seems we can remove this include.

hvy:
jenkins, test this please.

niboshi:
@durswd  Thanks!

PR #6187, opened 2019-02-06
Introduce LinkTestCase to simplify Link tests

### LINKED ISSUE #6178: Unified link test tools
A base class for link tests should be introduced to simplify the writing link tests and to increase the coverage. Similar to https://github.com/chainer/chainer/pull/3499 for functions.

Some points that must be respected are that.
- There are no constraints to `Link.__call___` inputs and outputs.
- Parameter initialization is link-dependent.
- Persistent value initialization is link-dependent.
- Persistent values can be of any type that's serializable.

hvy:
Fixes #6178.

Introduces 
- A base class for `chainer.Link` implementation tests, `chainer.testing.LinkTestCase`.
- `chainer.testing.test` to store common functions for `chainer.testing.{Function,Link}TestCase`.
- A "check_backward" function that allows passing parameters as ndarrays (instead of previous `chainer.Parameter`). This is needed since `chainer.Parameter` with ChainerX ndarrays cannot be updated in-place (`.array`).


chainer/testing/link.py
```diff
@@ -0,0 +1,80 @@
+import copy
+import unittest
+
+
+import chainer
+from chainer.testing import array as array_module
+from chainer.testing import function as function_module
+
+
+class LinkTestCase(unittest.TestCase):
```
hvy: Please ignore this class for now.

hvy:
Sample with BN, how it's link test would be written with a
- single class implementation https://github.com/chainer/chainer/pull/6187/files#diff-efa0a60dd7b75dd0da103086cfe48722
- implemented using separate classes https://github.com/chainer/chainer/pull/6187/files#diff-a10b666a884c3da8953b4a3ab3ecdfd4

tests/chainer_tests/links_tests/normalization_tests/test_batch_normalization_separate.py
```diff
@@ -0,0 +1,174 @@
...
+
+
+class BatchNormalizationLinkForward(LinkForward):
+    """Common logic for forward/backward and initialization tests."""
+
+    # Passed to forward_expect as numpy.ndarrays.
+    params = ['gamma', 'beta']
+
+    # Passed to forward_expect as is.
+    attrs = ['eps']
```
hvy: This `attrs` can probably be removed. `eps` can simply be saved by this object.

hvy:
Discussed offline that we should proceed with
- the unified implementation since it is simpler and sufficient in terms of functionality.
- having a single method to return all initializers arguments.

chainer/gradient_check.py
```diff
@@ -622,30 +631,47 @@ def perturb(data, direction):
...
+                    # Update the given Parameter in-place since the object is
+                    # held by the caller.
+                    param.array = data
 
             # Clear gradients to support func that calls backward inside of
             # itself.
             self._clear_grads(g_x_vars)
-            self._clear_grads(params)
+            if not self.params_to_func:
+                self._clear_grads(params)
```
hvy: Are these clears needed? 🤔

chainer/gradient_check.py
```diff
@@ -412,6 +412,7 @@ def __init__(
         self.eps = eps
         self.dtype = dtype
         self.detect_nondifferentiable = detect_nondifferentiable
+        self.params_to_func = params_to_func
```
niboshi: Move this line before the comment at line 411.

chainer/gradient_check.py
```diff
@@ -370,7 +370,7 @@ class _CheckBackward(object):
 
     def __init__(
             self, func, x_data, y_grad, params, eps, atol, rtol, no_grads,
-            dtype, detect_nondifferentiable):
+            dtype, detect_nondifferentiable, params_to_func=False):
```
niboshi: Remove the default value, as `_CheckBackward` is an internal implementation.

chainer/gradient_check.py
```diff
@@ -622,30 +631,47 @@ def perturb(data, direction):
...
                 else:
                     data = perturb(casted_data[j], directions[j])
                     g_x_vars.append(variable.Variable(data))
                     j += 1
+
             # Parameters
-            for i in range(len(params)):
-                params[i].data = perturb(
-                    casted_data[j + i], directions[j + i])
+            for i, param in enumerate(params):
```
niboshi: I think `for i in range(len(params)):` is better.
`param` is not always used.

chainer/gradient_check.py
```diff
@@ -812,6 +838,17 @@ def func(xs):
     ).run()
 
 
+def _check_backward_with_params(
```
niboshi: This function needs some comment: what's the difference, why it's needed, who's using it, etc.

chainer/testing/link.py
```diff
@@ -0,0 +1,549 @@
...
+    correspond to the actual initializer-like object used by the link.
+    In many cases, they are the same, except when passing ``None`` to the
+    constructor.
+    """
+
+    def __init__(self, first, second):
+        initializers._check_is_initializer_like(first)
+        initializers._check_is_initializer_like(second)
+
+        self.first = first
```
niboshi: I think `first` and `second` are too general.
Isn't there any better naming?

chainer/testing/link.py
```diff
@@ -0,0 +1,549 @@
...
+        returned by either ``generate_forward_backward_initializers`` or
+        ``generate_initializers`` depending on the test being run.
+
+    .. rubric:: Optional methods
+
+    Additionally, the concrete class can override the following methods. Some
+    must be overridden depending on the skip flags  ``skip_forward_test``,
+    ``skip_backward_test`` and ``skip_initializers_test``.
+
+    ``generate_forward_backward_initializers(self)``
```
niboshi: I feel `generate_forward_backward_initializers` is too long and not easily understood to test implementors.

tests/chainer_tests/links_tests/normalization_tests/test_batch_normalization.py
```diff
@@ -36,118 +39,175 @@ def _batch_normalization(expander, gamma, beta, x, mean, var, eps, test):
...
+
+        self.check_forward_options = {'atol': 1e-4, 'rtol': 1e-3}
+        self.check_backward_options = {'atol': 1e-4, 'rtol': 1e-3}
         if self.dtype == numpy.float16:
-            self.check_forward_optionss = {'atol': 1e-3, 'rtol': 1e-2}
-            self.check_backward_optionss = {'atol': 5e-1, 'rtol': 1e-1}
+            self.check_forward_options = {'atol': 1e-2, 'rtol': 1e-1}
+            self.check_backward_options = {'atol': 5e-1, 'rtol': 1e-1}
+
+    def before_test(self, test_name):
```
niboshi: I think this can be done in `setUp`.

tests/chainer_tests/links_tests/normalization_tests/test_batch_normalization.py
```diff
@@ -36,118 +39,175 @@ def _batch_normalization(expander, gamma, beta, x, mean, var, eps, test):
...
+    # ChainerX tests
+    + [
+        {'use_chainerx': True, 'chainerx_device': 'native:0'},
+        {'use_chainerx': True, 'chainerx_device': 'cuda:1'},
+        {'use_chainerx': True, 'chainerx_device': 'cuda:1'},
+    ])
+@_parameterize
+class BatchNormalizationTest(BatchNormalizationTestBase, testing.LinkTestCase):
+
+    contiguous = 'C'
```
niboshi: Is this needed?

tests/chainer_tests/links_tests/normalization_tests/test_batch_normalization.py
```diff
@@ -36,118 +39,175 @@ def _batch_normalization(expander, gamma, beta, x, mean, var, eps, test):
...
+        'use_cudnn': ['never', 'always'],
+        'cuda_device': [0],
+    }))
+@_parameterize
+class BatchNormalizationMultiGpuTest(
+        BatchNormalizationTestBase, testing.LinkTestCase):
+
+    skip_backward_test = True
+    skip_initializers_test = True
+    contiguous = 'C'
```
niboshi: Ditto.

tests/chainer_tests/links_tests/normalization_tests/test_batch_normalization.py
```diff
@@ -36,118 +39,175 @@ def _batch_normalization(expander, gamma, beta, x, mean, var, eps, test):
...
+@attr.multi_gpu(2)
+@testing.inject_backend_tests(
+    None,
+    testing.product({
+        'use_cuda': [True],
+        'use_cudnn': ['never', 'always'],
+        'cuda_device': [0],
+    }))
+@_parameterize
+class BatchNormalizationMultiGpuTest(
```
niboshi: Is this test required to manually written?
It seems it's something all the links should support.

tests/chainer_tests/links_tests/normalization_tests/test_batch_normalization.py
hvy: `before_test` is used for resampling.

chainer/gradient_check.py
hvy: Fixed.

chainer/gradient_check.py
hvy: Fixed.

chainer/gradient_check.py
hvy: I agree, thanks. Fixed.

chainer/gradient_check.py
hvy: Fixed.

hvy: Fixed.

tests/chainer_tests/links_tests/normalization_tests/test_batch_normalization.py
hvy: Although that's true, I am not sure why this particular link had this test in the first place, do you know why?

tests/chainer_tests/links_tests/normalization_tests/test_batch_normalization.py
hvy: It is, because of this bug https://github.com/cupy/cupy/issues/2061. I'll add a TODO comment.

tests/chainer_tests/links_tests/normalization_tests/test_batch_normalization.py
hvy: Thanks, it was there during debugging. Removed.

chainer/testing/link.py
hvy: What do you think about `before_init`, `after_init`?

chainer/testing/link.py
hvy: How about renaming it to `generate_initializers` and change the name of the existing `generate_initializers` to something else, preferably something that is less "stochastic" such as `create_initializers`? It's difficult to convey the difference without making the names longer (so, assuming users will read the docs).

chainer/testing/link.py
```diff
@@ -0,0 +1,549 @@
...
+            # When testing an initializer for a particular parameter, other
+            # initializers are set to the default initializer.
+
+            inits = [default_init, ] * len(params_inits)
+
+            for init in param_inits:
+                inits[i_param] = init
+                self._test_single_initializer(i_param, inits, backend_config)
+
+    def _test_single_initializer(self, i_param, inits, backend_config):
```
niboshi: Write comments about logic (before each block).

chainer/testing/link.py
```diff
@@ -0,0 +1,549 @@
...
+            # C-contiguous
+            return contig_arrays
+        assert False, (
+            'Invalid value of `contiguous`: {}'.format(self.contiguous))
+
+
+def _check_generated_initializer(init):
+    if isinstance(init, InitializerPair):
+        init = init.second
+        if init is None:
```
niboshi: Why comparing with `None`?
How about calling `_check_is_initializer_like`?

chainer/testing/link.py
niboshi: How about `agrument_value` vs `expected_initializer`?

chainer/testing/link.py
hvy: `None` actually passes `_check_is_initializer_like`, maybe it shouldn't.

chainer/testing/link.py
hvy: Changes as suggested.

chainer/testing/link.py
hvy: Fixed. Can you check if it's what you mean?

chainer/testing/link.py
niboshi: Discussed: `generate_params` and `get_initializers` respectively.

tests/chainer_tests/links_tests/normalization_tests/test_batch_normalization.py
niboshi: Discussed: 
* Resampling is not needed in this particular test. Currently we don't come up with links that need such resampling. That can be implemented by resampling in `create_link` if such links are found. 
* Call `before_test` once before each test.

tests/chainer_tests/links_tests/normalization_tests/test_batch_normalization.py
niboshi: Discussed:
Leave the test and add TODO to fix `BackendConfig`not to change the current device (and remove this test).

chainer/testing/link.py
hvy: Fixed as discussed.

tests/chainer_tests/links_tests/normalization_tests/test_batch_normalization.py
hvy: Fixed as disccused.

tests/chainer_tests/links_tests/normalization_tests/test_batch_normalization.py
hvy: Added comments as disccused.

chainer/testing/link.py
hvy: Fixed to not pass None.

chainer/gradient_check.py
```diff
@@ -583,28 +590,30 @@ def _directional_numeric_gradients(self, directions, y0_data):
...
-            [x for x, no_grad in six.moves.zip(x_vars, no_grads)
-             if not no_grad]
-            + list(params))
+
+        x_data_filtered = [
+            x.array for x, skip in six.moves.zip(x_vars, no_grads) if not skip]
 
         if dtype is None:
-            casted_data = [x.array for x in variables]
+            casted_data = [x for x in x_data_filtered + params_data]
```
niboshi: `casted_data` has both `ndarray`s and `Parameter`s. That looks weird. `data` stands for "array".

chainer/gradient_check.py
```diff
@@ -370,7 +370,7 @@ class _CheckBackward(object):
 
     def __init__(
             self, func, x_data, y_grad, params, eps, atol, rtol, no_grads,
-            dtype, detect_nondifferentiable):
+            dtype, detect_nondifferentiable, params_to_func):
         if dtype is not None and numpy.dtype(dtype).kind != 'f':
             raise ValueError('`dtype` is allowed only float type')
```
niboshi: Let's check the precondition: if `params_to_func` is `True`, `params` must be `ndarray`s.

chainer/testing/test.py
```diff
@@ -0,0 +1,180 @@
...
+        test_error_cls.fail(
+            '{}\n'
+            'Expected shapes and dtypes: {}\n'
+            'Actual shapes and dtypes:   {}\n'.format(
+                message,
+                utils._format_array_props(expected_arrays),
+                utils._format_array_props(actual_arrays)))
+
+
+# Utilities for contiguousness tests.
```
niboshi: These should not be here: these are closely coupled with the tests. It's better to put within `tests` directory.

chainer/testing/test.py
```diff
@@ -0,0 +1,180 @@
+import contextlib
```
niboshi: Isn't there any better naming than `test`?
Or how about just merging `function.py` and `link.py`?

chainer/gradient_check.py
```diff
@@ -370,7 +370,7 @@ class _CheckBackward(object):
 
     def __init__(
             self, func, x_data, y_grad, params, eps, atol, rtol, no_grads,
-            dtype, detect_nondifferentiable):
+            dtype, detect_nondifferentiable, params_to_func):
```
niboshi: Isn't there any better name than `params_to_func`?
I feel it's difficult to tell the meaning.

chainer/testing/test.py
hvy: Do you have any suggestions? We could merge them together but would still need a new name.

niboshi:
* `BatchNormalizationTestWithoutGammaAndBeta`
  Let's add TODO to rewrite this as well.
* `TestInitialize`
* `TestDefaultInitializer`
  How about removing them?

chainer/testing/link.py
```diff
@@ -0,0 +1,557 @@
...
+    In that case, the first element should correspond to the initializer-like
+    argument passed to the constructor of the link, and the second element
+    correspond to the actual initializer-like object used by the link.
+    In many cases, they are the same, except when passing ``None`` to the
+    constructor.
+    """
+
+    def __init__(self, argument_value, expected_initializer):
+        if argument_value is not None:
+            initializers._check_is_initializer_like(argument_value)
```
niboshi: I don't think this check is needed (it could be `str` for example)

chainer/testing/link.py
```diff
@@ -0,0 +1,557 @@
...
+            'Invalid value of `contiguous`: {}'.format(self.contiguous))
+
+
+def _check_generated_initializer(init):
+    if isinstance(init, InitializerPair):
+        init = init.expected_initializer
+    initializers._check_is_initializer_like(init)
+
+
+def _get_initializer(init):
```
niboshi: `_get_initializer_argument_value`?

hvy: Fixed.

chainer/gradient_check.py
hvy: Doesn't `casted_data` only contains "arrays" in the current implementation?

chainer/gradient_check.py
hvy: Thanks, fixed.

chainer/testing/test.py
hvy: Fixed.

chainer/testing/link.py
hvy: Fixed.

chainer/gradient_check.py
hvy: Fixed.

hvy:
Separated into 3 classes as discussed offline.

- `LinkTestCase`: Forward and backward specific methods and tests.
- `LinkInitializersTestCase`: Initializer specific methods and test.
- `LinkTestImpl`: Common logic for the above two.

A link test must inherit and implement `LinkTestImpl` and either of the former two.

tests/chainer_tests/links_tests/normalization_tests/test_batch_normalization.py
```diff
@@ -37,118 +29,206 @@ def _batch_normalization(expander, gamma, beta, x, mean, var, eps, test):
...
+    mean = mean[expander]
+    if test:
+        std = numpy.sqrt(var[expander])
+    else:
+        std = numpy.sqrt(var[expander] + eps)
+    y_expect = gamma * (x - mean) / std + beta
+    return y_expect
+
+
+class BatchNormalizationTestBase(testing.LinkTestImpl):
```
niboshi: Implementing in inheritance, I don't think `LinkTestImpl` is needed: the base implementation can simply inherit from `object`.

tests/chainer_tests/links_tests/normalization_tests/test_batch_normalization.py
```diff
@@ -37,118 +29,206 @@ def _batch_normalization(expander, gamma, beta, x, mean, var, eps, test):
...
+            with chainer.using_config('train', not self.test):
+                y = link(x, finetune=self.finetune)
+        return y,
+
+
+@testing.parameterize(*(testing.product_dict(
+    testing.product({
+        'dtype': [numpy.float16, numpy.float32, numpy.float64],
+    }))))
+@_inject_backend_tests
```
niboshi: Needs discussion: whether to parameterize backends in initializers test.

hvy:
Discussed and agreed offline:
- Hide `testing.LinkTestImpl` and let test writer inherit from `object` instead, treating it like a mixin.
- Keep backend parameterization for initializers test since some links (such as BN) might want it. It will still be possible to restrict the parameterization values.

tests/chainer_tests/links_tests/normalization_tests/test_batch_normalization.py
hvy: Leaving it as is according to offline discussions. https://github.com/chainer/chainer/pull/6187#issuecomment-471829926

tests/chainer_tests/links_tests/normalization_tests/test_batch_normalization.py
hvy: Fixed.

hvy:
PTAL.

chainer/testing/function_link.py
```diff
@@ -0,0 +1,1116 @@
...
+    def test_backward(self, backend_config):
+        """Tests backward computation."""
+        self.run_test_backward(backend_config)
+
+    def test_double_backward(self, backend_config):
+        """Tests double-backward computation."""
+        self.run_test_double_backward(backend_config)
+
+
+class _LinkTestImpl(object):
```
niboshi: How about `_LinkTestBase`?
`Impl` implies aggregation.

chainer/testing/function_link.py
```diff
@@ -0,0 +1,1116 @@
...
+
+    contiguous = None
+    backend_config = None
+
+    # List of parameter names represented as strings.
+    # I.e. ['gamma', 'beta'] for BatchNormalization.
+    param_names = []
```
niboshi: It's better to use tuple. It should be immutable.

chainer/testing/function_link.py
```diff
@@ -0,0 +1,1116 @@
...
+        if not isinstance(link, chainer.Link):
+            raise TypeError(
+                '`create_link` must return a chainer.Link object.')
+
+        link.to_device(backend_config.device)
+
+        return link
+
+    def _create_initialized_link(
+            self, inits, backend_config, return_inputs_outputs=False):
```
niboshi: >`return_inputs_outputs=False`

How about always returning it?

chainer/testing/function_link.py
```diff
@@ -0,0 +1,1116 @@
...
+        self.before_test('test_backward')
+
+        # avoid cyclic import
+        from chainer import gradient_check
+
+        def do_check():
+            inits = self._generate_params()
+            link = self._create_initialized_link(inits, backend_config)
+
+            def f(inputs, ps):
```
niboshi: How about defining it just before use?

chainer/testing/function_link.py
```diff
@@ -0,0 +1,1116 @@
...
+        params = _get_link_params(link, self.param_names)
+        params_np = [cpu_device.send(p.array) for p in params]
+
+        expected_outputs_np = self._forward_expected(inputs_np, params_np)
+
+        _check_forward_output_arrays_equal(
+            expected_outputs_np, outputs_xp,
+            'forward', LinkTestError, **self.check_forward_options)
+
+    def test_backward(self, backend_config):
```
niboshi: As the logic is not trivial, please write more code comments.

chainer/testing/function_link.py
```diff
@@ -0,0 +1,1116 @@
...
+        raise NotImplementedError('get_initializers is not implemented.')
+
+    def test_initializers(self, backend_config):
+        """Tests that the parameters of a links are correctly initialized."""
+
+        self.before_test('test_initializers')
+
+        params_inits = self._get_initializers()
+
+        for i_param, param_inits in enumerate(params_inits):
```
niboshi: How about leaving TODO to optimize this test? (to minimize the number of iterations)

chainer/testing/function_link.py
```diff
@@ -0,0 +1,1116 @@
...
+        inputs_xp = self._to_noncontiguous_as_needed(inputs_xp)
+        input_vars = tuple([chainer.Variable(i) for i in inputs_xp])
+        # Compute forward of the link and initialize its parameters.
+        output_vars = self._forward(link, input_vars, backend_config)
+        outputs_xp = [v.array for v in output_vars]
+
+        # Expected outputs are computed on the CPU so inputs and parameters
+        # must be transferred.
+        cpu_device = backend.CpuDevice()
+        params = _get_link_params(link, self.param_names)
```
niboshi: I think all the parameter in the link is needed (not only those specified in `self.param_names`).

chainer/initializers/__init__.py
```diff
@@ -110,3 +110,12 @@ def _get_initializer(initializer):
     if not callable(initializer):
         raise TypeError('invalid type of initializer: %s' % type(initializer))
     return initializer
+
+
+def _check_is_initializer_like(initializer):
```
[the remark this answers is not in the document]
hvy: Is the documentation for `None` correct? I don't think so since the default initializer differs between links.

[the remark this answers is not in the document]
hvy: I should maybe add a check for callable.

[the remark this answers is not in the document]
hvy: Added callable.

chainer/testing/function_link.py
hvy: Thanks, fixed.

chainer/testing/function_link.py
hvy: Thanks, fixed.

chainer/testing/function_link.py
hvy: Thanks, fixed.

chainer/testing/function_link.py
hvy: Fixed.

chainer/testing/function_link.py
hvy: Fixed.

chainer/testing/function_link.py
hvy: Fixed.

hvy:
Fixed conflicts.

chainer/testing/function_link.py
hvy: Agreed offline to pass the link object (transferred to CPU).

hvy: Fixed.

hvy:
PTAL.

chainer/testing/function_link.py
```diff
@@ -0,0 +1,1126 @@
...
+        inputs_xp = backend_config.get_array(inputs_np)
+        inputs_xp = self._to_noncontiguous_as_needed(inputs_xp)
+        input_vars = tuple([chainer.Variable(i) for i in inputs_xp])
+        # Compute forward of the link and initialize its parameters.
+        output_vars = self._forward(link, input_vars, backend_config)
+        outputs_xp = [v.array for v in output_vars]
+
+        # Expected outputs are computed on the CPU so the link must be
+        # transferred.
+        link.to_cpu()
```
niboshi: Let's use `to_device`.

chainer/initializers/__init__.py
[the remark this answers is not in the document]
niboshi: See the documentation:

>Unless explicitly specified, it is LeCunNormal with scale value 1.

It's consistent with the implementation.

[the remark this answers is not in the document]
hvy: Please correct me if I'm wrong but are you referring to this? https://github.com/chainer/chainer/blob/master/chainer/initializers/__init__.py#L104

chainer/testing/function_link.py
hvy: Fixed.

chainer/initializers/__init__.py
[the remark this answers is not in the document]
niboshi: Yes.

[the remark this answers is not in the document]
hvy: Fixed.

hvy:
PTAL.

chainer/testing/function_link.py
```diff
@@ -0,0 +1,1131 @@
...
+        #
+        # Note that for this to work, the expected parameter must be inferred
+        # deterministically.
+        expected_init = _get_expected_initializer(inits_orig[i_param])
+        expected_np = numpy.empty_like(param_np)
+        expected_init(expected_np)
+
+        # Compare the values of the expected and actual parameter.
+        _check_forward_output_arrays_equal(
+            (expected_np,), (param_np,), 'forward', LinkTestError,
```
niboshi: This is not the "outputs" of the "foward". I think error message is confusing.

tests/chainer_tests/testing_tests/test_function_link.py
```diff
@@ -441,4 +444,306 @@ def before_test(self, test_name):
...
+
+    """Incorrect implementation of dot (parameter initialization)."""
+
+    def __init__(self, in_size, out_size, initial_p=None):
+        # Ignores given initializer here.
+        super(DotLinkIncorrectInitialization, self).__init__(
+            in_size, out_size, initializers.Constant(0))
+
+
+class DotLinkTestImpl(object):
```
niboshi: `DotLinkTestBase`

tests/chainer_tests/links_tests/normalization_tests/test_batch_normalization.py
```diff
@@ -37,118 +29,207 @@ def _batch_normalization(expander, gamma, beta, x, mean, var, eps, test):
...
-            x = chainer.Variable(x_data)
-            y = self.link(x)
-            self.assertEqual(y.data.dtype, self.dtype)
+            y = link(x, finetune=self.finetune)
+        return y,
 
-        y_expect = _batch_normalization(
-            self.expander, self.gamma, self.beta, self.x, self.mean,
-            self.var, self.link.eps, self.test)
+    def forward_expected(self, inputs, link):
```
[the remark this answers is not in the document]
hvy: Fixed.

chainer/testing/function_link.py
hvy: How about generalizing `_check_forward_output_arrays_equal`?

hvy: Fixed.

tests/chainer_tests/testing_tests/test_function_link.py
hvy: Fixed.

hvy:
PTAL.

chainer/testing/function_link.py
```diff
@@ -0,0 +1,1173 @@
...
+        Typically a skip logic is implemented by conditionally raising
+        :class:`unittest.SkipTest`.
+        ``test_name`` is one of ``'test_forward'`` and ``'test_backward'``.
+
+    ``forward_expected(self, link, inputs)``
+        Implements the expectation of the target forward function.
+        ``link`` is the initialized link that was used to compute the actual
+        forward which the results of this method will be compared against.
+        The link is guaranteed to reside on the CPU.
+        ``inputs`` is a tuple of :class:`numpy.ndarray`\\ s and
```
niboshi: Remove "and"?

hvy: Thanks, fixed.

hvy:
@niboshi cleaned up history in #6559.

hvy:
Closing since https://github.com/chainer/chainer/pull/6559#event-2214243423 based on this PR has been merged.

Issue #6071, opened 2019-01-23
Rewrite function tests using `testing.FunctionTestCase`


niboshi:
We have introduced `testing.FunctionTestCase` (#3499).
This tool enables writing a function test for various backends (including ChainerX) in much simpler way.
This tool also provides an option to automatically dodge non-differentiable inputs.
We should apply this to all the existing function tests, and assure ChainerX support in those functions. 

~(Note: we should merge #6069 first to skip float16 in ChainerX backends.)~

## To get started
Please declare your will to work on a function, by posting a comment in this issue. If you're a collaborator, just write your name next to the function name.

### NumPy-compatible functions are not yet implemented
Write tentative implementation that falls back to numpy/cupy, in `chainerx/_fallback_workarounds.py`. For module-level functions (e.g. `chainerx.hstack`), write in `_populate_module_functions`. For `ndarray` methods, write in `_populate_ndarray`. Note that functions written in this way cannot support backprop in their own, but it's sufficient for computations in `FunctionNode.forward`, for example.
### Array type mismatch
In `FunctionNode` fallback (in which chainerx arrays are automatically converted to numpy/cupy and the corresponding `FunctionNode.forward` method is called), array attributes are also automatically converted between numpy/cupy and chainerx arrays. This is not the case if the array attribute is wrapped in a tuple, dict, etc. In this case, the attribute must be converted manually, using `backend.to_chainerx` and `backend.from_chainerx`.

76 done of 165 tests, 46% complete!

----

* [ ] functions_tests/pooling_tests/test_roi_max_align_2d.py
* [ ] functions_tests/pooling_tests/test_pooling_nd_kernel.py
* [ ] functions_tests/pooling_tests/test_max_pooling_2d.py
* [x] functions_tests/pooling_tests/test_unpooling_2d.py #6861
* [x] functions_tests/pooling_tests/test_average_pooling_2d.py #6211
* [ ] functions_tests/pooling_tests/test_roi_average_align_2d.py
* [ ] functions_tests/pooling_tests/test_spatial_pyramid_pooling_2d.py
* [ ] functions_tests/pooling_tests/test_roi_pooling_2d.py
* [ ] functions_tests/pooling_tests/test_roi_max_pooling_2d.py
* [x] functions_tests/pooling_tests/test_unpooling_nd.py #6861
* [ ] functions_tests/pooling_tests/test_roi_average_pooling_2d.py
* [x] functions_tests/pooling_tests/test_max_pooling_nd.py #3499
[...]

PR #5819, opened 2018-12-10
Simplify `F.triplet` test

niboshi:
Applies the function test case https://github.com/chainer/chainer/pull/3499.

PR #6195, opened 2019-02-07
Rewrite the tests of `F.triplet`

### LINKED ISSUE #6190: `F.triplet` test still fails sometime
The test of `F.triplet` sometime fails because of unreasonably large value of numerical grad result in FP16 double backward test. It may be because that input of the second order backward would contain some values around non-differentiable points.

crcrpar:
rel: #6071 
Resolves #6190 

This PR aims at simplifying and stabilizing the tests of `F.triplet` using `chainer.testing.FunctionTestCase`.

### DIFF
--- a/chainerx/_fallback_workarounds.py
+++ b/chainerx/_fallback_workarounds.py
@@ -136,6 +136,14 @@ def _populate_module_functions():
 
     chainerx.vstack = _vstack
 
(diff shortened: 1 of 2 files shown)


niboshi:
Oh, it's duplicate of  #5819.

PR #6197, opened 2019-02-08
Use `Variable._init_unchecked` instead of `__init__`

okapies:
Replace `__init__()` used in the core functionalities to instantiate `Variable` with `_init_unchecked()` introduced in #6033 to reduce the overhead of validations.


okapies:
Jenkins, test this please.

niboshi:
I think this method should be used sparingly, only in hot spots.
It should be avoided in `gradient_check` and serialization, for example.
I'm not sure for other cases. Maybe it's better to measure first.

okapies:
@niboshi Thanks. I'll check the result of performance testing.

okuta:
I agree @niboshi 's opinion.

okapies:
After the inspection of the performance profile, I'm going to close it because I couldn't find a remarkable overhead caused by `Variable.__init__` at the moment. The largest one is an usage in `_to_variable_with_chainerx_fallback_array` but it just account for 0.29%. I'll open an another PR if such an overhead is found.

PR #6201, opened 2019-02-08
[WIP] layer-wise allreduce

shu65:
This PR adds new optimizer with layer-wise allreduce.

### DIFF
--- a/chainer/variable.py
+++ b/chainer/variable.py
@@ -492,6 +492,7 @@ Actual: {0}'''.format(type(data))
         self._node = VariableNode(self, name)
         self._grad_var = None if grad is None else Variable(grad)
         self._loss_scale = None
+        self._post_backward_hooks = {}
 
     def __copy__(self):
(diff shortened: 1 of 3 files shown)


shu65:
In order to implement layer-wise allreduce, I want to change the backward as follows.


1．add hook to be executed when the backward ends
https://github.com/chainer/chainer/blob/db75dd987d5397dd851490339e5d496fe849b706/chainer/variable.py#L1186

2. search timing for hook execution
https://github.com/chainer/chainer/blob/db75dd987d5397dd851490339e5d496fe849b706/chainer/variable.py#L985-L1005

3. execute hook
https://github.com/chainer/chainer/blob/db75dd987d5397dd851490339e5d496fe849b706/chainer/variable.py#L1044

Known issue
- On the network where forget is used, it is impossible to determine the end timing of the backward

What do you think about this design?
@beam2d @toslunar

PR #6204, opened 2019-02-08
Optimized ChainerX-to-CuPy `ndarray` conversion

okapies:
Add `to_cupy` function to `chainerx`. This patch reduces the overhead of retrieving many properties of a chainerx array in `_array_to_gpu`. It retrieves the cached Python objects such like `numpy` and `cupy` from `pymodule.h`. It also introduces `GetDataSize` to calculate the data size of the array directly.

```python
>>> import timeit
>>> setup = """
... import cupy
... import chainerx as chx
... device = chx.get_device('cuda:0')
... array = chx.array(range(10), dtype=chx.float32, device=device)
... """
>>> before = """
... cupy.ndarray(
...     array.shape,
...     array.dtype,
...     cupy.cuda.MemoryPointer(
...         cupy.cuda.UnownedMemory(
...             array.data_ptr + a.offset,
...             array.data_size,
...             array,
...             array.device.index),
...         0),
...     strides=array.strides)
... """
>>> timeit.timeit(before, setup=setup, number=1000000)
35.3160393089056
>>> after = "chx.to_cupy(array)"
>>> timeit.timeit(after, setup=setup, number=1000000)
13.969784777611494
```

## Note
I will handle the optimization of `GetNumpyDtypeFromModule` (0f1b011) in another PR.


chainerx_cc/chainerx/python/array.cc
```diff
@@ -85,6 +86,29 @@ py::array MakeNumpyArrayFromArray(const ArrayBodyPtr& self, bool copy) {
     return py::array{dtype, shape, strides, ptr, py::cast(internal::MoveArrayBody(std::move(array)))};
 }
 
+py::object MakeCupyArrayFromArray(const py::module& m, py::handle self) {
+    Array array = Array{py::cast<ArrayBodyPtr>(self)};
+    const Device& device = array.device();
+    // TODO(okapies): rejects if array's device is not compatible with cupy
```
okapies: I have no idea how to check whether the device is compatible with CuPy or not...

niboshi:
Why is `py::get_or_create_shared_data` needed? It seems this function is undocumented.

okapies:
@niboshi We need to cache the result of `py::module::import` because of its overhead. `py::get_or_create_shared_data` is a variation of `get_shared_data` and `set_shared_data` documented in [this section](https://pybind11.readthedocs.io/en/master/advanced/misc.html?highlight=get_shared_data#partitioning-code-over-multiple-extension-modules). It is used to [hold a cache of the numpy dtypes](https://github.com/pybind/pybind11/blob/25abf7e/include/pybind11/numpy.h#L83-L109) in pybind itself. It seems it is a good and safe place to keep the user-defined data for pybind application like ChainerX.

Ref. https://gitter.im/pybind/Lobby?at=5c5d9ee028c89123cb954a32

niboshi:
@okapies Do you still think we need to use it? On slack you said simply holding `py::handle` should be enough.

okapies:
I think so because we can safely assume that `numpy_module` will be initialized after pybind becomes ready unlike using global static variable.

niboshi:
@okapies 
By using function-local static storage, it will be initialized on first usage (just as previously done in `MakeArrayIndex`).

okapies:
Ok, that makes sense. I'll fix it.

niboshi:
I think `numpy_module` and `cupy_module` are overkill, too.
`numpy` and `cupy` are only used as workarounds. Something like `static py::handle unowned_memory = py::module::import("cupy").attr("cuda").attr("memory").("UnownedMemory");` looks fine.

okapies:
Ok, fixed.

okapies:
Jenkins, test this please.

chainerx_cc/chainerx/python/array.cc
```diff
@@ -85,6 +86,28 @@ py::array MakeNumpyArrayFromArray(const ArrayBodyPtr& self, bool copy) {
     return py::array{dtype, shape, strides, ptr, py::cast(internal::MoveArrayBody(std::move(array)))};
 }
 
+py::object MakeCupyArrayFromArray(const py::module& m, py::handle self) {
+    Array array = Array{py::cast<ArrayBodyPtr>(self)};
```
niboshi: Simplify
```suggestion
    Array array{py::cast<ArrayBodyPtr>(self)};
```

chainerx_cc/chainerx/python/array.cc
```diff
@@ -85,6 +86,28 @@ py::array MakeNumpyArrayFromArray(const ArrayBodyPtr& self, bool copy) {
...
+    const Device& device = array.device();
```
niboshi: `Device&` is never const.
```suggestion
    Device& device = array.device();
```

chainerx_cc/chainerx/python/array.cc
```diff
@@ -135,7 +158,11 @@ void InitChainerxArray(pybind11::module& m) {
           py::arg("shape"),
           py::arg("dtype"),
           py::arg("device") = nullptr);
-    m.def("to_numpy", &MakeNumpyArrayFromArray, py::arg("array"), py::arg("copy") = true);
+    m.def("to_numpy",
+          [m](const ArrayBodyPtr& array, bool copy) { return MakeNumpyArrayFromArray(m, array, copy); },
+          py::arg("array"),
+          py::arg("copy") = true);
+    m.def("to_cupy", [m](py::handle array) { return MakeCupyArrayFromArray(m, array); }, py::arg("array"));
```
niboshi: Let's add `_` (`_to_cupy`) to indicate it's for internal use.

chainerx_cc/chainerx/python/array.cc
```diff
@@ -85,6 +86,28 @@ py::array MakeNumpyArrayFromArray(const ArrayBodyPtr& self, bool copy) {
     return py::array{dtype, shape, strides, ptr, py::cast(internal::MoveArrayBody(std::move(array)))};
 }
 
+py::object MakeCupyArrayFromArray(const py::module& m, py::handle self) {
```
niboshi: How about adding `TODO` to state that this is a temporary workaround.

chainerx_cc/chainerx/python/pymodule.h
```diff
@@ -0,0 +1,50 @@
+#pragma once
```
niboshi: I'm not sure if we need this file.
How about resolving each time directly?

chainerx_cc/chainerx/strides.cc
```diff
@@ -66,4 +66,19 @@ std::tuple<int64_t, int64_t> GetDataRange(const Shape& shape, const Strides& str
     return std::tuple<int64_t, int64_t>{first, last};
 }
 
+int64_t GetDataSize(const Shape& shape, const Strides& strides, size_t item_size) {
```
niboshi: Do we need this function?
It's quite straight-forward to use `GetDataRange` IMO (and it's not used so often).

chainerx_cc/chainerx/python/pymodule.h
okapies: Because resolving Python objects each time incurs larger performance overhead.

niboshi: Hmm, I don't think that matters.
By storing them in `static` function-local variables, it should be equivalent to this implementation in terms of peformance.

okapies: Ah, ok. Do you mean I should store the instance in each caller function instead of declaring in the separate file? It makes sense in terms of performance but I also think that collecting `py::module::import` things in one place is useful for later maintenance. Any idea?

niboshi: We agreed offline to define these functions in this file.

Let's give it more clear name.
How about `py_cached_objects.h`?

chainerx_cc/chainerx/python/pymodule.h
```diff
@@ -0,0 +1,50 @@
...
+
+#include "chainerx/macro.h"
+
+namespace chainerx {
+namespace python {
+namespace python_internal {
+
+namespace py = pybind11;
+
+inline py::handle& numpy_module() {
```
niboshi: We use camel case for free functions.
How about `GetCachedNumpyModule`, and so on?

niboshi: Let's return `py::handle` by value.

niboshi:
Jenkins, test this please

niboshi:
Jenkins, test this please

okapies:
Jenkins, test this please

okapies:
It seems the failure is caused by #6922. I'll try to fix it.

okapies:
Jenkins, test this please

niboshi:
LGTM.
I'm sorry for taking time to merge.

okapies:
Thanks!

PR #6205, opened 2019-02-08
Fix seq2seq example encoding problem on Python3

takagi:
On Python3, `open` built-in function uses platform dependent encoding when `encoding` parameter is not supplied. This can cause `UnicodeDecodeError` in seq2seq example.

This PR fixes `seq2seq.py` to use `io.open` instead of built-in `open` to supply `encoding` parameter on both Python2 and Python3 environments to make the example work properly.

### DIFF
--- a/examples/seq2seq/seq2seq.py
+++ b/examples/seq2seq/seq2seq.py
@@ -2,6 +2,7 @@
 
 import argparse
 import datetime
+import io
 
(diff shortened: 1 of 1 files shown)


okuta:
jenkins, test this please.

okuta:
LGTM!

PR #6213, opened 2019-02-11
Simplify `_check_grad_type`

niboshi:
Removed `is_var_gx` argument, etc.

### DIFF
--- a/chainer/_backprop_utils.py
+++ b/chainer/_backprop_utils.py
@@ -139,32 +139,34 @@ def backprop_step(
             raise ValueError(func._get_error_message(msg))
 
     for i, gx in six.moves.zip(target_input_indexes, gxs):
-        if gx is not None:
-            grad_inputs[func.inputs[i]].append(gx)
+        if gx is None or gx.array is None:
+            continue
 
-            if is_debug:
-                node_x = func.inputs[i]
-                g_input_list = grad_inputs[node_x]
-                if gx.shape != node_x.shape:
+        grad_inputs[func.inputs[i]].append(gx)
+
+        if is_debug:
+            node_x = func.inputs[i]
+            g_input_list = grad_inputs[node_x]
+            if gx.shape != node_x.shape:
+                raise ValueError(func._get_error_message(
+                    'shape of gradients returned from backward is '
+                    'incorrect: '
+                    'input-index={}, actual {} != expected {}'.format(
+                        i, gx.shape, node_x.shape)))
+            if gx is not None and g_input_list:
+                g_input = g_input_list[0]
+                if gx.shape != g_input.shape:
                     raise ValueError(func._get_error_message(
                         'shape of gradients returned from backward is '
                         'incorrect: '
                         'input-index={}, actual {} != expected {}'.format(
-                            i, gx.shape, node_x.shape)))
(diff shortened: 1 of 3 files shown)


Crissman:
Jenkins, test this please.

niboshi:
Fixed the test failure.

niboshi:
This PR slightly improves the performance of backprop.
19.1 ms -> 18.6 ms (using the same measurement as #6263)

niboshi:
This PR improves the performance of backprop.
19.1 ms -> 16.5 ms (using the same measurement as #6263)

chainer/_backprop_utils.py
```diff
@@ -139,32 +139,34 @@ def backprop_step(
             raise ValueError(func._get_error_message(msg))
 
     for i, gx in six.moves.zip(target_input_indexes, gxs):
-        if gx is not None:
-            grad_inputs[func.inputs[i]].append(gx)
+        if gx is None or gx.array is None:
```
toslunar: It seems that the change will make Chainer more tolerant of returning `Variable(None)`.  I'd rather not treat `Variable(None)` as `None`.

niboshi: Currently Chainer does allow `Variable(None)`, so I think this check is legitimate.
Do you have any alternative?

toslunar: To preserve the behavior for now,
```python
        if gx is None:
            continue
        grad_inputs[func.inputs[i]].append(gx)
        if is_debug and gx.array is None:
            raise ...
```

niboshi: Do you mean we're going to disallow `Variable(None)` in this PR?

toslunar: I mean that `Variable(None)` is checked if `is_debug` before the PR.  I've not checked older codes.

niboshi: Where is the check? I cannot find it.
Currently some function implementations return `Variable(None)`, so that would mean they're broken in debug mode.

toslunar: OK.  No "check" for `Variable(None)` is there but a line with `gx.shape` just fails in debug mode.

As we don't have a consensus on `Variable(None)`, I'd like code-fix/performance PRs to preserve the behavior.

niboshi: Currently Chainer does (or should) support `Variable(None)`, because some built-in functions return such variables.
So I'd say the current behavior is a bug.
Perhaps we can put `cat:bug` label to this PR.

toslunar: The PR turns `Variable(None)` into `None`.

> Currently Chainer does (or should) support `Variable(None)`, because some built-in functions return such variables.
> So I'd say the current behavior is a bug.

Then the fix for the bug could be
```python
        if gx is None:
            continue
        grad_inputs[func.inputs[i]].append(gx)
        if gx.array is None:
            continue
```

niboshi: Does that make difference?
I was not sure about that point but I thought it should not, so I fixed as the current implementation.

Crissman:
Jenkins, test this please

Crissman:
LGTM. Merging.

PR #6214, opened 2019-02-12
Use `_has_chainerx_array` in `Variable`

okuta:
This PR is performance improvement that relates #6016.

### DIFF
--- a/chainer/variable.py
+++ b/chainer/variable.py
@@ -608,7 +608,7 @@ class Variable(object):
         # _grad_var._data[0] and self._data[0].grad and recreates _grad_var
         # as necessary. (chainerx.ndarray.grad can be altered independently
         # from chainer)
-        if self.xp is chainerx:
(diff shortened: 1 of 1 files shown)


okapies:
It seems great but these kind of fix was once reverted based on the discussion in https://github.com/chainer/chainer/pull/6016#pullrequestreview-196876021. How do you think? @niboshi

niboshi:
@okapies 
I thought it's not worthwhile when it's viewed as performance optimization, but now I feel it's ok because this can be seen as refactoring (`_has_chainerx_array` is more straightforward to distinguish chainerx arrays than `xp is chainerx`).

niboshi:
In that sense I'd put `cat:code-fix` rather than `cat:performance`.

okuta:
Thank you for your advice. I added the label.

okuta:
jenkins, test this please.

niboshi:
LGTM

PR #6217, opened 2019-02-12
Omit `VariableNode` creation for `Variable`s with `requires_grad=False`

niboshi:
This PR cuts `VariableNode` creation for `Variable`s with `requires_grad=False`.

I cleaned up some less useful interfaces (`VariableNode.retain_data()` and `Variable.retain_data()`) and confined into a single method `VariableNode.retain_data()` with `data` argument.
Also added `data` argument to `VariableNode.__init__()`.

I assume they're interfaces for internal implementation. In that sense I think it's possible to go without the `no-compat` label, but strictly speaking this is `no-compat`.


niboshi:
What to discuss?

PR #6229, opened 2019-02-14
Remove `CHAINERX_VISIBILITY_HIDDEN`

niboshi:


hvy:
jenkins, test this please.

hvy:
It is required because pybind11 objects are hidden by default and ChainerX types holding these objects cannot have a public visibility. https://pybind11.readthedocs.io/en/stable/faq.html#someclass-declared-with-greater-visibility-than-the-type-of-its-field-someclass-member-wattributes

PR #6233, opened 2019-02-14
skip MultiprocessIterator finalization at exit

grafi-tt:
Finalization routine of `MultiprocessIterator` hangs when uncaught exception is triggered. Tested with Python 3.6.6 on Linux x86_64. Even though `with` statement is generally recommended, the hang is undesirable. This PR workaround the issue by skipping `MultiprocessIterator` finalization after `atexit` hook execution.

Such type of a hang occurs if `iterator.__del__` is called when the interpreter finalizing itself. In the following program, probably the traceback held by the exception object defers `__del__` call until the interpreter finalization.

```python
import chainer

class Dataset(chainer.dataset.DatasetMixin):
    def __len__(self):
        return 100

    def get_example(self, i):
        return i

def main():
    dataset = Dataset()
    iterator = chainer.iterators.MultiprocessIterator(dataset, 16, repeat=False, shuffle=False, n_processes=8)
    for batch in iterator:
        raise RuntimeError("Err")

main()
```

### DIFF
--- a/chainer/iterators/multiprocess_iterator.py
+++ b/chainer/iterators/multiprocess_iterator.py
@@ -1,4 +1,5 @@
 from __future__ import division
+import atexit
 import datetime
 import multiprocessing
 from multiprocessing import sharedctypes  # type: ignore
@@ -615,3 +616,13 @@ def _unpack(data, mem):
     elif t is _PackedNdarray:
         data = data.unpack(mem)
     return data
+
+
+# When the CPython interpreter exits, daemon threads abruptly exits without
+# any Python-level finalization. This makes `thread.is_alive` unreliable.
+# Pooled processes are also terminated by `multiprocessing` module.
+# As there is no way to ensure correct finalization, we simply skip
+# `MultiprocessIterator`'s finalization routine.
+@atexit.register
+def _force_skip_finalization():
+    MultiprocessIterator._finalized = True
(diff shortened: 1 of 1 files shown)


niboshi:
Just a cosmetic point, but I think it's better to add some comment on the declaration of `_finalized` that it should not be assigned `False` as an instance attribute (that would make the class attribute ignored).

grafi-tt:
@niboshi Thanks. I added comments.

Crissman:
Jenkins, test this please

kuenishi:
I'm afraid this workaround causes regression of https://github.com/chainer/chainer/pull/4137 in case of custom loop. With Trainer it'd be ok as it finalizes iterators.

grafi-tt:
@kuenishi Thanks. I think removing `__del__` on a major release is fine provided appropriate documents are added. It is a design choice to be made by the core team, so I couldn't make a decision.

I believe many Jupyter users create `MultiprocessIterator` instances without `with` statement just because it is faster, even though those who seriously care resource leak should already have used something like `with` statement. I think we should make enough notice if we get rid of `__del__`.

Crissman:
LGTM. Merging.

Crissman:
Reverted the merge to allow for discussion about removing __del__ as opposed to this fix.

Crissman:
Discussing with the Chainer team. Will either revert this, or remove `__del__` depending on the decision. Thanks!

Issue #6250, opened 2019-02-15
Allow `output_data` in `forward_postprocess`


tkerola:
Currently, only input data is passed to the forward postprocess function hook.
If we could pass output data as well, several use cases of function hooks would become easier, such as simply finding the number of outputs, or output dimensions.

An example interface could be as follows (`None` for backwards compatibility):
```
def forward_postprocess(self, function, in_data, output_data=None):
```

The above interface is trivial to implement, but it could cause some issues with backwards compatibility, as current hooks assume a fixed number of parameters. Please let me know what you think.

PR #6252, opened 2019-02-15
[WIP] Call forward_postprocess with output values if the hook impl supports

belltailjp:
This is an initial attempt to fix https://github.com/chainer/chainer/issues/6250

In order to keep backward compatibility, it checks signature of the specified hook.
In case the hook has only one argument, it calls without output values.
If it has 2, it calls with output value, and if it has more, an error is raised.

Although it's still work-in-progress, I want to hear opinions about the feature design.

TODO
* Code quality
* Tests and documentation

----

### DIFF
--- a/chainer/function_hook.py
+++ b/chainer/function_hook.py
@@ -1,3 +1,5 @@
+import inspect
+
 import chainer
 
 
@@ -114,9 +116,24 @@ class FunctionHook(object):
         if self.name in function_hooks:
             raise KeyError('hook %s already exists' % self.name)
 
+        self._wrap_legacy_hook()
         function_hooks[self.name] = self
         self.added(None)
         return self
+
+    def _wrap_legacy_hook(self):
+        # If the forward_postprocess does not seem to receive out_data,
+        # replace it by a wrapper function that transfers only in_data.
+        # This makes older hooks compatible with the current hook system.
+        sig = inspect.signature(type(self).forward_postprocess)
+        if len(sig.parameters) == 3:
+            self.forward_postprocess = \
+                lambda _self, func, in_data, out_data=None: \
+                type(self).forward_postprocess(_self, func, in_data)
+        elif len(sig.parameters) != 4:
+            msg = ('{}.forward_postprocess has unsupported '
+                   'number of arguments.'.format(type(self).__name__))
+            raise RuntimeError(msg)
 
     def __exit__(self, *_):
         chainer.get_function_hooks()[self.name].deleted(None)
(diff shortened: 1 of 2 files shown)


tkerola:
By offline discussion, we considered that `inspect` might be heavy to call each forward pass, and might be replaced by wrapping `self.function_hook` once instead in the `FunctionHook` constructor.

belltailjp:
This PR checks hook signature in every `FunctionNode` call but it turns out that cost of `inspection` is not trivial.
(ResNet50+GTX1080Ti case, it takes 4.46% of time just for inspection cost, which is a bit too heavy)

niboshi:
The inspection could be done on hook registration (`FunctionNode.add_hook`) as well.
I think it's better because it's less trickier (no need to replace `FunctionHook.forward_postprocess`).
Also doing that in the constructor would break existing hook that do not call `super().__init__` (none of the existing built-in hooks do that).

Another comment is that if possible, I'd like to change the "new" interface to something similar to link hooks, to prevent future compatibility issues.

belltailjp:
Thanks for the suggestion! I've changed the implementation to do inspection in hook registration time (both for global hook and FunctionNode individual hook).

> I'd like to change the "new" interface to something similar to link hooks,

Does this mean redesigned arguments of forward_postprocess (and others) like `forward_postprocess(func, name, args, kwargs, out)`?

niboshi:
@belltailjp 
>Does this mean redesigned arguments of forward_postprocess (and others) like `forward_postprocess(func, name, args, kwargs, out)`?

No, I want to change so that arguments would be bundled in a single object. The signature would be `forward_postprocess(args)` and callback implementation would use `arg` like `arg.function`, `arg.in_data` and so on. That way we would be able to introduce new arguments in the future, without breaking user code.

belltailjp:
> (func, name, args, kwargs, out)
Oh, yes I somehow misunderstood that they are formed into one `args`.
Let me try that, but it would require big changes, so should take time quite a bit.

niboshi:
@belltailjp 
The team discussed and the following is the current conclusion.

* Inspecting the signature might not be robust and better avoided. For example, it won't work for C-implemented functions.
* It's better to make a new interface: say, `FunctionNodeHook`.
  * `FunctionNodeHook` might be a bit too long... but might be more precise because it works with `FunctionNode`.
* OK to bundle arguments like `LinkHook`.

How do you think? Any opinions and/or suggestions are of course welcome!

niboshi:
What's the status of this PR?

PR #6254, opened 2019-02-16
add option `align` to `chainer.TimerHook.print_report()`

hitsgub:
The units of elapsed times are not aligned in current version.
It is hard to find a processing time bottleneck.
This PR adds option `align` to the member function `print_report()`.
For example,

xxx.print_report()

                   FunctionName  ElapsedTime  Occurrence
                 LinearFunction      1.24sec        3900
                           ReLU     593.05ms        2600
            SoftmaxCrossEntropy     824.11ms        1300
                       Accuracy     176.54ms         700

to

xxx.print_report(align=True)

                   FunctionName  ElapsedTime  Occurrence
                 LinearFunction      1.24sec        3900
                           ReLU       0.59sec       2600
            SoftmaxCrossEntropy       0.82sec       1300
                       Accuracy       0.18sec        700

### DIFF
--- a/chainer/function_hooks/timer.py
+++ b/chainer/function_hooks/timer.py
@@ -132,11 +132,31 @@ class TimerHook(function_hook.FunctionHook):
             second *= 1000.0
         return '%.2f%s' % (second, 'ns')
 
-    def print_report(self, file=sys.stdout):
-        """Prints a summary report of time profiling in functions."""
+    def _align(self, second):
+        """Returns align info: (factor, unit)."""
+        factor = 1
+        for unit in ['sec', 'ms', 'us']:
+            if second * factor >= 1:
+                return factor, unit
+            factor *= 1000.0
+        return factor, 'ns'
+
+    def print_report(self, align=False, file=sys.stdout):
+        """Prints a summary report of time profiling in functions.
+
+        Args:
+            align (bool): If `True`, units of times are aligned to the largest.
+        """
         entries = [['FunctionName', 'ElapsedTime', 'Occurrence']]
+        if align:
+            max_time = max(record['elapsed_time'] for record in self.summary().values())
+            factor, unit = self._align(max_time)
         for function_name, record in self.summary().items():
-            elapsed_time = self._humanized_time(record['elapsed_time'])
+            second = record['elapsed_time']
+            if align:
+                elapsed_time = '%3.2f%s' % (second * factor, unit)
+            else:
+                elapsed_time = self._humanized_time(second)
             occurrence = str(record['occurrence'])
             entries.append([function_name, elapsed_time, occurrence])
         entry_widths = []
--- a/tests/chainer_tests/function_hooks_tests/test_timer.py
+++ b/tests/chainer_tests/function_hooks_tests/test_timer.py
@@ -193,5 +193,17 @@ class TestTimerPrintReport(unittest.TestCase):
         actual = io.getvalue()
         six.assertRegex(self, actual, expect)
 
+    def test_print_report_align(self):
+        x = self.x
+        self.f.apply((chainer.Variable(x),))
+        self.f.apply((chainer.Variable(x),))
+        io = six.StringIO()
+        self.h.print_report(align=True, file=io)
+        expect = r'''\AFunctionName  ElapsedTime  Occurrence
+ +Exp +[0-9.\-e]+.s +[0-9]+$
+'''
+        actual = io.getvalue()
+        six.assertRegex(self, actual, expect)
+
 
 testing.run_module(__name__, __file__)


toslunar:
When I saw the title, I thought the PR adds `align='left'`.  How about giving a longer name, e.g. `align_units`?

hitsgub:
> When I saw the title, I thought the PR adds `align='left'`. How about giving a longer name, e.g. `align_units`?

Thanks, you're right.
I will fix it soon.

beam2d:
Rather than adding a boolean flag, how about adding an option to choose the units to print with a special value that indicates the same largest unit? I'm thinking something like `unit=str` where the values are either `us`, `ms`, `s`, or `auto`. To keep the current behavior, we may additionally support another special value, like `auto_foreach`, to choose the appropriate unit for each entry separately.

hitsgub:
> Rather than adding a boolean flag, how about adding an option to choose the units to print with a special value that indicates the same largest unit? I'm thinking something like `unit=str` where the values are either `us`, `ms`, `s`, or `auto`. To keep the current behavior, we may additionally support another special value, like `auto_foreach`, to choose the appropriate unit for each entry separately.

And I changed default behavior to `auto` in latest commit, please review it.

chainer/function_hooks/timer.py
```diff
@@ -132,11 +133,38 @@ def _humanized_time(self, second):
             second *= 1000.0
         return '%.2f%s' % (second, 'ns')
 
-    def print_report(self, file=sys.stdout):
-        """Prints a summary report of time profiling in functions."""
+    def _align(self, second):
```
beam2d: Now I feel this method name is misleading. How about, e.g., `_choose_unit`?

chainer/function_hooks/timer.py
```diff
@@ -132,11 +133,38 @@ def _humanized_time(self, second):
...
+        """Returns align info: (factor, unit)."""
+        factor = 1
+        for unit in ['sec', 'ms', 'us']:
+            if second * factor >= 1:
+                return factor, unit
+            factor *= 1000.0
+        return factor, 'ns'
```
beam2d: The logic looks duplicated with `_humanized_time`. Could you refactor them?

chainer/function_hooks/timer.py
```diff
@@ -132,11 +133,38 @@ def _humanized_time(self, second):
...
+            max_time = max(
+                record['elapsed_time'] for record in self.summary().values())
+            factor, unit = self._align(max_time)
+        elif unit != 'auto_foreach':
+            factor = self.table[unit]
         for function_name, record in self.summary().items():
-            elapsed_time = self._humanized_time(record['elapsed_time'])
+            second = record['elapsed_time']
+            if unit != 'auto_foreach':
+                elapsed_time = '%3.2f%s' % (second * factor, unit)
```
beam2d: This formatting logic is also duplicated with `_humanized_time`.

chainer/link_hooks/timer.py
```diff
@@ -129,11 +130,38 @@ def _humanized_time(self, second):
             second *= 1000.0
         return '%.2f%s' % (second, 'ns')
 
-    def print_report(self, file=sys.stdout):
-        """Prints a summary report of time profiling in links."""
+    def _align(self, second):
```
beam2d: ditto (method name)

chainer/link_hooks/timer.py
```diff
@@ -129,11 +130,38 @@ def _humanized_time(self, second):
...
+        """Returns align info: (factor, unit)."""
+        factor = 1
+        for unit in ['sec', 'ms', 'us']:
+            if second * factor >= 1:
+                return factor, unit
+            factor *= 1000.0
+        return factor, 'ns'
```
beam2d: ditto (logic duplication)

chainer/link_hooks/timer.py
```diff
@@ -129,11 +130,38 @@ def _humanized_time(self, second):
...
+            max_time = max(
+                record['elapsed_time'] for record in self.summary().values())
+            factor, unit = self._align(max_time)
+        elif unit != 'auto_foreach':
+            factor = self.table[unit]
         for link_name, record in self.summary().items():
-            elapsed_time = self._humanized_time(record['elapsed_time'])
+            second = record['elapsed_time']
+            if unit != 'auto_foreach':
+                elapsed_time = '%3.2f%s' % (second * factor, unit)
```
beam2d: ditto (logic duplication)

chainer/function_hooks/timer.py
hitsgub: That's good. I did it.

chainer/function_hooks/timer.py
hitsgub: Integrated them into `_choose_unit`.

chainer/function_hooks/timer.py
hitsgub: Changed to use `_choose_unit` and `elapsed_time = '%3.2f%s' % (second * factor, unit)` anytime.

chainer/link_hooks/timer.py
hitsgub: ditto

chainer/link_hooks/timer.py
hitsgub: ditto

chainer/link_hooks/timer.py
hitsgub: ditto

beam2d:
Jenkins, test this please.

beam2d:
LGTM!

beam2d:
Thank you!

PR #6255, opened 2019-02-16
Improve errors for disabled `Variable` operators

niboshi:
* Without messages, it's difficult to understand the cause.
* Python [officially suggests](https://docs.python.org/3/library/exceptions.html#NotImplementedError) that these methods should not be defined if not supported, but I think the built-in message is not informative enough: ```TypeError: '<' not supported between instances of 'Variable' and 'int'```
* Changed `NotImplementedError` to `TypeError` to match Python's built-in behavior.

#### Previously
```
NotImplementedError
```
#### This PR
```
TypeError: Variables do not support < operator. You could use `array` attribute instead.
```

### DIFF
--- a/chainer/variable.py
+++ b/chainer/variable.py
@@ -1478,37 +1478,46 @@ class Variable(object):
                 'method.')
         self._node.data = self._data[0]
 
+    def _error_nobp_op(self, op):
+        raise TypeError(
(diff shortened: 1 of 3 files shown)


okuta:
jenkins, test this please.

okuta:
LGTM!

PR #6261, opened 2019-02-17
Fix `backward` from a variable with `grad=None`

### LINKED ISSUE #6260: Backprop from a variable without grad causes backward() with all-None output_grads
Version: current master (d7adbfb45d46fafa6bd0bbb9f0cebe5894489437)

#### Code:
```py
import numpy
import chainer

a = numpy.ones((2, 3), numpy.float32)
x = chainer.Variable(a)
y = F.sin(x)
y.backward()
```

#### Result:
```
  File "/home/niboshi/t/b.py", line 8, in <module>
    y.backward()
  File "/data2/work/w/repos/chainer/chainer/variable.py", line 1400, in backward
  File "/data2/work/w/repos/chainer/chainer/variable.py", line 1591, in _backprop_to_all
    func, target_input_indexes, out_grad, in_grad)
  File "/data2/work/w/repos/chainer/chainer/_backprop_utils.py", line 116, in backprop_step
    target_input_indexes, grad_outputs)
  File "/data2/work/w/repos/chainer/chainer/functions/math/trigonometric.py", line 28, in backward
    return SinGrad().apply((x, grad_outputs[0]))
  File "/data2/work/w/repos/chainer/chainer/function_node.py", line 321, in apply
  File "/data2/work/w/repos/chainer/chainer/function_node.py", line 513, in forward
    return self.forward_cpu(inputs)
  File "/data2/work/w/repos/chainer/chainer/functions/math/trigonometric.py", line 37, in forward_cpu
    gx *= gy
TypeError: ufunc 'multiply' output (typecode 'O') could not be coerced to provided output parameter (typecode 'f') according to the casting rule ''same_kind''
```
`gy` at the last stack frame is `None`.

niboshi:
Fixes #6260


chainer/variable.py
```diff
@@ -1561,6 +1561,8 @@ def add_cand(cand):
         out_grad = tuple([grads.pop(y) for y in outputs])
         if not target_input_indexes:
             continue
+        if all([gy is None for gy in out_grad]):
```
okuta:
```suggestion
        for gy in out_grad:
            if gy is not None:
                break
        else:
```

toslunar:
Please explain the difference between
- the PR and 
- adding a check `y.grad is None` in `y.backward`.

niboshi:
> - adding a check `y.grad is None` in `y.backward`.

I didn't come up with this solution.
That may be okay but I think this PR's way is more straightfoward.
Even if we go the latter, I feel I want an assertion (`assert not all([gy is None for gy in out_grad])`) in this location.

Issue #6263, opened 2019-02-17
Backprop performance degradation


niboshi:
I found that backprop overhead is increasing lately.
Out of curiosity I measured backprop time in the past versions.

**(unit: ms)**

| | v1 | v2 | v3 | v4 | v5 | v6 |
|-|-|-|-|-|-|-|
| v1.24.0 | 5.9 |
| v2.1.0 | | 7.1 |
| v3.5.0 | | | 8.6 |
| v4.5.0 | | | | 9.2 |
| v5.0.0 <br /> v6.0.0a1 | | | | | 10.1 | 11.0
| v5.1.0 <br /> v6.0.0b1 | | | | | 11.1 | 12.5
| v5.2.0 <br /> v6.0.0b2 | | | | | 13.8 | 19.2
| v5 e730c2f591bb5d6035b990b8ebf6abfcf432ba48 <br /> master d7adbfb45d46fafa6bd0bbb9f0cebe5894489437 | | | | | 13.6 | 19.1 |

```py
import timeit

setup = '''
import numpy
import chainer

a = numpy.array([1, 2], numpy.float32)
x = chainer.Variable(a, requires_grad=True)
h = x
for i in range(1000):
    h = h + 1
z = h
z_grad = numpy.ones_like(z.array)
'''

stmt = '''
z.grad = z_grad
z.backward()
'''

n = 100
t = timeit.timeit(
    setup=setup,
    stmt=stmt,
    number=n) / n

print("{} us".format(t * 1e6))
```

niboshi:
After applying #6264 and #6213, it becomes 14.2ms (cf. 19.1ms in master d7adbfb45d46fafa6bd0bbb9f0cebe5894489437)

niboshi:
It's not clear whether this overhead is actually critical for real use cases.
I'll close this issue.

PR #6264, opened 2019-02-17
Avoid `contextmanager` in backprop

niboshi:
This improves the performance of backprop.

19.1ms -> 16.3ms (using the same measurement as #6263)


okuta:
jenkins, test this please.

okuta:
LGTM!

PR #6271, opened 2019-02-17
Cosmetic fix to `test_gradient_check`

niboshi:
Removes unnecessary call redirection.


okuta:
jenkins, test this please.

niboshi:
@okuta `cat:` labels are mutually exclusive (because they're used for release notes), and I think `cat:code-fix` is more appropriate because this PR does not change the behavior of the test.

okuta:
LGTM!

PR #6272, opened 2019-02-17
Improve performance of `is_arrays_compatible`

niboshi:
For single `numpy.ndarray` argument: 1101 ns -> 654 ns per call.
For three `numpy.ndarray` arguments: 1299 ns -> 820 ns per call.

The performance of backprop (#6263) slightly improves from 19.1 ms -> 18.4 ms.


chainer/__init__.py
```diff
@@ -160,25 +160,60 @@ def get_cpu_array_types():
...
+            has_ideep = has_ideep or isinstance(a, backends.intel64.mdarray)
+            if has_chainerx and has_ideep:
+                # chainerx arrays and intel64.mdarray are not compatible.
+                return False
+
+        if not has_chainerx:
+            # `types` is only initialized when needed.
+            # (i.e. at the first occurrence of non-None, non-ChainerX inputs)
+            if types is None:
+                if isinstance(a, backends.cuda.ndarray):
```
okuta: How about following code?
```
types = backends.cuda.ndarray
if not isinstance(a, types):
    types = get_cpu_array_types()
    if not isinstance(a, types):
        return False
```

chainer/__init__.py
```diff
@@ -160,25 +160,60 @@ def get_cpu_array_types():
...
+                        # TODO(niboshi): Perhaps better to make it an error but
+                        # in order to do that we won't be able to early-return
+                        # False from within the loop.
+                        return False
+
+            elif not isinstance(a, types):
+                # cpu/gpu arrays are mixed.
+                has_mixed_cpu_gpu = True
+
+    if not has_chainerx and has_mixed_cpu_gpu:
```
okuta:
```suggestion
    return has_chainerx or not has_mixed_cpu_gpu
```

niboshi: Why do you want this fix?
I want to leave the comment in the current code, and the code is more straightforward considering the comment. This condition can be considered as a kind of early-return case.

chainer/__init__.py
niboshi: Fixed.

toslunar:
The PR will conflict with #7520.

PR #6273, opened 2019-02-17
Use `continue` to save an indentation level

niboshi:
Just a cosmetic fix for readability.


niboshi:
Being fixed in #6213.

Issue #6275, opened 2019-02-18
cpplint is failing


hvy:
cpplint is failing with the recent updated https://github.com/cpplint/cpplint/releases which is affecting CI.

hvy:
Fixed by https://github.com/chainer/chainer/pull/6267.

PR #6267, opened 2019-02-17
Fix clang-tidy errors

niboshi:

PR #6276, opened 2019-02-18
Fix missing includes

hvy:
Fixes https://github.com/chainer/chainer/issues/6275.


niboshi:
This duplicates #6267.

hvy:
Ah thanks. I didn't find it since I was searching for "cpplint".

PR #6284, opened 2019-02-18
Unwrap ChainerX connected array from `Variable`

### LINKED ISSUE #6279: Get chainerx.ndarray with nodes from Variable
Support optimized backproppable calculation by unwrapping chainerx variables and directly manipulate arrays.

niboshi:
Fixes #6279 
Introduces `Variable.chx_view()` which returns a view (`chx.ndarray`) of the original array.

### DIFF
--- a/chainer/variable.py
+++ b/chainer/variable.py
@@ -868,6 +868,25 @@ class Variable(object):
             self._has_chainerx_array = False
 
     @property
+    def chainerx_array(self):
+        """A view of the raw ChainerX array.
+
+        In contrary to :data:`Variable.array` which is always disconnected,
+        the array represented by this attribute may be connected to the
+        computational graph.
+
+        It is a view, so it has a distinct gradient from the original array.
+
+        If this attribute is queried on a :class:`Variable` with a non-ChainerX
+        array, :class:`ValueError` will be raised.
+        """
+        if not self._has_chainerx_array:
+            raise ValueError(
+                'chainerx_array is not available for Variable with '
+                'non-ChainerX array.')
+        return self._data[0].view()
+
+    @property
     def data(self):
         # type: () -> tp.Optional[types.NdArray]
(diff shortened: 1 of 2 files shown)


chainer/variable.py
```diff
@@ -867,6 +867,24 @@ def array(self, d):
             self._data[0] = d
             self._has_chainerx_array = False
 
+    def chx_view(self):
```
hvy: Could you clarify why it's returning a view?

niboshi: If it is not, `var.chx_view().require_grad()` would affect the original array, making `var` in an invalid state.

hvy: Thanks for the explanation.

hvy: How about not abbreviating chainerx -> chx and name it `chainerx_view`? This is more consistent with other APIs.

niboshi: Changed to `chainex_array` after discussion. We may consider changing `chainerx` to `chx` throughout the Chainer code afterwards (e.g. `to_chainerx` -> `to_chx`).

hvy:
jenkins, test this please.

tests/chainer_tests/test_variable.py
```diff
@@ -2968,4 +2968,84 @@ def test_backward_cpu_lazy_grad_sum(self):
...
+        else:
+            y_arr.backward()
+
+        # x.grad is set
+        assert x.grad is not None
+        chainerx.testing.assert_array_equal_ex(
+            chainerx.array([2, 4], np.float32), x.grad)
+
+
+class TestVariableChainerxArrayView(unittest.TestCase):
```
hvy: We need the ChainerX attribute `@attr.chainerx` here.

hvy:
jenkins, test this please.

hvy:

PR #6285, opened 2019-02-18
Use some latest features for the WaveNet example

mitmul:
Replace Convolution2D with (k, 1)-size kernels with Convolution1D, etc.
And rename `__call__` methods to `forward`.


toslunar:
Are there other features to be used in the example?

toslunar:
Jenkins & flexCI, test this please.

PR #6289, opened 2019-02-18
Improve hook handling in backward

okuta:
I use #6263 

before
22.832 ms

after
21.415 ms

### DIFF
--- a/chainer/function_node.py
+++ b/chainer/function_node.py
@@ -1089,6 +1089,7 @@ def _backprop(outputs, inputs, grad_required, retain_grad, grads, loss_scale):
     ret_dict = {}
 
     is_debug = chainer.is_debug()
+    base_hooks = chainer.get_function_hooks().values()
     while candidate_funcs:
         func = pop_candidate()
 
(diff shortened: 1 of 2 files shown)


okuta:
jenkins, test this please.

chainer/function_node.py
```diff
@@ -1122,25 +1123,27 @@ def _backprop(outputs, inputs, grad_required, retain_grad, grads, loss_scale):
...
+            local_hooks = collections.OrderedDict(chainer.get_function_hooks())
+            local_hooks.update(func.local_function_hooks)
+            hooks = local_hooks.values()  # avoid six for performance
+        else:
+            hooks = base_hooks
 
-        in_data = tuple([x.data for x in func.inputs])
-        out_grad_data = tuple(
-            [None if g is None else g.data for g in gys])
+        in_data = [x.data for x in func.inputs]
```
mitmul: Is it OK not to use `array` attribute instead of `data`?

chainer/function_node.py
```diff
@@ -1122,25 +1123,27 @@ def _backprop(outputs, inputs, grad_required, retain_grad, grads, loss_scale):
...
+        out_grad_data = [None if g is None else g.data for g in gys]
```
mitmul: Ditto.

chainer/variable.py
```diff
@@ -1564,18 +1565,19 @@ def add_cand(cand):
         if not target_input_indexes:
             continue
 
-        in_data = tuple([x.data for x in inputs])
-        out_grad_array = tuple(
-            [None if g is None else g.array for g in out_grad])
-        hooks = chainer.get_function_hooks()
+        in_data = [x.data for x in inputs]
```
mitmul: Ditto.

chainer/function_node.py
toslunar: `x` is an instance of `VariableNode`, which only has `data`.

okuta:
jenkins, test this please.

chainer/function_node.py
mitmul: Oops, got it.

mitmul:
The error failed in the Jenkins jobs are irrelevant to this PR.

PR #6292, opened 2019-02-18
Improve performance of `using_device` for NumPy and Intel64 devices

niboshi:
1.86 us -> 0.54 us (only if `device` is either `CpuDevice` or `Intel64Device`).
```py
with chainer.using_device(device):
    pass
```


okuta:
jenkins, test this please.

okuta:
LGTM!

PR #6295, opened 2019-02-19
add Kuzushiji-MNIST dataset

wintercarver:
I'd like to suggest adding the Kuzushiji-MNIST dataset to Chainer.

The Kuzushiji-MNIST dataset is designed to be an
MNIST-like dataset focusing on handwritten Japanese
characters. All data features of KMNIST are identical to
MNIST, such as number of train and test images, image
size, etc with the intent of making it easy to run any
MNIST-based analysis or benchmarks against this new
dataset.

Details available here: https://github.com/rois-codh/kmnist

### DIFF
--- a/chainer/datasets/kmnist.py
+++ b/chainer/datasets/kmnist.py
@@ -1,0 +1,98 @@
+import os
+
+import numpy
+
+import chainer
+from chainer.dataset import download
+from chainer.datasets._mnist_helper import make_npz
+from chainer.datasets._mnist_helper import preprocess_mnist
+
+
+_kmnist_labels = [('o', u'\u304A'), ('ki', u'\u304D'),
+                  ('su', u'\u3059'), ('tsu', u'\u3064'),
+                  ('na', u'\u306A'), ('ha', u'\u306F'),
+                  ('ma', u'\u307E'), ('ya', u'\u3084'),
+                  ('re', u'\u308C'), ('wo', u'\u3092')]
+
+
+def get_kmnist_labels():
+    """Provide a list of labels for the KMNIST dataset.
+
+    Returns:
+        List of labels in the form of tuples. Each tuple contains the 
+        character name in romaji as a string value and the unicode codepoint 
+        for the character.
+
+    """
+    return list(_kmnist_labels)
+
+
+def get_kmnist(withlabel=True, ndim=1, scale=1., dtype=None,
+               label_dtype=numpy.int32, rgb_format=False):
+    """Gets the KMNIST dataset.
+
+    `Kuzushiji-MNIST (KMNIST) <http://codh.rois.ac.jp/kmnist/>`_ is a set of
+    hand-written Japanese characters represented by grey-scale 28x28 images.
+    In the original images, each pixel is represented by one-byte unsigned
+    integer. This function scales the pixels to floating point values in the
+    interval ``[0, scale]``.
+
+    This function returns the training set and the test set of the official
+    KMNIST dataset. If ``withlabel`` is ``True``, each dataset consists of
+    tuples of images and labels, otherwise it only consists of images.
+
+    Args:
+        withlabel (bool): If ``True``, it returns datasets with labels. In this
+            case, each example is a tuple of an image and a label. Otherwise,
+            the datasets only contain images.
+        ndim (int): Number of dimensions of each image. The shape of each image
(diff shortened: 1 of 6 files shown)


hvy:
Jenkins, test this please.

hvy:
I'm wondering if we should use the more verbose "kuzushiji_mnist" instead of "kmnist" if we we're to provide "kuzushiji_49" and "kuzushiji_kanji" in the future, for consistency that is. Another argument for this would be that "fashion_mnist" already follows this naming convention. @wintercarver what do you think?

chainer/datasets/kmnist.py
```diff
@@ -0,0 +1,98 @@
...
+def get_kmnist_labels():
+    """Provide a list of labels for the KMNIST dataset.
+
+    Returns:
+        List of labels in the form of tuples. Each tuple contains the
+        character name in romaji as a string value and the unicode codepoint
+        for the character.
+
+    """
+    return list(_kmnist_labels)
```
hvy: I think it's okay to skip this list casting, since it's already a list.

wintercarver: > I'm wondering if we should use the more verbose "kuzushiji_mnist" instead of "kmnist" if we we're to provide "kuzushiji_49" and "kuzushiji_kanji" in the future, for consistency that is. Another argument for this would be that "fashion_mnist" already follows this naming convention. @wintercarver what do you think?

I am OK with this. I actually started with "kuzushiji_" on everything but it felt a little long, however it would be more clear and "k" is not a very informative prefix. If you agree, I'll make the changes of `kmnist` -> `kuzushiji_mnist`.

chainer/datasets/kmnist.py
```diff
@@ -0,0 +1,98 @@
...
+
+_kmnist_labels = [('o', u'\u304A'), ('ki', u'\u304D'),
+                  ('su', u'\u3059'), ('tsu', u'\u3064'),
+                  ('na', u'\u306A'), ('ha', u'\u306F'),
+                  ('ma', u'\u307E'), ('ya', u'\u3084'),
+                  ('re', u'\u308C'), ('wo', u'\u3092')]
+
+
+def get_kmnist_labels():
+    """Provide a list of labels for the KMNIST dataset.
```
hvy: A suggestion, for consistency.
```suggestion
    """Provides a list of labels for the KMNIST dataset.
```

chainer/datasets/kmnist.py
hvy: I do agree!

hvy:
jenkins, test this please.

chainer/datasets/kuzushiji_mnist.py
```diff
@@ -0,0 +1,98 @@
...
+from chainer.dataset import download
+from chainer.datasets._mnist_helper import make_npz
+from chainer.datasets._mnist_helper import preprocess_mnist
+
+
+_kuzushiji_mnist_labels = [('o', u'\u304A'), ('ki', u'\u304D'),
+                           ('su', u'\u3059'), ('tsu', u'\u3064'),
+                           ('na', u'\u306A'), ('ha', u'\u306F'),
+                           ('ma', u'\u307E'), ('ya', u'\u3084'),
+                           ('re', u'\u308C'), ('wo', u'\u3092')]
```
hvy: Could you comment on why the Unicode hiragana should be included (and not just the romaji)?

wintercarver: The unicode codepoints aren't essential, but for those unfamiliar with the Japanese language an easy way to print and refer to the characters might be appreciated (i.e. seeing something classified as 'ma' might not mean much, but being able to see the original character easily could help). What do you think? Is consistent with Fashion MNIST that important in this context?

hvy: I see, thanks for the explanation. So you mean for actual visual verification? Consistency matters simply in terms of how much the users has to rewrite when switching from Fashion MNIST to Kuzushiji MNIST. Maybe this is not that much of an issue, regarding the fact that labels are mostly meant for practical visualizations in the first place.

hvy: Let's proceed with this PR as is.

hvy:
jenkins, test this please.

hvy:
LGTM!

PR #6304, opened 2019-02-19
Forbid calling empty `Sequential`

### LINKED ISSUE #6303: Empty Sequential has different semantics than other cases
As you can see from below, an empty `chainer.Sequential()` has different semantics: it wraps an argument with a tuple and return it. If at least one built-in links or functions exist, it returns a single `Variable` object.

I think it's better to forbid calling an empty `Sequential`.

#### Code
```py
import numpy
import chainer

x = numpy.array([1, 2], numpy.float32)
seq = chainer.Sequential()
ret = seq(x)

print(ret)
print(type(ret))
print(type(ret[0]))
print()

seq = chainer.Sequential(F.tanh)
ret = seq(x)
print(type(ret))
```
#### Result
```
(array([1., 2.], dtype=float32),)
<class 'tuple'>
<class 'numpy.ndarray'>

<class 'chainer.variable.Variable'>
```

niboshi:
Fixes #6303

### DIFF
--- a/chainer/sequential.py
+++ b/chainer/sequential.py
@@ -205,6 +205,9 @@ class Sequential(_link.ChainList):
             The output of the final layer in the given layers.
 
         """
+        if len(self._layers) == 0:
+            raise RuntimeError('Sequential does not have any layer.')
+
         for layer in self._layers:
             if isinstance(x, tuple):
                 x = layer(*x)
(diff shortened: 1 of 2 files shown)


okuta:
jenkins, test this please.

niboshi:
Sorry, I forgot `Test` prefix (it seems it doesn't matter, though).
Please retrigger jenkins.

okuta:
jenkins, test this please.

okuta:
jenkins, test this please.

okuta:
LGTM!

PR #6307, opened 2019-02-19
Simplify `F.sum` test

aksub99:
Simplified F.sum test using testing.FunctionTestCase.
Ref. #6071

### DIFF
--- a/tests/chainer_tests/functions_tests/math_tests/test_sum.py
+++ b/tests/chainer_tests/functions_tests/math_tests/test_sum.py
@@ -45,16 +45,10 @@ class TestSum(testing.FunctionTestCase):
 
     def forward(self, inputs, device):
         x, = inputs
-        print(x)
-        print("Forward:")
-        print(functions.sum(x, axis=self.axis, keepdims=self.keepdims))
         return functions.sum(x, axis=self.axis, keepdims=self.keepdims),
 
     def forward_expected(self, inputs):
         x, = inputs
-        print(x)
-        print("Expected:")
-        print(x.sum(axis=self.axis, keepdims=self.keepdims))
         expected = x.sum(axis=self.axis, keepdims=self.keepdims)
         expected = numpy.asarray(expected)
         return expected,    
--- a/tests/chainer_tests/functions_tests/math_tests/test_sum.py
+++ b/tests/chainer_tests/functions_tests/math_tests/test_sum.py
@@ -2,14 +2,8 @@ import unittest
 
 import numpy
 
-import chainer
-from chainer.backends import cuda
 from chainer import functions
-from chainer import gradient_check
 from chainer import testing
-from chainer.testing import attr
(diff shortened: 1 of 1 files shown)


aksub99:
@Crissman The outputs of forward() and forward_expected() seem to differ by a small amount, therefore causing the error. Could you please review the code and suggest why this might be so?

tests/chainer_tests/functions_tests/math_tests/test_sum.py
```diff
@@ -17,58 +17,41 @@
...
+        'use_cudnn': ['never', 'always'],
+        'cuda_device': [0, 1],
+    })
+    # ChainerX tests
+    + [
+        {'use_chainerx': True, 'chainerx_device': 'native:0'},
+        {'use_chainerx': True, 'chainerx_device': 'cuda:0'},
+        {'use_chainerx': True, 'chainerx_device': 'cuda:1'},
+    ])
+class TestSum(testing.FunctionTestCase):
```
crcrpar: It seems that this class lacks the relaxation of tolerance when self.dtype == numpy.float16.
I think this absence of tolerance for float16 relates to failures.

aksub99: Thanks @crcrpar . That managed to fix it !

aksub99:
The failures seem to be caused due to `test_average_pooling_nd` . 
@Crissman @crcrpar  Please review and confirm. Thanks in advance!

tests/chainer_tests/functions_tests/math_tests/test_sum.py
```diff
@@ -2,73 +2,57 @@
...
     'dtype': [numpy.float16, numpy.float32, numpy.float64],
 }))
-class TestSum(unittest.TestCase):
+@testing.fix_random()
+@testing.inject_backend_tests(
+    None,
+    # CPU tests
+    [
+        {},
+        {'use_ideep': 'always'},
```
crcrpar: It doesn't seem that [`F.sum`](https://github.com/chainer/chainer/blob/master/chainer/functions/math/sum.py#L11) use ideep so I don't think this case is necessary.

tests/chainer_tests/functions_tests/math_tests/test_sum.py
```diff
@@ -2,73 +2,57 @@
...
+    ]
+    # GPU tests
+    + testing.product({
+        'use_cuda': [True],
+        'use_cudnn': ['never', 'always'],
```
crcrpar: Like the above case, It doesn't seem that cudnn is used in `F.sum`.

aksub99: Thanks for reviewing @crcrpar ! Have made the required changes.

Crissman:

Crissman:
Merging!

PR #6317, opened 2019-02-20
Fix to clarify the description about initializer argument

okapies:
The `initializer` argument of the functions accept a scalar value, but it is not explicitly documented in some places. This patch fixes these docstrings.


niboshi:
``` :ref:`initializer <initializer>` ```  was introduced for this exact reason: to avoid repeating the same explanation everywhere.

okapies:
Thanks. I'll replace them with it.

niboshi:
Jenkins, test this please

niboshi:
LGTM.
I'm sorry for being late to merge.

PR #6324, opened 2019-02-20
Add `distributions.Independent`

ganow:
Typical DL frameworks which handle probabilistic distributions such as Tensorflow Probability and PyTorch Distributions have `Independent` distribution class. I implemented the same interface in Chainer.

`Independent` distribution reinterprets some of the batch dims of distribution as event dims. This is mainly useful for changing the shape of the result of `distribution.log_prob(x)`, and it makes code clearer and simpler. For example, the current official example of the VAE's objective function is like

```python
...
q_z = D.Normal(loc=mu(x), log_scale=ln_sigma(x))
z = q_z.sample(k)
p_x = D.Bernoulli(logit=h(z))
p_z = D.Normal(zeros_array, ones_array)

reconstr = F.mean(F.sum(p_x.log_prob(
    F.broadcast_to(x[None, :], (k,) + x.shape)), axis=-1))
kl_penalty = F.mean(F.sum(chainer.kl_divergence(q_z, p_z), axis=-1))
...
```

However, this code assumes the last dimension of `p_x`, `q_z`, and `p_z` should be treated as event dim. This code will not work when we change `q_z` from `D.Normal` to `D.MultivariateNormal`. By using `D.Independent`, we can rewrite the above code as

```python
...
q_z = D.Independent(D.Normal(loc=mu(x), log_scale=ln_sigma(x)))
z = q_z.sample(k)
p_x = D.Independent(D.Bernoulli(logit=h(z)), reinterpreted_batch_ndims=1)
p_z = D.Independent(
    D.Normal(zeros_array, ones_array), reinterpreted_batch_ndims=1)

reconstr = F.mean(p_x.log_prob(
    F.broadcast_to(x[None, :], (self.k,) + x.shape)))
kl_penalty = F.mean(chainer.kl_divergence(q_z, p_z))
...
```

This code is assumption-free, and we can treat distribution.event_shape appropriately.

### DIFF
--- a/chainer/distributions/independent.py
+++ b/chainer/distributions/independent.py
@@ -56,18 +56,20 @@ class Independent(distribution.Distribution):
     def covariance(self):
         '''Returns the covariance of the distribution based on the original
         i.i.d. distribution. By definition, the covariance of the new
-        distribution becomes block diagonal matrix. Let :math:`\\Sigma_x` be
-        the covariance matrix of the original random variable :math:`x
-        \\in R^d`, and :math:`x^{(1)}, x^{(2)}, \\cdots x^{(m)}` be the
-        :math:`m` i.i.d. random variables, new covariance matrix
-        :math:`\\Sigma_y` of :math:`y = [x^{(1)}, x^{(2)}, \\cdots, x^{(m)}]
-        \\in R^{md}` can be written as
+        distribution becomes block diagonal matrix. Let
+        :math:`\\Sigma_{\\mathbf{x}}` be the covariance matrix of the original
+        random variable :math:`\\mathbf{x} \\in \\mathbb{R}^d`, and
+        :math:`\\mathbf{x}^{(1)}, \\mathbf{x}^{(2)}, \\cdots \\mathbf{x}^{(m)}`
+        be the :math:`m` i.i.d. random variables, new covariance matrix
+        :math:`\\Sigma_{\\mathbf{y}}` of :math:`\\mathbf{y} =
+        [\\mathbf{x}^{(1)}, \\mathbf{x}^{(2)}, \\cdots, \\mathbf{x}^{(m)}] \\in
+        \\mathbb{R}^{md}` can be written as
 
         .. math::
             \\left[\\begin{array}{ccc}
-                    \\Sigma_{x^{1}} &         & 0               \\\\
-                                    & \\ddots &                 \\\\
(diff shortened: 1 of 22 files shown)


ganow:
> Could you separate a PR for adding missing `params` attributes ([6f0f368](https://github.com/chainer/chainer/commit/6f0f368955048a2de1ce75efbedace2df4dd0835))?

`Independent` class needs `xp` to compute `covariance` property, and `distribution.xp` refers `distribution.params`. which means that this implementation will not work without `params` property for some distributions. Should I split the PR even though?

chainer/distributions/independent.py
```diff
@@ -0,0 +1,245 @@
+from functools import reduce
```
toslunar: Chainer follows OpenStack Style Guidelines.

> Do not import objects, only modules

https://docs.openstack.org/hacking/latest/user/hacking.html#imports

chainer/distributions/independent.py
```diff
@@ -0,0 +1,245 @@
...
+import operator
+
+import numpy
+
+from chainer.backend import cuda
+from chainer import distribution
+from chainer.functions.array import reshape, repeat, transpose
```
toslunar: > [H301] Do not import more than one module per line

ganow: thank you for checking code style.

chainer/distributions/independent.py
```diff
@@ -0,0 +1,247 @@
...
+            num_repeat, axis=-4)
+        cov = reshape.reshape(
+            transpose.transpose(
+                cov, axes=(
+                    tuple(range(len(self.batch_shape))) + (-4, -2, -3, -1))),
+            self.batch_shape + (num_repeat * dim, num_repeat * dim))
+        block_indicator = self.xp.reshape(
+            self._block_indicator,
+            tuple([1] * len(self.batch_shape)) + self._block_indicator.shape)
+        return cov * block_indicator
```
toslunar: To make a block diagonal matrix, the backward of `F.diagonal` (or the backward of unary `F.einsum`) could be used.

https://github.com/chainer/chainer/blob/c8c9ee0d5cdad4726f01e6a354f775c555f74355/tests/chainer_tests/functions_tests/math_tests/test_einsum.py#L232-L254

toslunar:
Jenkins, test this please.

PR #5160, opened 2018-07-27
Add Binomial distribution

YoshikawaMasashi:
split from #4678 .

PR #6326, opened 2019-02-21
Adding binomial distribution to Chainer

hsezhiyan:
Adding binomial distribution to chainer.distributions.

I also created a test for this in chainer/tests/chainer_tests/distributions_tests/test_binomial.py

For some reason, the sample test within my test file gives a type casting issue. However, I verified this distribution is working by calling the distribution from another file and observing the result. 

Ex. 
dist = chainer.distributions.Binomial(n, p)
dist.sample_n(20) #this works as intended

If you can help me fix the test issue, it would be great, but it seems the distributions is working.

### DIFF
--- a/chainer/distributions/__init__.py
+++ b/chainer/distributions/__init__.py
@@ -2,6 +2,7 @@
 
 from chainer.distributions.bernoulli import Bernoulli  # NOQA
 from chainer.distributions.beta import Beta  # NOQA
+from chainer.distributions.binomial import Binomial # NOQA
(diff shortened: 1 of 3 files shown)


crcrpar:
This seems to be a duplicate of https://github.com/chainer/chainer/pull/5160.

hvy:
@hsezhiyan thanks for your PR, although it ended up being a duplicate. In the future, you could ask in our Slack channels for missing features. If you don't have an account already, there's a link in the [README.md](https://github.com/chainer/chainer/blob/master/README.md).

Issue #6198, opened 2019-02-08
FP16'ize MultiNodeBatchNormalization


kuenishi:
Subtask of #6161 .

PR #6331, opened 2019-02-21
Support for fp16 in MultiNodeBatchNormalization

shu65:
Solve https://github.com/chainer/chainer/issues/6198
This PR should be merged after https://github.com/chainer/chainer/pull/6333


shu65:
I will make another PR to support for mixed16

PR #6337, opened 2019-02-21
Dynamic loss scaling

anaruse:
This PR aims to support "dynamic" loss scaling.

Loss scaling is already available with current Chainer, but it is "static" loss scaling. You have to select an appropriate loss scaling factor before training, hence becoming kind of hyper parameter.

With this PR, there is no need to select scaling factor in advance. The appropriate scaling factor is dynamically determined by just adding following line to your training script.

    optimizer.loss_scaling('dynamic', interval=1000)

When "dynamic" loss scaling is enabled, whole grads of parameters are checked every iteration, more specifically after backward() and before update() to see if there are NaNs in the grads. Initially the scaling factor is set to 1 and it will be doubled every iteration as long as there is no NaN in the grads. Once NaN is detected in the grads, the scaling factor will get halved. After that, the scaling factor is slowly increased every iteration so that it gets doubled in the specified "interval".

"Static" loss scaling is also supported and can be used as follows.

    optimizer.loss_scaling('static', loss_scale=100.0)

Considering compatibility, `optimizer.use_loss_scale(100.0)` is also available.

### DIFF
--- a/chainer/optimizer.py
+++ b/chainer/optimizer.py
@@ -681,6 +681,33 @@ class Optimizer(object):
         """Sets loss scaling factor."""
         self.loss_scaling('static', loss_scale)
 
+    def is_safe_to_update(self):
+        if self._loss_scaling_mode is not 'dynamic':
+            return True
+        is_safe = True
+        for name, param in self.target.namedparams():
+            xp = param.device.xp
+            if not xp.all(xp.isfinite(param.grad)):
+                is_safe = False
+                self._loss_scaling_isnan_ever = True
+                warnings.warn(
+                    'Non finite number found in pram.grad of {}'
+                    ' (t:{}, loss_scale:{})'
+                    ''.format(name, self.t, self._loss_scale))
+        if not is_safe:
+            multiplier = 0.5
+        elif self._loss_scaling_isnan_ever:
+            multiplier = self._loss_scaling_multiplier
+        else:
+            multiplier = 2.0
+        self._loss_scale_next = max(1, min(self._loss_scale_max,
+                                           self._loss_scale * multiplier))
+        return is_safe
+
+    def update_loss_scale(self):
+        if self._loss_scaling_mode is 'dynamic':
+            self._loss_scale = self._loss_scale_next
+
 
 class GradientMethod(Optimizer):
     """Base class of all single gradient-based optimizers.
@@ -779,28 +806,7 @@ class GradientMethod(Optimizer):
         self.call_hooks('pre')
 
         self.t += 1
-
-        isnan = False
-        ls_multiplier = None
-        if self._loss_scaling_mode is 'dynamic':
-            for name, param in self.target.namedparams():
-                xp = param.device.xp
-                if xp.all(xp.isfinite(param.grad)):
-                    continue
-                isnan = True
-                warnings.warn('Non finite number found in grad of {}'
-                              ' (t:{}, loss_scale:{})'
-                              ''.format(name, self.t, self._loss_scale))
-            if isnan:
-                self._loss_scaling_isnan_ever = True
-                ls_multiplier = 0.5
-            else:
-                if self._loss_scaling_isnan_ever:
-                    ls_multiplier = self._loss_scaling_multiplier
-                else:
-                    ls_multiplier = 2.0
-
-        if not isnan:
+        if self.is_safe_to_update():
             for param in self.target.params():
                 param.update()
 
@@ -808,9 +814,7 @@ class GradientMethod(Optimizer):
 
         self.call_hooks('post')
 
-        if ls_multiplier is not None:
-            self._loss_scale = max(1.0, min(self._loss_scale_max,
-                                            self._loss_scale * ls_multiplier))
+        self.update_loss_scale()
 
     def use_cleargrads(self, use=True):
         """Enables or disables use of :func:`~chainer.Link.cleargrads` in `update`.
--- a/chainer/optimizer.py
+++ b/chainer/optimizer.py
@@ -457,8 +457,8 @@ class Optimizer(object):
     _pre_update_hooks = None
     _post_update_hooks = None
     _loss_scale = None
-    _loss_scaling_mode = None
     _loss_scale_max = 65504  # max representable value with fp16
+    _loss_scaling_is_dynamic = False
     use_auto_new_epoch = False
 
     def setup(self, link):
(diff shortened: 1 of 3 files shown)


beam2d:
Thanks for the PR! I have an idea for simplifying the interface: making the mode inferred from the option specified by the user. For example, `loss_scaling(interval=1000)` means dynamic scaling with interval 1000, while `loss_scaling(scale=100.0)` means static scaling with scale 100.0. Another idea would be to separate the method for each mode, e.g. `loss_scaling` for dynamic one and `static_loss_scaling` for static one. I prefer the former for its simplicity.

chainer/optimizer.py
```diff
@@ -652,9 +655,58 @@ def serialize(self, serializer):
...
+        if self._loss_scaling_mode is not 'dynamic':
+            return True
+        is_safe = True
+        for name, param in self.target.namedparams():
+            xp = param.device.xp
+            if not xp.all(xp.isfinite(param.grad)):
+                is_safe = False
+                self._loss_scaling_isnan_ever = True
+                warnings.warn(
+                    'Non finite number found in pram.grad of {}'
```
hvy: Ops a typo "pram".

hvy:
Sorry for the late comment. I think we want an additional argument to `loss_scaling(interval=1000)`, to specify its initial scale. How about `loss_scaling(scale=100.0, interval=1000)`, i.e. it turns dynamic when `interval` is specified? I am not too familiar with this technique though and it might be less obvious if we'd want the dynamic behavior as the default behavior (you'd probably have to pass `interval=None`).

anaruse:
I've just updated the PR based on the off-line discussion yesterday.
Please review it again.

chainer/optimizer.py
```diff
@@ -655,27 +655,27 @@ def serialize(self, serializer):
...
-            raise ValueError('loss_scale must be positive number')
-        if mode is 'dynamic' and interval < 1:
-            raise ValueError('interval must be greater equal to 1')
-
-        self._loss_scaling_mode = mode
-        self._loss_scale = loss_scale
-        self._loss_scaling_multiplier = math.pow(2.0, 1.0 / interval)
-        self._loss_scaling_isnan_ever = False
+        if scale is None:
+            self._loss_scaling_mode = 'dynamic'
```
hvy: Is it preferably to store this variable as a string and not a boolean flag such as `_loss_scaling_is_dynamic`?

chainer/optimizer.py
```diff
@@ -652,9 +655,58 @@ def serialize(self, serializer):
             if rule is not None:
                 rule.serialize(serializer[name])
 
+    def loss_scaling(self, interval=1000, scale=None):
+        """Sets loss scaling configurations.
+
+        Args:
+            interval (integer): Number of iterations until scaling factor gets
+                doubled. This is effective when "dynamic" loss scaling is used.
+            scale (float): Loss scaling factor. If None, "dynamic" loss scaling
```
hvy: A minor but how about the following (I'm not sure if this exceeds the maximum line length)?
```suggestion
            scale (float): Loss scaling factor. If ``None``, "dynamic" loss scaling
```

chainer/optimizer.py
```diff
@@ -652,9 +655,58 @@ def serialize(self, serializer):
             if rule is not None:
                 rule.serialize(serializer[name])
 
+    def loss_scaling(self, interval=1000, scale=None):
+        """Sets loss scaling configurations.
```
hvy: A minor but how about
```suggestion
        """Configures the loss scaling algorithm.
```

chainer/optimizer.py
```diff
@@ -652,9 +655,58 @@ def serialize(self, serializer):
...
+
+        Args:
+            interval (integer): Number of iterations until scaling factor gets
```
hvy: Sorry, again a minor.
```suggestion
            interval (int): Number of iterations until scaling factor gets
```

chainer/optimizer.py
```diff
@@ -652,9 +655,58 @@ def serialize(self, serializer):
...
+                doubled. This is effective when "dynamic" loss scaling is used.
+            scale (float): Loss scaling factor. If None, "dynamic" loss scaling
+                is used, otherwise "static" loss scaling is used.
+        """
+        if scale is None:
+            self._loss_scaling_mode = 'dynamic'
+            if interval < 1:
+                raise ValueError("interval must be greater equal to 1")
```
hvy: How about showing the actual value, e.g.
```suggestion
                raise ValueError("interval must be greater equal to 1. Actual: {}".format(interval))
```

chainer/optimizer.py
```diff
@@ -652,9 +655,58 @@ def serialize(self, serializer):
...
+            self._loss_scale = 1.0
+            self._loss_scaling_multiplier = math.pow(2.0, 1.0 / interval)
+            self._loss_scaling_isnan_ever = False
+        else:
+            self._loss_scaling_mode = 'static'
+            if scale <= 0:
+                raise ValueError("loss_scale must be positive number")
```
hvy: Could you stick to single quoted strings (except method/function documentation)?

hvy: Ditto regarding including the actual value in the error message.

chainer/optimizer.py
```diff
@@ -652,9 +655,58 @@ def serialize(self, serializer):
...
+            self._loss_scale = scale
+
     def set_loss_scale(self, loss_scale):
         """Sets loss scaling factor."""
-        self._loss_scale = loss_scale
+        self.loss_scaling('static', loss_scale)
```
hvy: Seems like it's passing a string according to the old version of `loss_scaling`. Could you change it according to the new version?

chainer/optimizer.py
```diff
@@ -652,9 +655,58 @@ def serialize(self, serializer):
...
+            return True
+        is_safe = True
+        for name, param in self.target.namedparams():
+            xp = param.device.xp
+            if not xp.all(xp.isfinite(param.grad)):
+                is_safe = False
+                self._loss_scaling_isnan_ever = True
+                warnings.warn(
+                    'Non finite number found in param.grad of {}'
+                    ' (t:{}, loss_scale:{})'
```
hvy: Is the user familiar with "t"?

chainer/optimizer.py
```diff
@@ -652,9 +655,58 @@ def serialize(self, serializer):
...
+                    ''.format(name, self.t, self._loss_scale))
```
hvy: Should you not break from the loop after this line in favor of performance?

chainer/optimizer.py
```diff
@@ -754,13 +806,16 @@ def update(self, lossfun=None, *args, **kwds):
         self.call_hooks('pre')
 
         self.t += 1
-        for param in self.target.params():
-            param.update()
+        if self.is_safe_to_update():
```
hvy: I get the impression from this line that the call only "checks if it is safe to update", but it actually does a lot more. E.g. `update_loss_scale` depends on this method being called before. Is it difficult to simplify the logic and decouple these methods?

chainer/optimizer.py
anaruse: Ah, sorry, I forgot to change that.

chainer/optimizer.py
anaruse: a string or a boolean, either would be fine with me. Would you prefer a boolean?

chainer/optimizer.py
anaruse: Thanks, that's better.

chainer/optimizer.py
anaruse: Probably, no.  It looks "iteration" should be used instead?

chainer/optimizer.py
anaruse: I did not put "break" here intentionally, since I thought it was better to show all layers which have 'NaN' in its grads.

anaruse:
Thank you for your detailed feedback, @hvy. I did code refactoring and clean-up based on your feedback. Could you review it again?

chainer/optimizer.py
hvy: I see, could you elaborate a bit in what sense that'd be useful to the user? I though the performance impact was more relevant here.

chainer/optimizer.py
hvy: Thanks for the change, I think the current boolean is simple and fine. Maybe a bit less prone to errors as well.

chainer/optimizer.py
```diff
@@ -652,9 +655,62 @@ def serialize(self, serializer):
...
+            self._loss_scaling_is_dynamic = True
+            if interval < 1:
+                raise ValueError('interval must be greater equal to 1. '
+                                 'Actual: {}'.format(interval))
+            self._loss_scale = 1.0
+            self._loss_scaling_multiplier = math.pow(2.0, 1.0 / interval)
+            self._loss_scaling_isnan_ever = False
+        else:
+            if scale <= 0:
+                raise ValueError('loss_scale must be positive number. '
```
hvy: A grammatical suggestion.
```suggestion
                raise ValueError('loss_scale must be a positive number. '
```

chainer/optimizer.py
```diff
@@ -652,9 +655,62 @@ def serialize(self, serializer):
...
+        Args:
+            interval (int): Number of iterations until scaling factor gets
+                doubled. This is effective when "dynamic" loss scaling is used.
+            scale (float): Loss scaling factor. If ``None``, "dynamic" loss
+                scaling is used, otherwise "static" loss scaling is used.
+        """
+        if scale is None:
+            self._loss_scaling_is_dynamic = True
+            if interval < 1:
+                raise ValueError('interval must be greater equal to 1. '
```
hvy:
```suggestion
                raise ValueError('interval must be greater than or equal to 1. '
```

chainer/optimizer.py
hvy: Thanks!

chainer/optimizer.py
hvy: Thanks!

chainer/optimizer.py
```diff
@@ -652,9 +655,62 @@ def serialize(self, serializer):
...
+        if not self._loss_scaling_is_dynamic:
+            return
+        for name, param in self.target.namedparams():
+            xp = param.device.xp
+            if not xp.all(xp.isfinite(param.grad)):
+                self._loss_scaling_isnan = True
+                self._loss_scaling_isnan_ever = True
+                warnings.warn(
+                    'Non finite number found in param.grad of {}'
+                    ' (iteration:{}, loss_scale:{})'
```
hvy: A style nitpick.
```suggestion
                    ' (iteration: {}, loss_scale: {})'
```

hvy:
Is it difficult to write tests for these APIs?

chainer/optimizer.py
```diff
@@ -750,16 +806,18 @@ def update(self, lossfun=None, *args, **kwds):
             del loss
 
         self.reallocate_cleared_grads()
-
+        self.check_nan_in_grads()
```
hvy: We can make it even less stateful by returning `self._loss_scaling_isnan` from this function instead of having it as an attribute. It can be passed to `update_loss_scale` and we then we no longer need the `is_safe_to_update` method. What do you say?

hvy: But maybe it's difficult considering the fact that `self_loss_scaling_isnan_ever` is very stateful.

chainer/optimizer.py
anaruse: Actually "break" helps reduce the amount of computation, but it is only when "NaN" is detected in the gradients and it is not a common case.  For example, if the interval is 1000, the probability of detecting NaN is about 0.1%.  Thus, I think it is not so important whether to break the loop or not in terms of performance.

anaruse:
Please wait for a while. I'm now considering how to test this API..

chainer/optimizer.py
hvy: That makes sense, thanks for the explanation. Let's leave it as is.

anaruse:
I've added some basic tests for static/dynamic loss scaling to existing test scripts: `test_optimizers.py` and `test_optimizers_by_linear_model.py`.  Could you review it?

hvy:
jenkins, test this please.

hvy:
LGTM!

PR #6338, opened 2019-02-22
Adding Skellam distribution to Chainer

hsezhiyan:
Adding Skellam distribution to Chainer. This distribution models the difference of two Poisson distributions.


hvy:
@YoshikawaMasashi Are there any policies within the distributions submodule, such as the use of SciPy and supporting various "xp" (NumPy/CuPy)?

YoshikawaMasashi:
@hsezhiyan Thank you for the PR.

@hsezhiyan @hvy 
I apologize for being late.

I think that `mean` and `variance` and `cdf` and `log_cdf` should be differentiable if it is possible.
If not so, document that says this properties don't support differential is needed.

And GPU(cupy) support is needed.

hvy:
Thanks @YoshikawaMasashi for the explanation. I haven't caught up the the `distributions` module yet and your comment helped. 

@hsezhiyan are you able to reflect the comments above to your PR? Feel free to let us know if you're having any issues./

PR #6215, opened 2019-02-12
Remove FP16 specific models from imagenet example

takagi:
This PR removes FP16 specific models from imagenet example as we now have the default dtype feature, which perfectly works to run the example in FP16 mode without any change of the script.

PR #6342, opened 2019-02-22
Fix dali_util in imagenet example for fp16

anaruse:
This PR aims to fix dtype mismatch issue in imagenet example when DALI and fp16 (chainer.config.dtype = float16) are used together.


takagi:
Following #6215.

takagi:
Thanks for the PR, LGTM! Currently we don't have a test of imagenet example with the corresponding options, which I will add shortly as another PR for [chainer/chainer-test](https://github.com/chainer/chainer-test).

PR #5396, opened 2018-09-27
Add Synchronized Iterator

levelfour:
## Overview

In some situations of distributed machine learning, it is preferable to

1. synchronize iterator seed: to preserve batch order among each process.
2. use private random number generator: to avoid interfering other random number generation, such as dropout.

In this PR, I added create_synchronized_iterator, which extends Chainer SerialIterator to synchronize the seed and make private random number generators during initialization.

## Note

This PR is a porting from chainer/chainermn#185.

PR #6345, opened 2019-02-22
Add Synchronized Iterator (Rebased version of #5396)

kuenishi:
As Travis test for #5396 fails with unknown reason I'm reopening with latest master.


kuenishi:
Jenkins, test this please.

PR #6362, opened 2019-02-26
[WIP] Add elish function

aksub99:
Implementation of ELiSH activation function.
Ref. #1448

### DIFF
--- a/chainer/functions/activation/elish.py
+++ b/chainer/functions/activation/elish.py
@@ -1,0 +1,44 @@
+import numpy
+
+from chainer.backends import cuda
+from chainer import function_node
+from chainer import utils
+from chainer.utils import type_check
+from chainer.functions.activation import swish
+
+def _sigmoid(x):
+    half = x.dtype.type(0.5)
+    return numpy.tanh(x * half) * half + half
+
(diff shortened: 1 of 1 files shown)


hvy:
Thanks for this PR. Although it is still in WIP, the necessity of this activation functions a part of the framework is a bit unclear. In particular, I can't find a lot of references to it and the significance is not obvious. Would you mind giving us a bit of background?

aksub99:
Sorry for not stating the reason for the choice earlier @hvy . I assumed that implementing functions from recent research (though not referenced to a lot) might allow more experimentation to be possible using the framework for research purposes by providing some flexibility on choices. Please correct me if this is not one of the intentions. I wouldn't mind working on another function instead :)

hvy:
Thanks for the quick response and the clarification. Providing various functions are indeed what we're aiming for but there's for instance a cost in maintaining public APIs. The consensus within the team is that this functions does not seem significant enough. We could be wrong of course, and don't want any discouragements. Maybe it can be reconsidered if it's being developed outside the framework (as a third party repository for instance) and it looks promising. Please let me close this ticket for now but you're very welcome to work on a different function. If unsure, feel free to open an issue first.

PR #6367, opened 2019-02-27
Add AdaBound optimizer

crcrpar:
This PR adds [AdaBound (ICLR 2019)](https://openreview.net/forum?id=Bkg3g2R9FX).


crcrpar:
Since this is not tested but the newer is tested, so let me close this.

PR #6370, opened 2019-02-27
Fix `str.join` `TypeError` in `FunctionTestCase` helper

hvy:
Fixes a typical `TypeError` when joining non `str`s. 

E.g.
```
In [1]: ''.join(type(i) for i in [1,2,3])
...
TypeError: sequence item 0: expected str instance, type found
```


niboshi:
Jenkins, test this please

niboshi:
LGTM

Issue #6325, opened 2019-02-20
Sequential out of bounds insertion is inconsistent


niboshi:
```py
from chainer import Sequential
from chainer import functions as F
from chainer import links as L

# IndexError
seq = Sequential(L.Linear(3))
seq.insert(100, L.Linear(4))

# works
seq = Sequential(F.sin)
seq.insert(100, F.cos)
```

PR #6374, opened 2019-02-27
Forbid out-of-range `insert` on `Sequence`

niboshi:
Fixes: #6325


mitmul:
Jenkins, test this please

mitmul:
@niboshi Could you resolve the conflicts? I will merge this after confirming all tests pass.

mitmul:
Jenkins, test this please

mitmul:
@niboshi The `insert()` method of the Python's list object can accept an index larger than the length of the list object. Is it OK not to align with the specification of Python's list here?

niboshi:
IMHO Python's specification of `list` was a failure (there may be other opinions though).
I think it's less error-prone to make such case an error.

@beam2d Do you have any opinion?

chainer/sequential.py
```diff
@@ -260,6 +260,12 @@ def extend(self, sequential):
             self.append(layer)
 
     def insert(self, i, layer):
+        n = len(self._layers)
+        if not (-n <= i < (n + 1)):
+            raise IndexError(
+                'Index out of range: {}'.format(i))
```
mitmul: How about additionally suggesting what should be used instead? For example,
```suggestion
                'Index out of range: {}. Please use .append() method instead.'.format(i))
```

mitmul: @niboshi Doesn't it make sense?

niboshi: When user encounteres this error, I think it's unlikely that he's trying to achieve the behavior of `append`.
For example, if a sequence `seq` has 2 elements, if you want to append a new layer, you can write either `seq.append(new_layer)` or `seq.insert(2, new_layer)` and both should work.
The error is caused when trying something like `seq.insert(3, new_layer)`. I think very few people would write this way. It's more likely that the code has flawed index handling.

mitmul: Got it. Thanks for explaining.

mitmul:
Jenkins, test this please

mitmul:
Jenkins, test this please

tests/chainer_tests/test_sequential.py
```diff
@@ -599,12 +593,69 @@ def test_flatten(self):
...
+    def test_index_out_of_bounds_empty_positive(self):
+        seq = chainer.Sequential()
+        with pytest.raises(IndexError):
+            seq.insert(1, functions.sin)
+
+    def test_index_out_of_bounds_empty_negative(self):
         seq = chainer.Sequential()
-        x = numpy.ones((2, 3), numpy.float32)
-        with pytest.raises(RuntimeError):
-            seq(x)
```
hvy: Seems like this test is irrelevant to this PR. Shouldn't we it?

hvy:
Basically LGTM. I added a comment regarding a deleted test.

niboshi:
@hvy Sorry, I restored the test. PTAL

tests/chainer_tests/test_sequential.py
```diff
@@ -607,4 +601,69 @@ def test_empty_sequential(self):
             seq(x)
 
 
+class SequentialInsert(unittest.TestCase):
```
hvy: How about adding tests inserting `Link`s (similar to the bug report https://github.com/chainer/chainer/issues/6325#issue-412548070)? The code paths are different for functions and links.

hvy: A nit.
```suggestion
class TestSequentialInsert(unittest.TestCase):
```

niboshi: Hmm, code paths are not different after the fix. What do you mean?
The test complexity would be greatly increase which I don't think worthwhile.

hvy: The paths are still different if the index is not out of bounds. How about restoring https://github.com/chainer/chainer/pull/6374/files#diff-a3da12ecfbd5149195d80196a0b251c1L424 ?

toslunar: `TestSequentialInsert` should test inserting `Link`s, while `Sequential.insert(_, link)` is called elsewhere in the file.  Having both `TestSequential.test_insert` and `TestSequentialInsert` is confusing.

hvy: I believe that's a perfectly valid point. My point was that we should still test inserting links, not necessarily with all combinations of invalid indices. We could do that by more or less moving the removed method to this class.

niboshi: It would be insufficient to literally restoring the single test.

niboshi: Fixed in 5030fb7c304cff0813d391596d1fa0114e436acd.

hvy:
Jenkins, test this please.

PR #3500, opened 2017-10-05
Fix some options of MNIST example not working

kmaehashi:

PR #6378, opened 2019-02-28
Use `args.out` in `train_cifar_custom_loop.py`

crcrpar:
Fix the training script with a custom loop to use `args.out` and modify the names of model and optimizer.

### DIFF
--- a/examples/cifar/train_cifar_custom_loop.py
+++ b/examples/cifar/train_cifar_custom_loop.py
@@ -37,7 +37,7 @@ def main():
                         help='Directory to output the result')
     parser.add_argument('--test', action='store_true',
                         help='Use tiny datasets for quick tests')
-    parser.add_argument('--resume', '-r', default='',
+    parser.add_argument('--resume', '-r', default=None, type=str,
                         help='Directory that has `vgg.model` and `vgg.state`')
     args = parser.parse_args()
 
@@ -77,7 +77,7 @@ def main():
     optimizer.setup(model)
     optimizer.add_hook(chainer.optimizer.WeightDecay(5e-4))
 
-    if args.resume != '':
+    if args.resume is not None:
         resume = args.resume
         if os.path.exists(resume):
             serializers.load_npz(os.path.join(resume, 'vgg.model'), model)
--- a/examples/cifar/train_cifar_custom_loop.py
+++ b/examples/cifar/train_cifar_custom_loop.py
@@ -7,6 +7,7 @@ training loop that manually computes the loss of minibatches and
 applies an optimizer to update the model.
 """
 import argparse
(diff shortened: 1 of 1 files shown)


hvy:
Do you think it's reasonable to fix the resuming as well? The `resume` argument is currently ignored. This could be a reference https://github.com/chainer/chainer/pull/3500.

examples/cifar/train_cifar_custom_loop.py
```diff
@@ -76,6 +77,17 @@ def main():
     optimizer.setup(model)
     optimizer.add_hook(chainer.optimizer.WeightDecay(5e-4))
 
+    if args.resume != '':
```
crcrpar: Which do you prefer?
1. `if args.resume != ''`
2. `if args.resume`

examples/cifar/train_cifar_custom_loop.py
```diff
@@ -76,6 +77,17 @@ def main():
...
+        resume = args.resume
+        if os.path.exists(resume):
+            serializers.load_npz(os.path.join(resume, 'vgg.model'), model)
+            serializers.load_npz(os.path.join(resume, 'vgg.state'), optimizer)
+        else:
```
crcrpar: Is this redundant?

examples/cifar/train_cifar_custom_loop.py
hvy: The former, since it's explicit about the type, so you can keep the current code.

examples/cifar/train_cifar_custom_loop.py
hvy: No, I think it's good (user friendly). Let's keep it as it. Preferably fix the MNIST example.

hvy:
jenkins, test this please.

examples/cifar/train_cifar_custom_loop.py
```diff
@@ -37,7 +38,7 @@ def main():
     parser.add_argument('--test', action='store_true',
                         help='Use tiny datasets for quick tests')
     parser.add_argument('--resume', '-r', default='',
```
hvy: Let's default to `None`, and in that case the `default` argument can be skipped altogether.

hvy:
Can we pass the `type` argument to the ones missing in the args?

crcrpar:
It works on both py27 & py36.

hvy:
Do you think it's better to be explicit about `default=None`? I think is cleaner to simply skip passing it.

crcrpar:
It's just my habit. Thank you.

hvy:
jenkins, test this please.

hvy:

PR #6391, opened 2019-03-01
Fix examples/.gitignore

crcrpar:
Current `examples/.gitignore` added by https://github.com/chainer/chainer/pull/2205 only ignores examples/result.
But I don't think this is expected behavior because every single example folder like `'mnist'` and `'dcgan'` has `README.md` and some users would run an experiment in one specific example folder.

IMO, it is natural to assume users to run a MNIST example by `python train_mnist.py ...` instead of `python mnist/train_mnist.py ...`.

### DIFF
--- a/examples/.gitignore
+++ b/examples/.gitignore
@@ -1,1 +1,1 @@
-result/
+**/result/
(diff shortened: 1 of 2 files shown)


Crissman:

PR #6393, opened 2019-03-01
Simplify `F.pad` test

ishanrai05:


Crissman:

toslunar:
The `chainer-py2` test is failing after this PR.  The cause seems the order of `with_requires` and `inject_backend_tests`.

takagi:
I've posted a PR to fix the problem as #6478 .

PR #6397, opened 2019-03-01
Fix to use right device for DALI iterator in imagenet example

takagi:
This PR fixes to use the right device in the imagenet example ran with the DALI iterator.


okuta:
LGTM!

takagi:
@okuta Would you merge this?

PR #6405, opened 2019-03-01
Refactor `roi_max_align_2d` and `roi_average_align_2d`

knorth55:
@Hakuyume can you review this?


knorth55:
@Hakuyume kindly ping.

chainer/functions/pooling/roi_average_align_2d.py
```diff
@@ -200,12 +202,21 @@ def forward_cpu(self, inputs):
                     w1, w2, w3, w4 = _get_bilinear_interp_params(
                         y, x, y_low, x_low, y_high, x_high)
 
-                    v1 = bottom_data[roi_batch_ind, c, y_low, x_low]
-                    v2 = bottom_data[roi_batch_ind, c, y_low, x_high]
-                    v3 = bottom_data[roi_batch_ind, c, y_high, x_low]
-                    v4 = bottom_data[roi_batch_ind, c, y_high, x_high]
+                    if w1 > 0 and y_low >= 0 and x_low >= 0:
```
Hakuyume: Do we need `w1 > 0` ?

chainer/functions/pooling/roi_average_align_2d.py
```diff
@@ -66,15 +68,15 @@ def _get_bilinear_interp_params(y, x, y_low, x_low, y_high, x_high):
         // empty
         return false;
     }
-    if (p <= 0) {
-        p = 0;
-    }
     low = (int)p;
```
Hakuyume: `(int)p` doesn't work when p is negative.
https://ideone.com/Hle9w3

chainer/functions/pooling/roi_average_align_2d.py
knorth55: we can skip this when `w1 = 0`

chainer/functions/pooling/roi_average_align_2d.py
knorth55: actually this is because of  `floor`.
Thank you I update to use `floor`.
```
(int)(-2.5) # -2.0
(int)floor(-2.5) # -3.0
```

chainer/functions/pooling/roi_average_align_2d.py
Hakuyume: If `w1 = 0`, the multiplied value is equals to 0. Does it improve performance?

Hakuyume: Yes. Since skipping and adding `0` are same, I wonder if comparing with `0` is more efficient than adding `0`. 
But this is a question and not important in this PR.

Hakuyume:
jenkins, test this please

toslunar:
Jenkins, test this please.

toslunar:
Jenkins, test this please.

toslunar:
The Jenkins failure
```
tests.chainer_tests.functions_tests.connection_tests.test_convolution_nd.TestConvolutionND_use_chainerx_false__chainerx_device_None__use_cuda_true__cuda_device_0__use_cudnn_always__cudnn_deterministic_false__autotune_false__cudnn_fast_batch_normalization_false__use_ideep_never_param_80_{b_dtype=float32,  dims=(3, 4, 3), autotune=False, cover_all=False, groups=1, dilate=1,  nobias=True, contiguous='C', W_dtype=float32,  x_dtype=float32}.test_backward
```
is unrelated to the PR.

PR #6414, opened 2019-03-04
Allow bool evaluation in `type_check`

niboshi:
I think it's useful when debugging.

Without this, it's difficult to debug in `check_type_forward` because you cannot write something like
```py
if x_type.dtype == numpy.float32:
    print("OK")
```
which would result in:
```
RuntimeError: Don't convert Expr to bool. Please call Expr.eval method to evaluate expression.
```

You cannot write something like this either (which is suggested by the error message above):
```py
if (x_type.dtype == numpy.float32).eval():
    print("OK")
```
which would result in:
```
AttributeError: 'bool' object has no attribute 'eval'
```

That's because `check_type_forward` is called twice in different modes, with different way of evaluation.
This is not intuitive.

### DIFF
--- a/chainer/utils/type_check.py
+++ b/chainer/utils/type_check.py
@@ -482,6 +482,12 @@ class BoolBinaryOperator(BinaryOperator, Testable):
                 '{0} {1} {2}'.format(self.lhs, self.exp, self.rhs),
                 '{0} {1} {2}'.format(left, self.inv, right))
 
+    def __bool__(self):
+        return self.eval()
+
+    def __nonzero__(self):
+        return self.__bool__()
+
 
 class InvalidType(Exception):
     """Raised when types of data for forward/backward are invalid.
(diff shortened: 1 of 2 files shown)


mitmul:
Jenkins, test this please

mitmul:
Jenkins, test this please

mitmul:
ChainerCI, test this please

mitmul:
Jenkins, test this please

mitmul:
Jenkins, test this please

toslunar:
> I think it's useful when debugging.

In general, this kind of proposal should be accompanied with an argument that the new benefit supersedes the existing one.

The existing benefit is explained:
https://github.com/chainer/chainer/blob/644d53fd02edcf15348919d6cfa78a510757b073/chainer/utils/type_check.py#L201-L213

niboshi:
@toslunar Updated the first post.

mitmul:
Jenkins, test this please

PR #6429, opened 2019-03-05
Properly pass device ID to DALI pipelines in imagenet example

takagi:
This PR fixes imagenet example to properly pass device ID, where #6397 was not enough.

This change is tested in https://github.com/chainer/chainer-test/pull/495.


takagi:
@okuta Would you review this? The change in #6397 was not enough.

okuta:
LGTM!

PR #6256, opened 2019-02-16
Add option `unit` to `CupyMemoryProfileHook.print_report()`

hitsgub:
Same as #6254.

The units of memory profile are not aligned in current version.
It is hard to find a bottleneck.
This PR adds option `align` to the member function `print_report()`.

PR #6433, opened 2019-03-05
Fix docment in `TimerHook`

hitsgub:
The PR renew the example output (L36--40) to present the behavior of the new (#6254) default value `unit='auto'`.
This change is same as https://github.com/chainer/chainer/pull/6256#pullrequestreview-210474886.


toslunar:
Jenkins, test this please.

PR #6435, opened 2019-03-06
Follow PEP8: use the fact that empty sequences are false

toslunar:
- It's possible that keep using `len(x)` is better for user inputs `x` (to assert their type).
- Please do not merge the PR until the coverage is checked.  `if len(x)` in the sense of `if x.shape[0]` cannot be replaced by `if x`.


Crissman:
Jenkins, test this please.

Crissman:
Jenkins, test this please.

Crissman:
Merging.

PR #6437, opened 2019-03-06
Add strides check in `NumpyOpTest`

niboshi:
Strides should be checked when using `NumpyOpTest`.

### DIFF
--- a/tests/chainerx_tests/op_utils.py
+++ b/tests/chainerx_tests/op_utils.py
@@ -126,7 +126,13 @@ class NumpyOpTest(_OpTest):
       * `forward_xp`: Forward implementation using both ChainerX and NumPy.
 
     It can have the same attributes as `chainer.testing.FunctionTestCase`.
+
+    This test also compares strides of forward output arrays with NumPy
+    outputs. Set ``check_numpy_strides_compliance`` attribute to ``False``
+    to skip this check.
     """
+
+    check_numpy_strides_compliance = True
 
     def forward_chainerx(self, inputs):
         return self.forward_xp(inputs, chainerx)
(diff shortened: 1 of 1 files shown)


chainer/testing/function.py
```diff
@@ -296,6 +306,12 @@ class FunctionTestCase(FunctionTestBase, unittest.TestCase):
         ``input_template`` is a tuple of template arrays. The returned arrays
         are expected to have the same shapes and dtypes as the template arrays.
 
+    ``check_forward_outputs(self, outputs, expected_outputs)``
+        Implements check logic of forward outputs. Typically additional check
+        can be done after calling ``super().check_forward_outputs``.
+        ``outputs`` and ``expected_outputs`` are tuples of arrays.
+        In case the check fails, ``FunctionTestCase`` should be raised.
```
hvy: Do you mean `FunctionTestError`?
```suggestion
        In case the check fails, ``FunctionTestError`` should be raised.
```

hvy:
Basically LGTM except the minor comment above.

chainer/testing/function.py
niboshi: Oops. Fixed.

niboshi:
Separated the introduction of `FunctionTestCase.check_forward_outputs()`: #6444

chainer/testing/function.py
```diff
@@ -69,6 +69,17 @@ def generate_grad_grad_inputs(self, inputs_template):
             for a in inputs_template])
         return grad_grad_inputs
 
+    def check_forward_outputs(self, expected_outputs, outputs):
```
toslunar: As far as I know, `assert actual == desired` is more common than the opposite.
```suggestion
    def check_forward_outputs(self, outputs, expected_outputs):
```

niboshi: Hmm, I also prefer the `actual, expected` order.
However, I think `expected, actual` is actually more common (perhaps historically, due to [JUnit](http://junit.sourceforge.net/javadoc/org/junit/Assert.html)).
(see also: ["historical note" on googletest primer](https://github.com/google/googletest/blob/master/googletest/docs/primer.md#binary-comparison) )

But we don't necessarily have to follow that convention.

niboshi: `numpy.testing.assert_array_equal` has `actual, expected` order.
I think it's ok to follow that.

hvy: Python, or `unittest`, was leaving "actual/expected" in favor of "first/second" so I think there's no real consensus but noticed now that NumPy follows `actual, expected` so maybe we can change?

toslunar: > However, I think `expected, actual` is actually more common (perhaps historically, due to [JUnit](http://junit.sourceforge.net/javadoc/org/junit/Assert.html)).

I didn't know this, sorry.:bow:

niboshi: NP, thank you for pointed out.
Changed the order in #6437.

hvy:
@niboshi could you rebase this PR now that https://github.com/chainer/chainer/pull/6444 is merged?

niboshi:
Rebased.

hvy:
jenkins, test this please.

hvy:
jenkins, test this please.

hvy:

PR #6439, opened 2019-03-06
Fix ResultTypeResolver multiple definitions

hvy:
Introduced `type_util.cc` and moved some `ResultTypeResolver` method definitions to it, to avoid multiple definitions error when including the header in multiple files.


asi1024:
Jenkins, test this please.

asi1024:
Jenkins, test this please.

PR #6440, opened 2019-03-06
Assert that input to array props formatter is a list or tuple

hvy:
Passing a single ndarray to `_format_array_props` yields unintuitive results. This PR simply asserts that it's never called that way.


Crissman:

PR #6445, opened 2019-03-06
Fix style of `.clang-tidy`

hvy:
so that's it's easier to work with (`Checks:` is getting longer).


hvy:
The style is borrowed from cmake project by the way https://github.com/Kitware/CMake/blob/master/.clang-tidy.

niboshi:
Jenkins, test this please

niboshi:
LGTM

PR #6450, opened 2019-03-06
Fix dead sphinx links

### LINKED ISSUE #6185: Dead links in documentation
https://docs.chainer.org/en/stable/guides/links.html

Some internal links are dead, e.g. `Link`, `F.linear`, `L.Linear`.

Crissman:
Closes #6185 and #6598.


chainer/link_hook.py
```diff
@@ -62,7 +62,7 @@ class LinkHook(object):
     :meth:`~chainer.LinkHook.forward_preprocess`
     (resp. :meth:`~chainer.LinkHook.forward_postprocess`)
     of all link hooks registered to this link are called before (resp. after)
-    :meth:`Link.forward() <chainer.Link.forward>` method of the link.
+    ``forward()`` method of the link.
```
niboshi: Instead of disabling cross reference, how about implementing `Link.forward` which just raises `NotImplementError`?

Crissman: I think having a link that when clicked provides no information disappoints the user. I think it's better to leave it unlinked, since the user has to provide the forward() method himself.

niboshi: By implementing `Link.forward`, we can write documentation there.

toslunar:
To keep git from managing `mean.npy` (1.5 MB), some commits should be squashed before the merge.

chainer/link_hook.py
toslunar: I'd like to keep it annotated (``:meth:`~chainer.Link.forward` `` looks better to me).  Being a dead link is informative here.

If `Link.forward` is defined, its docstring could give info to users.

A `Link` subclass often defines `forward`, but not always.  For example, when I define two equally-important calls, neither of them is named `forward`.

Crissman:
Reverted to :meth:`~chainer.Link.forward`. PTAL, and squash merge after testing.

chainer/link_hook.py
```diff
@@ -62,7 +62,7 @@ class LinkHook(object):
     :meth:`~chainer.LinkHook.forward_preprocess`
     (resp. :meth:`~chainer.LinkHook.forward_postprocess`)
     of all link hooks registered to this link are called before (resp. after)
-    :meth:`Link.forward() <chainer.Link.forward>` method of the link.
+    :meth:`~chainer.Link.forward` method of the link.
```
niboshi: Please revert to the previous markup. This would truncate the whole prefix to leave just `forward`.

chainer/link_hooks/timer.py
```diff
@@ -20,7 +20,7 @@
 
 
 class TimerHook(link_hook.LinkHook):
-    """Link hook for measuring elapsed time of :meth:`forward`.
+    """Link hook for measuring elapsed time of :meth:`~chainer.Link.forward`.
```
niboshi: This should also be marked up as ``` :meth:`Link.forward() <chainer.Link.forward>` ```, to add **`Link.`** prefix.

docs/source/examples/mnist.rst
```diff
@@ -113,28 +113,28 @@ Now let's create the :class:`~chainer.training.Updater` object !
...
+    Here, the model defined above is passed to :class:`~chainer.links.Classifier` and changed to a new :class:`~chainer.Chain`. :class:`~chainer.links.Classifier`, which in fact inherits from the :class:`~chainer.Chain` class, keeps the given :class:`~chainer.Chain` model in its ``predictor`` attribute. Once you give the input data and the corresponding class labels to the model by the ``()`` operator,
 
-    1. :meth:`~chainer.links.Classifier.forward` of the model is invoked. The data is then given to :attr:`~chainer.links.Classifier.predictor` to obtain the output ``y``.
-    2. Next, together with the given labels, the output ``y`` is passed to the loss function which is determined by :attr:`~chainer.links.Classifier.lossfun` argument in the constructor of :class:`~chainer.links.Classifier`.
+    1. :meth:`~chainer.links.Classifier.forward` of the model is invoked. The data is then given to ``predictor`` to obtain the output ``y``.
+    2. Next, together with the given labels, the output ``y`` is passed to the loss function which is determined by :attr:``lossfun`` argument in the constructor of :class:`~chainer.links.Classifier`.
```
niboshi: `:attr:` should be omitted from `lossfun`.

docs/source/guides/extensions.rst
```diff
@@ -170,8 +170,8 @@ This extension ``PolynomialShift`` takes five arguments.
...
+This extension calculates the number of iterations which will be performed during training by using ``stop_trigger``, ``batchsize``, and ``len_dataset``, then stores it as a property ``_maxiter``. This property will be used in the :meth:`~chainer.training.extensions.PolynomialShift.__call__` method to update the learning rate. The :meth:`~chainer.training.extensions.PolynomialShift.initialize` method obtains the initial learning rate from the optimizer given to the :class:`~chainer.training.Trainer` object. The :meth:`~chainer.training.extensions.PolynomialShift.serialize` method stores or recovers the properties, ``_t`` (number of iterations) and ``_last_value`` (the latest learning rate), belonging to this extension.
```
niboshi: `chainer.training.extensions.PolynomialShift` is not appropriate, because even with the same name, the one in the example is different from it.

docs/source/guides/functions.rst
```diff
@@ -135,8 +135,8 @@ MulAdd is simple and can be implemented as follows:
    w.grad = np.random.uniform(-1, 1, (3, 2)).astype(np.float32)
    w.backward()
 
-As per the warning above, the ``forward_cpu`` method returns a tuple of single element.
-Note that all arrays appearing in ``forward_cpu`` method are :class:`numpy.ndarray`.
+As per the warning above, the :meth:`~Function.forward_cpu` method returns a tuple of single element.
```
niboshi: `Function` -> `FunctionNode`.

docs/source/guides/type_checks.rst
```diff
@@ -26,16 +26,16 @@ These bugs are hard to fix.
...
 You can override this method to check the condition on types and shapes of arguments.
 
-:meth:`check_type_forward` gets an argument ``in_types``::
+:meth:`~chainer.Function.check_type_forward` gets an argument ``in_types``::
 
    def check_type_forward(self, in_types):
      ...
 
-``in_types`` is an instance of :class:`~utils.type_check.TypeInfoTuple`, which is a sub-class of :class:`tuple`.
+``in_types`` is an instance of :class:`~utils.type_check.TypeInfoTuple`, which is a sub-class of ``tuple``.
```
niboshi: ``` :class:`tuple` ``` is rendered correctly.

docs/source/guides/type_checks.rst
```diff
@@ -79,7 +79,7 @@ Detail of type information
 You can access three information of ``x_type``.
 
 * ``.shape`` is a tuple of ints. Each value is size of each dimension.
-* ``.ndim`` is :class:`int` value representing the number of dimensions. Note that ``ndim == len(shape)``
+* ``.ndim`` is :class:`~chainer.int` value representing the number of dimensions. Note that ``ndim == len(shape)``
```
niboshi: Ditto. ``` :class:`int` ``` is rendered correctly.

docs/source/reference/debug.rst
```diff
@@ -18,7 +18,7 @@ You can also enable or disable debug mode for the specific scope of code with :f
 
 See :ref:`configuration` for the details of Chainer's configuration mechanism.
 
-In debug mode, Chainer checks all results of forward and backward computation, and if it finds a NaN value, it raises :class:`RuntimeError`.
+In debug mode, Chainer checks all results of forward and backward computation, and if it finds a NaN value, it raises a ``RuntimeError``.
```
niboshi: Ditto. ``` :class:`RuntimeError` ``` is working.

Crissman:
Thanks for the thorough review. PTAL.

chainer/link.py
```diff
@@ -172,6 +172,12 @@ def __str__(self):
             cls=self.__class__.__name__, specs=specs,
         )
 
+    def forward(self):
+        """ This should be overwritten by the specific Link class.
```
niboshi: This looks insufficient as a documentation.
* What does this method do?
* What are arguments and return values?

As @toslunar said, it might be an option to leave it a dead link for now and fix it later.

Crissman: Ok. Leaving as dead link for now.

chainer/link_hooks/timer.py
```diff
@@ -20,7 +20,8 @@
 
 
 class TimerHook(link_hook.LinkHook):
-    """Link hook for measuring elapsed time of :meth:`forward`.
+    """Link hook for measuring elapsed time of
+    :meth:`Link.forward() <chainer.Link.forward>`.
```
niboshi: Summary line should not be separated.
Please join with `\` (and remove the extra spaces).

niboshi:
Jenkins, test this please

PR #6265, opened 2019-02-17
[WIP] Support mixed precision batch normalization

gwtnb:
Support mixed precision batch (re)normalization. Mixed precision BN reduces the risk of overflow. This PR depends on https://github.com/cupy/cupy/pull/2034.

TODOs:
- [ ] support iDeep
- [ ] consider memory consumption on batch renormalization
- [ ] consider compatibility of ChainerMN
- [ ] consider serialization

PR #6456, opened 2019-03-07
`mixed16` mode and its support in `L.BatchNormalization`

beam2d:
This PR introduces a new default dtype mode named `mixed16`. With this mode, Chainer basically uses `float16` as the default float dtype, but when Chainer decides that it is better to use higher precision representation for numerical stability and computational efficiency, it uses other float dtypes, esp. `float32`.

The batch normalization fix is based on #6265.

## General API

`chainer.mixed16` is a new dtype-like object. It can be passed to any numpy/cupy functions, where the dtype is interpreted as `float16`. Therefore, most existing code will just behave like the conventional FP16 mode. Specifying environment variable `CHAINER_DTYPE=mixed16` is equivalent to `chainer.global_config.dtype = chainer.mixed16`.

`chainer.get_dtype()` is extended to handle `mixed16` in a special manner. By passing an option `max_mixed16=np.float32`, we can get `float32` when the given (or default) dtype is `mixed16`. In this way, developers can choose the appropriate dtype when the users specify `mixed16`. Note that when `float16` is directly specified (either via `dtype` argument or `config.dtype`), `get_dtype()` still returns `float16` regardless of the `map_mixed16` option; i.e., `map_mixed16` is only used in `mixed16` case.

## Mixed precision batch normalization

cuDNN's batch normalization requires float32 beta/gamma parameters for mixed precision computation. Currently, Chainer (strictly speaking, CuPy) automatically casts these parameters to float32 on demand, which may cause a prohibitive overhead. In this PR, `L.BatchNormalization` is fixed to store these parameters in FP32 when `mixed16` dtype is specified, with which the conversion is not needed. Running mean and variance are also stored in float32 on mixed16 mode. `F.batch_normalization` and `F.fixed_batch_normalization` are fixed accordingly to accept arguments with mixed precisions.

### DIFF
--- a/chainer/__init__.py
+++ b/chainer/__init__.py
@@ -184,6 +184,18 @@ def is_arrays_compatible(arrays):
     return all([isinstance(a, types) for a in arrays])
 
 
+class _Mixed16(object):
+
+    dtype = numpy.dtype(numpy.float16)
+
+    def __repr__(self):
+        return "dtype('mixed16')"
+
+
+mixed16 = _Mixed16()
+"""Dtype-like object that represents 16/32 bits mixed precision float."""
+
+
 global_config.debug = bool(int(os.environ.get('CHAINER_DEBUG', '0')))
 global_config.cudnn_deterministic = False
 global_config.warn_nondeterministic = False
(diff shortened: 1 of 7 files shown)


gwtnb:
Jenkins, test this please

toslunar:
The design looks good to me.

beam2d:
Jenkins, test this please.

beam2d:
Jenkins, test this please.

toslunar:

toslunar:
Jenkins, test this please.

toslunar:
Jenkins, test this please.

toslunar:
Jenkins, test this please.

toslunar:
Jenkins, test this please.

toslunar:
Jenkins, test this please.

toslunar:
Jenkins, test this please.

mingxiaoh:
there is about 20% Convenet Performance regression on resnet50/101/152, googlenetv2/v3 and unet based on CPU-iDeep path, not check GPU yet.  Any comments here? @niboshi

PR #6462, opened 2019-03-07
Use __future__.division in imagenet example with Python2

takagi:
This PR follows #6429 to use `__future__.division` so that the imagenet example correctly runs with DALI in Python2. This change is tested in chainer/chainer-test#495.


okuta:
LGTM!

kmaehashi:
Removed from backport as #6567 will be backported.

PR #6465, opened 2019-03-07
Link test case param inits

hvy:
Trying a variation of https://github.com/chainer/chainer/pull/6187 where initializers are parameterized with normal parameterization.


hvy:
Closing as this was merely a test.

Issue #6423, opened 2019-03-04
ChainerX needs to support more routines


hvy:
**_Update_**
We've introduced an `Op` registration and dispatch mechanism. This practically means that `Device` methods will be replaced by `Op` implementations. E.g. `Device::Arange` -> `ArangeOp : public Op`. The following description will soon be updated to take this into account.

``` cpp
class FuncOp : public Op {
public:
    static const char* name() { return "Func"; }
    // Call is overridden per device. Does not create a graph but merely performs device computations such as calling a kernel in case of CUDA.
    // Call can have any signature.
    virtual void Call(..., const Array& out) = 0;  
    // Another definition would be
    // virtual Array Call(..., const nonstd::optional<Array>& out) = 0;  
};

class CudaFuncOp : public FuncOp {
// Override Call.
};

CHAINERX_REGISTER_OP_CUDA(FuncOp, CudaFuncOp);  // Allows backend.CallOp<FuncOp>(...);

Array Func(...) {  // A routine called `Func`.
    Array out = ....;
    {
        NoBackpropModeScope scope{};
        device.backend().CallOp<ArangeOp>(..., out);
    }
    // Create graph.
    return out;
}

```

---

The current repository of  backpropable operations, or ["routines"](https://github.com/chainer/chainer/tree/master/chainerx_cc/chainerx/routines) in ChainerX is still limited. We'd like to open this as a "contribution-welcome"-labeled issue for any contributor to introduce new routines and take part in the early development of ChainerX. 

### References

- [ChainerX routines spreadsheet](https://docs.google.com/spreadsheets/d/1B4E78tw9Awgpcdn5G7zsQ8NVFYJdOoJlIQg42QxKNfU) including statuses and priorities.
- [Available device methods](https://github.com/chainer/chainer/blob/master/chainerx_cc/chainerx/device.h) (low-level data manipulating operations)

### What kinds of routines are missing?

Routines that need to be implemented probably fall into either of the following two categories.

- NumPy compatible `{numpy,chainerx}` functions and `{chainerx,numpy}.ndarray` methods.
- Deep learning routines such as convolutions, pooling, RNN-type routines, etc.

Please make sure that it's not already implemented by checking the list of available routines.

### How do you start implementing routines?

1. Make sure you can build ChainerX. [Instructions here](https://docs.chainer.org/en/latest/chainerx/install/index.html).

[...]

PR #6471, opened 2019-03-08
Implementation of sigmoid for chainerx

dido1998:
Ref Issue: #6423


aksub99:
@dido1998 I'd suggest you checkout to another branch rather than committing directly to the master branch of your forked repository.

Issue #4800, opened 2018-05-22
Revise trainer extension tutorial


niboshi:
Current tutorial demonstrates [how to implement a trainer extension](https://docs.chainer.org/en/v5.0.0a1/guides/extensions.html#write-a-class-inherited-from-the-extension-class), by taking `PolynomialShift` as an example.

Now that we have `PolynomialShift` implemented (https://github.com/chainer/chainer/pull/4693) as an official trainer extension, we should revise it with another example.

PR #6475, opened 2019-03-09
Creating sigmoid_shift.py under training/extensions

hsezhiyan:
Adding sigmoid shift in chainer/training/extensions.

This extension reduces a specified attribute of training (learning rate, alpha/beta values, etc.) using the following formula: e^{-1 * rate * i} / (e^{-1 * rate * i} + 1), where `i` is the iteration number.

Based on this formula, the specified attribute will approach 0 as training proceeds.

### DIFF
--- a/chainer/training/extensions/sigmoid_shift.py
+++ b/chainer/training/extensions/sigmoid_shift.py
@@ -1,0 +1,88 @@
+from __future__ import division
+
+import numpy
+
+from chainer.training import extension
+
+
+class SigmoidShift(extension.Extension):
+
+    """Trainer extension to sigmoid shift an optimizer attribute.
+
+    This extension decreases the specified attribute of
+    an optimizer using the sigmoid function. This extension will
+    set the corresponding attribute to:
+    ``init_value * e^{-1 * rate * i} / (e^{-1 * rate * i} + 1)``,
+    at the ``i``-th iteration, where rate is the rate of shift.
+
+    This extension is also called before the training loop starts
+    by default.
+
+    Args:
+        attr (str): Name of the attribute to shift.
+        rate (float): Rate of sigmoid shift.
+        init (float): Initial value of the attribute. If it is ``None``, the
+            extension extracts the attribute at the first call and uses it as
+            the initial value.
+        target (float): Target value of the attribute. If the attribute reaches
+            this value, the shift stops.
+        optimizer (~chainer.Optimizer): Target optimizer to adjust the
+            attribute. If it is ``None``, the main optimizer of the updater is
+            used.
+    """
+
+    invoke_before_training = True
+
+    def __init__(self, attr, rate, init=None, target=None,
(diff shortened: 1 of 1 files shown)


chainer/training/extensions/sigmoid_shift.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+class SigmoidShift(extension.Extension):
+
+    """Trainer extension to sigmoid shift an optimizer attribute.
+
+    This extension decreases the specified attribute of
+    an optimizer using the sigmoid function. This extension will
+    set the corresponding attribute to:
+    ``init_value * e^{-1 * rate * i} / (e^{-1 * rate * i} + 1)``,
```
toslunar: Could you use `:math:` (inline) or `.. math::` (display)?

chainer/training/extensions/sigmoid_shift.py
```diff
@@ -0,0 +1,88 @@
+from __future__ import division
```
toslunar: Is the line required?

chainer/training/extensions/sigmoid_shift.py
```diff
@@ -0,0 +1,88 @@
...
+        rate (float): Rate of sigmoid shift.
+        init (float): Initial value of the attribute. If it is ``None``, the
+            extension extracts the attribute at the first call and uses it as
+            the initial value.
+        target (float): Target value of the attribute. If the attribute reaches
+            this value, the shift stops.
+        optimizer (~chainer.Optimizer): Target optimizer to adjust the
+            attribute. If it is ``None``, the main optimizer of the updater is
+            used.
+    """
```
toslunar: Could you cite a reference for the method?

chainer/training/extensions/sigmoid_shift.py
toslunar: OK to keep the formula in Python code as the other `*Shift` extensions do so.

chainer/training/extensions/sigmoid_shift.py
hsezhiyan: I don't believe this method is used specifically in any papers, and is therefore experimental.

toslunar:
> I don't believe this method is used specifically in any papers, and is therefore experimental.

Because the method seems not enough recognized to increase the maintenance cost, I'm negative to adding the extension to `chainer.training.extensions`.

Nevertheless it may be a good opportunity to replace the guide [Trainer Extensions - Write a class inherited from the Extension class](https://docs.chainer.org/en/latest/guides/extensions.html#write-a-class-inherited-from-the-extension-class) with this `SigmoidShift`.  Please see #4800.

Crissman:
@hsezhiyan Closing this PR due to the experimental nature of SigmoidShift. If you'd like to move this extension to the Trainer Extensions, please let me know the PR #, and we'll proceed.

Thanks!

PR #6482, opened 2019-03-11
fix NCCL version check error in ChainerMN

shu65:
'nccl.get_version()' in the CuPy master returns 0 when using the old version (such as 2.2) of NCCL.
It is the cause of a bug in `PureNcclCommunicator` in ChainerMN.
This PR fixes it.


shu65:
Jenkins, test this please

kuenishi:
Several things we discussed offline:
* We'll deprecate NCCL 2.2 and older version at Chainer v6
* We'll drop NCCL 2.2 support at Chainer v7
* We'll change CuPy and Chainer(MN) according to deprecation schedule above.

kuenishi:
Test failure is not related to ChainerMN part.

PR #6487, opened 2019-03-11
Simplify `F.squeeze` test

ishanrai05:
Simplifies `squeeze_test` using `testing.FunctionTestCase`

Solves part of issue #6071

### DIFF
--- a/tests/chainer_tests/functions_tests/array_tests/test_squeeze.py
+++ b/tests/chainer_tests/functions_tests/array_tests/test_squeeze.py
@@ -4,13 +4,29 @@ import numpy
 
 from chainer.backends import cuda
 from chainer import functions
-from chainer import gradient_check
 from chainer import testing
 from chainer.testing import attr
 from chainer.utils import type_check
-import chainerx
 
 
+@testing.inject_backend_tests(
+    None,
+    # CPU tests
+    [
+        {},
+    ]
+    # GPU tests
+    + testing.product({
+        'use_cuda': [True],
+        'use_cudnn': ['never', 'always'],
+        'cuda_device': [0, 1],
+    })
+    # ChainerX tests
+    + testing.product({
+        'use_chainerx': [True],
+        'chainerx_device': ['native:0', 'cuda:0', 'cuda:1'],
+    })
+)
 @testing.parameterize(*testing.product_dict(
     [
         {'axis': None, 'out_shape': (3,)},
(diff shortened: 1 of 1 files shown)


Crissman:

tests/chainer_tests/functions_tests/array_tests/test_squeeze.py
```diff
@@ -27,50 +43,28 @@
...
             self.check_forward_options = {'atol': 5e-4, 'rtol': 5e-3}
-            self.check_backward_options = {
-                'dtype': numpy.float64, 'atol': 2 ** -4, 'rtol': 2 ** -4}
-
-    def check_forward(self, x_data):
-        y = functions.squeeze(x_data, axis=self.axis)
-        expected = numpy.squeeze(self.x, axis=self.axis)
-        testing.assert_allclose(y.data, expected, **self.check_forward_options)
+            self.check_backward_options.update({
+                'atol': 2 ** -4, 'rtol': 2 ** -4})
```
toslunar: Do not mix styles (`options = {...}` and `options.update({...})`).  I'd always use `update` because `self.check_forward_options`, `self.check_backward_options`, and `self.check_double_backward_options` are `{}` by default.

BTW, does the test require large tolerances for float16?  Did you check double-backward tests pass without adjusting tolerances?

ishanrai05: Thanks.. @toslunar . Made the required changes. I'll keep your words in mind. 

Also, I did try adjusting the relative and absolute tolerances after your comment and they turn out to be extremely low. Should I update the `atol` and `rtol` as well?

toslunar: It's no problem to leave too large tolerances.  Could you set `self.check_double_backward_options` if you found the double-backward tests are possible to fail.  Double backward is often (but not always) less accurate than backward.

ishanrai05: @toslunar  I find that double backward tests are likely to fail when tolerances are less than the order of 1e-13 which is again extremely low. Should I update the `self.check_double_backward_options`. If yes then what should be the order of magnitude for absolute and relative tolerance values?

Also is there any specific paradigm to follow while setting up or calculating tolerances?

toslunar: No.



> Should I update the `self.check_double_backward_options`.

No.

> Also is there any specific paradigm to follow while setting up or calculating tolerances?

`chainer.testing.condition.repeat` is often used to confirm stability of tests.

ishanrai05: @toslunar Thanks. I think the commit is ready to be merged then.

toslunar:
Jenkins, test this please.

PR #6489, opened 2019-03-12
Fix broken imagenet example

### LINKED ISSUE #5631: ChainerMN's ImageNet example is broken because of update in pickle module
* Conditions

  - Chainer >= 5.0.0

* Code to reproduce
The ImageNet example on Python >= 3.6

```
import multiprocessing
p = multiprocessing.Process(target=lambda *x: x, args=())
p.start()
```

```
  File "/home/kfukuda/.pyenv/versions/anaconda3-4.3.1/lib/python3.6/multiprocessing/process.py", line 105, in start
    self._popen = self._Popen(self)
  File "/home/kfukuda/.pyenv/versions/anaconda3-4.3.1/lib/python3.6/multiprocessing/context.py", line 223, in _Popen
    return _default_context.get_context().Process._Popen(process_obj)
  File "/home/kfukuda/.pyenv/versions/anaconda3-4.3.1/lib/python3.6/multiprocessing/context.py", line 291, in _Popen
    return Popen(process_obj)
  File "/home/kfukuda/.pyenv/versions/anaconda3-4.3.1/lib/python3.6/multiprocessing/popen_forkserver.py", line 35, in __init__
    super().__init__(process_obj)
  File "/home/kfukuda/.pyenv/versions/anaconda3-4.3.1/lib/python3.6/multiprocessing/popen_fork.py", line 20, in __init__
    self._launch(process_obj)
  File "/home/kfukuda/.pyenv/versions/anaconda3-4.3.1/lib/python3.6/multiprocessing/popen_forkserver.py", line 47, in _launch
    reduction.dump(process_obj, buf)
  File "/home/kfukuda/.pyenv/versions/anaconda3-4.3.1/lib/python3.6/multiprocessing/reduction.py", line 60, in dump
    ForkingPickler(file, protocol).dump(obj)
_pickle.PicklingError: Can't pickle <function <lambda> at 0x7fb633f35e18>: attribute lookup <lambda> on __main__ failed
```
For details, see https://docs.python.org/3.6/library/pickle.html

```
The following types can be pickled:

* None, True, and False
* integers, floating point numbers, complex numbers
* strings, bytes, bytearrays
* tuples, lists, sets, and dictionaries containing only picklable objects
* functions defined at the top level of a module (using def, not lambda)
* built-in functions defined at the top level of a module
* classes that are defined at the top level of a module
* instances of such classes whose __dict__ or the result of calling __getstate__() is picklable (see section Pickling Class Instances for details).
```

Thus lambda is not pickable.

keisukefukuda:
This PR fixes #5631 by rewriting a dummy function from a lambda to an explicitly defined local function.

### DIFF
--- a/docs/source/chainermn/tutorial/tips_faqs.rst
+++ b/docs/source/chainermn/tutorial/tips_faqs.rst
@@ -15,7 +15,7 @@ start child processes, with a process explicitly forked right after, **before
 communicator is created** as follows::
 
   multiprocessing.set_start_method('forkserver')
-  p = multiprocessing.Process(target=lambda *x: x, args=())
+  p = multiprocessing.Process()
   p.start()
   p.join()
 
--- a/examples/chainermn/imagenet/train_imagenet.py
+++ b/examples/chainermn/imagenet/train_imagenet.py
@@ -88,10 +88,6 @@ class TestModeEvaluator(extensions.Evaluator):
         return ret
 
 
-def _mp_dummy_func(*x):
-    return x
-
-
 def main():
     # Check if GPU is available
     # (ImageNet example does not support CPU execution)
@@ -147,7 +143,7 @@ def main():
     # https://chainermn.readthedocs.io/en/stable/tutorial/tips_faqs.html#using-multiprocessiterator
     # This must be done *before* ``chainermn.create_communicator``!!!
     multiprocessing.set_start_method('forkserver')
-    p = multiprocessing.Process(target=_mp_dummy_func, args=())
+    p = multiprocessing.Process()
     p.start()
     p.join()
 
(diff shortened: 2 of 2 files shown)


keisukefukuda:
Jenkins, test please.

keisukefukuda:
Jenkins, test this please

shu65:
A dummy function is not used in ChinaerCV example.
https://github.com/chainer/chainercv/blob/3726368c49b5d9476eafdc3081d8cf7db607e06e/examples/classification/train_imagenet_multi.py#L89

We may not need the dummy function.

shu65:
Jenkins, test this please

shu65:
I confirmed that the unittests for chainermn were passed

PR #6492, opened 2019-03-12
Improve contribution docs

Crissman:
I've had a few questions and comments about people having issues building documentation locally. Usually this is related to not having installed or updated local code. I've edited the `Contribution` doc to make the solution more clear.


hvy:
@niboshi Can you take a look at this PR?

docs/source/contribution.rst
```diff
@@ -431,15 +431,22 @@ To build the documentation, you need to install `Sphinx <http://www.sphinx-doc.o
 
   $ pip install sphinx sphinx_rtd_theme
 
-Then you can build the documentation in HTML format locally::
+Docstrings (documentation comments in the source code) are collected from the installed Chainer module.
+If Chainer has not been installed yet, then ``make html`` will yield an error that the reference sources cannot
+be read, such as ``chainer.AbstractSerializer``.
```
niboshi: >such as ``chainer.AbstractSerializer``.

What does this mean?

docs/source/contribution.rst
```diff
@@ -431,15 +431,22 @@ To build the documentation, you need to install `Sphinx <http://www.sphinx-doc.o
...
+
+Alternatively, if you have modified docstrings, make sure to install the module before building the
```
niboshi: >Alternatively,

Is this needed?
It sounds like there are two methods that are explaned. What are those?

docs/source/contribution.rst
Crissman: It's a function I often saw an error on. Edited to remove specific reference.

docs/source/contribution.rst
Crissman: Both situations have the same remedy--install from top of GitHub directory. I edited to provide clarity.

docs/source/contribution.rst
```diff
@@ -431,15 +431,22 @@ To build the documentation, you need to install `Sphinx <http://www.sphinx-doc.o
 
   $ pip install sphinx sphinx_rtd_theme
 
-Then you can build the documentation in HTML format locally::
+Docstrings (documentation comments in the source code) are collected from the installed Chainer module.
+If Chainer has not been installed yet, then ``make html`` will yield an error that the reference sources cannot
+be read.
+
+Another reason it may be necessary to reinstall Chainer is when you have edited docstrings and want to see those changes reflected in the generated html. In that case,  make sure to install the module before building the
```
niboshi: This sentence sounds a bit unexpected to me, because there's no mention about "reinstalling" above.
Actually if Chainer is installed with `-e` option as demonstrated below, I think no reinstalling is needed.
Am I missing something?

Crissman: Thanks for the comment. I've edited to make more clear. PTAL!

docs/source/contribution.rst
```diff
@@ -431,15 +431,21 @@ To build the documentation, you need to install `Sphinx <http://www.sphinx-doc.o
 
   $ pip install sphinx sphinx_rtd_theme
 
-Then you can build the documentation in HTML format locally::
+Docstrings (documentation comments in the source code) are collected from the installed Chainer module.
+In order to generate the html files, Chainer must be installed in develop mode. If Chainer has not been installed in develop mode, then ``make html`` will yield an error that the reference sources cannot
```
niboshi: >Chainer must be installed in develop mode

This statement is not correct either: if Chainer is installed in non-develop (more precisely, non-editable) mode, modules are collected from "installed" files. The build process works just fine. The problem is that if a developer modifies files in the checked-out source files, which are not "installed" ones, the edits will not be reflected.
Let's be more precise and clear.

docs/source/contribution.rst
```diff
@@ -431,15 +431,21 @@ To build the documentation, you need to install `Sphinx <http://www.sphinx-doc.o
...
+be read.
+
+If you have edited docstrings and want to see those changes reflected in the generated html, it is necessary to reinstall Chainer in develop mode again to see the changes reflected.
```
niboshi: If Chainer is installed in editable mode, I think reinstalling is not needed.

Crissman: You're correct. Confirmed.

docs/source/contribution.rst
Crissman: Edited accordingly.

docs/source/contribution.rst
Crissman: Actually, reinstallation is often required. For an example, please see this gist: https://gist.github.com/Crissman/d5d60509b678cc2ec15d71d3923b507f#file-gistfile1-txt

docs/source/contribution.rst
niboshi: Hmm, is this comment reflected?
It's still saying "Chainer must be installed in develop mode".

niboshi:
ping

docs/source/contribution.rst
Crissman: Yes. "In order to see changes in checked-out source files..." In other words, only if you want to see local changes, do you need to install in develop mode.

docs/source/contribution.rst
Crissman: Not required, and that line removed.

Crissman:
Pong! See comments above. I believe this is ready for merge.

docs/source/contribution.rst
niboshi: Thanks, I confirmed.
I don't know why I gave the above comment... perhaps I was looking at a wrong diff.

niboshi:
Jenkins, test this please
@Crissman Btw, please do not remove `st:` labels by yourself.

niboshi:
https://docs.chainer.org/en/v7.0.0a1/contribution.html#documentation

It seems we already have a note:

clip

Crissman:
Noted. I consolidated the notes, and clarified what should be done to fix.

docs/source/contribution.rst
```diff
@@ -275,19 +275,19 @@ To build the documentation, you need to install `Sphinx <http://www.sphinx-doc.o
 
   $ pip install sphinx sphinx_rtd_theme
 
-Then you can build the documentation in HTML format locally::
+
+.. note::
+
+   Docstrings (documentation comments in the source code) are collected from the installed Chainer module. If you have edited docstrings in checked-out source files and want to see those changes reflected in the generated html, Chainer must be installed in develop mode to see those changes reflected in the generated documentation. To do this use `pip install -e .`from the the top of the Chainer directory.
+
+Move to the docs directory, and make the html::
```
niboshi: The word "html" appears here at the first time, so "the html" sounds a bit too sudden. 
I think the original sentence was okay.

Crissman:
Review suggestions reflected. Please review.

niboshi:
Thanks.
Jenkins, test this please

docs/source/contribution.rst
```diff
@@ -275,6 +275,14 @@ To build the documentation, you need to install `Sphinx <http://www.sphinx-doc.o
 
   $ pip install sphinx sphinx_rtd_theme
 
+
+.. note::
+
+   Docstrings (documentation comments in the source code) are collected from the installed Chainer module.
+   If you have edited docstrings in checked-out source files and want to see those changes reflected in the generated html,
+   Chainer must be installed in develop mode to see those changes reflected in the generated documentation.
+   To do this use `pip install -e .` from the the top of the Chainer directory.
```
niboshi:
```suggestion
   To do this use ``pip install -e .`` from the the top of the Chainer directory.
```

Crissman:
Jenkins, test this please.

niboshi:
LGTM

PR #6410, opened 2019-03-02
Simplify `F.vstack` test

ishanrai05:
Simplifies `test_vstack` using `function.FunctionTestCase`

Solves part of issue #6071

PR #6494, opened 2019-03-12
Fix input of `F.vstack` to a list of ndarrays

crcrpar:
#6410 changed inputs to an `ndarray` from a list of `ndarray`s.


toslunar:
Jenkins, test this please.

PR #6496, opened 2019-03-12
Elementwise `Power` for ChainerX

dido1998:
The double backward test seems to be failing. @hvy @aksub99  could you check this? Ref issue: #6423


hvy:
It is probably because you are using device methods in the backward definitions, instead of routines. 

```c++
    // ...
    x1.device().Pow(x1, x2 - 1, pow_x1);
    // should be replaced with something like
    // Array pow_x1 = ElementwisePower(x1, x2 - 1);
```

Similar to the Chainer `FunctionNode`s, the backward computations must extend the computational graph such that the backward itself can be backwarded through during double backward. Since you are allocating new arrays and using the device functions, the computational graphs are not extended.

chainerx_cc/chainerx/routines/math.cc
```diff
@@ -575,6 +577,47 @@ Array Sqrt(const Array& x) {
     return out;
 }
 
+Array ElementwisePower(const Array& x1, const Array& x2) {
```
hvy: How about using a simpler and more consistent name such as `Pow`?

dido1998:
Thank you! I'll make the changes you suggested.

chainerx_cc/chainerx/routines/math.cc
dido1998: Yes, I'll change it.

chainerx_cc/chainerx/routines/math.cc
```diff
@@ -594,28 +593,47 @@ Array ElementwisePower(const Array& x1, const Array& x2) {
...
         if (BackwardBuilder::Target bt = bb.CreateTarget(1)) {
             bt.Define([x1_tok = bb.RetainInput(0), out_tok = bb.RetainOutput(0)](BackwardContext& bctx) {
+                const Array& gout = *bctx.output_grad();
                 const Array& x1 = bctx.GetRetainedInput(x1_tok);
                 const Array& out = bctx.GetRetainedOutput(out_tok);
-                Array logx1 = EmptyLike(x1, x1.device());
-                x1.device().Log(x1, logx1);
 
-                bctx.input_grad() = *bctx.output_grad() * out * logx1;
+                bctx.input_grad() = gout * out * Log(x1);
```
dido1998: I have done as you suggested, but still, the double backward test is failing.

hvy: Would it be possible to provide some info about e.g. inputs and an actual failure message?

hvy:
I have some questions, which some of could be addressed in separate PRs but I also think would make sense to include here.

- Rename `chainerx.power` to `chainerx.pow` (the latter is used in PyTorch so it'd be more familiar to users).
- Alias `chainerx.power` with `__pow__` so that `**` can be used in Python.
- Implement `Power(chainerx::Scalar, chainerx::Array)` also.

Could I hear your thoughts about those?

chainerx_cc/chainerx/routines/math.cc
dido1998: The inputs are - 
a = [[2, 6],[4,6]] b = [[3, 5],[4, 2]]

Here is the test error - 
`
[ RUN      ] ForEachBackend/MathTest.PowerDoubleBackward/0
2: unknown file: Failure
2: C++ exception with description "Numerical error in backward on inputs (out of 3): 0, 1
2: Backprop ID: <default>
2: Atol: 1e-05  Rtol: 0.0001
2: Error[0]:
2: array([[-0.19791481, -0.        ],
2:        [28.06966120, -0.        ]], shape=(2, 2), dtype=float64, device='cuda:0')
2: Backward gradients[0]:
2: array([[   24.30505552,    -0.        ],
2:        [-6255.52749001,     0.        ]], shape=(2, 2), dtype=float64, device='cuda:0')
2: Numerical gradients[0]:
2: array([[   24.50297033,     0.        ],
2:        [-6283.59715121,     0.        ]], shape=(2, 2), dtype=float64, device='cuda:0')
2: Eps[0] (perturbation in numerical gradients):
2: array([[0.1, 0.1],
2:        [0.1, 0.1]], shape=(2, 2), dtype=float64, device='cuda:0')
2: Error[1]:
2: array([[ -0.05591017,  -0.        ],
2:        [124.21004340,  -0.        ]], shape=(2, 2), dtype=float64, device='cuda:0')
2: Backward gradients[1]:
2: array([[   13.59493467,    -0.        ],
2:        [-8999.68248524,     0.        ]], shape=(2, 2), dtype=float64, device='cuda:0')
2: Numerical gradients[1]:
2: array([[   13.65084483,     0.        ],
2:        [-9123.89252864,     0.        ]], shape=(2, 2), dtype=float64, device='cuda:0')
2: Eps[1] (perturbation in numerical gradients):
2: array([[0.1, 0.1],
2:        [0.1, 0.1]], shape=(2, 2), dtype=float64, device='cuda:0')
2: " thrown in the test body.
2: [  FAILED  ] ForEachBackend/MathTest.PowerDoubleBackward/0, where GetParam() = "cuda" (13 ms)
2: [ RUN      ] ForEachBackend/MathTest.PowerDoubleBackward/1
2: unknown file: Failure
2: C++ exception with description "Numerical error in backward on inputs (out of 3): 0, 1
2: Backprop ID: <default>
2: Atol: 1e-05  Rtol: 0.0001
2: Error[0]:
2: array([[-0.19791481, -0.        ],
2:        [28.06966120, -0.        ]], shape=(2, 2), dtype=float64, device='native:0')
2: Backward gradients[0]:
2: array([[   24.30505552,    -0.        ],
2:        [-6255.52749001,     0.        ]], shape=(2, 2), dtype=float64, device='native:0')
2: Numerical gradients[0]:
2: array([[   24.50297033,     0.        ],
2:        [-6283.59715121,     0.        ]], shape=(2, 2), dtype=float64, device='native:0')
2: Eps[0] (perturbation in numerical gradients):
2: array([[0.1, 0.1],
2:        [0.1, 0.1]], shape=(2, 2), dtype=float64, device='native:0')
2: Error[1]:
2: array([[ -0.05591017,  -0.        ],
2:        [124.21004340,  -0.        ]], shape=(2, 2), dtype=float64, device='native:0')
2: Backward gradients[1]:
2: array([[   13.59493467,    -0.        ],
2:        [-8999.68248524,     0.        ]], shape=(2, 2), dtype=float64, device='native:0')
2: Numerical gradients[1]:
2: array([[   13.65084483,     0.        ],
2:        [-9123.89252864,     0.        ]], shape=(2, 2), dtype=float64, device='native:0')
2: Eps[1] (perturbation in numerical gradients):
2: array([[0.1, 0.1],
2:        [0.1, 0.1]], shape=(2, 2), dtype=float64, device='native:0')
2: " thrown in the test body.
2: [  FAILED  ] ForEachBackend/MathTest.PowerDoubleBackward/1, where GetParam() = "native" (4 ms)

`

dido1998:
- I can change it to chainerx.pow, I had kept it as chainerx.power as numpy uses power.

- Yes, I will alias chainerx.power with __pow__.

-

chainerx_cc/chainerx/routines/math.cc
```diff
@@ -594,28 +593,47 @@ Array ElementwisePower(const Array& x1, const Array& x2) {
...
         }
-        bb.Finalize();
-        return out;
+    
+    bb.Finalize();
+    }
+    return out;
+}
+
+Array Pow(const Array& x1, const Scalar x2) {
```
dido1998: I have implemented `pow(Array, Scalar)` here. Are you referring to something else?

chainerx_cc/chainerx/routines/math.cc
dido1998: Fixed the issue.

chainerx_cc/chainerx/python/routines.cc
```diff
@@ -596,6 +596,14 @@ void InitChainerxMath(pybind11::module& m) {
           py::arg("x"),
           py::arg("axis") = nullptr);
     m.def("sqrt", [](const ArrayBodyPtr& x) { return MoveArrayBody(Sqrt(Array{x})); }, py::arg("x"));
+    m.def("pow",
+      [](const ArrayBodyPtr& x1, const ArrayBodyPtr& x2) { return MoveArrayBody(Pow(Array{x1}, Array{x2})); },
+      py::arg("x1"),
+      py::arg("x2"));
+    m.def("pow",
```
hvy: You are right about `numpy.power`. Sorry but could you change it back to `chainerx.power` and similarly `Power` to the C++ API?

chainerx_cc/chainerx/routines/math.cc
hvy: I meant the reversed `pow(Scalar, Array)`.

tests/chainerx_tests/unit_tests/routines_tests/test_math.py
```diff
@@ -581,6 +581,58 @@ def forward_xp(self, inputs, xp):
...
+            self.check_backward_options = {'atol': 5e-4, 'rtol': 5e-3}
+            self.check_double_backward_options = {'atol': 5e-3, 'rtol': 5e-1}
+
+    def generate_inputs(self):
+        return self.lhs, self.rhs
+
+    def forward_xp(self, inputs, xp):
+        if xp is numpy:
+            return xp.power(inputs[0], inputs[1]),
+        return xp.pow(inputs[0], inputs[1]),
```
hvy: You can parameterize `is_module` to test both `chainerx.power` and `**` similar to this test https://github.com/chainer/chainer/blob/master/tests/chainerx_tests/unit_tests/routines_tests/test_math.py#L295

hvy:
Wrote some comments, could you take another look?

dido1998:
Yes, I will make the suggested changes.

chainerx_cc/chainerx/routines/math.cc
vishalgolcha: @dido1998 could you specify what the issue was with the tests earlier i was trying to get a workaround but couldn't.

chainerx_cc/chainerx/routines/math.cc
vishalgolcha: @hvy  @dido1998  i have added the reversed pow but it fails the DoubleBackward test can you take a look.

dido1998: @vishalgolcha is your error the same as above?

dido1998:
@hvy I would like to know your thoughts on these -

- For the `power(chainerx::array, chainerx::scalar)` I chose to implement the elementwise power operator as a device methods as this has been done for other ops like multiply.  

- For `power(chainerx::scalar, chainerx::array)` I didn't implement the device method (converted scalar to array) as that makes it consistent with other ops for which only the `AS` method is implemented.

If you want I could change both to have their own device method or both to reuse `Power(chainerx::Array, chainerx::Array)`.

hvy:
@dido1998 thanks for checking.

- That sounds reasonable to me.
- Which "other" ops are you referring to in this case? I think it's okay to implement it as a device method too to avoid unnecessary array allocation.

Some additional thoughts.

- Can you reuse the `*Binary*` helpers in `*math.cc*` ?
- It seems like some irrelevant formatting changes are included. Are you maybe using the default clang-format? ChainerX uses a slightly customized one (such as longer line length). The config is included in the project repository.

dido1998:
- I was talking about the scalar, array case as in multiply or divide. I will implement it as a device method.

- I will try to reuse the `*Binary*` helpers.

- I ran the `run-clang-format.sh` and it used the default clang-format. I will run the clang-format command and explicitly specify the config file `.clang-format`.

dido1998:
- I think reuse of the `*Binary*` will still require the same amount code to be written. As, in contrast to 'Add', the backward computation for `power(scalar, array)`, `power(array, scalar)` and `power(array, array)` are all different.

- The chainerx test is failing due to failure MaxPoolND.

- I will fix the formatting issues by tomorrow.

hvy:
For the record, as for division with `scalar / Array`, it's not yet supported, and for multiplication, we can simply `a * b` -> b * a` to reuse `Array * scalar` for `scalar * Array`. So this case with power is a bit different.

dido1998:
@hvy , could you check this?

dido1998:
@hvy  could have a look at this if this can be merged?

tests/chainerx_tests/unit_tests/routines_tests/test_math.py
```diff
@@ -1798,6 +1798,86 @@ def func(self, xp, a):
 
 
 @op_utils.op_test(['native:0', 'cuda:0'])
+@pytest.mark.parametrize('lhs, rhs', [
+    (numpy.asarray([[1., 2.], [4., 5.]]), numpy.asarray([1., 3.])),
+    (numpy.asarray([1., 2., 3.]), numpy.asarray([3., 4., 5.]))
+])
+@pytest.mark.parametrize('contiguous', [None, 'C'])
+class TestPow(op_utils.NumpyOpTest):
```
hvy: As for the tests in this PR, could you parameterize them similarly to e.g. https://github.com/chainer/chainer/blob/137224ebbd92298e6a6c58e79b7c6886217d51df/tests/chainerx_tests/unit_tests/routines_tests/test_math.py#L1373-L1409 to covert shape and value edge cases as well?

chainerx_cc/chainerx/cuda/cuda_device.h
```diff
@@ -229,7 +231,7 @@ class CudaDevice : public Device {
     std::shared_ptr<void> AllocatePinnedMemory(size_t bytesize);
 
     // Asynchronous transfer from host to this device, w.r.t. host, using temporary pinned memory.
-    // The current device must be set to this device, prior to calling this function.
+    // The current device must be set to this device, prior to calling this  function.
```
hvy: This seems like a small mistake. Can you revert it?
```suggestion
    // The current device must be set to this device, prior to calling this function.
```

chainerx_cc/chainerx/device.h
```diff
@@ -210,7 +210,9 @@ class Device {
     virtual void Square(const Array& x, const Array& out) = 0;
 
     virtual void Sqrt(const Array& x, const Array& out) = 0;
-
+    virtual void Pow(const Array& x1, const Array& x2, const Array& out) = 0;
+    virtual void PowAS(const Array& x1, Scalar x2, const Array& out) = 0;
+    virtual void PowSA(Scalar x1, const Array& x2, const Array& out) = 0;
```
hvy: Ditto regarding spacing.

chainerx_cc/chainerx/native/native_device.h
```diff
@@ -116,7 +116,9 @@ class NativeDevice : public Device {
     void Square(const Array& x, const Array& out) override;
 
     void Sqrt(const Array& x, const Array& out) override;
-
+    void Pow(const Array& x1, const Array& x2, const Array& out) override;
+    void PowAS(const Array& x1, Scalar x2, const Array& out) override;
+    void PowSA(Scalar x1, const Array& x2, const Array& out) override;
```
hvy: Ditto regarding spacing.

chainerx_cc/chainerx/routines/math.cc
```diff
@@ -501,7 +502,8 @@ Array AMax(const Array& a, const OptionalAxes& axis, bool keepdims) {
 
     BackwardBuilder bb{"amax", a, out};
     if (BackwardBuilder::Target bt = bb.CreateTarget(0)) {
-        // a and out are used only for restoring the mask. We don't need graph nodes.
+        // a and out are used only for restoring the mask. We don't need graph
+        // nodes.
```
hvy: Is this change required? It looks like an application of a different formatter.

chainerx_cc/chainerx/native/native_device/misc.cc
```diff
@@ -35,6 +35,41 @@ void NativeDevice::Sqrt(const Array& x, const Array& out) {
...
+        Elementwise<const T, T>(Impl{static_cast<T>(x2)}, x1, out);
+    });
+}
+
+void NativeDevice::PowSA(Scalar x1, const Array& x2, const Array& out) {
+    CheckDevicesCompatible(x2, out);
+    VisitNumericDtype(out.dtype(), [&](auto pt) {
+        using T = typename decltype(pt)::type;
+        struct Impl {
+            void operator()(int64_t /*i*/, T x1, T& out) { out = chainerx::Pow(x2, x1); }
```
hvy: Is the order of `x2` and `x1` correct here if we're taking `x2 ** x1`?

chainerx_cc/chainerx/python/routines.cc
```diff
@@ -294,7 +294,10 @@ void InitChainerxIndexing(pybind11::module& m) {
                   return MoveArrayBody(
                           Take(Array{a}, Array{MakeArrayFromNumpyArray(py::cast<py::array>(indices), a->device())}, axis.value()));
               }
-              throw py::type_error{"only integers, slices (`:`), sequence, numpy.ndarray and chainerx.newaxis (`None`) are valid indices"};
+              throw py::type_error{
+                      "only integers, slices (`:`), sequence, "
+                      "numpy.ndarray and chainerx.newaxis (`None`) "
+                      "are valid indices"};
```
hvy: Ditto regarding applying a formatter with different rules..

hvy:
I wrote some comments could you take a look?

hvy:
It seems like you are applying clang-format with a different set of rules. Did you fix it? Since there's a set of rules included in this repository which you should be able to utilize, automatically.

dido1998:
Yes, I am having some problems to get it working. I will try to get it working. I will make the suggested changes.

dido1998:
@hvy, I have included OP implementations. I  have also updated the test as you asked. One particular test is failing. I observed that this test fails sometimes and passes sometimes on my system. Do you know why this might be happening?

hvy:
Thanks for you updates. The gradient check is stochastic and you're in addition generating random inputs meaning that the tests results may vary between runs. I think it's in this case a matter of precision. Do you know for what particular configurations (e.g. dtypes) the tests are failing?

dido1998:
I saw that the tests were failing for float16. I reduced the tolerance and it passed on my system every time that I ran it.

hvy:
Jenkins, test this please.

chainerx/__init__.pyi
```diff
@@ -780,6 +784,7 @@ def squared_difference(x1: tp.Any, x2: tp.Any) -> ndarray: ...
 
 def sqrt(x: ndarray) -> ndarray: ...
 
+def power(x1: ndarray, x2: ndarray) -> ndarray: ...
```
hvy: Should this be `tp.Any` too?
```suggestion
def power(x1: tp.Any, x2: tp.Any) -> ndarray: ...
```

tests/chainerx_tests/unit_tests/routines_tests/test_math.py
```diff
@@ -1899,6 +1899,124 @@ def func(self, xp, a):
...
+            self.check_forward_options.update({'rtol': 5e-3, 'atol': 5e-3})
+            self.check_backward_options.update({'rtol': 5e-3, 'atol': 5e-3})
+            self.check_double_backward_options.update(
+                {'rtol': 5e-3, 'atol': 5e-3})
+
+    def func(self, xp, a, b):
+        if self.is_module:
+            return xp.power(a, b)
+        else:
+            return a**b
```
hvy: A nitpick but let's be consistent with the whitespaces around the operator.
```suggestion
            return a ** b
```

hvy:
Are you able to look at the Jenkins failures? There are some tolerance errors with fp32, conventional errors and also some still unaddressed comment above. Let me know if things are unclear.

dido1998:
@hvy I just saw the jenkins errors. Currently, I am reducing the tolerance for float16, should I reduce it for all dtypes?

hvy:
It depends on how often they fail. Tolerances should be relaxed to reduce the failure rate but not more than necessary. They should also be considered independently for each dtype. Let me rerun the CI as the previous results seem to be gone!

hvy:
Ah, sorry but could you resolve the conflicts?

hvy:
Jenkins, test this please.

chainerx_cc/chainerx/native/native_device/misc.cc
```diff
@@ -16,6 +16,56 @@ namespace chainerx {
 namespace native {
 namespace {
 
+class NativePowKernel : public PowKernel {
+public:
+    void Call(const Array& x1, const Array& x2, const Array& out) {
```
hvy: Ops, missing `override` here.
```suggestion
    void Call(const Array& x1, const Array& x2, const Array& out) override {
```

chainerx_cc/chainerx/native/native_device/misc.cc
```diff
@@ -16,6 +16,56 @@ namespace chainerx {
...
+            Elementwise<const T, const T, T>(Impl{}, x1, x2, out);
+        });
+    }
+};
+
+CHAINERX_NATIVE_REGISTER_KERNEL(PowKernel, NativePowKernel);
+
+class NativePowASKernel : public PowASKernel {
+public:
+    void Call(const Array& x1, Scalar x2, const Array& out) {
```
hvy: Ditto.
```suggestion
    void Call(const Array& x1, Scalar x2, const Array& out) override {
```

chainerx_cc/chainerx/native/native_device/misc.cc
```diff
@@ -16,6 +16,56 @@ namespace chainerx {
...
+            Elementwise<const T, T>(Impl{static_cast<T>(x2)}, x1, out);
+        });
+    }
+};
+
+CHAINERX_NATIVE_REGISTER_KERNEL(PowASKernel, NativePowASKernel);
+
+class NativePowSAKernel : public PowSAKernel {
+public:
+    void Call(Scalar x1, const Array& x2, const Array& out) {
```
hvy: Ditto.
```suggestion
    void Call(Scalar x1, const Array& x2, const Array& out) override {
```

chainerx_cc/chainerx/routines/math.cc
```diff
@@ -878,6 +880,88 @@ Array Sqrt(const Array& x) {
...
+    if (BackwardBuilder::Target bt = bb.CreateTarget(0)) {
+        bt.Define([out_tok = bb.RetainOutput(0), x1](BackwardContext& bctx) {
+            const Array& out = bctx.GetRetainedOutput(out_tok);
+
+            bctx.input_grad() = *bctx.output_grad() * out * std::log(static_cast<double>(x1));
+        });
+    }
+    bb.Finalize();
+    return out;
+}
```
hvy: Regarding using `Binary` you can take a look at the latest master as it's has division for scalar / array, i.e. "SA" which should be quite similar to this?

tests/chainerx_tests/unit_tests/routines_tests/test_math.py
```diff
@@ -2431,6 +2431,124 @@ def func(self, xp, a):
...
+        if self.is_module:
+            if self.is_scalar_rhs:
+                return xp.power(a, scalar)
+            else:
+                return xp.power(scalar, a)
+        else:
+            if self.is_scalar_rhs:
+                return a ** scalar
+            else:
+                return scalar ** a
```
hvy: Shouldn't we test non-floats such as integers too? Note that an int raised to a negative int will invoke a `ValueError` for NumPy while for CuPy it won't. I think it's okay for ChainerX to follow CuPy.

hvy:
Sorry for the delay but I added some more comments. If you think it's difficult, maybe I can take over this PR and continue with the rest. Please let me know in that case.

dido1998:
I have incorporated some of the changes. I think you can take over this pr now. In that case, I will begin working on the GSOC project.

chainerx_cc/chainerx/numeric.h
```diff
@@ -63,4 +63,13 @@ CHAINERX_DEFINE_NATIVE_FLOAT16_FALLBACK_UNARY(Log, std::log)
...
 CHAINERX_DEFINE_NATIVE_FLOAT16_FALLBACK_UNARY(Sqrt, std::sqrt)
 
+template <typename T>
+inline T Pow(T x1, T x2) {
+    return std::pow(x1, x2);
+}
+template <>
+inline chainerx::Float16 Pow<chainerx::Float16>(chainerx::Float16 x1, chainerx::Float16 x2) {
+    return Float16{std::pow(static_cast<float>(x1), static_cast<float>(x2))};
+}
```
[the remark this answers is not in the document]
hvy: Fixed.

chainerx_cc/chainerx/python/routines.cc
hvy: Fixed.

tests/chainerx_tests/unit_tests/routines_tests/test_math.py
hvy: Fixed.

chainerx_cc/chainerx/routines/math.cc
hvy: Fixed.

chainerx/__init__.pyi
```diff
@@ -322,6 +322,10 @@ class ndarray:
 
     def __neg__(self) -> ndarray: ...
 
+    def __pow__(self, arg0: tp:Any) -> ndarray: ...
+
+    def __rpow__(self, arg0: tp.Any) -> ndarray: ...
```
asi1024: How about sorting in lexicographical order of attribute names for consistency?

hvy: Thanks for the comment, I didn't notice that. Should be fixed now.

chainerx_cc/chainerx/scalar.h
```diff
@@ -104,6 +104,13 @@ class Scalar {
         return this->UnwrapAndCast<int64_t>() + rhs.UnwrapAndCast<int64_t>();
     }
 
+    Scalar operator-(const Scalar& rhs) const {
```
hvy: Introduced for this computation https://github.com/chainer/chainer/pull/6496/files#diff-82d97e7f86c1e4c6e00277b3f56cbfd8R918

chainerx_cc/chainerx/cuda/numeric.cuh
```diff
@@ -77,5 +77,22 @@ CHAINERX_DEFINE_CUDA_FLOAT16_FALLBACK_UNARY(Log10, std::log10)
 CHAINERX_DEFINE_CUDA_FLOAT16_FALLBACK_UNARY(Sqrt, std::sqrt)
 CHAINERX_DEFINE_CUDA_FLOAT16_FALLBACK_UNARY(Fabs, std::fabs)
 
+template <typename T>
+__device__ inline T Power(T x1, T x2) {
+    return T{rint(pow(static_cast<double>(x1), static_cast<double>(x2)))};
```
asi1024: How about implementing with loop with O(log(x2)) times?
`std::pow` sometime causes a little error.
```py
>>> numpy.power(numpy.array([123456789]), numpy.array([2]))
array([15241578750190521])
>>> chainerx.power(chainerx.array([123456789]), chainerx.array([2]))
array([15241578750190520], shape=(1,), dtype=int64, device='native:0')
```

chainerx_cc/chainerx/routines/math.cc
```diff
@@ -878,6 +879,72 @@ Array Sqrt(const Array& x) {
...
+            const Array& x1 = bctx.GetRetainedInput(x1_tok);
+            const Array& out = bctx.GetRetainedOutput(out_tok);
+            const Array& gin = *bctx.output_grad() * out * Log(x1);
+            bctx.input_grad() = x2_dtype != gin.dtype() ? gin.AsType(x2_dtype) : gin;
+        });
+    }
+    bb.Finalize();
+}
+
+void PowerASImpl(const Array& x1, Scalar x2, const Array& out) {
```
asi1024: Is this behavior intended?
```py
>>> chainerx.power(chainerx.array([2]), -1)
array([0], shape=(1,), dtype=int64, device='native:0')
```
How about raising an error in such a case?

hvy:
Some offline discussions lead to the following which I'll address soon.
- Int raised to int should be 0, similar to CuPy, for both native and cuda backends.
- Boolean base/power -> Error. There are little to no use-cases. Reconsider if such cases would appear.
- With the native backend, do not use `std::pow` for integers.

hvy:
Made the above changes, PTAL.

chainerx_cc/chainerx/routines/math.cc
hvy: Discussed to leave it as is. Similar to e.g. CuPy.

chainerx_cc/chainerx/cuda/numeric.cuh
hvy: Thanks for the suggestion, fixed accordingly.

chainerx_cc/chainerx/routines/math.cc
```diff
@@ -906,6 +910,10 @@ void PowerImpl(const Array& x1, const Array& x2, const Array& out) {
 }
 
 void PowerASImpl(const Array& x1, Scalar x2, const Array& out) {
+    if (x1.dtype() == Dtype::kBool || x2.kind() == DtypeKind::kBool) {
```
asi1024: How about using `CheckArithmeticDtypes`?

chainerx_cc/chainerx/native/native_device/misc.cc
```diff
@@ -53,6 +53,46 @@ class NativeSqrtKernel : public SqrtKernel {
 
 CHAINERX_NATIVE_REGISTER_KERNEL(SqrtKernel, NativeSqrtKernel);
 
+CHAINERX_NATIVE_REGISTER_ELTWISE_DTYPE_BINARY_KERNEL(PowerKernel, { out = chainerx::Power(x1, x2); }, VisitNumericDtype);
```
asi1024: This device implementation might be better to be added in `arithmetic.cc`

chainerx_cc/chainerx/routines/math.cc
hvy: Fixed.

chainerx_cc/chainerx/native/native_device/misc.cc
hvy: Agree, fixed. We need to fix these files really.

hvy:
Fixed, PTAL.

tests/chainerx_tests/unit_tests/routines_tests/test_math.py
```diff
@@ -2431,6 +2431,233 @@ def func(self, xp, a):
...
+        'in_dtypes,out_dtype': (
+            _make_same_in_out_dtypes(2, chainerx.testing.float_dtypes)),
+        'input_lhs': ['random', float('inf'), -float('inf'), float('nan')],
+        'input_rhs': ['random', float('inf'), -float('inf'), float('nan')],
+        'is_module': [False],
+        'skip_backward_test': [True],
+        'skip_double_backward_test': [True],
+    })
+))
+class TestPower(BinaryMathTestBase, op_utils.NumpyOpTest):
```
asi1024: (Discussed with @hvy)
We decided to avoid using `random` backward/double-backward test and make a deterministic test because `pow` has many edge cases.
1. The gradient becomes so large when LHS is close to `0`.
1. `pow` returns a finite value for discrete inputs. (LHS >= 0 or fmod(RHS, 1) == 0)

hvy:
Jenkins, test this please.

hvy:
Jenkins, test this please.

hvy:
@asi1024 could you take a look at the test failures?

hvy:
Jenkins, test this please.

hvy:
Jenkins, test this please.

hvy:
@asi1024 do you want or should I resolve the conflicts?

hvy:
Jenkins, test this please.

hvy:
Thanks everyone for the long running PR. Tests passed and it LGTM.

chainerx/__init__.pyi
```diff
@@ -324,6 +324,8 @@ class ndarray:
 
     def __or__(self, arg0: tp.Any) -> ndarray: ...
 
+    def __pow__(self, arg0: tp:Any) -> ndarray: ...
```
toslunar: The type of `arg0` should be `tp.Any`.

PR #6506, opened 2019-03-13
Deprecate old NCCL versions and related communicators

kuenishi:
From https://github.com/chainer/chainer/pull/6482#issuecomment-472272599 :
* We'll deprecate NCCL 2.2 and older version at Chainer v6
* We'll drop NCCL 2.2 support at Chainer v7
* We'll change Chainer(MN) according to deprecation schedule above.

I concluded we don't have to change any of cupy code and range of supported NCCL versions. Thoughts?

### DIFF
--- a/tests/chainermn_tests/communicator_tests/test_communicator.py
+++ b/tests/chainermn_tests/communicator_tests/test_communicator.py
@@ -487,7 +487,7 @@ class TestDifferentDtype(unittest.TestCase):
 
     def setup(self, gpu):
         if gpu:
-            self.communicator = chainermn.create_communicator('hierarchical')
+            self.communicator = chainermn.create_communicator('flat')
             self.device = self.communicator.intra_rank
             chainer.cuda.get_device_from_id(self.device).use()
         else:
@@ -767,7 +767,7 @@ class TestNonContiguousArray(unittest.TestCase):
 
     def setup(self, gpu):
         if gpu:
-            self.communicator = chainermn.create_communicator('hierarchical')
+            self.communicator = chainermn.create_communicator('flat')
             self.device = self.communicator.intra_rank
             chainer.cuda.get_device_from_id(self.device).use()
         else:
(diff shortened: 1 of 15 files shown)


kuenishi:
Still wondering:
* Are there use cases `two_dimensional` cannot be replaced with `flat` or `pure_nccl`?
* Does NCCL works with Omni-path equipped environment? (if yes, we also cannot deprecate hierarchical yet)
* Has Intel-MPI become CUDA-aware? (if yes, we'll deprecate non-cuda aware comm instead of hierarchical)

kuenishi:
Jenkins, test this please.

shu65:
We are confirming whether users are often using these communicators.

keisukefukuda:
Some feedbacks from users:

* Xeon Phi (Oakforest Pacs at UTokyo): `flat` is enough. Basically Intel MPI takes care of low-level optimization. Actually even more advanced optimizations are possible, but they depend on the architecture details and generations. It does not make much sense to optimize collective operations on ChainerMN layer.
* Intel Omni-Path environment (Tsubame 3.0 at Tokyo Tech): NCCL works. No need of Hierarchical, Two-dimensional communicators.

As a result, I think that we can safely mark the communicators deprecated.

shu65:
Thank you!

shu65:
I confirmed that the unittests for ChainerMN are passed.

shu65:
maehashi-san's comments:

1. avoid CUDA API call in import 
https://github.com/chainer/chainer/pull/6506/files#diff-004aa318ddb12cb358a0ea7e38d8e562R16
2. add a description to Upgrade Guide

kuenishi:
Jenkins, test this please.

kuenishi:
Jenkins, test this please.

shu65:
Jenkins, test this please.

tests/chainermn_tests/communicator_tests/test_communicator.py
```diff
@@ -772,3 +772,16 @@ def test_alltoall_cpu(self):
...
+
+@chainer.testing.attr.gpu
+def test_deprecation():
+    with chainer.testing.assert_warns(DeprecationWarning):
+        chainermn.create_communicator('hierarchical')
+
+    with chainer.testing.assert_warns(DeprecationWarning):
+        chainermn.create_communicator('two_dimensional')
+
+    if mpi_comm.is_intra:
```
shu65: I have gotten the same error.
How about changing to the following?
https://github.com/chainer/chainer/blob/58a2a9696d4884e0b1d8eb777c55f63cabab2a38/tests/chainermn_tests/communicator_tests/test_communicator.py#L221-L224

shu65:
Jenkins, test this please.

PR #6509, opened 2019-03-13
Add reference and warning to `F.swish` document

fiarabbit:
Improved documentation.
- Added reference to the original paper, because `F.swish` did not contain the reference.
- Added warning to use `L.Swish`, because Swish is an activation function which should be trained.


Crissman:
Given the original paper had `beta` as a trainable variable, I think warning is appropriate here. @toslunar let's discuss.

Crissman:
Jenkins, test this please.

toslunar:
We discussed it offline:+1:

Crissman:
Merging!

Issue #6514, opened 2019-03-14
Typo in docstring for classification_summary


yewang:
There appears to be a typo in the docstring for the "classification_summary" function (located in functions/evaluation/classification_summary.py).

The correct definition for recall is TP / (TP + FN), however the current docstring writes "TN" instead of "FN".

Also, after amending the definition, a few lines later should mention the definition of "FN" alongside "TP", "FP", and "TN".

The actual code seems to make the correct calculation.

toslunar:
Fixed in #6515.

PR #6515, opened 2019-03-14
Fixes typo in docstring for classification_summary

yewang:
Address issue https://github.com/chainer/chainer/issues/6514 by fixing the typo in the definition for recall in the docstring. Also add definition for "fn" as false negatives".

### DIFF
--- a/chainer/functions/evaluation/classification_summary.py
+++ b/chainer/functions/evaluation/classification_summary.py
@@ -75,12 +75,13 @@ def classification_summary(y, t, label_num=None, beta=1.0, ignore_label=-1):
     This function calculates the following quantities for each class.
 
     - Precision: :math:`\\frac{\\mathrm{tp}}{\\mathrm{tp} + \\mathrm{fp}}`
-    - Recall: :math:`\\frac{\\mathrm{tp}}{\\mathrm{tp} + \\mathrm{tn}}`
+    - Recall: :math:`\\frac{\\mathrm{tp}}{\\mathrm{tp} + \\mathrm{fn}}`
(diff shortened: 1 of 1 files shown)


kmaehashi:
Jenkins, test this please.

chainer/functions/evaluation/classification_summary.py
```diff
@@ -75,12 +75,13 @@ def classification_summary(y, t, label_num=None, beta=1.0, ignore_label=-1):
...
 
     - Precision: :math:`\\frac{\\mathrm{tp}}{\\mathrm{tp} + \\mathrm{fp}}`
-    - Recall: :math:`\\frac{\\mathrm{tp}}{\\mathrm{tp} + \\mathrm{tn}}`
+    - Recall: :math:`\\frac{\\mathrm{tp}}{\\mathrm{tp} + \\mathrm{fn}}`
     - F beta Score: The weighted harmonic average of Precision and Recall.
     - Support: The number of instances of each ground truth label.
 
-    Here, ``tp``, ``fp``, and ``tn`` stand for the number of true positives,
-    false positives, and true negative, respectively.
+    Here, ``tp``, ``fp``, ``tn``, and ``fn`` stand for the number of true
```
toslunar: Would you like to mention the number of true negatives (`tn`), while it doesn't appear in the formulas?

yewang: Good point. I think it's possible to remove mention of `tn`. However, one could always keep it for the sake of completeness or in case other equations for other metrics are added later.

I don't have strong feelings about it either way.

The more essential part of this PR is to fix the typo in the equation above.

PR #6516, opened 2019-03-14
Add option for caching `x_hat` to GroupNormalization

crcrpar:
This PR aims at enabling GroupNormalization FunctionNode to cache `x_hat` in order to skip normalizing `x` of backward by adding an option.

### DIFF
--- a/chainer/links/normalization/group_normalization.py
+++ b/chainer/links/normalization/group_normalization.py
@@ -32,7 +32,7 @@ class GroupNormalization(link.Link):
             If a scalar, the vector is filled by it.
             If ``numpy.ndarray``, the vector is set by it.
         cache_x_hat (bool): If ``True``, cache normalized input ``x`` for
-            faster backward. The default value is ``Nonee`` because Group
+            faster backward. The default value is ``False`` because Group
             Normalization is a normalization technique for training with
             very small batchsizes.
 
--- a/chainer/links/normalization/group_normalization.py
+++ b/chainer/links/normalization/group_normalization.py
@@ -31,6 +31,10 @@ class GroupNormalization(link.Link):
             If ``None``, then the vector is filled by 0.
             If a scalar, the vector is filled by it.
             If ``numpy.ndarray``, the vector is set by it.
+        cache_x_hat (bool): If ``True``, cache normalized input ``x`` for
+            faster backward. The default value is ``Nonee`` because Group
+            Normalization is a normalization technique for training with
+            very small batchsizes.
 
     Attributes:
         groups (int): The number of channel groups.
@@ -42,18 +46,19 @@ class GroupNormalization(link.Link):
     """
 
     def __init__(self, groups, size=None, eps=1e-5, initial_gamma=None,
-                 initial_beta=None):
+                 initial_beta=None, cache_x_hat=False):
         super(GroupNormalization, self).__init__()
         if initial_gamma is None:
             initial_gamma = 1
         if initial_beta is None:
             initial_beta = 0
 
+        self.groups = groups
+        self.eps = eps
+        self.cache_x_hat = cache_x_hat
         with self.init_scope():
-            self.groups = groups
             self.gamma = variable.Parameter(initial_gamma)
             self.beta = variable.Parameter(initial_beta)
-            self.eps = eps
 
         if size is not None:
             self._initialize_params(size)
@@ -85,4 +90,4 @@ class GroupNormalization(link.Link):
             self._initialize_params(channels)
 
         return group_normalization.group_normalization(
-            x, self.groups, self.gamma, self.beta, self.eps)
+            x, self.groups, self.gamma, self.beta, self.eps, self.cache_x_hat)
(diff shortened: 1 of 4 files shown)


hvy:
Do you know if we have other functions with similar performance/memory-tradeoff customization interfaces? Just as a reference.

crcrpar:
I do not know such functions.

This PR is a step to implement Layer Normalization and Instance Normalization on top of GoupNormalization FunctionNode for me.

hvy:
So it is not just about performance optimization? Could you elaborate a bit your latter motivation?

crcrpar:
The motivation is to optimizer performance of course.

As #5924 says, layer normalization and instance normalization can be considered as a special case of group normalization. So, I thought it would be easier to maintain/optimize implementations of both normalizations if they were implemented on top of the GroupNormalization with this caching option.

crcrpar:
**Condition**
- batch size = 128
- groups = 16

**Script & Model**
- [examples/cifar/train_cifar.py
](https://github.com/chainer/chainer/blob/master/examples/cifar/train_cifar.py)
- [examples/cifar/models/VGG.py](https://github.com/chainer/chainer/blob/master/examples/cifar/models/VGG.py): All BNs are replaced with GroupNormalization

### GPU
Without cache
```
CupyMemoryProfileHook
                             FunctionName        UsedBytes  AcquiredBytes  Occurrence
                       GroupNormalization  44375554048.00B  264241152.00B        2212
                                    _XHat  11079680000.00B          0.00B        1106
                          _ScaleShiftGrad  11082754048.00B   58720256.00B        1106
                                _XHatGrad  11097600000.00B   92274688.00B        1106
...
TimerHook
                             FunctionName  ElapsedTime  Occurrence
                       GroupNormalization      2.06sec        2212
                                    _XHat      0.22sec        1106
                          _ScaleShiftGrad      0.56sec        1106
                                _XHatGrad      0.26sec        1106
```

With cache
```
                             FunctionName        UsedBytes  AcquiredBytes  Occurrence
                       GroupNormalization  33295874048.00B  222298112.00B        2212
                                    _XHat            0.00B          0.00B        1106
                          _ScaleShiftGrad  11082754048.00B          0.00B        1106
                                _XHatGrad  11097600000.00B   92274688.00B        1106
...
TimerHook
                             FunctionName  ElapsedTime  Occurrence
                       GroupNormalization      1.84sec        2212
                                    _XHat      0.02sec        1106
                          _ScaleShiftGrad      0.56sec        1106
                                _XHatGrad      0.26sec        1106
```

---

### CPU
Without cache
```
                             FunctionName  ElapsedTime  Occurrence
                       GroupNormalization      9.23sec         280
                                    _XHat      0.78sec         140
                          _ScaleShiftGrad      1.73sec         140
                                _XHatGrad      2.88sec         140
```

With cache
```
                             FunctionName  ElapsedTime  Occurrence
                       GroupNormalization      8.41sec         280
                                    _XHat      0.00sec         140
                          _ScaleShiftGrad      1.72sec         140
                                _XHatGrad      2.85sec         140
```

hvy:
For the record, I added the discussion label since I wanted to discuss whether this interface is okay.

toslunar:
The interface looks good to me.

@hvy Do you have in mind for concerns or other interfaces?

chainer/links/normalization/group_normalization.py
```diff
@@ -31,6 +31,10 @@ class GroupNormalization(link.Link):
             If ``None``, then the vector is filled by 0.
             If a scalar, the vector is filled by it.
             If ``numpy.ndarray``, the vector is set by it.
+        cache_x_hat (bool): If ``True``, cache normalized input ``x`` for
+            faster backward. The default value is ``Nonee`` because Group
```
toslunar:
```suggestion
            faster backward. The default value is ``False`` because Group
```

crcrpar: Thank you.

crcrpar:
If this PR is inappropriate from the perspective of the maintenance cost, interface, design, or something other, feel free to close.

hvy:
Sorry for the delayed comment. These types of trade-offs are nice for users to be able to control but it stretches beyond this function and applies to others as well (?) so I was being careful. I like the idea that this convention can be generalized to `cache_<any intermediate var>` but am not sure that's the level of control we want to give the function caller. It slightly feels like exposing internals to the public API. Is "x_hat" not to be treated as an internal maybe?

hvy:
I'm also wondering what this interface will look like to users using e.g. ChainerX; must it also provide a similar API or is it okay that the argument is ignored? 🤔

crcrpar:
IMO, group normalization (GN) differs from the other `Link`s in that GN
derives partially from the concern of memory consumption of Batch
Normalization. Also, given that previously GN did cache the `x_hat`
(because it was implemented on the top of the `F.batch_normalization`),
this option makes sense.

crcrpar:
> what this interface will look like to users using e.g. ChainerX
Do you mean ChainerX C++ API in the future, or ChainerX use in conventional
Chainer API?

As to the former, I’d like the implementation to support this cache
option.  As to the latter, as we currently don’t support ChaimerX in GN, we
have to talk about, for instance, how to make ChainerX path of
`F.group_normalization` compatible with the existing numpy/cupy
implementation.

toslunar:
As Chainer is under the maintenance phase, unfortunately we decided not to merge this feature. Thanks again for your contribution.

PR #6519, opened 2019-03-14
Document fix in default initializer

hvy:
A minor change to the documentation for the default initializer behavior in case of `None`. 

Initializers can be usually be "specified" as arguments in `Link.__init__` and I think this change makes it a bit clearer what it means.


Crissman:

Crissman:
Jenkins issue is unrelated. LGTM. Merging.

PR #6529, opened 2019-03-14
[WIP] Add `dtype` argument in `F.sum`

niboshi:
For reducing overheads in some cases.


niboshi:
This feature is not only for speed.
It simplifies a code where casting is necessary.
(as in `F.softmax_cross_entropy`).

PR #6530, opened 2019-03-14
Include cuDNN function name in error message

niboshi:
cuDNN functions tend to be difficult to debug.
This PR includes cuDNN function names in error messages.


hvy:
What is the status of this PR?

niboshi:
Rebased and removed WIP. PTAL

niboshi:
@hvy Can you review?

chainerx_cc/chainerx/cuda/cudnn.h
```diff
@@ -10,21 +12,40 @@
 #include "chainerx/macro.h"
 #include "chainerx/stack_vector.h"
 
+#define CHAINERX_CUDA_CUDNN_CALL_WITH_HANDLE(handle, func, args...)                                                                    \
```
hvy: Actually, I don't think we need this macro? We can simply use `CHAINERX_CUDA_CUDNN_CALL` passing the handle similar to any other argument to the cuDNN routine. It's easier from the caller perspective I think.

chainerx_cc/chainerx/cuda/cudnn.h
```diff
@@ -10,21 +12,40 @@
...
     cudnnStatus_t error() const noexcept { return status_; }
 
 private:
     cudnnStatus_t status_{};
 };
 
-void CheckCudnnError(cudnnStatus_t status);
+namespace cudnn_detail {
+
+inline void CheckCudnnError(const char* func_name, cudnnStatus_t status) {
```
hvy: A detail but how about passing `func_name` as the later argument as it's less significant (in most cases never used).

hvy:
Sorry for the delay, I wrote some comments. It's worth noting that we might want similar features for cuBLAS and soon cuSOLVER?

chainerx_cc/chainerx/cuda/cudnn.h
niboshi: I don't think `CudnnHandle::Call` can be bypassed.
It has a concurrency control.

chainerx_cc/chainerx/cuda/cudnn.h
niboshi: What do you mean by never used? It's mandatory.

hvy: It's only used when the condition evaluates to true, that's what I meant...

niboshi: Oh does that matter? `func_name` is more "stable" parameter, so I think it's natural to come first.

chainerx_cc/chainerx/cuda/cudnn.h
hvy: Ah, that is true of course. Thanks.

chainerx_cc/chainerx/cuda/cudnn.h
hvy: I thought so. Or at least, I'm not sure what you mean with "stable", but the intent of this functions is to check the status, doesn't `status` have a higher importance? 🤔

emcastillo:
Jenkins, test this please

emcastillo:
Some segfaults? 
```
Aborted                 (core dumped) 
```

emcastillo:
Let's close this, Thanks!

PR #5489, opened 2018-10-15
[WIP] Snapshot extension with automatic resume

kmaehashi:
To better handle preemption caused by cluster manager, provide option to automatically resume from the last-saved snapshot.

PR #6531, opened 2019-03-15
AutoSnapshot

kuenishi:
In the pursuit for more intuitive and reliable snapshotting


kmaehashi:
Related: #5489

kuenishi:
After discussion with @beam2d , we've got following direction:

* add `autoload=False` and `snapshot_retain=-1`-like options to both `snapshot` and `snapshot_object` functions.
* if `snapshot_retain > 0` stale snapshots are automatically removed every time a new snapshot was created.
* If `autoload=True` the extension automatically finds the newest snapshot that matches the filename pattern (regarding timestamp) and loads the content to the object.

Issue #6168, opened 2019-02-02
Support mixed precision training in chainer examples


MannyKayy:
Not technically an issue but it maybe worthwhile to warn users of overflow issues if they enable fp16 and have exponential functions in their model.

The older VAE example code directly calls [F.gaussian](https://github.com/chainer/chainer/blob/ef13d538969768497d7a0cfb4aec58ec0f3dac25/examples/vae/net.py#L58) resultings in NaNs with FP16.

The latest VAE example code relies on chainer.distributions, which doesn't yet support FP16 and throws an error whenever [D.Normal()](https://github.com/chainer/chainer/blob/master/examples/vae/net.py#L68) is called, because cupy's random.normal function only operates with float32's. But once/if it does support FP16, similar issues will be encountered. Just as with the Adam FP16 warning, the F.exp overflow issues may be worth noting somewhere.

MannyKayy:
After testing some more of the chainer examples (specifically those with functions that support all dtypes), it appears that most of them are unstable to train with anything but "float32".

takagi:
Thanks for your suggestion. I will check the examples with FP16 enabled and make a plan to provide them work properly.

takagi:
Example|FP16 supported|Work with FP16|Note
--|--|--|--
caffe_export|||
chainermn|||
cifar|||
dcgan|||
glance|||
image_captioning|||
imagenet|||
memnn|||
mnist|||
modelzoo|||
pix2pix|||
pos|||
ptb|||
reinforcement_learning|||
sentiment|||
seq2seq|||
serialization|||
static_graph_optimizations|||
text_classification|||
vae|||
wavenet|||
word2vec|||

anaruse:
I've just run imagenet example with RN50 in fp32 and fp16, and found loss value became `NaN` soon when fp16.  This is a bit strange, since it worked fine when I did similar test a few month ago.  

```
config.dtype: float32
epoch       iteration   main/loss   validation/main/loss  main/accuracy  validation/main/accuracy  lr
1           32          4.461       38.1201               0.0458984      0.0414201                 0.01
2           64          3.93517     3.80036               0.0710449      0.0815921                 0.01
3           96          3.37447     3.25447               0.125732       0.151766                  0.01
```

```
config.dtype: float16
epoch       iteration   main/loss   validation/main/loss  main/accuracy  validation/main/accuracy  lr
1           32          nan         nan                   0              0                         0.01
2           64          nan         nan                   0              0                         0.01
[...]

PR #6536, opened 2019-03-15
Fix wavenet example to support the default dtype

takagi:
Rel. #6168.

This PR fixes the wavenet example to support the default dtype feature, especially performing in FP16 mode.


niboshi:
Jenkins, test this please

niboshi:
LGTM

PR #6540, opened 2019-03-15
Fix a off-by-one in slicing of chainerx::Array

shinh:


shinh:
I confirmed the test is valid (i.e., it failed without the fix), PTAL.

hvy:
jenkins, test this please.

hvy:
Thanks for the quick fix. The CI failure is unrelated and it LGTM.

PR #6541, opened 2019-03-15
Implement array vs array functionality to chainerx.minimum

aksub99:
Implemented array vs array functionality to chainerx.minimum
Ref. #6423


aksub99:
@asi1024 Could you please let me know if the Travis failures are related to this PR?

asi1024:
Thanks for the PR! Could you fix for ClangFormat like the following link?
https://travis-ci.org/chainer/chainer/jobs/506722734#L753

aksub99:
@asi1024 Made the required changes.

asi1024:
Could you add a test in `tests/chainerx_tests/unit_tests/routines_tests/test_math.py`?

aksub99:
Unintentionally committed some mistakes in the python test for chainerx.minimum . Am in the process of rectifying them.

chainerx_cc/chainerx/routines/math.cc
```diff
@@ -532,6 +532,37 @@ Array IfGreaterElse(const Array& x1, Scalar x2, Scalar pos, const Array& neg) {
...
+        NoBackpropModeScope scope{};
+        x1.device().IfGreaterElseAAAA(x1, x2, pos, neg, out);
+    }
+
+    {
+        BackwardBuilder bb{"if_greater_else", {pos, neg}, out};
+        if (BackwardBuilder::Target bt = bb.CreateTarget(0)) {
+            bt.Define([x1 = x1.AsGradStopped(), x2](BackwardContext& bctx) {
+                const Array& gout = *bctx.output_grad();
+                bctx.input_grad() = IfGreaterElse(x1, x2, ZerosLike(gout, gout.device()), gout);
```
asi1024:
```suggestion
                bctx.input_grad() = IfGreaterElse(x1, x2, gout, ZerosLike(gout, gout.device()));
```

chainerx_cc/chainerx/routines/math.cc
```diff
@@ -544,6 +575,10 @@ Array Minimum(const Array& x1, Scalar x2) {
 
 Array Minimum(Scalar x1, const Array& x2) { return Minimum(x2, x1); }
 
+Array Minimum(const Array& x1, const Array& x2) {
+    return IfGreaterElse(x1, x2, x2, x1);  // x1 > x2 ? x2 : x1
```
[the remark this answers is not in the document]
aksub99: Sure! Will work on it @asi1024 .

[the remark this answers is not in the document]
aksub99: @asi1024 Do I define `BroadcastQuaternary` to handle 4 inputs in case of `Maximum` and `Minimum` ?

> It fails on the tests for different shapes. Could you broadcast the inputs with `BroadcastBinary`?

aksub99:
`
/home/akshay/chainer/chainerx_cc/chainerx/routines/math.cc:61:7: note:   template argument deduction/substitution failed:
/home/akshay/chainer/chainerx_cc/chainerx/routines/math.cc:596:54: note:   couldn't deduce template parameter ‘Impl’
     return BroadcastQuaternary(&IfGreaterElse, x1, x2);  // x1 > x2 ? x2 : x1
`

I get this error after making the above changes. I think this is probably because the return type of `IfGreaterElse` is `Array` and not void and hence is not directly deduced. @asi1024 Could you suggest a method of overcoming this?

asi1024:
Hmm, how about defining `MinimumImpl(const Array& x1, const Array& x2, const Array& out)` and using `BroadcastBinary`? we may not need to define `BroadcastQuaternary` only for this case.

aksub99:
@asi1024 Could you please review changes?

chainerx_cc/chainerx/routines/math.cc
```diff
@@ -525,6 +525,69 @@ Array IfGreaterElse(const Array& x1, Scalar x2, Scalar pos, const Array& neg) {
...
+void MinimumImpl(const Array& x1, const Array& x2, const Array& out) {
+    CheckEqual(x1.dtype(), x2.dtype());
+    CheckEqual(x1.shape(), x2.shape());
+    {
+        NoBackpropModeScope scope{};
+        x1.device().IfGreaterElseAAAA(x1, x2, x2, x1, out);
+    }
+
+    {
+        BackwardBuilder bb{"if_greater_else", {x2, x1}, out};
```
asi1024: This backward implementation seems to duplicate with `IfGreaterElse`'s one. How about defining `void IfGreaterElseImpl(const Array& x1, const Array& x2, const Array& pos, const Array& neg, const Array& out)` and just calling `IfGreaterElseImpl(x1, x2, x2, x1, out)` in this function?

aksub99: That makes sense. Thanks for the input @asi1024 .I'll work on that.

aksub99: Could `IfGreaterElse` be removed then?

aksub99:
Is this fine @asi1024 ?

chainerx_cc/chainerx/routines/math.cc
```diff
@@ -525,6 +525,77 @@ Array IfGreaterElse(const Array& x1, Scalar x2, Scalar pos, const Array& neg) {
...
+    CheckEqual(x1.dtype(), x2.dtype());
+    CheckEqual(x1.shape(), x2.shape());
+    Array out = EmptyLike(x1, x1.device());
+    {
+        NoBackpropModeScope scope{};
+        x1.device().IfGreaterElseAAAA(x1, x2, pos, neg, out);
+    }
+
+    {
+        BackwardBuilder bb{"if_greater_else", {pos, neg}, out};
```
asi1024: This backward implementation also duplicates with `IfGreaterElseImpl`. How about the following implementation?
```
CheckEqual(x1.dtype(), x2.dtype());
CheckEqual(x1.shape(), x2.shape());
Array out = EmptyLike(x1);
IfGreaterElseImpl(x1, x2, pos, neg, out);
return out;
```

aksub99: Defining `Array out` within `MinimumImpl` shadows the `out` parameter. Also, the return statement will require `Array` return type instead of `void`. So, would an implementation like this be sufficient?
`
CheckEqual(x1.dtype(), x2.dtype());
CheckEqual(x1.shape(), x2.shape());
IfGreaterElseImpl(x1, x2, pos, neg, out);
`

aksub99:
Could you please have a look at the question above @asi1024 ?

chainerx_cc/chainerx/routines/math.cc
asi1024: The current implementation of `MinimumImpl` seems to be OK, but both `IfGreaterElse` and `IfGreaterElseImpl` have backward implementations.

aksub99: @asi1024 Sorry, I probably misunderstood your previous comment. I will make the required modifications to `IfGreaterElseImpl`.

aksub99: > This backward implementation also duplicates with `IfGreaterElseImpl`. How about the following implementation?
> 
> ```
> CheckEqual(x1.dtype(), x2.dtype());
> CheckEqual(x1.shape(), x2.shape());
> Array out = EmptyLike(x1);
> IfGreaterElseImpl(x1, x2, pos, neg, out);
> return out;
> ```

As I understand it, `IfGreaterElseImpl` has to make use of `IfGreaterElse` in its backward. So, wouldn't replicating the backward implementations in `IfGreaterElse` and `IfGreaterElseImpl` be the only possible option apart from removing `IfGreaterElse` completely and only having `IfGreaterElseImpl` ? Sorry if I'm missing out on something.

asi1024: Ah, now I understand it. How about retaining a bitmask `x1 > x2` instead of raw data of `x1` and `x2`? It not only resolve the problem in my comment but also reduce memory consumption. The implementation will be as follows:
```
const Array& gout = *bctx.output_grad();
bctx.input_grad() = gout * mask.AsType(gout.dtype());  // or (LogicalNot(mask).Astype(...))
```

aksub99: Great. Thanks for the idea @asi1024 . Will make the required changes.

aksub99: @asi1024 Sorry for the delayed reply. Do you suggest retention of the bitmask in `IfGreaterElse` only and not `IfGreaterElseImpl` to avoid duplication of their backwards?

aksub99: mask_impl
@asi1024 This is my attempt at an implementation of your idea. Please let me know if any modifications can be made? Note: I've left `IfGreaterElseImpl` unchanged. (Still possesses the normal backward implementation).

asi1024: If you rewrite `IfGreaterElseImpl` in the same way, `IfGreaterElse` seems to be able to be removed.

asi1024: `mask` and `not_mask` should be of `bool` dtype. How about the following implementation?
```
Array mask = Greater(x1, x2);
Array not_mask = LogicalNot(mask);
```

asi1024: A while ago, some PRs about mixed-dtypes are merged in the master branch. Could you merge master branch and support mixed-dtypes in `chainerx.minimum`?

aksub99: I have made the required changes to the masks. I have also rebased the branch. Hope that is okay.

aksub99:
@asi1024 I will support mixed dtypes in the next commit.

aksub99:
@asi1024 Could you please review?

aksub99:
@asi1024 Could you please review?

chainerx_cc/chainerx/routines/math.cc
```diff
@@ -599,6 +637,14 @@ Array Minimum(const Array& x1, Scalar x2) {
 
 Array Minimum(Scalar x1, const Array& x2) { return Minimum(x2, x1); }
 
+Array Minimum(const Array& x1, const Array& x2) {
+    Dtype dtype = GetArithmeticResultDtype(x1, x2);
+    if (GetKind(dtype) != DtypeKind::kFloat) {
+        dtype = internal::GetDefaultDtype(DtypeKind::kFloat);
```
asi1024: This function seems better to return integers when integer arrays are given as the arguments, as NumPy's one does so.

chainerx_cc/chainerx/routines/math.cc
```diff
@@ -587,6 +588,43 @@ Array IfGreaterElse(const Array& x1, Scalar x2, Scalar pos, const Array& neg) {
 
 }  // namespace
 
+namespace {
+
+void IfGreaterElseImpl(const Array& x1, const Array& x2, const Array& pos, const Array& neg, const Array& out) {
+    CheckEqual(x1.dtype(), x2.dtype());
```
asi1024:
```suggestion
```

aksub99:
When I test my implementation locally, I get the following warning. @asi1024 Could you please take a look?
`
chainer/testing/_bundle.py:0
  /home/akshay/chainer/chainer/testing/_bundle.py:0: PytestWarning: cannot collect test class "TestMinimum_param_139_{x1_shape=(2,), x2_shape=(3, 2), in_dtypes=('float64', 'float64'), out_dtype='float64'}" because it has a __init__ constructor
`

tests/chainerx_tests/unit_tests/routines_tests/test_math.py
```diff
@@ -1729,3 +1729,58 @@ def test_max(is_module, xp, device, input, axis, dtype):
...
+            self.check_backward_options.update({'rtol': 3e-3, 'atol': 3e-3})
+            self.check_double_backward_options.update(
+                {'rtol': 3e-3, 'atol': 3e-3})
+
+    def generate_inputs(self):
+        x1_shape = self.x1_shape
+        x2_shape = self.x2_shape
+        x1_dtype = self.x1_dtype
+        x2_dtype = self.x2_dtype
+        x1 = numpy.random.uniform(-1, 1, x1_shape).astype(x1_dtype)
```
asi1024: How about using `array_utils.uniform`? The lower bound and upper bound are not appropriate for non-float inputs.

aksub99: Do you suggest having a wider range to allow for more integers? (For example, -5, 5)

tests/chainerx_tests/unit_tests/routines_tests/test_math.py
```diff
@@ -1729,3 +1729,58 @@ def test_max(is_module, xp, device, input, axis, dtype):
         return xp.max(a, axis)
     else:
         return a.max(axis)
+
+
+@chainer.testing.parameterize(*(
+    chainer.testing.product([
+        chainer.testing.from_pytest_parameterize(
```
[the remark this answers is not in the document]
aksub99: If I'm not mistaken, the following implementation is also very similar to the one I've used. Aren't the arguments taken in of the same type here as well? Sorry if I'm missing something here.
https://github.com/chainer/chainer/blob/7b40b4a4dae424bfba39d3c41077207b5c128f9d/tests/chainerx_tests/unit_tests/routines_tests/test_connection.py#L34

tests/chainerx_tests/unit_tests/routines_tests/test_math.py
aksub99: That makes sense. Will make the change. Thanks for the suggestion.

tests/chainerx_tests/unit_tests/routines_tests/test_math.py
```diff
@@ -1729,3 +1729,58 @@ def test_max(is_module, xp, device, input, axis, dtype):
...
+
+        if (x1_dtype == 'float16' or x2_dtype == 'float16'):
+            self.check_backward_options.update({'rtol': 3e-3, 'atol': 3e-3})
+            self.check_double_backward_options.update(
+                {'rtol': 3e-3, 'atol': 3e-3})
+
+    def generate_inputs(self):
+        x1_shape = self.x1_shape
+        x2_shape = self.x2_shape
+        x1_dtype = self.x1_dtype
```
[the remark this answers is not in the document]
aksub99: @asi1024 Aren't they defined from the parametrizations? The above assignment is very similar to the following.
https://github.com/chainer/chainer/blob/7b40b4a4dae424bfba39d3c41077207b5c128f9d/tests/chainerx_tests/unit_tests/routines_tests/test_connection.py#L104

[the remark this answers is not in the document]
asi1024: Sorry, I'm mistaken. they are defined certainly.

tests/chainerx_tests/unit_tests/routines_tests/test_math.py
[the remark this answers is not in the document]
asi1024: The line you commented seems needed at least. Could you try it?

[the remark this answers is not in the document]
aksub99: Sorry, I don't get what line you refer to. Could you please elaborate?

tests/chainerx_tests/unit_tests/routines_tests/test_math.py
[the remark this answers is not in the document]
asi1024: `self.x1_shape` and `self.x2_shape` are defined but dtypes are not.

tests/chainerx_tests/unit_tests/routines_tests/test_math.py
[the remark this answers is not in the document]
asi1024: We should parameterize devices at least. How about implementing like as follows?
```py
@op_utils.op_test(['native:0', 'cuda:0'])
@chainer.testing.parameterize(*(
    chainer.testing.product({
        'x1_shape,x2_shape': [
            ((3, 2), (3, 2)),
            ((), ()),
            ((3, 2), (3, 1)),
            ((2,), (3, 2)),
        ],
        'in_dtypes,out_dtype': _in_out_dtypes_arithmetic
    })
))
class TestMinimum(BinaryMathTestBase, op_utils.NumpyOpTest):

    def generate_inputs(self):
        x1_dtype, x2_dtype = self.in_dtypes
        x1 = array_utils.uniform(self.x1_shape, x1_dtype)
        x2 = array_utils.uniform(self.x2_shape, x2_dtype)
        return x1, x2

    def func(self, xp, x1, x2):
        return xp.minimum(x1, x2)
```
However, this test might be still weak. You can add more tests (e.g., tests for `inf` or `nan`) if you want to do.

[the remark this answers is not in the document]
aksub99: Could you please take a look at my changes? I've also added inf and nan tests to strengthen the tests.

chainerx_cc/chainerx/routines/math.cc
```diff
@@ -1,4 +1,5 @@
 #include "chainerx/routines/math.h"
+#include "chainerx/routines/logic.h"
```
asi1024: Duplicates with L22.
```suggestion
```

tests/chainerx_tests/unit_tests/routines_tests/test_math.py
```diff
@@ -2006,3 +2006,53 @@ def test_max_invalid_shapes_and_axis(device, array, axis, dtype, is_module):
...
+        'in_dtypes,out_dtype': (
+            _make_same_in_out_dtypes(2, chainerx.testing.float_dtypes)),
+        'input_lhs': ['random', float('inf'), -float('inf'), float('nan')],
+        'input_rhs': ['random', float('inf'), -float('inf'), float('nan')],
+        'is_module': [False],
+        'skip_backward_test': [True],
+        'skip_double_backward_test': [True],
+    })
+))
+class TestMinimum(BinaryMathTestBase, op_utils.ChainerOpTest):
```
asi1024:
```suggestion
class TestMinimum(BinaryMathTestBase, op_utils.NumpyOpTest):
```

aksub99:
There seems to be a `core dumped` error while running `test_math.py`. Any idea why it may be possible @asi1024 ?

aksub99:
@asi1024 Could you please review the modified tests?

asi1024:
Jenkins, test this please.

asi1024:
The test failures are unrelated to this PR. LGTM! @aksub99 Thank you very much!

PR #6547, opened 2019-03-18
Use large eps in test for Linear function

gwtnb:
When calculating numerical differentiation

```
(f(x + eps) - f(x)) / eps
```

of linear functions, large `eps` stabilizes the calculation. The value of `eps` does not affect the result except for rounding error, and `f(x + eps) - f(x)` with small `eps` tends to have large error.

This stabilization can be used not only for `Linear`, but `Convolution2D`, etc.

### DIFF
--- a/tests/chainer_tests/functions_tests/connection_tests/test_linear.py
+++ b/tests/chainer_tests/functions_tests/connection_tests/test_linear.py
@@ -38,10 +38,15 @@ from chainer.testing import backend
 class TestNonparameterizedLinear(testing.FunctionTestCase):
 
     def setUp(self):
-        self.check_backward_options = {'atol': 1e-2, 'rtol': 1e-2}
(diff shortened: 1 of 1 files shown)


gwtnb:
This PR relates to #6574.

gwtnb:
Jenkins, test this please

beam2d:
Note that `F.linear` is actually NOT linear, but bilinear. Large eps is also valid for quadratic functions, thus the tests pass with this fix.

niboshi:
@gwtnb Could you clarify the intention of this PR?
Currently `F.linear` doesn't have any numerical stability issue.

niboshi:
ping

gwtnb:
I intended that this PR is for reference implementation. If unnecessary, it is ok to close.

niboshi:
Thanks for the clarification. Closing.

PR #6552, opened 2019-03-18
Add FP16 support to send/recv

belldandyxtq:
This PR add FP16 to send/recv function pair in ChainerMN for mpi.

### DIFF
--- a/tests/chainermn_tests/functions_tests/test_point_to_point_communication.py
+++ b/tests/chainermn_tests/functions_tests/test_point_to_point_communication.py
@@ -8,70 +8,14 @@ import numpy
 import pytest
 
 import chainermn
-import chainermn.functions
+from chainer.functions import sigmoid
+from chainer.functions import mean_squared_error as mse
 
 
 class Param(object):
     def __init__(self, param):
         self.dtype = None
         self.__dict__.update(param)
-
-
-class Variables(object):
-    def __init__(self, gpu, param):
-        self.gpu = gpu
-        self.communicator = None
-        self.rank_send = 0
-        self.rank_recv = 0
-        self.f = None
-        self.evaluation = None
-        self.x = None
-        self.model = None
-        self.entire_model = None
-        self.device = None
-
-        self.setup(param)
-
-    def setup(self, param):
-        if self.gpu:
-            self.communicator = chainermn.create_communicator('flat')
-            device = self.communicator.intra_rank
-            chainer.cuda.get_device_from_id(device).use()
-        else:
-            self.communicator = chainermn.create_communicator('naive')
-            device = -1
-
-        if self.communicator.size < 2:
-            pytest.skip('This test is for multinode')
-
-        self.rank_send = (self.communicator.rank + 1) % self.communicator.size
-        self.rank_recv = (self.communicator.rank - 1) % self.communicator.size
-
-        # Activation function.
-        self.f = chainer.functions.sigmoid
-
-        # Evaluation function.
-        self.evaluation = chainer.functions.mean_squared_error
-
-        with chainer.using_config('dtype', param.dtype):
-            self.x = chainer.Variable(
-                numpy.arange(10).reshape(1, 10).astype(param.dtype) / 10)
-            self.model = chainer.links.Linear(
-                10, 10, initialW=self._init_w(self.communicator.rank))
-            self.entire_model = [chainer.links.Linear(
-                10, 10, initialW=self._init_w(l))
-                for l in range(self.communicator.size)]
-            self.device = device
-
-            if device >= 0:
-                self.x.to_gpu()
-                self.model.to_gpu()
-                for model in self.entire_model:
-                    model.to_gpu()
-
-    def _init_w(self, l):
-        return 1.0 * numpy.arange(100).reshape(10, 10) \
-            / ((l + 1) * 100)
 
 
 params = [Param(p) for p in [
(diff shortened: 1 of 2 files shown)


tests/chainermn_tests/functions_tests/test_point_to_point_communication.py
```diff
@@ -12,6 +12,13 @@
...
+class Param(object):
+    def __init__(self):
+        self.x = None
+        self.model = None
+        self.entire_model = None
+
+
 class TestPointToPointCommunication(unittest.TestCase):
 
     def setup(self, gpu):
```
shu65: This setup is too large. Could you split using `pytest.mark.parametrize` ?
Please refer to the following link.
https://github.com/chainer/chainer/blob/master/tests/chainermn_tests/communicator_tests/test_communicator.py#L305

belldandyxtq: It seems `parametrize` does not work with `unittest`[link](https://github.com/pytest-dev/pytest/issues/3484)
If we want to do that, we need to remove the `unittest`

tests/chainermn_tests/functions_tests/test_point_to_point_communication.py
```diff
@@ -12,221 +11,251 @@
...
+    {
+        'dtype': numpy.float16,
+    }, {
+        'dtype': numpy.float32,
+    }]]
+
+
+def create_communicator(gpu, param):
+    if gpu:
+        communicator = chainermn.create_communicator('hierarchical')
```
kuenishi: 'hierarchical' is to be deprecated in next release. Use 'flat' instead for default communicator.

tests/chainermn_tests/functions_tests/test_point_to_point_communication.py
```diff
@@ -12,221 +11,251 @@
...
+        pytest.skip('This test is for multinode')
+
+    return communicator
+
+
+def check_communication(gpu, param):
+    communicator = create_communicator(gpu, param)
+    rank_send = (communicator.rank + 1) % communicator.size
+    rank_recv = (communicator.rank - 1) % communicator.size
+    train = TrainEnv(gpu, param, communicator)
```
kuenishi: I don't think `TrainEnv` is intuitive. I think we concluded `create_model`-ish name was more intuitive.
`f` and `evaluation` are not a variable so they don't necessarily have to be held in the object.

tests/chainermn_tests/functions_tests/test_point_to_point_communication.py
```diff
@@ -12,221 +11,251 @@
...
+
+    else:
+        y = chainermn.functions.recv(communicator, rank_recv)
+        err = chainermn.functions.send(y, communicator, rank_send)
+        err.backward()
+
+
+@pytest.mark.parametrize('param', params)
+def test_tuple_communication1_cpu(param):
+    check_tuple_communication(1, False, param)
```
kuenishi: Why don't you just parameterize the length 1 and 2 here?

tests/chainermn_tests/functions_tests/test_point_to_point_communication.py
kuenishi: Say, like 
```python
from chainer.functions import sigmoid
import chainer.functions.mean_squared_error as mse
```

tests/chainermn_tests/functions_tests/test_point_to_point_communication.py
```diff
@@ -9,224 +8,262 @@
...
+        communicator = create_communicator(gpu, param)
+        rank_send = (communicator.rank + 1) % communicator.size
+        rank_recv = (communicator.rank - 1) % communicator.size
+        x = create_x(param, communicator)
+        (model, entire_model) = create_models(param, communicator)
+
+        if communicator.rank == 0:
             # Input process.
-            y = self.f(self.model(self.x))
+            y = function(model(x))
```
kuenishi: Isn't `sigmoid(model(x))` just enough?

tests/chainermn_tests/functions_tests/test_point_to_point_communication.py
```diff
@@ -9,224 +8,262 @@
...
+            grad = model.W.grad
 
             # Compute the expected gradient.
-            x_ = self.x
-            for l in range(self.communicator.size):
-                x_ = self.f(self.entire_model[l](x_))
-            err_ = self.evaluation(x_, self.x)
+            x_ = x
+            for l in range(communicator.size):
+                x_ = function(entire_model[l](x_))
```
kuenishi: ditto

tests/chainermn_tests/functions_tests/test_point_to_point_communication.py
```diff
@@ -9,224 +8,262 @@
...
+                y, communicator, rank_send)
 
             # Unless delegate_variable is used, backprop would stop here.
-            x = chainermn.functions.recv(
-                self.communicator, self.rank_recv,
-                delegate_variable=dlg)
-            err = self.evaluation(x, t)
+            x_ = chainermn.functions.recv(communicator, rank_recv,
+                                          delegate_variable=dlg)
+            err = evaluation(x_, t)
```
kuenishi: ditto; `mse(x_, t)`

tests/chainermn_tests/functions_tests/test_point_to_point_communication.py
belldandyxtq: I want to avoid hard coding. Using 'function' can be more convenient when we want to change to a different one, say ReLU. The same goes to 'evaluation'.

belldandyxtq:
Jenkins, test this please

PR #6554, opened 2019-03-18
Fix `parameterize_pytest` to allow parameterizing with tuples

asi1024:
This PR makes the following code work.
```py
shapes = [
    (1, 2),
    (3, 4, 5)
]

@chainer.testing.parameterize_pytest('shape', shapes)
class TestFoo...
```

PR #6535, opened 2019-03-15
Avoid exposing `chainerx.Scalar` to Python

asi1024:
~Depends on #6481.~

PR #6555, opened 2019-03-18
Rewrite ChainerX arithmetics tests

asi1024:
Depends on ~#6535~ and #6554.


asi1024:
Duplicate with #6566.

PR #6557, opened 2019-03-18
Fix more corner cases in chainerx::Slice

shinh:
I think there were three issues. The first issue was introduced by myself (sorry) when I fixed another issue. See the first three commits for details of these bugs. I think the semantics of GetStart/GetStop were unclear so it was somewhat difficult for me to identify these issues, so I've added comments on these functions in the fourth commit.

PTAL


hvy:
Tried locally generating permutations of slices and they passed (while failing on master obviously).

hvy:
jenkins, test this please.

hvy:
Jenkins failure seems unrelated.

niboshi:
For record (and reference), this part roughly corresponds to the following code in CuPy:
https://github.com/cupy/cupy/blob/v6.0.0b3/cupy/core/internal.pyx#L203-L208

hvy:
LGTM!

PR #6562, opened 2019-03-18
Use float division instead of `__future__.division` for Python2

takagi:
Follows #6462 .

This PR fixes the imagenet example to use float division instead of `__future__.division` for Python2 to compute cropping size.


takagi:
Jenkins, test this please.

niboshi:
LGTM

PR #6581, opened 2019-03-20
Fix rrelu document

### LINKED ISSUE #6503: [document] The input argument name in `F.rrelu` is inconsistent
https://docs.chainer.org/en/stable/reference/generated/chainer.functions.rrelu.html#chainer.functions.rrelu

The slope for x < 0 is named `a` in the body text, but `r` is used for the input argument.

> Randomized Leaky Rectified Liner Unit function.
> This function is expressed as
> f(x)=max(x,ax),
> where a is a random number sampled from a uniform distribution U(l,u).
> ...
> ・r (N-dimensional array or None) – The r to be used for rrelu. The shape and dtype must be the same as x[0] and should be on the same device. If r is not specified or set to None, a r will be generated randomly according to the given l and u. If r is specified, l and u will be ignored.

fiarabbit:
This PR resolves #6503

### DIFF
--- a/chainer/functions/activation/rrelu.py
+++ b/chainer/functions/activation/rrelu.py
@@ -99,10 +99,12 @@ def rrelu(x, l=1. / 8, u=1. / 3, **kwargs):
     .. math:: f(x)=\\max(x, rx),
 
     where :math:`r` is a random number sampled from a uniform distribution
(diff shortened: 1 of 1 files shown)


toslunar:
Jenkins, test this please.

PR #6582, opened 2019-03-20
Remove commented out code from `_BNMode`

hvy:
Removed commented out code from `_BNMode` used in batch normalization.


toslunar:
Note: the next line is the current code for `self.cudnn_dtype_ok`.

PR #6593, opened 2019-03-20
Fix dimension check of `chainerx::Linear`

crcrpar:
This PR fixes a weird dimension check in chainerx/routines/connection where chainerx::Linear evaluates `w.ndim()` twice and the first evaluation does not make sense.


niboshi:
Thanks!
Jenkins, test this please

niboshi:
LGTM

PR #6597, opened 2019-03-21
[WIP] Add `Pad` to chainerx

vermashresth:
Issue #6423 
Added the Pad routine for chainerx. Currently implemented only `constant` mode. @hvy I had several doubts regarding the design.

Since the `pad_width` and `constant_value` variables can be both a single value or an array, there would be lot of possible function definitions (currently I have implemented two). Is it fine to just define them all as I have done or should it be done in some other way?  

Should padding be in array manipulation or creation module or be a different module altogether because it will contain several helper functions for different modes?

I think this PR needs a lot more work so I prepared a list. Please let me know your thoughts on it.
- [x] Add Python tests
- [x] Add support for array of tuple `pad_width` and its variations. This might need creation routines like `tile`.
- [ ] Improve error handling in Pad
- [ ] Enable `reflection` and `edge` modes
- [ ] Move Pad to separate module


chainerx_cc/chainerx/python/routines.cc
```diff
@@ -350,6 +350,23 @@ void InitChainerxManipulation(pybind11::module& m) {
           [](const ArrayBodyPtr& a, int8_t axes) { return MoveArrayBody(Transpose(Array{a}, {axes})); },
           py::arg("a"),
           py::arg("axes") = nullptr);
+    m.def("pad",
+          [](const ArrayBodyPtr& a, int8_t pad_width, const std::string& mode, int64_t constant_values) {
+              return MoveArrayBody(Pad(Array{a}, pad_width, mode, constant_values));
+          },
+          py::arg("a"),
```
niboshi: Let's use argument name "array" to align with `numpy.pad`.

chainerx_cc/chainerx/routines/manipulation.h
```diff
@@ -22,6 +23,8 @@ Array RollAxis(const Array& a, int8_t axis, int8_t start = 0);
 // Returns a transposed view of the array.
 Array Transpose(const Array& a, const OptionalAxes& axes = nonstd::nullopt);
 
+Array Pad(const Array& a, int8_t pad_width, const std::string& mode, int64_t constant_values);
```
niboshi: Let's use `enum` instead of `std::string`. (just like `chainerx::AveragePoolPadMode` defined in `chainerx/routines/pooling.h`)

(For Python binding, it's ok to use `str`.)

niboshi: Let's use "array" as argument name here, too.

chainerx_cc/chainerx/python/routines.cc
vermashresth: Done!

chainerx_cc/chainerx/routines/manipulation.h
vermashresth: I have changed padding mode to use `enum`.

vermashresth:
Hi! @niboshi 
I have added python tests for the Pad routine and have made the function definition resemble `numpy.pad`. Could you please take a look and see if they are fine?

chainerx_cc/chainerx/routines/manipulation.h
```diff
@@ -22,6 +28,11 @@ Array RollAxis(const Array& a, int8_t axis, int8_t start = 0);
 // Returns a transposed view of the array.
 Array Transpose(const Array& a, const OptionalAxes& axes = nonstd::nullopt);
 
+// Returns an array padded by pad_width amount
+Array Pad(const Array& a, int8_t pad_width, PadMode mode, int64_t constant_values);
```
niboshi: The types of `pad_width` and `constant_values` are independent.
For example, one should be able to pass `int8_t` and `std::vector<std::vector<int64_t>>`.
One way would be to introduce specialized classes `class PadWidths` and `class ConstantValues` which are implicitly convertible from `int8_t`/`std::vector<int8_t>`/`std::vector<std::vector<int8_t>>` (replace `int8_t` with `int64_t` for `ConstantValues`).

chainerx_cc/chainerx/routines/manipulation_test.cc
```diff
@@ -0,0 +1,1081 @@
+#include "chainerx/routines/manipulation.h"
```
niboshi: We no longer test routines in C++ tests.
Could you remove this file?

chainerx_cc/chainerx/routines/manipulation.h
```diff
@@ -11,6 +12,11 @@
 
 namespace chainerx {
 
+enum class PadMode {
+    constant = 1,
```
niboshi: As a convention, please name the members like `kConstant`.

chainerx_cc/chainerx/routines/manipulation.cc
```diff
@@ -115,6 +115,99 @@ Array Transpose(const Array& a, const OptionalAxes& axes) {
...
+Array Pad(const Array& array, int8_t pad_width, PadMode mode, int64_t constant_values) {
+    std::vector<std::vector<int8_t>> new_pad_width = NormalizeShape(array, pad_width);
+    std::vector<std::vector<int64_t>> new_constant_values = NormalizeShape(array, constant_values);
+    return Pad(array, new_pad_width, mode, new_constant_values);
+}
+
+Array Pad(const Array& array, std::vector<int8_t> pad_width, PadMode mode, std::vector<int64_t> constant_values) {
+    Array newarray;
+    newarray = array.Copy();
+    if (mode == PadMode::constant) {
```
niboshi: Please use `switch` and write `CHAINERX_NEVER_REACH();` in the `default:` clause.

chainerx_cc/chainerx/routines/manipulation.cc
```diff
@@ -115,6 +115,99 @@ Array Transpose(const Array& a, const OptionalAxes& axes) {
...
+}
+}  // namespace
+
+Array Pad(const Array& array, int8_t pad_width, PadMode mode, int64_t constant_values) {
+    std::vector<std::vector<int8_t>> new_pad_width = NormalizeShape(array, pad_width);
+    std::vector<std::vector<int64_t>> new_constant_values = NormalizeShape(array, constant_values);
+    return Pad(array, new_pad_width, mode, new_constant_values);
+}
+
+Array Pad(const Array& array, std::vector<int8_t> pad_width, PadMode mode, std::vector<int64_t> constant_values) {
```
niboshi: Pass `std::vector` by const reference.

chainerx_cc/chainerx/routines/manipulation.cc
```diff
@@ -115,6 +115,99 @@ Array Transpose(const Array& a, const OptionalAxes& axes) {
...
+    Array newarray;
+    newarray = array.Copy();
+    if (mode == PadMode::constant) {
+        for (int8_t i = 0; i < array.shape().ndim(); ++i) {
```
niboshi: You can simply write `array.ndim()`.

chainerx_cc/chainerx/routines/manipulation.h
```diff
@@ -11,6 +12,11 @@
...
+    edge,
```
niboshi: `edge` and `reflection` are not implemented.
How about removing them so that no one would accidently use them?
(Perhaps we can write TODO here to implement them)

tests/chainerx_tests/unit_tests/routines_tests/test_manipulation.py
```diff
@@ -181,6 +181,40 @@ def test_rollaxis_invalid(in_shape, axis, start):
...
+    def generate_inputs(self):
+        a = array_utils.shaped_arange(self.shape, 'float32')
+        return a,
+
+    def forward_xp(self, inputs, xp):
+        a, = inputs
+        pad_width = self.pad_width
+        mode = self.mode
+        constant_values = self.constant_values
+        if xp is numpy:
```
niboshi: Do you need this condition?

tests/chainerx_tests/unit_tests/routines_tests/test_manipulation.py
```diff
@@ -181,6 +181,40 @@ def test_rollaxis_invalid(in_shape, axis, start):
         else:
             chainerx.rollaxis(a, axis, start)
 
+@op_utils.op_test(['native:0', 'cuda:0'])
+@chainer.testing.parameterize_pytest('shape,pad_width,mode,constant_values', [
+    ((1,), 1, "constant", 0),
```
niboshi: Use single quotes for string literals for consistency.

chainerx_cc/chainerx/routines/manipulation.h
vermashresth: Could you please elaborate a bit how to implement this? I have been stuck with this since past few days.

tests/chainerx_tests/unit_tests/routines_tests/test_manipulation.py
vermashresth: Done.

tests/chainerx_tests/unit_tests/routines_tests/test_manipulation.py
vermashresth: Right, I have removed it now!

chainerx_cc/chainerx/routines/manipulation.h
vermashresth: Sure, I have done that now.

chainerx_cc/chainerx/routines/manipulation.cc
vermashresth: Done.

chainerx_cc/chainerx/routines/manipulation.cc
vermashresth: Done.

chainerx_cc/chainerx/routines/manipulation.h
vermashresth: Done.

chainerx_cc/chainerx/routines/manipulation_test.cc
vermashresth: Oh okay. Didn't notice that.

chainerx_cc/chainerx/routines/manipulation.cc
vermashresth: Sure, done that now.

chainerx_cc/chainerx/python/routines.cc
```diff
@@ -355,6 +355,54 @@ void InitChainerxManipulation(pybind11::module& m) {
...
+                  pad_mode = PadMode::constant;
+              } else {
+                  throw py::value_error{"Currently only `constant` padding mode is supported"};
+              }
+              return MoveArrayBody(Pad(Array{array}, pad_width, pad_mode, constant_values));
+          },
+          py::arg("array"),
+          py::arg("pad_width"),
+          py::arg("mode"),
+          py::arg("constant_values"));
```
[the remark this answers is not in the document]
vermashresth: Done.

chainerx_cc/chainerx/python/routines.cc
```diff
@@ -355,6 +355,54 @@ void InitChainerxManipulation(pybind11::module& m) {
           py::arg("a"),
           py::arg("axis"),
           py::arg("start") = 0);
+    m.def("pad",
+          [](const ArrayBodyPtr& array, int8_t pad_width, const std::string& mode, int64_t constant_values) {
```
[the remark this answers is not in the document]
vermashresth: Done.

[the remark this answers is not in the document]
vermashresth: Done.

vermashresth:
Hi @niboshi 

Sorry for such a long delay. I had exams going on.
I have implemented the custom class parameters for the pad function. I think there is some code duplication in the definition of two classes, so would it be a good idea to use template class? Please let me know your thoughts.

As for the python binding function, I tried a lot but couldn't come up with a clean implementation and had to write return statement for all possible combinations. I referred to the [binding function for `Split`](https://github.com/chainer/chainer/blob/master/chainerx_cc/chainerx/python/routines.cc#L450:L538) to see how dynamic type checking and casting could be done but there there was only one parameter, so things were fine. I tried something like independently checking for both variables but couldn't use `return` statement from `if block` of one variable since the function call needs both of them together.

It would be great if you could suggest something for this.

niboshi:
Thank you for fixes! We were in a long vacation. I'm sorry for delayed reply.

>would it be a good idea to use template class?

That sounds good to me.

>I tried something like independently checking for both variables but couldn't use `return` statement from `if block` of one variable since the function call needs both of them together.

Seems like you can separate conversion logics as respective functions. How about that?

Also, flake8 and clang-format are complaining in the CI. Please fix code formatting as they suggest.

PR #6611, opened 2019-03-24
Use `F.matmul` instead of `F.batch_matmul` in memnn example

takagi:
This PR fixes the memnn example to use `F.matmul` rather than `F.batch_matmul`, which is deprecated now. With `F.matmul`, the memnn example can work in FP16 mode.

### DIFF
--- a/examples/memnn/memnn.py
+++ b/examples/memnn/memnn.py
@@ -116,7 +116,7 @@ class Memory(object):
         tc = self.TC(inds)
         tm = F.broadcast_to(tm, self.m.shape)
         tc = F.broadcast_to(tc, self.c.shape)
-        p = F.softmax(F.matmul(self.m + tm, u[..., None]))
+        p = F.softmax(F.matmul(self.m + tm, F.expand_dims(u, u.ndim)))
         o = F.matmul(F.swapaxes(self.c + tc, 2, 1), p)
         o = F.squeeze(o, -1)
         u = o + u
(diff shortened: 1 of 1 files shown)


examples/memnn/memnn.py
```diff
@@ -116,8 +116,8 @@ def query(self, u):
         tc = self.TC(inds)
         tm = F.broadcast_to(tm, self.m.shape)
         tc = F.broadcast_to(tc, self.c.shape)
-        p = F.softmax(F.batch_matmul(self.m + tm, u))
-        o = F.batch_matmul(F.swapaxes(self.c + tc, 2, 1), p)
+        p = F.softmax(F.matmul(self.m + tm, u[..., None]))
```
toslunar: `F.expand_dims` performs better on backprop.

takagi: Fixed!

examples/memnn/memnn.py
```diff
@@ -116,8 +116,8 @@ def query(self, u):
         tc = self.TC(inds)
         tm = F.broadcast_to(tm, self.m.shape)
         tc = F.broadcast_to(tc, self.c.shape)
-        p = F.softmax(F.batch_matmul(self.m + tm, u))
-        o = F.batch_matmul(F.swapaxes(self.c + tc, 2, 1), p)
+        p = F.softmax(F.matmul(self.m + tm, F.expand_dims(u, u.ndim)))
```
toslunar: How about this?
```suggestion
        p = F.softmax(F.matmul(self.m + tm, F.expand_dims(u, -1)))
```

takagi: Sure!

toslunar:
Jenkins, test this please.

PR #6615, opened 2019-03-25
Fix pix2pix example to work with default dtype

takagi:
Rel #6168 .

This PR fixes the pix2pix example to work with the default dtype feature.


examples/pix2pix/updater.py
```diff
@@ -54,8 +59,9 @@ def update_core(self):
         w_in = 256
         w_out = 256
 
-        x_in = xp.zeros((batchsize, in_ch, w_in, w_in)).astype('f')
-        t_out = xp.zeros((batchsize, out_ch, w_out, w_out)).astype('f')
+        dtype = self.dtype
+        x_in = xp.zeros((batchsize, in_ch, w_in, w_in)).astype(dtype)
+        t_out = xp.zeros((batchsize, out_ch, w_out, w_out)).astype(dtype)
 
         for i in range(batchsize):
```
crcrpar: I think we can use `chainer.dataset.concat_examples` to create minibatches like this ↓ instead of copying one by one after preparing 2 zeroed arrays.
```
x_in, t_out = self.converter(self.get_iterator('main').next(), device=self.device)
```

What do you think?

takagi: Definitely. I'll fix the example as this!

takagi: Fixed!

crcrpar: Kudos.

PR #6618, opened 2019-03-25
Improve docs and exception message in `F.erfcx`, `F.erfcinv` and `F.erfinv`

kmaehashi:


chainer/functions/math/erfcx.py
```diff
@@ -27,7 +27,7 @@ def forward_cpu(self, x):
                 _erfcx_cpu = special.erfcx
             except ImportError:
                 raise ImportError('SciPy is not available. Forward computation'
-                                  ' of erfcx can not be done.')
+                                  ' of erfcx cannot be done.')
```
toslunar: Could you add 'in CPU' to make it consistent with `erfinv` and `erfcinv`?

kmaehashi: Changed to be consistent with `erfinv` and `erfcinv`.

toslunar:
Jenkins, test this please.

PR #6620, opened 2019-03-25
remove redefinition of mean

kshitij12345:
Remove a redefinition of mean in pool.cc

### DIFF
--- a/chainerx_cc/chainerx/native/native_device/pool.cc
+++ b/chainerx_cc/chainerx/native/native_device/pool.cc
@@ -20,6 +20,7 @@
 #include "chainerx/routines/indexing.h"
 #include "chainerx/routines/math.h"
 #include "chainerx/routines/pooling.h"
+#include "chainerx/routines/statistics.h"
 #include "chainerx/scalar.h"
 #include "chainerx/shape.h"
 #include "chainerx/stack_vector.h"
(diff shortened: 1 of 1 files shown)


kshitij12345:
I was also wondering since mean is very much implemented, should we expose it to Python API and write some test cases for it against `numpy`.

chainerx_cc/chainerx/native/native_device/pool.cc
```diff
@@ -220,7 +214,7 @@ class NativeAveragePoolForwardBackward : public chainerx::AveragePoolForwardBack
 
         switch (pad_mode_) {
             case AveragePoolPadMode::kZero:
-                Mean(col, kernel_axes, out);
+                out = Mean(col, kernel_axes);
```
hvy: This `Mean` is a routine meaning that it has its own logic for deciding the output dtype (and connect the computational graph) which is not really what we want. Here we just want to compute the result.

hvy:
Thanks the this PR! I wrote it in the comment but the `Mean` here is not a mere redefinition but introduced for a different purpose. As stated in the TODO, I proper solution would be to define `Device::Mean`.

hvy:
On a different topic, exposing `Mean` and also `Var` with Python bindings would be great and indeed welcomed as PRs. :)

chainerx_cc/chainerx/native/native_device/pool.cc
kshitij12345: Oh, I see. I totally missed it. Thank you

kshitij12345:
Oh see. Failed to notice that. Closing the PR for now. Will work on exposing to Python Api. And after that maybe try the device implementation for mean.

PR #6636, opened 2019-03-26
Destroy NCCL communicator after every use 

keisukefukuda:
This PR fixes NCCL_UNHANDLED_CUDA_ERROR during CI. 

The recently-added mixed16 tests creates PureNcclCommunicator repeatedly in very short intervals. This may cause unhandled cuda error from NCCL library.

### DIFF
--- a/tests/chainermn_tests/communicator_tests/test_communicator.py
+++ b/tests/chainermn_tests/communicator_tests/test_communicator.py
@@ -247,7 +247,7 @@ def destroy_communicator(comm):
     destory NCCL communicator after every use."""
     if hasattr(comm, 'nccl_comm') and comm.nccl_comm is not None:
         comm.nccl_comm.destroy()
-    comm.nccl_comm = None
+        comm.nccl_comm = None
     if hasattr(comm, 'intra_nccl_cojmm') and comm.intra_nccl_comm is not None:
         comm.intra_nccl_comm.destroy()
         comm.intra_nccl_comm = None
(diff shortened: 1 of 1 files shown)


kuenishi:
I would rather suggest

```python
if hasattr(communicator, 'nccl_comm'):
    communicator.nccl_comm.destroy()
```
than casually using `del` and `gc.collect()` in Python.

shu65:
It is a good idea!
I confirmed that the unittest for chainermn were passed when using the following.
```
    if hasattr(communicator, 'nccl_comm'):
        communicator.nccl_comm.destroy()
    if hasattr(communicator, 'intra_nccl_comm'): # for hierarchical, single, and two_dimensional
        communicator.intra_nccl_comm.destroy()
```

tests/chainermn_tests/communicator_tests/test_communicator.py
```diff
@@ -239,6 +239,20 @@ def create_communicator(param, use_gpu):
...
 
+def destroy_communicator(comm):
+    """Destroy internal NCCL communicator.
+
+    When too many NCCL communicator are alive, NCCL produces
+    unhandled CUDA error. To avoid this, we need to make sure to
+    destory NCCL communicator after every use."""
+    if hasattr(comm, 'nccl_comm') and comm.nccl_comm is not None:
+        comm.nccl_comm.destroy()
+    comm.nccl_comm = None
```
shu65: Is this line indent wrong?

keisukefukuda: Fixed! thanks!

shu65:
Jenkins, test this please.

PR #6639, opened 2019-03-26
Fix ChainerX CBLAS link issue

### LINKED ISSUE #6607: Undefined reference when building chainerx with BLAS support
I cannot compile the unit tests of chainerx when I enable BLAS support of the native backend.

## Conditions

* blas: OpenBLAS
* chainer: the current master (the commit id is `c8c9ee0d5cdad4726f01e6a354f775c555f74355`).
* cmake version: 3.14.0
* g++ version: 8.2.1
* Platform: Linux (`Linux 5.0.2-arch1-1-ARCH #1 SMP PREEMPT Thu Mar 14 18:47:49 UTC 2019 x86_64 GNU/Linux`)

## Shell Commands to reproduce

Following commands (almost the same as [the documentation of chainerx](https://github.com/chainer/chainer/blob/master/docs/source/chainerx/contribution.rst)) reproduce the error.

```bash
# I assume that OpenBLAS is already installed on this computer.
$ cd chainer
$ cd chainerx_cc
$ mkdir -p build
$ cd build
$ cmake -DCHAINERX_BUILD_TEST=ON -DCHAINERX_ENABLE_BLAS=ON ..
$ make -j $(nproc)
...
/usr/bin/ld: libchainerx.so: undefined reference to `cblas_sgemm'
/usr/bin/ld: libchainerx.so: undefined reference to `cblas_dgemm'
```

## Cause

I think the reason for this bug is that `libchainerx` does not have the proper dependency on shared libraries.
Although `cblas_sgemm` and `clbas_dgemm` are functions of `libcblas`, `libchainerx` does not have dependency on `libcblas`.

```bash
$ ldd build/chainerx/libchainerx.so
        linux-vdso.so.1 (0x00007ffc61b8c000)
        libdl.so.2 => /usr/lib/libdl.so.2 (0x00007f002cdf6000)
        libopenblas.so.3 => /usr/lib/libopenblas.so.3 (0x00007f002bbc7000)
        libpthread.so.0 => /usr/lib/libpthread.so.0 (0x00007f002bba6000)
        libstdc++.so.6 => /usr/lib/libstdc++.so.6 (0x00007f002ba17000)
        libm.so.6 => /usr/lib/libm.so.6 (0x00007f002b892000)
        libgcc_s.so.1 => /usr/lib/libgcc_s.so.1 (0x00007f002b878000)
        libc.so.6 => /usr/lib/libc.so.6 (0x00007f002b6b2000)
        /usr/lib64/ld-linux-x86-64.so.2 (0x00007f002d123000)
        libgomp.so.1 => /usr/lib/libgomp.so.1 (0x00007f002b682000)
```

Adding dependency on `libcblas` solves this error. The following is a workaround patch I tested in my environment.

```patch
--- a/chainerx_cc/chainerx/native/CMakeLists.txt
+++ b/chainerx_cc/chainerx/native/CMakeLists.txt
@@ -47,7 +47,7 @@ if(${BLAS_FOUND})
         message(STATUS "Found BLAS (include directory: ${BLAS_INCLUDE_DIRS})")
         add_definitions(-DCHAINERX_ENABLE_BLAS=1)
         include_directories(${BLAS_INCLUDE_DIRS})
-        target_link_libraries(chainerx_native ${BLAS_LIBRARIES})
+        target_link_libraries(chainerx_native ${BLAS_LIBRARIES} /usr/lib/libcblas.so)
     else()
         message(WARNING "Found the BLAS library but not the include directory. Skipping BLAS.")
     endif()
```

grafi-tt:
This should fix #6607.

@HiroakiMikami Could you try ChainerX build with this branch? I don't have Arch Linux environment.

Probably the cause of the error is Arch Linux's OpenBLAS is built [without CBLAS API](https://git.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/openblas#n22). CBLAS functions are separately provided in libcblas package.

This PR

* guards `<cblas.h>` by `extern "C"` and
* searches `libcblas.so` in `FindCBLAS.cmake` if required.

Those changes also fixes the issue with ATLAS on Ubuntu 18.04. ATLAS ships `cblas.h` without `__cplusplus` checking, so guards should be added at include place.

Note that the PR don't address header file detection. Notably MKL wouldn't work since it ships `mkl_cblas.h`. Though it is a different issue, I think it is better to bundle our own `cblas.h` in ChainerX as [numpy does](https://github.com/numpy/numpy/blob/master/numpy/core/src/common/npy_cblas.h).

### DIFF
--- a/chainerx_cc/cmake/FindBLAS.cmake
+++ b/chainerx_cc/cmake/FindBLAS.cmake
@@ -409,12 +409,28 @@ endif ()
 if (BLA_VENDOR STREQUAL "OpenBLAS" OR BLA_VENDOR STREQUAL "All")
   if(NOT BLAS_LIBRARIES)
     # OpenBLAS (http://www.openblas.net)
+    # ChainerX modification: check `cblas_sgemm` since ChianerX uses CBLAS API.
     check_fortran_libraries(
       BLAS_LIBRARIES
       BLAS
-      sgemm
+      cblas_sgemm
       ""
       "openblas"
+      ""
+      )
+  endif()
+endif ()
+
+if (BLA_VENDOR STREQUAL "OpenBLAS" OR BLA_VENDOR STREQUAL "All")
+  if(NOT BLAS_LIBRARIES)
+    # ChainerX modification: search libcblas if `cblas_sgemm` not found.
+    # OpenBLAS might be built without cblas functions (e.g. on Arch Linux.)
+    check_fortran_libraries(
+      BLAS_LIBRARIES
+      BLAS
+      cblas_sgemm
+      ""
+      "openblas;cblas"
       ""
       )
   endif()
--- a/chainerx_cc/cmake/FindBLAS.cmake
+++ b/chainerx_cc/cmake/FindBLAS.cmake
@@ -450,6 +450,22 @@ if (BLA_VENDOR STREQUAL "FLAME" OR BLA_VENDOR STREQUAL "All")
   endif()
 endif ()
 
+if (BLA_VENDOR STREQUAL "ATLAS" OR BLA_VENDOR STREQUAL "All")
+  if(NOT BLAS_LIBRARIES)
+    # BLAS in ATLAS library? (http://math-atlas.sourceforge.net/)
+    # ChainerX modification: search libcblas as well and check the function
+    # `cblas_sgemm`, since ATLAS provides libcblas as a separate library.
+    check_fortran_libraries(
+      BLAS_LIBRARIES
+      BLAS
+      cblas_sgemm
+      ""
+      "f77blas;atlas;cblas"
+      ""
+      )
+  endif()
+endif ()
+
 # BLAS in PhiPACK libraries? (requires generic BLAS lib, too)
 if (BLA_VENDOR STREQUAL "PhiPACK" OR BLA_VENDOR STREQUAL "All")
   if(NOT BLAS_LIBRARIES)
@@ -729,28 +745,6 @@ if (BLA_VENDOR STREQUAL "Generic" OR BLA_VENDOR STREQUAL "All")
   endif()
 endif ()
 
-# NOTE: ChainerX does not support ATLAS because we encountered
-# `undefined reference 'cblas_sgemm'` errors, at least on Ubuntu16.04.
-# Also, it is well known that ATLAS is significantly slower than OpenBLAS.
-if (BLA_VENDOR STREQUAL "ATLAS" OR BLA_VENDOR STREQUAL "All")
-  if(NOT BLAS_LIBRARIES)
-    # BLAS in ATLAS library? (http://math-atlas.sourceforge.net/)
-    check_fortran_libraries(
-      BLAS_LIBRARIES
-      BLAS
-      dgemm
-      ""
-      "f77blas;atlas"
-      ""
-      )
-    if(BLAS_LIBRARIES)
-      message(WARNING "ATLAS is found (library: ${BLAS_LIBRARIES}), but ChainerX does not support ATLAS.")
-      set(BLAS_LIBRARIES FALSE)
-    endif()
-  endif()
-endif ()
-
-# Set BLAS_FOUND
 if(BLA_F95)
   find_package_handle_standard_args(BLAS REQUIRED_VARS BLAS95_LIBRARIES)
   set(BLAS95_FOUND ${BLAS_FOUND})


chainerx_cc/cmake/FindBLAS.cmake
```diff
@@ -171,6 +171,21 @@ macro(Check_Fortran_Libraries LIBRARIES _prefix _name _flags _list _thread)
...
   endif()
+  if(_libraries_work)
+    # NOTE: ChainerX uses CBLAS API. Search `libcblas.so` if CBLAS functions are
+    # not provided in the blas library found.
+    set(CMAKE_REQUIRED_LIBRARIES ${_flags} ${${LIBRARIES}} ${_thread})
+    check_function_exists("cblas_sgemm" ${_prefix}${_combined_name}_HAS_CBLAS)
+    if(NOT ${_prefix}${_combined_name}_HAS_CBLAS)
+        find_library(${_prefix}${_combined_name}_CBLAS_LIBRARY cblas)
+        if(${_prefix}${_combined_name}_CBLAS_LIBRARY)
+          set(${LIBRARIES} ${${LIBRARIES}} ${_prefix}${_combined_name}_CBLAS_LIBRARY)
```
HiroakiMikami: This line appends the variable name (e.g., BLAS_openblas_CBLAS_LIBRARY) to the `LIBRARIES` variable.
`set(${LIBRARIES} ${${LIBRARIES}} ${${_prefix}${_combined_name}_CBLAS_LIBRARY})` seems correct.

HiroakiMikami:
This PR almost works for me, but there is one trivial mistake I commented.

grafi-tt:
@HiroakiMikami Thank you for pointing out my silly mistake. I fixed the mistake and confirmed that `FindBLAS` works well.

@hvy Could you review the patch?

grafi-tt:
ping

hvy:
Sorry for the delayed review. I'll try to take a look at it today.

hvy:
Jenkins, test this please.

chainerx_cc/cmake/FindBLAS.cmake
```diff
@@ -171,6 +171,21 @@ macro(Check_Fortran_Libraries LIBRARIES _prefix _name _flags _list _thread)
...
+    # NOTE: ChainerX uses CBLAS API. Search `libcblas.so` if CBLAS functions are
+    # not provided in the blas library found.
+    set(CMAKE_REQUIRED_LIBRARIES ${_flags} ${${LIBRARIES}} ${_thread})
+    check_function_exists("cblas_sgemm" ${_prefix}${_combined_name}_HAS_CBLAS)
+    if(NOT ${_prefix}${_combined_name}_HAS_CBLAS)
+        find_library(${_prefix}${_combined_name}_CBLAS_LIBRARY cblas)
+        if(${_prefix}${_combined_name}_CBLAS_LIBRARY)
+          set(${LIBRARIES} ${${LIBRARIES}} ${${_prefix}${_combined_name}_CBLAS_LIBRARY})
+        else()
+          set(_libraries_work FALSE)
```
hvy: Could you explain why this is needed?

grafi-tt: * If `cblas_sgemm` symbol not found in the blas library (e.g. `libopenblas.so`), then `${_prefix}${_combined_name}_HAS_CBLAS` is set false.
* In this case, check whether `libcblas.so` exists. This file should provides CBLAS API.
* If `libcblas.so` not found, then we couldn't use CBLAS API (only FORTRAN API available.)  So we revert `_li
brafy_work` flag.

This situation may happen when user is running Arch Linux. It strips CBLAS API from OpenBLAS, and provides the API by a separate package [`cblas`](https://www.archlinux.org/packages/extra/x86_64/cblas/) that contains `libcblas.so`.

hvy:
Jenkins, test this please.

chainerx_cc/cmake/FindBLAS.cmake
hvy: Thanks for the explanation. I just want to make sure but the vendor for which this path is passed is OpenBLAS? Is it the case that no other vendors are available (e.g. ATLAS) in the environment for which this issue tries to address? If not, you could use `BLA_VENDOR ` to specify the candidate vendors explicitly. For reference, this seems like a step towards a less ad-hoc approach to me, but this also depends on the answers to those questions. https://cmake-basis.github.io/apidoc/latest/FindBLAS_8cmake_source.html

grafi-tt: ATLAS has the same issue. ATLAS officially provides `libatlas.so` and `libcblas.so` separately.

https://packages.ubuntu.com/ja/bionic/amd64/libatlas3-base/filelist

Debian and Ubuntu also build `libblas.so` to support BLAS library switching. Since we blacklisted `libatlas.so`, the issue hadn't appear on those distributions.

https://salsa.debian.org/science-team/atlas/blob/a53f04c697a75dd39487fe331d49eea13effefa7/debian/rules#L178-182

hvy: I see. I just find the current change a bit ad-hoc since it's non-trivial and not visible outside `Check_Fortran_Libraries`. How about adding another argument to it, for additional libraries to search for (and changing the `_name` passed to the function for OpenBLAS and ATLAS to the "missing" symbol?

hvy:
In case the comment got slipped away, could you address https://github.com/chainer/chainer/pull/6639#discussion_r289241981 @grafi-tt ?

hvy:
@HiroakiMikami could you check if the current changes solves your issue?

hvy:
Jenkins, test this please.

grafi-tt:
Sorry for late response. I rebased the changes onto master, and squashed commits for simpler history.

The test failure is not related to the change. Could you please check again? @hvy @HiroakiMikami

PR #6647, opened 2019-03-27
 Environment variable to set ChainerX Python binding build type 

niboshi:
Required in #6646 

Set `CHAINERX_BUILD_TYPE=Debug` when installing Chainer to build ChainerX in debug mode.

Possible values: `Debug` or `Release` (default)

### DIFF
--- a/chainerx_build_helper.py
+++ b/chainerx_build_helper.py
@@ -37,6 +37,21 @@ class CMakeBuild(build_ext.build_ext):
             self.build_extension(ext)
 
     def build_extension(self, ext):
+        # Decide the build type: release/debug
+        build_type = os.getenv('CHAINERX_BUILD_TYPE', None)
+        if build_type is not None:
+            # Use environment variable
+            pass
+        elif self.debug:
+            # Being built with `python setup.py build --debug`
+            build_type = 'Debug'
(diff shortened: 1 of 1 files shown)


chainerx_build_helper.py
```diff
@@ -37,36 +37,44 @@ def run(self):
...
+        elif self.debug:
+            # Being built with `python setup.py build --debug`
+            build_type = 'Debug'
+        elif os.getenv('READTHEDOCS', None) == 'True':
+            # on ReadTheDocs
+            build_type = 'Debug'
+        else:
+            # default
+            build_type = 'Release'
+
```
asi1024: How about checking `assert build_type in ('Debug', 'Release')`?

niboshi: I don't think it's something to be done here.
`cmake` should handle that.

asi1024: Currently, cmake seems not to check it.

niboshi: Hmm, I think that should be addresed separately.
Those who build chainerx_cc directly should also receive an error.

asi1024:
Jenkins, test this please.

asi1024:
Test failures seem unrelated. LGTM.

PR #6432, opened 2019-03-05
Mixed dtype: arithmetics

asi1024:
~Merge #6481 first.~
Close #6385.

PR #6651, opened 2019-03-27
Suppress numpy warnings in math tests

asi1024:
Depends on #6432.


niboshi:
How about using `numpy.seterr`?
(Also, use single quotes.)

niboshi:
Jenkins, test this please

niboshi:
LGTM

PR #5742, opened 2018-11-29
add spectral normalization linkhook

crcrpar:
This PR aims at accessing #5702.

PR #6655, opened 2019-03-27
Improve `link_hooks.SpectralNormalization`

crcrpar:
This PR enables `chainer.link_hooks.SpectralNormalziation` to support ideep, chainerx, and multiple GPUs.

### DIFF
--- a/chainer/link_hooks/spectral_normalization.py
+++ b/chainer/link_hooks/spectral_normalization.py
@@ -2,7 +2,6 @@ import numpy
 
 import chainer
 from chainer import backend
-from chainer.backends import intel64
 from chainer import configuration
 import chainer.functions as F
 from chainer import link_hook
@@ -275,12 +274,10 @@ class SpectralNormalization(link_hook.LinkHook):
             self.v = v
             with chainer.using_device(link.device):
                 if configuration.config.train:
-                    if (link.xp is chainerx or
-                            isinstance(getattr(link, vector_name),
-                                       intel64.ideep.mdarray)):
+                    if link.xp is chainerx:
                         getattr(link, vector_name)[:] = u
                     else:
-                        link.xp.copyto(getattr(link, vector_name), u)
+                        backend.copyto(getattr(link, vector_name), u)
         return W
 
     def reshape_W(self, W):
--- a/chainer/link_hooks/spectral_normalization.py
+++ b/chainer/link_hooks/spectral_normalization.py
@@ -8,7 +8,7 @@ from chainer import link_hook
 import chainer.links as L
 from chainer import variable
 import chainerx
-from chainerx._fallback_workarounds import _from_chx as from_chx_workaround
+from chainerx import _fallback_workarounds as fallback
 
 
 def l2normalize(xp, v, eps=1e-12):
@@ -239,7 +239,7 @@ class SpectralNormalization(link_hook.LinkHook):
             weight_matrix = self.reshape_W(initialW.array)
             # TODO(crcrpar): Remove this when chainerx supports SVD.
             if link.xp is chainerx:
-                xp, device, array = from_chx_workaround(weight_matrix)
+                xp, device, array = fallback._from_chx(weight_matrix)
                 if xp is numpy:
                     _, s, _ = numpy.linalg.svd(array)
                 else:
--- a/chainer/link_hooks/spectral_normalization.py
+++ b/chainer/link_hooks/spectral_normalization.py
@@ -11,7 +11,7 @@ import chainerx
 from chainerx import _fallback_workarounds as fallback
 
 
-def l2normalize(xp, v, eps=1e-12):
+def l2normalize(xp, v, eps):
     """Normalize a vector by its L2 norm.
 
     Args:
@@ -34,7 +34,7 @@ def l2normalize(xp, v, eps=1e-12):
 
 
 def update_approximate_vectors(
-        weight_matrix, u, n_power_iteration=1, eps=1e-12):
+        weight_matrix, u, n_power_iteration, eps):
     """Update the first left and right singular vectors.
 
     This function updates the first left singular vector `u` and
@@ -108,7 +108,9 @@ class SpectralNormalization(link_hook.LinkHook):
         n_power_iteration (int): Number of power iteration.
             The default value is 1.
         eps (float): Numerical stability in norm calculation.
-            The default value is 1e-12.
+            The default value is 1e-6 for the compatibility with
+            mixed precision training. The value used in the author's
+            implementation is 1e-12.
         use_gamma (bool): If ``True``, weight scaling parameter gamma which is
             initialized by initial weight's max singular value is introduced.
         factor (float, None): Scaling parameter to divide maximum singular
@@ -154,7 +156,7 @@ class SpectralNormalization(link_hook.LinkHook):
 
     name = 'SpectralNormalization'
 
-    def __init__(self, n_power_iteration=1, eps=1e-12, use_gamma=False,
+    def __init__(self, n_power_iteration=1, eps=1e-6, use_gamma=False,
                  factor=None, weight_name='W', name=None):
         assert n_power_iteration > 0
         self.n_power_iteration = n_power_iteration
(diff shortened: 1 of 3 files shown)


crcrpar:
I think I should modify the initialization of this hook because epsilon value is hardcoded (`1e-12`) currently and it'll cause some problems.

https://github.com/chainer/chainer/pull/6655/files#diff-38a057a4e66f368529e19a843118a8e8R157

chainer/link_hooks/spectral_normalization.py
```diff
@@ -250,8 +272,12 @@ def normalize_weight(self, link, *args, **kwargs):
             W = W / sigma
         if not configuration.config.in_recomputing:
             self.v = v
-            if configuration.config.train:
-                link.xp.copyto(getattr(link, vector_name), u)
+            with chainer.using_device(link.device):
+                if configuration.config.train:
+                    if link.xp is chainerx:
```
toslunar: Why don't you add the comment `# TODO(crcrpar): Remove this when chainerx supports copyto`?

tests/chainer_tests/link_hooks_tests/test_spectral_normalization.py
```diff
@@ -1,14 +1,20 @@
...
+import shutil
+import tempfile
 import unittest
 
 import numpy
 import pytest
 
 import chainer
-from chainer.backends import cuda
+from chainer.backends._cpu import _to_cpu as to_cpu
```
toslunar: Could you import only modules?
`from chainer.backends import _cpu` or `import chainer.backends._cpu as cpu_backend`, etc.

tests/chainer_tests/link_hooks_tests/test_spectral_normalization.py
```diff
@@ -35,6 +41,8 @@ def test_invalid_shaped_weight(self):
 
 class BaseTest(object):
 
+    allclose_options = {'atol': 1e-6, 'rtol': 1e-4}
```
toslunar: Could you avoid setting a mutable object to an attribute?  How about setting in `__init__` or `setUp`.  (The subclasses might have to call `super`'s method.)

tests/chainer_tests/link_hooks_tests/test_spectral_normalization.py
```diff
@@ -176,23 +159,120 @@ def check_u_not_updated_in_test(self, gpu):
...
+        # Test attributes are the same.
+        orig_weight = to_cpu(getattr(layer1, hook1.weight_name).array)
+        orig_vector = to_cpu(getattr(layer1, hook1.vector_name))
+        numpy.testing.assert_array_equal(
+            orig_weight, getattr(layer2, hook2.weight_name).array)
+        numpy.testing.assert_array_equal(
+            orig_vector, getattr(layer2, hook2.vector_name))
+        testing.assert_allclose(y1.array, y2.array, **self.allclose_options)
+
+        shutil.rmtree(root)
```
toslunar: Could you make sure the line is executed even if the test fails?  Do you think `chainer.utils.tempdir` can be used here?

tests/chainer_tests/link_hooks_tests/test_spectral_normalization.py
crcrpar: I forgot to remove this.
I set this because I mistakenly used `numpy.testing.assert_allclose` instead of `testing.assert_allclose` because I overlooked it.

tests/chainer_tests/link_hooks_tests/test_spectral_normalization.py
crcrpar: I didn't know the existence of `chainer.utils.tempdir`. Thank you.

tests/chainer_tests/link_hooks_tests/test_spectral_normalization.py
crcrpar: Sure.

chainer/link_hooks/spectral_normalization.py
```diff
@@ -1,9 +1,14 @@
...
+
+import chainer
 from chainer import backend
 from chainer import configuration
 import chainer.functions as F
 from chainer import link_hook
 import chainer.links as L
 from chainer import variable
+import chainerx
+from chainerx._fallback_workarounds import _from_chx as from_chx_workaround
```
crcrpar: uh-oh...
Do I have to fix this as follows?
```python
import chainerx._fallback_workarounds as fallback

...

xp, device, array = fallback._from_chx(some_array)
```

toslunar: Yes, please.

chainer/link_hooks/spectral_normalization.py
```diff
@@ -225,20 +237,34 @@ def _prepare_parameters(self, link, input_variable=None):
...
+            else:
+                _, s, _ = link.xp.linalg.svd(weight_matrix)
             with link.init_scope():
                 link.gamma = variable.Parameter(s[0], ())
         self._initialized = True
 
-    def normalize_weight(self, link, *args, **kwargs):
+        # N.B.(crcrpar): Tentative adjustment for half precision.
+        if initialW.dtype == numpy.float16:
+            self.eps = 1e-6
```
toslunar: Could you make it documented?  `__init__` can have a default `eps=None` that sets `self.eps = 1e-6` if float16 or `self.eps = 1e-12` otherwise.

toslunar: I discussed it offline, and I found Chainer team reached the consensus (when `eps` of `Adam` is discussed) that hyper-parameters should not depend on precision.

Making `eps=1e-6` default looks good (before v6 release).

crcrpar: That makes sense to me.

One thing I'm concerned with is that the original implementation uses 1e-12.
ref: https://github.com/pfnet-research/sngan_projection/blob/master/source/functions/max_sv.py#L5

chainer/link_hooks/spectral_normalization.py
```diff
@@ -108,7 +108,9 @@ class SpectralNormalization(link_hook.LinkHook):
         n_power_iteration (int): Number of power iteration.
             The default value is 1.
         eps (float): Numerical stability in norm calculation.
-            The default value is 1e-12.
+            The default value is 1e-6 for the compatibility with
```
crcrpar: I changed eps to 1e-6 and mentioned the author's implementation use a different value.
https://github.com/chainer/chainer/pull/6655#discussion_r273389355

toslunar:
Jenkins, test this please.

erhuodaosi:
Hi,

How to install chainer V6.0.0rc1?the latest chainer  version seems to be V5..4.0 by the command of pip install chianer.
DCIAN8%9%QHY4F%X9T8 (XX
I would very appreciate if anyone could help me!Thank you!

Thanks,

erhuodaosi:
It seems that chainer V5.4.0 does not support link_hooks.SpectralNormalization

toslunar:
@erhuodaosi Try `pip install --upgrade --pre chainer --user`.  (`--pre` is required for pre-releases.)

erhuodaosi:
@toslunar 
Hi,

Thank you for your timely reply,I have solved this problem and install chainer V6.0.0rc1 eventually.
But when I use the function of chainer.link_hooks.spectralnormalization',it shows no module name called,how could I solve the problem ?
SPTXW4JT7LR}S{LW}E}C7CO
UO_MNY1KQ8A1)WM2AY58J D

I would very appreciate if anyone could help me!Thank you!

Thanks,

toslunar:
@erhuodaosi 
https://docs.chainer.org/en/v6.0.0rc1/reference/generated/chainer.link_hooks.SpectralNormalization.html has the doc (of the link hook in the version `6.0.0rc1`).
BTW, the link hook was added in #5742, but the enhancement here will be available in the version `6.0.0`.  Google Forum or Slack is a better place to ask a question (see https://github.com/chainer/chainer/blob/master/README.md).

erhuodaosi:
@toslunar
Hi,

Thank you very much!

Thanks,

erhuodaosi:
@toslunar
Hi,

I'm sorry to bother you again,but I have a question in chainer.To be honest,I am a green hand in chainer.If I have a ndarry(cupy.core.core.ndarray) which shape is (1,256,256),how could I change the shape into (1,1,256,256)?

I would very appreciate if you could help me,thank you!

Thanks,

erhuodaosi:
@toslunar
Hi,

I‘m sorry to bother you again,but I have some questions in chainer.To be honest,I am a green hand in chainer. If I have a  'numpy.ndarray' which size is (256,256,1),how to conver it to a image?

I would very appreciate if you could help me,thank you!

Thanks,

PR #6656, opened 2019-03-27
BatchNormalization to only allocate dummy mean and var in cuDNN path

hvy:
`F.batch_normalization` always allocates and updates dummy mean and var if they aren't given by the caller. Those dummies are only required when calling cuDNN so I changed the logic to condition these allocations/updates.


asi1024:
Jenkins, test this please.

asi1024:
LGTM.

PR #6668, opened 2019-03-27
Add `chainerx.meshgrid`

kshitij12345:
This PR intends to add Meshgrid to ChainerX.

* Meshgrid is implemented using existings functions and methods i.e. No device specific implementation.

**TO-DO** 
* ~~Computing Gradient~~
* ~~Update Documentation~~


kshitij12345:
Please review and let me know if it is on the right track.

chainerx_cc/chainerx/routines/creation.cc
```diff
@@ -343,4 +344,62 @@ Array Linspace(
...
+        std::swap(arr_reshapes[1][0], arr_reshapes[1][1]);
+        std::swap(broadcast_shape[0], broadcast_shape[1]);
+    } else if (kind == "ij") {
+    } else {
+        throw ValueError{"Valid values for indexing are \"xy\" or \"ij\""};
+    }
+
+    std::vector<Array> reshaped_arr(arrays.size());
+    for (unsigned int i = 0; i < arrays.size(); i++) {
+        reshaped_arr[i] = arrays[i].Reshape(arr_reshapes[i]);
```
kshitij12345: From what I understand here, Reshape returns a new Array of given shape. But instead is there any `view` function that will return a reshaped view of the same array as values of data remains untouched in further operations.

hvy: Reshape actually checks whether a copy is needed or not, it depends on the inputs. https://github.com/chainer/chainer/blob/master/chainerx_cc/chainerx/routines/manipulation.cc#L233

hvy:
Thanks for the review. We'll try to review this PR as soon as possible but as meshgrid is not of high priority it might delay. FYI, we've created a prioritized spreadsheet for routines to implement. https://docs.google.com/spreadsheets/d/1B4E78tw9Awgpcdn5G7zsQ8NVFYJdOoJlIQg42QxKNfU/edit#gid=0

chainerx_cc/chainerx/routines/creation.cc
kshitij12345: Oh that is great.

kshitij12345:
@hvy Oh sure. But please do review whenever you have time. 

And Thanks for the sheet will try something out this weekend.

kshitij12345:
@hvy Just a headsup that, `Tan,ArcSin,ArcCos,ArcTan`  have been added in #6703. Sheet does not reflect that.

Also please have a look at #6687 , it implements `prod`.

chainerx_cc/chainerx/python/routines.cc
```diff
@@ -275,6 +275,17 @@ void InitChainerxCreation(pybind11::module& m) {
           py::arg("endpoint") = true,
           py::arg("dtype") = nullptr,
           py::arg("device") = nullptr);
+    m.def("meshgrid",
+          [](py::sequence arrays, const nonstd::optional<std::string>& indexing) {
```
hvy: Let's stick to the same names as NumPy where it makes sense, i.e. `xi`.

chainerx_cc/chainerx/python/routines.cc
```diff
@@ -275,6 +275,17 @@ void InitChainerxCreation(pybind11::module& m) {
...
+              std::vector<Array> xs;
+              xs.reserve(arrays.size());
+              std::transform(arrays.begin(), arrays.end(), std::back_inserter(xs), [](const auto& item) {
+                  return Array{py::cast<ArrayBodyPtr>(item)};
+              });
+              return MoveArrayBodies(Meshgrid(xs, indexing));
+          },
+          py::arg("arrays"),
```
hvy: Ditto.

chainerx_cc/chainerx/python/routines.cc
```diff
@@ -275,6 +275,17 @@ void InitChainerxCreation(pybind11::module& m) {
...
+          py::arg("indexing") = nullptr);
```
hvy: Can we default to an actual `xy` value as done in NumPy?

chainerx/_docs/routines.py
```diff
@@ -417,6 +417,30 @@ def _docs_creation():
 .. seealso:: :func:`numpy.diagflat`
 """)
 
+    _docs.set_doc(
+        chainerx.meshgrid,
+        """meshgrid(arrays, indexing="xy")
```
hvy: A nitpick.
```suggestion
        """meshgrid(arrays, indexing='xy')
```

chainerx/_docs/routines.py
```diff
@@ -417,6 +417,30 @@ def _docs_creation():
...
+Return coordinate matrices from coordinate vectors.
```
hvy: Ditto (although NumPy says "Return").
```suggestion
Returns coordinate matrices from coordinate vectors.
```

chainerx_cc/chainerx/routines/creation.cc
```diff
@@ -345,4 +347,119 @@ Array Linspace(
...
+    }
+
+    // Referred from numpy documentation and source.
+    std::string kind = indexing.value_or("xy");
+    if (kind == "xy") {
+        std::swap(arr_reshapes[0][0], arr_reshapes[0][1]);
+        std::swap(arr_reshapes[1][0], arr_reshapes[1][1]);
+        std::swap(broadcast_shape[0], broadcast_shape[1]);
+    } else if (kind == "ij") {
+    } else {
```
hvy: Could you write a comment above this line, just to notify the reader that this empty branch is intentional?

chainerx_cc/chainerx/routines/creation.h
```diff
@@ -120,4 +122,6 @@ Array Linspace(
         const nonstd::optional<Dtype>& dtype = nonstd::nullopt,
         Device& device = GetDefaultDevice());
 
+std::vector<Array> Meshgrid(const std::vector<Array>& arrays, const nonstd::optional<std::string>& indexing);
```
hvy: How about defining an enum for the indexing argument similar to https://github.com/chainer/chainer/blob/master/chainerx_cc/chainerx/routines/pooling.h#L18-L21 and take an instance of that as the argument instead? The input validation can be done in the Python binding glue code and would not be needed here.

tests/chainerx_tests/unit_tests/routines_tests/test_creation.py
```diff
@@ -1173,3 +1176,37 @@ def test_copy(xp, shape, dtype, device, is_module):
...
+            arrs += (numpy.linspace(randint(-10, 0), randint(1, 10),
+                                    randint(3, 7)).astype(self.dtype),)
+
+        return arrs
+
+    def forward_xp(self, inputs, xp):
+        if xp is numpy:
+            outs = xp.meshgrid(*inputs, indexing=self.indexing)
+        elif xp is chainerx:
+            outs = xp.meshgrid(list(inputs), indexing=self.indexing)
```
hvy: Can you alight ChainerX's meshgrid binding so that this conditioning is not needed? I think you can use `py::args` instead of `py::sequence` but I have not tried.

chainerx_cc/chainerx/routines/creation.cc
```diff
@@ -345,4 +347,119 @@ Array Linspace(
...
+    bb.Finalize();
+}
+
+}  // namespace
+
+std::vector<Array> Meshgrid(const std::vector<Array>& arrays, const nonstd::optional<std::string>& indexing) {
+    Shape shape;
+    Shape broadcast_shape;
+    std::vector<Shape> arr_reshapes;
+    std::vector<Array> grid_arrs(arrays.size());
```
hvy: A detail but we haven't used the "arrs" abbreviation in ChainerX. Could you simply use "out" which we often use or something along "grid_arrays"?

chainerx_cc/chainerx/routines/creation.cc
```diff
@@ -345,4 +347,119 @@ Array Linspace(
...
+    // (n1, 1, 1) , Vector 2 as (1, n2, 1)
+    //
+    // Step 2: Broadcast each vector to the shape
+    // if (indexing == "ij") -> (n1, n2, n3)
+    // else if (indexing == "xy") -> (n2, n1, n3)
+    // Note : For "xy" only n1 and n2 swap their places
+    //        all others are same as "ij"
+
+    // Step 1
+    for (const auto& array : arrays) {
```
hvy: A detail but we are quite conservative with `auto`. Could you be explicit with the type for consistency/to avoid confusion?

chainerx_cc/chainerx/routines/creation.cc
```diff
@@ -345,4 +347,119 @@ Array Linspace(
...
+        shape.push_back(1);
+        broadcast_shape.push_back(array.GetTotalSize());
+    }
+
+    // Shape for each array based on number of arrays.
+    for (unsigned int i = 0; i < arrays.size(); i++) {
```
hvy: Another detail but for consistency. Also, we in general use pre-increment`
```suggestion
    for (size_t i = 0; i < arrays.size(); ++i) {
```

chainerx_cc/chainerx/routines/creation.cc
```diff
@@ -345,4 +347,119 @@ Array Linspace(
...
+    } else if (kind == "ij") {
+    } else {
+        throw ValueError{R"(Valid values for indexing are "xy" or "ij")"};
+    }
+
+    {
+        NoBackpropModeScope scope;
+
+        std::vector<Array> reshaped_arr(arrays.size());
+        for (unsigned int i = 0; i < arrays.size(); i++) {
```
hvy: Ditto.

chainerx_cc/chainerx/routines/creation.cc
```diff
@@ -345,4 +347,119 @@ Array Linspace(
...
+            reshaped_arr[i] = arrays[i].Reshape(arr_reshapes[i]);
+        }
+
+        // Step 2
+        for (unsigned int i = 0; i < arrays.size(); i++) {
```
hvy: Ditto.

chainerx_cc/chainerx/routines/creation.cc
```diff
@@ -345,4 +347,119 @@ Array Linspace(
...
+        std::swap(broadcast_shape[0], broadcast_shape[1]);
+    } else if (kind == "ij") {
+    } else {
+        throw ValueError{R"(Valid values for indexing are "xy" or "ij")"};
+    }
+
+    {
+        NoBackpropModeScope scope;
+
+        std::vector<Array> reshaped_arr(arrays.size());
```
hvy: Could you create an empty container, reserve and emplace_back? To reduce number of shared_ptrs created?

chainerx_cc/chainerx/routines/creation.cc
```diff
@@ -345,4 +347,119 @@ Array Linspace(
     return out;
 }
 
+namespace {
+
+// Defines the backward pass for Meshgrid.
+void MeshgridBackward(const std::vector<Array>& inputs, const std::vector<Array>& outputs, const std::string& kind) {
```
hvy: How about `MeshgridGrad` which itself has a routine-like interface and behavior. We have other routines which are implemented that way. Then, the graph would be created in `Meshgrid`, but inside the backward definition, `MeshgridGrad` would be called to do the computation of the gradient. What do you think?

hvy:
Sorry for the delay. I wrote some comments, mostly about the interfaces and not so much on the implementation. Though you'll see some nitpicks. Let me know what you think.

tests/chainerx_tests/unit_tests/routines_tests/test_creation.py
```diff
@@ -1173,3 +1176,37 @@ def test_copy(xp, shape, dtype, device, is_module):
...
+
+@op_utils.op_test(['native:0', 'cuda:0'])
+@chainer.testing.parameterize_pytest('indexing', ['xy', 'ij'])
+@chainer.testing.parameterize_pytest('input_arrs', [2, 3, 4, 5, 6])
+class TestMeshgrid(op_utils.NumpyOpTest):
+
+    check_numpy_strides_compliance = False
+
+    def setup(self, dtype):
+        # TODO(kshitij12345) : Remove when #6621 is in.
```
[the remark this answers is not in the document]
kshitij12345: Removed.

tests/chainerx_tests/unit_tests/routines_tests/test_creation.py
```diff
@@ -1,7 +1,9 @@
 from io import StringIO
 import sys
 import tempfile
+from random import randint
```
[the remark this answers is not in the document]
kshitij12345: Fixed.

chainerx_cc/chainerx/python/routines.cc
kshitij12345: Done.

chainerx_cc/chainerx/python/routines.cc
kshitij12345: Using `py::args` now.

chainerx_cc/chainerx/python/routines.cc
kshitij12345: Done.

chainerx/_docs/routines.py
kshitij12345: Done.

chainerx/_docs/routines.py
kshitij12345: Done.

chainerx_cc/chainerx/routines/creation.cc
kshitij12345: Refactored.

chainerx_cc/chainerx/routines/creation.h
kshitij12345: Input validation moved to binding code.
Have directly used string instead as we are certain that there will only be 2 modes, I felt having a `enum` for these two would be over-kill. Your thoughts?

tests/chainerx_tests/unit_tests/routines_tests/test_creation.py
kshitij12345: Done.

chainerx_cc/chainerx/routines/creation.cc
kshitij12345: Changed to `grid_arrays`.

chainerx_cc/chainerx/routines/creation.cc
kshitij12345: Done.

chainerx_cc/chainerx/routines/creation.cc
kshitij12345: Done.

chainerx_cc/chainerx/routines/creation.cc
kshitij12345: Done.

chainerx_cc/chainerx/routines/creation.cc
kshitij12345: Done.

chainerx_cc/chainerx/routines/creation.cc
kshitij12345: Done.

chainerx_cc/chainerx/routines/creation.cc
kshitij12345: Turns out, that we can rely on existing machinery so there is no need for a `BackwardBuilder`.

kshitij12345:
Pending
* Invalid test cases.
* Validating the `indexing` arguments value.

chainerx_cc/chainerx/python/routines.cc
```diff
@@ -275,6 +275,20 @@ void InitChainerxCreation(pybind11::module& m) {
           py::arg("endpoint") = true,
           py::arg("dtype") = nullptr,
           py::arg("device") = nullptr);
+    m.def("meshgrid", [](py::args xi, py::kwargs kwargs) {
+        std::vector<Array> xs;
+        std::string indexing("xy");
```
hvy: Could you use list initialization? Just a better practice. https://stackoverflow.com/questions/18222926/why-is-list-initialization-using-curly-braces-better-than-the-alternatives
```suggestion
        std::string indexing{"xy"};
```

chainerx_cc/chainerx/routines/creation.h
hvy: I see your point but I must say I think it's a better API design choice reducing possible errors. It also simplifies input validation.

chainerx_cc/chainerx/python/routines.cc
kshitij12345: oops sorry. I tend to miss that. Fixed.

chainerx_cc/chainerx/routines/creation.h
kshitij12345: Have used `enum`, have to agree it is more clear in terms of valid choices. Thank You.

chainerx_cc/chainerx/routines/creation.h
```diff
@@ -120,4 +122,11 @@ Array Linspace(
         const nonstd::optional<Dtype>& dtype = nonstd::nullopt,
         Device& device = GetDefaultDevice());
 
+enum class MeshgridIndexingMode {
+    xy = 1,
+    ij,
```
hvy: How about `MeshgridIndexingMode::{kCartesian,kMatrix}` ? A bit clearer I think. And prefixing with `k` is per convention.

chainerx_cc/chainerx/routines/creation.cc
```diff
@@ -345,4 +347,60 @@ Array Linspace(
     return out;
 }
 
+std::vector<Array> Meshgrid(const std::vector<Array>& arrays, MeshgridIndexingMode mode) {
+    Shape shape;
+    Shape broadcast_shape;
+    std::vector<Shape> arr_reshapes;
```
hvy: `arr_reshapes` can also be `reserve`d? And a detail but how about keeping the names somewhat aligned `s/arr/array`.

chainerx_cc/chainerx/routines/creation.cc
```diff
@@ -345,4 +347,60 @@ Array Linspace(
...
+
+    // Step 1
+    for (const Array& array : arrays) {
+        shape.emplace_back(1);
+        broadcast_shape.emplace_back(array.GetTotalSize());
+    }
+
+    // Shape for each array based on number of arrays.
+    for (size_t i = 0; i < arrays.size(); ++i) {
+        Shape temp_shape(shape.begin(), shape.end());
```
hvy: List initialization here too?

chainerx_cc/chainerx/routines/creation.cc
```diff
@@ -345,4 +347,60 @@ Array Linspace(
...
+        temp_shape[i] = arrays[i].GetTotalSize();
+        arr_reshapes.push_back(temp_shape);
```
hvy: This is also a detail and you may or may not skip this, but we almost always prefer `emplace_back` over `push_back` since there are basically not pros with the latter and the latter falls back to the former.

chainerx_cc/chainerx/routines/creation.h
```diff
@@ -3,6 +3,8 @@
 #include <cstddef>
 #include <cstdint>
 #include <memory>
+#include <string>
```
hvy: Ah, strings are no longer used?

kshitij12345: Fixed.

chainerx_cc/chainerx/routines/creation.cc
kshitij12345: Oh Thank you. Fixed.

chainerx_cc/chainerx/routines/creation.cc
kshitij12345: Fixed.

chainerx_cc/chainerx/routines/creation.cc
kshitij12345: reserve done.
> And a detail but how about keeping the names somewhat aligned s/arr/array.

Didn't understand this part exactly.
Have renamed `arr_reshapes` to `broadcasted_array_shapes`

chainerx_cc/chainerx/routines/creation.h
kshitij12345: Done.

chainerx_cc/chainerx/error.h
```diff
@@ -99,4 +99,10 @@ class GradientCheckError : public ChainerxError {
     using ChainerxError::ChainerxError;
 };
 
+// Error on using invalid contexts.
+class ValueError : public ChainerxError {
```
hvy: Sorry but where is this exception thrown?

chainerx_cc/chainerx/routines/creation.cc
```diff
@@ -5,6 +5,7 @@
 #include <cstddef>
 #include <cstdint>
 #include <memory>
+#include <string>
```
hvy: It seems like `<string>` is not used?

chainerx_cc/chainerx/error.h
kshitij12345: Oops. Used to throw it previously. Now we don't. Will remove.

chainerx_cc/chainerx/routines/creation.cc
kshitij12345: Oops. Sorry missed this one.

hvy:
`chx.meshgrid()` (call with no arguments) causes a segfault. Could you fix it behave similar to NumPy, i.e. return an empty list?

kshitij12345:
Fixed. 
No worries. Always happy to learn and get things right. Thank You.

hvy:
Jenkins, test this please.

hvy:
Jenkins, test this please.

kshitij12345:
@hvy could you please have a look and test this one.

chainerx_cc/chainerx/routines/creation.cc
```diff
@@ -341,4 +341,73 @@ Array Linspace(
...
+    std::vector<Array> grid_arrays;
+
+    // special cases
+    // similar behavior to numpy.
+    if (arrays.empty()) {
+        return grid_arrays;
+    }
+
+    if (arrays.size() == 1) {
+        grid_arrays.emplace_back(arrays[0]);
```
hvy: Can you flatten the array to be NumPy compatible here?

hvy:
@kshitij12345 I'm really sorry for the delayed comment. It basically LGTM but I left a smaller comment. If you are busy, I can take other the PR so just let me know. Thanks.

kshitij12345:
@hvy No worries. I would like to continue working on this PR as well as keep contributing to the project.

hvy:
Jenkins, test this please.

hvy:
Jenkins, test this please.

hvy:
Jenkins, test this please.

hvy:
Jenkins, test this please.

hvy:
@kshitij12345 And merged at last. Thanks for your contribution as always.

hvy:
@kshitij12345 And merged at last. Thanks for your contribution as always.

PR #6677, opened 2019-03-28
Minimize the Travis CI matrix

beam2d:
I would like to redefine the purpose of Travis CI to quick automatic checks for all contributors. In this regard, recent situations of Travis CI's long job queue is not desirable. This change reduces the number of jobs running on Travis CI by reducing the jobs following the redefined purpose.

- Checks that are covered by Jenkins will only run for PRs (master/stable branches are removed).
- Windows tests are removed (we will use our new CI infrastructure for them).
- Only one combination of Ubuntu/py environment is used for each PR (namely Ubuntu16.04 py35).
- macOS py27 test will run for each PR instead of macOS py35 (for wider py versions coverage combined with Ubuntu test).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ matrix:
       python: "3.5"
       env:
       - CHAINER_TRAVIS_TEST="python-static-check"
+      if: type in (pull_request)
 
     - name: "C++ Static Check"
       language: python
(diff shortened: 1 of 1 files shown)


niboshi:
Jenkins, test this please

niboshi:
Test failures are irrelevant.
LGTM

PR #6678, opened 2019-03-28
Add `WeightStandardization` link hook

hitsgub:
This PR adds [Weight Standardization](https://arxiv.org/abs/1903.10520).
Please wait for the experimental result of CIFAR-10 until tomorrow.

### DIFF
--- a/chainer/functions/normalization/_standardize.py
+++ b/chainer/functions/normalization/_standardize.py
@@ -87,18 +87,15 @@ class Standardize(function_node.FunctionNode):
 
 
 def _standardize(x, eps=1e-6):
-    """Standardize for `Weight standardization
-    <https://arxiv.org/abs/1903.10520>`_.
+    """Weight Standardization function.
 
     This function implements a "weight standardization"
     which standardizes the input weights by statistics
     that are computed except the first axis.
 
-
     Args:
         x (:class:`~chainer.Variable` or :ref:`ndarray`): Weight vectors.
             e.g., the input of :func:`~chainer.functions.convolution_2d`.
-
 
     Returns:
         ~chainer.Variable: The output variable which has the same shape
--- a/chainer/functions/normalization/_standardize.py
+++ b/chainer/functions/normalization/_standardize.py
@@ -33,18 +33,19 @@ class Standardize(function_node.FunctionNode):
         squ_x_mu = xp.square(x_mu)
         var = xp.mean(squ_x_mu, axis=axes, keepdims=True)
         if xp in (numpy, cuda.cupy):
-            std = xp.sqrt(var, dtype=x.dtype) + self.eps
+            std_noeps = xp.sqrt(var, dtype=x.dtype)
         else:
-            std = xp.sqrt(var) + self.eps
+            std_noeps = xp.sqrt(var)
+        std = std_noeps + self.eps
         inv_std = 1. / std
         x_hat = x_mu * inv_std
-        return x_mu, var, inv_std, x_hat
+        return x_mu, var, std_noeps, inv_std, x_hat
 
     def forward(self, inputs):
         self.retain_inputs((0,))
         xp = backend.get_array_module(*inputs)
         x, = inputs
-        x_mu, var, inv_std, x_hat = self._compute(xp, x)
+        x_mu, var, std_noeps, inv_std, x_hat = self._compute(xp, x)
         return x_hat,
 
     def backward(self, indexes, grad_outputs):
@@ -53,13 +54,13 @@ class Standardize(function_node.FunctionNode):
         gy, = grad_outputs
         axes = tuple(six.moves.range(1, len(x.shape)))
 
-        x_mu, var, inv_std, x_hat = self._compute(F, x)
+        x_mu, var, std_noeps, inv_std, x_hat = self._compute(F, x)
 
-        g_inv_std = F.sum(gy * x_mu, axis=axes, keepdims=True)
         g_x_mu_1 = gy * inv_std
 
+        g_inv_std = F.sum(gy * x_mu, axis=axes, keepdims=True)
         g_std = g_inv_std * (- 1. / var)
-        g_var = g_std * 0.5 * inv_std
+        g_var = g_std * 0.5 / std_noeps
 
         n_units = x.size / x.shape[0]
         g_squ_x_mu = g_var * (1. / n_units)
--- a/chainer/functions/normalization/_standardize.py
+++ b/chainer/functions/normalization/_standardize.py
@@ -62,7 +62,7 @@ class Standardize(function_node.FunctionNode):
         self.retain_inputs((0,))
         x, = inputs
         xp = backend.get_array_module(x)
-        x_mu, std_noeps, std, x_hat = self._compute(xp, x)
+        x_mu, std_noeps, std, x_hat, axes = self._compute(xp, x)
         return x_hat,
 
     def backward(self, indexes, grad_outputs):
--- a/chainer/functions/normalization/_standardize.py
+++ b/chainer/functions/normalization/_standardize.py
@@ -33,12 +33,12 @@ class Standardize(function_node.FunctionNode):
         squ_x_mu = xp.square(x_mu)
         var = xp.mean(squ_x_mu, axis=axes, keepdims=True)
         if xp is numpy:
-            std = numpy.sqrt(var + self.eps, dtype=x.dtype)
-            inv_std = 1. / std
+            std = numpy.sqrt(var, dtype=x.dtype) + self.eps
         elif xp is cuda.cupy:
-            inv_std = cuda.cupyx.rsqrt(var + self.eps, dtype=x.dtype)
+            std = cuda.cupyx.sqrt(var, dtype=x.dtype) + self.eps
         else:
-            inv_std = xp.rsqrt(var + self.eps)
+            std = xp.sqrt(var) + self.eps
+        inv_std = 1. / std
         x_hat = x_mu * inv_std
         return x_mu, var, inv_std, x_hat
 
--- a/chainer/functions/normalization/_standardize.py
+++ b/chainer/functions/normalization/_standardize.py
@@ -56,8 +56,7 @@ class Standardize(function_node.FunctionNode):
         else:
             std_noeps = xp.sqrt(var)
         std = std_noeps + self.eps
-        inv_std = xp.reciprocal(std)
-        x_hat = x_mu * inv_std
+        x_hat = x_mu / std
         return x_mu, std_noeps, std, x_hat
 
     def forward(self, inputs):
--- a/chainer/functions/normalization/_standardize.py
+++ b/chainer/functions/normalization/_standardize.py
@@ -3,7 +3,6 @@ import six
 from chainer import backend
 from chainer import function_node
 import chainer.functions
-from chainer import utils
 from chainer.utils import type_check
 
 
@@ -31,7 +30,7 @@ class Standardize(function_node.FunctionNode):
     """Standardization for `Weight standardization
         <https://arxiv.org/abs/1903.10520>`_"""
 
-    def __init__(self, eps=1e-5):
+    def __init__(self, eps=1e-6):
         self.eps = eps
 
     def check_type_forward(self, in_types):
@@ -55,7 +54,7 @@ class Standardize(function_node.FunctionNode):
         else:
             std_noeps = xp.sqrt(var, dtype=x.dtype)
         std = std_noeps + x.dtype.type(self.eps)
-        x_hat = utils.force_array(x_mu / std)
+        x_hat = x_mu / std
         return x_mu, std_noeps, std, x_hat, axes
 
     def forward(self, inputs):
@@ -71,29 +70,23 @@ class Standardize(function_node.FunctionNode):
         gy, = grad_outputs
 
         x_mu, std_noeps, std, x_hat, axes = self._compute(F, x)
-
         g_x_mu_1 = std * gy
-
         g_std = F.mean((x_mu * gy), axis=axes, keepdims=True)
 
         # _standardize with eps has continuous backward. However,
         # the backward is not differentiable for the indices of zero vectors.
         # To avoid nan in double backward, do not compute outside of mask.
         mask = std_noeps.array != 0
-        g_var, = _SetItemZero(mask).apply((
-            g_std[mask] / std_noeps[mask],))
+        g_var, = _SetItemZero(mask).apply((g_std[mask] / std_noeps[mask],))
 
         g_x_mu_2 = g_var * x_mu
-
         g_x_1 = (g_x_mu_1 - g_x_mu_2) / (std ** 2)
         g_mu = F.mean(g_x_1, axis=axes, keepdims=True)
 
-        g_x = g_x_1 - g_mu
-
-        return g_x,
+        return g_x_1 - g_mu,
 
 
-def _standardize(x, eps=1e-5):
+def _standardize(x, eps=1e-6):
     """Standardize for `Weight standardization
     <https://arxiv.org/abs/1903.10520>`_.
 
--- a/chainer/functions/normalization/_standardize.py
+++ b/chainer/functions/normalization/_standardize.py
@@ -5,6 +5,7 @@ from chainer import backend
 from chainer.backends import cuda
 from chainer import function_node
 import chainer.functions
+from chainer import utils
 from chainer.utils import type_check
 
 
@@ -51,18 +52,18 @@ class Standardize(function_node.FunctionNode):
         x_mu = x - mu
         squ_x_mu = xp.square(x_mu)
         var = xp.mean(squ_x_mu, axis=axes, keepdims=True)
-        if xp in (numpy, cuda.cupy):
-            std_noeps = xp.sqrt(var, dtype=x.dtype)
-        else:
+        if xp is chainer.functions:
             std_noeps = xp.sqrt(var)
-        std = std_noeps + self.eps
-        x_hat = x_mu / std
+        else:
+            std_noeps = xp.sqrt(var, dtype=x.dtype)
+        std = std_noeps + x.dtype.type(self.eps)
+        x_hat = utils.force_array(x_mu / std)
         return x_mu, std_noeps, std, x_hat
 
     def forward(self, inputs):
         self.retain_inputs((0,))
-        xp = backend.get_array_module(*inputs)
         x, = inputs
+        xp = backend.get_array_module(x)
         x_mu, std_noeps, std, x_hat = self._compute(xp, x)
         return x_hat,
 
--- a/chainer/functions/normalization/_standardize.py
+++ b/chainer/functions/normalization/_standardize.py
@@ -56,15 +56,14 @@ class Standardize(function_node.FunctionNode):
         else:
             std_noeps = xp.sqrt(var)
         std = std_noeps + self.eps
-        inv_std = 1. / std
-        x_hat = x_mu * inv_std
-        return x_mu, var, std_noeps, inv_std, x_hat
+        x_hat = x_mu / std
+        return x_mu, std_noeps, std, x_hat
 
     def forward(self, inputs):
         self.retain_inputs((0,))
         xp = backend.get_array_module(*inputs)
         x, = inputs
-        x_mu, var, std_noeps, inv_std, x_hat = self._compute(xp, x)
+        x_mu, std_noeps, std, x_hat = self._compute(xp, x)
         return x_hat,
 
     def backward(self, indexes, grad_outputs):
@@ -73,12 +72,12 @@ class Standardize(function_node.FunctionNode):
         gy, = grad_outputs
         axes = tuple(six.moves.range(1, len(x.shape)))
 
-        x_mu, var, std_noeps, inv_std, x_hat = self._compute(F, x)
+        x_mu, std_noeps, std, x_hat = self._compute(F, x)
 
-        g_x_mu_1 = gy * inv_std
+        g_x_mu_1 = gy / std
 
         g_inv_std = F.sum(gy * x_mu, axis=axes, keepdims=True)
-        g_std = g_inv_std * (- 1. / var)
+        g_std = g_inv_std * (- 1. / std ** 2)
 
         # _standardize with eps has continuous backward. However,
         # the backward is not differentiable for the indices of zero vectors.
--- a/chainer/functions/normalization/_standardize.py
+++ b/chainer/functions/normalization/_standardize.py
@@ -70,7 +70,7 @@ class Standardize(function_node.FunctionNode):
         F = chainer.functions
         x, = self.get_retained_inputs()
         gy, = grad_outputs
-        axes = tuple(six.moves.range(1, len(x.shape)))
+        axes = tuple(six.moves.range(1, x.ndim))
 
         x_mu, std_noeps, std, x_hat = self._compute(F, x)
 
--- a/chainer/functions/normalization/_standardize.py
+++ b/chainer/functions/normalization/_standardize.py
@@ -86,8 +86,7 @@ class Standardize(function_node.FunctionNode):
         g_x_mu_2 = g_var * x_mu
 
         g_x_1 = (g_x_mu_1 - g_x_mu_2) / (std ** 2)
-        g_mu, = _SetItemZero(mask).apply((
-            F.mean(g_x_1[mask], axis=axes, keepdims=True),))
+        g_mu = F.mean(g_x_1, axis=axes, keepdims=True)
 
         g_x = g_x_1 - g_mu
 
--- a/chainer/functions/normalization/_standardize.py
+++ b/chainer/functions/normalization/_standardize.py
@@ -1,8 +1,6 @@
-import numpy
 import six
 
 from chainer import backend
-from chainer.backends import cuda
 from chainer import function_node
 import chainer.functions
 from chainer import utils
@@ -47,7 +45,7 @@ class Standardize(function_node.FunctionNode):
 
     def _compute(self, xp, x):
         # xp: numpy, cupy, or chainer.functions
-        axes = tuple(six.moves.range(1, len(x.shape)))
+        axes = tuple(six.moves.range(1, x.ndim))
         mu = xp.mean(x, axis=axes, keepdims=True)
         x_mu = x - mu
         squ_x_mu = xp.square(x_mu)
@@ -58,7 +56,7 @@ class Standardize(function_node.FunctionNode):
             std_noeps = xp.sqrt(var, dtype=x.dtype)
         std = std_noeps + x.dtype.type(self.eps)
         x_hat = utils.force_array(x_mu / std)
-        return x_mu, std_noeps, std, x_hat
+        return x_mu, std_noeps, std, x_hat, axes
 
     def forward(self, inputs):
         self.retain_inputs((0,))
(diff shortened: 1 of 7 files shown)


chainer/functions/normalization/weight_standardization.py
```diff
@@ -0,0 +1,101 @@
...
+        type_check.expect(
+            x_type.dtype.kind == 'f',
+            x_type.ndim >= 2,
+        )
+
+    def _compute(self, xp, x):
+        # xp: numpy, cupy, or chainer.functions
+        axes = tuple(six.moves.range(1, len(x.shape)))
+        mu = xp.mean(x, axis=axes, keepdims=True)
+        x_mu = x - _broadcast_to(xp, mu, x.shape)
```
toslunar: Automatic broadcast could be used (#4679).

chainer/functions/normalization/weight_standardization.py
```diff
@@ -0,0 +1,101 @@
...
+        squ_x_mu = xp.square(x_mu)
+        var = xp.mean(squ_x_mu, axis=axes, keepdims=True)
+        std = xp.sqrt(var + self.eps)
+        inv_std = 1. / std
+        x_hat = x_mu * _broadcast_to(xp, inv_std, x_mu.shape)
```
toslunar: ditto

chainer/functions/normalization/weight_standardization.py
```diff
@@ -0,0 +1,101 @@
+from chainer import cuda
+from chainer import function_node
+import chainer.functions
+from chainer.utils import type_check
+import six
```
toslunar: Could you group imports?
```
import six

from chainer import cuda
import chainer.functions
```

chainer/links/connection/convolution_2d.py
```diff
@@ -179,8 +191,11 @@ def forward(self, x):
         """
         if self.W.array is None:
             self._initialize_params(x.shape[1])
+        W = self.W
+        if self.standardizeW:
+            W = weight_standardization.weight_standardization(W, self.eps)
```
toslunar: I'd like to implement the feature by `LinkHook`.  See #5742 for example.  It's OK to separate a PR for it (in order to add `F.weight_standardization` first).

chainer/links/connection/convolution_2d.py
```diff
@@ -51,6 +54,10 @@ class Convolution2D(link.Link):
             convolution operation will be executed independently. Input channel
             size ``in_channels`` and output channel size ``out_channels`` must
             be exactly divisible by this value.
+        standardizeW (bool): A flag to use the weight standardization from the
+            paper: `Weight Standardization <https://arxiv.org/abs/1903.10520>`_
+        eps (float): Small value for the numerical stability in
```
toslunar: The name `eps` is too broad to be used in weight standardization.

chainer/links/connection/convolution_2d.py
crcrpar: https://github.com/crcrpar/chainer/blob/weight-standardization-hook/chainer/link_hooks/weight_standardization.py
Though it's in progress, I have tried to implement this in Link hook, so feel free to copy this if helpful.

chainer/functions/normalization/weight_standardization.py
hitsgub: If using `chainer.functions.broadcast_to`, we will get `chainer.variable` instead of `cupy.ndarray`.
Do you suggest it is better to use `variable` isntead of `ndarray` in those computation?

chainer/links/connection/convolution_2d.py
hitsgub: Thanks @crcrpar !
Your code is very helpful for my new implementation in `link_hook`.

chainer/links/connection/convolution_2d.py
hitsgub: Since re-implementing as `link_hook`, This arguments `eps` in `Convolution2D` is removed.

chainer/functions/normalization/weight_standardization.py
toslunar: I meant `x_mu = x - mu` works.

toslunar: I'm sorry.  I should have commented to the lines in `backward`.

chainer/functions/normalization/standardize.py
```diff
@@ -0,0 +1,104 @@
...
+    def backward(self, indexes, grad_outputs):
+        F = chainer.functions
+        x, = self.get_retained_inputs()
+        gy, = grad_outputs
+        axes = tuple(six.moves.range(1, len(x.shape)))
+
+        x_mu, var, inv_std, x_hat = self._compute(F, x)
+
+        g_inv_std = F.sum(gy * x_mu, axis=axes, keepdims=True)
+        g_x_mu_1 = gy * F.broadcast_to(inv_std, gy.shape)
```
toslunar:
```suggestion
        g_x_mu_1 = gy * inv_std
```

chainer/functions/normalization/standardize.py
```diff
@@ -0,0 +1,104 @@
...
+
+        g_std = g_inv_std * (- 1. / var)
+        g_var = g_std * 0.5 * inv_std
+
+        n_units = x.size / x.shape[0]
+        g_squ_x_mu = F.broadcast_to(g_var * (1. / n_units), x.shape)
```
toslunar:
```suggestion
        g_squ_x_mu = g_var * (1. / n_units)
```

chainer/functions/normalization/standardize.py
hitsgub: Thank you, and I notice same duplication in `layer_normalization`.
I will PR this.

chainer/functions/normalization/weight_standardization.py
hitsgub: Ah! I see, I fixed it.

hitsgub:
> * It's possible that the name of function would be `standardize(x, eps, axis)`.
> * I'm wondering if `cudnn.batch_normalization_forward_training` speeds up while the moving averages are not needed.

hmm...
If using `cudnn.batch_normalization_forward_training`, 
```
normalized_weight, _, _ = (
    cudnn.batch_normalization_forward_training(
        weight, gamma, beta, running_mean, running_var, None, None, eps, 0, ...))
```
Four more vector all 1 or 0 is needed (gamma, beta, running_mean, running_var).
I wonder memory consumption, how about your opinion?

tests/chainer_tests/link_hooks_tests/test_weight_standardization.py
```diff
@@ -0,0 +1,190 @@
...
+
+    def test_deleted_cpu(self):
+        self.check_deleted(False)
+
+    @attr.gpu
+    def test_deleted_gpu(self):
+        self.check_deleted(True)
+
+
+class TestEmbedID(unittest.TestCase, BaseTest):
```
crcrpar: I don't think this necessary because WeightStandardization is proposed for 2D convolutional layers followed by Batch/Group Normalization.

What do you think?

tests/chainer_tests/link_hooks_tests/test_weight_standardization.py
```diff
@@ -0,0 +1,190 @@
...
+        layer.add_hook(hook)
+        if self.lazy_init:
+            with chainer.using_config('train', False):
+                layer(self.x)
+
+
+@testing.parameterize(*testing.product({
+    'lazy_init': [True, False],
+}))
+class TestLinear(unittest.TestCase, BaseTest):
```
crcrpar: Rephrase the comment for `TestEmbedID`.

tests/chainer_tests/link_hooks_tests/test_weight_standardization.py
```diff
@@ -0,0 +1,190 @@
...
+            size=(self.bs, self.in_size)).astype(numpy.float32)
+        self.layer = L.Linear(self.out_size)  # Lazy initialization
+        in_size = None if self.lazy_init else self.in_size
+        self.layer = L.Linear(in_size, self.out_size)
+        self.hook = WeightStandardization()
+
+
+@testing.parameterize(*testing.product({
+    'lazy_init': [True, False],
+    'link': [L.Convolution1D, L.Deconvolution1D],
```
crcrpar: Do we need a test for Deconvolution layer?

tests/chainer_tests/link_hooks_tests/test_weight_standardization.py
```diff
@@ -0,0 +1,190 @@
...
+            in_channels, self.out_channels, **conv_init_args)
+        self.x = numpy.random.normal(
+            size=(5, self.in_channels, 4)).astype(numpy.float32)
+        self.hook = WeightStandardization()
+        self.out_size = self.out_channels  # For compatibility
+
+
+@testing.parameterize(*testing.product({
+    'lazy_init': [True, False],
+    'link': [L.Convolution2D, L.Deconvolution2D],
```
crcrpar: ditto

tests/chainer_tests/link_hooks_tests/test_weight_standardization.py
```diff
@@ -0,0 +1,190 @@
...
+            in_channels, self.out_channels, **conv_init_args)
+        self.x = numpy.random.normal(
+            size=(5, self.in_channels, 4, 4)).astype(numpy.float32)
+        self.hook = WeightStandardization()
+        self.out_size = self.out_channels  # For compatibility
+
+
+@testing.parameterize(*testing.product({
+    'lazy_init': [True, False],
+    'link': [L.Convolution3D, L.Deconvolution3D],
```
crcrpar: ditto

chainer/functions/normalization/standardize.py
```diff
@@ -0,0 +1,93 @@
...
+        squ_x_mu = xp.square(x_mu)
+        var = xp.mean(squ_x_mu, axis=axes, keepdims=True)
+        std = xp.sqrt(var + self.eps)
+        inv_std = 1. / std
+        x_hat = x_mu * inv_std
+        return x_mu, var, inv_std, x_hat
+
+    def forward(self, inputs):
+        self.retain_inputs((0,))
+        xp = cuda.get_array_module(*inputs)
```
crcrpar: https://github.com/chainer/chainer/pull/6678/files#r270065458 is for this.
```suggestion
        xp = backend.get_array_module(*inputs)
```

chainer/functions/normalization/standardize.py
```diff
@@ -0,0 +1,93 @@
...
+        g_mu = F.sum(g_x_1, axis=axes, keepdims=True) * (- 1.)
+
+        g_x_2 = g_mu * (1. / n_units)
+
+        g_x = g_x_1 + g_x_2
+
+        return g_x,
+
+
+def standardize(x, eps=1e-5):
```
toslunar: At the moment, could you make the function private?

- From the NumPy point of view, the default axis can be `None`, which means all the axes.
- To be consistent with `F.normalize`, it might be reasonable to make `axis=1` the default.
- There seems no consensus on the axis for deconvolutions yet.

hitsgub:
Please wait for a while response to reviews.
This is experimental result.

- ResNet20(3-3-3) with pre-activation
- 300epoch with reducing learning rate, 0-149: 0.05, 150-224: 0.005, 225-299: 0.0005.
- NesterovAG, batchsize=128

weight_standardization

Like said in original paper, GN+WS and BN+WS are better than GN.
BN is the best in this experiments because batchsize is large.

chainer/functions/normalization/standardize.py
```diff
@@ -0,0 +1,93 @@
...
+        )
+
+    def _compute(self, xp, x):
+        # xp: numpy, cupy, or chainer.functions
+        axes = tuple(six.moves.range(1, len(x.shape)))
+        mu = xp.mean(x, axis=axes, keepdims=True)
+        x_mu = x - mu
+        squ_x_mu = xp.square(x_mu)
+        var = xp.mean(squ_x_mu, axis=axes, keepdims=True)
+        std = xp.sqrt(var + self.eps)
```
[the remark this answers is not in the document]
hitsgub: Thanks, `rsqrt` seems to be better.

chainer/functions/normalization/standardize.py
hitsgub: Sure, I agree with your opinion that the function `standardize` is irregular.
About deconvolution, I plan to see off correspondence to deconvolutions (since done our experiment or publication of new paper).
Do you mean remove this line?
https://github.com/chainer/chainer/pull/6678/files/fa2567f0e9ba4eacb3a232b02a8083e3cb3801df#diff-75561f28993ff21bfbaad90b8b4ccb74R204

tests/chainer_tests/link_hooks_tests/test_weight_standardization.py
hitsgub: Not be needed, removed it.

chainer/link_hooks/weight_standardization.py
```diff
@@ -0,0 +1,67 @@
...
+class WeightStandardization(link_hook.LinkHook):
+    """Weight Standardization link hook implementation.
+
+    This hook standardizes a weight by *weight statistics*.
+
+    See: Siyuan Qiao et. al., `Weight Standardization
+    <https://arxiv.org/abs/1903.10520>`_
+
+    Args:
+    eps (int): Numerical stability in standard deviation calculation.
```
crcrpar:
```suggestion
    eps (float): Numerical stability in standard deviation calculation.
```

I know the SN hook link implementation has this typo and I'm fixing this now. 🙇

chainer/link_hooks/weight_standardization.py
```diff
@@ -0,0 +1,67 @@
+import chainer.functions as F
+from chainer import link_hook
+
+
+class WeightStandardization(link_hook.LinkHook):
+    """Weight Standardization link hook implementation.
```
crcrpar: Personally, it should be noted that this hook is originally proposed for 2D convolution layers followed by mainly GN and sometimes BN.
Also, I want some disclaimer that this hook is able to handle layers such as N-dimensional convolutional, linear and embedding layers but there is no guarantee that this hook helps training.

chainer/functions/normalization/standardize.py
toslunar: > Do you mean remove this line?
> https://github.com/chainer/chainer/pull/6678/files/fa2567f0e9ba4eacb3a232b02a8083e3cb3801df#diff-75561f28993ff21bfbaad90b8b4ccb74R204

Yes.  Then could you do either of the following?

- rename `standardize` to `_standardize`
- implement it in `chainer/functions/normalization/_standardize.py`

toslunar:
Another design choice is
```tex
$(x - \mu) / (\sigma + \epsilon)$
```
vs
```tex
$(x - \mu) / \sqrt{\sigma^2 + \epsilon}$
```

The former is visually cleaner (when an article has a max-page limit), but I believe the latter is mathematically cleaner.  The authors used the former.

chainer/functions/normalization/standardize.py
hitsgub: Thank you! Sure.

chainer/link_hooks/weight_standardization.py
hitsgub: thank you!

hitsgub:
> Another design choice is
> 
> ```
> $(x - \mu) / (\sigma + \epsilon)$
> ```
> 
> vs
> 
> ```
> $(x - \mu) / \sqrt{\sigma^2 + \epsilon}$
> ```
> 
> The former is visually cleaner (when an article has a max-page limit), but I believe the latter is mathematically cleaner. The authors used the former.

Hmm, github code by authors also uses the former.
Next commit will follow the author, and it can return later if there is problem.
https://github.com/joe-siyuan-qiao/pytorch-classification/blob/e6355f829e85ac05a71b8889f4fff77b9ab95d0b/models/layers.py#L20

chainer/functions/normalization/_standardize.py
```diff
@@ -0,0 +1,100 @@
...
+        # xp: numpy, cupy, or chainer.functions
+        axes = tuple(six.moves.range(1, len(x.shape)))
+        mu = xp.mean(x, axis=axes, keepdims=True)
+        x_mu = x - mu
+        squ_x_mu = xp.square(x_mu)
+        var = xp.mean(squ_x_mu, axis=axes, keepdims=True)
+        if xp is numpy:
+            std = numpy.sqrt(var, dtype=x.dtype) + self.eps
+        elif xp is cuda.cupy:
+            std = cuda.cupyx.sqrt(var, dtype=x.dtype) + self.eps
```
crcrpar: `sqrt` is implemented in `cupy`, not `cupyx` (cupyx functions are not supported by numpy). I'm sorry to have you revise this section again and again.
```suggestion
            std = cuda.cupy.sqrt(var, dtype=x.dtype) + self.eps
```

crcrpar: `sqrt` is implemented in `cupy`, not `cupyx` (cupyx functions are not supported by numpy). I'm sorry to have you revise this section again and again.
```suggestion
            std = cuda.cupy.sqrt(var, dtype=x.dtype) + self.eps
```

chainer/link_hooks/weight_standardization.py
```diff
@@ -0,0 +1,75 @@
+from chainer.functions.normalization._standardize import _standardize
+from chainer import link_hook
+
+
+class WeightStandardization(link_hook.LinkHook):
+    """Weight Standardization link hook implementation.
+
+    This hook standardizes a weight by *weight statistics*.
+
+    This is a normalization technique that smooths the landscape like batch
```
crcrpar: Could you get this sentence clearer?
It is a bit hard for me to guess what kind of landscape does this technique smooth.

chainer/link_hooks/weight_standardization.py
```diff
@@ -0,0 +1,75 @@
...
+    normalization but without relying on large batch sizes during training.
+    Weight Standardization is originally proposed for 2D convolution layers
+    followed by mainly group normalization and sometimes batch normalization.
+    Noted that this hook is able to handle layers such as N-dimensional
```
crcrpar: nit-picking
```suggestion
    Note that this hook is able to handle layers such as N-dimensional
```

docs/source/reference/functions.rst
```diff
@@ -302,6 +302,7 @@ Normalization functions
    chainer.functions.layer_normalization
    chainer.functions.local_response_normalization
    chainer.functions.normalize
+   chainer.functions.standardize
```
crcrpar: If I understand https://github.com/chainer/chainer/pull/6678#discussion_r270714744 correctly, this would be either `chainer.functions._standardize` or removed.

hitsgub: ah! I've forgotten. thank you.

toslunar:
Computing the true gradient needs some efforts.  The implementation of `F.normalize` could be referred, if you'd like to write the code.
- #4190: fix for large eps
- #4769: fix for norm = 0

Other options I come up with are:
- Drop support for large eps
  - TODOs: Remove tests and add notes to the docstring
- Implement standardize with existing Chainer functions
  - TODOs: (It could be mentioned that the backprop fails if some std = 0.)
- Add eps to variance
  - TODOs: Explain the difference in the docstring.  The default value doesn't have to be `1e-5`.

hitsgub:
> Computing the true gradient needs some efforts. The implementation of `F.normalize` could be referred, if you'd like to write the code.
> 
> * #4190: fix for large eps
> * #4769: fix for norm = 0
> 
> Other options I come up with are:
> 
> * Drop support for large eps
>   
>   * TODOs: Remove tests and add notes to the docstring
> * Implement standardize with existing Chainer functions
>   
>   * TODOs: (It could be mentioned that the backprop fails if some std = 0.)
> * Add eps to variance
>   
>   * TODOs: Explain the difference in the docstring.  The default value doesn't have to be `1e-5`.

Thank you for your suggestion!
I have tried to fix for large eps and norm = 0, following `F.normalize`.

tests/chainer_tests/functions_tests/normalization_tests/test_standardize.py
```diff
@@ -1,15 +1,51 @@
...
+
+@testing.parameterize(*filter(
+    _is_good_param,
+    testing.product({
+        'ch_out': [1, 5],
+        'size': [10, 20],
+        'dtype': [numpy.float32, numpy.float16],
+    })
+    + testing.product([
+        # nonzeros (optional int): max number of nonzero elems in input
```
toslunar: Could you test with arrays many of whose elements are the same value, which is not necessarily zero?  For standardization, small std has to be tested, whereas `F.normalize` is tested with small norm.

hitsgub: OK, now I do understand.
For test of `standardization`, valid test cases is that *all* of the elements are same (or slightly different) value, don't we?

toslunar: The new test param, (maybe `nonconsts`) would run over
- `None`: corresponds to the generic cases
- `0`: corresponds to the cases where all of the elements are the same
- `2`: corresponds to in-between cases

chainer/link_hooks/weight_standardization.py
```diff
@@ -0,0 +1,81 @@
...
+from chainer import link_hook
+
+
+class WeightStandardization(link_hook.LinkHook):
+    """Weight Standardization link hook implementation.
+
+    This hook standardizes a weight by *weight statistics*.
+
+    This link hook implements a "weight standardization" which computes the
+    mean and variance along axis "output channels", then normalize by these
```
crcrpar:
```suggestion
    mean and variance along axis "output channels", then normalizes by these
```

chainer/link_hooks/weight_standardization.py
```diff
@@ -0,0 +1,81 @@
...
+    statistics.
+    That improves training by reducing the Lipschitz constants of the loss and
+    the gradients like batch normalization but without relying on large batch
+    sizes during training. Specifically, in micro-batch training, weight
+    standardization significantly outperforms other normalization methods like
```
crcrpar: Isn't Group Normalization necessary for Weight Standardization to overperform over BN?

hitsgub: Oh, thank you, it has gone out while changing sentences sometimes.

crcrpar: Thanks for update.

tests/chainer_tests/functions_tests/normalization_tests/test_standardize.py
```diff
@@ -0,0 +1,111 @@
...
+        'chainerx_device': ['native:0', 'cuda:0'],
+    })
+)
+class TestStandardize(testing.FunctionTestCase):
+
+    def setUp(self):
+        self.skip_double_backward_test = self.same in ('equal', 'near')
+        if self.same == 'equal':
+            # Make self.x have same values
+            self.x[...] = self.x[0]
```
crcrpar: When is the attribute of ‘x’ defined and used?

crcrpar: I guess this processing should be done in `generate_inputs` method
and this test currently uses inputs generated randomly.

chainer/functions/normalization/_standardize.py
```diff
@@ -0,0 +1,123 @@
...
+        xp = backend.get_array_module(*inputs)
+        x, = inputs
+        x_mu, std_noeps, std, x_hat = self._compute(xp, x)
+        return x_hat,
+
+    def backward(self, indexes, grad_outputs):
+        F = chainer.functions
+        x, = self.get_retained_inputs()
+        gy, = grad_outputs
+        axes = tuple(six.moves.range(1, len(x.shape)))
```
crcrpar: `x.ndim` is faster.
```suggestion
        axes = tuple(six.moves.range(1, x.ndim))
```

tests/chainer_tests/link_hooks_tests/test_weight_standardization.py
```diff
@@ -0,0 +1,185 @@
...
+        hook = WeightStandardization(weight_name=wrong_Weight_name)
+        with pytest.raises(ValueError):
+            self.layer.add_hook(hook)
+
+    def test_raises(self):
+        with pytest.raises(NotImplementedError):
+            with WeightStandardization():
+                self.layer(self.x)
+
+    def test_invalid_shaped_weight(self):
```
crcrpar: Does WeightStandardization hook have some limitations of the shape of a weight?

hitsgub: There is no limitations. Thanks.

crcrpar:
Is https://github.com/chainer/chainer/pull/6678/commits/c3deacf23044408639b6fc7e9f5c97a193184975 for performance enhancement?

hitsgub:
Something wrong in my differentiation implementation even if `same = 'no'` (non-zero variance)...
Someone could you review in this differentiation?
(Also I continue to check implementation.)

crcrpar:
The tolerance for fp16 avoided the backward tests. 🤔 

ref: https://github.com/hitsgub/chainer/compare/weight_standardization...crcrpar:modify_ws?expand=1

hitsgub:
> The tolerance for fp16 avoided the backward tests failures of fp32&64🤔
> 
> ref: https://github.com/hitsgub/chainer/compare/weight_standardization...crcrpar:modify_ws?expand=1

Thank you @crcrpar !

hitsgub:
Following other normalization functions, double backward implementation of `~chainer.functions._standardize` is temporarily suspended.

crcrpar:
@toslunar Could you review this?

chainer/link_hooks/weight_standardization.py
```diff
@@ -0,0 +1,80 @@
...
+            The default value is 1e-5.
+        weight_name (str): Link's weight name to appky this hook. The default
+            value is ``'W'``.
+        name (str or None): Name of this hook. The default value is
+            ``'WeightStandardization'``.
+    """
+
+    name = 'WeightStandardization'
+
+    def __init__(self, eps=1e-6, weight_name='W', name=None):
```
toslunar: Could you follow the authors' default value?
```suggestion
    def __init__(self, eps=1e-5, weight_name='W', name=None):
```

chainer/functions/normalization/_standardize.py
```diff
@@ -0,0 +1,109 @@
...
+        g_x_mu_2 = g_var * x_mu
+        g_x_1 = (g_x_mu_1 - g_x_mu_2) / (std ** 2)
+        g_mu = F.mean(g_x_1, axis=axes, keepdims=True)
+
+        return g_x_1 - g_mu,
+
+
+def _standardize(x, eps=1e-6):
+    """Standardize for `Weight standardization
+    <https://arxiv.org/abs/1903.10520>`_.
```
toslunar: Could you make the summary fit in a line?

chainer/link_hooks/weight_standardization.py
```diff
@@ -0,0 +1,80 @@
+from chainer.functions.normalization._standardize import _standardize
```
toslunar: Could you import a module, not a function?
```suggestion
from chainer.functions.normalization import _standardize
```
and use `_standardize._standardize` later.

toslunar:
Jenkins, test this please.

tests/chainer_tests/functions_tests/normalization_tests/test_standardize.py
```diff
@@ -0,0 +1,106 @@
...
+
+    def generate_inputs(self):
+        shape = self.ch_out, self.size
+        if self.same in ('equal', 'near'):
+            # Make self.x have same values
+            x = numpy.ones(shape, self.dtype)
+            x *= numpy.random.uniform(-1, 1)
+            if self.same == 'near':
+                # Make self.x have slightly different values
+                zero_scale = 10. ** numpy.random.randint(-40, -3)
```
toslunar: The scale is too small to add to `x`.

chainer/functions/normalization/_standardize.py
hitsgub: If it fits on one line, it was 85 characters.

crcrpar: I think it is OK if the summary is "Weight standardization function" like `F.batch_noramlization` since you have written the paper and its link at the end of this docstring.

tests/chainer_tests/functions_tests/normalization_tests/test_standardize.py
hitsgub: Changed the way to make an array that has small variance (according to the number of significant digits of float64).

chainer/functions/normalization/_standardize.py
hitsgub: Thanks! I fixed it.

toslunar:
Jenkins, test this please.

toslunar:
Jenkins, test this please.

toslunar:
CIs, test this please.

toslunar:
CIs, test this please.

toslunar:
Jenkins, test this please.

PR #6679, opened 2019-03-28
Fix conflicts between #6432 and #6486

toslunar:

PR #6683, opened 2019-03-28
fix test input

kshitij12345:
Probably a missing case of refactoring the code.

### DIFF
--- a/tests/chainerx_tests/unit_tests/routines_tests/test_math.py
+++ b/tests/chainerx_tests/unit_tests/routines_tests/test_math.py
@@ -1500,14 +1500,14 @@ def test_log_softmax_invalid(device, a_shape, axis, dtype):
     # Special shapes
     chainer.testing.product({
         'shape': [(), (0,), (1,), (2, 0, 3), (1, 1, 1), (2, 3)],
-        'dtype': _expected_float_dtypes_math_functions,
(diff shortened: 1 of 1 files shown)


tests/chainerx_tests/unit_tests/routines_tests/test_math.py
```diff
@@ -1500,14 +1500,14 @@ def test_log_softmax_invalid(device, a_shape, axis, dtype):
     # Special shapes
     chainer.testing.product({
         'shape': [(), (0,), (1,), (2, 0, 3), (1, 1, 1), (2, 3)],
-        'dtype': _expected_float_dtypes_math_functions,
+        'in_dtypes,out_dtype': _in_out_float_dtypes_math_functions,
```
kshitij12345: Should be `_in_out_dtypes_math_functions` based on other tests, however there is a type mismatch between `numpy` and `chainerx` So have kept it as `_in_out_float_dtypes_math_functions` as was the case before this diff.

toslunar: - Thanks for explaining the remaining issue.
- The PR seems a dup of #6679.

kshitij12345: Oh my bad. Did not notice it. Closing this one.

PR #5094, opened 2018-07-12
Use `F.cumprod` in backward of `F.prod`

toslunar:
Merge #5074 first.

PR #3031, opened 2017-07-17
Implement prod function

ronekko:
I implemented `F.prod` based on [`F.sum`](https://github.com/chainer/chainer/blob/d62be84d0fa4375d0fe18e5e6d40bc02c1847e5e/chainer/functions/math/sum.py).

PR #6687, opened 2019-03-28
[chainerx] add `prod`

kshitij12345:
This PR intends to add Prod to ChainerX.

It is implemented with reference of Sum.

Done :
* Native Device Implementation.
* Cuda Device Implementation.
* Backward Function.
* Python bindings for `chx.prod`.
* Python bindings for `array.prod`.
* Tests

**TO-DO**
~~Update Documentation~~


kshitij12345:
8 tests for compute_prod fail. All are the cases where `dtype` is `float16` and `device` is `cuda`.
image


Numpy returns `0.0` while ChainerX returns `nan`.
The inputs for the failed cases are [here](https://github.com/chainer/chainer/blob/fb35b515ef05d58c79a1c390e1b7f716a199f71e/tests/chainerx_tests/unit_tests/routines_tests/test_math.py#L1277-L1278).

From my guess, the hotspots should maybe [here](https://github.com/chainer/chainer/blob/fb35b515ef05d58c79a1c390e1b7f716a199f71e/chainerx_cc/chainerx/cuda/cuda_device/reduction.cu#L91) or [here](https://github.com/chainer/chainer/blob/c5b075ba46c285c0bcf7466d77c6abda5c158617/chainerx_cc/chainerx/cuda/float16.cuh#L55)

Note : All other types for the same inputs pass.

Please Help.

kshitij12345:
@niboshi , could you help me figure why double_backward is failing but not backward? I am not sure which part of backward is wrong exactly. Thank You.

chainerx_cc/chainerx/routines/math.cc
```diff
@@ -477,6 +477,50 @@ Array Sum(const Array& a, const OptionalAxes& axis, bool keepdims) {
...
+
+    Array out = internal::EmptyReduced(a.shape(), out_dtype, sorted_axis, keepdims, a.device());
+    {
+        NoBackpropModeScope scope{};
+        a.device().backend().CallKernel<ProdKernel>(a, sorted_axis, out);
+    }
+
+    BackwardBuilder bb{"prod", a, out};
+    if (BackwardBuilder::Target bt = bb.CreateTarget(0)) {
+        bt.Define([sorted_axis, a, out, in_shape = a.shape(), keepdims](BackwardContext& bctx) {
```
niboshi: Not sure if it's relevant, but I think `a` and `out` should be captured using `bb.RetainInput(0)` and `bb.RetainOutput(0)` respectively?

chainerx_cc/chainerx/native/native_device/reduction.cc
```diff
@@ -76,6 +76,31 @@ class NativeSumKernel : public SumKernel {
...
+        a.device().CheckDevicesCompatible(a, out);
+
+        auto do_prod = [&a, &axis, &out](auto in_pt, auto out_pt) {
+            using In = typename decltype(in_pt)::type;
+            using Out = typename decltype(out_pt)::type;
+            using Accum = std::conditional_t<std::is_same<Out, Float16>{}, float, Out>;
+            struct Impl {
+                Accum Identity() { return Accum{1}; }
+                Accum MapIn(In in, int64_t /*index*/) { return static_cast<Accum>(in); }
+                void Reduce(Accum next, Accum& accum) { accum *= next; }
```
niboshi: (This is not relevant to the error)
It won't compile in gcc, complaining `*=` is invalid for `bool`s.

As `Accum` (and `Out`) is never actually `bool`, let's replace `VisitDtype` (the first one) at the line 98 with `VisitNumericDtype`, so that compilers won't try to generate the code for `bool`.

https://github.com/chainer/chainer/pull/6687/files#diff-5cc617697b5522640bcf9858d0c42b73R98

chainerx_cc/chainerx/routines/math.cc
kshitij12345: Done.

chainerx_cc/chainerx/native/native_device/reduction.cc
kshitij12345: Done.

kshitij12345:
@niboshi , after merging the latest master and  addressing the above comments, the `double_backward` issue is resolved except for the  few random `float16` cases.

The `double_backward` for `float16` fails spurriously many times. Please review.

chainerx_cc/chainerx/routines/math.cc
```diff
@@ -513,6 +513,53 @@ Array Sum(const Array& a, const OptionalAxes& axis, bool keepdims) {
...
+            CHAINERX_ASSERT(std::is_sorted(sorted_axis.begin(), sorted_axis.end()));
+
+            if (!(in_shape.ndim() == 0 || sorted_axis.empty() || keepdims)) {
+                Shape out_shape_broadcastable = gout.shape();
+                for (auto axis : sorted_axis) {
+                    out_shape_broadcastable.insert(out_shape_broadcastable.begin() + axis, 1);
+                }
+                bctx.input_grad() = (out.Reshape(out_shape_broadcastable) * gout.Reshape(out_shape_broadcastable) / a);
+            } else {
+                bctx.input_grad() = out * gout / a;
```
toslunar: Could you avoid `0 / 0`?  There are at least two ways (#3031, #5094) to implement backward that allows forward inputs to contain `0`.  It might be fine to leave `backward` of `chainerx.prod` totally unimplemented.

kshitij12345: Sure. Do note that it'll take me some time to resume work on this PR.

niboshi:
We're no longer merging feature PRs.
Let me close the PR.

PR #6696, opened 2019-03-29
Support Tagged communication with MPI_Status.

y1r:
This PR aims to support Tagged communication of send_obj and recv_obj on `MpiCommunicatorBase`.
MPI_Status support is also included.

Please check https://github.com/mpi4py/mpi4py/blob/master/src/mpi4py/MPI/Comm.pyx#L1168 .
Default behavior of mpi4py's recv without tag is interpreted as `ANY_TAG`, therefore I followed it.


keisukefukuda:
Jenkins, test this please

PR #6699, opened 2019-03-30
Fix ChainerX optimzer fallback for non-default devices

hvy:
This PR fixes optimizer fallback from ChainerX with non-default devices. The issue was that the current device was never set when falling back meaning that running MNIST with a non-default device e.g. `--device='cuda:2'` lead to an array-current-device mismatch error.


asi1024:
Jenkins, test this please.

asi1024:
Test failures seem unrelated. LGTM.

PR #6709, opened 2019-04-01
Properly handle FP16 in `D.Normal`

takagi:
Rel. 6168.

cuRAND supports only FP32 and FP64. This PR fixes `D.Normal`, which uses `cupy.random`, to handle FP16 properly to sample.


toslunar:
Jenkins, test this please.

PR #6715, opened 2019-04-01
[chainerx]Implementation of Absolute

dido1998:
ref issue: #6423


chainerx_cc/chainerx/python/routines.cc
```diff
@@ -608,6 +608,7 @@ void InitChainerxMath(pybind11::module& m) {
     m.def("tanh", [](const ArrayBodyPtr& x) { return MoveArrayBody(Tanh(Array{x})); }, py::arg("x"));
     m.def("sin", [](const ArrayBodyPtr& x) { return MoveArrayBody(Sin(Array{x})); }, py::arg("x"));
     m.def("cos", [](const ArrayBodyPtr& x) { return MoveArrayBody(Cos(Array{x})); }, py::arg("x"));
+    m.def("abs", [](const ArrayBodyPtr& x) { return MoveArrayBody(Absolute(Array{x})); }, py::arg("x"));
```
hvy: I would prefer to align with Chainer in this case (and your C++ implementation), since for NumPy `np.abs is np.absolute`. What do you say?
```suggestion
    m.def("absolute", [](const ArrayBodyPtr& x) { return MoveArrayBody(Absolute(Array{x})); }, py::arg("x"));
```

chainerx_cc/chainerx/routines/math.cc
```diff
@@ -770,6 +770,17 @@ Array Cos(const Array& x) {
     return out;
 }
 
+Array Absolute(const Array& x) {
+    Dtype dtype = GetMathResultDtype(x.dtype());
+    Array out = Empty(x.shape(), dtype, x.device());
```
hvy: You're allocating an array here which is discarded when being assigned the sum of `x_flip_1` and `x_flip_2`.

hvy:
Thanks for the PR as always. I added some minor comments. Could you take a look?

chainerx_cc/chainerx/python/routines.cc
hvy: It might be fine also to alias. `m.attr("abs") = m.attr("absolute");`.

tests/chainerx_tests/unit_tests/routines_tests/test_math.py
```diff
@@ -1630,7 +1630,7 @@ def func(self, xp, a):
     # Special shapes
     chainer.testing.product({
         'shape': [(), (0,), (1,), (2, 0, 3), (1, 1, 1), (2, 3)],
-        'in_dtypes,out_dtype': _in_out_dtypes_math_functions,
+        'in_dtypes,out_dtype': _in_out_float_dtypes_math_functions,
```
hvy: Can you revert this change? It seems unrelated to your PR.

hvy:
Can you write a Python test that checks that `abs` and `absolute` are equal, i.e. `assert chainerx.abs is chainerx.absolute`?

dido1998:
yes, I will do this.

dido1998:
I have added the assertion test. But, I was not able to reproduce the python test errors (dtype mismatch) locally.

dido1998:
I will check it again.

dido1998:
@hvy, I have fixed all the errors. Could you have a look?

hvy:
Jenkins, test this please.

hvy:
Jenkins, test this please.

dido1998:
The failure seems to be in TestMaxPool

hvy:
Indeed, the failures seem unrelated although there were a couple of them except max pool. Thanks for the long running PR, LGTM!

PR #6732, opened 2019-04-03
Fix: documentation for using cmake to build Python and C++ for ChainerX

sierreis:
The path to the CMakeLists.txt moved relative to the top level build directory I think, so it should now be `../chainerx_cc` instead of `..`, that's how I managed to get it to work.

### DIFF
--- a/docs/source/chainerx/contribution.rst
+++ b/docs/source/chainerx/contribution.rst
@@ -112,14 +112,14 @@ Test coverage
 -------------
 
 We use `gcov <https://gcc.gnu.org/onlinedocs/gcc/Gcov.html>`_ to the measure C++ code coverage.
-Build the Python package in ``Debug`` mode, and build C++ test suite as:
+From the root of the project, build the Python package in ``Debug`` mode, and build C++ test suite as:
 
 .. code-block:: console
 
     $ python setup.py build --debug --build-temp ./build --build-lib ./build develop
     $ mkdir -p build
     $ cd build
-    $ cmake -DCMAKE_BUILD_TYPE=Debug -DCHAINERX_BUILD_PYTHON=1 -DCHAINERX_ENABLE_COVERAGE ..
+    $ cmake -DCMAKE_BUILD_TYPE=Debug -DCHAINERX_BUILD_PYTHON=1 -DCHAINERX_ENABLE_COVERAGE ../chainerx_cc
     $ make
 
 Run both the Python and the C++ test suite:
--- a/docs/source/chainerx/contribution.rst
+++ b/docs/source/chainerx/contribution.rst
@@ -119,7 +119,7 @@ From the root of the project, build the Python package in ``Debug`` mode, and bu
     $ python setup.py build --debug --build-temp ./build --build-lib ./build develop
     $ mkdir -p build
     $ cd build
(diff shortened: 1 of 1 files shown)


docs/source/chainerx/contribution.rst
```diff
@@ -112,14 +112,14 @@ Test coverage
...
-Build the Python package in ``Debug`` mode, and build C++ test suite as:
+From the root of the project, build the Python package in ``Debug`` mode, and build C++ test suite as:
 
 .. code-block:: console
 
     $ python setup.py build --debug --build-temp ./build --build-lib ./build develop
     $ mkdir -p build
     $ cd build
-    $ cmake -DCMAKE_BUILD_TYPE=Debug -DCHAINERX_BUILD_PYTHON=1 -DCHAINERX_ENABLE_COVERAGE ..
+    $ cmake -DCMAKE_BUILD_TYPE=Debug -DCHAINERX_BUILD_PYTHON=1 -DCHAINERX_ENABLE_COVERAGE ../chainerx_cc
```
hvy: This is an error that already existed before this PR but I think `-DCHAINERX_ENABLE_COVERAGE` needs a value. Could you fix that too? If not, we can address it in a separate PR.
```suggestion
    $ cmake -DCMAKE_BUILD_TYPE=Debug -DCHAINERX_BUILD_PYTHON=1 -DCHAINERX_ENABLE_COVERAGE=1 ../chainerx_cc
```

hvy:
Thanks for the PR. It indeed needs to be fixed. It basically LGTM but I added a smaller comment. Could you take a look?

docs/source/chainerx/contribution.rst
sierreis: Sure, I'll update that line and ensure there aren't any issues with the build and coverage checks.

sierreis:
One thing, that I ran into is that I needed to explicitly set the version of `gcc` and `g++`, since the system default is a higher version the build warning might differ resulting in build failures on Travis. Do you think the documentation should also include `CC=gcc-5 CXX=g++-5` prior to the `cmake` command?

Also, I'm having some issues running pytest. I'm not sure why but when I run:

`python setup.py build --debug --build-temp ./build --build-lib ./build develop`

Python expects the `_core.so` library to be located at `/chainerx/_core.so`. The build, however, creates `/build/chainerx/_core.so` which doesn't seem to be found by default by Python. If I copy `_core.so` to `/chainerx/_core.so` then everything seems to work. Do you know why this is the case?

sierreis:
The build failure in this instance is for macos and has to do with one of the installation dependencies missing. The error message given by Travis is:

`ERROR: The Python ssl extension was not compiled. Missing the OpenSSL lib?`

I'll try and rerun the build again to check if this has been resolved.

hvy:
Could you try with the latest master to see if the library path problem persists?

hvy:
I'd like to merge this PR more or less as it is right now but would prefer having the `Force rebuild` commit reverted. Could you do that?

hvy:
Jenkins, test this please.

hvy:
We discussed this issue offline with the team and for now concluded to remove this coverage section entirely https://github.com/chainer/chainer/pull/7462. Sorry for the long running PR going back and forth but please let me close this PR. If you still have issues, of course, please feel to open another ticket. You probably know this but there's also a separate section for how to install ChainerX https://github.com/chainer/chainer/blob/master/docs/source/chainerx/install/index.rst

PR #6442, opened 2019-03-06
Change intermediate dtype of `Adam` for float16 parameters to float32

niboshi:
Fixes #6299

With this fix, `Adam` no longer complains the default `eps` to be too small for float16 parameters.

Similar fix to #6366

PR #6738, opened 2019-04-03
Use larger eps in mnist example

takagi:
Rel. #6168.

This PR fixes the mnist example to use larger eps for `Adam` optimizer to make it work in FP16 mode as well.

### DIFF
--- a/examples/mnist/train_mnist.py
+++ b/examples/mnist/train_mnist.py
@@ -69,7 +69,9 @@ def main():
     device.use()
 
     # Setup an optimizer
-    optimizer = chainer.optimizers.Adam()
+    # Here used a larger eps in case of FP16 mode, the default value is enough
+    # for FP32 mode.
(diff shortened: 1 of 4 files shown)


niboshi:
I think this problem has been solved in #6442, hasn't it?

takagi:
It seems that it is still unstable with `CHAINER_DTYPE=float16` and the default Adam eps=`1e-8`. Would you try it too?

takagi:
Although I posted this PR, I wonder which Chainer function causes `Nan`, maybe not Adam.

takagi:
`RuntimeError: NaN is detected on forward computation of SoftmaxCrossEntropy`

takagi:
I found that `Linear` leads to infinity, which causes `NaN` in `SoftmaxCrossEntropy `. I close this PR because it seems not an essential fix.

Issue #6661, opened 2019-03-27
In examples, use device spec instead of gpu id


toslunar:
Examples with `--gpu` does not support chainerx and intel64 backends.
Add `--device` option to the examples and use `chainer.get_device` to support all the backends.

See [examples/mnist/train_mnist.py](https://github.com/chainer/chainer/blob/master/examples/mnist/train_mnist.py) for an example with `--device`.

hvy:
For the record, all examples are fixed except for the ones that are located under `examples//chainermn`.

hvy:
@toslunar Are ChainerMN examples within the scope of this PR?

toslunar:
> Are ChainerMN examples within the scope of this PR?

No.

PR #6739, opened 2019-04-03
Support ChainerX in VAE example

niboshi:
Part of #6661


niboshi:
It doesn't work with ChainerX CUDA device >=1.
I'm investigating.

takagi:
LGTM! I tested it on my local and it workd.

Issue #6692, opened 2019-03-29
Rewrite device functions using dynamic op registration


niboshi:
* Rewrite existing op implementations with dynamic op registry #6675.
* Remove the corresponding `Device` virtual functions.
.
* [x] connection #6833
* [x] creation https://github.com/chainer/chainer/pull/6745
* [x] indexing #6718 
* [x] linalg #6746
* [x] logic #6727 
* [x] manipulation https://github.com/chainer/chainer/projects/2#card-20071153
* [x] math
  * [x] arithmetic #6723 
  * [x] trigonometric #6729 
  * [x] other #6842
* [x] normalization https://github.com/chainer/chainer/pull/6719
* [x] pooling https://github.com/chainer/chainer/pull/6800
* [x] sorting #6675
* [x] statistics https://github.com/chainer/chainer/projects/2#card-20071812
* [x] `Array` member functions #6834

niboshi:
TODO: after finishing above tasks, we should review them once again. We introduced some additional convension after merging some of them.

niboshi:
Also, we should rename macros:
* `CHAINERX_REGISTER_OP_CUDA` -> `CHAINERX_CUDA_REGISTER_OP`
* `CHAINERX_REGISTER_OP_NATIVE` -> `CHAINERX_NATIVE_REGISTER_OP`

hvy:
Renamed in https://github.com/chainer/chainer/pull/6865.

hvy:
Renamed "op" to "kernel" in https://github.com/chainer/chainer/pull/6944.

niboshi:
Seems like this issue has been resolved.

PR #6745, opened 2019-04-04
Op registration: creation

hvy:
Part of #6692.

Note that `Device::AsType` and `Device::Fill` are not covered in this PR. Although similar to other device methods covered in this PR, these methods do not have corresponding routines and are instead exposed as `Array` methods `Array::AsType`, `Array::Fill`.


niboshi:
Jenkins, test this please

niboshi:
LGTM

PR #6754, opened 2019-04-04
Fix Travis with macOS

toslunar:
The PR fixes Travis fails to install python (2.7.10 and 3.5.1) using pyenv.  I reproduced the same error on my MacBook by
```
brew install openssl@1.1
pyenv install 2.7.10  # fail
brew uninstall openssl@1.1
pyenv install 2.7.10  # pass
```

Updating `openssl@1.1` did not fix the problem.

The update of pyenv, which includes pyenv/pyenv#1302, probably caused the error, while I've not figured out why openssl 1.1 makes the installation fail.

### DIFF
--- a/scripts/ci/travis/steps.sh
+++ b/scripts/ci/travis/steps.sh
@@ -51,7 +51,7 @@ step_before_install_chainer_test() {
     # Remove oclint as it conflicts with GCC (indirect dependency of hdf5)
     if [[ $TRAVIS_OS_NAME = "osx" ]]; then
         brew update >/dev/null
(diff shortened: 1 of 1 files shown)


hvy:
Jenkins, test this please.

hvy:
It seems like it is addressed in the official documentation. https://github.com/pyenv/pyenv/wiki/Common-build-problems#error-the-python-ssl-extension-was-not-compiled-missing-the-openssl-lib

hvy:
Jenkins, test this please.

hvy:
Let me merge this fix as it seems to work. LGTM!

PR #6762, opened 2019-04-05
Allow snapshot_object have condition and writer option

kuenishi:
This commit also aligns behaviour of `snapshot_object` to `snapshot` for future potential extensions of both functions.

### DIFF
--- a/chainer/training/extensions/_snapshot.py
+++ b/chainer/training/extensions/_snapshot.py
@@ -4,8 +4,11 @@ from chainer.training.extensions import snapshot_writers
 from chainer.utils import argument
 
 
-def snapshot_object(target, filename, savefun=npz.save_npz, **kwargs):
-    """Returns a trainer extension to take snapshots of a given object.
+def snapshot_object(target, filename, **kwargs):
+    """snapshot_object(target, filename, savefun=None, \
+*, condition=None, writer=None, snapshot_on_error=False)
+
+    Returns a trainer extension to take snapshots of a given object.
 
     This extension serializes the given object and saves it to the output
     directory.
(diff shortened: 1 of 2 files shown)


chainer/training/extensions/_snapshot.py
```diff
@@ -4,7 +4,7 @@
 from chainer.utils import argument
 
 
-def snapshot_object(target, filename, savefun=npz.save_npz, **kwargs):
+def snapshot_object(target, filename, **kwargs):
```
crcrpar: To make the documentation more readable, could you change the docstring as follows?

```python
def snapshot_object(target, filename, **kwargs):
    """snapshot_object(target, filename, *, savefun=npz.save_npz, \
condition=None, writer=None, snapshot_on_error=False)

    Returns a trainer extension to take snapshots of a given object.

    ...
```

niboshi:
Made priority high.

chainer/training/extensions/_snapshot.py
crcrpar: Done

chainer/training/extensions/_snapshot.py
```diff
@@ -4,8 +4,11 @@
 from chainer.utils import argument
 
 
-def snapshot_object(target, filename, savefun=npz.save_npz, **kwargs):
-    """Returns a trainer extension to take snapshots of a given object.
+def snapshot_object(target, filename, **kwargs):
```
toslunar: Are you sure to make a no-compat change by forcing an argument to be a keyword?  The function is publicly available as `chainer.training.extensions.snapshot_object`, while this file `chainer/training/extensions/_snapshot.py` looks private.

kuenishi: Thanks for good catch. I knew it was public function and caring default behaviour but less careful about custom savefun case. I added another commit, which might seem strange that the default value of `savefun` being `None`, but it preserves previous behaviour when either `None` or some function is passed.

toslunar:
Jenkins, test this please.

PR #6767, opened 2019-04-05
 Append local version if not released 

ktns:
Hi, I've wrote a patch to append the local version to the version string if chainer is installed from a git repository.

Why I wrote this patch is to ensure the (pre-)released version is properly installed even after installing the later development tree which might have the same version as the (pre-)released one from a repository.
(e.g. As of writing this, `pip install git+https://github.com/chainer/chainer@master && pip install chainer==6.0.0rc1` does not install the pre-released version because the latest master also has `6.0.0rc1` as its version and `pip` conclude that the requirement is already met).


chainer/_version.py
```diff
@@ -1,5 +1,86 @@
...
+# A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+# OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+# SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+# LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+_released_version = '6.0.0rc1'
```
toslunar: Could you move `_released_version` and `_optional_dependencies` to the top of the file?  They are constants to be edited.  The NumPy copyright notice does not have to be at the beginning of the file.

tests/chainer_tests/test_version.py
```diff
@@ -0,0 +1,10 @@
+import unittest
+import packaging.version
+
+import chainer
+
+
+class TestVersion(unittest.TestCase):
+    def test_pep440(self):
+        version = packaging.version.parse(chainer.__version__)
+        self.assertIsInstance(version, packaging.version.Version)
```
toslunar: Chainer has a rule to make each test file runnable.
```suggestion
        self.assertIsInstance(version, packaging.version.Version)


testing.run_module(__name__, __file__)
```

tests/chainer_tests/test_version.py
```diff
@@ -0,0 +1,10 @@
+import unittest
+import packaging.version
```
toslunar: Could you group imports?  `packaging` is not a built-in module.
```suggestion

import packaging.version
```

tests/chainer_tests/test_version.py
```diff
@@ -0,0 +1,14 @@
+import unittest
+
+import packaging.version
+
+import chainer
+
```
toslunar:
```suggestion
from chainer import testing

```

toslunar:
Jenkins, test this please.

toslunar:
It seems chainer-test needs to be fixed first.  Please wait.

PR #6768, opened 2019-04-05
Fix `Evaluator` for `chainer.dataset.converter`

niboshi:
`Evaluator` was not aware of `chainer.dataset.converter` (#5832).


asi1024:
Jenkins, test this please.

asi1024:
LGTM.

Issue #6608, opened 2019-03-24
Simple way to add Basic Ops in ChainerX


kshitij12345:
While working on the PR to add `sin` and `cos` functionality. I found that most of the code was same except for a verysmall part.

Also going through the PR #6590 , it seems that a lot of same code needs to be added/modified at various places.

So I tried to do it with a simple Macro based on Tanh implementation.
```C++
// Can be improved !!??
#define UnaryOp_Cuda(func, func_def)                                                \
    namespace {                                                                     \
    template <typename T>                                                           \
    struct func##Impl {                                                             \
        using CudaType = cuda_internal::DataType<T>;                                \
        __device__ func_def                                                         \
    };                                                                              \
    }                                                                               \
                                                                                    \
    void CudaDevice::func(const Array& x, const Array& out) {                       \
        CheckDevicesCompatible(x, out);                                             \
        CudaSetDeviceScope scope{index()};                                          \
        const Array& x_cast = x.dtype() == out.dtype() ? x : x.AsType(out.dtype()); \
        VisitFloatingPointDtype(out.dtype(), [&](auto pt) {                         \
            using T = typename decltype(pt)::type;                                  \
            Elementwise<const T, T>(func##Impl<T>{}, x_cast, out);                  \
        });                                                                         \
    }
```
which allows to define a new Operator simply by
```C++
UnaryOp_Cuda(Sqrt, void operator()(int64_t /*i*/, CudaType x, CudaType& out) { out = cuda::Sqrt(x); })
```
From a glance it can also be extended for many standard Binary Operators.

I have tried this with macro for native and converted similar Unary Operators and it built successfully. All changes can be found [here](https://github.com/kshitij12345/chainer/tree/unary_op_macro)

Also it seems important to have simple code generation script to facilitate these simple cases.
```yaml
- name : Sqrt
- operator_type : Unary
- transform : out = Sqrt(in)

- name : Cube
- operator_type : Unary
- transform : out = (in * in * in)
```

The above yaml example could be easily parsed to generate Header file and Implementation for Standard Operation for Cuda and Native or any other backend( Unary , Binary ). It will also facilitate making bulk changes with ease.

Would love to know your views on this.
Thank You.

niboshi:
Thank you for suggestion!

It looks useful as there are many unary ops.

We will need versions for both float-point dtypes and all dtypes (the latter for ops like `Negative`).
How about something like the following?

```c++
[...]

PR #6776, opened 2019-04-05
[chainerx] add `a/sinh`,`a/cosh`

kshitij12345:
Add `sinh, cosh, arcsinh, arccosh`.

Reference : #6423


chainerx_cc/chainerx/cuda/cuda_device/hyperbolic.cu
```diff
@@ -0,0 +1,103 @@
...
+namespace {
+
+template <typename T>
+struct CudaUnaryOp {
+    T (*func)(T);
+
+    CudaUnaryOp(T (*func)(T)) : func{func} {}
+
+    __device__ void operator()(int64_t /*i*/, CudaType x, CudaType& out) { out = func(x); }
+};
```
kshitij12345: The above part is slightly different from usual please review.

chainerx_cc/chainerx/native/native_device/hyperbolic.cc
```diff
@@ -0,0 +1,90 @@
...
+namespace {
+
+template <typename T>
+struct NativeUnaryOp {
+    T (*func)(T);
+
+    NativeUnaryOp(T (*func)(T)) : func{func} {}
+
+    void operator()(int64_t /*i*/, T x, T& out) { out = func(x); }
+};
```
kshitij12345: The above part is slightly different from usual. Please review

chainerx_cc/chainerx/cuda/cuda_device/hyperbolic.cu
kshitij12345: If this looks ok, this struct can be used for other UnaryOps instead of define their own Impl

chainerx_cc/chainerx/cuda/CMakeLists.txt
```diff
@@ -32,6 +32,7 @@ CUDA_ADD_LIBRARY(chainerx_cuda STATIC
     cuda_device/dot.cu
     cuda_device/exp_log.cu
     cuda_device/fill.cu
+    cuda_device/hyperbolic.cu
```
asi1024: There might be no need to split files with `trigonometric.cu`. How about merging them?

chainerx_cc/chainerx/native/native_device/hyperbolic.cc
asi1024: It looks not good to me because it seems to worsen performance because the inline expansion will not work well.

chainerx_cc/chainerx/cuda/cuda_device/hyperbolic.cu
asi1024: It looks good to me.

asi1024: This struct may be dropped when a simpler way is introduced (like the issue you suggested: #6608) in the near future.

asi1024:
Could you resolve conflicts?

chainerx_cc/chainerx/cuda/cuda_device/hyperbolic.cu
kshitij12345: Will keep it as it is for now.

chainerx_cc/chainerx/native/native_device/hyperbolic.cc
kshitij12345: Will revert it to the standard approach. Or should we make `struct` for `numeric ops` as done for Cuda

chainerx_cc/chainerx/cuda/CMakeLists.txt
kshitij12345: Oh will merge them with `trignometric`

kshitij12345:
Yeah will resolve the conflicts and merge `hyperbolic` with `trignometric`.
Also will revert to standard way for `NativeImpl`

kshitij12345:
@asi1024 Have merged `hyperbolic` to `trignometric`.

Also, reverted to standard `NativeImpl` style.

Actually now, I am curious whether you have checked that the functions are getting inlined in assembly as I have heard that `inline` is just a hint for the compiler and it may choose to skip it.

Thank You.

asi1024:
Yes, `inline` is just a hint for the compiler, but my comment didn't mean that.
If `UnaryOpImpl` takes a function as `void*` type, The function which should be called is determined at runtime.

asi1024:
Jenkins, test this please.

kshitij12345:
@asi1024 The compilation is failing, will look into it in the evening.



> Yes, `inline` is just a hint for the compiler, but my comment didn't mean that.
> If `UnaryOpImpl` takes a function as `void*` type, The function which should be called is determined at runtime.

Umm, I didn't exactly understand the second bit. Also, isn't the function type, `T(*)(T)`, sorry if my doubt seems to amateur. 
Thank You

asi1024:
Could you resolve conflicts again?

asi1024:
Jenkins, test this please.

asi1024:
I can't reproduce this CI failure locally, but `std::asinh` and `std::acosh` may not be defined as a device function in this CUDA version (=9.2). How about try rewriting with `std::exp` as temporary support?

kshitij12345:
Just a heads up,

Maybe this may or may not help.

The system that I have also has Cuda-9.2
```bash
$ nvcc --version
nvcc: NVIDIA (R) Cuda compiler driver
Copyright (c) 2005-2018 NVIDIA Corporation
Built on Tue_Jun_12_23:07:04_CDT_2018
Cuda compilation tools, release 9.2, V9.2.148
```
And the code compiles on it.

However, the gcc version is 6.5.0
```bash
$gcc --version
gcc (Ubuntu 6.5.0-2ubuntu1~16.04) 6.5.0 20181026
Copyright (C) 2017 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

asi1024:
Could you resolve conflict again?

asi1024:
Your several changes in `numeric.h` and `cuda/numeric.cuh` seems to have gone when you resolve conflicts.

kshitij12345:
Oh. Right. Thanks. Will do it locally in few mins.

asi1024:
Jenkins, test this please.

asi1024:
Could you resolve conflict?

asi1024:
Jenkins, test this please.

asi1024:
Jenkins failure seems unrelated to this PR.

asi1024:
LGTM! Thank you very much!

PR #6780, opened 2019-04-06
Fix `chainerx.astype` casting from `float16` to `bool` in CUDA

### LINKED ISSUE #6778: [bug] [chainerx] Incorrect cast to bool when (`dtype == float16' and 'device==cuda')
When a `float16, cuda` array is cast as `bool`, the result is inverted i.e. `0 -> True` while `Other Values -> False`

Eg:
```python
import numpy as np
import chainerx as chx

dtypes = ['int8','int16','int32', 'int64', 'float16', 'float32', 'float64']
device = ['native', 'cuda']

for dev in device:
    for dtype in dtypes:
        a = np.array([1,2,0], dtype=dtype)
        c = chx.array(a, device = dev)
        if not all(chx.array(a.astype(bool), device=dev) == c.astype(bool)):
            print('Device : {} and Dtype : {}'.format(dev, dtype))

a = chx.array([1., 2, 4, 0], dtype='float16', device='cuda')
print(a.astype(bool))
# Output -> [ False, False, False, True]
```

kshitij12345:
Fixes #6778

### DIFF
--- a/chainerx_cc/chainerx/cuda/float16.cuh
+++ b/chainerx_cc/chainerx/cuda/float16.cuh
@@ -28,7 +28,7 @@ public:
     explicit __host__ Float16(Scalar v) : Float16{static_cast<chainerx::Float16>(v)} {}
     explicit __host__ Float16(chainerx::Float16 v) : Float16{v.data(), FromDataTag{}} {}
 
-    explicit __device__ operator bool() const { return *this == Float16{0}; }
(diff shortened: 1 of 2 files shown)


niboshi:
Thank you for the fix!
Jenkins, test this please

niboshi:
Jenkins, test this please

niboshi:
LGTM

PR #6782, opened 2019-04-07
Add infinity check in debug mode

takagi:
Currently Chainer checks the existence of `NaN` in the outputs of Chainer functions in the debug mode. This PR adds infinity check as well, which is helpful on debugging in FP16 mode since FP16 values are easily become infinity values.

### DIFF
--- a/chainer/backend.py
+++ b/chainer/backend.py
@@ -36,6 +36,23 @@ def _contains_nan(x):
         return False
 
 
+def _contains_inf(x):
+    """Returns whether the input array has inf values.
+
+    Args:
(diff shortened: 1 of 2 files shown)


niboshi:
It's better not to backport. It breaks backward compatibility.

toslunar:
> It's better not to backport. It breaks backward compatibility.

I agree.

Moreover I'm not sure whether infs are so bad to raise an error.

takagi:
Okay, I close this PR for now and might revisit later. The issue I have is FP16 and its numerical stability, I'm not still sure.

PR #6786, opened 2019-04-07
Implement Trace, Diagonal, Diagonalflat for ChainerX

sierreis:
This PR should implement the trace operator. It is a work in progress at the moment though the native forward pass is functional.

Todo:
- [x] Native Forward Pass
- [x] Cuda Forward Pass
- [x] Backward Definition
- [x] Python Binding
- [x] Tests
- [x] Documentation


niboshi:
Thank you for the PR!
Can't `Trace` implemented as `Sum(Diag(x))` which would be much simpler?

sierreis:
Yea, I think that will work for the case when the input is 2D, though [`trace`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.trace.html) in numpy works with higher dimensional inputs. I think internally they have another function [`diagonal`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.diagonal.html) much like [`diag`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.diag.html) but for higher dimensions. I couldn't find something similar in ChainerX, could you point it out if it exists (or something like it)? If not, I can try and break the implementation down into these chunks if that makes sense. Thanks!

niboshi:
@sierreis Currently ChainerX does not have `diagonal`. As you stated, above I think it's better to implement `diagonal` first and use it in `trace`. Could you implement in this way?

sierreis:
Sure, the current push should break it down using `diagonal` and the existing `sum` op. I will work on testing it soon and ensuring it passes all the checks.

chainerx_cc/chainerx/array_body_leak_detection.cc
```diff
@@ -55,12 +55,12 @@ bool ArrayBodyLeakTracker::IsAllArrayBodiesFreed(std::ostream& os) const {
     return true;
 }
 
-ArrayBodyLeakDetectionScope ::ArrayBodyLeakDetectionScope(ArrayBodyLeakTracker& tracker) {
```
niboshi: Even though these fixes seem appropriate, they're irrelevant to this PR.
Could you omit those fixes from this PR? (Just one or two might be okay but it's too many).

sierreis: Sure, I'll revert these changes.

niboshi:
Thanks!

niboshi:
Please fix compile errors in Travis.

PR #6788, opened 2019-04-08
Add `namedpersistent`

crcrpar:
This PR adds `namedpersistent` method to `chainer.Link`, `chainer.Chain`, and `chainer.ChainList`.

Recent GANs sometimes tweak batch normalizations' moving averages. So, this method will be useful to some extent and pytorch implements a corresponding method.

- [x] write tests: (EDIT)To avoid unprecedented side effect, define almost the same chain in a TestChain.test_namedpersistent.


muupan:
This feature is also desired for ChainerRL to enable sharing persistent values among multiple processes using shared arrays.

kuenishi:
I marked this as medium priority as this is important for my experiments, although we have temporary implementation for this feature.

beam2d:
bump. @kmaehashi Do you have time to review it?

chainer/link.py
```diff
@@ -1202,6 +1226,15 @@ def namedparams(self, include_uninit=True):
             for path, param in link.namedparams(include_uninit):
                 yield prefix + path, param
 
+    def namedpersistent(self):
+        # type: () -> tp.Iterator[tp.Tuple[str, tp.NdArray]]
+        for ret in super(ChainList, self).namedpersistent():
+            yield ret
+        for idx, link in enumerate(self._children):
+            prefix = '/{}'.format(idx)
```
toslunar:
```suggestion
            prefix = '/%d' % idx
```
to be consistent with `namedparams` and `namedlinks`.

crcrpar:
The failure came from chainerx tests.

kmaehashi:
As Chainer is under the maintenance phase, unfortunately we decided not to merge this feature. Thanks again for your contribution.

PR #6800, opened 2019-04-08
Op registration: pooling

hvy:
Part of #6692

- ForwardBackward class for max and average pooling removed and replaced with Ops.
- (CUDA)`PoolImpl` class removed and replaced its methods with free functions `Pool` and `PoolGrad`.
- (CUDA) As max pooling requires grad grad, also introduced `MaxPoolGradGrad` function.


chainerx_cc/chainerx/routines/pooling.h
```diff
@@ -1,13 +1,45 @@
...
+    virtual std::tuple<Array, std::shared_ptr<MaxPoolState>> Call(
+            const Array& x, bool return_state, const nonstd::optional<Array>& out) = 0;
+};
+
+class MaxPoolGradOp : public Op {
+public:
+    static const char* name() { return "MaxPoolGrad"; }
+
+    virtual std::tuple<Array, std::shared_ptr<MaxPoolState>> Call(
+            const Array& gout, const std::shared_ptr<MaxPoolState>& state, bool return_state, const nonstd::optional<Array>& gx) = 0;
```
hvy: This looks a bit obfuscated (but it's consistent with how the state is returned and how the logic is controlled via a boolean argument)? 🤔

kshitij12345: Ummm will `using shared_state = std::shared_ptr<MaxPoolState>` be nicer?

hvy: That might help a bit but I was also concerned with the actual return and input values of the declaration.  However, they are probably all needed.

hvy:
We need to expose the concrete native/CUDA state classes to allow users to call the device ops directly without going through the routines. Should we fix that now?

hvy:
Fixed it now and placed the state classes in `{native,cuda}_device.h`.

niboshi:
Jenkins, test this please

niboshi:
@hvy 
Oops, plase fix the conflict.

hvy:
Thanks, fixed the conflicts.

niboshi:
Jenkins, test this please

niboshi:
Jenkins, test this please

niboshi:
Jenkins, test this please

niboshi:
LGTM

hvy:
Removing `to-be-backported` as we decided not to backport non-critical PRs.

Issue #6027, opened 2019-01-18
`Variable.unchain` is not reliable if the creator has multiple outputs


toslunar:
The documentation of the method `Variable.unchain` is correct: 

> Deletes the reference to the creator of this variable.

This does not always stop the creator from collecting the grad of the variable.

* Code to reproduce
```
>>> import numpy as np; import chainer; import chainer.functions as F
>>> x = chainer.Variable(np.arange(4).astype('f').reshape(2, 2)); x
variable([[0., 1.],
          [2., 3.]])
>>> h0, h1 = F.split_axis(x, [1], axis=0); h0, h1
(variable([[0., 1.]]), variable([[2., 3.]]))
>>> y = F.sum(h0 + h1)
>>> h0.unchain()
>>> y.backward()
>>> x.grad, h0.grad, h1.grad
(array([[1., 1.],
       [1., 1.]], dtype=float32), None, None)
```

Rishav1:
This issue arises as the backward pass gathers gradients [from all output nodes](https://github.com/chainer/chainer/blob/1130eea1fe018eb3b61d112cade283ea2ccf5731/chainer/variable.py#L1030-L1031), even from output nodes on whom unchain() have been called. In order to curb this behavior, I think we can zero out gradients in [this line](https://github.com/chainer/chainer/blob/1130eea1fe018eb3b61d112cade283ea2ccf5731/chainer/variable.py#L1031) (without poping from _grads_) for nodes that have no _creator_ (i.e. have been unchained).

toslunar:
@Rishav1 Thanks for the investigation and the concrete suggestion.  Are you interested in making it a pull request?
- Because my example is artificial, could you show your example where the issue matters, if you found the issue somewhere else?  A realistic example would raise the priority of the issue.
- In fact, I'm waiting for PRs to fix other backprop issues are merged.  The PRs #6348, #5952, #5722 might be relevant.  `#5274` and `#5724` are probably not.  If I fix the issue, I'll work after all of these.

Rishav1:
@toslunar Thanks for responding to my comment. Take a look at this example similar to yours.

```python
>>> import numpy as np; import chainer; import chainer.functions as F
>>> x = chainer.Variable(np.arange(4).astype('f').reshape(2, 2)); x
variable([[0., 1.],
          [2., 3.]])
>>> h0, h1 = F.split_axis(x, [1], axis=0); h0, h1
(variable([[0., 1.]]), variable([[2., 3.]]))
[...]

PR #6804, opened 2019-04-08
Fix unchain gradient pull

Rishav1:
This change would use Null gradients for child nodes that have been
unchained. References issue #6027. As opposed to current behavior:

```python
>>> import numpy as np; import chainer; import chainer.functions as F
>>> x = chainer.Variable(np.arange(4).astype('f').reshape(2, 2)); x
variable([[0., 1.],
          [2., 3.]])
>>> h0, h1 = F.split_axis(x, [1], axis=0); h0, h1
(variable([[0., 1.]]), variable([[2., 3.]]))
>>> y = F.sum(h0)
>>> z = F.sum(h1)
>>> h0.unchain()
>>> x.grad, h0.grad, h1.grad
(None, None, None)
>>> y.backward()
>>> x.grad, h0.grad, h1.grad
(None, array([[1., 1.]], dtype=float32), None)
>>> z.backward()
>>> x.grad, h0.grad, h1.grad
(array([[1., 1.],
       [1., 1.]], dtype=float32), None, None)
```
we would get this:

```python
>>> import numpy as np; import chainer; import chainer.functions as F
>>> x = chainer.Variable(np.arange(4).astype('f').reshape(2, 2)); x
variable([[0., 1.],
          [2., 3.]])
>>> h0, h1 = F.split_axis(x, [1], axis=0); h0, h1
(variable([[0., 1.]]), variable([[2., 3.]]))
>>> y = F.sum(h0)
>>> z = F.sum(h1)
>>> h0.unchain()
>>> x.grad, h0.grad, h1.grad
(None, None, None)
>>> y.backward()
>>> x.grad, h0.grad, h1.grad
(None, array([[1., 1.]], dtype=float32), None)
>>> z.backward()
>>> x.grad, h0.grad, h1.grad
(array([[0., 0.],
       [1., 1.]], dtype=float32), array([[1., 1.]], dtype=float32), None)
```

### DIFF
--- a/chainer/variable.py
+++ b/chainer/variable.py
@@ -1614,7 +1614,9 @@ def _backprop_to_all(outputs, retain_grad, loss_scale):
             i for i, x in enumerate(inputs) if x.requires_grad
         ])
         outputs = [y() for y in func.outputs]  # access via weak ref
-        out_grad = tuple([grads.pop(y) if y.creator_node else None
+        out_grad = tuple([grads.pop(y)
+                          if y is not None and y.creator_node is not None
+                          else None
                           for y in outputs])
         if not target_input_indexes:
             continue
--- a/chainer/variable.py
+++ b/chainer/variable.py
@@ -1614,7 +1614,8 @@ def _backprop_to_all(outputs, retain_grad, loss_scale):
             i for i, x in enumerate(inputs) if x.requires_grad
         ])
         outputs = [y() for y in func.outputs]  # access via weak ref
-        out_grad = tuple([grads.pop(y) if y.creator_node else None for y in outputs])
+        out_grad = tuple([grads.pop(y) if y.creator_node else None
+                          for y in outputs])
         if not target_input_indexes:
             continue
 
--- a/chainer/variable.py
+++ b/chainer/variable.py
@@ -1614,7 +1614,7 @@ def _backprop_to_all(outputs, retain_grad, loss_scale):
             i for i, x in enumerate(inputs) if x.requires_grad
         ])
         outputs = [y() for y in func.outputs]  # access via weak ref
-        out_grad = tuple([grads.pop(y) for y in outputs])
+        out_grad = tuple([grads.pop(y) if y.creator_node else None for y in outputs])
         if not target_input_indexes:
             continue
 
--- a/tests/chainer_tests/test_function_node.py
+++ b/tests/chainer_tests/test_function_node.py
@@ -869,29 +869,6 @@ class TestGradValueCheck(unittest.TestCase):
             chainer.grad([y], [x], [None], [None, None])
 
 
-class TestUnchainSplit(unittest.TestCase):
-
-    def test_unchain_split(self):
-        x = chainer.Variable(numpy.arange(4).astype('f').reshape(2, 2))
-        h0, h1 = chainer.functions.split_axis(x, [1], axis=0)
-        y = chainer.functions.sum(h0)
-        z = chainer.functions.sum(h1)
-        w = y + z
-        h0.unchain()
-
-        dy_dh0 = numpy.array([[1., 1.]])
-        dz_dh1 = numpy.array([[1., 1.]])
-        dy_dx = None
-        dz_dx = numpy.array([[0., 0.], [1., 1.]])
-        dw_dx = numpy.array([[0., 0.], [1., 1.]])
-
-        assert numpy.all(chainer.grad([y], [h0])[0].array == dy_dh0)
-        assert numpy.all(chainer.grad([z], [h1])[0].array == dz_dh1)
-        assert chainer.grad([y], [x])[0] is dy_dx
-        assert numpy.all(chainer.grad([z], [x])[0].array == dz_dx)
-        assert numpy.all(chainer.grad([w], [x])[0].array == dw_dx)
-
-
 class GradTestBase(object):
 
     shape = 3,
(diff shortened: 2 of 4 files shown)


chainer/variable.py
```diff
@@ -1614,7 +1614,8 @@ def add_cand(cand):
             i for i, x in enumerate(inputs) if x.requires_grad
         ])
         outputs = [y() for y in func.outputs]  # access via weak ref
-        out_grad = tuple([grads.pop(y) for y in outputs])
+        out_grad = tuple([grads.pop(y) if y.creator_node else None
```
toslunar: - `y` can be `None` if the output (node) is deleted.
- `func is None` is better because `func.__bool__` might be overridden.
```suggestion
        out_grad = tuple([grads.pop(y)
                          if y is not None and y.creator_node is not None
                          else None
```

tests/chainer_tests/test_function_node.py
```diff
@@ -869,6 +869,29 @@ def test_length_check(self):
             chainer.grad([y], [x], [None], [None, None])
 
 
+class TestUnchainSplit(unittest.TestCase):
```
toslunar: Since the test is somewhat corner-case, I'd place this test after the basic cases.  Could you move it to just after `TestGradComplex` or just before `TestGradV3Compat1`?

Could you rename the class?
```suggestion
class TestGradUnchainSplit(unittest.TestCase):
```

tests/chainer_tests/test_function_node.py
```diff
@@ -869,6 +869,29 @@ def test_length_check(self):
...
+        w = y + z
+        h0.unchain()
+
+        dy_dh0 = numpy.array([[1., 1.]])
+        dz_dh1 = numpy.array([[1., 1.]])
+        dy_dx = None
+        dz_dx = numpy.array([[0., 0.], [1., 1.]])
+        dw_dx = numpy.array([[0., 0.], [1., 1.]])
+
+        assert numpy.all(chainer.grad([y], [h0])[0].array == dy_dh0)
```
toslunar: Could you use `testing.assert_allclose`, to be ready for adding GPU tests in future?

tests/chainer_tests/test_variable.py
```diff
@@ -507,6 +507,17 @@ def test_unchain(self):
         assert ret[1].rank == old_rank
         self.check_backward((ret[1],), (ret[2],), (ret[3],), False)
 
+    def test_unchain_split(self):
+        if not len(self.x.shape) > 0:
```
toslunar:
```suggestion
        if self.x.ndim == 0:
```

tests/chainer_tests/test_variable.py
```diff
@@ -507,6 +507,17 @@ def test_unchain(self):
...
+            return
+        ret = get_variable(np, self.x)
+        ret.grad = np.zeros_like(ret.data)
+        y1, y2 = F.split_axis(ret, [5], axis=0)
+        y1.unchain()
+        z1, z2 = F.sum(y1), F.sum(y2)
+        z1.backward()
+        self.check_backward((ret,), (y1, y2,), (z2,), False)
```
toslunar: To test `y1.unchain()`, it's more natural to test that `y1` become an "input".  While it'll be less realistic, I recommend something like
```
w = z1 + z2
self.check_backward((ret, y1), (y2, z1, z2), (w,), False)
```

tests/chainer_tests/test_variable.py
```diff
@@ -507,6 +507,17 @@ def test_unchain(self):
         assert ret[1].rank == old_rank
         self.check_backward((ret[1],), (ret[2],), (ret[3],), False)
 
+    def test_unchain_split(self):
+        if not len(self.x.shape) > 0:
+            return
+        ret = get_variable(np, self.x)
```
toslunar: It looks ok to use `x` for the variable (without the confusion with `self.x`).

Rishav1: `x` is the class variable.

tests/chainer_tests/test_variable.py
Rishav1: Okay, this is better.

tests/chainer_tests/test_variable.py
Rishav1: Done.

tests/chainer_tests/test_function_node.py
Rishav1: Ok.

tests/chainer_tests/test_function_node.py
Rishav1: Alright.

tests/chainer_tests/test_variable.py
```diff
@@ -507,6 +507,17 @@ def test_unchain(self):
...
 
+    def test_unchain_split(self):
+        if self.x.ndim == 0:
+            return
+        ret = get_variable(np, self.x)
+        ret.grad = np.zeros_like(ret.data)
+        y1, y2 = F.split_axis(ret, [5], axis=0)
+        y1.unchain()
+        z1, z2 = F.sum(y1), F.sum(y2)
+        w = z1 + z2
```
toslunar: Sorry.  I've just found the backward tests in the class don't rely on default grads.  Could you set grad for each variable, including `w`?

Rishav1: Okay, I guess I missed it too. Added it.

toslunar:
Jenkins, test this please.

PR #6807, opened 2019-04-09
Use intermediate dtype in `F.mean_absolute_error` for FP16

### LINKED ISSUE #6702: F.mean_absolute_error numerically unstable with float16 arrays
In #5053, float16 support has been enabled for [F.mean_absolute_error](https://github.com/chainer/chainer/blob/master/chainer/functions/loss/mean_absolute_error.py), but it seems to produce NaN values quite easily. Usually this happens when using big batch sizes and/or if the absolute error difference is large.

The calculation is done by summing over all the absolute differences, and then dividing by the number of elements in the array. However, it appears that the summing can produce large numbers outside the possible range for `float16`. The current implementation can be broken down as follows:

```python
    x0, x1 = inputs
    self.diff = x0 - x1
    diff = self.diff.ravel()
    abs_diff = abs(diff)
    summed_abs_diff = abs_diff.sum()  # numerically unstable, can result in inf
    mean_abs_error = np.array(summed_abs_diff / diff.size, dtype=diff.dtype)
    return mean_abs_error
```

Code to reproduce error:

```python
import chainer.functions as F
import numpy as np

a = np.full(shape=(64,1,16,16), fill_value=2, dtype=np.float16)
b = np.full(shape=(64,1,16,16), fill_value=-2, dtype=np.float16)

loss = F.mean_absolute_error(a,b)
# /home/user/.local/share/virtualenvs/.../lib/python3.6/site-packages/numpy/core/_methods.py:36: RuntimeWarning: overflow encountered in reduce
#  return umr_sum(a, axis, dtype, out, keepdims, initial)
# variable(inf)

loss = F.mean_absolute_error(a.astype("float32"), b.astype("float32"))
# variable(4.)
```

Note that the actual loss (4) would still be valid in the float16 range, it is just that summing over many values results in an `inf`, which cannot then be divided to get a proper number.

Workaround ideas:

I've noticed the new `mixed16` mode that was implemented in #6456, and was wondering if there might be a similar way to do the intermediate calculations in `float32`, and cast the result back into `float16`? Thoughts?

System info:
```
Platform: Linux-4.15.0-46-generic-x86_64-with-debian-buster-sid
Chainer: 6.0.0b3
NumPy: 1.16.2
CuPy:
  CuPy Version          : 6.0.0b3
  CUDA Build Version    : 10000
  CUDA Driver Version   : 10010
  CUDA Runtime Version  : 10000
  cuDNN Build Version   : 7402
  cuDNN Version         : 7402
  NCCL Build Version    : 2307
  NCCL Runtime Version  : 2307
iDeep: Not Available
```

takagi:
Close #6702.

This PR fixes `F.mean_absolute_error` to use an intermediate dtype for FP16 inputs. `numpy.mean` is not used because old numpy does not use extra precision for FP16.

https://docs.scipy.org/doc/numpy-1.9.2/reference/generated/numpy.mean.html
https://docs.scipy.org/doc/numpy-1.16.1/reference/generated/numpy.mean.html

### DIFF
--- a/chainer/functions/loss/mean_absolute_error.py
+++ b/chainer/functions/loss/mean_absolute_error.py
@@ -2,8 +2,18 @@ import numpy
 
 import chainer
 from chainer import backend
+from chainer.backends import cuda
 from chainer import function_node
 from chainer.utils import type_check
+
+
+def _get_intermediate_dtype(dtype):
+    # Returns the dtype for intermediate calculation.
+    # For float16 input, float32 is used.
+    # Otherwise the same dtype as the parameter is used.
+    if dtype == numpy.float16:
+        return numpy.float32
+    return dtype
 
 
 class MeanAbsoluteError(function_node.FunctionNode):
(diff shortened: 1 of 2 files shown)


kmaehashi:
pfnCI, test this please.

kmaehashi:
pfnCI, test this please.

PR #6815, opened 2019-04-09
[chainerx][tests] fix skipped_backward tests to return as PASS

kshitij12345:
If `skip_backward/skip_double_back` is False, the test
is marked as skipped, even when the Forward is checked
and passes successfully. This change will make sure that,
even these tests are marked as PASS by Pytest.

I have a feeling that I saw an issue regarding the same.
But I am unable to find it. Would put a reference to it if
I am able to do so.

### DIFF
--- a/tests/chainerx_tests/op_utils.py
+++ b/tests/chainerx_tests/op_utils.py
@@ -99,7 +99,7 @@ class OpTest(chainer.testing.function_link.FunctionTestBase):
         if self.skip_double_backward_test:
             return
 
-        super(OpTest, self).run_test_backward(backend_config)
+        super(OpTest, self).run_test_double_backward(backend_config)
 
 
 class ChainerOpTest(OpTest):
--- a/tests/chainerx_tests/op_utils.py
+++ b/tests/chainerx_tests/op_utils.py
@@ -85,7 +85,7 @@ class OpTest(chainer.testing.function_link.FunctionTestBase):
         if self.skip_forward_test:
             raise unittest.SkipTest('skip_forward_test is set')
 
-        super(OpTest, self).run_test_forward
+        super(OpTest, self).run_test_forward(backend_config)
 
     def run_test_backward(self, backend_config):
         # Skipping Backward -> Test PASS
--- a/tests/chainerx_tests/op_utils.py
+++ b/tests/chainerx_tests/op_utils.py
@@ -1,5 +1,6 @@
 import inspect
 import sys
+import unittest
 
 import numpy
 import pytest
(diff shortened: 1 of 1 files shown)


kshitij12345:
Had no clue there were tests for `test` functions. Nice. Will set status to WIP till I figure what's going wrong.

kshitij12345:
@hvy Please review.

hvy:
Thanks for the PR. I agree that the current behavior can be improved. I'm wondering though if the current change will cause a test to be marked as passed even when all of forward, backward and double backward tests are skipped. Shouldn't we simply generate separate results/tests for each of them? That's consistent with how `unittest.TestCase` behaves and wouldn't be any ambiguous. cc @niboshi

niboshi:
@hvy 
>Shouldn't we simply generate separate results/tests for each of them?

Currently the three tests (forward/backward/double-backward) run sequentially in order to test expected errors (i.e. in `op_utils.NumpyOpTest`). In such tests, it should be checked that ChainerX impl. raises errors similarly to numpy, in forward call. If the error is detected, subsequent backward/double-backward tests must be skipped. It couldn't be possible if tests of the three were separated.

hvy:
Sorry this is not a direct answer to the comment above but I'd like to clarify if this PR addresses the `FunctionTestCase` (not ChainerX-related) or the `op_test` (ChainerX). I think we want to fix the latter while keeping the former as it is?

kshitij12345:
@hvy @niboshi  Sorry I don't know much about the tests in Chainer, but are there other cases besides, `forward, backward, double-backward`. If all three are skipped due to the class being marked with `skip` or `skip-if`, then the test would be marked as skipped. While if all are skipped manually then it is due to the writer's conscious decision, in which case test being marked as Success doesn't seem wrong?

Do let me know if I am understanding anything incorrectly.

hvy:
@kshitij12345 No, forward, backward and double backward are the only ones. And yes, that's a point too. We had a general offline discussion within the team and though that it'd be better for now to leave the Chainer tests as they are, while changing the ChainerX tests (`OpTest`) to be rendered as PASS in case `skip_backward` / `skip_double_backward` is `True`. In case `skip_forward` is `True`, we should render it as SKIP. Does that make sense? And if so, could you adapt your implementation to it?

The points is that the `OpTest` is running all of forward, backward and double backward tests in a single test. On the other hand, Chainer's tests using `FunctionTestCase` separates the above three into different tests with their different results of PASS/SKIP/etc. So we have to treat the slightly differently although they're using the same `skip_...` flags. Ideally, we should consider a common test utilities for both Chainer and ChainerX.

kshitij12345:
Oh I see.

tests/chainerx_tests/op_utils.py
```diff
@@ -79,6 +80,27 @@ def forward_chainerx(self, inputs):
         raise NotImplementedError(
             'Op test implementation must override `forward_chainerx`.')
 
+    def run_test_forward(self, backend_config):
```
kshitij12345: Will work the same even if this method is not overloaded, but included it so that behaviour one does'nt have to dig into `FunctionTestBase` to know the behaviour when `forward` is skipped.

kshitij12345:
There are some errors. Will try to fix and notify.

kshitij12345:
@hvy Please review

hvy:
Jenkins, test this please.

hvy:
Jenkins, test this please.

hvy:
Jenkins, test this please.

tests/chainerx_tests/op_utils.py
```diff
@@ -79,6 +80,27 @@ def forward_chainerx(self, inputs):
...
+            return
+
+        super(OpTest, self).run_test_backward(backend_config)
+
+    def run_test_double_backward(self, backend_config):
+        # Skipping Double Backward -> Test PASS
+        if self.skip_double_backward_test:
+            return
+
+        super(OpTest, self).run_test_backward(backend_config)
```
hvy: Ops, could you fix this?
```suggestion
        super(OpTest, self).run_test_double_backward(backend_config)
```

hvy:
LGTM except the comment above. Could you take a look?

tests/chainerx_tests/op_utils.py
kshitij12345: Haha. Sorry. 

Thank You.

hvy:
Jenkins, test this please.

kshitij12345:
https://github.com/chainer/chainer/projects/2#card-19090847

PR #6832, opened 2019-04-10
Simplify numeric.h

asi1024:


chainerx_cc/chainerx/cuda/numeric.cuh
```diff
@@ -22,89 +22,28 @@ __device__ inline bool IsInf(cuda::Float16 value) { return value.IsInf(); }
...
+DECLARE_CUDA_FLOAT16_FALLBACK_UNARY(Tanh, std::tanh(x))
+DECLARE_CUDA_FLOAT16_FALLBACK_UNARY(Sin, std::sin(x))
+DECLARE_CUDA_FLOAT16_FALLBACK_UNARY(Cos, std::cos(x))
+DECLARE_CUDA_FLOAT16_FALLBACK_UNARY(Tan, std::tan(x))
+DECLARE_CUDA_FLOAT16_FALLBACK_UNARY(Arcsin, std::asin(x))
+DECLARE_CUDA_FLOAT16_FALLBACK_UNARY(Arccos, std::acos(x))
+DECLARE_CUDA_FLOAT16_FALLBACK_UNARY(Arctan, std::atan(x))
+DECLARE_CUDA_FLOAT16_FALLBACK_UNARY(Exp, std::exp(x))
+DECLARE_CUDA_FLOAT16_FALLBACK_UNARY(Log, std::log(x))
+DECLARE_CUDA_FLOAT16_FALLBACK_UNARY(Sqrt, std::sqrt(x))
```
hvy: I think prefixing with `DEFINE` is more correct. What do you say?

hvy: Looks simpler and sufficient but are these cases where this isn't appropriate?
```suggestion
DECLARE_CUDA_FLOAT16_FALLBACK_UNARY(Sqrt, std::sqrt)
```

hvy:
Change the label from "code-fix" since this PR actually fixes Square.

chainerx_cc/chainerx/cuda/numeric.cuh
asi1024: Sometimes `x` may appear twice in the expression in the future.

hvy: I see, could you for the record also mention them here?

asi1024: Fixed.

asi1024: Fixed to remove `x`. In the above case, it is OK to define a unary function ahead and pass that as the second argument to the macro.

hvy:
Jenkins, test this please.

hvy:
Jenkins failure seem unrelated. Waiting for Travis.

hvy:
LGTM!

hvy:
Removing `to-be-backported` as we decided not to backport non-critical PRs.

PR #6855, opened 2019-04-11
Relax tolerances and avoid non-differentiable points for FP16 in triplet loss tests

hvy:


crcrpar:
Does this relate to https://github.com/chainer/chainer/pull/5819 and #6190?

hvy:
Ah, I did not know that it was addressed already. I think it's okay to solve it using the non-differentiable-dodging also.

crcrpar:
If the pair of relaxation and dodge_nondifferentiable enough for F.triplet test to settle, then I would like 5819 to adopt this relaxation and get merged.

hvy:
Yes, that might be better as the non-differentiable points are known.

tests/chainer_tests/functions_tests/loss_tests/test_triplet.py
```diff
@@ -21,10 +21,13 @@
 class TestTriplet(unittest.TestCase):
 
     def setUp(self):
+        # Sample differentiable inputs
         eps = 1e-3
+        if self.dtype == numpy.float16:
+            eps *= 10
```
niboshi: Is there any reason you wrote this way instead of `eps = 1e-2`?

niboshi:
Let's merge this PR first, as it's not clear if `dodge_nondifferentiable` works with this fix.

tests/chainer_tests/functions_tests/loss_tests/test_triplet.py
hvy: No, either should be fine. Do you want me to change it?

niboshi: Yes, please.

hvy: Fixed.

hvy:
PTAL.

niboshi:
Jenkins, test this please

niboshi:
LGTM

PR #6856, opened 2019-04-11
Add automatic management of snapshots (deletion and load)

kuenishi:
This branch adds two new feature to snapshot system of Chainer. The
first feature is automatic deletion of old snapshot files during the
training. This is done by adding a post-save hook interface to
snapshot writers and if the number of snapshots to retain is set as
positive integers, the snapshot writers automatically lists all files
in Trainer output directory and matches the filename format.  One
limitation is that the automatic matching is only for integers but
does not work for arbitrary formatting of objects or string. For
example, if the filename format is defined as
`snapshot_iter_{.updater.iteration}` , the automatic matching system
picks files named as `snapshot_iter_10`, `snapshot_iter_12000` and so
on.

Another feature is automatic loading of snapshot target at extension
initialization. If `autoload` option is set as True it scans the
Trainer output directory and finds matches the filename format. For
example, if files named as `snapshot_iter_10` and
`snapshot_iter_12000` are found at startup, it chooses the latter
as the latest snapshot and loads the data to target.

Note: this depends on #6762 to prevent conflict. Also this is more better version of  #6531 .


kuenishi:
Note: `find_latest_snapshot` and `find_stale_snapshots` are exposed for future extension of swapping writer and lister that access HDFS (and other storage systems).

niboshi:
It looks the two features are independent.
Could you separate PRs to make review process simpler?

kuenishi:
My intention to mark this as 6.0.0 milestone after feature fix is because this has high priority in our internal cluster system; It is important 1) to help Chainer users to casually get prepared for job preemption by our internal job scheduler system (along with automatic start), and 2) to save our storage spaces by automatically deleting stale snapshots, which is anticipated to increase and to occupy more space if we encourage taking snapshots more by (1).

The reason I didn't split the PR was both new features stem from one core internal function `_find_snapshot_files`. It defines behaviours of both autocleanup and autoload so we should discuss their spec at the same context. If I split this into two PRs then the scope and place of discussion might be confused.

I want to discuss several points such as:
- It only matches integers but not string defined by the format, do we need more smart matching?
- The newly exposed function names are a bit long, can they be named more cooler?
- How's the idea itself?

niboshi:
We agreed to backport this PR.

kuenishi:
Jenkins, test this please.

niboshi:
Sorry for delay.
To me, the collection part (i.e. regex part) looks a bit too adhoc.
It also does not always work (as you stated), which is confusing.
I think this inverse problem is simply too difficult and we should find another way round.
I don't have a good alternative idea, though.
Maybe the problem is that the original spec of `snapshot` is too general... We should have limited to some specific directory.

niboshi:
Discussed with the team:

* We don't want to add another interface with shrunk specification (e.g. limit the directory of snapshot files) as that would affect many users and their environment (e.g. their auxiliary scripts handling snapshot files).
* As the regexp method is too complicated, one alternative idea is to recognize only the prefix and the suffix of the format (it's easily done by detecting `{` and `}` respectively).
* Users may request more sophiscated parser (like the current regexp method). In that case we may make the listing functionality pluggable, but not necessarily in this PR.

@kuenishi  How do you think?

kuenishi:
@niboshi that'd be fine. Let me have time to work on it.

kuenishi:
@niboshi I pushed new implementation that only uses prefix and suffix matching by detected `{}` . Please check it out.

chainer/training/extensions/_snapshot.py
```diff
@@ -108,6 +193,15 @@ def snapshot(savefun=None,
             used.
         snapshot_on_error (bool): Whether to take a snapshot in case trainer
             loop has been failed.
+        num_retain (int): Number of snapshot files to retain
+            through the cleanup. Must be positive integer. Automatic deletion
+            of old snapshots only works when the filename is string.
+        lister (callable): A function to find files from Trainer output
+            directory.
+        loadfun (callable): A function to load file content to the target.
```
niboshi: Hmm, do you have to specify this?
As `savefun` is already given, and `savefun` and `loadfun` are essentially the two aspects of a single logic, I feel it's bizzare to have to give both...

chainer/training/extensions/snapshot_writers.py
```diff
@@ -57,6 +57,29 @@ def save(self, filename, outdir, target, savefun, **kwds):
             savefun(tmppath, target)
             shutil.move(tmppath, os.path.join(outdir, filename))
 
+        self.post_save(filename, outdir, target, savefun, **kwds)
+
+    def add_hook(self, hook_fun):
```
niboshi: The interface of the hook also needs more careful design.
Current form prohibits future extensions (like adding an argument to a hook function).
In order to avoid that, we should consider interface like `LinkHook` (it encupsulates arguments in a single instance).

chainer/training/extensions/_snapshot.py
```diff
@@ -1,9 +1,85 @@
...
 from chainer.training import extension
 from chainer.training.extensions import snapshot_writers
 from chainer.utils import argument
 
 
+def _find_snapshot_files(fmt, path, lister, get_ts):
+    # Only prefix and suffix match
+
+    prefix = fmt.split('{')[0]
+    suffix = fmt.split('}')[-1]
```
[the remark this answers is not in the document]
kuenishi: Doesn't that mean it accepts prefix like `abc` for file `abc{def....` ? Longer prefix and suffix would be nicer as formats like `abc}def{ghi` is unformattable and thus illegal.

chainer/training/extensions/_snapshot.py
```diff
@@ -1,9 +1,85 @@
...
+
+    snapshot_files = _find_snapshot_files(fmt, path, lister, get_ts)
+
+    if len(snapshot_files) > 0:
+        _, filename = snapshot_files[-1]
+        return filename
+
+
+def find_stale_snapshots(fmt, path, num_retain,
+                         lister=os.listdir, get_ts=None):
```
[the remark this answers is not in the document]
kuenishi: Future extension to other file systems like HDFS or AWS S3 will need this feature but might be confusion. Will do that

chainer/training/extensions/_snapshot.py
kuenishi: I understand that. This is needed for future extension as I mentioned above, but the abstraction might be unmature. I want further discussion about this, separate from this PR.

chainer/training/extensions/snapshot_writers.py
kuenishi: I see that idea. Possible solutions for this (adding cleanup to the writer) are:

1. Discuss the general interface for snapshot writers extension
2. Make this hook built-in and do not expose as an API, say like `def _add_cleanup_hook()`

I would prefer 2 as there are few possibility to extend the snapshot writers or we just haven't yet learned enough use cases for this. What do you think?

chainer/training/extensions/_snapshot.py
kuenishi: I don't have good idea on how to obtain the loading function, as `savefun` itself doesn't indicate anything to computers...

kuenishi:
@niboshi I removed `listdir` and `get_ts` abstraction, but still want to hear thoughts on prefix and suffix matchinng, hook and `loadfun` .

chainer/training/extensions/_snapshot.py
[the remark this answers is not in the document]
niboshi: I see, I was misunderstanding about the format.
I found that it still fails to parse formats including `{{` or `}}`, but I think we can just ignore such cases. Could you add a TODO about this?

chainer/training/extensions/_snapshot.py
niboshi: How about omitting `loadfun` in this PR and assuming `savefun` is `save_npz`?
Or do you want to specify something other than `load_npz`?
If so, we would have to either reconsider the design of `savefun/loadfun`, or make `loadfun` a hidden argument.

kuenishi:
@niboshi ping?

chainer/training/extensions/_snapshot.py
kuenishi: Assuming `save_npz` would be nice for now. I'll remove it tomorrow.

chainer/training/extensions/snapshot_writers.py
niboshi: Option 2. looks nice.
I assume users won't be able to insert their custom hooks, right?

niboshi:
@kuenishi 
Sorry for being late. I commented on the last thread https://github.com/chainer/chainer/pull/6856#discussion_r298432858.
Could you proceed with that?

chainer/training/extensions/_snapshot.py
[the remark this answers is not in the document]
kuenishi: Addressed at 1f0c81e

chainer/training/extensions/_snapshot.py
kuenishi: Addressed at 1f0c81e , but with no safeguard...

chainer/training/extensions/snapshot_writers.py
kuenishi: Done with 1f0c81e . Technically possible for users to add custom hooks, by you know, Python is hackable anywhere. I meant 'do not expose' as 1. undocumented, 2. method name starts with `_` .

kuenishi:
I think it's ready for review again :)

chainer/training/extensions/_snapshot.py
```diff
@@ -1,9 +1,72 @@
+import os
+
+import chainer
 from chainer.serializers import npz
 from chainer.training import extension
 from chainer.training.extensions import snapshot_writers
 from chainer.utils import argument
 
 
+def _find_snapshot_files(fmt, path):
```
niboshi: Please write in a comment what this function is expected to return.

chainer/training/extensions/_snapshot.py
```diff
@@ -1,9 +1,72 @@
...
+
+    def _prepend_mtime(f):
+        t = os.stat(os.path.join(path, f)).st_mtime
+        return (t, f)
+
+    return sorted(_prepend_mtime(file) for file in matched_files)
+
+
+def find_latest_snapshot(fmt, path):
+    """Finds the latest snapshots in a directory
```
niboshi: Ditto. Write about the return value.

chainer/training/extensions/_snapshot.py
```diff
@@ -1,9 +1,72 @@
...
+            only examined. Also, files' staleness is judged
+            by timestamps. The default is metime.
+        path (str): a directory path to search for snapshot files.
+
+    """
+    snapshot_files = _find_snapshot_files(fmt, path)
+
+    if len(snapshot_files) > 0:
+        _, filename = snapshot_files[-1]
+        return filename
```
niboshi: There's no `return` statement after this `if` clause. Please return `None` explicitly.

chainer/training/extensions/_snapshot.py
```diff
@@ -1,9 +1,72 @@
...
+
+
+def find_stale_snapshots(fmt, path, num_retain):
+    """Finds stale snapshots in a directory
```
niboshi: Ditto. Write about the return value.

chainer/training/extensions/_snapshot.py
```diff
@@ -45,6 +108,12 @@ def snapshot_object(target, filename, savefun=None, **kwargs):
             used.
         snapshot_on_error (bool): Whether to take a snapshot in case trainer
             loop has been failed.
+        num_retain (int): Number of snapshot files to retain
+            through the cleanup. Must be positive integer. Automatic
+            deletion of old snapshots only works when the filename is string.
+        autoload (boolean): With this enabled, the extension automatically
```
niboshi:
```suggestion
        autoload (bool): With this enabled, the extension automatically
```

chainer/training/extensions/_snapshot.py
```diff
@@ -45,6 +108,12 @@ def snapshot_object(target, filename, savefun=None, **kwargs):
...
+            finds the latest snapshot and loads the data to the target.
+            Automatic loading only works when the filename is string.
```
niboshi:
```suggestion
            Automatic loading only works when the filename is a string.
```

chainer/training/extensions/_snapshot.py
```diff
@@ -108,6 +177,14 @@ def snapshot(savefun=None,
             used.
         snapshot_on_error (bool): Whether to take a snapshot in case trainer
             loop has been failed.
+        num_retain (int): Number of snapshot files to retain
+            through the cleanup. Must be positive integer. Automatic deletion
+            of old snapshots only works when the filename is string.
+        autoload (boolean): With this enabled, the extension
```
niboshi:
```suggestion
        autoload (bool): With this enabled, the extension
```

chainer/training/extensions/_snapshot.py
```diff
@@ -108,6 +177,14 @@ def snapshot(savefun=None,
...
+            automatically finds the latest snapshot and loads the data
+            to the target.  Automatic loading only works when the
+            filename is string. It assumes snapshots generated by
```
niboshi:
```suggestion
            filename is a string. It assumes snapshots generated by
```

niboshi:
```suggestion
            filename is string. It is assumed that snapshots are generated by
```

chainer/training/extensions/_snapshot.py
```diff
@@ -108,6 +177,14 @@ def snapshot(savefun=None,
...
+            ``save_npz``.
```
niboshi:
```suggestion
            :func:`chainer.serializers.save_npz`.
```

chainer/training/extensions/_snapshot.py
```diff
@@ -213,6 +293,36 @@ def __init__(
...
+
+            outdir = trainer.out
+            filename = find_latest_snapshot(self.filename, outdir)
+            if filename is None:
+                if chainer.is_debug():
+                    print("No snapshot file that matches {} was found"
+                          .format(self.filename))
+            else:
+                snapshot_file = os.path.join(outdir, filename)
+                npz.load_npz(snapshot_file, target)
```
niboshi: Please write a comment that describes:
* Currently it is assumed snapshots are saved by `save_npz`.
* In order to support general snapshot loaders, we need to first reconstruct the design of savefun/loadfun.

chainer/training/extensions/_snapshot.py
```diff
@@ -213,6 +293,36 @@ def __init__(
         self.condition = condition
         self.writer = writer
         self._snapshot_on_error = snapshot_on_error
+        self.num_retain = num_retain
+        self.autoload = autoload
+
+    def initialize(self, trainer):
+        if self.autoload:
```
niboshi: Please write a comment about what's being done in this code block.

chainer/training/extensions/_snapshot.py
```diff
@@ -213,6 +293,36 @@ def __init__(
...
+            target = trainer if self._target is None else self._target
+
+            outdir = trainer.out
+            filename = find_latest_snapshot(self.filename, outdir)
+            if filename is None:
+                if chainer.is_debug():
+                    print("No snapshot file that matches {} was found"
```
niboshi: Please use single quotes for consistency.

chainer/training/extensions/_snapshot.py
```diff
@@ -213,6 +293,36 @@ def __init__(
...
+                          .format(self.filename))
+            else:
+                snapshot_file = os.path.join(outdir, filename)
+                npz.load_npz(snapshot_file, target)
+                if chainer.is_debug():
+                    print("Snapshot loaded from", snapshot_file)
+
+        if hasattr(self.writer, '_add_cleanup_hook') and \
```
niboshi: Ditto. Please write a comment about what's being done in this code block.

chainer/training/extensions/_snapshot.py
```diff
@@ -213,6 +293,36 @@ def __init__(
...
+            filename = find_latest_snapshot(self.filename, outdir)
+            if filename is None:
+                if chainer.is_debug():
+                    print("No snapshot file that matches {} was found"
+                          .format(self.filename))
+            else:
+                snapshot_file = os.path.join(outdir, filename)
+                npz.load_npz(snapshot_file, target)
+                if chainer.is_debug():
+                    print("Snapshot loaded from", snapshot_file)
```
niboshi: Ditto. Please use single quotes

chainer/training/extensions/_snapshot.py
niboshi: As per PEP-8, please avoid using `\` and use parentheses instead.
Please place the operator `and` after the line break.

chainer/training/extensions/_snapshot.py
```diff
@@ -213,6 +293,36 @@ def __init__(
...
+
+        if hasattr(self.writer, '_add_cleanup_hook') and \
+           self.num_retain > 0 and isinstance(self.filename, str):
+
+            def _cleanup(filename, outdir, target, savefun, **kwds):
```
niboshi: This is no longer exposed to users. Please remove unnecessary arguments.
(it looks no argument is needed?)

chainer/training/extensions/snapshot_writers.py
```diff
@@ -57,6 +57,29 @@ def save(self, filename, outdir, target, savefun, **kwds):
             savefun(tmppath, target)
             shutil.move(tmppath, os.path.join(outdir, filename))
 
+        self._post_save(filename, outdir, target, savefun, **kwds)
+
+    def _add_cleanup_hook(self, hook_fun):
+        """Adds post-save hook function.
```
niboshi: Please update the comment. It is no longer exposed to users and it looks some sentences are obsolete.

examples/mnist/train_mnist.py
```diff
@@ -44,8 +44,6 @@ def main():
                         'negative integer, NumPy arrays are used')
     parser.add_argument('--out', '-o', default='result',
                         help='Directory to output the result')
-    parser.add_argument('--resume', '-r', type=str,
```
niboshi: Should we remove this argument?
Some users might want to control the resume behavior.
I think we should give users a choice from the following:
* Resume from a specific snapshot (`--resume`).
* Enable `autoload` and load the latest snapshot if available.
* Disable `autoload` and always start a new training.

So, how about retaining `--resume` argument, and introducing another argument to enable `autoload` feature?

kuenishi:
@niboshi Thanks for sole and careful review, I was happy to fix all of them. I hope now they're all fixed.

niboshi:
>Note: `find_latest_snapshot` and `find_stale_snapshots` are exposed for future extension of swapping writer and lister that access HDFS (and other storage systems).

Could you elaborate this point? I feel they cannot be reused as they currently are in HDFS etc., partly because `_find_snapshot_files` is internal.

chainer/training/extensions/_snapshot.py
```diff
@@ -1,9 +1,96 @@
...
+    Args:
+        fmt (str): format string to match with file names of
+            existing snapshots, where prefix and suffix are
+            only examined. Also, files' staleness is judged
+            by timestamps. The default is metime.
+        path (str): a directory path to search for snapshot files.
+
+    Returns:
+        A sorted list of pair of ``mtime, filename``, whose file
+        name that matched the format ``fmt`` directoly under ``path``.
```
niboshi: A nitpick 
```suggestion
        name that matched the format ``fmt`` directly under ``path``.
```

chainer/training/extensions/snapshot_writers.py
```diff
@@ -57,6 +57,28 @@ def save(self, filename, outdir, target, savefun, **kwds):
...
+
+        Technically, arbitrary user-defined hook can be called, but
+        this is intended for cleaning up stale snapshots.
+
+        Args:
+            hook_fun (callable): callable funtion to be called
+                right after save is done. It takes no arguments.
+
+        """
+        if not hasattr(self, '_post_save_hooks'):
```
niboshi: How about initializing it with `_post_save_hooks = []` in `Writer.__init__`?

chainer/training/extensions/snapshot_writers.py
```diff
@@ -57,6 +57,28 @@ def save(self, filename, outdir, target, savefun, **kwds):
...
+            self._post_save_hooks = []
+        self._post_save_hooks.append(hook_fun)
+
+    def _post_save(self):
+        if hasattr(self, '_post_save_hooks'):
```
niboshi: By applying the comment above, this check can be omitted.

tests/chainer_tests/training_tests/extensions_tests/test_snapshot.py
```diff
@@ -97,4 +106,147 @@ def exception_raiser(trainer):
         self.assertTrue(os.path.exists(self.filename))
 
 
+@pytest.mark.parametrize('fmt', ['snapshot_iter_{}',
+                                 'snapshot_iter_{}.npz',
+                                 '{}_snapshot_man_suffix.npz'])
+def test_find_snapshot_files(fmt):
```
niboshi: Please write tests using `unittest.TestCase` for consistency.
We could allow pytest-style tests, but it should at least be consistent within a file.

chainer/training/extensions/snapshot_writers.py
kuenishi: I didn't do that probably because  Writer didn't have `__init__()` method, but will do that.

tests/chainer_tests/training_tests/extensions_tests/test_snapshot.py
kuenishi: I've updated to all class-based style, though I'm not sure the usage of parameterized tests is right.

chainer/training/extensions/snapshot_writers.py
kuenishi: Removed; thanks

chainer/training/extensions/_snapshot.py
kuenishi: I included the fix on this in my commit.

kuenishi:
> Could you elaborate this point? I feel they cannot be reused as they currently are in HDFS etc., partly because _find_snapshot_files is internal.

At the first aim I wanted to include new feature addition on pluggable writers to transparently switch backend storage from POSIX to HDFS, but IIRC we agreed at offline chat that those feature addition should be another PR. My original comment was before that chat, if I really remember correctly.

Anyhow I updated the PR onto the comments, thanks again @niboshi

tests/chainer_tests/training_tests/extensions_tests/test_snapshot.py
niboshi: It looks good. Thanks!

niboshi:
@kuenishi Thank you for explanation. So I think `find_latest_snapshot` and `find_stale_snapshots` can be made internal.

kuenishi:
yes, indeed. I should have done it

niboshi:
Thanks!
Jenkins, test this please

niboshi:
Could you check the failure in Travis CI?

kuenishi:
Sure. Test failures in Jenkins were all from ChainerX so IMO they're flaky tests (not sure). I'll dig in test failure from Travis.

kuenishi:
@niboshi it was a new bug in the test code made during the review process. Now fixed.

kuenishi:
Jenkins, test this please.

tests/chainer_tests/training_tests/extensions_tests/test_snapshot.py
```diff
@@ -142,7 +142,7 @@ def test_find_latest_snapshot(self):
         for file in files[:-1]:
             file = os.path.join(self.path, file)
             open(file, 'w').close()
-        time.sleep(10e-3)
+            time.sleep(10e-2)
```
niboshi: 10e-2 = 0.1?
0.1 * 100 = 10 seconds seem a bit too long for a single test.
Why is it needed?

kuenishi: All these feature uses timestamps to sort the snapshot files. To make sure files' mtimes are correctly sorted in filesystems which have 1-second resolution e.g. HFS+ or ext3, I put sleep at each iteration. For example, format like `trainer_snapshot_{.iteration}` and filenames like `trainer_snapshot_9` and `trainer_snapshot_99` need timestamp to sort out the latest one without using integer knowledge in the format. If you think it's too long, I can retry Travis test with `10e-3` seconds interval.

kuenishi: I also of course understand that 10 seconds is too long for one test.  Will try 10e-3 second.

kuenishi:
It looks `10e-3` is enough as an interval and I think it's good to go.

niboshi:
Another approach for better determinacy would be to [set file timestamps manually](https://docs.python.org/3/library/os.html#os.utime).
How do you think?

kuenishi:
I didn't know `os.utime` and it'd be awesome! I pushed the change and carefully updated the test code, putting `utime` call to the place where it's really needed.

kuenishi:
Jenkins, test this please.
@niboshi I finally got the Travis test pass. Hopefully it also passes other CIs..

tests/chainer_tests/training_tests/extensions_tests/test_snapshot.py
```diff
@@ -97,4 +106,160 @@ def exception_raiser(trainer):
...
+
+            # Same comment applies here. See comment in ``TestFindSnapshot``
+            t = base_timestamp + i
+            os.utime(file, (t, t))
+
+        stale = list(_find_stale_snapshots(fmt, self.path, retain))
+        assert max(length-retain, 0) == len(list(stale))
+        stales = [fmt.format(i) for i in range(0, max(length-retain, 0))]
+        for lhs, rhs in zip(stales, stale):
+            lhs == rhs
```
niboshi: `assert` is missing? 
You may as well just write `assert stales == stale`.
Perhaps it's better to rename `stales` to something like `expected`.

tests/chainer_tests/training_tests/extensions_tests/test_snapshot.py
```diff
@@ -97,4 +106,160 @@ def exception_raiser(trainer):
...
+        assert trainer._done
+
+        pattern = os.path.join(trainer.out, "snapshot_iter_*")
+        found = [path for path in glob.glob(pattern)]
+        assert retain == len(found)
+        found.sort()
+
+        for lhs, rhs in zip(['snapshot_iter_{}'.format(i)
+                             for i in range(8, 10)], found):
+            lhs == rhs
```
niboshi: Ditto.
You could compare two lists directly here, too.

kuenishi:
mmmm... sorry about my test code being flaky. Now I hope it fixed.

kuenishi:
/test
Now this time around it should work

kuenishi:
@niboshi Thanks for patient review, this time all tests seem passing to me.

tests/chainer_tests/training_tests/extensions_tests/test_snapshot.py
```diff
@@ -97,4 +106,168 @@ def exception_raiser(trainer):
...
+        for file in itertools.chain(noise, files, noise2):
+            file = os.path.join(self.path, file)
+            open(file, 'w').close()
+
+        snapshot_files = _find_snapshot_files(self.fmt, self.path)
+
+        expected = sorted([self.fmt.format(i) for i in range(1, 100)])
+        assert len(snapshot_files) == 99
+        timestamps, snapshot_files = zip(*snapshot_files)
+        expected == sorted(list(snapshot_files))
```
niboshi: `assert` is still missing 😁

tests/chainer_tests/training_tests/extensions_tests/test_snapshot.py
```diff
@@ -97,4 +106,168 @@ def exception_raiser(trainer):
...
+            open(file, 'w').close()
+
+            # Same comment applies here. See comment in ``TestFindSnapshot``
+            t = base_timestamp + i
+            os.utime(file, (t, t))
+
+        stale = list(_find_stale_snapshots(fmt, self.path, retain))
+        assert max(length-retain, 0) == len(list(stale))
+        expected = [fmt.format(i) for i in range(0, max(length-retain, 0))]
+        expected == stale
```
niboshi: Ditto.

tests/chainer_tests/training_tests/extensions_tests/test_snapshot.py
```diff
@@ -97,4 +106,170 @@ def exception_raiser(trainer):
...
+        for i, file in enumerate(files[:-1]):
+            file = os.path.join(self.path, file)
+            open(file, 'w').close()
+
+            # Same comment applies here. See comment in ``TestFindSnapshot``
+            t = base_timestamp + i
+            os.utime(file, (t, t))
+
+        file = os.path.join(self.path, fmt.format(length))
+        open(file, 'w').close()
```
niboshi: Why does the last item have to be handled differently?

niboshi:
Jenkins, test this please

niboshi:
LGTM!

kuenishi:
Thanks again for working with long review rounds :)

PR #6859, opened 2019-04-11
[WIP] Support ChainerX in MNIST model parallel example

niboshi:
Part of #6661

TODO: Needs to fix `StandardUpdater` not to transfer the model parameters.

See this comment for more detail: https://github.com/chainer/chainer/pull/4139#issuecomment-366684813


niboshi:
#7225

PR #6865, opened 2019-04-11
Rename `CHAINERX_REGISTER_OP_{NATIVE,CUDA}` to `CHAINERX_{NATIVE,CUDA}_REGISTER_OP`

hvy:
Fixes https://github.com/chainer/chainer/issues/6864

PR #6870, opened 2019-04-11
[chainerx] std op macro

kshitij12345:
References #6608


chainerx_cc/chainerx/cuda/cuda_device/std_ops.h
```diff
@@ -0,0 +1,60 @@
+#pragma once
```
hvy: Why not include the definitions in these files to the existing `{native,cuda}/op_regist.h` ?

hvy:
Thanks for continuing with the implementation. Could you take a look at the comments above?

chainerx_cc/chainerx/cuda/cuda_device/std_ops.h
```diff
@@ -0,0 +1,60 @@
...
+    CHAINERX_CUDA_REGISTER_ELTWISE_DTYPE_UNARY_OP(func, func_def, VisitFloatingPointDtype)
+
+#define CHAINERX_CUDA_REGISTER_ELTWISE_UNARY_OP(func, func_def) CHAINERX_CUDA_REGISTER_ELTWISE_DTYPE_UNARY_OP(func, func_def, VisitDtype)
+
+#define CHAINERX_CUDA_REGISTER_ELTWISE_DTYPE_BINARY_OP(func, func_def, visit_dtype)                 \
+                                                                                                    \
+    template <typename T>                                                                           \
+    struct func##Impl {                                                                             \
+        using CudaType = cuda_internal::DataType<T>;                                                \
+        __device__ void operator()(int64_t /*i*/, CudaType x1, CudaType x2, CudaType& out) func_def \
```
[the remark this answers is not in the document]
kshitij12345: Yeah, I was planning for that actually. Yesterday I was getting `unused-variable` error. Will have a new to have a new macro with index available. Sounds right?

chainerx_cc/chainerx/cuda/cuda_device/std_ops.h
kshitij12345: Oh my bad, since the files that I changed had them, I didn't notice the missing header. Thanks.

chainerx_cc/chainerx/cuda/cuda_device/std_ops.h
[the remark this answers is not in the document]
hvy: I'm wondering if it'd be possible with this macros https://github.com/chainer/chainer/blob/master/chainerx_cc/chainerx/macro.h#L15-L17 🤔 .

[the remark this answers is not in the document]
kshitij12345: I tried `void operator()(int64_t i, T x, T& out) { CHAINERX_ASSERT(func_def); }` . But compiler still complains.

I don't think it will work as I also tried a minimal version [here](https://tio.run/##XU9LS8NAEL7vrxgohF1ogr02aSFI6UmUBPUgsqy7k7qQbMI@ihDy1415HLTOZZjvyciuiy9SjuNGG1kHhZDp1nmLojmSX0w6r2r9cQM1wn/@Ba4ofWuPhGwUVtoglM9Pxaks@UN@XzzSJEkY0GurFaNv79CDRR@sgUrUDlMYKIMoAsr5S87z4lxyzhghsx6CCQ4Vd6Gz6BzVxsPXFqYVQRs86wlMc9vWzwwcYHc3RbOUDITMtkZoQxlZHTMwqdLl@Nex2y7RKSzk9P1@L@fELFstwzh@y6oWFzfGr2htaw9rQtwJKxr0aH8A) which doesn't compile. Let me know if I am doing something wrong.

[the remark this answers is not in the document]
kshitij12345: https://stackoverflow.com/a/20077422/5602957

How about the second method using unnamed parameter in the template. I tried that and does work.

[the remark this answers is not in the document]
hvy: Ah I meant to pass the index as the argument to `CHAINERX_ASSERT`. Something along the lines of the following inside the macro definition.
```cpp
{
    CHAINERX_ASSERT(i);
    func_def;
}
```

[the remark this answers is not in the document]
kshitij12345: As per the stackoverflow answer. 
```
template <class T>
void ignore_unused(T&) {} 

//later...
auto x = /* ... */;
ignore_unused(x);
```
This is better than `(void)i`.
Your thoughts?

I am unable to find the other link which neatly explained why this is better comparted to `void` casting. Will share the link once I find it.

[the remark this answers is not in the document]
hvy: I see. The point is that we already have `(void)variable;` occurrences in our source code so it's consistent and fine for this PR. However, if there is a general better approach, we might consider applying that to all (or wherever possible) of the occurrence.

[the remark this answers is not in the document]
kshitij12345: Have used `(void)i`.

I was also thinking about allowing the implementer to also be able to cast the based on `input and output` variable.

Maybe pass in a lambda that takes in input and output and return casted version of both?
For example on 
```
[](auto x, auto output) -> std::tuple<Array, Array> {
    // cast and return
};
```

Your thoughts?

hvy:
Jenkins, test this please.

chainerx_cc/chainerx/cuda/cuda_device/std_ops.h
hvy: Sorry for pinging but I understand this is still a TODO?

chainerx_cc/chainerx/cuda/cuda_device/std_ops.h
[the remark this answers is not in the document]
hvy: Do you have concrete cases where this applies? We briefly discussed this offline but thought that it could wait.

hvy:
Could you give us a status of this PR? We'd actually like to merge it as soon as possible.

kshitij12345:
Have included the header as well. So if dtype is not needed. Will remove
the [WIP]

chainerx_cc/chainerx/cuda/cuda_device/std_ops.h
kshitij12345: Its been taken care off

hvy: Ah, what I meant with this comment in the first place was that the `std_ops.h` are not needed and that the content of this file can be moved to the existing `op_regist.h` instead.

hvy:
Please take a look at comment https://github.com/chainer/chainer/pull/6870#discussion_r275200230.

chainerx_cc/chainerx/cuda/cuda_device/std_ops.h
kshitij12345: Oh sorry, I misunderstood that. Will fix it in few hours

hvy:
Sorry the former op registration macro has been renamed in https://github.com/chainer/chainer/pull/6865. Could you solve those conflicts?

kshitij12345:
Sure on it,

chainerx_cc/chainerx/cuda/cuda_device/arithmetic.cu
```diff
@@ -68,28 +46,7 @@ public:
...
-        CudaSetDeviceScope scope{device.index()};
-        VisitNumericDtype(out.dtype(), [&](auto pt) {
-            using T = typename decltype(pt)::type;
-            Elementwise<const T, const T, T>(SubtractImpl<T>{}, x1_cast, x2_cast, out);
-        });
-    }
-};
-
-CHAINERX_CUDA_REGISTER_OP(SubtractOp, CudaSubtractOp);
+CHAINERX_CUDA_REGISTER_ELTWISE_DTYPE_BINARY_OP(Subtract, { out = ArithmeticOps<CudaType>::Subtract(x1, x2); }, VisitNumericDtype);
```
hvy: I'm wondering if we should use the macro as follows, requiring the whole class name including the "Op" suffix. Benefits are:
- More consistent with `CHAINERX_{NATIVE,CUDA}_REGISTER_OP`.
- Better grep-ability.
- Appending "Op" is convention rather than an enforcement.

What do you think?
```suggestion
CHAINERX_CUDA_REGISTER_ELTWISE_DTYPE_BINARY_OP(SubtractOp, { out = ArithmeticOps<CudaType>::Subtract(x1, x2); }, VisitNumericDtype);
```

chainerx_cc/chainerx/cuda/op_regist.h
```diff
@@ -6,3 +6,69 @@
 // Register an op statically in CudaBackend.
 #define CHAINERX_CUDA_REGISTER_OP(key_op_cls, op_cls) \
     static chainerx::internal::OpRegistrar<chainerx::cuda::CudaBackend, key_op_cls, op_cls> s_cuda_backend_op_##op_cls{};
+
+#define CHAINERX_CUDA_REGISTER_ELTWISE_DTYPE_UNARY_OP(func, func_def, visit_dtype)      \
```
hvy: I think the following better describe the usage of the parameters. The "op_name" sort of depends on https://github.com/chainer/chainer/pull/6870/files#r275277089. 
```suggestion
#define CHAINERX_CUDA_REGISTER_ELTWISE_DTYPE_UNARY_OP(op_name, op_body, visit_dtype)      \
```

chainerx_cc/chainerx/cuda/cuda_device/arithmetic.cu
kshitij12345: Makes sense. Ties up with the original `Op` structure as well.

chainerx_cc/chainerx/cuda/op_regist.h
kshitij12345: True. Makes sense.

hvy:
Jenkins, test this please.

hvy:
Jenkins, test this please.

chainerx_cc/chainerx/cuda/op_regist.h
hvy: Sorry I now noticed that we could align the first argument name with  `CHAINERX_{NATIVE,CUDA}_REGISTER_OP`, but we should do that in a separate PR.

hvy:
Jenkins, test this please.

hvy:
Failures are unrelated. LGTM!

chainerx_cc/chainerx/cuda/op_regist.h
hvy: Addressed in https://github.com/chainer/chainer/pull/6941

hvy:
May I modify the title of PR to be more descriptive, since it (and all other PRs) will be rendered in the release note?

kshitij12345:
All yours

PR #6872, opened 2019-04-11
Fix type check of `F.where`

### LINKED ISSUE #6258: `type_check` can cause assertion failure
Version: current master

### Problem
```py
# Note that dtype of the second argument is int64.
F.where(y.array < 0, numpy.array(0), y)
```

This code causes an error like this:
```
  File "/home/niboshi/t/a.py", line 12, in <module>
    y = F.where(y.array == 0, numpy.array(0), y)
  File "/data2/work/w/repos/chainer/chainer/functions/array/where.py", line 87, in where
    z, = Where(condition).apply((x, y))
  File "/data2/work/w/repos/chainer/chainer/function_node.py", line 297, in apply
  File "/data2/work/w/repos/chainer/chainer/function_node.py", line 400, in _check_data_type_forward
  File "/data2/work/w/repos/chainer/chainer/functions/array/where.py", line 23, in check_type_forward
    x_type.dtype == y_type.dtype,
  File "/data2/work/w/repos/chainer/chainer/utils/type_check.py", line 545, in expect
    assert isinstance(expr, Testable)
AssertionError
```

#### Cause
`F.where`'s forward type check:
```py
        type_check.expect(
            condition.dtype == numpy.bool_,
            x_type.dtype == y_type.dtype,
        )
```
The above code fails at the second condition, causing `InvalidType` in the light-mode type check.
`FunctionNode` then catches the error and triggeres the non-light-mode type checking. The `assert` statement is evaluated only in this mode, and it fails because the first condition, `condition.dtype == numpy.bool_` is of type `bool`.

### LINKED ISSUE #6359: Fix `F.where` type checking
Related: #6258

The check for `condition` should be done in `__init__`.

toslunar:
Close #6359 and fix the example in https://github.com/chainer/chainer/issues/6258#issue-411056784.

Before the PR, `F.where` accepts the following for the `condition` argument.
- `bool` if `type_check` is skipped,
- `numpy.bool_`, and
- `ndarray` with `dtype=numpy.bool_`.

The PR also makes it accept `bool` always.


niboshi:
As pointed out this morning, I think it's more straight-forward to check `condition.dtype` in `__init__` unless there are other cases it's better to perform checks in `check_type_forward`.

niboshi:
@toslunar Can you proceed?

toslunar:
The new function `_make_variable_from_array` is an obvious combination of `make_variable` and `_get_type`.  No refactoring has done because the performance is highly concerned in the module.

> other cases

For example, it's no longer a good design to pass `t` to the function node for `F.select_item(x, t)`.  To remove `t` from the inputs, type_check_forward would get the type of `t` from the array.

niboshi:
>The new function `_make_variable_from_array` is an obvious combination of `make_variable` and `_get_type`.  No refactoring has done because the performance is highly concerned in the module.

Are you saying performance would be a concern if the check was in `__init__`? Could you explain a bit more?

>For example, it's no longer a good design to pass `t` to the function node for `F.select_item(x, t)`.  To remove `t` from the inputs, type_check_forward would get the type of `t` from the array.

If `t` were removed from inputs, it would be passed to `SelectItem.__init__`, wouldn't it?

toslunar:
> > The new function `_make_variable_from_array` is an obvious combination of `make_variable` and `_get_type`.  No refactoring has done because the performance is highly concerned in the module.
> 
> Are you saying performance would be a concern if the check was in `__init__`? Could you explain a bit more?

I was saying the PR is more straight-forward than it looks.  For the consistency, I'd rather not split type checks into two method.

niboshi:
I see.
Now I think it's better to check at `check_type_forward` because otherwise it wouldn't be possible to check consistency of shapes.

chainer/utils/type_check.py
```diff
@@ -572,6 +572,18 @@ def make_variable(value, name):
         return Variable(value, name)
 
 
+def _make_variable_from_array(array, name):
```
niboshi: How about making this function public?

chainer/functions/array/where.py
```diff
@@ -14,9 +14,12 @@ def __init__(self, condition):
         self.condition = condition
 
     def check_type_forward(self, in_types):
-        type_check.expect(in_types.size() == 2)
+        type_check._argname(in_types, ('x', 'y'))
         x_type, y_type = in_types
-        condition = self.condition
+        condition = type_check._make_variable_from_array(
+            # allow scalar `condition`
+            chainer.utils.force_array(self.condition),
```
niboshi: I think it's better to avoid having to call `force_array` explicitly, as it's nonessential and not so cheap.
How about something like `type_check._make_variable_from_array_or_scalar`?

toslunar:
> The PR also makes it accept bool always.

Let me drop it from this PR, because I'd like to fix the type check of `F.where` in v6 by a backport.

chainer/utils/type_check.py
niboshi: How about this comment?

niboshi:
Please write some tests.

toslunar:
I'd rather not forbid `np.bool_` because a user can easily get this, e.g. from `cond = z.array > 0` if `z.shape == ()`.

> > The PR also makes it accept bool always.
> 
> Let me drop it from this PR, because I'd like to fix the type check of `F.where` in v6 by a backport.

`bool` looks as OK as `np.bool_`.

niboshi:
Jenkins, test this please

niboshi:
This fix won't fix #6258, will it?

niboshi:
Jenkins, test this please

niboshi:
Test failed, but it seems irrelevant?
```
2019-06-07 13:24:19.115062 STDOUT 2776] | [TIMEOUT] /chainer/tests/chainer_tests/functions_tests/evaluation_tests/test_accuracy.py (61 seconds)
```

toslunar:
> This fix won't fix #6258, will it?

Right.  This PR only fixes the example there.  (I updated the description of the PR.)

niboshi:
Retrying the test.

niboshi:
LGTM

PR #6879, opened 2019-04-12
Add `RandWire`

hitsgub:
This PR adds [RandWire](https://arxiv.org/abs/1904.01569).
Please wait for the experimental results.

### DIFF
--- a/chainer/links/__init__.py
+++ b/chainer/links/__init__.py
@@ -41,6 +41,8 @@ from chainer.links.connection.n_step_rnn import NStepRNNReLU  # NOQA
 from chainer.links.connection.n_step_rnn import NStepRNNTanh  # NOQA
 from chainer.links.connection.parameter import Parameter  # NOQA
 from chainer.links.connection.peephole import StatefulPeepholeLSTM  # NOQA
+from chainer.links.connection.randwire import RandWire  # NOQA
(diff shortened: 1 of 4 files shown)


hitsgub:
ToDo:

- __init__
- test
- doc

hitsgub:
Errors reported by travis-ci are seems to rely on another parts.

hitsgub:
This is experimental result.

ResNet56(9-9-9) / RandWire56(9-9-9) with pre-activation
300epoch with reducing learning rate, 0-149: 0.05, 150-224: 0.005, 225-299: 0.0005.
NesterovAG, batchsize=128
RandWire_20190424090620

ResNet and RandWires seem to be same performance, like original paper.

hvy:
Thanks for the quick implementation of the paper and the experimental results. The code looks clean and we'd like to make it available to other users but are wondering if this isn't better introduced as an [example to the Chainer repository](https://github.com/chainer/chainer/tree/master/examples). If we observe that the community leans towards integrating RandWire into various kinds of experiments, we could reconsider introducing this as a feature of Chainer. By then, the desired interface might have changed and we could adapt to the more matured needs although I see that you've already worked on generalizing it. We could of course introduce this as an experimental feature, but again, cannot be sure how this algorithm (and maybe related but other AutoML algorithms) might evolve over time.

PR #6884, opened 2019-04-12
Make `InvalidType` picklable

zaltoprofen:
The following error is often occurred while using multiprocessing, then hide details of InvalidType exception.
This PR fixes this problem.

```pycon
>>> from chainer.utils.type_check import InvalidType
>>> import pickle
>>> pickle.loads(pickle.dumps(InvalidType('foo', 'bar')))
TypeError: __init__() missing 1 required positional argument: 'actual'
```

### DIFF
--- a/tests/chainer_tests/utils_tests/test_type_check.py
+++ b/tests/chainer_tests/utils_tests/test_type_check.py
@@ -1,3 +1,4 @@
+import pickle
 import sys
 import unittest
 import warnings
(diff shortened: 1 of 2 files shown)


tests/chainer_tests/utils_tests/test_type_check.py
```diff
@@ -396,4 +397,13 @@ def test_numpy_cupy_mixed_2(self):
         self.assertFalse(T.same_types(x, y, z))
 
 
+class TestInvalidType(unittest.TestCase):
+    def test_pickle(self):
+        exc = T.InvalidType('foo', 'bar')
```
niboshi: Let's initialize with a specific `msg`, instead of the default `None`.

niboshi:
Jenkins, test this please

niboshi:
LGTM
Thanks!

PR #6866, opened 2019-04-11
Fix `filename` arg of `PlotReport`

toslunar:
This PR fixes the renamed `filename` argument of `PlotReport` hasn't been effective.

PR #6928, opened 2019-04-15
[backport] Fix `filename` arg of `PlotReport`

toslunar:
Backport of #6866

### DIFF
--- a/chainer/training/extensions/variable_statistics_plot.py
+++ b/chainer/training/extensions/variable_statistics_plot.py
@@ -187,7 +187,7 @@ class VariableStatisticsPlot(extension.Extension):
                  plot_mean=True, plot_std=True,
                  percentile_sigmas=(
                      0, 0.13, 2.28, 15.87, 50, 84.13, 97.72, 99.87, 100),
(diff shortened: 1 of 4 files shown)


toslunar:

chainer/training/extensions/variable_statistics_plot.py
```diff
@@ -187,17 +187,15 @@ def __init__(self, targets, max_sample_size=1000,
                  plot_mean=True, plot_std=True,
                  percentile_sigmas=(
                      0, 0.13, 2.28, 15.87, 50, 84.13, 97.72, 99.87, 100),
-                 trigger=(1, 'epoch'), filename='statistics.png',
+                 trigger=(1, 'epoch'), filename=None,
```
niboshi: I think this causes an unwanted change in the documentation.

niboshi: I'll address it in another PR.

niboshi:
LGTM

chainer/training/extensions/variable_statistics_plot.py
niboshi: #6932

PR #6940, opened 2019-04-16
Fix `HuberLoss.forward` avoid loss of significance

asi1024:


chainer/functions/loss/huber_loss.py
```diff
@@ -30,15 +30,16 @@ def forward(self, inputs):
...
-        diff = utils.force_array(x0 - x1, dtype)
+        linear_part = utils.force_array(x0 - x1, dtype)
         delta = dtype.type(self.delta)
 
-        xp.abs(diff, out=diff)
-        y = utils.force_array(xp.square(diff), dtype)
-        diff -= delta
-        xp.maximum(diff, 0, dtype=dtype, out=diff)
-        xp.square(diff, out=diff)
-        y -= diff
```
asi1024: In the previous implementation, the loss of significance occurred.
`y` and `diff` is O(a^2), and the result is O(a) (where `a` is `x0 - x1`)

toslunar:
Jenkins, test this please.

PR #6943, opened 2019-04-16
[backport] Allow `snapshot_object` have `condition` and `writer` option

kuenishi:
Manual backport of #6762 .

### DIFF
--- a/chainer/training/extensions/_snapshot.py
+++ b/chainer/training/extensions/_snapshot.py
@@ -4,8 +4,11 @@ from chainer.training.extensions import snapshot_writers
 from chainer.utils import argument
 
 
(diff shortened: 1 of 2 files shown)


niboshi:
The diff itself looks good to me.

@toslunar Do you think we should backport this feature?
I think this PR is a `feature` addition, not `enhancement`.

toslunar:
While it's not usual to backport this, let me wait for the decision whether #6856 is backported to v6.

niboshi:
@toslunar We agreed to backport #6856, so let's merge this.

Jenkins, test this please

niboshi:
LGTM

kuenishi:
Thanks!

PR #6946, opened 2019-04-16
Fix test parameters in ChainerX math tests

niboshi:
Skips backward/double-backward tests for inputs that results in non-finite outputs.

Differentiable domains are different among inverse trigonometric functions.

### DIFF
--- a/tests/chainerx_tests/unit_tests/routines_tests/test_math.py
+++ b/tests/chainerx_tests/unit_tests/routines_tests/test_math.py
@@ -1,3 +1,5 @@
+import unittest
+
 import chainer
 import numpy
 import pytest
@@ -21,8 +23,11 @@ class IgnoreNumpyFloatingPointError(object):
 
 class UnaryMathTestBase(object):
 
+    input = None
+
     def setup(self):
         in_dtype, = self.in_dtypes
+        in_kind = numpy.dtype(in_dtype).kind
 
         if numpy.dtype(in_dtype).kind != 'f':
             self.skip_backward_test = True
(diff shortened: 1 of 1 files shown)


niboshi:
This PR blocks #6947.

tests/chainerx_tests/unit_tests/routines_tests/test_math.py
```diff
@@ -21,8 +23,11 @@ def __exit__(self, *args):
 
 class UnaryMathTestBase(object):
 
+    input = None
```
asi1024: This fix is for the tests for array-scalar binary operation, but this class is also used for other functions which take a single array for the argument. How about defining a new 
derived class for array-scalar tests?

tests/chainerx_tests/unit_tests/routines_tests/test_math.py
```diff
@@ -1780,7 +1810,7 @@ def func(self, xp, a):
         return xp.sqrt(a)
 
 
-_trignometric_hyperpolic_params = \
+_trigonometric_hyperpolic_params = \
```
hvy: A nitpick but should it be "hyperbolic"?

tests/chainerx_tests/unit_tests/routines_tests/test_math.py
niboshi: I don't necessarily understand the point, but `self.input` is already used in thi class. Such refactoring could be done separately.

tests/chainerx_tests/unit_tests/routines_tests/test_math.py
niboshi: Fixed.

asi1024:
Jenkins, test this please.

imos:
(PFN CI failed because master is updated just after the previous request.)
pfnCI, please test.

imos:
pfnCI, please test
(PFN CI failed because master is updated just after the previous request.)

asi1024:
LGTM.

PR #6947, opened 2019-04-16
Detect non-finite outputs in `check_backward`

niboshi:
Numerical gradients cannot be calculated for functions that outputs non-finite numbers.
This PR detects that as early as possible, to make debug easier. 

Merge #6946 first.


hvy:
Could you add a test for this?

niboshi:
Sorry for delay. I added a test.

hvy:
Thanks for the tests. Could you just fix `TestPower` to skip backward for nan yielding parameterizations (Travis is failing due to this)?

PR #6948, opened 2019-04-16
Increase the default columns in Travis CI

niboshi:
In Travis CI, pytest renders the progress of the tests in 80 columns.
This requires quite a bit of scroll to reach the end of outputs.

Merge #6949 first.

### DIFF
--- a/scripts/ci/travis/run-tests.sh
+++ b/scripts/ci/travis/run-tests.sh
@@ -36,6 +36,9 @@ export REPO_DIR
 export WORK_DIR
 export CHAINER_BASH_ENV
 
+# Increase the default columns for better browsability
+export COLUMNS=120
+
 
 run_prestep() {
     # Failure immediately stops the script.


beam2d:
We need consensus on the appropriate column width.

niboshi:
160 fits my screen nicely.
Even in narrower screens, the rendering won't be that bad.

hvy:
It fits my screen as well. I'm okay with 160.

hvy:
#6949 is now merged.

beam2d:
It seems my laptop can only show 125 characters in a line on Travis. 160 chars are too long for me.

hvy:
I see, thanks for the quick response. Is 120 reasonable?

kmaehashi:
Max per line was 128 on my laptop. 120 sounds nice.

niboshi:
@beam2d @kmaehashi Do you mean 120 is actually better?
Breaking lines won't make things that bad. I'd say we can even go with 500, for example.
I'll revert to 120 if you don't agree.

kmaehashi:
> This requires quite a bit of scroll to reach the end of outputs.

How about using fold feature to collapse outputs in `.travis.yml`?

sample: https://travis-ci.org/kmaehashi/chainer/builds/521187533
code: https://github.com/chainer/chainer/compare/master...kmaehashi:travis-fold?expand=1

niboshi:
@kmaehashi Oh, that looks useful. But I think the two fixes are complementary. We can apply both.

kmaehashi:
Absolutely. As folding the output helps reducing the amount of scroll, can we use 120 instead of 160? 
Actually I don't like wrapped lines either.

niboshi:
OK.

niboshi:
Changed to 120.

kmaehashi:
Thanks! I'll make separate PR for collapse.

hvy:
Jenkins, test this please.

hvy:
@beam2d Is the current change of width of 120 fine with you?

beam2d:
Yes! It fits nicely. Thanks!

hvy:
LGTM!

PR #6950, opened 2019-04-16
Remove `-v` option from Travis CI ChainerMN tests

niboshi:
The same motivation as #6948

With `-v` option, each test item name is printed in a separate line.
Without it, the progress is rendered with a file name and dots (`.`) in each line.


beam2d:
What is the effect of removing `-v`? (I want such an explanation to be written in the description for smoother review and look-back)

niboshi:
@beam2d Updated the first post.

niboshi:
Now that #6961 is merged, I don't think we need this fix any more.

PR #6952, opened 2019-04-17
Use `cublasHgemm` in float16 `chainerx::Dot`

asi1024:
This PR improves performance.


asi1024:
#6960 will resolve this issue.

PR #6956, opened 2019-04-17
Fix VGG download message

niboshi:
"Now loading caffemodel (usually it may take few minutes)"
* This message have been printed **after** downloading.
* Fix the sentence a little.
* Make the code more readable.


toslunar:
How long does it take to convert the Caffe model to npz?  If it takes few minutes, I'd lie to see both "downloading" and "converting".

niboshi:
In my test
* Download took 49.6 seconds (depends heavily on network)
* Conversion took 14.1 seconds

How do you think?

PR #6972, opened 2019-04-18
Fix cuDNN descriptor double destroy

hvy:
cuDNN descriptor move semantics were implicitly defined and could lead to double free/destroy. This issue was present in `DeriveBatchNormTensorDescriptor` but was hidden because of NRVO. 

This PR fixes those issues by simply disallowing move and copy, except for the ones currently required. We can allow more as needed and this is explained in the TODOs.

### DIFF
--- a/chainerx_cc/chainerx/cuda/cudnn.cc
+++ b/chainerx_cc/chainerx/cuda/cudnn.cc
@@ -1,7 +1,5 @@
 #include "chainerx/cuda/cudnn.h"
 
-#include <utility>
-
 #include <cudnn.h>
 #include <nonstd/optional.hpp>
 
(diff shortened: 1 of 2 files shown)


chainerx_cc/chainerx/cuda/cudnn.cc
```diff
@@ -96,6 +98,11 @@ CudnnTensorDescriptor::~CudnnTensorDescriptor() {
     }
 }
 
+CudnnTensorDescriptor::CudnnTensorDescriptor(CudnnTensorDescriptor&& other) {
+    std::swap(desc_, other.desc_);
+    other.desc_ = nullptr;
```
niboshi: How about `desc_{other.desc_}` in the initializer list?

hvy: Ah that's simpler. Thanks, fixed!

hvy:
PTAL.

niboshi:
Jenkins, test this please

hvy:
Jenkins, test this please.

niboshi:
LGTM

PR #6857, opened 2019-04-11
Support device specifiers in MNIST data parallel example

niboshi:
~Part of #6661~

~TODO: Needs `Variable.addgrad` support for ChainerX~

Some more internal fixes are needed to support ChainerX.

PR #6982, opened 2019-04-18
Support ChainerX in `F.copy`

niboshi:
ChainerX support in `F.copy`.

(Required in #6857, separated from #6859)

~TODO: test~

### DIFF
--- a/chainer/functions/array/copy.py
+++ b/chainer/functions/array/copy.py
@@ -1,13 +1,17 @@
+import chainer
+from chainer import backend
 from chainer.backends import cuda
 from chainer import function_node
 from chainer.utils import type_check
+import chainerx
 
 
 class Copy(function_node.FunctionNode):
 
     """Copies the input variable onto the specified device."""
 
-    def __init__(self, out_device):
+    def __init__(self, in_device, out_device):
+        self._in_device = in_device
         self.out_device = out_device
 
     def check_type_forward(self, in_types):
@@ -15,16 +19,18 @@ class Copy(function_node.FunctionNode):
 
     def forward(self, inputs):
         x, = inputs
-        self._in_device = cuda.get_device_from_array(x).id
-        if int(self.out_device) == -1:
-            return cuda.to_cpu(x),
-        else:
-            return cuda.to_gpu(x, self.out_device),
+        return self.out_device.send(x),
+
+    def forward_chainerx(self, inputs):
+        x, = inputs
+        return x.to_device(self.out_device.device),
 
     def backward(self, indexes, grad_outputs):
-        return Copy(self._in_device).apply(grad_outputs)
+        f = Copy(self.out_device, self._in_device)
+        return f.apply(grad_outputs)
 
 
+# TODO(niboshi): Link from `dst` to an appropriate device specifier docs.
 def copy(x, dst):
     """Copies the input variable onto the specified device.
 
@@ -36,7 +42,7 @@ def copy(x, dst):
     Args:
         x (:class:`~chainer.Variable` or :ref:`ndarray`):
             Variable to be copied.
-        dst (int): Target device specifier.
+        dst: Target device specifier.
 
     Returns:
         ~chainer.Variable: Output variable.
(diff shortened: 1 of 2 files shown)


hvy:
A drop by comment but should we not implement `forward_chainerx` instead?

chainer/functions/array/copy.py
```diff
@@ -36,24 +41,38 @@ def copy(x, dst):
...
     """
-    y, = Copy(dst).apply((x,))
+    in_device = backend.get_device_from_array(
+        x.array if isinstance(x, chainer.Variable) else x)
+    out_device = chainer.get_device(dst)
+
+    is_chainerx = in_device.xp is chainerx
+    if is_chainerx != (out_device.xp is chainerx):
+        raise RuntimeError(
+            'F.copy does not support a copy between a non-ChainerX device and '
```
hvy: A major nitpick but I think the following is a more common way to put it.
```suggestion
            'F.copy does not support copies between non-ChainerX devices and '
```

chainer/functions/array/copy.py
```diff
@@ -36,24 +41,38 @@ def copy(x, dst):
...
+            'a ChainerX device.\n'
```
hvy: Following the comment above.
```suggestion
            'ChainerX devices.\n'
```

chainer/functions/array/copy.py
```diff
@@ -36,24 +41,38 @@ def copy(x, dst):
...
+        >>> x = chainer.Variable(x_arr)
+        >>> x.device
+        '@numpy'
+        >>> y = F.copy(x, '@cupy:0') # from CPU (NumPy) to GPU 0 (CuPy)
+        >>> y.device
+        '@cupy:0'
+
+    .. note::
+        Copies between non-ChainerX devices and ChainerX devices are not
+        supported.
```
hvy: Could you explain why it's not supported (maybe not in the documentation but in this PR)?

niboshi: Because it would be extremely difficult to support backprop.

chainer/functions/array/copy.py
niboshi: Thanks, fixed.

chainer/functions/array/copy.py
niboshi: Fixed.

tests/chainer_tests/functions_tests/array_tests/test_copy.py
```diff
@@ -18,97 +18,184 @@ def _to_gpu(x, device_id):
...
+        assert backend.get_device_from_array(x_grad) == src_device
+        numpy.testing.assert_array_equal(_numpy_device.send(x_grad), self.gy)
+
+    def test_double_backward(self, src_backend_config, dst_backend_config):
+        x = src_backend_config.get_array(self.x)
+        gy = dst_backend_config.get_array(self.gy)
+        ggx = src_backend_config.get_array(self.ggx)
+        dst_device = dst_backend_config.device
+
+        x_var = chainer.Variable(x, requires_grad=True)
```
hvy: Do we need the `requires_grad=True`? I think it increases the reader's load a bit since there's no such argument passed in `test_backward`.

tests/chainer_tests/functions_tests/array_tests/test_copy.py
```diff
@@ -18,97 +18,184 @@ def _to_gpu(x, device_id):
...
+
+        y_var = functions.copy(x_var, dst_device)
 
-        self.assertEqual(self.x_data.dtype, self.dtype)
-        numpy.testing.assert_array_equal(self.x_data, cuda.to_cpu(y.data))
+        # TODO(niboshi): Remove this workround after Variable.grad.setter is
+        # fixed so that it calls gy.require_grad() internally.
```
hvy: Can we address this TODO now by constructing a `Variable` gradient and call the `grad_var` setter instead (and preferably similarly in other gradient assignments in this file for consistency)?

niboshi: What's the point of your suggestion?
I think both ways are equally correct.

tests/chainer_tests/functions_tests/array_tests/test_copy.py
niboshi: `requires_grad=True` is essential in this test and it's better to make it explicit.

chainer/functions/array/copy.py
hvy: I see, it doesn't get along very well with the `forward_chainerx` forward implementation separation and the fallback mechanism I guess.

tests/chainer_tests/functions_tests/array_tests/test_copy.py
hvy: I think my suggestion is simpler since it's xp-agnostic but then I also meant we could remove the TODO comment since it's not a temporary workaround.

tests/chainer_tests/functions_tests/array_tests/test_copy.py
hvy: Is it not the case for `test_backward` ?

tests/chainer_tests/functions_tests/array_tests/test_copy.py
niboshi: I'm not persuaded.
Both ways are xp-agnostic and both are equally correct as I said.
It shouldn't matter whether we need TODO, because it doesn't mean it's a wrong way.

tests/chainer_tests/functions_tests/array_tests/test_copy.py
niboshi: It is.

tests/chainer_tests/functions_tests/array_tests/test_copy.py
hvy: Okay, let's leave it as is since the TODO will be addressed most likely and in that case `.grad` will be simpler.

hvy:
Could you take a look at the Travis failures?

niboshi:
#7145 is needed for `TestCopyCudaDummyDevice`.
(the behavior of `F.copy` is not changed in this PR).

niboshi:
Instead of relying on #7145, I fixed `F.copy` to handle `DummyDevice` in its own.

hvy:
Jenkins, test this please.

hvy:
LGTM.

hvy:
Jenkins, test this please.

Issue #6912, opened 2019-04-15
`test_math` is flaky


imos:
This is part of https://github.com/chainer/chainer/issues/6903 effort.

- Failure example: https://ci.preferred.jp/chainer.py37/2259/
- Failure test: chainerx_tests/unit_tests/routines_tests/test_math.py
- Failure test case: `test_Tan_param_17_{contiguous=None, in_dtypes=('float16',), out_dtype='float16', input=2, shape=(2, 3)}[native:0]`
- Tolerance:
```
E           Not equal to tolerance rtol=0.001, atol=0.001
E           
E           Mismatch: 100%
E           Max absolute difference: 0.00203618
E           Max relative difference: 0.00200944
```

asi1024:
Already fixed in #6798.

asi1024:
@imos Can I close this issue?

imos:
Sure. I'll recreate an issue if the same issue happens for the same test case.

imos:
It seems the test is still flaky in the master branch.

- Failure example: https://ci.preferred.jp/chainer.py37/2762/
- Failure test: chainerx_tests/unit_tests/routines_tests/test_math.py
- Failure test case: `test_LogSumExp_param_1_{keepdims=False}_param_1_{shape=(2,), axis=-1}_param_0_{in_dtypes=('float16',), out_dtype='float16'}[cuda:0]`
- Tolerance:
```
E       Not equal to tolerance rtol=0.001, atol=0.001
E       
E       Mismatch: 100%
E       Max absolute difference: 0.00116
E       Max relative difference: 0.011566
```

imos:
FYI, this is another example:

- Failure example: https://ci.preferred.jp/chainer.py37/2758/
- Failure test case: `test_Min_param_105_{in_dtypes=('float16',), out_dtype='float16', is_module=False, shape=(4, 2), axis=-2}[cuda:0]`
- Tolerance:
```
E           Not equal to tolerance rtol=0.01, atol=0.01
E           
E           Mismatch: 100%
E           Max absolute difference: 95.84712927
E           Max relative difference: 1179.51587459
```

asi1024:
- Failure example: https://travis-ci.org/chainer/chainer/jobs/522915728#L3603
- Failure test case: `test_Abs_param_24_{shape=(), contiguous='C', out_dtype='float32', input='random', in_dtypes=('float32',)}[native:0]`

PR #6988, opened 2019-04-18
Fix flaky `TestLogSumExp`

asi1024:
Part of #6912.

Currently, the pre-mapped array is casted to a float16 array before reduction. This loss of trailing digits seems avoidable by supporting reduction kernel for `LogSumExp` for each device.


niboshi:
Jenkins, test this please

niboshi:
LGTM

PR #6993, opened 2019-04-18
Use cuDNN in ReLU

okuta:
`cudnnActivationBackward` requests input layer data. Current ReLU implementation uses CuPy one.
But, In ReLU case we can pass output layer data as input layer data.

- Speed up with FP16. cuDNN method is faster than CuPy.
- Reduce memory consumption with cuDNN mode.


asi1024:
Jenkins, test this please.

asi1024:
Jenkins failure seems unrelated to this PR. LGTM.

PR #7003, opened 2019-04-19
Fix unit selection of CupyMemoryProfiler

kmaehashi:
Currently output unit of CupyMemoryProfiler always becomes bytes.
This issue is introduced in #6256.


chainer/function_hooks/cupy_memory_profile.py
```diff
@@ -129,10 +129,12 @@ def summary(self):
     def _choose_unit(self, size):
         """Choose optimal unit."""
         denomi = 1
+        if size <= 0:
+            return denomi, self._units[0]
```
toslunar: :+1:

chainer/function_hooks/cupy_memory_profile.py
```diff
@@ -129,10 +129,12 @@ def summary(self):
...
         for unit in self._units[:-1]:
-            if size / denomi >= 1:
+            if size / (denomi * 1024.0) < 1:
```
toslunar: How about fixing the previous line?
`for unit in self._units:`

chainer/function_hooks/cupy_memory_profile.py
```diff
@@ -129,10 +129,12 @@ def summary(self):
...
                 return denomi, unit
-            denomi *= 1024.0
+            denomi *= 1024
```
toslunar: It seems `from __future__ import division` will be needed (L170--172).

chainer/function_hooks/cupy_memory_profile.py
kmaehashi: Sorry, I couldn't get your point. Will that simplify this code?

chainer/function_hooks/cupy_memory_profile.py
kmaehashi: Thanks, changed to always return denomi as float, as it was the original spec.

chainer/function_hooks/cupy_memory_profile.py
toslunar: I'm sorry.  I was confused with `self._units[::-1]`.

toslunar:
CIs, test this please.

kmaehashi:
The error is not related to this PR.

kmaehashi:
pfnCI, test this please.

PR #7016, opened 2019-04-21
[chainerx] add `isfinite`

kshitij12345:
References : #6423

### DIFF
--- a/chainerx/__init__.pyi
+++ b/chainerx/__init__.pyi
@@ -660,6 +660,9 @@ def identity(
 def is_available(): ...
 
 
+def isfinite(x: ndarray) -> ndarray: ...
+
+
 def isinf(x: ndarray) -> ndarray: ...
 
 
--- a/chainerx/_docs/routines.py
+++ b/chainerx/_docs/routines.py
@@ -1391,6 +1391,24 @@ Note:
 """)
 
     _docs.set_doc(
+        chainerx.isfinite,
+        """isfinite(x)
+Test element-wise for finiteness (not infinity or not Not a Number).
+
+Args:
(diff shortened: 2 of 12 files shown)


hvy:
I'm wondering if we should skip introducing the `numeric.{h,cuh}` `IsFinite` since we can reuse existing ones in the kernels. There should be no difference in performance and it's less to maintain. What do you say?

hvy:
It otherwise LGTM.

kshitij12345:
> I'm wondering if we should skip introducing the `numeric.{h,cuh}` `IsFinite` since we can reuse existing ones in the kernels. There should be no difference in performance and it's less to maintain. What do you say?

My reasoning was that generally these functions have their peculiarities so I feel that it is safer to wrap library implementations. But if you feel that, it would be better to use existing ones then I will remove the definitions in `numeric`.

hvy:
Thanks for the quick response. Do you have any concrete cases where either would behave unexpectedly? Wrapping is more robust to future changes but as for now, both approaches are basically logically identical.

hvy:
Jenkins, test this please.

kshitij12345:
> Do you have any concrete cases where either would behave unexpectedly? Wrapping is more robust to future changes but as for now, both approaches are basically logically identical.

I guess, `isfinite` is not as hairy , I don't think there are any cases that this would misbehave.

kshitij12345:
Do let me know, if anything is remaining. I tried to find is `isfinite` in `chainer.functions` but couldn't find it, so I assume we don't need to overload this one.

hvy:
Then it LGTM 👍

PR #7017, opened 2019-04-21
[chainerx] add bitwise ops

kshitij12345:
References #6423 

To-Do:
* ~~In-place Version.~~
* ~~Documentation.~~
* ~~Array -Scalar~~
* ~~Array-Scalar-Inplace-Test~~


asi1024:
These routines look not support broadcasts. Could you add tests for broadcast after #7023 is merged?

kshitij12345:
Ah, right thanks for pointing that out.

chainerx_cc/chainerx/routines/math.cc
```diff
@@ -1110,4 +1110,30 @@ Array IsInf(const Array& x) {
...
 }
 
+template <typename Impl>
+inline Array BitwiseImpl(Impl&& impl, const Array& x1, const Array& x2) {
+    CHAINERX_ASSERT(GetKind(x1.dtype()) != DtypeKind::kFloat);
+    CHAINERX_ASSERT(GetKind(x2.dtype()) != DtypeKind::kFloat);
+    // promote to which ever is bigger
+    Array x1_cast = x1.dtype() < x2.dtype() ? x1.AsType(x2.dtype()) : x1;
+    Array x2_cast = x2.dtype() < x1.dtype() ? x2.AsType(x1.dtype()) : x2;
+    Dtype out_dtype = x1_cast.dtype();
```
asi1024: We can write it simpler as follows:
```
Dtype out_dtype = std::max(x1_dtype(), x2_dtype());
const Array& x1_cast = ...
const Array& x2_cast = ...
```
However, I think it is confusing that the dtype conversion rule for bitwise operations is different from that for other arithmetic operations.

asi1024: I discussed it with a team member and we decide to use the same dtype rule as other arithmetic operations. Could you use `GetArithmeticResultDtype`?

kshitij12345: Oh sure.

kshitij12345:
@asi1024 is it ok if we do Array-Vs-Scalar in other PR?

asi1024:
Yes, please!

kshitij12345:
Actually on second thought, adding Array-Scalar in this PR itself.

chainerx_cc/chainerx/routines/math.cc
```diff
@@ -1140,4 +1140,84 @@ Array IsFinite(const Array& x) {
     return out;
 }
 
+template <typename Impl>
+inline Array BitwiseImpl(Impl&& impl, const Array& x1, const Array& x2) {
+    CHAINERX_ASSERT(GetKind(x1.dtype()) != DtypeKind::kFloat);
```
asi1024: Could you raise `DtypeError`?
`CHAINERX_ASSERT` is used only for the test, and does not expose errors to Python.

chainerx_cc/chainerx/routines/math.cc
```diff
@@ -1140,4 +1140,84 @@ Array IsFinite(const Array& x) {
...
+
+template <typename Kernel>
+inline void ApplyBitwiseImpl(const Array& x1, const Array& x2, const Array& out) {
+    NoBackpropModeScope scope;
+    CheckEqual(x1.shape(), x2.shape());
+    x1.device().backend().CallKernel<Kernel>(x1, x2, out);
+}
+
+template <typename Kernel>
+inline void ApplyBitwiseASImpl(const Array& x1, Scalar x2, const Array& out) {
```
asi1024: Could you check dtypes?

tests/chainerx_tests/unit_tests/routines_tests/test_math.py
```diff
@@ -1349,6 +1349,341 @@ def func_scalar(self, xp, a, scalar):
         a /= scalar
 
 
+_bwise_in_out_dtypes = dtype_utils._permutate_dtype_mapping([
```
asi1024: How about naming `_in_out_dtypes_bitwise` and so on for the consistency?

asi1024: Could you add cases for un-mixed dtypes?

tests/chainerx_tests/unit_tests/routines_tests/test_math.py
```diff
@@ -1349,6 +1349,341 @@ def func_scalar(self, xp, a, scalar):
...
+    (('float64',), numpy.int8, 'float64'),
+    (('float16',), numpy.int64, 'float16'),
+]
+
+_bwise_in_out_scalar_dtypes = [
+    dtypes for dtypes in _in_out_dtypes_array_int_scalar
+    if dtypes not in _bwise_scalar_invalid
+]
+
+_bitwise_params = (
```
asi1024: Could you merge master and check if broadcast works well?

tests/chainerx_tests/unit_tests/routines_tests/test_math.py
kshitij12345: https://github.com/chainer/chainer/blob/7ee1103ffcc002aa45749befc2a39f6f99e3345f/tests/chainerx_tests/unit_tests/routines_tests/test_math.py#L1427

un-mixed dtypes are used here, should we add them here?

tests/chainerx_tests/unit_tests/routines_tests/test_math.py
kshitij12345: If I am correct broadcasting is already checked using,

https://github.com/chainer/chainer/blob/7ee1103ffcc002aa45749befc2a39f6f99e3345f/tests/chainerx_tests/unit_tests/routines_tests/test_math.py#L295-L329

chainerx/__init__.pyi
```diff
@@ -290,8 +292,14 @@ class ndarray:
 
     def __iadd__(self, arg0: tp.Any) -> ndarray: ...
 
+    def __iand__(self, arg0: tp.Any) -> ndarray: ...
+
     def __imul__(self, arg0: tp.Any) -> ndarray: ...
 
+    def __ior__(self, arg0: tp.Any) -> ndarray: ...
+
+    def __ixor__(self, arg0: tp.Any) -> ndarray: ...
```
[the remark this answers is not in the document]
kshitij12345: Oop, sure will.

chainerx/_docs/routines.py
```diff
@@ -1444,6 +1444,60 @@ def _docs_math():
 .. seealso:: :data:`numpy.isinf`
 """)
 
+    _docs.set_doc(
+        chainerx.bitwise_and,
+        """bitwise_and(x1, x2)
+Compute the bit-wise AND of two arrays element-wise.
+
+Args:
+    x1 (~chainerx.ndarray or scalar): Input array.
```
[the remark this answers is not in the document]
kshitij12345: Sure.

chainerx_cc/chainerx/routines/math.cc
kshitij12345: Oh, sure

chainerx_cc/chainerx/routines/math.cc
kshitij12345: Sure

tests/chainerx_tests/unit_tests/routines_tests/test_math.py
asi1024: Sorry, I overlooked.

tests/chainerx_tests/unit_tests/routines_tests/test_math.py
asi1024: The case in your comment is for in-place version. Isn't it needed also in `_in_out_dtypes_bitwise`?

kshitij12345: The line in the comment is used for inplace as well as usual.

https://github.com/chainer/chainer/blob/7ee1103ffcc002aa45749befc2a39f6f99e3345f/tests/chainerx_tests/unit_tests/routines_tests/test_math.py#L1397

https://github.com/chainer/chainer/blob/7ee1103ffcc002aa45749befc2a39f6f99e3345f/tests/chainerx_tests/unit_tests/routines_tests/test_math.py#L1427

So while checking broadcastable shapes, we use same `dtype` i.e. un-mixed dtypes for both the input. While we check mixed dtypes, 

https://github.com/chainer/chainer/blob/7ee1103ffcc002aa45749befc2a39f6f99e3345f/tests/chainerx_tests/unit_tests/routines_tests/test_math.py#L1405

https://github.com/chainer/chainer/blob/7ee1103ffcc002aa45749befc2a39f6f99e3345f/tests/chainerx_tests/unit_tests/routines_tests/test_math.py#L1434

asi1024: Certainly, the whole `_in_out_dtypes_bitwise` covers all cases for un-mixed and mixed dtypes. However, L1397 is the test for broadcast, and not for same dtypes. L1405 is the test for all dtypes combination, so it seems better to include not only mixed cases but also un-mixed cases.

kshitij12345: Oh. Sure. On it.

kshitij12345: Done. I have also added cases for `uint8` and `{int8, int16, int32, int64}`

asi1024:
Jenkins, test this please.

asi1024:
Jenkins failure seems unrelated to this PR. LGTM!

PR #7027, opened 2019-04-22
Support Sphinx 2.0 and use absolute path to support the latest RTD

kmaehashi:
* In Sphinx v2, `Sphinx.warn` has been removed and https://www.sphinx-doc.org/ja/master/extdev/logging.html API should be used instead. Sphinx v2 is used in recent 
 RTD projects configured to use Conda.
* Inside the RTD container, the current directory is `docs/source`, not under `docs/`. Fixed `_autosummary_check` to avoid using relative directory.


toslunar:
> * Inside the RTD container, the current directory is the repo root, not under `docs/`. Fixed `_autosummary_check` to avoid using relative directory.

As far as I read the docs of RTD, the build process runs `sphinx-build -b html . _build/html` after it finds `conf.py`.  Did you mean RTD runs `sphinx-build -b html docs/source _build/html` under root?  It's possible that RTD runs `sphinx-build -b html . _build/html` under `docs/source`, which also fails with the previous `_autosummary_check`.

kmaehashi:
Resolved conflict.

> As far as I read the docs of RTD, the build process runs `sphinx-build -b html . _build/html` after it finds `conf.py`.  Did you mean RTD runs `sphinx-build -b html docs/source _build/html` under root?  It's possible that RTD runs `sphinx-build -b html . _build/html` under `docs/source`, which also fails with the previous `_autosummary_check`.

Ah yes, I was wrong. The current directory is `docs/source` and it was not working from the beginning.

toslunar:
Jenkins, test this please.

PR #7032, opened 2019-04-22
[chainerx] Adds vsplit

ishanrai05:
Solves part of issue #6423

### DIFF
--- a/chainerx_cc/chainerx/routines/manipulation.cc
+++ b/chainerx_cc/chainerx/routines/manipulation.cc
@@ -633,6 +633,26 @@ std::vector<Array> DSplit(const Array& ary, std::vector<int64_t> indices) {
     return Split(ary, std::move(indices), 2);
 }
 
+std::vector<Array> VSplit(const Array& ary, int64_t sections) {
+    if (sections < 1) {
+        throw DimensionError("Number of sections must be larger than 0.");
+    }
+
+    if (ary.ndim() < 2) {
+        throw DimensionError("vsplit only works on arrays of 2 or more dimensions.");
+    }
+
+    return Split(ary, sections, 0);
+}
+
+std::vector<Array> VSplit(const Array& ary, std::vector<int64_t> indices) {
+    if (ary.ndim() < 2) {
+        throw DimensionError("vsplit only works on arrays of 2 or more dimensions.");
+    }
+
+    return Split(ary, std::move(indices), 0);
+}
+
 Array Swapaxes(const Array& a, int8_t axis1, int8_t axis2) {
     Shape shape = a.shape();
     Strides strides = a.strides();
--- a/tests/chainerx_tests/unit_tests/routines_tests/test_manipulation.py
+++ b/tests/chainerx_tests/unit_tests/routines_tests/test_manipulation.py
@@ -773,6 +773,47 @@ class TestDSplit(op_utils.NumpyOpTest):
 
 
 @op_utils.op_test(['native:0', 'cuda:0'])
+@chainer.testing.parameterize_pytest('shape,indices_or_sections', [
+    ((6, 4, 2), [1, 2, 4]),
+    ((6, 4, 2), [2, 4, 6]),
+    ((6, 4, 2), [1, 5, 7]),
+    ((6, 4, 2), [2, -3, -5]),
+    ((6, 4, 2), [2, 8, 10]),
+    ((8, 6, 4, 2), [2, 4, 4, 6]),
+    ((8, 6, 4, 2), [2, 6, 6, 8]),
+    ((8, 6, 4, 2), [1, 4, 4, 6]),
+    ((8, 6, 4, 2), [1, 4, 5, 6]),
+    ((8, 6, 6, 4, 4, 2), [1, 3, 5, 7]),
+    # indices with 1-d numpy array
+    ((8, 6, 4, 2), numpy.array([1, 2, 3])),
+    # indices with (4,)-shape numpy array
+    ((8, 6, 4, 2), numpy.array([1, 2, 3, 4])),
+    ((8, 8, 6, 6, 4, 2), numpy.array([1, 2, 3], numpy.int32)),
+])
+class TestVSplit(op_utils.NumpyOpTest):
+
+    def setup(self):
+        # TODO(ishanrai05): There's a bug in backward of split() in which the
+        # gradient shape differs from the input if indices are not in the
+        # sorted order. Fix this.
+        indices_or_sections = self.indices_or_sections
+        if (isinstance(indices_or_sections, list) and
+                sorted(indices_or_sections) != indices_or_sections):
+            self.skip_backward_test = True
+            self.skip_double_backward_test = True
+
+    def generate_inputs(self):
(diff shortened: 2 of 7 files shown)


hvy:
Hm, I'd like to reduce the code duplication. Do you have any ideas how, or I can maybe help you out.

ishanrai05:
@hvy Yes. I was thinking if we could reduce the code duplication in the python binding part. I already used the predefined Split function to reduce some of the code in chainerx implementation. 
I'd like to do it myself, if you guide me. :)

PS: I am really sorry for the late reply. I was busy with end sems.

hvy:
How about simply creating a free function that takes `split_sections` and `split_indices` as arguments which wraps the logic of the currently duplicated code. It would of course also need the arguments `ary` and `indices_or_sections`.

hvy:
Ah, yes in the binding. Sorry I wasn't clear enough.

ishanrai05:
@hvy Hi. I think this is ready to be merged now. Can you take a look?

chainerx_cc/chainerx/routines/manipulation.cc
```diff
@@ -633,6 +633,26 @@ std::vector<Array> DSplit(const Array& ary, std::vector<int64_t> indices) {
     return Split(ary, std::move(indices), 2);
 }
 
+std::vector<Array> VSplit(const Array& ary, int64_t sections) {
+    if (sections < 1) {
+        throw DimensionError("Number of sections must be larger than 0.");
+    }
+
+    if (ary.ndim() < 2) {
+        throw DimensionError("vsplit only works on arrays of 2 or more dimensions.");
```
hvy: vsplit should work with 1-dimensional arrays also (since it's split over axis 0). Could you do that?

chainerx_cc/chainerx/routines/manipulation.cc
```diff
@@ -633,6 +633,26 @@ std::vector<Array> DSplit(const Array& ary, std::vector<int64_t> indices) {
...
+    }
+
+    return Split(ary, sections, 0);
+}
+
+std::vector<Array> VSplit(const Array& ary, std::vector<int64_t> indices) {
+    if (ary.ndim() < 2) {
+        throw DimensionError("vsplit only works on arrays of 3 or more dimensions.");
```
hvy: Ditto. The error message seems inconsistent with the logic by the way so it'd be good if you could fix it simultaneously.

tests/chainerx_tests/unit_tests/routines_tests/test_manipulation.py
```diff
@@ -772,6 +772,49 @@ def forward_xp(self, inputs, xp):
...
+    ((2, 4, 6, 8), numpy.array(2.0)),  # sections with numpy scalar, float
+    ((2, 4, 6, 8), 2.0),  # float type sections, without fraction
+    # indices with empty numpy indices
+    ((2, 4, 8, 10), numpy.array([], numpy.int32)),
+    ((2, 4, 5, 10), numpy.array([], numpy.float64)),
+])
+class TestVSplit(op_utils.NumpyOpTest):
+
+    def setup(self):
+        # TODO(niboshi): There's a bug in backward of split() in which the
```
hvy: Although copied, I prefer having the actual author's name before the comment (i.e. niboshi is not responsible for these lines of code). Could you write your own?

hvy:
Basically LGTM but I left some smaller comments. Can you take a look at them?

chainerx_cc/chainerx/routines/manipulation.cc
ishanrai05: The numpy implementation of [vsplit](https://github.com/numpy/numpy/blob/v1.16.1/numpy/lib/shape_base.py#L922-L972
) says it works with tow or more than two dimensions.

ishanrai05:
@hvy  ...

chainerx_cc/chainerx/routines/manipulation.cc
hvy: Sorry you're right. Your current implementation looks correct.

tests/chainerx_tests/unit_tests/routines_tests/test_manipulation.py
```diff
@@ -772,6 +772,49 @@ def forward_xp(self, inputs, xp):
...
+    ((2, 4, 8), [2, 5]),
+    ((2, 4, 10), [1, 4]),
+    ((2, 6), numpy.array([1, 2])),  # indices with 1-d numpy array
+    ((2, 6), numpy.array([2])),  # indices with (1,)-shape numpy array
+    ((2, 4, 8), numpy.array(2)),  # sections numpy scalar
+    ((2, 4, 6, 8), numpy.array(2.0)),  # sections with numpy scalar, float
+    ((2, 4, 6, 8), 2.0),  # float type sections, without fraction
+    # indices with empty numpy indices
+    ((2, 4, 8, 10), numpy.array([], numpy.int32)),
+    ((2, 4, 5, 10), numpy.array([], numpy.float64)),
```
hvy: I find it acceptable to reuse the params from dsplit but we should make sure that the dimension along which this routine splits has appropriate length (i.e. it should be larger than 2 w.r.t. to `indices_or_sections`).

hvy:
Sorry, I did not notice that this PR was ready for another review. I added a single comment if you could take a look.

hvy:
Could you make similar tests to dsplit? More specifically, the dimension that is split, only has lengths of 2. You see that the shapes for the dsplit have larger length axes over which they are split.

hvy:
Sorry if I was unclear, I meant you latter. The length of this dimension should be greater than 2.

tests/chainerx_tests/unit_tests/routines_tests/test_manipulation.py
```diff
@@ -772,6 +772,44 @@ def forward_xp(self, inputs, xp):
...
+    ((2, 4, 6), [2, -3, -5]),
+    ((2, 4, 6), [2, 8, 10]),
+    ((2, 4, 6), [4, 2, 6]),
+    ((2, 4, 6), [1, 3, 5]),
+    ((2, 4, 5), [1, -3, -5]),
+    ((2, 4, 8), [2, 5, 7, 6]),
+    ((2, 4, 10), [1, 4, 6, 8]),
+    ((2, 4, 6), numpy.array([1, 2, 3])),  # indices with 1-d numpy array
+    ((2, 4, 6, 8), numpy.array([1, 2, 3, 4])),  # indices with (4,)-shape numpy array
+    ((2, 4, 8, 10), numpy.array([1, 2, 3], numpy.int32)),
```
hvy: Can you change the length of the first dimension (which vsplit splits over) to something larger, e.g. 6, 8 or 10 depending on the `indices_or_sections`?

ishanrai05: @hvy don't you think that will slow down the tests?

hvy: How about reducing the lengths of the dimensions that are not of interest?

hvy:
Thanks, LGTM! Let me verify the changes with CI. Jenkins, test this please.

PR #6621, opened 2019-03-25
Allow backprop on disconnected arrays 

niboshi:
~Fixes #6612~  Fixes #8069

~Depends on #6658 and #6659.~

PR #7034, opened 2019-04-23
Simplify `F.cast` test

hvy:
Part of https://github.com/chainer/chainer/issues/6071.

~Depends on a change in gradient check in #6621 . In particular https://github.com/chainer/chainer/pull/6621/files#diff-643267efb2f2b096c96a529d2818281dR1016~

### DIFF
--- a/chainer/gradient_check.py
+++ b/chainer/gradient_check.py
@@ -968,6 +968,14 @@ def check_double_backward(func, x_data, y_grad, x_grad_grad, params=(),
         gys = inputs[n_x:]
 
         ys = _as_tuple(func(*xs))
+
+        # `gys` (inputs to `first_order_grad` forward function) may have been
+        # casted to float64 by `numerical_grad`. For certain functions demoting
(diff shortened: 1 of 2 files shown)


Crissman:
@hvy As I read #6071, this is no longer blocked by that PR, and ok to proceed with review and merge, correct?

hvy:
I think the CI will fail, but haven't tested it for a while. It should depend on a change in the gradient check.

hvy:
This PR is no longer blocked and ready for review, PTAL. Since the PR on which this PR used to depend was closed, I simply included the necessary changes in this PR instead.

Crissman:
Jenkins, test this please.

Crissman:
ImosCI has the same issue with `chainerx.testing`. Do we need to fix imosCI?

hvy:
Hm, could you reproduce the same failure locally? It seems to fail only for certain environments.

Crissman:
Jenkins, test this please.

Crissman:
Travis issue looks unrelated.

PR #7037, opened 2019-04-23
Fix subview of zero-sized arrays

hvy:
A sub-view of a zero-sized array (e.g. a sub-view of shape `(2, 0)` of an array one that has a shape of `(7. 0)`) should always have an offset of 0. Having a non-zero offset would yield an invalid address when for instance creating a new array based on it.

In short, this fix was missing for ChainerX.
https://github.com/cupy/cupy/blob/master/cupy/core/_routines_indexing.pyx#L322-L323

~_TODO: Need to add tests._~ Done.

### DIFF
--- a/chainerx_cc/chainerx/routines/indexing.cc
+++ b/chainerx_cc/chainerx/routines/indexing.cc
@@ -103,6 +103,11 @@ Array At(const Array& a, const std::vector<ArrayIndex>& indices) {
         out_strides.emplace_back(a.strides()[i]);
     }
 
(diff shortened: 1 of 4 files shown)


asi1024:
Jenkins, test this please.

hvy:
Jenkins, test this please.

asi1024:
Jenkins, test this please.

hvy:
Jenkins, test this please.

asi1024:
LGTM.

PR #7045, opened 2019-04-25
`forward_chainerx` for `F.cast`

hvy:
`F.cast` to use utilize `chainerx.astype` instead of falling back to NumPy/CuPy for ChainerX inputs.

~Depends on https://github.com/chainer/chainer/pull/7034 (that adds tests).~


hvy:
#7034 is not merged and this PR is ready for review.

PR #7046, opened 2019-04-25
Fix `At` output offset

hvy:
Fixed offset calculation for `At`. The current implementation is similar to getitem of CuPy https://github.com/cupy/cupy/blob/master/cupy/core/_routines_indexing.pyx#L267.


asi1024:
Jenkins, test this please.

asi1024:
LGTM.

PR #7047, opened 2019-04-25
Fix handling of ndarray offsets

hvy:
When converting a ChainerX array into a CuPy array, the base pointer was incremented by the offset. This PR changes the behavior such that the base pointer is kept and the offset is specified via the "offset" parameter.

### DIFF
--- a/chainerx_cc/chainerx/python/array.cc
+++ b/chainerx_cc/chainerx/python/array.cc
@@ -99,7 +99,7 @@ py::object MakeCupyArrayFromArray(const py::module& m, py::handle self) {
     const Strides& strides = array.strides();
 
     // NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast)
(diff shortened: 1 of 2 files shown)


okapies:
Jenkins, test this please.

hvy:
Jenkins, test this please.

hvy:
Jenkins, test this please.

hvy:
Fixed an error in `Stack` and `StackGrad` that became present due to this change (they've existed since before).

okapies:
LGTM. I think they can be solved in the same PR because they are related to the handling of `offset`. Just rename the title is enough I think.

hvy:
Okay, I'll do so.

hvy:
Fixed the name to be a bit more "general".

hvy:
Should be backported but after https://github.com/chainer/chainer/pull/7046.

asi1024:
LGTM.

asi1024:
I will create a backport PR after #7054 is merged.

PR #7058, opened 2019-04-26
Support mixed dtypes in `Stack`

hvy:
Fixes https://github.com/chainer/chainer/issues/7050.

Also, simplified the backward of `Stack` to use `Split` and removed `StackGrad`.

### DIFF
--- a/chainerx_cc/chainerx/routines/manipulation.cc
+++ b/chainerx_cc/chainerx/routines/manipulation.cc
@@ -492,53 +492,6 @@ Array Concatenate(const std::vector<Array>& arrays, nonstd::optional<int8_t> axi
     return ConcatenateImpl(raveled_arrays, 0);
 }
 
-namespace {
-std::vector<Array> StackGrad(const Array& gout, int8_t axis) {
-    Shape shape{gout.shape()};
-    Strides strides{gout.strides()};
-    size_t dim = shape[axis];
-    int64_t step = strides[axis];
-    int64_t offset = gout.offset();
-    bool is_empty = gout.GetTotalSize() == 0;
-    shape.erase(shape.begin() + axis);
-    strides.erase(strides.begin() + axis);
-
-    std::vector<Array> gxs;
-    std::vector<ConstArrayRef> gxs_refs{};
-    gxs.reserve(dim);
-    gxs_refs.reserve(dim);
-    {
-        NoBackpropModeScope scope{};
-        Dtype dtype = gout.dtype();
-        Device& device = gout.device();
-        for (size_t i = 0; i < dim; ++i) {
-            gxs.emplace_back(internal::MakeArray(shape, strides, dtype, device, gout.data(), offset));
-            gxs_refs.emplace_back(gxs.back());
-            if (!is_empty) {
-                offset += step;
-            }
-        }
-    }
-
-    {
-        BackwardBuilder bb{"stack-grad", gout, gxs_refs};
-        if (BackwardBuilder::Target bt = bb.CreateTarget()) {
-            bt.Define([axis](BackwardContext& bctx) {
-                std::vector<Array> ggxs;
-                ggxs.reserve(bctx.output_count());
-                for (size_t i = 0; i < bctx.output_count(); ++i) {
-                    // TODO(imanishi): Check if bctx.output_grad(i) is not nullopt.
-                    ggxs.emplace_back(*bctx.output_grad(i));
-                }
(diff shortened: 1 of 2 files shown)


chainerx_cc/chainerx/routines/manipulation.cc
```diff
@@ -555,32 +508,38 @@ Array Stack(const std::vector<Array>& arrays, int8_t axis) {
...
-            int64_t out_offset = 0;
-            for (const Array& array : arrays) {
-                Array sliced_out = internal::MakeArray(array.shape(), strides, dtype, device, out.data(), out_offset);
+    int64_t step = strides[axis];
+    strides.erase(strides.begin() + axis);
+    {
+        NoBackpropModeScope scope{};
+        int64_t out_offset = 0;
+        for (const Array& array : arrays) {
+            if (out.GetTotalSize() != 0) {
```
hvy: Moved the condition to inside the loop since we need to loop over inputs to collect their dtypes for the backward anyway.

chainerx_cc/chainerx/routines/manipulation.cc
```diff
@@ -591,11 +550,18 @@ Array Stack(const std::vector<Array>& arrays, int8_t axis) {
     {
         BackwardBuilder bb{"stack", array_refs, out};
         if (BackwardBuilder::Target bt = bb.CreateTarget()) {
-            bt.Define([axis](BackwardContext& bctx) {
+            bt.Define([axis, in_shape = array.shape(), in_dtypes = std::move(in_dtypes)](BackwardContext& bctx) {
                 const Array& gout = *bctx.output_grad();
-                std::vector<Array> gxs = StackGrad(gout, axis);
+                Dtype gout_dtype = gout.dtype();
+                std::vector<Array> gxs = Split(gout, gout.shape()[axis], axis);
```
hvy: Simpler to reuse `Split` for the backward computation instead of defining a new `StackGrad`.

tests/chainerx_tests/unit_tests/routines_tests/test_manipulation.py
```diff
@@ -431,15 +431,28 @@ def setup(self):
     def generate_inputs(self):
         return _make_inputs(self.shapes, self.dtypes)
 
+    def join(self, inputs, xp):
+        # Calls and returns the result of joining routines e.g. xp.concatenate,
+        # xp.stack and xp.vstack.
+        raise NotImplementedError()
```
hvy: Generalized `join` so that it can be used for `vstack` and `hstack` which have different parameters from `concatenate` and `stack`.

chainerx_cc/chainerx/routines/manipulation.cc
```diff
@@ -492,61 +492,14 @@ Array Concatenate(const std::vector<Array>& arrays, nonstd::optional<int8_t> axi
...
-                bctx.input_grad() = Stack(ggxs, axis);
-            });
-        }
-        bb.Finalize();
-    }
-    return gxs;
-}
-}  // namespace
-
 Array Stack(const std::vector<Array>& arrays, int8_t axis) {
```
asi1024: How about just using `Concatenate`?

hvy: How about addressing that in a separate PR? I don't necessarily think that it's within the scope of this PR. Although I did refactor the gradient computation in a similar way, but that was mostly to simply the mixed dtype support there.

hvy: Fixed, PTAL.

chainerx_cc/chainerx/routines/manipulation.cc
```diff
@@ -480,129 +480,26 @@ Array ConcatenateImpl(const std::vector<Array>& arrays, int8_t axis) {
...
-                ggxs.reserve(bctx.output_count());
-                for (size_t i = 0; i < bctx.output_count(); ++i) {
-                    // TODO(imanishi): Check if bctx.output_grad(i) is not nullopt.
-                    ggxs.emplace_back(*bctx.output_grad(i));
-                }
-                bctx.input_grad() = Stack(ggxs, axis);
-            });
-        }
-        bb.Finalize();
+    if (!axis.has_value()) {
```
hvy: Just reversed the condition (logic inside the indented scope) since the case where no axis is given is the "special" case.

asi1024:
Jenkins, test this please.

asi1024:
Jenkins failure seems unrelated to this PR. LGTM.

PR #7063, opened 2019-04-29
[chainerx] add basic loss functions

kshitij12345:
Add basic loss functions

Not sure how to go about writing test. Should we use prebuilt chainer functions to compare or write numpy versions of the same. I believe, first would be better option as they are well-tested already.


asi1024:
Thank you for the PR! I think that the first seems better too. It would be good if you use `op_utils.ChainerOpTest`.

kshitij12345:
Sure will try it that way.

chainerx_cc/chainerx/routines/loss.cc
```diff
@@ -0,0 +1,59 @@
...
+#include "chainerx/routines/creation.h"
+#include "chainerx/routines/math.h"
+#include "chainerx/routines/type_util.h"
+#include "chainerx/scalar.h"
+#include "chainerx/shape.h"
+#include "chainerx/strides.h"
+
+namespace chainerx {
+
+Array MeanAbsoluteError(const Array& y, const Array& targ) {
```
hvy: Are the variables names of the arguments following come convention? I'm wondering if we shouldn't stick to the ones similar to Chainer, i.e. `x0` and `x1`.

kshitij12345: Have renamed them as of now to be similar to `chainer`. However, from what I have seen in other frameworks, its generally `{y_pred, predictions, input}` and `{y_true, labels, target}` which I guess are better given that the function is used for calculating loss. Your thoughts?

hvy:
Having the reduction argument in Chainer is mostly due to a historical reason. I briefly discussed it with the team but we thought it might be better simplifying the interfaces by simple not taking any reduction option as arguments (as long as those functions can still be defined). Could you change it that way (it should be a lot simpler)?

chainerx_cc/chainerx/routines/loss.cc
hvy: For routines that don't depend on the order such as MSE, let's stick with `x1`, `x2`. In cases the order matters, `x` and `t` is a convention quite often used by Chainer and we should stick to that one if possible. I see the point in having more verbose names too but it should be fine for common functions such as the ones in this PR.

chainerx_cc/chainerx/python/routines.cc
```diff
@@ -1002,13 +1003,46 @@ void InitChainerxPooling(pybind11::module& m) {
           py::arg("pad_mode") = "ignore");
 }
 
+void InitChainerxLoss(pybind11::module& m) {
+    m.def("mean_absolute_error",
+          [](const ArrayBodyPtr& x0, const ArrayBodyPtr& x1) { return MoveArrayBody(MeanAbsoluteError(Array{x0}, Array{x1})); },
+          py::arg("x0"),
+          py::arg("x1"));
```
hvy: Could you start with 1 instead of 2 (similar to other ChainerX routines)?

chainerx_cc/chainerx/python/routines.cc
```diff
@@ -1002,13 +1003,46 @@ void InitChainerxPooling(pybind11::module& m) {
...
+    m.def("mean_squared_error",
+          [](const ArrayBodyPtr& x0, const ArrayBodyPtr& x1) { return MoveArrayBody(MeanSquaredError(Array{x0}, Array{x1})); },
+          py::arg("x0"),
+          py::arg("x1"));
```
hvy: Ditto.

chainerx_cc/chainerx/routines/loss.cc
```diff
@@ -0,0 +1,68 @@
...
+    if (reduce == "sum") {
+        return loss.Sum();
+    } else if (reduce == "mean") {
+        return loss.Mean();
+    } else {
+        return loss;
+    }
+}
+
+Array HuberLoss(const Array& x, const Array& t, Scalar delta, const std::string& reduce) {
```
hvy: Let's rename the arguments to `X1` and `x2`.

chainerx_cc/chainerx/python/routines.cc
kshitij12345: Instead of 0 right? `x1` and `x2`?

kshitij12345:
> Having the reduction argument in Chainer is mostly due to a historical reason. I briefly discussed it with the team but we thought it might be better simplifying the interfaces by simple not taking any reduction option as arguments (as long as those functions can still be defined). Could you change it that way (it should be a lot simpler)?

What would be the default behaviour in that case? The case with default value of `reduction`?

chainerx_cc/chainerx/python/routines.cc
hvy: Yes!

hvy:
Sorry my point was that these functions don't perform any reduction at all.

kshitij12345:
> Sorry my point was that these functions don't perform any reduction at all.

Sorry I am a bit confused, `gaussian_kl_divergence` does perform reduction right?

hvy:
`gaussian_kl_divergence` is element-wise up to this point https://github.com/chainer/chainer/blob/master/chainer/functions/loss/vae.py#L58-L61 (referencing Chainer's code but you see the algorithm). My suggestion is to skip this part entirely in the ChainerX version of it, just returning the element-wise loss. It'll be up to the caller to take the sum or the mean.

chainerx_cc/chainerx/routines/loss.cc
```diff
@@ -0,0 +1,55 @@
...
+
+namespace chainerx {
+
+Array MeanAbsoluteError(const Array& x1, const Array& x2) { return Fabs(x1 - x2).Mean(); }
+
+Array MeanSquaredError(const Array& x1, const Array& x2) { return SquaredDifference(x1, x2).Mean(); }
+
+Array GaussianKLDivergence(const Array& mu, const Array& ln_var) { return (Square(mu) + Exp(ln_var) - ln_var - 1) * 0.5; }
+
+Array HuberLoss(const Array& x1, const Array& x2, Scalar delta, const std::string& reduce) {
```
hvy: Could you remove the reduction argument from `HuberLoss` similar to `GaussianKLDivergence`?

hvy:
Jenkins, test this please.

kshitij12345:
@hvy do note that `huber_loss` is quite flaky.

hvy:
Do you know the cause and for what dtypes is it flaky?

kshitij12345:
Not exactly certain but it might be related to https://github.com/chainer/chainer/issues/7241. Will try to dig into it. This week.

kshitij12345:
> Do you know the cause and for what dtypes is it flaky?

It was probably due to `Absolute` being non-differentiable at 0. 
We get zeros in the input as we generate input from normal distribution centred at 0.
Realised that when `MeanAbsoluteError` also failed. Have added `dodge_nondifferentiable = True` for `HuberLoss` as well as `MeanAbsoluteError`.

chainerx_cc/chainerx/routines/loss.cc
```diff
@@ -0,0 +1,28 @@
...
+#include "chainerx/routines/misc.h"
+#include "chainerx/scalar.h"
+
+namespace chainerx {
+
+Array MeanAbsoluteError(const Array& x1, const Array& x2) { return Absolute(x1 - x2).Mean(); }
+
+Array MeanSquaredError(const Array& x1, const Array& x2) { return SquaredDifference(x1, x2).Mean(); }
+
+Array GaussianKLDivergence(const Array& mu, const Array& ln_var) { return (Square(mu) + Exp(ln_var) - ln_var - 1) * 0.5; }
```
hvy: A detail but let's be consistent with `mu` and `mean`. I think the latter is better as it's used in Chainer but open for discussions.

kshitij12345: I agree.

hvy:
I'm very sorry for a delayed input but we for now have a consensus to keep the APIs simple in terms of how these types of routines reduce. More specifically, we're not performing any reductions at all (users can perform arbitrary reductions afterwards). This sort of goes agains `Mean{Squared,Absolute}Error`. I discussed this with @beam2d and thought that we actually don't need to provide these two routines. `MeanAbsoluteError` could be replaced with `AbsoluteError`.

Could we hear your opinion, and would it be okay simply to omit these two reducing routines from this PR and introduce a `AbsoluteError`? (The other two routines should of course be there)

It might however be important to let the users find `SquaredDifference` from both the "loss" and the math section in the documentations. It might be beyond the scope of this PR though.

kshitij12345:
Umm honestly, I feel 50-50 on this decision. I agree the API should be minimal and sufficient. Also reducing the output from `{Squared, Absolute}` would be trivial and will allow user to do whatever he wants.

But also the fact that MSE is like a staple loss (especially for regression and compression)  in the machine learning community. It is always present in almost any framework you take. So I take it that everyone assumes it to be present. 
Case can also be made if user really wants some other reduction output, it is easy to get since these functions are relatively trivial to implement.

So are you sure we should remove them the reduction from them?

hvy:
Expanding the API in the future is always possible so I believe this is the right approach for the time being. I absolutely understand your point in that the users might expect MSE to be there so we might reconsider this again. It's reasonable by the way to e.g. in the docs of `squared_error` mention how MSE should be implemented. That'd make it searchable.

hvy:
Jenkins, test this plese.

hvy:
Jenkins, test this please.

hvy:
Thanks for the changes. It basically LGTM so let me merge after merging becomes possible (after release). There are some minor errors in the docs but I could fix them later (unless you want to of course). I think it's okay to delete `squared_difference` by the way but that should probably be addressed in a separate PR.

kshitij12345:
Oh. Sure.
I would be more than happy if you could fix the docs. 

Shouldn't both exist? For explicitly expressing the behaviour.
We can have a single function in C++ while in Python use it to alias the other?

Thank You.

hvy:
Got it. I see your point but at the same time think it'd then be better to reconsider the name 🤔

hvy:
Jenkins, test this please.

hvy:
Jenkins, test this please.

hvy:
Failures seem unrelated. LGTM!

Issue #7057, opened 2019-04-26
[Bug] ValueError: Function `...` (BernoulliLogProb) has a bug.


seiyab:
## Overview
When I calculate a gradient of `log_prob` of Bernoulli distribution, I got error message `(BernoulliLogProb) has a bug.`.

## Conditions
  - Chainer version ... 6.0.0rc1 
  - CuPy version ... I don't use CuPy.
  - Numpy version ... 1.16.2
  - OS/Platform Linux-4.9.125-linuxkit-x86_64-with-Ubuntu-18.04-bionic
  - CUDA/cuDNN version ... I don't use CUDA/cuDNN.

## Code to reproduce
```
import numpy as np
import chainer as ch
from chainer import distributions as dist
from chainer import functions as F

param_p = ch.Parameter(np.array(0.5))
bernoulli = dist.Bernoulli(param_p)
samples = F.cast(bernoulli.sample((10,)), np.float)
loss = np.sum(bernoulli.log_prob(samples))
loss.backward()
```
## Error messages, stack traces, or logs
```
Traceback (most recent call last):          
  File "bernoulli_bag.py", line 10, in <module>                                          
    loss.backward()   
  File "/usr/local/lib/python3.6/dist-packages/chainer/variable.py", line 1433, in backward
    _backprop_to_all([(node, grad_var)], retain_grad, loss_scale)                        
  File "/usr/local/lib/python3.6/dist-packages/chainer/variable.py", line 1667, in _backprop_to_all
    _check_grad_type(func, x, True, gx_elem.array)                                       
  File "/usr/local/lib/python3.6/dist-packages/chainer/variable.py", line 83, in _check_grad_type
    raise typ(detail + msg)                 
ValueError: Function `<property object at 0x7f592daf4a98>` (BernoulliLogProb) has a bug. 

Please report this error to the issue tracker with the stack trace,                      
the information of your environment, and your script:                                    
https://github.com/chainer/chainer/issues/new.                                           
Shape of data and grad mismatch             
grad: (10,) != data: ()  
```

seiyab:
I'll send a pull request.

PR #7064, opened 2019-05-01
Fix a bug in `Bernoulli.log_prob`

seiyab:
Fixed a bug mentioned in https://github.com/chainer/chainer/issues/7057 .

This bug was caused by unintended broadcasting in the backward method.

I added a test that reproduces the bug and fixed it referencing this code https://github.com/chainer/chainer/blob/c92db3396d55a2a351d1bee1a47e33e85dd144b2/chainer/functions/math/basic_math.py#L176  .

### DIFF
--- a/chainer/distributions/bernoulli.py
+++ b/chainer/distributions/bernoulli.py
@@ -6,9 +6,9 @@ from chainer.backends import cuda
 from chainer import distribution
 import chainer.distributions.utils
 from chainer.functions.activation import sigmoid
-from chainer.functions.math import sum
 from chainer.functions.math import exponential
 from chainer.functions.math import logarithm_1p
+from chainer.functions.math import sum
 from chainer import utils
 from chainer.utils import cache
 
--- a/chainer/distributions/bernoulli.py
+++ b/chainer/distributions/bernoulli.py
@@ -5,6 +5,7 @@ from chainer import backend
 from chainer.backends import cuda
 from chainer import distribution
 import chainer.distributions.utils
+from chainer.functions import sum_to
 from chainer.functions.activation import sigmoid
 from chainer.functions.math import exponential
 from chainer.functions.math import logarithm_1p
@@ -53,7 +54,7 @@ class BernoulliLogProb(chainer.function_node.FunctionNode):
         nan_dlogit[self.logit_isminf] = numpy.nan
         dlogit += nan_dlogit
 
-        return gy * dlogit, None
+        return sum_to(gy * dlogit, logit.shape), None
 
 
 def _bernoulli_log_prob(logit, x, binary_check=False):
--- a/chainer/distributions/bernoulli.py
+++ b/chainer/distributions/bernoulli.py
@@ -5,8 +5,8 @@ from chainer import backend
 from chainer.backends import cuda
 from chainer import distribution
 import chainer.distributions.utils
-from chainer.functions import sum_to
 from chainer.functions.activation import sigmoid
+from chainer.functions.math import sum
 from chainer.functions.math import exponential
 from chainer.functions.math import logarithm_1p
 from chainer import utils
@@ -54,7 +54,7 @@ class BernoulliLogProb(chainer.function_node.FunctionNode):
         nan_dlogit[self.logit_isminf] = numpy.nan
         dlogit += nan_dlogit
 
-        return sum_to(gy * dlogit, logit.shape), None
+        return sum.sum_to(gy * dlogit, logit.shape), None
 
 
 def _bernoulli_log_prob(logit, x, binary_check=False):
--- a/chainer/distributions/bernoulli.py
+++ b/chainer/distributions/bernoulli.py
@@ -6,6 +6,7 @@ from chainer.backends import cuda
 from chainer import distribution
 import chainer.distributions.utils
 from chainer.functions.activation import sigmoid
+from chainer.functions.array import where
 from chainer.functions.math import exponential
 from chainer.functions.math import logarithm_1p
 from chainer.functions.math import sum
@@ -47,12 +48,11 @@ class BernoulliLogProb(chainer.function_node.FunctionNode):
         dlogit = x - 1. / (1. + exponential.exp(-logit))
 
         # extreme logit
-        nan_dlogit = xp.zeros_like(dlogit.array)
+        nan = xp.array(xp.nan).astype(dlogit.dtype)
+        logit_isinf = xp.bitwise_or(self.logit_ispinf, self.logit_isminf)
+        dlogit = where.where(logit_isinf, nan, dlogit)
         if self.binary_check:
-            nan_dlogit[self.invalid] = numpy.nan
-        nan_dlogit[self.logit_ispinf] = numpy.nan
-        nan_dlogit[self.logit_isminf] = numpy.nan
-        dlogit += nan_dlogit
+            dlogit = where.where(self.invalid, nan, dlogit)
 
         return sum.sum_to(gy * dlogit, logit.shape), None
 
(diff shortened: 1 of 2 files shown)


chainer/distributions/bernoulli.py
```diff
@@ -5,6 +5,7 @@
 from chainer.backends import cuda
 from chainer import distribution
 import chainer.distributions.utils
+from chainer.functions import sum_to
```
niboshi: We do not allow importing non-module objects: https://docs.openstack.org/hacking/latest/user/hacking.html#imports

Please import like `from chainer.functions.math import sum` and use `sum.sum_to`.

seiyab:
@niboshi 
Thank you for your review.
I fixed the import statement.

niboshi:
Jenkins, test this please

niboshi:
Jenkins, test this please

niboshi:
@seiyab Could you check test failures?

seiyab:
@niboshi 
I'll check them.

I'm sorry that I tested with only python3.
I'll reproduce them in my environment and fix them.
Now I cannot reproduce failure on my python2 environment. I might ask it when I think that I'm stuck.

Some of failures seem not to be related. (e.g. [test_convolution_2d](https://jenkins.preferred.jp/job/chainer/job/chainer_pr/699/TEST=chainer-py2,label=mn1-p100/testReport/junit/tests.chainer_tests.functions_tests.connection_tests.test_convolution_2d/TestConvolution2DFunction_use_chainerx_false__chainerx_device_None__use_cuda_true__cuda_device_0__use_cudnn_always__cudnn_deterministic_true__autotune_false__cudnn_fast_batch_normalization_false__use_ideep_never_param_31_%7Bcover_all=False,%20groups=1,%20dilate=1,%20nobias=False,%20c_contiguous=False,%20W_dtype=float32,%20x_dtype=float16%7D/test_double_backward/)
I will check causes but I might also ask it when I think that I'm stuck.

niboshi:
@seiyab Yes, currently some of our tests occasionally fail (including convolution_2d)... Please ignore them. Sorry for inconvenience.

seiyab:
@niboshi 
I see. I'll check test_convolution_2d a little just to make sure.

And I reproduced the future warning on my python2 environment downgrading numpy version.
I'll fix.

seiyab:
@niboshi 
I resolved the warning in my python2 environment (python 2.7.16, numpy 1.12.1).
Of course I also tested on python3.

I have two questions.

1. Is it better that tests for infinity logits exist?
https://github.com/chainer/chainer/pull/7064/commits/a0256a672e30a6866de2e23c9370953975bf6724 changes the implementation to handle infinity logits.
It is not tested as far as I know.
I manually confirmed the behavior but I'm not convinced that it always works well.

2. Is it better that resolve the warning caused by infinity logits?
When I manually confirmed the behavior of infinity logits, I got a RuntimeWarning message.
It might be better that resolve the message since the code explicitly handles infinity logits as p=1.

Both of them might be included this PR or resolved in new PRs.

I confirmed the behavior of logit containing an infinity value like following.
```
>>> import chainer as ch             
>>> import numpy as np                                                                      
>>> from chainer.distributions.bernoulli import Bernoulli
>>> logit = ch.Variable(np.array([0, np.inf]))                                              
>>> x = ch.Variable(np.array([1., 1.]))       
>>> b = Bernoulli(logit=logit, binary_check=True)                                           
>>> logprob = b.log_prob(x)
/src/chainer/distributions/bernoulli.py:27: RuntimeWarning: invalid value encountered in multiply
  y = logit * (x - 1) - xp.log(xp.exp(-logit) + 1)                                          
>>> logprob            
variable([-0.69314718,  0.        ])          
>>> sum(logprob).backward()                   
>>> logit.grad         
array([0.5, nan])  
```
The result above is same as Chainer 6.0.0rc1.

toslunar:
>     1. Is it better that tests for infinity logits exist?
>        [a0256a6](https://github.com/chainer/chainer/commit/a0256a672e30a6866de2e23c9370953975bf6724) changes the implementation to handle infinity logits.
>        It is not tested as far as I know.
>        I manually confirmed the behavior but I'm not convinced that it always works well.

I'd like to have tests that the backward just runs, while I don't think it's mandatory to check values of grads in the PR because the *forward* tests are weak and the PR fixes the backward.  The backward tests may hard-code the expected grads since `gradient_check.numerical_grad` doesn't support infinities.

On forward tests, infinity logits are tested via the testing parameter `self.extreme_values=True` of `TestBernoulli`, where several issues have been left because of low-priority.

- For non-binary inputs there should be 8 test methods: {`prob`, `log_prob`} × {`binary_check=True`, `binary_check=False`} × {`cpu`, `gpu`}, but there are 4.
- Value tests for non-binary inputs are missing.  Such tests could be added to `check_log_prob_binary_check` and `check_prob_binary_check`, whose name would be renamed to something like `*_non_binary`.

seiyab:
@toslunar 
Thank you for your response.
Let me confirm your indications.

Does
> I'd like to have tests that the backward just runs...

mean
> It seems better to have tests that confirm backward method doesn't raise error when logit have infinity value.

rather than
> It seems better to have tests that hard-code the expected grads for logits of infinity value.

or
> The existing tests that checks grads for finite values are enough.

?

About forward, now I think there is no need to worry about it in this PR.
I'll send another PR if I want to fix them.

toslunar:
@seiyab 
> Does
> 
> > I'd like to have tests that the backward just runs...
> 
> mean
> 
> > It seems better to have tests that confirm backward method doesn't raise error when logit have infinity value.

Yes.  Thanks for clarifying it.

seiyab:
@toslunar 
I added a test.

tests/chainer_tests/distributions_tests/test_bernoulli.py
```diff
@@ -156,5 +163,16 @@ def test_double_backward_gpu(self):
             cuda.to_gpu(self.logit), cuda.to_gpu(self.x),
             cuda.to_gpu(self.gy), cuda.to_gpu(self.ggx))
 
+    def test_backward_where_logit_has_infinite_values(self):
+        self.logit += numpy.inf
```
toslunar: For the readability,
```suggestion
        self.logit[...] = numpy.inf
```
looks better to me.

tests/chainer_tests/distributions_tests/test_bernoulli.py
```diff
@@ -156,5 +163,16 @@ def test_double_backward_gpu(self):
...
+        with numpy.errstate(invalid='ignore'):
+            log_prob = distributions.bernoulli._bernoulli_log_prob(
+                self.logit, self.x)
+
+        log_prob.backward()
+
+        # just confirm that the backward method runs without raising error.
+        assert True
```
toslunar: How does the line confirm it?

tests/chainer_tests/distributions_tests/test_bernoulli.py
seiyab: Thank you.
I'll fix it.

tests/chainer_tests/distributions_tests/test_bernoulli.py
seiyab: I'm sorry for misleading code.

Is the following without assert statement clearer?
```
# just confirm that the backward method runs without raising error.
log_prob.backward()
```

What I wanted to express is following.
> This test method just confirms that the backward method runs without raising error.
> So assert statement is always passed.
> `assert True` explicitly expresses there is nothing to assert.

toslunar: In the repository `assert True` is not used.  How about replacing the last 4 lines with the following?
```
        # just confirm that the backward method runs without raising error.
        log_prob.backward()
```

seiyab: I adopted that. Thank you.

tests/chainer_tests/distributions_tests/test_bernoulli.py
seiyab: I fixed it.

toslunar:
Jenkins, test this please.

PR #7067, opened 2019-05-04
[chainerx] add indexing where

kshitij12345:
References : #6423 

To-Do:
* ~~Tests.~~
* ~~Documentation.~~

### DIFF
--- a/chainerx/_docs/routines.py
+++ b/chainerx/_docs/routines.py
@@ -450,16 +450,17 @@ Note:
     _docs.set_doc(
         chainerx.where,
         """where(condition, x, y)
-Return elements chosen from x or y depending on condition.
+Return elements chosen from ``x`` or ``y`` depending on condition.
 
 Args:
-    condition (~chainerx.ndarray): Where True, yield x, otherwise yield y.
+    condition (~chainerx.ndarray): Where True, yield ``x``, otherwise
+    yield ``y``.
     x (~chainerx.ndarray): Values from which to choose.
     y (~chainerx.ndarray): Values from which to choose.
 
 Returns:
     :func:`~chainerx.ndarray`: An array with elements
-    from x where condition is True, and elements from y elsewhere.
+    from ``x`` where condition is True, and elements from ``y`` elsewhere.
 
 Note:
     During backpropagation, this function propagates the gradient of the
--- a/chainerx/_docs/routines.py
+++ b/chainerx/_docs/routines.py
@@ -462,9 +462,6 @@ Returns:
     from x where condition is True, and elements from y elsewhere.
 
 Note:
-    x, y and condition need to be broadcastable to some shape.
-
-Note:
     During backpropagation, this function propagates the gradient of the
     output array to the input array ``x`` and ``y``.
 
--- a/chainerx/_docs/routines.py
+++ b/chainerx/_docs/routines.py
@@ -447,6 +447,30 @@ Note:
 .. seealso:: :func:`numpy.take`
 """)
 
+    _docs.set_doc(
+        chainerx.where,
+        """where(condition, x, y)
+Return elements chosen from x or y depending on condition.
+
+Args:
+    condition (~chainerx.ndarray): Where True, yield x, otherwise yield y.
+    x (~chainerx.ndarray): Values from which to choose.
+    y (~chainerx.ndarray): Values from which to choose.
+
+Returns:
+    :func:`~chainerx.ndarray`: An array with elements
+    from x where condition is True, and elements from y elsewhere.
+
+Note:
+    x, y and condition need to be broadcastable to some shape.
+
+Note:
+    During backpropagation, this function propagates the gradient of the
+    output array to the input array ``x`` and ``y``.
+
+.. seealso:: :func:`numpy.where`
+""")
+
 
 def _docs_linalg():
     _docs.set_doc(
--- a/chainerx_cc/chainerx/cuda/cuda_device/indexing.cu
+++ b/chainerx_cc/chainerx/cuda/cuda_device/indexing.cu
@@ -25,7 +25,6 @@
 #include "chainerx/kernels/indexing.h"
 #include "chainerx/macro.h"
 #include "chainerx/routines/indexing.h"
-#include "chainerx/routines/type_util.h"
 #include "chainerx/shape.h"
 
 namespace chainerx {
@@ -281,7 +280,7 @@ public:
         device.CheckDevicesCompatible(condition, x, y, out);
         const Array& condition_cast = condition.dtype() != Dtype::kBool ? condition.AsType(Dtype::kBool) : condition;
 
-        Dtype out_dtype = ResultType(x, y);
+        Dtype out_dtype = out.dtype();
         const Array& x_cast = x.dtype() != out_dtype ? x.AsType(out_dtype) : x;
         const Array& y_cast = y.dtype() != out_dtype ? y.AsType(out_dtype) : y;
 
--- a/chainerx_cc/chainerx/cuda/cuda_device/indexing.cu
+++ b/chainerx_cc/chainerx/cuda/cuda_device/indexing.cu
@@ -278,15 +278,12 @@ public:
     void Call(const Array& condition, const Array& x, const Array& y, const Array& out) override {
         Device& device = x.device();
         device.CheckDevicesCompatible(condition, x, y, out);
-        Dtype dtype = std::max(x.dtype(), y.dtype());
-        const Array& x_cast = x.dtype() == dtype ? x : x.AsType(dtype);
-        const Array& y_cast = y.dtype() == dtype ? y : y.AsType(dtype);
-        const Array& out_cast = out.dtype() == dtype ? out : out.AsType(dtype);
+        const Array& condition_cast = condition.dtype() != Dtype::kBool ? condition.AsType(Dtype::kBool) : condition;
 
         CudaSetDeviceScope scope{device.index()};
-        VisitDtype(dtype, [&](auto x_pt) {
+        VisitDtype(out.dtype(), [&](auto x_pt) {
             using T = typename decltype(x_pt)::type;
-            Elementwise<const bool, const T, const T, T>(WhereImpl<T>{}, condition, x_cast, y_cast, out_cast);
+            Elementwise<const bool, const T, const T, T>(WhereImpl<T>{}, condition_cast, x, y, out);
         });
     }
 };
--- a/chainerx_cc/chainerx/cuda/cuda_device/indexing.cu
+++ b/chainerx_cc/chainerx/cuda/cuda_device/indexing.cu
@@ -267,6 +267,32 @@ public:
 
 CHAINERX_CUDA_REGISTER_KERNEL(AddAtKernel, CudaAddAtKernel);
 
+template <typename T>
+struct WhereImpl {
+    using CudaType = cuda_internal::DataType<T>;
+    __device__ void operator()(int64_t /*i*/, bool condition, CudaType x, CudaType y, CudaType& out) { out = condition ? x : y; }
+};
+
+class CudaWhereKernel : public WhereKernel {
+public:
+    void Call(const Array& condition, const Array& x, const Array& y, const Array& out) override {
+        Device& device = x.device();
+        device.CheckDevicesCompatible(condition, x, y, out);
+        Dtype dtype = std::max(x.dtype(), y.dtype());
+        const Array& x_cast = x.dtype() == dtype ? x : x.AsType(dtype);
+        const Array& y_cast = y.dtype() == dtype ? y : y.AsType(dtype);
+        const Array& out_cast = out.dtype() == dtype ? out : out.AsType(dtype);
+
+        CudaSetDeviceScope scope{device.index()};
+        VisitDtype(dtype, [&](auto x_pt) {
+            using T = typename decltype(x_pt)::type;
+            Elementwise<const bool, const T, const T, T>(WhereImpl<T>{}, condition, x_cast, y_cast, out_cast);
+        });
+    }
+};
+
+CHAINERX_CUDA_REGISTER_KERNEL(WhereKernel, CudaWhereKernel);
+
 }  // namespace
 }  // namespace cuda
 }  // namespace chainerx
--- a/chainerx_cc/chainerx/cuda/cuda_device/indexing.cu
+++ b/chainerx_cc/chainerx/cuda/cuda_device/indexing.cu
@@ -25,6 +25,7 @@
 #include "chainerx/kernels/indexing.h"
 #include "chainerx/macro.h"
 #include "chainerx/routines/indexing.h"
+#include "chainerx/routines/type_util.h"
 #include "chainerx/shape.h"
 
 namespace chainerx {
@@ -280,10 +281,14 @@ public:
         device.CheckDevicesCompatible(condition, x, y, out);
         const Array& condition_cast = condition.dtype() != Dtype::kBool ? condition.AsType(Dtype::kBool) : condition;
 
+        Dtype out_dtype = ResultType(x, y);
+        const Array& x_cast = x.dtype() != out_dtype ? x.AsType(out_dtype) : x;
+        const Array& y_cast = y.dtype() != out_dtype ? y.AsType(out_dtype) : y;
+
         CudaSetDeviceScope scope{device.index()};
         VisitDtype(out.dtype(), [&](auto x_pt) {
             using T = typename decltype(x_pt)::type;
-            Elementwise<const bool, const T, const T, T>(WhereImpl<T>{}, condition_cast, x, y, out);
+            Elementwise<const bool, const T, const T, T>(WhereImpl<T>{}, condition_cast, x_cast, y_cast, out);
         });
     }
 };
--- a/chainerx_cc/chainerx/routines/indexing.cc
+++ b/chainerx_cc/chainerx/routines/indexing.cc
@@ -201,13 +201,10 @@ Array Take(const Array& a, const Array& indices, int8_t axis) {
 
 Array Where(const Array& condition, const Array& x, const Array& y) {
     Dtype out_dtype = ResultType(x, y);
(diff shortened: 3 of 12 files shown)


chainerx_cc/chainerx/routines/math.cc
```diff
@@ -1176,4 +1176,28 @@ Array IsFinite(const Array& x) {
...
+    Shape _false_sh = x_b.shape();
+    std::fill(_false_sh.begin(), _false_sh.end(), 1);
+
+    Scalar{false};
+
+    const Array& _false = Zeros(_false_sh, Dtype::kBool, condition.device());
+    Array out = Empty(condition.shape(), x.dtype(), x.device());
+    IfGreaterElseImpl(condition, _false.BroadcastTo(x_b.shape()), x_b, y_b, out);
+
+    return out;
```
kshitij12345: Is this ok? Is there a better alternative way? should we just opt for a `where`?

chainerx_cc/chainerx/routines/math.cc
```diff
@@ -1176,4 +1176,28 @@ Array IsFinite(const Array& x) {
     return out;
 }
 
+Array Where(const Array& condition, const Array& x, const Array& y) {
+    if (condition.dtype() != Dtype::kBool) {
```
asi1024: How about casting `condition` to bool dtype if the given condition is not a bool array?

chainerx_cc/chainerx/routines/math.cc
```diff
@@ -1176,4 +1176,28 @@ Array IsFinite(const Array& x) {
...
+        throw DtypeError{"condition should be an array of dtype bool."};
+    }
+
+    if (x.dtype() != y.dtype()) {
+        throw DtypeError{"dtype of x and y should be same."};
+    }
+
+    const Array& x_b = x.BroadcastTo(condition.shape());
+    const Array& y_b = y.BroadcastTo(condition.shape());
```
asi1024: How about broadcasting to the following shape?
```cpp
Shape out_shape = internal::BroadcastShapes(condition.shape(), internal::BroadcastShapes(x.shape(), y.shape()));
```

chainerx_cc/chainerx/routines/math.cc
asi1024: We want the optimized `where` because it will be used in many other functions.

kshitij12345: Have used custom kernel for `where`

chainerx_cc/chainerx/routines/math.cc
kshitij12345: Done. However do note that as per 
https://github.com/onnx/onnx/blob/master/docs/Operators.md#Where

`condition` should be `bool`

chainerx_cc/chainerx/routines/math.cc
kshitij12345: Done.

chainerx_cc/chainerx/cuda/cuda_device/indexing.cu
```diff
@@ -267,6 +267,32 @@ public:
...
+
+class CudaWhereKernel : public WhereKernel {
+public:
+    void Call(const Array& condition, const Array& x, const Array& y, const Array& out) override {
+        Device& device = x.device();
+        device.CheckDevicesCompatible(condition, x, y, out);
+        Dtype dtype = std::max(x.dtype(), y.dtype());
+        const Array& x_cast = x.dtype() == dtype ? x : x.AsType(dtype);
+        const Array& y_cast = y.dtype() == dtype ? y : y.AsType(dtype);
+        const Array& out_cast = out.dtype() == dtype ? out : out.AsType(dtype);
```
asi1024: The output will not be updated when `out.dtype == dtype` is false.

chainerx_cc/chainerx/routines/indexing.cc
```diff
@@ -198,4 +198,41 @@ Array Take(const Array& a, const Array& indices, int8_t axis) {
     return out;
 }
 
+Array Where(const Array& condition, const Array& x, const Array& y) {
+    const Array& condition_cast = condition.dtype() != Dtype::kBool ? condition.AsType(Dtype::kBool) : condition;
```
asi1024: Could you move this cast to kernel definition of each device?

chainerx_cc/chainerx/routines/indexing.cc
```diff
@@ -198,4 +198,41 @@ Array Take(const Array& a, const Array& indices, int8_t axis) {
...
+    {
+        NoBackpropModeScope scope;
+        x.device().backend().CallKernel<WhereKernel>(condition_cast, x_b, y_b, out);
+    }
+
+    BackwardBuilder bb{"where", {x_b, y_b}, out};
+    if (BackwardBuilder::Target bt = bb.CreateTarget(0)) {
+        bt.Define([condition_cast](BackwardContext& bctx) {
+            const Array& gout = *bctx.output_grad();
+            bctx.input_grad() = Where(condition_cast, gout, ZerosLike(gout));
```
asi1024: This line can be improved by using array-scalar Where. Could you write TODO?

chainerx_cc/chainerx/cuda/cuda_device/indexing.cu
kshitij12345: Actually this case doesn't happen as we make output, the same-type as `x and y`. Also note that, we don't allow `x` and `y` to be of different `dtypes` similar to onnx description. However numpy promotes the lower dtype. Should we also do the same?

chainerx_cc/chainerx/routines/indexing.cc
kshitij12345: Sure.

chainerx_cc/chainerx/routines/indexing.cc
kshitij12345: Will do that.

chainerx_cc/chainerx/cuda/cuda_device/indexing.cu
asi1024: For consistency with other functions, this function should support mixed dtype. For example, `chainerx::Concatenate` supports mixed dtype whereas onnx's `Concat` requires the same types.

chainerx_cc/chainerx/routines/math.cc
asi1024: For consistency with logical functions (e.g., `logical_not`), the condition should allow all dtypes.

kshitij12345: Oh. Have done it in the recent commits.

chainerx_cc/chainerx/cuda/cuda_device/indexing.cu
kshitij12345: Oh. Done.

chainerx_cc/chainerx/routines/indexing.cc
```diff
@@ -199,35 +200,31 @@ Array Take(const Array& a, const Array& indices, int8_t axis) {
...
 
 Array Where(const Array& condition, const Array& x, const Array& y) {
-    const Array& condition_cast = condition.dtype() != Dtype::kBool ? condition.AsType(Dtype::kBool) : condition;
-
-    if (x.dtype() != y.dtype()) {
-        throw DtypeError{"dtype of x and y should be same."};
-    }
+    Dtype out_dtype = ResultType(x, y);
+    const Array& x_cast = x.dtype() != out_dtype ? x.AsType(out_dtype) : x;
+    const Array& y_cast = y.dtype() != out_dtype ? y.AsType(out_dtype) : y;
```
asi1024: Could you move these casts to device implementation?

tests/chainerx_tests/unit_tests/routines_tests/test_indexing.py
```diff
@@ -204,6 +205,50 @@ def forward_xp(self, inputs, xp):
         return b,
 
 
+_in_out_dtypes_where = dtype_utils._permutate_dtype_mapping([
```
asi1024: How about using `dtype_utils.result_dtypes_two_arrays`?

tests/chainerx_tests/unit_tests/routines_tests/test_indexing.py
```diff
@@ -214,31 +259,38 @@ def forward_xp(self, inputs, xp):
...
-        if self.dtypes == 'float16':
+        if x_dtype == 'float16' or y_dtype == 'float16':
             self.check_backward_options.update({'rtol': 1e-3, 'atol': 1e-3})
 
     def generate_inputs(self):
-        x = numpy.random.uniform(-1, 1, self.x_shape).astype(self.dtypes)
-        y = numpy.random.uniform(-1, 1, self.y_shape).astype(self.dtypes)
+        (x_dtype, y_dtype) = self.in_dtypes
+        x = numpy.random.uniform(-1, 1, self.x_shape).astype(x_dtype)
+        y = numpy.random.uniform(-1, 1, self.y_shape).astype(y_dtype)
```
asi1024: How about using `array_utils.uniform` for integer types?

tests/chainerx_tests/unit_tests/routines_tests/test_indexing.py
```diff
@@ -214,31 +259,38 @@ def forward_xp(self, inputs, xp):
...
         condition = \
             numpy.random.uniform(-1, 1,
                                  self.cond_shape).astype(self.condition_dtypes)
         return (x, y, condition)
 
     def forward_xp(self, inputs, xp):
         x, y, condition = inputs
```
asi1024: The given condition might be better to be of bool dtype. How about operating >0.5 in `generate_inputs`?

chainerx_cc/chainerx/routines/indexing.cc
kshitij12345: Done. However, have kept 
`Dtype out_dtype = ResultType(x, y);` as is since it is consumed for while initializing `out`.

tests/chainerx_tests/unit_tests/routines_tests/test_indexing.py
kshitij12345: Done. Thanks.

tests/chainerx_tests/unit_tests/routines_tests/test_indexing.py
kshitij12345: Done. Used it for generating `x` and `y` and `condition`.

tests/chainerx_tests/unit_tests/routines_tests/test_indexing.py
kshitij12345: Moved it to `generate_inputs` and cast `condition > 0.5` to the `condition_dtype`. 
Earlier input `condition` to `xp.where` was always bool as you mentioned. Thanks.

chainerx/_docs/routines.py
```diff
@@ -447,6 +447,30 @@ def _docs_indexing():
...
+    condition (~chainerx.ndarray): Where True, yield x, otherwise yield y.
+    x (~chainerx.ndarray): Values from which to choose.
+    y (~chainerx.ndarray): Values from which to choose.
+
+Returns:
+    :func:`~chainerx.ndarray`: An array with elements
+    from x where condition is True, and elements from y elsewhere.
+
+Note:
+    x, y and condition need to be broadcastable to some shape.
```
asi1024:
```suggestion
    ``x``, ``y`` and ``condition`` need to be broadcastable to some shape.
```

asi1024: By the way, I think this note is not needed because other docs also don't have such a note.

chainerx_cc/chainerx/routines/indexing.cc
```diff
@@ -198,4 +199,36 @@ Array Take(const Array& a, const Array& indices, int8_t axis) {
     return out;
 }
 
+Array Where(const Array& condition, const Array& x, const Array& y) {
+    Dtype out_dtype = ResultType(x, y);
+    Shape out_shape = internal::BroadcastShapes(condition.shape(), internal::BroadcastShapes(x.shape(), y.shape()));
+    Array out = Empty(out_shape, out_dtype, x.device());
+    const Array& x_b = x.BroadcastTo(out_shape);
+    const Array& y_b = y.BroadcastTo(out_shape);
```
asi1024: `condition` also should be broadcasted.

tests/chainerx_tests/unit_tests/routines_tests/test_indexing.py
```diff
@@ -201,3 +203,61 @@ def forward_xp(self, inputs, xp):
...
+    # Broadcast Shapes
+    ((2, 3), (1, 3), (1, 3)),
+    ((2, 3), (2, 1), (1, 3)),
+    ((2, 3), (2, 3), (1, 3)),
+])
+@chainer.testing.parameterize_pytest(
+    'in_dtypes,out_dtype', dtype_utils.result_dtypes_two_arrays
+)
+@chainer.testing.parameterize_pytest(
+    'condition_dtype', chainerx.testing.integral_dtypes)
```
asi1024: How about adding tests for float conditions?
```suggestion
    'condition_dtype', chainerx.testing.all_dtypes)
```

chainerx/_docs/routines.py
kshitij12345: Oh. Will remove it.

chainerx_cc/chainerx/routines/indexing.cc
kshitij12345: Done.

tests/chainerx_tests/unit_tests/routines_tests/test_indexing.py
kshitij12345: Done.

chainerx_cc/chainerx/cuda/cuda_device/indexing.cu
```diff
@@ -267,6 +268,33 @@ public:
...
+};
+
+class CudaWhereKernel : public WhereKernel {
+public:
+    void Call(const Array& condition, const Array& x, const Array& y, const Array& out) override {
+        Device& device = x.device();
+        device.CheckDevicesCompatible(condition, x, y, out);
+        const Array& condition_cast = condition.dtype() != Dtype::kBool ? condition.AsType(Dtype::kBool) : condition;
+
+        Dtype out_dtype = ResultType(x, y);
```
asi1024: Could you fix as follows?
```suggestion
        Dtype out_dtype = out.dtype();
```

kshitij12345: Ahh. Right. On it.

asi1024: If the device functions are called from `chainerx::Where`, this fix does not change the behavior. However, if not, `x` and `y` seems to be better to be casted to `out.dtype`.

tests/chainerx_tests/unit_tests/routines_tests/test_indexing.py
```diff
@@ -201,3 +203,64 @@ def forward_xp(self, inputs, xp):
...
+            self.skip_double_backward_test = True
+
+        if x_dtype == 'float16' or y_dtype == 'float16':
+            self.check_backward_options.update({'rtol': 1e-3, 'atol': 1e-3})
+
+    def generate_inputs(self):
+        (x_dtype, y_dtype) = self.in_dtypes
+        x = array_utils.uniform(self.x_shape, x_dtype)
+        y = array_utils.uniform(self.y_shape, y_dtype)
+        condition = array_utils.uniform(self.cond_shape, self.condition_dtype)
```
asi1024: `self.condition` should be of `self.condition_dtype`, but this `condition` should be float array.

chainerx_cc/chainerx/cuda/cuda_device/indexing.cu
kshitij12345: True, this is more consistent, as in the case if someone calls `out` with a different `dtype` than inferred from `ResultType(x, y)`, that can be a problem.

tests/chainerx_tests/unit_tests/routines_tests/test_indexing.py
kshitij12345: On it.

asi1024: Could you fix to `numpy.random.uniform(0, 1, size=self.cond_shape, 'float32')`?
`array_utils.uniform` does not guarantee the range of return value.

kshitij12345: Done.

chainerx/_docs/routines.py
```diff
@@ -447,6 +447,27 @@ def _docs_indexing():
 .. seealso:: :func:`numpy.take`
 """)
 
+    _docs.set_doc(
+        chainerx.where,
+        """where(condition, x, y)
+Return elements chosen from x or y depending on condition.
```
asi1024:
```suggestion
Return elements chosen from ``x`` or ``y`` depending on condition.
```

chainerx/_docs/routines.py
```diff
@@ -447,6 +447,27 @@ def _docs_indexing():
...
+
+Args:
+    condition (~chainerx.ndarray): Where True, yield x, otherwise yield y.
```
asi1024:
```suggestion
    condition (~chainerx.ndarray): Where True, yield ``x``, otherwise yield ``y``.
```

chainerx/_docs/routines.py
```diff
@@ -447,6 +447,27 @@ def _docs_indexing():
...
+    x (~chainerx.ndarray): Values from which to choose.
+    y (~chainerx.ndarray): Values from which to choose.
+
+Returns:
+    :func:`~chainerx.ndarray`: An array with elements
+    from x where condition is True, and elements from y elsewhere.
```
asi1024:
```suggestion
    from ``x`` where condition is True, and elements from ``y`` elsewhere.
```

asi1024:
Jenkins, test this please.

asi1024:
LGTM. Thank you for the PR!

PR #7077, opened 2019-05-08
Revert #7017

niboshi:
Currently CI is broken in master.
I haven't succeeded in replication, but it seems #7017 is the first PR to cause those errors.
Trying to see if a revert fixes the problem.


niboshi:
Jenkins, test this please

niboshi:
The problem remains.

niboshi:
#7083

PR #7080, opened 2019-05-08
Fix `Reshape` copy condition

hvy:
Copies were made by `chainerx::Reshape` where views would suffice, due to an unnecessarily strict check. This PR fixes that check.

A side note. `chainerx::ExpandDims` simply calls `chainexr::Reshape` to add an axis. This should in practice not incur a copy. However, it could in certain cases. The following assertion would e.g. fail. This is fixed and tested in this PR.

```python
import chainerx as chx

a = chx.random.uniform(-1, 1, (2, 1, 3)).T
b = a.reshape(1, 3, 1, 2)  # Just an "expand dims".
assert a.data_ptr == b.data_ptr  # Would fail, meaning a copy was created.
```


asi1024:
Jenkins, test this please.

asi1024:
LGTM.

PR #7083, opened 2019-05-08
Exclude protobuf 3.8.0rc1 from dependencies

niboshi:
Currently CI is broken in master.
It turned out that cause is protobuf==3.8.0rc1 being used in CIs.
In some CIs we install Chainer with `python setup.py install`, which causes pre-release dependencies to be installed (pypa/setuptools/issues/855).

This PR is a temporary workaround.
We should probably always use pip for installing chainer.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,7 +28,11 @@ requirements = {
         'typing_extensions',
         'filelock',
         'numpy>=1.9.0',
-        'protobuf>=3.0.0',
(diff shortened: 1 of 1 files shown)


hvy:
Jenkins, test this please.

hvy:
Hm, convolution_2d tests are failing on pfnCI but is it even related to this PR?

hvy:
LGTM!

PR #7087, opened 2019-05-08
Support ChainerX in Sentiment example

hvy:
Part of #6661.

- [x] train_sentiment.py (improved by ~87%)
- [x] train_recursive_minibatch.py (improved by ~5%)
- [ ] ~test_thin_stack.py~ Tests are already failing, actually. I think it should be addressed in a separate ticket. https://github.com/chainer/chainer/issues/7100


examples/sentiment/train_recursive_minibatch.py
```diff
@@ -240,6 +235,8 @@ def main():
         ['epoch', 'main/loss', 'validation/main/loss',
          'main/accuracy', 'validation/main/accuracy', 'elapsed_time']))
 
+    trainer.extend(extensions.ProgressBar(update_interval=10))
```
hvy: Not required for this PR but there are other examples including this line and I don't think it hurts (but rather helps).

niboshi: How about separating a PR?

examples/sentiment/train_recursive_minibatch.py
```diff
@@ -215,18 +211,17 @@ def main():
...
-    if args.gpu >= 0:
-        model.to_gpu()
+    model.to_device(device)
 
     optimizer = chainer.optimizers.AdaGrad(0.1)
     optimizer.setup(model)
 
     updater = training.StandardUpdater(
-        train_iter, optimizer, device=None, converter=convert)
+        train_iter, optimizer, converter=convert, device=device)
```
niboshi: This order of arguments is inconsistent with `train_sentiment.py`.
I have no preference but inconsistency is not good.

examples/sentiment/train_sentiment.py
```diff
@@ -118,10 +118,18 @@ def evaluate(model, test_trees):
         acc_root, result['correct_root'], result['total_root']))
 
 
+@chainer.dataset.converter()
+def convert(batch, _):
+    return batch
```
niboshi: It's not related to the PR but I just wonder why `device` can be omitted. Do you know anything?

hvy: The batch contains `str`s actually, not ndarrays.

examples/sentiment/train_recursive_minibatch.py
hvy: I changed the order since it was already "inconsistent" with the `Evaluator` constructor below but then let me align all occurrences to this one instead because that's the order of the arguments defined in the constructors of the `StandardUpdater` and `Evaluator`.

hvy: Fixed.

examples/sentiment/train_recursive_minibatch.py
hvy: I feel this is quite trivial but okay, let me do so.

hvy: Reverted this change.

hvy: Separated to https://github.com/chainer/chainer/pull/7103.

hvy:
Thanks for the comments, they should be addressed now. PTAL!

examples/sentiment/train_sentiment.py
niboshi: Hmm, I see. thanks.

niboshi:
Jenkins, test this please

examples/sentiment/train_recursive_minibatch.py
niboshi: We recently confirmed each PR should focus on a single topic. Let's stick with that policy.

niboshi:
Jenkins, test this please

PR #7095, opened 2019-05-09
Add `F.arctanh`

muupan:
This function is intended to be used to model a Gaussian distribution squashed by tanh as described in https://arxiv.org/abs/1812.05905.


niboshi:
Double-backprop test failed in travis. Could you check it?
```
FAIL tests/chainer_tests/functions_tests/math_tests/test_arctanh.py::TestArctanh_param_0_{dtype=float16, shape=(3, 2)}::test_double_backward_cpu
```

muupan:
I set `eps=1e-3` for backward and double backward tests and locally confirmed that `pytest -m "not gpu" tests/chainer_tests/functions_tests/math_tests/test_arctanh.py` succeeded 100 times in a row.

niboshi:
Jenkins, test this please

niboshi:
LGTM
Thank you!

PR #7101, opened 2019-05-09
Fix trigger constructors to raise errors instead of assertion failures

hvy:
Some triggers with public interfaces used to assert on the user-given inputs. This PR changes those assertions to checks that raise errors.


niboshi:
Jenkins, test this please

niboshi:
LGTM

PR #7103, opened 2019-05-09
Add progress bar to sentiment analysis example

hvy:
Added progress bar for user-friendliness. Other examples also use this progress bar (setting).

Also, please see https://github.com/chainer/chainer/pull/7087#discussion_r282409705.

### DIFF
--- a/examples/sentiment/train_recursive_minibatch.py
+++ b/examples/sentiment/train_recursive_minibatch.py
@@ -240,6 +240,8 @@ def main():
         ['epoch', 'main/loss', 'validation/main/loss',
          'main/accuracy', 'validation/main/accuracy', 'elapsed_time']))
 
+    trainer.extend(extensions.ProgressBar(update_interval=10))
+
     trainer.run()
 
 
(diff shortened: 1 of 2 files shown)


examples/sentiment/train_recursive_minibatch.py
```diff
@@ -240,6 +240,8 @@ def main():
         ['epoch', 'main/loss', 'validation/main/loss',
          'main/accuracy', 'validation/main/accuracy', 'elapsed_time']))
 
+    trainer.extend(extensions.ProgressBar(update_interval=10))
```
niboshi: Why `update_interval=10`?

hvy: Do you mean why I'm not using the default value of 100? I thought that 100 was a bit too long, when testing locally. 1 would be fine too for this example I'd say. Do you have other suggestions?

hvy: For the record, various examples are already using various intervals.

niboshi: >For the record, various examples are already using various intervals.

I'm aware of that. I was just curious how you decided to use 10 in this particular example.

>I thought that 100 was a bit too long, when testing locally. 1 would be fine too for this example I'd say.

So this answers my question.

niboshi:
Jenkins, test this please

niboshi:
Jenkins, test this please

PR #7107, opened 2019-05-09
Fix construction of `std::shared_ptr` with custom deleter in `chianer_interop.cc`

hvy:
Instead of creating a `std::shared_ptr` with a custom deleter, an unintended constructor of `py::object` was called (argument were implicitly converted) without any custom deleters being set. This PR fixes this bug.


niboshi:
Jenkins, test this please

niboshi:
LGTM

PR #6085, opened 2019-01-25
Fix cupy import failure detection

kmaehashi:
In some environment, the second import failure of CuPy may raise `AttributeError` instead of `ImportError` if CUDA driver DLLs are unavailable.

```py
Python 3.7.1 (v3.7.1:260ec2c36a, Oct 20 2018, 14:57:15) [MSC v.1915 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import cupy
Traceback (most recent call last):
  File "C:\Development\venv\kmaehashi-test\lib\site-packages\cupy\__init__.py", line 11, in <module>
    from cupy import core  # NOQA
  File "C:\Development\venv\kmaehashi-test\lib\site-packages\cupy\core\__init__.py", line 1, in <module>
    from cupy.core import core  # NOQA
  File "cupy\core\core.pyx", line 1, in init cupy.core.core
  File "C:\Development\venv\kmaehashi-test\lib\site-packages\cupy\cuda\__init__.py", line 4, in <module>
    from cupy.cuda import compiler  # NOQA
  File "C:\Development\venv\kmaehashi-test\lib\site-packages\cupy\cuda\compiler.py", line 11, in <module>
    from cupy.cuda import device
  File "cupy\cuda\runtime.pxd", line 5, in init cupy.cuda.device
ImportError: DLL load failed: The specified module could not be found.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Development\venv\kmaehashi-test\lib\site-packages\cupy\__init__.py", line 32, in <module>
    six.reraise(ImportError, ImportError(msg), exc_info[2])
  File "C:\Development\venv\kmaehashi-test\lib\site-packages\six.py", line 692, in reraise
    raise value.with_traceback(tb)
  File "C:\Development\venv\kmaehashi-test\lib\site-packages\cupy\__init__.py", line 11, in <module>
    from cupy import core  # NOQA
  File "C:\Development\venv\kmaehashi-test\lib\site-packages\cupy\core\__init__.py", line 1, in <module>
    from cupy.core import core  # NOQA
  File "cupy\core\core.pyx", line 1, in init cupy.core.core
  File "C:\Development\venv\kmaehashi-test\lib\site-packages\cupy\cuda\__init__.py", line 4, in <module>
    from cupy.cuda import compiler  # NOQA
[...]

PR #7114, opened 2019-05-10
Fast integer scale unpooling

tkerola:
I added a code path for faster unpooling when the output shape is an integer multiple of the input shape. This happens e.g. when upsampling the previous result of 2x2 max pooling, which is a common use case. This PR speeds up the forward pass about 1.7-3.2 times for this case.

MWE:
With batch and channel size 1, this PR makes the unpooling 1.7 times faster:
```
# Prepare input data
In [1]: import numpy as np; import cupy; import chainer.functions as F; x = cupy.arange(1, 128*64+1).reshape((1, 1, 64, 128)).astype(np.float32)

# Without this PR (master branch)
In [2]: %timeit F.unpooling_2d(x, 2, outsize=(x.shape[2]*2, x.shape[3]*2))
178 µs ± 18.2 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

# With this PR
In [3]: %timeit F.unpooling_2d(x, 2, outsize=(x.shape[2]*2, x.shape[3]*2))
99.2 µs ± 93.2 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

With larger batch and channel size, this PR makes the unpooling 3.2 times faster:
```
# Prepare input data
In [1]: import numpy as np; import cupy; import chainer.functions as F; x = cupy.arange(1, 128*32*128*64+1).reshape((128, 32, 64, 128)).astype(np.float32)

# Without this PR (master branch)
In [2]: %timeit F.unpooling_2d(x, 2, outsize=(x.shape[2]*2, x.shape[3]*2))
20.1 ms ± 1.82 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

# With this PR
In [3]: %timeit F.unpooling_2d(x, 2, outsize=(x.shape[2]*2, x.shape[3]*2))
6.17 ms ± 17.8 ns per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

### DIFF
--- a/chainer/functions/pooling/unpooling_2d.py
+++ b/chainer/functions/pooling/unpooling_2d.py
@@ -1,7 +1,7 @@
 import numpy
 import numpy.lib.stride_tricks
 try:
-    import cupy.lib.stride_tricks
+    import cupy.lib.stride_tricks  # NOQA
 except ImportError:
     pass
 
@@ -21,7 +21,7 @@ class Unpooling2D(pooling_2d.Pooling2D):
                  outsize=None, cover_all=True):
         super(Unpooling2D, self).__init__(ksize, stride, pad, cover_all)
         self.outh, self.outw = (None, None) if outsize is None else outsize
-        self._use_integer_scale_forward = False
+        self._use_int_scale_forward = False
 
     def check_type_forward(self, in_types):
         n_in = in_types.size()
@@ -63,7 +63,7 @@ class Unpooling2D(pooling_2d.Pooling2D):
             self.outh // h == self.kh and self.outw // w == self.kw and
             self.ph == 0 and self.pw == 0 and
                 self.sx == self.kh and self.sy == self.kw):
-            self._use_integer_scale_forward = True
+            self._use_int_scale_forward = True
             return self._integer_scale_forward(x[0])
         xp = backend.get_array_module(*x)
         col = xp.tile(x[0][:, :, None, None],
@@ -92,7 +92,7 @@ class Unpooling2DGrad(function_node.FunctionNode):
         self.outh = unpooling2d.outh
         self.outw = unpooling2d.outw
         self.cover_all = unpooling2d.cover_all
-        self._use_integer_scale_forward = unpooling2d._use_integer_scale_forward
+        self._use_int_scale_forward = unpooling2d._use_int_scale_forward
 
     def _integer_scale_forward(self, gy):
         xp = backend.get_array_module(gy)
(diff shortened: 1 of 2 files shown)


mitmul:
Jenkins, test this please

chainer/functions/pooling/unpooling_2d.py
```diff
@@ -44,6 +59,12 @@ def forward(self, x):
         if self.outw is None:
             self.outw = conv.get_deconv_outsize(
                 w, self.kw, self.sx, self.pw, cover_all=self.cover_all)
+        if (self.outh % h == 0 and self.outw % w == 0 and
+            self.outh // h == self.kh and self.outw // w == self.kw and
+            self.ph == 0 and self.pw == 0 and
```
mitmul: Is it difficult to support the case when ph or pw is greater than 0?

chainer/functions/pooling/unpooling_2d.py
```diff
@@ -1,4 +1,9 @@
 import numpy
+import numpy.lib.stride_tricks
+try:
+    import cupy.lib.stride_tricks  # NOQA
+except ImportError:
```
mitmul: Could you change this line to catch the error with `Exception` ?
Because if the CuPy installation is failed, the broken CuPy may emit `AttributeError` when importing is failed. (ref: https://github.com/chainer/chainer/pull/6085)
```suggestion
except Exception:
```

chainer/functions/pooling/unpooling_2d.py
tkerola: Hmm, if we have non-zero padding, then the result will not have a uniform pattern, so I don't think we can speed up this cause with stride tricks, but would need another solution, so I think this is out of the scope of this PR.

tkerola:
@mitmul Thanks for the review! I fixed the import exception and commented about non-zero padding. Can you please check it?

chainer/functions/pooling/unpooling_2d.py
tkerola: Ok, I realized that it is possible to use stride tricks when `pad % 2 == 0`, but I don't think we can use it with odd padding. I'll update the PR.

tkerola: Ok, I added support for even padding.

mitmul:
Jenkins, test this please

chainer/functions/pooling/unpooling_2d.py
mitmul: Great, thanks!

mitmul:
All the tests on the ChainerCI and examples/docs tests on Jenkins have passed. LGTM

tkerola:
Hmm, it seems like the test for group normalization failed - unrelated to this PR. Can you please run the tests again?

tkerola:
Jenkins, test this please

mitmul:
@tkerola This PR passed all the tests. Please only see the ImosCI test jobs basically and only for the example and docs tests, please check the Jenkins tests. Thanks. (FYI, "Jenkins, test this please" triggers ImosCI and Jenkins at the same time)

PR #7117, opened 2019-05-11
[WIP] Support TabularDataset in training utilities

Hakuyume:
merge after #7116


chainer/dataset/tabular/join.py
```diff
@@ -0,0 +1,54 @@
...
+    def keys(self):
+        return tuple(key for dataset in self._datasets for key in dataset.keys)
+
+    @property
+    def mode(self):
+        return self._datasets[0].mode
+
+    def get_examples(self, indices, key_indices):
+        if key_indices is None:
+            key_indices = six.moves.range(len(self.keys))
```
okapies: I think we can directly return results of `get_examples(indices, None)` if `key_indices is None`.

```suggestion
            examples = []
            for d in self._datasets:
                examples.extend(d.get_examples(indices, None))
            return tuple(examples)
```

chainer/dataset/tabular/slice.py
```diff
@@ -0,0 +1,143 @@
...
+        if isinstance(args, tuple):
+            indices, keys = args
+        else:
+            indices = args
+            keys = None
+
+        return Slice(self._dataset, indices, keys)
+
+
+def _as_indices(indices, len_):
```
okapies: I profiled `_as_indices` by using MNIST example (same as #5954) and found that checking *all* values in `indices` degrade the performance. The following code improves by about 5%. I think we need a discussion about the validation spec around it.

```python
    if isinstance(indices, slice) or len(indices) == 0:
        return indices

    if isinstance(indices[0], (bool, np.bool_)):
        if not len(indices) == len_:
            raise ValueError('The number of booleans is '
                             'different from the length of dataset')
        return [i for i, index in enumerate(indices) if index]

    return [(i + len_ if i < 0 else i) for i in indices]
```

chainer/dataset/convert.py
```diff
@@ -211,6 +211,15 @@ def concat_examples(batch, device=None, padding=None):
     if not batch:
         raise ValueError('batch is empty')
 
+    # batch from TabularDataset.fetch
+    if isinstance(batch, tuple):
+        return tuple(
+            to_device(device, _concat_arrays(array, None)) for array in batch)
```
okapies: It seems we can bypass `_concat_arrays` because `array` is just an `ndarray` (tabular array) in this place.

chainer/dataset/convert.py
```diff
@@ -211,6 +211,15 @@ def concat_examples(batch, device=None, padding=None):
...
+    elif isinstance(batch, dict):
+        return {
+            key: to_device(device, _concat_arrays(batch[key], None))
```
okapies: Ditto.

```suggestion
            key: to_device(device, batch[key])
```

chainer/dataset/tabular/slice.py
Hakuyume: I prefer strict checks since the mixture of ints and bools will be confusing for users.
Perhaps, we can provide an environment variable to disable strict checking for expert users (like `CHAINER_TYPE_CHECK`)

FYI: `numpy` checks only the first element of the list to determine the type of list. Actually, numpy accepts mixture of ints and bools.

chainer/iterators/serial_iterator.py
```diff
@@ -74,6 +75,9 @@ def __next__(self):
         if indices is None:
             raise StopIteration
 
+        if isinstance(self.dataset, TabularDataset):
```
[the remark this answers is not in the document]
Hakuyume: As you pointed out, many computation can be cached. But is this computation so critical?

chainer/dataset/convert.py
Hakuyume: `array` is not always an ndarray. It can be a list.

chainer/iterators/serial_iterator.py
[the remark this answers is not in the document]
okapies: Not critical, but `__next__` sit in the inner side of nested loop of training and evaluation, and `isinstance` is relatively slower than other operations. Caching is reasonable and no side-effect I think.

chainer/dataset/tabular/slice.py
okapies: I think we can prevent the mixture by documentation. It seems Chainer checks just the head of list/array for its element in other cases (such like numpy). I like very much the strict manner too, but invoking `isinstance` for every item is little bit overkill in this situation.

chainer/dataset/convert.py
okapies: I see. But `_concat_arrays` is originally for converting list of arrays in one column to an array, and causes another 5% overhead. The case you pointed out should be handled in a separate logic I think.

Hakuyume: Here, I want to convert a list of scalars/ndarrays into an ndarray. I think this behaviour is same as `_concat_arrays`. What is the intention of separated logic?

chainer/dataset/tabular/slice.py
Hakuyume: @beam2d How do you think of this?

Anyway, this modification should be in another PR.

chainer/dataset/tabular/join.py
Hakuyume: I updated in #7116

chainer/dataset/tabular/slice.py
Hakuyume: > FYI: numpy checks only the first element of the list to determine the type of list. Actually, numpy accepts mixture of ints and bools.

Sorry. This was wrong. numpy treat `indices` as index if at least one of elements is integer. Only if all elements are booleans, numpy treat it as booleans. So even if we make the behaviour numpy-compat, we need call `isinstance` for all elements.

Hakuyume:
```python
np.arange(3)[[False, False, True]]  # array([2])
np.arange(3)[[False, False, 1]]  # array([0, 0, 1])
```

Hakuyume: @okapies I add mixed indices support in #7251. If the indices is a list of ints, `isinstance` is called only once.

chainer/dataset/convert.py
okapies: Hmm, it seems there is three cases to handle input:

1. An ndarray (of scalars). It doesn’t need any concatanation. This is the typical case in which `TabularDataset` removes the performance overhead.

2. A list of scalars. It needs a conversion from a list to an ndarray. We just need the following lines in `_concat_arrays` (IMO it should be done in `fetch`):

```python
    if not isinstance(arrays[0], chainer.get_array_types()):
        arrays = numpy.asarray(arrays)
```

3. A list of N-dimension ndarrays. We should convert it to (N+1)-dimension ndarray done in `_concat_arrays` before. But it also can be done in `fetch` (if we update it’s signature as `None -> ndarray`).

Any ideas?

okapies: Note that we have a chance to simplify the implementation of 3rd case if cupy/cupy#2081 is merged.

Hakuyume: In the term of performance,
even if we call concat in `fetch`, we cannot remove the overhead. 
Since the output of `get_examples` can be both list and ndarray. we should handle both cases. So the logic won't change.

In the term of design.
I think users can call `fetch` outside of updater. For example, users may want to check intermediate values for debugging. In that case, we cannot assume that all values can be concatenated.

Hakuyume: In addition, evaluation of some models are not require concatenated data and can accept list of arrays as its input. In that case, forcing concat in `fetch` is not good.

okapies: I see. In this sense, does it make sense for you?

```suggestion
            to_device(device, (
                array if isinstance(array, chainer.get_array_types())
                else _concat_arrays(array, None)
            )) for array in batch)
```

chainer/dataset/convert.py
okapies:
```suggestion
            key: to_device(device, (
                batch if isinstance(batch, chainer.get_array_types())
                else _concat_arrays(batch[key], None)))
```

chainer/dataset/convert.py
Hakuyume: Ah, I see. I understand why you don't want to use `_concat_arrays` as is. I'll make another helper for this situation.

Hakuyume: I added `_as_array` that can bypass `ndarray` as is.

Hakuyume:
I split this PR into #7428 and #7430.

PR #7118, opened 2019-05-12
Add basic activation functions

aksub99:
Added `clipped_relu`, `crelu` and `elu`.
TODO: Add tests, Add docs
Ref. #6423

### DIFF
--- a/chainerx_cc/chainerx/python/routines.cc
+++ b/chainerx_cc/chainerx/python/routines.cc
@@ -1019,7 +1019,7 @@ void InitChainerxActivation(pybind11::module& m) {
           "x"_a,
           "z"_a = 20.0);
     m.def("crelu",
-          [](const ArrayBodyPtr& x, nonstd::optional<int8_t> axis) { return MoveArrayBody(Crelu(Array{x}, axis)); },
+          [](const ArrayBodyPtr& x, int8_t axis) { return MoveArrayBody(Crelu(Array{x}, axis)); },
           "x"_a,
           "axis"_a = 1);
 }
--- a/chainerx_cc/chainerx/python/routines.cc
+++ b/chainerx_cc/chainerx/python/routines.cc
@@ -623,7 +623,7 @@ void InitChainerxManipulation(pybind11::module& m) {
 
 void InitChainerxActivation(pybind11::module& m) {
     m.def("clipped_relu", [](const ArrayBodyPtr& x, Scalar z) { return MoveArrayBody(ClippedRelu(Array{x}, z)); }, "x"_a, "z"_a = 20.0);
-    m.def("crelu", [](const ArrayBodyPtr& x, int8_t axis) { return MoveArrayBody(Crelu(Array{x}, axis)); }, "x"_a, "axis"_a = 1);
+    m.def("crelu", [](const ArrayBodyPtr& x, int8_t axis) { return MoveArrayBody(CRelu(Array{x}, axis)); }, "x"_a, "axis"_a = 1);
     m.def("sigmoid", [](const ArrayBodyPtr& x) { return MoveArrayBody(Sigmoid(Array{x})); }, "x"_a);
     m.def("relu", [](const ArrayBodyPtr& x) { return MoveArrayBody(Relu(Array{x})); }, "x"_a);
     m.def("leaky_relu",
--- a/chainerx_cc/chainerx/python/routines.cc
+++ b/chainerx_cc/chainerx/python/routines.cc
@@ -1015,7 +1015,7 @@ void InitChainerxActivation(pybind11::module& m) {
           "x"_a,
           "slope"_a = 0.2);
     m.def("clipped_relu", [](const ArrayBodyPtr& x, Scalar z) { return MoveArrayBody(ClippedRelu(Array{x}, z)); }, "x"_a, "z"_a = 20.0);
-    m.def("crelu", [](const ArrayBodyPtr& x) { return MoveArrayBody(Crelu(Array{x})); }, "x"_a);
+    m.def("crelu", [](const ArrayBodyPtr& x, nonstd::optional<int8_t> axis) { return MoveArrayBody(Crelu(Array{x}, axis)); }, "x"_a, "axis"_a = 1);
 }
 
 }  // namespace
--- a/chainerx_cc/chainerx/python/routines.cc
+++ b/chainerx_cc/chainerx/python/routines.cc
@@ -9,7 +9,6 @@
 
 #include <nonstd/optional.hpp>
 
-#include "chainerx/routines/activation.h"
 #include "chainerx/array.h"
 #include "chainerx/axes.h"
 #include "chainerx/constant.h"
@@ -21,6 +20,7 @@
 #include "chainerx/macro.h"
 #include "chainerx/routines/activation.h"
 #include "chainerx/routines/arithmetic.h"
+#include "chainerx/routines/activation.h"
 #include "chainerx/routines/binary.h"
 #include "chainerx/routines/connection.h"
 #include "chainerx/routines/creation.h"
@@ -1014,8 +1014,14 @@ void InitChainerxActivation(pybind11::module& m) {
           [](const ArrayBodyPtr& x, Scalar slope) { return MoveArrayBody(LeakyRelu(Array{x}, slope)); },
           "x"_a,
           "slope"_a = 0.2);
-    m.def("clipped_relu", [](const ArrayBodyPtr& x, Scalar z) { return MoveArrayBody(ClippedRelu(Array{x}, z)); }, "x"_a, "z"_a = 20.0);
-    m.def("crelu", [](const ArrayBodyPtr& x, nonstd::optional<int8_t> axis) { return MoveArrayBody(Crelu(Array{x}, axis)); }, "x"_a, "axis"_a = 1);
+    m.def("clipped_relu",
+          [](const ArrayBodyPtr& x, Scalar z) { return MoveArrayBody(ClippedRelu(Array{x}, z)); },
+          "x"_a,
+          "z"_a = 20.0);
+    m.def("crelu",
+          [](const ArrayBodyPtr& x, nonstd::optional<int8_t> axis) { return MoveArrayBody(Crelu(Array{x}, axis)); },
+          "x"_a,
+          "axis"_a = 1);
 }
 
 }  // namespace
--- a/chainerx_cc/chainerx/python/routines.cc
+++ b/chainerx_cc/chainerx/python/routines.cc
@@ -20,7 +20,6 @@
 #include "chainerx/macro.h"
 #include "chainerx/routines/activation.h"
 #include "chainerx/routines/arithmetic.h"
-#include "chainerx/routines/activation.h"
 #include "chainerx/routines/binary.h"
 #include "chainerx/routines/connection.h"
 #include "chainerx/routines/creation.h"
--- a/chainerx_cc/chainerx/python/routines.cc
+++ b/chainerx_cc/chainerx/python/routines.cc
@@ -622,6 +622,8 @@ void InitChainerxManipulation(pybind11::module& m) {
 }
 
 void InitChainerxActivation(pybind11::module& m) {
+    m.def("clipped_relu", [](const ArrayBodyPtr& x, Scalar z) { return MoveArrayBody(ClippedRelu(Array{x}, z)); }, "x"_a, "z"_a = 20.0);
+    m.def("crelu", [](const ArrayBodyPtr& x, int8_t axis) { return MoveArrayBody(Crelu(Array{x}, axis)); }, "x"_a, "axis"_a = 1);
     m.def("sigmoid", [](const ArrayBodyPtr& x) { return MoveArrayBody(Sigmoid(Array{x})); }, "x"_a);
     m.def("relu", [](const ArrayBodyPtr& x) { return MoveArrayBody(Relu(Array{x})); }, "x"_a);
     m.def("leaky_relu",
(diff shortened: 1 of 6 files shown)


chainerx_cc/chainerx/routines/CMakeLists.txt
```diff
@@ -1,4 +1,5 @@
 add_library(chainerx_routines STATIC
+    activation.h
```
kshitij12345: Should be `activation.cc`.

aksub99: Thanks! Made the change.

aksub99:
I'm not exactly sure how `elu` can be defined in `activation.cc` using the available device methods. Any ideas will be helpful.

kshitij12345:
I guess https://github.com/chainer/chainer/blob/fc97d355d75b04b31cf88d46f2949b61a391167f/chainerx_cc/chainerx/routines/math.cc#L671

can be used.

However unsure about if this is better than a custom kernel?

hvy:
It depends on how common these activation functions are among researchers and developers today and with that in mind, I think it's fine reusing existing routines. Or is it the case that recent models relying on these functions?

aksub99:
@hvy As far as I am aware, these functions might not be too commonly used and reusing the device methods might be okay as you suggested. What are your views @kshitij12345 ?

kshitij12345:
@hvy knows much better, so I'll one up his view.
However, I'll like to make a feature request for in place version of `relu`. But i guess, it is out of scope for this PR?

hvy:
In that case, proceeding with reusing existing ones sound fine for now.

As for the in-place version of ReLU, feel free to open a new issue, preferably with motivations or some concrete example.

aksub99:
I guess `elu` can't be defined using the available device methods. I did try using `IfGreaterElse`, but `elu` would require `IfGreaterElseASAS` to be defined, which is not available as yet. Should I go ahead and implement `IfGreaterElseASAS` or define a custom kernel?
If there is a way to define it using the available device methods, please let me know.

aksub99:
I'm not sure if `elu` can be defined using the existing device methods. I did try using `IfGreaterElse` and `IfLessElse` but we would then require them to be defined with parameters being (Array, Scalar, Array, Array). Do I go ahead and implement `IfGreaterElseASAA` or is it better to go for a custom kernel?
Please let me know if the definition is possible using the existing device methods itself.

chainerx_cc/chainerx/routines/activation.cc
```diff
@@ -0,0 +1,41 @@
...
+#include "chainerx/kernels/math.h"
+#include "chainerx/macro.h"
+#include "chainerx/routines/creation.h"
+#include "chainerx/routines/math.cc"
+#include "chainerx/routines/math.h"
+#include "chainerx/routines/manipulation.h"
+#include "chainerx/routines/type_util.h"
+#include "chainerx/scalar.h"
+#include "chainerx/shape.h"
+#include "chainerx/strides.h"
```
hvy: There are some redundant includes. Could you make sure to only include the ones that are actually required in this file?

chainerx_cc/chainerx/routines/activation.cc
```diff
@@ -0,0 +1,41 @@
...
+
+namespace chainerx {
+Array ClippedRelu(const Array& x, Scalar z) {
+    return Minimum(Maximum(0, x), z);
+}
+
+Array Crelu(const Array& x, nonstd::optional<int8_t> axis) {
```
hvy: I think Crelu can be implement with a single memory allocation similar to `F.crelu` in Chainer. Could you write a TODO to optimize the implementation? You are of course welcome to implement it that way if you'd like to.

chainerx_cc/chainerx/routines/activation.cc
```diff
@@ -0,0 +1,41 @@
...
+}
+
+Array Elu(const Array& x, Scalar alpha) {
+    Array zero = ZerosLike(x, x.device());
+    Array out = ZerosLike(x, x.device());
+    IfGreaterElseImpl(x, zero, x, alpha * (Exp(x) - 1), out);
+    return out;
+}
+
+}
```
hvy: Could you fix these `No newline at end of file`?

chainerx_cc/chainerx/routines/activation.h
```diff
@@ -0,0 +1,18 @@
...
+
+namespace chainerx {
+
+    Array ClippedRelu(const Array& x, Scalar z);
+
+    Array Crelu(const Array& x, nonstd::optional<int8_t> axis);
+
+    Array Elu(const Array& x, Scalar alpha);
+
+}
```
hvy: Ditto.

chainerx_cc/chainerx/python/routines.cc
```diff
@@ -9,6 +9,7 @@
 
 #include <nonstd/optional.hpp>
 
+#include "chainerx/routines/activation.h"
```
hvy: The includes should be ordered lexicographically. Could you fix it? Usually, it's fixed automatically if you apply clang-format 🤔

chainerx_cc/chainerx/python/routines.cc
```diff
@@ -1002,9 +1003,16 @@ void InitChainerxPooling(pybind11::module& m) {
           py::arg("pad_mode") = "ignore");
 }
 
+void InitChainerxActivation(pybind11::module& m) {
+    m.def("clipped_relu", [](const ArrayBodyPtr& x, Scalar z) { return MoveArrayBody(ClippedRelu(Array{x}, z)); }, py::arg("x"), py::arg("z") = 20.0);
+    m.def("crelu", [](const ArrayBodyPtr& x, nonstd::optional<int8_t> axis) { return MoveArrayBody(Crelu(Array{x}, axis)); }, py::arg("x"), py::arg("axis") = 1);
```
hvy: Chainer's `F.crelu` does not work for `axis=None` so we can make the `axis` argument here a regular `int8_t` without being an optional. Let me know if I'm wrong.

chainerx_cc/chainerx/routines/activation.cc
```diff
@@ -0,0 +1,41 @@
...
+Array Crelu(const Array& x, nonstd::optional<int8_t> axis) {
+    std::vector<Array> c{x, Negative(x)};
+    Array concat = Concatenate(c, axis);
+    return Relu(concat);
+}
+
+Array Elu(const Array& x, Scalar alpha) {
+    Array zero = ZerosLike(x, x.device());
+    Array out = ZerosLike(x, x.device());
+    IfGreaterElseImpl(x, zero, x, alpha * (Exp(x) - 1), out);
```
hvy: We have a `chainerx::Expm1` now. Could you use it here?

hvy: Does this line compile? `IfGreaterElseImpl` is "hidden" inside `math.cc` and the signature won't match. I think you actually need to implement a version of `IfGreaterElse` that matches our needs here.

hvy:
I wrote some comments if you could take a look. Oh and please note that the Travis CI results are available from the link at the bottom of this PR. You'll see that there are style errors there.

aksub99:
```
/home/akshay/chainer/chainerx_cc/chainerx/routines/misc.cc:113:7: error: ‘chainerx::Array chainerx::{anonymous}::IfGreaterElse(const chainerx::Array&, const chainerx::Array&, const chainerx::Array&, const chainerx::Array&, const chainerx::Array&)’ defined but not used [-Werror=unused-function]
 Array IfGreaterElse(const Array& x1, const Array& x2, const Array& pos, const A
       ^
```
I get this error when I tried defining `IfGreaterElse` in `misc.cc` and `misc.h`. What could be the reason for this?

kshitij12345:
Your definition of `IfGreaterElse` is in the anonymous namespace {`chainerx::{anonymous}::IfGreaterElse`} and hence it lives there, meaning it is visible only inside the translation unit.
Just move it outside the `anonymous` namespace and it should ideally compile.

aksub99:
Thanks @kshitij12345 for the reply.
@hvy Could you let me know what would be the best option here?

chainerx_cc/chainerx/routines/misc.cc
```diff
@@ -116,6 +116,11 @@ void MaximumImpl(const Array& x1, const Array& x2, const Array& out) { IfGreater
 
 }  // namespace
 
+Array IfGreaterElse(const Array& x1, const Array& x2, const Array& pos, const Array& neg, const Array& out) {
+    IfGreaterElseImpl(x1, x2, pos, neg, out);
+    return out;
+}
```
kshitij12345: I guess the signature of the function should be
`void IfGreaterElse(const Array& x1, const Array& x2, const Array& pos, const Array& neg, const Array& out)`
with current implementation we mostly return a copy of `out` as well as alter the referenced argument as well. (Not sure)

Or alternatively don't take `out` as input, but initialise `out` in function and return it.
Current implementation is a bit confusing to me.
@hvy your opinion?

aksub99:
@kshitij12345 Any idea why the backward tests are failing?

kshitij12345:
On a quick look, not sure why it is failing assuming that all the used functions have correct `backward` implemented.

However I'll suggest that you use, 
https://github.com/chainer/chainer/blob/7710c442d3ada3723da0b036879099a625f02f31/tests/chainerx_tests/op_utils.py#L105

which will allow you to test against existing `chainer.functions`.
Similar to 
https://github.com/chainer/chainer/blob/7710c442d3ada3723da0b036879099a625f02f31/tests/chainerx_tests/unit_tests/routines_tests/test_pooling.py#L29

chainerx_cc/chainerx/routines/misc.cc
hvy: For Elu, we're actually looking for a ASAA? Since zero can be a scalar. But AAAA is also fine for now if that allows us to reuse existing code. And, if this is to be exposed as a routine, yes out should be allocated inside the routine and returned.

hvy:
I'm wondering if we should separate Elu from this PR to keep it simple. I discussed the exposer of IfGreaterElse and we agreed it's fine to expose it as a routine, though preferably in a separate PR (also separate from Elu). It should contain Python bindings and tests. What do you think? After that PR is merged, we can continue with Elu and leaky relu in https://github.com/chainer/chainer/pull/7351.

aksub99:
@hvy I agree it might be simpler if `Elu` is separated and `IfGreaterElse` as well. Will split them accordingly.

aksub99:
@hvy I agree that separating `IfGreaterElse` from this PR would simplify things. Will split them accordingly.

kshitij12345:
Actually do we really need to expose `IfGreaterElse` with Python Bindings? I am sorry that I forgot that we have `Where` implemented which I assume we can use here with `>` operator with `Zeros_Like` since `>` does not support comparison against a `Scalar` (which would be nice to have). `IfGreaterElse` is essentially a special case of `Where`
Your thoughts? @aksub99 @hvy

aksub99:
Thanks for the idea @kshitij12345 ! I guess `Where` could be used for `Elu` as well as `LeakyRelu`  that I'm implementing in #7351 . @hvy Do you think it's advisable to use `Where` for these routines?

aksub99:
I have removed `Elu` from this PR as suggested. I will Create a new PR adding it.

hvy:
Right, as suggested above, we should be able to use `Where` for `Elu`. Sorry for the delayed response. You could add it back to this PR if you'd like but sending a separate PR would also be fine, or preferred.

aksub99:
I have modified test parametrizations similar to `Relu` and have created a separate PR for `Elu`. The tests are still having some `dtype` mismatch errors that I'm not sure how to fix.

hvy:
The CI is failing because it won't compile. Can you fix it?

aksub99:
The `clipped_relu` tests fail for `bool` type inputs. `Chainer's` `clipped_relu` also does not support `bool` type inputs. So, can these tests be avoided?

hvy:
Do you mean when the scalar `z` is a `bool`? Can you cast it according to the other input?

hvy:
Sorry but could you also resolve the conflicts?

hvy:
Thanks for resolving the conflicts but is it ready for review? It seems like the boolean scalar issue is still there, am I correct?

aksub99:
Yes. I'm still working on resolving the boolean errors. Will commit when
it's ready.

hvy:
Okay, no hurries but I just wanted to confirm. Thank you.

aksub99:
@hvy I have fixed the errors. Please review:)

hvy:
Jenkins, test this please.

hvy:
Jenkins, test this please.

chainerx_cc/chainerx/routines/activation.cc
```diff
@@ -18,13 +15,29 @@
...
 
 namespace chainerx {
 
+Array ClippedRelu(const Array& x, Scalar z) {
+    Dtype dtype = internal::GetMathResultDtype(x.dtype());
+    const Array& x_cast = x.dtype() == dtype ? x : x.AsType(dtype);
+    return Minimum(Maximum(0, x_cast), z);
+}
+
+Array Crelu(const Array& x, int8_t axis) {
```
hvy: Maybe it should be CRelu, according to existing ChainerX "ReLU-conventions"? Not a strong opinion and I can fix it in a separate PR if you want.

aksub99: Sure @hvy . I'll fix it.

hvy:
Jenkins, test this please.

hvy:
Failure is unrelated. LGTM!

PR #7079, opened 2019-05-08
[WIP] Split math.cc

asi1024:
Split large math.cc into several files.

- arithmetic.cc
  - chainerx.negative
  - chainerx.add
  - chainerx.subtract
  - chainerx.multiply
  - chainerx.divide

- explog.cc
  - chainerx.exp
  - chainerx.log
  - chainerx.log10

- reduction.cc
  - chainerx.sum
  - chainerx.logsumexp

- trigonometric.cc
  - chainerx.sin
  - chainerx.cos
  - chainerx.tan
  - chainerx.arcsin
  - chainerx.arccos
  - chainerx.arctan
  - chainerx.arctan2

- hyperbolic.cc
  - chainerx.sinh
  - chainerx.cosh
  - chainerx.tanh
  - chainerx.arcsinh
  - chainerx.arccosh

- rounding.cc
  - chainerx.ceil
  - chainerx.floor

- misc.cc
  - chainerx.maximum
  - chainerx.clip
  - chainerx.sqrt
  - chainerx.square

- binary.cc
  - chainerx.bitwise_and
  - chainerx.bitwise_or
  - chainerx.bitwise_xor

- activation.cc
  - chainerx.relu
  - chainerx.log_softmax

PR #7126, opened 2019-05-13
Add `expm1` and `exp2`

aksub99:
Add `expm1` and `exp2` to chainerx.
TODO: Add tests, Add docs
Ref. #6423


kshitij12345:
How about a custom kernel using
https://en.cppreference.com/w/cpp/numeric/math/expm1
which should be more stable.

hvy:
Yes we should as suggested by @kshitij12345 implement these routines using the standard library functions, for the reason already stated. Thanks for pointing that out, and sorry @aksub99 but could you modify your implementation accordingly? Note that this PR then shouldn't depends on #6496 .

aksub99:
Thanks for the inputs @kshitij12345 @hvy . Will modify the implementation accordingly.

kshitij12345:
Just a heads up. Maybe you should wait till #7079 and its related PR get in, otherwise there might be conflicts.

aksub99:
The backward implementation of `Exp2` requires `Log` to be defined for `Scalar` values. Should I implement that in this PR itself?

niboshi:
@aksub99 Are you referring to `Log(2.0)`?
This part is just a constant. You can simply write `std::log(2.0)`.

chainerx_cc/chainerx/cuda/numeric.cuh
```diff
@@ -62,8 +74,6 @@ CHAINERX_DEFINE_CUDA_FLOAT16_FALLBACK_UNARY(Arcsin, std::asin)
 CHAINERX_DEFINE_CUDA_FLOAT16_FALLBACK_UNARY(Arccos, std::acos)
 CHAINERX_DEFINE_CUDA_FLOAT16_FALLBACK_UNARY(Arctan, std::atan)
 CHAINERX_DEFINE_CUDA_FLOAT16_FALLBACK_UNARY(Exp, std::exp)
-CHAINERX_DEFINE_CUDA_FLOAT16_FALLBACK_UNARY(Expm1, std::expm1)
-CHAINERX_DEFINE_CUDA_FLOAT16_FALLBACK_UNARY(Exp2, std::exp2)
```
niboshi: Doesn't this macro work?
As `std::expm1` and `std::exp2` have `float` overloads, we don't have to use special versions like `std::expm1f`.

aksub99: I made this change because I was getting the following error earlier for `float` type inputs.
```
home/akshay/chainer/chainerx_cc/chainerx/cuda/numeric.cuh(40): error: calling a constexpr __host__ function("expm1") from a __device__ function("Expm1") is not allowed. The experimental flag '--expt-relaxed-constexpr' can be used to allow this.

/home/akshay/chainer/chainerx_cc/chainerx/cuda/numeric.cuh(49): error: calling a constexpr __host__ function("exp2") from a __device__ function("Exp2") is not allowed. The experimental flag '--expt-relaxed-constexpr' can be used to allow this.
```

niboshi: I see. Thank you for clarification!

niboshi:
LGTM.
Could you mark the PR as ready for review?

aksub99:
Sure. Writing tests for the implementations is still pending though.

niboshi:
Oh, I missed that.

niboshi:
Thanks!
Jenkins, test this please

niboshi:
Jenkins, test this please

PR #7141, opened 2019-05-14
[chainerx] add random creation functions

kshitij12345:
This PR intends to add random sampling functions.

### DIFF
--- a/chainerx_cc/chainerx/kernels/misc.h
+++ b/chainerx_cc/chainerx/kernels/misc.h
@@ -15,6 +15,13 @@ public:
     virtual void Call(const Array& out, Scalar value) = 0;
 };
 
+class UniformKernel : public Kernel {
+public:
+    static const char* name() { return "Uniform"; }
+
+    virtual void Call(const Array& out, Scalar a, Scalar b) = 0;
+};
+
 // Casts the elements from one array to the other dtype, and store into the other.
 class AsTypeKernel : public Kernel {
 public:
--- a/chainerx_cc/chainerx/native/native_device/fill.cc
+++ b/chainerx_cc/chainerx/native/native_device/fill.cc
@@ -1,6 +1,7 @@
 #include "chainerx/native/native_device.h"
 
 #include <cstdint>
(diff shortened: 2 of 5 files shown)


kshitij12345:
As of now only **Native Kernel** for `uniform` is implemented.

I would request you to review and let me know if it is on the right track or has it gone terribly wrong.

Also as for **Cuda Kernel** idea is to use `cuRAND`. Would like to know the opinion regarding the same.

hvy:
Thanks for the PRs as always. I wouldn't say it's terribly wrong nor is using cuRAND a bad idea (CuPy uses it so it would be the first goto). We need to introduce some mechanism to control the seed.

kshitij12345:
> We need to introduce some mechanism to control the seed.
True. How does `chainer` support that? Similar to `numpy`?

crcrpar:
AFAIK, there is no such function to set seed in Chainer.
Before ChainerX, it is enough to set seed in NumPy (& CuPy).

kshitij12345:
Oh thank you very much.
Digging a bit further,
https://stackoverflow.com/questions/5836335/consistently-create-same-random-numpy-array/5837352#5837352
There should also be something akin to `RandomState` .

hvy:
I guess we first need to decide who `RandomState` is owned by and how it it accessed by multiple threads. I'd be open for suggestion but could otherwise consider this with the team and get back to you. It depends on the priority but my understanding is that this is not urgent until we'd e.g. want to implement other random routines such as dropout in C++.

I'm not sure if it should be configurable via Chainer. It's not straight forward to maintain consistency between e.g. Chainer and ChainerX if the same configuration can be controlled from both `chainer` and `chainerx`.

kshitij12345:
Oh.
As of what I thought, user can create multiple `RandomState` objects and one will be a global instance which will be wrapped for `chainerx.normal/chainerx.uniform`. Management of other instances will fall under user. 
But subtlety of the managing global instance looks to be a deeper issue. Do discuss with the team as per convenience.
Thank You.

PR #7145, opened 2019-05-15
`get_device` returns `CpuDevice` for `cuda.DummyDevice`

niboshi:
Before this PR, it returned a seemingly invalid device:

```
>>> chainer.get_device(chainer.backends.cuda.DummyDevice)
<GpuDevice (cupy):-1>
```


hvy:
Jenkins, test this please.

hvy:
LGTM, let me merge this PR once merging is allowed again.

niboshi:
I think this PR is not a good way to handle `DummyDevice`. We should move away from it because Chainer now has its own device abstraction layer.

PR #7146, opened 2019-05-15
Add a flag to detect access to grad before calling `cleargrad`

niboshi:
Currently `Parameter.grad` is initialized with NaN if `cleargrad` is not called. This is to notify users that they mistakenly forgot to call `cleargrad`.

This PR is an attempt to detect that by a flag.


chainer/variable.py
```diff
@@ -1932,10 +1948,12 @@ def as_array(obj):
     return obj
 
 
-def _recover_parameter(data, name, grad, initializer, update_rule, device):
+def _recover_parameter(
+        data, name, grad, grad_valid, initializer, update_rule, device):
```
niboshi: Does it cause snapshot incompatibility?

chainer/variable.py
```diff
@@ -948,6 +963,12 @@ def grad(self):
         and error.
 
         """
+        if not self._grad_valid:
+            raise RuntimeError(
+                'Cannot retrieve Variable.grad because it\'s not valid. '
```
Crissman:
```suggestion
                'Cannot retrieve Variable.grad because it must be set '
```

chainer/variable.py
```diff
@@ -948,6 +963,12 @@ def grad(self):
...
+                'Either it must be set beforehand or Variable.cleargrad() '
```
Crissman:
```suggestion
                'beforehand or Variable.cleargrad() must be called.')
```

chainer/variable.py
```diff
@@ -948,6 +963,12 @@ def grad(self):
...
+                'must be called.')
```
Crissman:
``````

Crissman:
Please look at comments and resolve conflicts. 🙇

chainer/variable.py
niboshi: How about "... because neither it's set beforehand nor Variable.cleargrad() is called" ?

Crissman: `Neither... nor...` is difficult to understand English. I think my first revision is more clear.

niboshi: Hmm.
I feel your version sounds to suggest the cause is "it must be ..." and does not state whether "it actually is".

niboshi: How about removing "because it's not valid" from the original version, leaving two sentences?

Crissman: So,
```Cannot retrieve Variable.grad. Either it must be set beforehand or Variable.cleargrad() must be called.```

I'm ok with that.

Crissman: Actually, for more clarity, should be,

```Cannot retrieve Variable.grad. Either it must be set manually or Variable.cleargrad() must be called beforehand.```

niboshi: Thank you for correction. Fixed as such.

Crissman:
Jenkins, test this please.

Crissman:
Jenkins failure looks unrelated.

PR #7148, opened 2019-05-15
Avoid using `cuda.DummyDevice` and `cuda.get_device_from_array`

niboshi:
Now that Chainer has its own device abstraction, we should avoid using `DummyDevice` outside `backends.cuda` module.

Also replaces `cuda.get_device_from_array` with `backend.get_device_from_array`.
Some uses are left for respective reasons. (cupy-specific code, being handled in another PR (#6982), etc.)

### DIFF
--- a/chainer/backend.py
+++ b/chainer/backend.py
@@ -30,8 +30,9 @@ def _contains_nan(x):
 
     """
     if x.dtype.kind in ('f', 'c'):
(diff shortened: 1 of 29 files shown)


hvy:
Jenkins, test this please.

hvy:
LGTM!

PR #7149, opened 2019-05-15
Add `Device.name` property

niboshi:
I think it's useful.

This PR also improves backend device tests.


kmaehashi:
Do you think the attribute name `name` is appropriate? I think we're currently calling it `device specifier`.
https://docs.chainer.org/en/latest/reference/generated/chainer.get_device.html?highlight=get_device#chainer.get_device

niboshi:
"Device specifier" is more general. It can be `int`, for example.
`Device.name` is meant to be more like unique identifiers.

C.f. `chainerx.Device` also has the same attribute.
```
>>> dev = chainerx.get_device('native')
>>> dev.name
'native:0'
```

kmaehashi:
OK, that makes sense.
Jenkins, test this please.

kmaehashi:
@niboshi Could you resolve conflicts?

niboshi:
Resolved.

hvy:
Seems to be conflicting again.

niboshi:
Resolved.

kmaehashi:
pfnCI, test this please.

PR #7150, opened 2019-05-15
Add TabularDataset.transform/transform_batch

Hakuyume:
~merge after #7115~
~merge after #7275~


Hakuyume:
jenkins, test this please.

Hakuyume:
jenkins, test this please

Hakuyume:
jenkins, test this please

chainer/dataset/tabular/_transform.py
```diff
@@ -0,0 +1,120 @@
...
+            if self._dataset.mode is tuple:
+                out_example = self._transform(*in_example)
+            elif self._dataset.mode is dict:
+                out_example = self._transform(
+                    **dict(six.moves.zip(self._dataset.keys, in_example)))
+
+            if not isinstance(out_example, (tuple, dict)):
+                out_example = out_example,
+            if isinstance(out_example, tuple):
+                self._mode = tuple
```
beam2d: When `_transform` is implemented incorrectly and it returns an object of different types depending on the example, `self._mode` may be changed over time. Is it intended? If not, how about raising an error in such a case?

chainer/dataset/tabular/_transform.py
```diff
@@ -0,0 +1,120 @@
...
+        else:
+            len_ = len(indices)
+
+        if key_indices is None:
+            key_indices = six.moves.range(len(self._keys))
+
+        in_examples = self._dataset.get_examples(indices, None)
+
+        if self._dataset.mode is tuple:
+            self._mode = tuple
```
beam2d: Maybe this line should be removed? (ditto for the elif branch)

chainer/dataset/tabular/_transform.py
```diff
@@ -0,0 +1,120 @@
+import six
+
+from chainer.dataset.tabular import tabular_dataset
+
+
+class _Transform(tabular_dataset.TabularDataset):
+
+    def __init__(self, dataset, keys, transform):
```
beam2d: Is it possible/reasonable to make `keys` automatically inferred like `mode`?

Hakuyume: No. If `transform` returns a `tuple`, we cannot know the name of keys. If `transform` returns a `dict`, we cannot know the order of keys.

chainer/dataset/tabular/_transform.py
Hakuyume: Current implementation assumes `_transform` is correctly implemented. OK, I'll add a check.

chainer/dataset/tabular/_transform.py
Hakuyume: Yes. Thank you.

Hakuyume:
@beam2d Thank you for your comments. I updated my code.

chainer/dataset/tabular/_transform.py
beam2d: That makes sense. I still think the information of key names and the definition of transform are tightly connected, so it'd be good if the user only needs to specify `transform` and then the keys are automatically set. How about passing keys via an attribute/property of `transform`?

Hakuyume: > How about passing keys via an attribute/property of transform?

Do you mean the following?
```python
def transform(a, b, c):
    d = a + b
    e = b + c
    return d, e

transform.keys = ('d', 'e')
dataset.transform(transform)
```
I have two concerns.
1. Is is common to set an attribute to a function instance?
2. How can users specify `keys` for a lambda function?

Hakuyume: Perhaps, we should think about this PR (https://github.com/chainer/chainer/pull/7361) at the same time.

beam2d: I thought a similar approach as Trainer extensions, like:
```py3
class MyTransform:
    keys = 'd', 'e'
    def __call__(self, a, b, c):
        ...

# or

@tabular.make_transform(keys=('d', 'e'))  # it sets `keys` attribute
def my_transform(a, b, c):
    ...
```
For lambda functions and the cases when the user wants to override the `keys` attribute provided by the transform impl, we can keep optional `keys` argument to be specified at `transform` (this is also similar to `Trainer.extend`).

I haven't checked #7361 yet, so I will look through it. Thanks for pointing that out.

Hakuyume: If we provide a decorator `make_transform`, users can use it for lambda or overriding `keys`.
```python
tabular.make_transform(keys=('d', 'e'))(lambda a, b, c: ...)
```

Hakuyume: I think it will be confusing if there are more than one way to specify `keys`.

Hakuyume: `SimpleDataset` in https://github.com/chainer/chainer/pull/7276 also requires `key` and data of a column. This API also should be considered.

beam2d: I want to clarify my point of view. I'm mainly thinking about reusing the same transform function at multiple places (or by multiple users), like "let's use the same preprocessing as was done in the XXX project." In this case, there is no more than one way to specify `keys`; there are just two choices of keys for the user: the default keys, or custom (overriding) keys. The default keys exist for pushing "convention over configuration".

When the author of the transform just uses it immediately, there is no difference between the default keys and the overriding keys in effect, so you may feel there are multiple ways to do the same thing. Lambda function usage is an extreme case. I think just using the `keys` option of `transform` is much more natural than applying the decorator to the lambda, so I believe it won't be so confusing (but it's just my intuition).

As for `SimpleDataset.add_column` in #7276, who calls this method is always the same as who designs the dataset, so it's natural to always specify the key here. I don't think it affects the design choice for `transform`.

Hakuyume: > I want to clarify my point of view. I'm mainly thinking about reusing the same transform function at multiple places (or by multiple users), like "let's use the same preprocessing as was done in the XXX project."

Unlike trainer extensions, I think reusing of transform functions is not so common. So, `keys` attribute is not so useful. In most cases, the author of transform is same as the user.

beam2d: I see. If it is not so bothersome in most cases, then it's fine to leave it as is.

Hakuyume: If we find it is bothersome, we can add that functionality.

beam2d:
Jenkins, test this please.

chainer/dataset/tabular/_transform.py
beam2d: OK, thanks for discussion!

beam2d:
LGTM. Jenkins failure is not related.

PR #7153, opened 2019-05-16
Support device specifiers in static graph example 

hvy:
Part of #6661.

All of the sub-examples of the static graph optimization example, specifying a ChainerX device fail and need separate fixes. For instance, this error is raised. https://github.com/chainer/chainer/blob/master/chainer/graph_optimizations/static_graph.py#L738

### DIFF
--- a/examples/static_graph_optimizations/cifar/train_cifar.py
+++ b/examples/static_graph_optimizations/cifar/train_cifar.py
@@ -49,15 +49,14 @@ def main():
     args = parser.parse_args()
 
     device = chainer.get_device(args.device)
+    if device.xp is chainerx:
+        sys.stderr.write('This example does not support ChainerX devices.\n')
+        sys.exit(1)
 
     print('Device: {}'.format(device))
     print('# Minibatch-size: {}'.format(args.batchsize))
     print('# epoch: {}'.format(args.epoch))
     print('')
-
-    if device.xp is chainerx:
-        sys.stderr.write('This example does not support ChainerX devices.\n')
-        sys.exit(1)
 
     device.use()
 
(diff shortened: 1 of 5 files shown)


examples/static_graph_optimizations/cifar/train_cifar.py
```diff
@@ -21,29 +23,44 @@
...
 
-    print('GPU: {}'.format(args.gpu))
+    device = chainer.get_device(args.device)
+
+    print('Device: {}'.format(device))
     print('# Minibatch-size: {}'.format(args.batchsize))
     print('# epoch: {}'.format(args.epoch))
     print('')
 
+    if device.xp is chainerx:
```
niboshi: How about doing this check before printing training configurations?

examples/static_graph_optimizations/cifar/train_cifar.py
```diff
@@ -80,19 +94,21 @@ def main():
...
 
     # Reduce the learning rate by half every 25 epochs.
     trainer.extend(extensions.ExponentialShift('lr', 0.5),
                    trigger=(25, 'epoch'))
 
     # Dump a computational graph from 'loss' variable at the first iteration
     # The "main" refers to the target link of the "main" optimizer.
-    trainer.extend(extensions.DumpGraph('main/loss'))
+    # TODO(hvy): Support ChainerX
+    if not isinstance(device, chainer.backend.ChainerxDevice):
```
niboshi: `device.xp is chainerx` is better because it's easier to comprehend for users.

examples/static_graph_optimizations/cifar/train_cifar.py
hvy: Fixed.

examples/static_graph_optimizations/cifar/train_cifar.py
hvy: Fixed.

hvy: I guess you want to change this style for all xp instances such as https://github.com/chainer/chainer/blob/master/examples/imagenet/train_imagenet.py#L155 ?

hvy:
PTAL.

niboshi:
Jenkins, test this please

examples/static_graph_optimizations/cifar/train_cifar.py
niboshi: Perhaps.

niboshi:
LGTM

PR #7170, opened 2019-05-17
Remove backslashes to continue lines of link targets

toslunar:
Follow the example in ReST reference:
```
External hyperlinks, like `Python
<http://www.python.org/>`_.
```
http://docutils.sourceforge.net/docs/user/rst/quickref.html#external-hyperlink-targets


niboshi:
Jenkins, test this please

niboshi:
LGTM

Issue #7172, opened 2019-05-17
`Link.serialize` does not support ChainerX


hvy:
`Link.serialize` assumes that the ndarrays held by the link are either NumPy or CuPy ndarrays and breaks when they're ChainerX ndarrays. 

#### Relevant code
https://github.com/chainer/chainer/blob/master/chainer/link.py#L661-L664

#### Reproduction
```python
import numpy as np
import chainer as ch
import chainerx as chx

# Save link parameters
l = ch.Link()
with l.init_scope():
    l.x = ch.Parameter(np.random.rand(2, 3))
ch.serializers.npz.save_npz('model.npz', l)

# Try to load parameters and fail
l = ch.Link()
with l.init_scope():
    l.x = ch.Parameter()
l.to_device('native')

ch.serializers.npz.load_npz('model.npz', l, path='')  # AttributeError: 'chainerx.ndarray' object has no attribute 'set'
```

PR #7175, opened 2019-05-17
`Link.serialize` to support ChainerX

hvy:
Fixes https://github.com/chainer/chainer/issues/7172.

### DIFF
--- a/tests/chainer_tests/test_link.py
+++ b/tests/chainer_tests/test_link.py
@@ -2466,4 +2466,42 @@ class TestLinkOverrideToDeviceMethods(unittest.TestCase):
         assert l.child.to_method_called == 1
 
 
+class TestSerialize(unittest.TestCase):
+    def setUp(self):
+        self.array = numpy.array([1, 2, 3], dtype=numpy.float32)
+        self.serializer = mock.MagicMock(return_value=self.array)
+
+        link = chainer.Link()
+        with link.init_scope():
+            link.x = chainer.Parameter()
+            link.y = chainer.Parameter()
+        link.add_persistent('z', None)
+        self.link = link
+
+    def check_serialize(self, device_spec):
+        array = self.array
+        link = self.link
+        serializer = self.serializer
+
+        link.to_device(device_spec)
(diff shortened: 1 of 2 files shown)


tests/chainer_tests/test_link.py
```diff
@@ -2466,4 +2466,43 @@ def test_to_intel64(self):
         assert l.child.to_method_called == 1
 
 
+class TestSerialize(unittest.TestCase):
+    def setUp(self):
+        self.array = numpy.array([1, 2, 3], dtype=numpy.float32)
+        self.serializer = mock.MagicMock(return_value=self.array)
+
+        # Create an empty link to populate.
```
niboshi: Is this comment appropriate? `link` is not empty a few lines below.

tests/chainer_tests/test_link.py
```diff
@@ -2466,4 +2466,43 @@ def test_to_intel64(self):
...
+        link.to_device(device_spec)
+        link.serialize(serializer)
+
+        self.assertEqual(serializer.call_count, 3)
+        cpu_device = chainer.backend.CpuDevice()
+        numpy.testing.assert_array_equal(cpu_device.send(link.x.array), array)
+        numpy.testing.assert_array_equal(cpu_device.send(link.y.array), array)
+        numpy.testing.assert_array_equal(cpu_device.send(link.z), array)
+
+    def test_serialize_numpy(self):
```
niboshi: Why not use `inject_backend_config`?

tests/chainer_tests/test_link.py
hvy: Removed the comment.

tests/chainer_tests/test_link.py
hvy: Thanks, that actually helped me find a bug when the device is a ChainerX CUDA device. Fixed that bug and the test.

hvy:
Thanks for the comments. PTAL.

niboshi:
Jenkins, test this please

PR #7185, opened 2019-05-19
Change `F.dropout` to use cuDNN by default

### LINKED ISSUE #6266: Why `F.dropout` usually does not use CuDNN?
Currently, `F.dropout` uses cudnn only when `using_config('use_cudnn', 'always')`
https://github.com/chainer/chainer/blob/d7adbfb45d46fafa6bd0bbb9f0cebe5894489437/chainer/functions/noise/dropout.py#L46

But IMO, cudnn should be enabled when `>=auto`.

crcrpar:
Close #6266

### DIFF
--- a/tests/chainer_tests/functions_tests/noise_tests/test_dropout.py
+++ b/tests/chainer_tests/functions_tests/noise_tests/test_dropout.py
@@ -8,7 +8,6 @@ from chainer import functions
 from chainer import gradient_check
 from chainer import testing
 from chainer.testing import attr
-from chainer.testing import backend
 
 
 @testing.parameterize(
@@ -17,7 +16,7 @@ from chainer.testing import backend
     {'dtype': numpy.float64, 'ratio': 0.5},
     {'dtype': numpy.float64, 'ratio': 0.0},
 )
-@backend.inject_backend_tests(
+@testing.inject_backend_tests(
     ['test_forward', 'test_backward', 'test_double_backward',
      'test_immutable'],
     # CPU tests
--- a/chainer/functions/noise/dropout.py
+++ b/chainer/functions/noise/dropout.py
@@ -83,7 +83,7 @@ class Dropout(function_node.FunctionNode):
         return y,
 
     def backward(self, x, gy):
-        if chainer.should_use_cudnn('==always', 5000) and self._use_cudnn:
+        if chainer.should_use_cudnn('>=auto', 5000) and self._use_cudnn:
             return DropoutGradCuDNN(self.states, self.dropout_ratio).apply(gy)
         else:
             return DropoutGrad(self.mask).apply(gy)
--- a/chainer/functions/noise/dropout.py
+++ b/chainer/functions/noise/dropout.py
@@ -44,7 +44,7 @@ class Dropout(function_node.FunctionNode):
         return y,
 
     def forward_gpu(self, x):
-        if (chainer.should_use_cudnn('==always', 5000)
+        if (chainer.should_use_cudnn('>=auto', 5000)
                 and x[0].flags.c_contiguous
                 and self.mask is None
                 and not self.return_mask):


tests/chainer_tests/functions_tests/noise_tests/test_dropout.py
```diff
@@ -183,7 +184,8 @@ def test_cpu_ideep(self):
     def test_gpu(self):
         self.x = cuda.to_gpu(self.x)
         self.mask = cuda.to_gpu(self.mask)
-        self._check()
+        with chainer.using_config('use_cudnn', 'never'):
+            self._check()
```
niboshi: How about adding a new test for `use_cudnn='always'` (as `test_gpu_cudnn`)?
Perhaps it's better to use `inject_backend_tests`.

crcrpar: I added tests, but it is possibly implemented in a wrong way.

tests/chainer_tests/functions_tests/noise_tests/test_dropout.py
```diff
@@ -144,46 +148,49 @@ def test_immutable(self, backend_config):
...
-        mask = self.mask if self.specify_mask else None
-        with chainer.using_config('train', self.train):
-            out, out_mask = functions.dropout(
-                self.x, 0.5, mask=mask, return_mask=True)
+    def test_check(self, backend_config):
+        mask = None
+        if self.specify_mask:
+            mask = self.mask
 
+        x, mask = backend_config.get_array((self.x.copy(), mask))
```
niboshi: No need to `copy()` I guess.

tests/chainer_tests/functions_tests/noise_tests/test_dropout.py
niboshi: Thanks!

tests/chainer_tests/functions_tests/noise_tests/test_dropout.py
```diff
@@ -144,46 +148,49 @@ def test_immutable(self, backend_config):
...
+        with chainer.using_config('train', self.train), backend_config:
+            out, out_mask = functions.dropout(
+                x, 0.5, mask=mask, return_mask=True)
         if self.train:
-            assert isinstance(out_mask, type(out.array))
-            if mask is None:
+            if mask is None and backend_config.use_cudnn != 'always':
+                assert isinstance(out_mask, type(out.array))
```
niboshi: Shouldn't this assertion always satisfied?

niboshi: Sorry, I mean it should be always checked.

crcrpar: How can we get mask from cudnn implementation?

crcrpar:
```bash
Platform: Linux-4.9.0-8-amd64-x86_64-with-debian-9.8
Chainer: 7.0.0a1
NumPy: 1.16.2
CuPy:
  CuPy Version          : 7.0.0a1
  CUDA Build Version    : 10000
  CUDA Driver Version   : 10000
  CUDA Runtime Version  : 10000
  cuDNN Build Version   : 7401
  cuDNN Version         : 7401
  NCCL Build Version    : None
  NCCL Runtime Version  : None
iDeep: Not Available

In [2]: x = cp.random.uniform(-1, 1, (10, 32)).astype(cp.float32)                                                                                                  
In [3]: with chainer.using_config('use_cudnn', 'always'): y, out_mask = F.dropout(x, return_mask=True)                                                             
In [4]: out_mask is None                                                                                                                                           
Out[4]: True
```

crcrpar: How will I handle cudnn enabled forward?

niboshi: Hmm, how about enabling cuDNN only when `not return_mask`?
You will need to pass `return_mask` to the `Dropout` class.

niboshi: I think this is a preexisting bug.
Could you separate a PR, as we need to backport the fix?

niboshi: Or it's ok to just dodge the bug in this PR and fix it at some other time.

crcrpar: OK, I'll send a PR to make `Dropout` FunctionNode consider `return_mask`, then I'll add tests that check cuDNN correctly used/unused.

Is my understanding correct?

niboshi: Yes. Thank you very much!

niboshi:
Jenkins, test this please

niboshi:
Jenkins, test this please

niboshi:
LGTM

PR #7194, opened 2019-05-20
Use CHAINER_READTHEDOCS instead of READTHEDOCS

kmaehashi:
I set `CHAINER_READTHEDOCS=1` in RTD configuration (only Chainer maintainers can access this page): https://readthedocs.org/dashboard/chainer/environmentvariables/
Note: it seems environment variables cannot be controlled via `readthedocs.yml`.

### DIFF
--- a/chainerx_build_helper.py
+++ b/chainerx_build_helper.py
@@ -45,8 +45,8 @@ class CMakeBuild(build_ext.build_ext):
         elif self.debug:
             # Being built with `python setup.py build --debug`
             build_type = 'Debug'
-        elif os.getenv('CHAINER_READTHEDOCS', None) == 'True':
-            # on ReadTheDocs
+        elif os.getenv('CHAINER_READTHEDOCS', None) == '1':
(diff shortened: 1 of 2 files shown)


chainerx_build_helper.py
```diff
@@ -45,7 +45,7 @@ def build_extension(self, ext):
         elif self.debug:
             # Being built with `python setup.py build --debug`
             build_type = 'Debug'
-        elif os.getenv('READTHEDOCS', None) == 'True':
+        elif os.getenv('CHAINER_READTHEDOCS', None) == 'True':
```
beam2d: How about using `'0'` / `'1'` for the value instead of `'True'`? (just for consistency with other environment variables)

beam2d: Or is it better to keep the consistency with the `READTHEDOCS` variable?

kmaehashi: Fixed and configured `CHAINER_READTHEDOCS=1` on RTD.

kmaehashi:
Closing as #7363 is better than this PR.

PR #7198, opened 2019-05-20
Call `require_grad()` on ChainerX `Variable.grad` setter 

### LINKED ISSUE #7190: Make `Variable.grad.setter` call `arr.require_grad()` for ChainerX array in order to align the double-backprop behavior with NumPy/CuPy
Original issue: The following code fails at the 2 commented lines. It succeeds for `numpy` and `cupy`.

```py
import chainerx as xp
import chainer


x_arr = xp.array([1, 2], 'float32')
gy_arr = xp.array([3, 4], 'float32')
ggx_arr = xp.array([5, 6], 'float32')

x = chainer.Variable(x_arr, requires_grad=True)
y = F.identity(x)
y.grad = gy_arr

gy = y.grad_var
y.backward(enable_double_backprop=True)


assert x.grad_var.requires_grad is True  # fail
x.grad_var.grad = ggx_arr  # error
x.grad_var.backward()


print(gy.grad_var)
```

niboshi:
Fixes #7190

~Depends on #7199~

### DIFF
--- a/chainer/variable.py
+++ b/chainer/variable.py
@@ -912,8 +912,15 @@ class Variable(object):
 
         self.array = d
 
-    def _set_chainerx_grad(self, g):
-        # Assigns chainerx.ndarray.grad
+    def _set_chainerx_grad(self, g, from_grad_var):
+        # Assigns chainerx.ndarray.grad.
+        #
+        # If the main array is connected to the graph, in order to enable
+        # double-backprop, the grad will also be backprop-required
+        # (a view is created not to affect the given grad).
+        # If the given grad is from a grad_var, this operation is skipped,
+        # as the status of the given grad reflects the necessity of
+        # double-backprop.
         assert self.xp is chainerx
         if not self._requires_grad and g is not None:
             raise RuntimeError(
@@ -927,13 +934,15 @@ class Variable(object):
         elif arr.is_backprop_required():
             # If g is grad-stopped, require grad on it.
             # Make a view in order not to affect the input.
-            if g is not None and not g.is_backprop_required():
+            if (g is not None
+                    and not from_grad_var
+                    and not g.is_backprop_required()):
                 g = g.view().require_grad()
             arr.set_grad(g)
 
     def _set_grad_without_check(self, g):
         if self._has_chainerx_array:
-            self._set_chainerx_grad(g)
+            self._set_chainerx_grad(g, False)
             self._grad_var = None
             return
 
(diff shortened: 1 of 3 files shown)


hvy:
Hm, BN tests are failing. 🤔

chainer/variable.py
```diff
@@ -925,6 +925,9 @@ def _set_chainerx_grad(self, g):
                 raise RuntimeError(
                     'Cannot set a gradient to an empty variable')
         elif arr.is_backprop_required():
+            # If g is grad-stopped, require grad on it.
+            if g is not None and not g.is_backprop_required():
+                g.require_grad()
```
hvy: Discussed offline to create a view of `g` so that the user's gradient isn't modified. This is to keep similar semantics to the array passed as the data to `Variable.__init__`.

chainer/variable.py
```diff
@@ -925,6 +925,10 @@ def _set_chainerx_grad(self, g):
                 raise RuntimeError(
                     'Cannot set a gradient to an empty variable')
         elif arr.is_backprop_required():
+            # If g is grad-stopped, require grad on it.
+            # Make a view in order not to affect the input.
+            if g is not None and not g.is_backprop_required():
+                g = g.view().require_grad()
```
hvy: `_set_chainerx_grad` is called from the `grad_var` setter too. If the argument variable to a call to it holds a `chainerx.ndarray` which doesn't require any gradients, it will be modified to require gradients here. Maybe pass a third flag argument to this method to control it?

```python
import chainerx as chx
import chainer as ch
v = ch.Variable(chx.ones(3, dtype='f'), requires_grad=True)
g = ch.Variable(chx.ones(3, dtype='f'), requires_grad=False)
v.grad_var = g
v.grad_var.requires_grad  # `True`, but should be `False`. It'll be `False` if `chx` is replaced with e.g. `numpy`.

```

tests/chainer_tests/test_variable.py
```diff
@@ -757,30 +757,57 @@ def test_copydata_from_to_uninitialized_parameters(
...
+            backend_config.xp.testing.assert_array_equal(v.grad, g)
 
     def test_grad_var(self, backend_config):
         x = backend_config.get_array(
             np.random.uniform(-1, 1, self.shape).astype(np.float32))
         g = backend_config.get_array(
             np.random.uniform(0.1, 10, self.shape).astype(np.float32))
-        v = chainer.Variable(x)
+        v = chainer.Variable(x, requires_grad=self.requires_grad)
         gv = chainer.Variable(g)
```
hvy: Can you parameterize `requires_grad` for `gv` and check its state too?

chainer/variable.py
niboshi: Fixed.

tests/chainer_tests/test_variable.py
niboshi: Fixed.

hvy:
Jenkins, test this please.

hvy:
LGTM!

hvy:
LGTM, failure seems unrelated.

PR #7209, opened 2019-05-20
Remove `PlotReport.available()` check in glance example

niboshi:
This example already requires `matplotlib`, so this check has no effect.
More importantly, it is nonessential for learning Chainer at a glance.


Crissman:
Jenkins, test this please.

Crissman:
Jenkins issue looks unrelated.

PR #7211, opened 2019-05-20
[RFC] Improve coding style in examples

niboshi:
* Avoid `from chainer import training` which I think is not normally done by ordinary users.
  * Additionally I don't think this is good in examples, because it makes less apparent about where `training.updaters.StandardUpdater` is acually located.
* When breaking lines, put a single argument in each line. This is easier to read and recommended in [OpenStack guideline](https://docs.openstack.org/hacking/latest/user/hacking.html#calling-methods).
* Avoid `a if b else c`, which is difficult to read.
* Some other minor fixes.

### DIFF
--- a/examples/mnist/train_mnist.py
+++ b/examples/mnist/train_mnist.py
@@ -4,7 +4,6 @@ import argparse
 import chainer
 import chainer.functions as F
 import chainer.links as L
-from chainer import training
 from chainer.training import extensions
 import chainerx
 
@@ -54,6 +53,7 @@ def main():
     args = parser.parse_args()
 
     device = chainer.get_device(args.device)
+    device.use()
 
     print('Device: {}'.format(device))
     print('# unit: {}'.format(args.unit))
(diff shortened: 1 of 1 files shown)


niboshi:
Requesting for comments.
If these fixes are acceptable, I would like to apply to all other examples.

crcrpar:
> * Avoid `from chainer import training` which I think is not normally done by ordinary users.

I don't think so personally, because one famous chainer implementation does this.
https://github.com/pfnet-research/sngan_projection/blob/e84b1a5f604de5fec268f37c3f26478e80b7f475/train.py#L6-L9

niboshi:
@crcrpar 
It seems the code was based on one of Chainer examples.
I guess not many people would have written this way by their own.
That may be just a guess, but there is more concrete reason to avoid that (restating what I wrote above):
* For those who are not familiar with Chainer, it's more difficult to figure out actual location of identifiers like `training.updaters.StandardUpdater`. They have to go back to the top  of the script to check.

It would be fine if we were going to say it's a convention (like `F` or `extensions`), but I don't think this is the case.

PR #7212, opened 2019-05-20
Improve code in `testing.backend.BackendConfig`

niboshi:


chainer/testing/backend.py
```diff
@@ -64,20 +61,16 @@ def _adjust_params(self):
...
     def device(self):
         if self._device is None:
             if self.use_cuda:
-                device = chainer.get_device(
-                    chainer.backends.cuda.Device(self.cuda_device))
+                device = backend.GpuDevice.from_device_id(self.cuda_device)
             elif self.use_chainerx:
-                device = chainer.get_device(self.chainerx_device)
+                device = backend.ChainerxDevice(
+                    chainerx.get_device(self.chainerx_device))
```
hvy: I prefer the previous simpler version 🤔

niboshi: Yes, and that makes the code safer.
For example, if `self.chainerx_device` was `1` for some reason, the previous code would not cause any error here.

hvy: Okay, thank you for the explanation. I think that the user input validation should take place before reaching this line if any but that's maybe another story.

hvy:
Jenkins, test this please.

hvy:
Failures seems unrelated. LGTM!

PR #7213, opened 2019-05-20
Print only non-default parameters in `testing.backend.BackendConfig`

niboshi:
To more easily figure out the configuration.


niboshi:
It was pointed out that if a default parameter that is significant for a test is explicitly given, this PR won't print that parameter. We may distinguish explicit parameters, but I don't have a strong motivation to implement it for now. So I'm closing the issue.

PR #7215, opened 2019-05-20
Fix sentiment example test

### LINKED ISSUE #7100: `test_thin_stack.py` is failing
The following test file in the sentiment analysis example fails. https://github.com/chainer/chainer/blob/master/examples/sentiment/test_thin_stack.py

It is a part of an example and is thus not tested by the CI. I have not yet investigated the cause nor the effects but the loss on the training data seems to decrease.

niboshi:
Fixes #7100

### DIFF
--- a/examples/sentiment/test_thin_stack.py
+++ b/examples/sentiment/test_thin_stack.py
@@ -64,7 +64,7 @@ class TestThinStackGet(unittest.TestCase):
 
         self.assertIsNone(i.grad)
         # Thin stack reueses the same gradient array.
-        self.assertIs(s.grad, t.grad)
+        self.assertIs(s.grad, gt_data)
 
     def test_backward_cpu(self):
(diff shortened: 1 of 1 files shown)


hvy:
Thanks for the fix. Was wondering if you could you motivate the introduction of the new test directory?

niboshi:
Simply it's inconvenience to be in separate directories if we're going to test it in CI.
That can be a separate PR if you prefer.

hvy:
Thanks for the clarification. It sound to me like the scope is a bit widened although it makes perfect sense to run this test in the CI... How about discussing how to run them in the CI in a different issue, making this PR minimal?

niboshi:
OK. Removed the move of the file.

hvy:
Jenkins, test this please.

hvy:
Jenkins, test this please.

hvy:
Failure is unrelated. LGTM!

PR #7216, opened 2019-05-20
Clear outputs in `FunctionNode.unchain`

niboshi:
From the documentation, I expect `outputs` to be cleared as well (even though they're weak refs).

Additionally I put an error message for ChainerX fallback function nodes.


beam2d:
Jenkins, test this please.

beam2d:
The CI error is not related. LGTM.

PR #7225, opened 2019-05-21
Support device specifiers in MNIST model parallel example

asi1024:
Some more internal fixes are needed to support ChainerX.

### DIFF
--- a/examples/mnist/train_mnist_model_parallel.py
+++ b/examples/mnist/train_mnist_model_parallel.py
@@ -1,11 +1,13 @@
 #!/usr/bin/env python
 import argparse
+import sys
 
 import chainer
 import chainer.functions as F
 import chainer.links as L
 from chainer import training
 from chainer.training import extensions
+import chainerx
 
 import train_mnist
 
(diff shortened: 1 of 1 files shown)


examples/mnist/train_mnist_model_parallel.py
```diff
@@ -13,47 +13,47 @@
...
 
             # the input size, n_units, is inferred
             self.second0 = train_mnist.MLP(n_units // 2, n_out)
             self.second1 = train_mnist.MLP(n_units // 2, n_out)
 
-            if gpu0 >= 0:
-                self.second0.to_gpu(gpu0)
-            if gpu1 >= 0:
-                self.second1.to_gpu(gpu1)
+            if device0.device.id >= 0:
```
niboshi: I think this condition can be omitted.

examples/mnist/train_mnist_model_parallel.py
```diff
@@ -13,47 +13,47 @@
...
+                self.second0.to_device(device0)
+            if device1.device.id >= 0:
```
niboshi: Ditto.

niboshi:
I guess this example does not work with ChainerX.
Please output an error message and exit, just like #6857.

niboshi:
Jenkins, test this please

niboshi:
LGTM

PR #7234, opened 2019-05-22
Check device instance using `xp` when possible

hvy:
Should be slightly faster, easier to read. 

Note that `CpuDevice` and `Intel64Device` cannot be distinguished so those occurrences are left unchanged..

### DIFF
--- a/chainer/backends/_chainerx.py
+++ b/chainer/backends/_chainerx.py
@@ -37,9 +37,10 @@ class ChainerxDevice(_backend.Device):
     def from_fallback_device(device):
         # TODO(niboshi): Write unit test
         assert isinstance(device, _backend.Device)
-        if isinstance(device, _cpu.CpuDevice):
+        assert not isinstance(device, intel64.Intel64Device)
+        if device.xp is numpy:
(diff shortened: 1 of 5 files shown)


chainer/backends/_chainerx.py
```diff
@@ -37,9 +37,10 @@ def from_array(array):
     def from_fallback_device(device):
         # TODO(niboshi): Write unit test
         assert isinstance(device, _backend.Device)
-        if isinstance(device, _cpu.CpuDevice):
+        assert not isinstance(device, intel64.Intel64Device)
+        if device.xp is numpy:
             return ChainerxDevice(chainerx.get_device('native', 0))
-        if isinstance(device, cuda.GpuDevice):
+        if device.xp is cuda.cupy:
```
niboshi: The previous logic is preferable.
* This method is irrelevant to the existence of `Intel64Device`. It should not appear in the code.
* In this implementation it causes assertion error instead of `RuntimeError` for an `Intel64Device` input.

hvy: Reverted change.

niboshi:
Jenkins, test this please

PR #7257, opened 2019-05-23
Add a test to reproduce the bcast deadlock problem

### LINKED ISSUE #7093: pytest does not detect Open MPI's hang issue on MNIST
The test does not detect the GPU-direct hang issue of Open MPI

As described in #5740 and #7086 , some Open MPI versions have bugs that cause ChainerMN MNIST example to hang. However, the test suite of ChainerMN ( `communicator_test` in particular) does not detect the problem. 

Actually, in my experiments
* MNIST hangs
* `communicator_test` passes
* `gpudirect_bug.c` fails with an error in MPI_Init()

(`gpudirect_bug.c` is from https://github.com/open-mpi/ompi/issues/3972)

Thus, we need further investigation.

The goal of this issue is
* Find the root cause of MNIST hang.
  * We assume the root cause is Open MPI's issue 3972, but not sure
  * The problem may include environment-dependency.
* Detect the error in the test suite (i.e. communicator_test)

keisukefukuda:
This PR fixes #7093 by adding a test to reproduce deadlock in mpi4py's Bcast function.

### DIFF
--- a/tests/chainermn_tests/communicator_tests/test_mpi.py
+++ b/tests/chainermn_tests/communicator_tests/test_mpi.py
@@ -40,13 +40,14 @@ class _TimeoutThread(threading.Thread):
                 m = re.search(r'ompi:version:full:(\S+)', out)
                 version = m.group(1)
 
-                msg = "\n\n***** ERROR:" \
+                msg = "\n\n***** ERROR: " \
                       "It looks like you are using " \
-                      "Open MPI version {}." \
-                      "It is known that the following Open MPI versions " \
-                      "have a bug that cause MPI_Bcast() deadlock when " \
-                      "GPUDirect is used: " \
-                      "3.0.0, 3.0.1, 3.0.2, 3.1.0, 3.1.1, 3.1.2\n"
+                      "Open MPI version {}.\n" \
+                      "***** It is known that the following Open MPI " \
+                      "versions have a bug \n" \
+                      "***** that cause MPI_Bcast() deadlock " \
+                      "when GPUDirect is used: \n" \
+                      "***** 3.0.0, 3.0.1, 3.0.2, 3.1.0, 3.1.1, 3.1.2\n"
                 if self.rank == 1:
                     # Rank 1 prints the error message,
                     # Rank 0 may finish Bcast() immediately without deadlock,
--- a/tests/chainermn_tests/communicator_tests/test_mpi.py
+++ b/tests/chainermn_tests/communicator_tests/test_mpi.py
@@ -75,7 +75,7 @@ class TestBcastDeadlock(unittest.TestCase):
 
         self.queue = queue.Queue(maxsize=1)
 
-    def teardown(self):
+    def tearDown(self):
         pass
 
     @chainer.testing.attr.gpu
@@ -101,5 +101,3 @@ class TestBcastDeadlock(unittest.TestCase):
         mpi_comm.barrier()
         self.queue.put(True)
         assert True
-
-        self.teardown()
--- a/tests/chainermn_tests/communicator_tests/test_mpi.py
+++ b/tests/chainermn_tests/communicator_tests/test_mpi.py
@@ -1,10 +1,14 @@
 import os
-import queue
+try:
+    import queue
+except ImportError:
+    import Queue as queue
 import re
 import signal
 from subprocess import CalledProcessError
 from subprocess import PIPE
 from subprocess import Popen
+import sys
 import threading
 import unittest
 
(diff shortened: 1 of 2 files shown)


tests/chainermn_tests/communicator_tests/test_mpi.py
```diff
@@ -0,0 +1,104 @@
...
+                p = Popen(['ompi_info', '--all', '--parsable'], stdout=PIPE)
+                out, err = p.communicate()
+                if type(out) == bytes:
+                    out = out.decode('utf-8')
+                m = re.search(r'ompi:version:full:(\S+)', out)
+                version = m.group(1)
+
+                msg = "\n\n***** ERROR:" \
+                      "It looks like you are using " \
+                      "Open MPI version {}." \
```
belldandyxtq: Can you add a line break here?

tests/chainermn_tests/communicator_tests/test_mpi.py
```diff
@@ -0,0 +1,104 @@
...
+                      "It is known that the following Open MPI versions " \
+                      "have a bug that cause MPI_Bcast() deadlock when " \
+                      "GPUDirect is used: " \
```
belldandyxtq: And here as well

belldandyxtq:
I have tested the following MPI versions
```
2.1.6     succeeded  (no error)
3.0.0     succeeded (hang, got error message)
3.0.1     failed? (core dumped)
3.0.2     failed? (core dumped)
3.0.3     failed (core dumped, NOT SUPPOSED)
3.1.0     succeeded (hang, got error message)
3.1.3     succeeded  (no error)
3.1.4     succeeded  (no error)
```
Will add `3.0.4, 3.1.1, 3.1.2` results
Command:
```
mpienv use $VERSION
mpienv exec -np 2 -npernode 1 -hostfile host --mca orte_base_help_aggregate 0 pytest test_mpi.py -s
```

Environment:
```
platform linux -- Python 3.7.2, pytest-4.6.2, py-1.8.0, pluggy-0.12.0
```

tests/chainermn_tests/communicator_tests/test_mpi.py
belldandyxtq: Can you add a line break here?

tests/chainermn_tests/communicator_tests/test_mpi.py
belldandyxtq: line break here

tests/chainermn_tests/communicator_tests/test_mpi.py
```diff
@@ -0,0 +1,104 @@
...
+            chainer.cuda.get_device_from_id(self.device).use()
+        else:
+            self.device = -1
+
+        if self.communicator.size < 2:
+            pytest.skip('This test is for at least two processes')
+
+        self.queue = queue.Queue(maxsize=1)
+
+    def teardown(self):
```
belldandyxtq: How about using the `tearDown`?

tests/chainermn_tests/communicator_tests/test_mpi.py
```diff
@@ -0,0 +1,104 @@
...
+
+        # This Bcast() cause deadlock if the underlying MPI has the bug.
+        th = _TimeoutThread(self.queue, self.communicator.rank)
+        th.start()
+        mpi_comm.Bcast(ptr, root=0)
+        mpi_comm.barrier()
+        self.queue.put(True)
+        assert True
+
+        self.teardown()
```
belldandyxtq: So we do not need to call it explicitly

tests/chainermn_tests/communicator_tests/test_mpi.py
keisukefukuda: Fixed!

tests/chainermn_tests/communicator_tests/test_mpi.py
keisukefukuda: Fixed by ↑

tests/chainermn_tests/communicator_tests/test_mpi.py
keisukefukuda: Improved the error message so the message would be:
```
***** ERROR: It looks like you are using Open MPI version 3.0.0.
***** It is known that the following Open MPI versions have a bug
***** that cause MPI_Bcast() deadlock when GPUDirect is used:
*****  3.0.0, 3.0.1, 3.0.2, 3.1.0, 3.1.1, 3.1.2
```

keisukefukuda:
I will update the MPI version list according to our internal & detailed tests.

belldandyxtq:
pfnci, test this please

belldandyxtq:
LGTM, 
The errors are irrelative to this PR.

PR #7258, opened 2019-05-23
Alias StackVector<int64_t, kMaxNdim> to Dims

### LINKED ISSUE #7343: Add type alias to StackVector<int64_t, kMaxNdim>
Relates to https://github.com/chainer/chainer/projects/2#card-16909303

Some other issue relates
- Argument of pooling kernels isn't typed `const&`
- Type of arguments named `stride` should be `Strides`
- Capacity of pad  should be double size of the dimension since it specifies begin and end pad
- There should be better name than Dims (some doesn't relate to dimensions)

take-cheeze:
This is a changes for: https://github.com/chainer/chainer/projects/2#card-16909303
closes #7343

There is few questions I want to ask:
- Is there any reason why `Dims` isn't typed `const&` in pooling kernels?
- Shouldn't the type of arguments named `stride` be `Strides` ?


take-cheeze:
Other things that needs to be discussed:
- Shouldn't be used in pad since it needs double size of the dimension
- There should be better name than `Dims` (some doesn't relate to dimensions)

chainerx_cc/chainerx/dims.h
```diff
@@ -8,10 +8,12 @@
 
 namespace chainerx {
 
-// Formatter to print StackVector containing integral elements as e.g. '[]' or '[1, 2, 3]'.
+typedef StackVector<int64_t, kMaxNdim> Dims;
```
kshitij12345: Wouldn't `using` be better ?

hvy:
> Is there any reason why Dims isn't typed const& in pooling kernels?

It indeed seems better to use const&s. Could you send a separate PR, or file an issue?

> Shouldn't the type of arguments named stride be Strides ?

Do you mean the stride argument to the convolutions? These strides are different "strides" from what `Strides` is intended to represent. They are strides of the convolution kernel.

> Shouldn't be used in pad since it needs double size of the dimension

Hm, why is that? We currently don't allow specifying different padding values for e.g. left and right.

> There should be better name than Dims (some doesn't relate to dimensions)

Could you point me to where it doesn't relate to dimensions? The name "Dims" doesn't necessarily take the usage into account and is merely an alias for the underlying type (i.e. `StackVector<int64_t, kMaxNdim>`).

How about also renaming `CudaStackVector` to `CudaDims`?

take-cheeze:
> Hm, why is that? We currently don't allow specifying different padding values for e.g. left and right.

Sorry it seems I confused with ONNX. Never mind.

hvy:
Ah, it seems like ONNX allows fine grained control of these values. I did not know that.

hvy:
Jenkins, test this please.

hvy:
Jenkins, test this please.

hvy:
Hm, the Jenkins log was 852MB. Rerunning the build. 🤔

hvy:
LGTM! The Jenkins failure seem unrelated.

PR #7259, opened 2019-05-23
Add cmake to setup.py install requirements

Crissman:
Now that ChainerX is part of the base build, we should add `cmake` to the requirements.


Crissman:
After discussion, decided to add this to Chainer install docs. Closing, to replace with new PR.

PR #7271, opened 2019-05-23
DataParallel with custom train-loop

### LINKED ISSUE #7264: [Enhancement] DataParallel without trainer
Hi, 
I've implemented a version of DataParallel, which automatically handles the following things:

* replication of the model across different devices
* data scattering across thoose devices
* the gradient accumulation before optimizer update
* the update of all replicated parameters after the optimizer step

My implementation is based on the [DataParallel Example without a trainer](https://docs.chainer.org/en/stable/guides/gpu.html#data-parallel-computation-on-multiple-gpus-without-trainer) and the [PyTorch Version](https://pytorch.org/docs/stable/nn.html#torch.nn.DataParallel).

Would you welcome a PR including this or is this something you don't want?

justusschock:
This adds a `DataParallel` chain, optimizer hooks for gradient accumulation and parameter replication and an optimizer wrapper, which automatically registers this hooks.

I didn't really know, how to test this, so I came up with some functional tests.

The gathering and scattering somehow mimics the behavior of [pytorch](https://pytorch.org/docs/stable/nn.html#torch.nn.DataParallel). The whole code is inspired by the pytorch version and the [data-parallel example without a trainer](https://docs.chainer.org/en/stable/guides/gpu.html#data-parallel-computation-on-multiple-gpus-without-trainer).

I'm also not sure, whether the attribute forwarding inside the optimizer wrapper should be done by properties (which just works, but maybe missleads code completion in IDEs) or if we should explicitly write down the functions including all parameters (which would increase the effort to maintain this, if the API changes).

Closes #7264 

EDIT: I don't know, why `chainer.Optimizer` isn't found. This worked for me in my own implementation with chainer installed via pip... Will investigate this tomorrow

### DIFF
--- a/chainer/optimizer_hooks/data_parallel.py
+++ b/chainer/optimizer_hooks/data_parallel.py
@@ -23,7 +23,7 @@ class DataParallelOptimizerCumulateGradientsHook(object):
                 summed across the replications
 
         """
-        if isinstance(optimizer.target, chainer.links.model.DataParallel):
+        if isinstance(optimizer.target, chainer.links.DataParallel):
             for module in optimizer.target.modules[1:]:
                 optimizer.target.modules[0].addgrads(module)
 
@@ -48,6 +48,6 @@ class DataParallelOptimizerUpdateModelParameters(object):
                 updated across the replications
 
         """
-        if isinstance(optimizer.target, chainer.links.model.DataParallel):
+        if isinstance(optimizer.target, chainer.links.DataParallel):
             for module in optimizer.target.modules[1:]:
                 module.copyparams(optimizer.target.modules[0])
--- a/chainer/optimizer_hooks/data_parallel.py
+++ b/chainer/optimizer_hooks/data_parallel.py
@@ -12,7 +12,7 @@ class DataParallelOptimizerCumulateGradientsHook(object):
     call_for_each_param = False
     timing = 'pre'
 
-    def __call__(self, optimizer: chainer.Optimizer):
+    def __call__(self, optimizer):
         """
         Summing up all parameters if the target is an instance of
         ~chainer.links.model.DataParallel
@@ -38,7 +38,7 @@ class DataParallelOptimizerUpdateModelParameters(object):
     call_for_each_param = False
     timing = "post"
 
-    def __call__(self, optimizer: chainer.Optimizer):
+    def __call__(self, optimizer):
         """
         Copying all parameters across the model replicas after optimizer update
 
--- a/chainer/optimizer_hooks/data_parallel.py
+++ b/chainer/optimizer_hooks/data_parallel.py
@@ -1,0 +1,53 @@
+import chainer
+import chainer.links.model
+
+
+class DataParallelOptimizerCumulateGradientsHook(object):
+    """
+    A hook which sums up all replication's gradients in a
+    DataParallel-Scenario
+    """
+
+    name = "DataParallelCumulateGradients"
+    call_for_each_param = False
+    timing = 'pre'
+
+    def __call__(self, optimizer: chainer.Optimizer):
+        """
+        Summing up all parameters if the target is an instance of
+        ~chainer.links.model.DataParallel
+
+        Args
+            optimizer (~chainer.Optimizer):
+                the optimizer holding the target, whose gradients should be
+                summed across the replications
+
+        """
+        if isinstance(optimizer.target, chainer.links.model.DataParallel):
+            for module in optimizer.target.modules[1:]:
+                optimizer.target.modules[0].addgrads(module)
+
+
+class DataParallelOptimizerUpdateModelParameters(object):
+    """
+    A hook to replicate all parameters from the root model, to all
+    model-replicas after the optimizer step
+    """
+
+    name = "DataParallelUpdateModelParams"
+    call_for_each_param = False
+    timing = "post"
+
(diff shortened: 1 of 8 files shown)


justusschock:
Okay, to be honest, I have absolutely no idea, why the imports don't work, because they should (and they are if I install the latest PyPi Release).

chainer/optimizers/data_parallel.py
```diff
@@ -0,0 +1,148 @@
+import chainer
+import chainer.optimizer_hooks
+
```
kshitij12345: Should probably import `optimizer` similar to how it is done here


https://github.com/justusschock/chainer/blob/f8c9f1507de5828422131a69e1e866f330f3fd86/chainer/optimizers/rmsprop.py#L5

This removed import error atleast from this file.

chainer/optimizer_hooks/data_parallel.py
```diff
@@ -0,0 +1,53 @@
...
+    """
+    A hook which sums up all replication's gradients in a
+    DataParallel-Scenario
+    """
+
+    name = "DataParallelCumulateGradients"
+    call_for_each_param = False
+    timing = 'pre'
+
+    def __call__(self, optimizer: chainer.Optimizer):
```
kshitij12345: I guess you should remove the type annotation as `chainer` supports Python 2.7

https://docs.chainer.org/en/stable/install.html

Plus it will also fix the error.

justusschock:
Finally, this should be ready for review (last commit just added docstrings, which were ommited due to their simplicity, but tests passed before).

Thanks @kshitij12345 for the hint regarding the import statements. This solved it.

cc @beam2d

kuenishi:
Let me as several questions; what is the rationale behind adding another multi-process training, or use case for yet another multi-GPU feature in Chainer?  We already have [MultiprocessParallelUpdater](https://docs.chainer.org/en/stable/reference/generated/chainer.training.updaters.MultiprocessParallelUpdater.html) and ChainerMN.

Especially, the use case mentioned in #7264 can easily be done with what we have in `chainermn`, included in Chainer ( [MNIST example](https://github.com/chainer/chainer/blob/master/examples/chainermn/mnist/train_mnist.py) , [ImageNet example](https://github.com/chainer/chainercv/blob/master/examples/classification/train_imagenet_multi.py)  ) by just installing `mpi4py` and underlying MPI package. MPI is easy to install in Linux distributions such as Ubuntu and so on. ChainerMN is full-featured distribution feature that supports scattering training data, aggregating gradients and updating the parameters after optimization (and multi-node batch normalization, model parallelism, etc).

Sorry that we don't have exact example or documentation for custom loop, but it must be quite easy to port single-GPU training code to parallel training IMO. I'd like you to take a look at [our tutorial](https://docs.chainer.org/en/stable/chainermn/tutorial/index.html) and would be very happy to answer questions.

justusschock:
From what I understood (and correct me if I'm wrong here), chainermn was designed to enable distributed training and thus always has the overhead of inter-process (and maybe even inter-host) communication. 

IMO, it is not necessary to work with this overhead and this may complicate the whole behavior. Also ChainerMN only scatters whole datasets (atleast that's the only thing I found so far), while the approach here scatters on a per-batch basis.

Another benefit would be, that this module would behave like a typical single GPU model.

I want to be honest here: I have never tried the ChainerMN approach for this, because I only saw the examples you linked (working with predefined trainers and evaluators) and found it kind of non-trivial (or maybe just unintuitive) to convert it a custom train-loop, which is why I came up with this (python-only) solution.

It's up to you, whether this fits chainers general scope or not :)

beam2d:
In most cases from my experience, performance in multi-GPU training is much better with the multi-process approach (ChainerMN and MultiprocessParallelUpdater) than the single-process approach (ParallelUpdater), so I generally recommend the former. There are (at least) two reasons for that:

1) The multi-process approach parallelizes data loading and preprocessing while the single-process approach does not. This point influences performance in many cases.
2) The multi-process approach launches kernels for multiple GPUs in parallel while the single-process approach does it sequentially. In particular, the kernels for the second GPU are not launched until the forward code for the first GPU is finished. If there are more than two GPUs, the latter GPU has a longer idle time. This may degrade the device utilization. Moreover, when the number of kernels launched for single GPU is large, it may hit the maximum length of the CUDA kernel launch queue (I remember it's around 1,000); in this case, the host side code blocks until the first kernel finishes, and the performance is much more degraded.

Based on these considerations, honestly speaking, I'm currently not positive to introduce yet another way to write single-process multiple-GPU code. The easiness of coding might count, but as @kuenishi noted, ChainerMN changed the situation. It only requires two or three points of user code to be modified for supporting multi-GPU (and even multi-host) execution. I believe it's as easy as the changes needed for using DataParallel in this PR (for some users, setting up MPI environment might be a bottleneck for using it, though).

justusschock:
Right, so I'll close this PR and either try to switch to the multiprocess approach or host my version in one of my repos :)

Thanks for your feedback!

kuenishi:
Just FYI, carefully comparing these two examples would be a hint for porting your custom loop to parallel training:
- https://github.com/chainer/chainer/blob/master/examples/chainermn/mnist/train_mnist.py
- https://github.com/chainer/chainer/blob/master/examples/mnist/train_mnist.py

Feel free to ask any question in Chainer Slack: https://bit.ly/join-chainer-slack or forum: https://groups.google.com/forum/#!forum/chainer

PR #7275, opened 2019-05-24
Use `_` for private classes under `chainer.dataset.tabular`

Hakuyume:
e.g. `chainer.dataset.tabular.slice.Slice` -> `chainer.dataset.tabular._slice._Slice`

### DIFF
--- a/chainer/dataset/tabular/__init__.py
+++ b/chainer/dataset/tabular/__init__.py
@@ -1,2 +1,2 @@
-from chainer.dataset.tabular import as_mode  # NOQA
-from chainer.dataset.tabular import slice  # NOQA
+from chainer.dataset.tabular import _as_mode  # NOQA
+from chainer.dataset.tabular import _slice  # NOQA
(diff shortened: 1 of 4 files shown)


Hakuyume:
jenkins, test this please.

okapies:
@beam2d Someone pointed out in today's meeting that it doesn't need to prefix the classes instead of the functions. Any idea?

Hakuyume:
I personally prefer adding prefix because I want users not to use these classes directly.

Hakuyume:
The failure of jenkins seems not related to this PR.

beam2d:
Basically, private names only need at least one name in the module path prefixed with `_`, so `chainer.dataset.tabular._slice.Slice` should also be treated as a private name (using it requires to import a module prefixed with `_`, which is not "public"). Whether to also prefix `Slice` with `_` or not is a matter of taste (I don't have any preference here).

Hakuyume:
> Whether to also prefix Slice with _ or not is a matter of taste (I don't have any preference here).

So, is this PR ok?

beam2d:
Yes, it looks good to me.

PR #7276, opened 2019-05-24
Add `chainer.dataset.tabular.DelegateDataset`

Hakuyume:
merge after ~#7150~, ~#7361~, #7847 

This is a helper class to build `TabularDataset` (like `DatasetMixin` or `GetterDataset` in ChainerCV).
I don't think the name is good. Suggestions are welcome.


Hakuyume:
jenkins, test this please

yuyu2172:
I suggest naming the proposed dataset as `CustomDataset` instead of `SimpleDataset`.

beam2d:
I think the feature and implementation of this class somehow overlap with those of `from_data` and am considering if there is a better separation of them. One idea would be

- to support callable columns in `from_data`, and
- to make `SimpleDataset` concentrate on providing an inheritable implementation of an existing base dataset (i.e., `self._dataset`)

In this case, we may name the class `DelegateDataset` (just an idea). It has almost the same feature as the current `SimpleDataset` except for how to specify the base dataset. For example, instead of adding columns with `add_column`, it may just accept the base dataset (or optionally the arguments to `from_data`) in `__init__`. Does this design make sense?

Hakuyume:
> to support callable columns in from_data, and

To support callable, the length should be given. In `SimpleDataset`, the author of a dataset can pass this information by implementing `__len__`. I think this is difficult in `from_data`.

Hakuyume:
One solution is adding `from_callable(keys, func, length)`. With this helper, we can write as following
```python
super().__init__(from_data(('a', array)).join(from_callable('b', func_b, 100).join(from_callable('c', func_c, 100)))
```

Hakuyume:
Another option:
Change the signature of `from_data` from `from_data(*args, **kwargs)` to `from_data(data, len=None)`. 
- `from_data((array, callable))`
- `from_data({'a': array, 'b': callable})`
- `from_data((callable,), len=100)`
- `from_data({'a': callable}, len=100)`

beam2d:
Thank you for the analysis and proposals. I think it's natural to combine a callable with the length as a column. How about letting `from_data` accept a pair of callable and length as a single column?

beam2d:
It could also accept a callable that implements `__len__` instead of a pair of callable and length, but not sure if it's useful.

Hakuyume:
> Thank you for the analysis and proposals. I think it's natural to combine a callable with the length as a column. How about letting from_data accept a pair of callable and length as a single column?

If a user want to pass more than one callable, will it be `from_data((callable_a, 100), (callable_b, 100))`?

Hakuyume:
How about following API?
- `from_data(array)`  -> unary, name of key is randomly set
- `from_data(callable, len)`  -> unary, name of key is randomly set
- `from_data((key, array))` -> unary
- `from_data((key, callable), len)` -> unary
- `from_data((array_a, array_b))` -> tuple, names of keys are randomly set
- `from_data((array, callable))` -> tuple, names of keys are randomly set
- `from_data((callable_a, callable_b), len)` -> tuple, names of keys are randomly set
- `from_data(((key_a, array_a), (key_b, array_b)))` -> tuple
- `from_data(((key_a, array), (key_b, callable)))` -> tuple
- `from_data(((key_a, callable_a), (key_b, callable_b)), len)` -> tuple
- `from_data({key_a: array_a, key_b: array_b})` -> dict, order of keys are randomly set
- `from_data({key_a: array, key_b: callable})` -> dict, order of keys are randomly set
- `from_data({key_a: callable_a, key_b: callable_b}, len)` -> dict, order of keys are randomly set

beam2d:
> If a user want to pass more than one callable, will it be from_data((callable_a, 100), (callable_b, 100))?

I thought it'd be ok to omit the length if that is provided by another column.

Anyway, your last proposal looks good to me.

Hakuyume:
> I thought it'd be ok to omit the length if that is provided by another column.

You mean accepting `from_data((callable_a, 100), callable_b)`? It looks strange to me.

beam2d:
Yeah, that may look strange, but it allows to write `from_data(column_with_len, callable)`, which looks natural to me. What I thought is that only at least one column should know its length, and we can treat `(callable, length)` as a column that knows its length, while a sole `callable` as a column that does not know its length.

Hakuyume:
I'm using a tuple of `(key, callable)`  as a column with its name. Accepting `(callable, len)` will be confusing.

Hakuyume:
Giving `key` as an element of tuple is reasonable because it can be specified for each column. On the other hand, `len` should be specified for entire dataset.

beam2d:
Ah, that makes sense. Another way along that direction is to provide a special type that combines a callable and a length (`to_column(callable, len)`), but it looks overkill to me, so providing `len` argument in `from_data` looks better to me, then.

Hakuyume:
@beam2d I updated `from_data` to support callable. Please check #7847.

Hakuyume:
@beam2d I also updated this PR to use `from_data`.

beam2d:
LGTM.
Jenkins, test this please.

Issue #6094, opened 2019-01-26
Check output in example tests


niboshi:
In example CI, currently it's only checked that the examples can be run without errors. No output is checked.
https://github.com/chainer/chainer-test/blob/fc5520350944b93d769986e30e697d6e0092a94d/test_example.sh

We could improve this test to check output using regular expressions, for example.
(We could have been able to detect an issue like https://github.com/chainer/chainer/pull/6091.)

niboshi:
TODO: First write a test for MNIST.

PR #7277, opened 2019-05-25
Improve example tests

niboshi:
This is an attempt to improve example tests in two points:
- Some examples have `--test` argument, which is irrelevent to ordinary readers and often confusing. This PR eliminates the necessity of it by altering script files on the fly.
- Moves the test logic from https://github.com/chainer/chainer-test to the chainer repository, to make it easier to maintain.

This is also preparation of #6094 (see #7280)


niboshi:
#7475 has been merged.

PR #7280, opened 2019-05-25
Check output in example tests 

### LINKED ISSUE #6094: Check output in example tests

niboshi:
Fixes #6094

Depends on #7277


niboshi:
I found pytest-repeat useful to investigate the acceptable fluctuations of the statistics, but it does not support `unittest` style tests. So I wrote the test in pytest-style.

examples/tests/test_mnist.py
```diff
@@ -0,0 +1,38 @@
...
+
+epoch       main/loss   validation/main/loss  main/accuracy  validation/main/accuracy  elapsed_time
+1           {a        } {b                  } {c           } {d                      } {e         }
+''',  # NOQA
+        a=(float, lambda x: 0.6 < x < 1.5),
+        b=(float, lambda x: 0.3 < x < 0.6),
+        c=(float, lambda x: 0.62 < x < 0.82),
+        d=(float, lambda x: 0.83 < x < 0.98),
+        e=(float, lambda x: 0. < x < 100.),
+    )
```
beam2d: I would like to find information for choosing this new format method.
1) Possible other solutions
2) Pros and cons of these solutions (and thus why the above code is good)

niboshi: Actually I didn't come up with any other reasonable solutions.
I'll try to list possibilities any way.

- Write in each test a logic to check the output line by line

  - Pros: What's being done in each step is clear from test implementation.
  - Cons: Prohibitively difficult to write/maintain test cases.

- Use a regular expression to capture the numbers, and use lambdas to check the numerical ranges.

  - Pros: Being a standard notation
  - Cons: Difficult to express the output with a regular expression. It's also not easy to interpret for human. Possibilities to not be able to express with the syntax.

niboshi:
Rebased as #7475 has been merged.

niboshi:
PTAL

niboshi:
Conflict resolved.

niboshi:
matplotlib cannot be removed from requirements. I add a step to install matplotlib in Travis CI.

scripts/ci/travis/run-tests.sh
```diff
@@ -174,6 +174,7 @@ case "${CHAINER_TRAVIS_TEST}" in
             install)
                 run_prestep install_chainer_test_deps
                 run_prestep chainer_install_from_sdist
+                run_prestep chainer_example_tests
```
emcastillo: shouldn't be `install_chainer_example_tests`?

niboshi: Oops... Force-pushed in f586e60396ce1e671d717a5d446460b571fef803.
Thanks!

emcastillo:
Jenkins, test this please

PR #7281, opened 2019-05-25
Fix `F.absolute_error` for ChainerX

crcrpar:
Referencing #6071.

### DIFF
--- a/tests/chainer_tests/functions_tests/loss_tests/test_absolute_error.py
+++ b/tests/chainer_tests/functions_tests/loss_tests/test_absolute_error.py
@@ -41,6 +41,12 @@ from chainer import utils
 class TestAbsoluteError(testing.FunctionTestCase):
 
     def setUp(self):
+        # Declare x0 and x1 here for `test_backward_non_default_gpu`
+        self.x0 = numpy.random.uniform(-1, 1, self.shape).astype(self.dtype)
+        # Add sufficient margin to prevent computational error
+        diff = numpy.random.uniform(-1, 1, self.shape).astype(self.dtype)
+        diff[abs(diff) < 0.02] = 0.5
+        self.x1 = numpy.asarray(self.x0 + diff)
         if self.dtype == numpy.float16:
             self.check_forward_options.update({'atol': 1e-3, 'rtol': 1e-3})
             self.check_backward_options.update({'atol': 5e-2, 'rtol': 5e-2})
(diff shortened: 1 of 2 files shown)


niboshi:
Thanks!
Jenkins, test this please

niboshi:
Oh, it seems `F.absolute_error` needs a fix for ChainerX.
Could you implement `AbsoluteError.forward_chainerx()`?

(Fallback impl `AbsoluteError.forwrad` assigns `self.diff` as `{numpy/cupy}.ndarray`, but that causes array type mismatch in `backward`).

niboshi:
Thanks! Jenkins, test this please

niboshi:
Oh, ChainerX does not support built-in `abs` yet.
How about using `chainerx.abs` instead?

crcrpar:
Soffy for bothering you. Somehow my environment is broken while chainer and chainerx of 7.0.0a1 are installed/imported w/o any error so that I could not find the absence of chainerx's built-in `abs` support.

niboshi:
@crcrpar No problem at all.
Jenkins, test this please.

niboshi:
Jenkins, test this please.

niboshi:
Please separate `test_backward_non_default_gpu` as another test case.
It does not use any feature of `FunctionTestCase`.
Input sampling is not meant to be done in `setUp`.

tests/chainer_tests/functions_tests/loss_tests/test_absolute_error.py
```diff
@@ -26,74 +26,89 @@
...
+class TestAbsoluteError(testing.FunctionTestCase):
 
     def setUp(self):
-        self.x0 = numpy.random.uniform(-1, 1, self.shape).astype(self.dtype)
+        if self.dtype == numpy.float16:
+            self.check_forward_options.update({'atol': 1e-3, 'rtol': 1e-3})
+            self.check_backward_options.update({'atol': 5e-2, 'rtol': 5e-2})
+            self.check_double_backward_options.update(
+                {'atol': 5e-2, 'rtol': 5e-2})
+        if self.double_backward_options:
```
niboshi: This check does not seem reliable.
It's not guaranteed that `self.double_backward_options` is empty by default.

tests/chainer_tests/functions_tests/loss_tests/test_absolute_error.py
```diff
@@ -26,74 +26,89 @@
...
+        # Add sufficient margin to prevent computational error
+        diff = numpy.random.uniform(-1, 1, self.shape).astype(self.dtype)
+        diff[abs(diff) < 0.02] = 0.5
+        self.x1 = numpy.asarray(self.x0 + diff)
+        self.gy = numpy.random.random(self.shape).astype(self.dtype)
 
-    # test for #4669
     @attr.multi_gpu(2)
-    def test_backward_non_default_gpu(self):
+    def test_backward_non_default_gpu(self, backend_config):
```
niboshi: Previously this test did not use `@testing.inject_backend_tests`.
I think there's no strong motivation to run this test for various backends.

niboshi:
Thanks!
Jenkins, test this please

niboshi:
I changed the title of the PR because it contains a bug fix for ChainerX.

PR #7285, opened 2019-05-27
Implement clip function in C++

take-cheeze:


chainerx/math/misc.py
```diff
@@ -1,33 +0,0 @@
-import chainerx
-
-
-# TODO(sonots): Implement in C++
-def clip(a, a_min, a_max):
-    """Clips the values of an array to a given interval.
```
asi1024: Could you move this docstring to `chainerx/_docs/routines.py`?

hvy:
Does this also support "None" as min and max? Encountered an error trying to do this using the current temporary Python implementation since it probably wasn't considered. NumPy does https://docs.scipy.org/doc/numpy/reference/generated/numpy.clip.html

hvy:
Just a general tip but could you prefix the PR title with "[WIP]" (or create a the PR as a draft PR when first creating it)? That'd help the reviewers to know when to actually start reviewing.

take-cheeze:
@hvy Now should support `None` arguments too.
Do I need to update the document too?

hvy:
Thanks for the quick fixes. 
1. Overloads might be fine, but you can also use `nonstd::optional` for values that may take "None". I Think that might be simpler in this case. You'll only need a single definition without any overloads.
2. Ah you changes the Chainer's behavior too. I'm sorry to bother but that's probably better done in a separate PR... Could you?

hvy:
Yes the documents should reflect the current logic.

hvy:
Oh, could you resolve the conflicts?

hvy:
Please let me know when it's ready for review again.

take-cheeze:
@hvy I should be fixed now.

chainerx_cc/chainerx/python/routines.cc
```diff
@@ -696,6 +696,56 @@ void InitChainerxMath(pybind11::module& m) {
           [](const ArrayBodyPtr& x1, const ArrayBodyPtr& x2) { return MoveArrayBody(Minimum(Array{x1}, Array{x2})); },
           py::arg("x1"),
           py::arg("x2"));
+    m.def("clip",
+          [](const ArrayBodyPtr& a, Scalar a_min, const nonstd::optional<Scalar>& a_max) {
+              if (!a_max) {
```
hvy: Is the negation equal to calling `!a_max.has_value()` ? We often use the more verbose `has_value`. Sorry these are all details but it helps keeping the code consistent.

chainerx_cc/chainerx/routines/misc.h
```diff
@@ -25,4 +25,9 @@ Array Minimum(const Array& x1, Scalar x2);
 Array Minimum(Scalar x1, const Array& x2);
 Array Minimum(const Array& x1, const Array& x2);
 
+Array Clip(const Array& x1, Scalar min, Scalar max);
+Array Clip(const Array& x1, const Array& a_min, Scalar max);
+Array Clip(const Array& x1, Scalar min, const Array& a_max);
+Array Clip(const Array& x1, const Array& min, const Array& max);
```
hvy: We're usually okay with propagating optional arguments to the routines (C++ interface). Do you have a strong reason not to do so in this case, basically (with acceptable overhead) moving the error checking from the Python glue code to C++. The benefit is that the glue code becomes simpler and the interfaces more similar.

chainerx_cc/chainerx/python/routines.cc
hvy: Can you reduce the "overloads" to 4?

chainerx_cc/chainerx/routines/misc.h
hvy: How about using the same variables names in NumPy?

hvy:
Wrote some comments if you could take a look.

chainerx_cc/chainerx/routines/misc.h
take-cheeze: `a_` prefix seems to mean it may take array object so I think it's shouldn't be used in language that supports function overloading.

take-cheeze: If the glue code requires more than 10 line I would use optional typed argument but it's quite short.
Polluting overloads of `Clip` function seems more complex to me. If I'm using ChainerX directly from C++ I don't want many unused overloads around the error messages.
In this case keeping complexity to glue code seems to make the total complexity less to me.

Though if the glue code to the python is generate fully automatic I'll define the optional interface.
And it's not too late to define numpy like interface after a demand of somebody.

take-cheeze:
Thanks for the review!
Fixed some of it and responded with some comment.

chainerx_cc/chainerx/routines/misc.h
hvy: Thanks for your explanation. I see your point and also think it's important to think of the C++ and Python interfaces separately. It does make sense to keep the interface non-optional for the C++ especially in terms of the code quality of the implementation but I'm at the same time not sure if the average (Python +)C++ user is happier having to write the current glue code logic themselves.

take-cheeze:
Seems travis failure is fixed after merging upstream

PR #7287, opened 2019-05-27
Make `chx.backward` not cause error even if backprop is not required

### LINKED ISSUE #7272: Make `chx.backward` not cause error even if backprop is not required
Context: https://github.com/chainer/chainer/issues/6612#issuecomment-495080834

niboshi:
Fixes #7272

### DIFF
--- a/chainerx_cc/chainerx/backward.cc
+++ b/chainerx_cc/chainerx/backward.cc
@@ -111,10 +111,12 @@ std::unordered_map<OpNode*, std::vector<uint8_t>> CreateSubgraph(
 
         // Initialize op node queue starting from outputs.
         for (const std::shared_ptr<ArrayNode>& array_node : output_array_nodes) {
-            forward_op_nodes.emplace(array_node.get(), nullptr);  // Outputs have no forward op nodes.
-            const std::shared_ptr<OpNode>& op_node = array_node->creator_op_node();
-            if (op_node != nullptr) {
-                PushNodeIfNotSeen(candidate_op_nodes, op_node.get(), seen_op_nodes);
+            if (array_node != nullptr) {
+                forward_op_nodes.emplace(array_node.get(), nullptr);  // Outputs have no forward op nodes.
+                const std::shared_ptr<OpNode>& op_node = array_node->creator_op_node();
+                if (op_node != nullptr) {
+                    PushNodeIfNotSeen(candidate_op_nodes, op_node.get(), seen_op_nodes);
+                }
             }
         }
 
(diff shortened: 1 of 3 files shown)


chainerx_cc/chainerx/backward.cc
```diff
@@ -211,27 +213,34 @@ class BackwardImpl {
...
+        // destruction order problem.
+        static const std::shared_ptr<ArrayNode> kNullArrayNode{nullptr};
+
         for (const Array& output : outputs) {
-            if (!output.IsBackpropRequired(backprop_id)) {
-                throw ChainerxError{"Cannot start backprop from an array whose gradient is not required (on graph '", backprop_id, "')"};
+            if (output.IsBackpropRequired(backprop_id)) {
+                output_array_nodes_.emplace_back(internal::GetArrayBody(output)->GetArrayNode(backprop_id));
+            } else {
+                output_array_nodes_.emplace_back(kNullArrayNode);
```
hvy: You can just write as follows (and avoid the static shared_ptr)?
```cpp
output_array_nodes_.reserve(outputs.size());
for (const Array& output : outputs) {
    output_array_nodes_.emplace_back(internal::GetArrayBody(output)->GetArrayNode(backprop_id));
}
```

chainerx_cc/chainerx/backward.cc
```diff
@@ -211,27 +213,34 @@ class BackwardImpl {
...
             }
-            output_array_nodes_.emplace_back(internal::GetArrayBody(output)->GetArrayNode(backprop_id));
         }
 
-        // Check if backward is possible for the given graph, in this context.
-        // It is not possible if a graph from an outer scope has already been backpropped.
-        backprop_id.context().CheckBackpropAllowed(backprop_id);
+        if (!output_array_nodes_.empty()) {
```
hvy: A minor but `output_array_nodes_` should have the same size here as `outputs`. If `outputs` is empty, this ctor body does essentially nothing. How about checking that in the beginning with an early return?

chainerx_cc/chainerx/backward_test.cc
```diff
@@ -334,15 +334,29 @@ TEST_F(BackpropTest, BackwardWithComplicatedRanks) {
...
     auto xs = MakeFullArrays({1}, {2.0f, 3.0f});
     auto y1 = xs[0] * xs[1];  // without graph
-    EXPECT_THROW(Backward(y1), ChainerxError);
+    Backward(y1);
+
+    EXPECT_THROW(xs[0].GetGrad(), ChainerxError);
+    EXPECT_THROW(xs[1].GetGrad(), ChainerxError);
+}
+
+TEST_F(BackpropTest, BackwardFromArrayWithAndWithoutNode) {
```
hvy: Are the comments in this test reversed ("with"/"without")?

niboshi: Oops, I thought the comments were wrong, but I was wrong.

chainerx_cc/chainerx/backward.cc
niboshi: Fixed as suggested.

chainerx_cc/chainerx/backward.cc
niboshi: Done.

niboshi:
PTAL.

hvy:
Jenkins, test this please.

hvy:
LGTM!

PR #7292, opened 2019-05-27
Remove not tested part from dropout.py

crcrpar:
I don't think the removed block is tested/used.
Also, the comment of "if we already have a dropout mask" seems obsolete/outdated because ~~cuDNN forward is never called with self.mask being None if `return_mask` is `True`~~ cuDNN forward is called only when mask==None.
 
@bonprosoft Do you remember any background/motivation?

### DIFF
--- a/chainer/functions/noise/dropout.py
+++ b/chainer/functions/noise/dropout.py
@@ -47,13 +47,6 @@ class Dropout(function_node.FunctionNode):
                 and x[0].flags.c_contiguous
                 and self.mask is None):
             self._use_cudnn = True
-
-            if hasattr(self, 'states'):
-                # if we already have a dropout mask,
-                # the forward operation is equal to backward.
-                return cuda.get_cudnn_dropout_states().backward(
-                    None, x[0], self.dropout_ratio, self.states),
-
             self.states, y = cuda.get_cudnn_dropout_states().forward(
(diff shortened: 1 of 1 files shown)


bonprosoft:
Thank you for mentioning.

> if we already have a dropout mask

In the comment, I mean "if there is a cuDNN state for the dropout instance, use the same one for next forwarding". I think the code is required to use the same dropout mask for another input array. The following code doesn't seem to work on your PR:
```py
import chainer
import cupy as cp
from chainer import functions, testing

x = cp.random.uniform(size=(5, 5))

with chainer.using_config('use_cudnn', 'always'):
    d = functions.noise.dropout.Dropout(0.5)
    y1, = d.forward_gpu(x)
    assert(d._use_cudnn)
    for i in range(10):
        y2, = d.forward_gpu(x)
        assert(cp.allclose(y1, y2))
```

I think it is tested in `check_immutable` test. Is it correct?
https://github.com/chainer/chainer/blob/65d10dd914ce423c38cf04f00857ef5f0f6252ec/tests/chainer_tests/functions_tests/noise_tests/test_dropout.py#L129-L137

crcrpar:
@bonprosoft 
Ah, I didn't understand how to use the feature at all and I feel it better to document it.
Therefore, I will close this PR and make a PR that adds documentation.

Thank you very much.

PR #7296, opened 2019-05-28
Support `None` arguments in `chainerx.clip` and `chainerx.ndarray.clip`

hvy:
A temporary fix (until https://github.com/chainer/chainer/pull/7285 is merged) that allows `chainerx.clip` and `chainerx.ndarray.clip` to take `None` interval arguments.

### DIFF
--- a/chainerx/math/misc.py
+++ b/chainerx/math/misc.py
@@ -30,10 +30,13 @@ def clip(a, a_min, a_max):
     .. seealso:: :func:`numpy.clip`
 
     """
-    if a_min is None:
-        a_min = a.min()
+    if a_min is None and a_max is None:
+        raise ValueError('Must set either a_min or a_max.')
 
-    if a_max is None:
-        a_max = a.max()
+    if a_min is not None:
+        a = chainerx.maximum(a, a_min)
 
-    return -chainerx.maximum(-chainerx.maximum(a, a_min), -a_max)
+    if a_max is not None:
+        a = chainerx.minimum(a, a_max)
+
+    return a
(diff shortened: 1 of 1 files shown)


niboshi:
What's the motivation of this PR?

hvy:
The PTB example using ChainerX, temporarily avoiding `unchain_backward`, reaching this code. https://github.com/chainer/chainer/blob/master/chainer/optimizer_hooks/gradient_clipping.py#L66 . It's also a NumPy compat issue.

chainerx/math/misc.py
```diff
@@ -30,4 +30,10 @@ def clip(a, a_min, a_max):
     .. seealso:: :func:`numpy.clip`
 
     """
+    if a_min is None:
+        a_min = a.min()
```
asi1024:
```suggestion
        return chainerx.minimum(a, a_max)
```

chainerx/math/misc.py
```diff
@@ -30,4 +30,10 @@ def clip(a, a_min, a_max):
...
+
+    if a_max is None:
+        a_max = a.max()
```
asi1024:
```suggestion
        return chainerx.maximum(a, a_mix)
```

chainerx/math/misc.py
hvy: Does this work when `a_max` is `None`?

chainerx/math/misc.py
hvy: Do you meant `a_min`?

asi1024:
Jenkins, test this please.

asi1024:
Jenkins failure seems unrelated to this PR.

asi1024:
LGTM.

PR #7270, opened 2019-05-23
Split reduction routines from math.cc

asi1024:
Part of #7143.
Depends on #7269.

PR #7261, opened 2019-05-23
Support bool dtypes in `chainerx::{Max,Min}imum`

asi1024:
This PR makes `chainerx::Maximum` and `chainerx::Minimum` accept Boolean dtypes like as `chainerx::Greater` and `chainerx::Less` do.

Issue #7143, opened 2019-05-15
math.cc is large


asi1024:
Split large routines/math.cc into several files.

- [x] arithmetic.cc #7415 
  - chainerx.negative
  - chainerx.add
  - chainerx.subtract
  - chainerx.multiply
  - chainerx.divide
  - chainerx.reciprocal

- [x] explog.cc #7317
  - chainerx.exp
  - chainerx.log
  - chainerx.log10

- [x] reduction.cc #7270
  - chainerx.sum
  - chainerx.logsumexp
  - chainerx.logsumexp
  - chainerx.log_softmax

- [x] trigonometric.cc #7171 
  - chainerx.sin
  - chainerx.cos
  - chainerx.tan
  - chainerx.arcsin
  - chainerx.arccos
  - chainerx.arctan
  - chainerx.arctan2

- [x] hyperbolic.cc #7171
  - chainerx.sinh
  - chainerx.cosh
  - chainerx.tanh
  - chainerx.arcsinh
  - chainerx.arccosh

- [x] rounding.cc #7407
  - chainerx.ceil
  - chainerx.floor

- [x] misc.cc #7298
  - chainerx.sqrt
  - chainerx.square
  - chainerx.squared_difference
  - chainerx.absolute
  - chainerx.fabs
  - chainerx.sign
  - chainerx.maximum
  - chainerx.minimum

- [x] binary.cc #7128 
  - chainerx.bitwise_and
  - chainerx.bitwise_or
  - chainerx.bitwise_xor

- [x] activation.cc #7496
  - chainerx.relu
  - chainerx.sigmoid

- [x] statistics.cc #7222 
  - chainerx.amin
  - chainerx.amax
  - chainerx.mean
  - chainerx.var

- [x] logic.cc #7444 
  - chainer.isnan
  - chainer.isinf
  - chainer.isfinite

hvy:
Finished!

PR #7298, opened 2019-05-28
Split misc routines from math.cc

asi1024:
Part of #7143.
~Depends on #7261.~
~Depends on #7270.~


chainerx_cc/chainerx/native/native_device/misc.cc
```diff
@@ -7,51 +7,27 @@
...
-        VisitFloatingPointDtype(out.dtype(), [&](auto pt) {
-            using T = typename decltype(pt)::type;
-            struct Impl {
-                void operator()(int64_t /*i*/, T x, T& out) { out = chainerx::Sqrt(x); }
-            };
-            Elementwise<const T, T>(Impl{}, x_cast, out);
-        });
-    }
-};
+CHAINERX_NATIVE_REGISTER_ELTWISE_FLOAT_UNARY_KERNEL(CeilKernel, { out = chainerx::Ceil(x); });
```
niboshi: I think `ceil` and `floor` should be in `rounding`.

asi1024: I'll work on it in another PR.

niboshi: Then why did you move this in this PR?

asi1024: These kernels were defined here before this PR. This PR just rewrites the definitions of kernels in `*_device/misc.cc` using macro. Should we revert them?

niboshi: OK, I understand, but it was difficult to notice in the review. Maybe it's better to separate the fix, or at least leave some comment.

chainerx_cc/chainerx/routines/misc.cc
```diff
@@ -0,0 +1,239 @@
...
+#include "chainerx/backward_builder.h"
+#include "chainerx/backward_context.h"
+#include "chainerx/dtype.h"
+#include "chainerx/error.h"
+#include "chainerx/graph.h"
+#include "chainerx/kernels/misc.h"
+#include "chainerx/macro.h"
+#include "chainerx/routines/creation.h"
+#include "chainerx/routines/logic.h"
+// #include "chainerx/routines/manipulation.h"
```
niboshi: Remove it

chainerx_cc/chainerx/routines/misc.cc
```diff
@@ -0,0 +1,239 @@
...
+#include "chainerx/routines/routines_util.h"
+// #include "chainerx/routines/statistics.h"
```
niboshi: Ditto

niboshi:
Jenkins, test this please

chainerx_cc/chainerx/routines/misc.cc
asi1024: Fixed.

niboshi:
Jenkins, test this please

PR #7306, opened 2019-05-28
Fix robots.txt to allow indexing root

kmaehashi:
ReadTheDocs returns 302 instead of 301 for the root, so it may be indexed by search engine.

```
$ curl --head https://docs.chainer.org/
HTTP/2 302
...
```

We should allow indexing it.


Crissman:
Jenkins, test this please.

kmaehashi:
pfnCI, test this please.

Crissman:
Jenkins failure looks unrelated.

Issue #7312, opened 2019-05-29
`optimizer.update()` fails when link has un-initialized parameter


corochann:
Testing under chainer==6.0.0 with chainerx built and cupy installed.

Code to reproduce
```
import numpy as np

import chainer
import chainer.links as L


class Model(chainer.Chain):

    def __init__(self):
        super(Model, self).__init__()
        with self.init_scope():
            self.l1 = L.Linear(10, 1)
            # l2 is not initialized
            self.l2 = L.Linear(None, 1)

    def forward(self, x):
        # l2 is not used, and never initialized.
        return self.l1(x)


# device = chainer.get_device(-1)  # When using numpy, below code works.
device = chainer.get_device('native')

model = Model()
model.to_device(device)

optimizer = chainer.optimizers.Adam()
optimizer.setup(model)

x = np.zeros((1, 10), dtype=np.float32)
x = device.send(x)
loss = model(x)
loss.backward()
optimizer.update()
```

```
# Traceback (most recent call last):
#   File "tmp3.py", line 34, in <module>
#     optimizer.update()
#   File "/home/user/.pyenv/versions/anaconda3-5.1.0/lib/python3.6/site-packages/chainer/optimizer.py", line 815, in update
#     param.update()
#   File "/home/user/.pyenv/versions/anaconda3-5.1.0/lib/python3.6/site-packages/chainer/variable.py", line 1877, in update
#     self.update_rule.update(self)
#   File "/home/user/.pyenv/versions/anaconda3-5.1.0/lib/python3.6/site-packages/chainer/optimizer.py", line 223, in update
#     self.update_core(param)
#   File "/home/user/.pyenv/versions/anaconda3-5.1.0/lib/python3.6/site-packages/chainer/optimizer.py", line 240, in update_core
#     self.update_core_chainerx(param)
#   File "/home/user/.pyenv/versions/anaconda3-5.1.0/lib/python3.6/site-packages/chainer/optimizer.py", line 283, in update_core_chainerx
#     backend_name = param.array.device.backend.name
# AttributeError: 'NoneType' object has no attribute 'device'
```

PR #7313, opened 2019-05-29
fix chainerx optimizer update error when parameter is not initialized

corochann:
resolves https://github.com/chainer/chainer/issues/7312.

### DIFF
--- a/chainer/optimizer.py
+++ b/chainer/optimizer.py
@@ -280,8 +280,6 @@ class UpdateRule(object):
 
         """
         grad_array = param.grad
-        if grad_array is None:
-            return
         backend_name = param.array.device.backend.name
         if backend_name not in ('native', 'cuda'):
             raise RuntimeError(
@@ -310,7 +308,9 @@ class UpdateRule(object):
         temp_param = variable.Variable._init_unchecked(
             param._chainerx_fallback_array, is_chainerx_array=False)
 
-        temp_param._set_grad_without_check(backend.from_chx(grad_array))
+        if grad_array is not None:
+            temp_param._set_grad_without_check(
+                backend.from_chx(grad_array))
 
         # Update
         self.update_core(temp_param)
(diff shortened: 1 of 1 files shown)


chainer/optimizer.py
```diff
@@ -308,9 +310,7 @@ def update_core_chainerx(self, param):
         temp_param = variable.Variable._init_unchecked(
             param._chainerx_fallback_array, is_chainerx_array=False)
 
-        if grad_array is not None:
```
corochann: I also refactored here, but I'm not sure what kind of case is considered originally.
Please review.

chainer/optimizer.py
```diff
@@ -280,6 +280,8 @@ def update_core_chainerx(self, param):
 
         """
         grad_array = param.grad
+        if grad_array is None:
+            return
```
niboshi: I think it's better to write the logic to skip the update at `Parameter.update()`, instead of here.
In that case, maybe you can write `assert param.array is not None` at the beginning of this method.

I also think this method should not early-return even if `grad_array is None`.
Mathematically that is equivalent to `grad == 0`. That might mean something for some optimizers.
`update_core` should be called even in that case.

niboshi:
Thank you for the PR! I left a comment.

corochann:
I added `param.array is None` check at `update` method.
If my understanding was wrong and if it is faster for you to implement by yourself, rather than discussing too much, it is also fine for me to just close this PR and let you fix this.

niboshi:
Thank you for the fix.

We discussed internally and now we think `update_core` should NOT be skipped in both cases (`param.array is None` (=uninitialized) and `param.grad is None`).
So, we should skip some lines in `UpdateRule.update_core_chainerx()` if `param.array is None`.
We also need unit tests.
Could you do that?

We're not in a hurry, but if you think you don't have time for this issue, please feel free to let us take over.

corochann:
I see. I think I could not consider all the test cases to cover by this change.
So I would like you to take over.

niboshi:
OK, I fixed the issue in #7492.
I'll close this one.

PR #7324, opened 2019-05-29
Add `model_device` and `input_device` arguments to `StandardUpdater`

### LINKED ISSUE #7322: Add to `StandardUpdater` an option to skip model transfer
Background: #7220

>* [ ] c. `StandardUpdater` transfers the model parameters, which causes device mismatch.
>   - Add `model_device` and `input_device` options to `StandardUpdater` (the existing `device` option will be an alias of the former), to specify the devices to transfer the model and the input data respectively. If `model_device` is omitted, it skips the model transfer. If `input_device` is omitted, `model_device` is used to transfer the data.

niboshi:
Fixes #7322

~Depends on #7326~

### DIFF
--- a/chainer/training/updaters/standard_updater.py
+++ b/chainer/training/updaters/standard_updater.py
@@ -1,15 +1,20 @@
 import six
 
+import chainer
 from chainer import backend
-from chainer.backends import cuda
 from chainer.dataset import convert
 from chainer.dataset import iterator as iterator_module
 from chainer.training import _updater
+from chainer.utils import argument
 
 
 class StandardUpdater(_updater.Updater):
 
-    """Standard implementation of Updater.
+    """StandardUpdater(\
+iterator, optimizer, converter=convert.concat_examples, model_device=None, \
+loss_func=None, loss_scale=None, auto_new_epoch=True, *, input_device=None)
+
+    Standard implementation of Updater.
 
     This is the standard implementation of :class:`~chainer.training.Updater`.
     It accepts one or more training datasets and one or more optimizers.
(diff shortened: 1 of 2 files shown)


chainer/training/updaters/standard_updater.py
```diff
@@ -32,8 +37,10 @@ class StandardUpdater(_updater.Updater):
             extracted by the main iterator and the ``device`` option are passed
             to this function. :func:`chainer.dataset.concat_examples` is used
             by default.
-        device: Device to which the training data is sent. Negative value
-            indicates the host memory (CPU).
+        model_device (device specifier):
```
hvy: I'm wondering if `link_device` ("link" as in `optimizer.setup`) or `target_device` ("target" as in optimizer) don't include more familiar terminologies? "model" is in my understanding a convention among Chainer-users to name their links but maybe it's used in Chainer too.

chainer/training/updaters/standard_updater.py
```diff
@@ -62,13 +76,51 @@ class StandardUpdater(_updater.Updater):
...
+
     """
 
     def __init__(self, iterator, optimizer, converter=convert.concat_examples,
                  device=None, loss_func=None, loss_scale=None,
-                 auto_new_epoch=True):
+                 auto_new_epoch=True, **kwargs):
+        # `device` argument is for backward compatibility.
+        # If it's given (either positional or keyword), and the device(s) are
+        # GpuDevice, StandardUpdater skips GPU-to-GPU transfer for the model.
```
hvy: It's probably better to elaborate on what we mean by the plural devices here. You mean the device argument and the device on which the model already resides?

niboshi:
Replaced with  #7472

Issue #7305, opened 2019-05-28
Strict scatter_dataset and more flexible multinode evaluator


kuenishi:
The rationale behind https://github.com/chainer/chainercv/pull/868 is

- Current `scatter_dataset` implicitly appends duplicate data for odd number of test dataset to communicatgor size
- Some evaluators does not allow naive summation or average of results of scattered test data

The case above is calculating standard mAP in object detection but not easy with current MultiNodeEvaluator.

We have to understand more about above case and other cases such as object segmentation, NLP and so on. RL is not using Chainer's trainer scheme so far, could be out of scope.

iwiwi:
FYI: duplicate with the former one https://github.com/chainer/chainermn/issues/274

kuenishi:
I want to discuss the design with ...
- WIP branch: https://github.com/chainer/chainer/compare/master...kuenishi:strict-multinode-evaluator
- ChainerCV's example: https://gist.github.com/kuenishi/fa20a88881beffd3cccbd6c15d62522e

On...
- converter, eval_func, aggreation funcs are to be injected as function
- Design simplicity and level of abstraction (too close to ChainerCV's apply_to_iterator?)

kuenishi:
After offline discussion:
- inheritance style is more preferred than plugging callable objects on callbacks
- no need for generator; debuggable is preferred and give a list to aggregator
- ProgressHook to be given from users
- naming: eval_func & aggregate -> calc_local & aggregate

kuenishi:
Additional requiest:

- Hope it works in non-MPI env dummy-like single proc communicator or `comm=None`

PR #7289, opened 2019-05-27
Fix a bug that root is ignored in  scatter_dataset and bcast

keisukefukuda:
This PR fixes a critical bug, which is related to `scatter_dataset` and `bcast` routines.

`scatter_dataset()` has a parameter called `root`. It is expected that `root` is passed to the underlying MPI collective routines, in particular `Bcast()`. However, the routines are actually hard-coded so the `root` is always rank 0 and the `root` parameter is ignored. The test did not detect the bug.

The PR fixes the bug in two steps. 
First, the test is fixed. The problem in `check_scatter_dataset`  was that all ranks have the same data when the test begins, so if the `root` argument is ignored, rank 0 performed `bcast` and all process obtained the data so the test succeeded.

Second, the code is fixed.

PR #7327, opened 2019-05-29
Add strict mode to scatter_dataset as well as scatter_index

kuenishi:
This PR addresses first half of #7305  by introducing strict mode to existing `scatter_dataset` as well as `scatter_index` to prevent broadcasting underlying dataset. Fix in #7289 is included.


keisukefukuda:
pfnci, test this please

keisukefukuda:
pfnci, test this please.

chainermn/datasets/scatter.py
```diff
@@ -33,6 +34,8 @@ def scatter_dataset(dataset, comm, root=0, shuffle=False,
             If ``None``, the permutation is changed randomly.
         max_buf_len (int): Max buffer size to be used at broadcasting
             binaries. Must not be larger than 2147483647.
+        strict (bool): Strictly scatters the index and avoid data duplication
```
keisukefukuda: I guess we need to document what happens when `len(dataset) % comm.size != 0`. 

The reason why the previous `scatter_dataset()` allows duplication is that all processes must have the same number of samples to keep the iteration count correct.  I'm not 100% sure, but using `strict` option comes with a cost of breaking the restriction ?

kuenishi: Updated the document.

keisukefukuda:
pfnci, test this please.

iwiwi:
I don't think `strict` is descriptive enough. How about something like `force_equal_lengths` ?

kuenishi:
Then how about `allow_duplicates` ? I think it's more natural in terms of phrasing software concepts.

keisukefukuda:
+1 for `force_equal_length` or `force_even_split` .

Also, I'm afraid that the docstring is still not enough. I think it should mention *why* and *when* it should be used, and why the default value is `False` (or `True`).

kuenishi:
@keisukefukuda ping

keisukefukuda:
pfnci, test this please.

kuenishi:
jenkins, test this please.

keisukefukuda:
LGTM

PR #7328, opened 2019-05-29
Fix imagenet example to use DALI with ChainerX

anaruse:
This PR makes it available to use DALI with ChainerX in imagenet example.


niboshi:
Awaiting #7489.

kmaehashi:
#7489 has been merged.

